PR #17, opened 2014-11-14
Added aliases to commands so they can be referenced by their short name

jasdel:
This is based on gdb's command list of short names for commands.

Commands aliases were added for:
- break: b
- continue: c
- step: si
- next: n
- print: p

Also updated .gitignore to ignore built dlv file in cmd/dlv


Wphelp98737:
بنام خدا و با سلام خدمت شما 
لطفا مرا کمک و راهنمایی بفرمائید، متشکرم

PR #21, opened 2014-11-14
Basic command/function completion

ebfe:
This is more of a RFC/experiment and probably a bit rough around the edges. It
switches from the goreadline binding to golang.org/x/crypto/ssh/terminal for
terminal handling and implements basic command and function name completion
(for `clear` and `break`).


derekparker:
Going to close this out - I think we may want to just switch to a pure Go readline, and maybe make some modifications to allow command completion and so on.

PR #39, opened 2015-01-01
Better handling of struct pointers

epipho:
- `print structptr.member` now works
- attempting to print a pointer variable that is nil will no longer panic
- attempting to print a member variable of a nil struct will no longer panic
- added a bunch more tests for error conditions

### DIFF
--- a/proctl/variables.go
+++ b/proctl/variables.go
@@ -394,7 +394,7 @@ func findDwarfEntry(name string, reader *dwarf.Reader, member bool) (*dwarf.Entr
 }
 
 func (thread *ThreadContext) evaluateStructMember(parentEntry *dwarf.Entry, reader *reader.Reader, memberName string) (*Variable, error) {
-	parentInstr, err := instructionsForEntry(parentEntry)
+	parentAddr, err := thread.extractVariableDataAddress(parentEntry, reader)
 	if err != nil {
 		return nil, err
 	}
@@ -406,7 +406,7 @@ func (thread *ThreadContext) evaluateStructMember(parentEntry *dwarf.Entry, read
 	}
 
 	// Seek reader to the type information so members can be iterated
-	_, err = reader.SeekToType(parentEntry, true, false)
+	_, err = reader.SeekToType(parentEntry, true, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


proctl/variables_test.go
```diff
@@ -35,26 +37,31 @@ func TestVariableEvaluation(t *testing.T) {
...
+		{"i8", "1", "int8", nil},
+		{"f32", "1.2", "float32", nil},
+		{"a6.Baz", "8", "int", nil},
+		{"a7.Baz", "5", "int", nil},
+		{"a8.Baz", "feh", "struct string", nil},
+		{"a9.Baz", "nil", "int", errors.New("a9 is nil")},
+		{"a9.NonExistent", "nil", "int", errors.New("a9 has no member NonExistent")},
+		{"a8", "main.FooBar2 {Bur: 10, Baz: feh}", "main.FooBar2", nil}, // reread variable after member
+		{"i32", "[2]int32 [1 2]", "[2]int32", nil},
+		{"NonExistent", "", "", errors.New{"could not find symbol value for NonExistent"}}
```
derekparker: Missing ending comma here, won't compile.

proctl/variables.go
```diff
@@ -490,27 +498,71 @@ func (thread *ThreadContext) extractVariableFromEntry(entry *dwarf.Entry) (*Vari
...
+}
 
-		fctx := fde.EstablishFrame(regs.PC())
-		cfaOffset := fctx.CFAOffset() + int64(regs.SP())
+// Extracts the value from the instructions given in the DW_AT_location entry.
+// We execute the stack program described in the DW_OP_* instruction stream, and
+// then grab the value from the other processes memory.
+func (thread *ThreadContext) extractValue(instructions []byte, addr int64, typ interface{}) (string, error) {
+	var err error
+	address := addr
```
derekparker: Might as well just use `addr` here. It isn't used after being assigned to the local, so we shouldn't have to preserve it.

proctl/variables_test.go
epipho: Whoops, had that fixed locally and didn't commit. *hangs head in shame*

derekparker:
Thanks, if you don't mind squashing this, I'll merge it in.

epipho:
Squashed.

derekparker:
Merged.

PR #40, opened 2015-01-01
Breakpoints now have ids. Consolidated location lookup logic

epipho:
- Gave breakpoints a monotonically increasing id (similar to gdb) for easier manipulation. 
- Centralized looking up address by location (file+line, function) to DebuggedProcess.FindLocation
- FindLocation supports arbitrary addresses and breakpoint ids
- Added BreakByLocation and ClearByLocation to DebuggedProcess and changed the commands to call them instead.
- BreakByLocation can now also use an address `b 0x400d60` will work
- ClearByLocation can take a breakpoint id or address (as well as the existing file + line and function)

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -4,14 +4,11 @@ package command
 
 import (
 	"bufio"
-	"debug/gosym"
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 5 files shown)


epipho:
I also unified a few uintptr vs uint64 types for addresses. Ended up going toward uint64 as that is what the golang libraries use.

PR #42, opened 2015-01-14
Add process control for Linux x86

malix0:
Hi Derek,

I need to run Delve under Linux x86 (32 bit). So I just made very little change to you code. I jusr create a symlink proctl_linux_x86.go -> proctl_linux_amd64.go and copy threads_linux_amd64.go to threads_linux_x86.go. In the last file I only chenge the name of the register. I know that this is just a dirty fix but it works.

Sincerly
Massimo

### DIFF
--- a/proctl/proctl_linux_x86.go
+++ b/proctl/proctl_linux_x86.go
@@ -1,0 +1,1 @@
+proctl_linux_amd64.go
--- a/proctl/threads_linux_x86.go
+++ b/proctl/threads_linux_x86.go
@@ -1,0 +1,37 @@
+package proctl
+
+import "syscall"
+
+type Regs struct {
+	regs *syscall.PtraceRegs
+}
+
+func (r *Regs) PC() uint64 {
+	return r.regs.PC()
+}
+
+func (r *Regs) SP() uint64 {
+	return uint64(r.regs.Esp)
+}
+
+func (r *Regs) SetPC(tid int, pc uint64) error {
+	r.regs.SetPC(pc)
+	return syscall.PtraceSetRegs(tid, r.regs)
+}
+
+func registers(tid int) (Registers, error) {
+	var regs syscall.PtraceRegs
+	err := syscall.PtraceGetRegs(tid, &regs)
+	if err != nil {
+		return nil, err
+	}
+	return &Regs{&regs}, nil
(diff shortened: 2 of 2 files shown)


derekparker:
Hmm.. do all of the tests pass?

If they do, this commit will need to be reworked, but I'll merge it in after feedback & testing it out myself on a 32 bit machine.

malix0:
Hi Derek,

 I think I was too hasty with the pull request. The tests effectively does
not work. But I want to fix it. Can you recommend some links where I
can understand
how the program debugging works under linux 32-bit?

Thanks and excuse me.

PS: It's really curious that in less than a month, you're the second person
of Chicago with which I have ever spoken in my entire life, and both of
your names is Derek. As I said to him, excuse me, but I speak english like
I eat maccaroni. :-)

2015-01-14 19:05 GMT+01:00 Derek Parker <notifications@github.com>:

derekparker:
So, there used to be a bunch of places in `proctl/variables.go` that assumed 8 byte pointers, however most if not all of that should be gone there. I know there are still assumptions about register naming, everything assumes 64 bit registers, so that will have to be reworked. Other than that, I would just run the tests and fix whatever is broken.

There may be other places where pointer size is assumed, and it could be that stack frame setup in `dwarf/frame` may be making some assumptions.

If you have any questions please let me know, I'd really like to help you work on this if you're interested.

malix0:
Hi Derek,

I'm reading DWARF 2.0 specification (
http://www.dwarfstd.org/doc/dwarf-2.0.0.pdf) Because I need to understand
how debugger works. I try to follow your indication and try to run the test
but I get error accessing the registers "could not determine kind of name
for C.DR_LEN_8". Indeed inside debugreg.h it's defined only for 64bit
system.

#ifdef __x86_64__
# define DR_LEN_8 (0x8)
#endif

Where I can find information about register?

Thanks in advance
Massimo




PS: This may be interesting http://ernestmicklei.com/tag/golang/

2015-01-16 21:52 GMT+01:00 Derek Parker <notifications@github.com>:

derekparker:
@malix0 that last issue to mentioned should be fixed on master.

derekparker:
Going to close this PR due to lack of activity.

PR #44, opened 2015-01-20
Fix panic in readstring

epipho:
- Fixes panic if trying to read a string over 16 bytes
- readString now works on sliced strings correctly
- Added tests for both


derekparker:
Thanks!

PR #46, opened 2015-01-20
Better handling of arrays and slices

epipho:
Built on #45 
- Generalized arrays and slices, print/eval supports arrays and slices of all types.


derekparker:
This is great, thanks!

I have an idea for an optimization, where the entire array is read into memory and then the values are parsed from that. The issue is there are potential edge cases with pointers where memory could have to be read anyways for dereferencing, so I'll open an issue to keep this in mind.

PR #47, opened 2015-01-21
info vars command

epipho:
The last of the info commands, prints out all package level variables and values.  With this code, running info vars after starting the testvariables does not show any unparseable or unreadable variables! Built on #45 and #46 

- Adds support for UndefinedType which will always have a value of (unknown)
- Adds support for VoidType which will always have a value of (void)
- Fixes reading null fn pointers.
- When a TypedefType is found, recurse down until finding something that is not a typedef, some types are typedefs of typedefs etc.

The code is complete but I found a bug during testing. For certain data structures that are very deep (linked lists) and structures that are large and have many nested structs, extractVariableFromEntry goes crazy continually recursively reading nested structs until the application panics and runs out of memory.

We should probably stop recursing after N levels (2?). What should we display in that case? Perhaps ... to indicate there is more there, but we can't display it. Once proper handling of evalVariable for package variables is complete the user could dig in that way.

### DIFF
--- a/dwarf/reader/reader.go
+++ b/dwarf/reader/reader.go
@@ -10,18 +10,18 @@ type Reader struct {
 	depth int
 }
 
-// New returns a reader for the specified dwarf data
+// New returns a reader for the specified dwarf data.
 func New(data *dwarf.Data) *Reader {
 	return &Reader{data.Reader(), 0}
 }
 
-// Seek moves the reader to an arbitrary offset
+// Seek moves the reader to an arbitrary offset.
 func (reader *Reader) Seek(off dwarf.Offset) {
 	reader.depth = 0
 	reader.Reader.Seek(off)
 }
 
-// SeekToEntry moves the reader to an arbitrary entry
+// SeekToEntry moves the reader to an arbitrary entry.
 func (reader *Reader) SeekToEntry(entry *dwarf.Entry) error {
 	reader.Seek(entry.Offset)
 	// Consume the current entry so .Next works as intended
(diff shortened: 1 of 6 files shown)


derekparker:
Yeah, that's a good point and I'm on board with your proposed solution. That could even extend to other large types, i.e. huge arrays, instead of just blasting those onto the screen, print the first N and then shorten it with a `...`.

derekparker:
@epipho did you want me to merge this in and then work on the large data structure issue separately or did you want me to wait to merge this until that fix is in?

epipho:
I would wait as there is a chance using the command causes OOM.  I am in
the process of digging out of snowpacalypse here so I will take care of it
in the next few days.

derekparker:
Sounds good. You're on the east coast? Good luck with that blizzard! Hope you're not too badly affected.

epipho:
Added a limit of 2 levels of recursion for structs. Once reached, any structs will print {...} instead of their list of fields.
Added a 64 element limit to printing arrays. Adds ...+x more  to the end of the print statement.

Added tests for both.

dwarf/reader/reader.go
```diff
@@ -144,3 +144,28 @@ func (reader *Reader) NextMemberVariable() (*dwarf.Entry, error) {
 	// No more items
 	return nil, nil
 }
+
+// NextPackage moves the reader to the next debug entry that describes a package variable
```
derekparker: Change to `NextPackageVariable`, and add a period at the end (nit-picky I know, sorry).

proctl/variables.go
```diff
@@ -492,6 +497,7 @@ func (thread *ThreadContext) extractVariableFromEntry(entry *dwarf.Entry) (*Vari
 
 	val, err := thread.extractValue(instructions, 0, t, true)
 	if err != nil {
+		fmt.Printf("NAME: %s\n", n)
```
derekparker: Accidentally committed?

proctl/variables.go
```diff
@@ -897,6 +934,28 @@ func (thread *ThreadContext) FunctionArguments() ([]*Variable, error) {
 	return thread.variablesByTag(dwarf.TagFormalParameter)
 }
 
+// PackageVariables returns the name, value, and type of all package variables in the application
```
derekparker: Could you add a period to the end of the comment, also I tend to like to group exported methods together, so maybe move this method as well as `LocalVariables` and `FunctionArguments` up a bit in the file. (Again, nit-picky I know but I feel it keeps things a bit more organized).

derekparker:
Just some minor nitpicky things but otherwise looks good.

epipho:
Cleaned up. No worries about nitpicky stuff, better code in the long run.

derekparker:
Merged.

PR #48, opened 2015-01-21
Dont panic if .debug_frame section is not found.

epipho:
Attempting to debug a process with no symbols causes dlv to panic because .debug_frame section is missing. Fixed up logic to log + exit instead.

Last PR until you get a chance to review everything, thanks for all your work on this.

### DIFF
--- a/proctl/proctl_linux_amd64.go
+++ b/proctl/proctl_linux_amd64.go
@@ -109,13 +109,17 @@ func (dbp *DebuggedProcess) findExecutable() (*elf.File, error) {
 func (dbp *DebuggedProcess) parseDebugFrame(exe *elf.File, wg *sync.WaitGroup) {
 	defer wg.Done()
 
-	debugFrame, err := exe.Section(".debug_frame").Data()
-	if err != nil {
-		fmt.Println("could not get .debug_frame section", err)
+	if sec := exe.Section(".debug_frame"); sec != nil {
+		debugFrame, err := exe.Section(".debug_frame").Data()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


proctl/proctl_linux_amd64.go
```diff
@@ -109,13 +109,17 @@ func (dbp *DebuggedProcess) findExecutable() (*elf.File, error) {
...
-		fmt.Println("could not get .debug_frame section", err)
+	if sec := exe.Section(".debug_frame"); sec != nil {
+		debugFrame, err := exe.Section(".debug_frame").Data()
+		if err != nil {
+			fmt.Println("could not get .debug_frame section", err)
+			os.Exit(1)
+		}
+		dbp.FrameEntries = frame.Parse(debugFrame)
+	} else {
+		fmt.Println("No debug symbols found")
```
derekparker: Could you change this to something like "Error: Executable does not contain .debug_frame section" or something along those lines, just to make it more clear exactly why we're bailing.

derekparker:
Yeah, I like this. I think I ran into this in the wild a while ago, when I first started testing Delve on different types of binaries, but I can't remember what caused the `.debug_frame` section to be missing.

I'm curious what kind of binary you were trying to debug which caused this?

epipho:
I was testing the new info vars command and I tried to load a compiled version of docker as I had it on hand.  You should be able to repro the crash by compiling with -s to omit debug info.

derekparker:
Cool, makes sense. I'll merge this in.

derekparker:
Merged.

Issue #43, opened 2015-01-20
Error running tests


epipho:
Tests are failing on master and I have not yet found a cause.

```
new thread spawned 1476
new thread spawned 1477
new thread spawned 1487
new thread spawned 1488
Hello, World!
new thread spawned 1497
new thread spawned 1498
new thread spawned 1508
new thread spawned 1534
new thread spawned 1535
new thread spawned 1536
foo
Hello, World!
foo
foo
Hello, World!
new thread spawned 1545
new thread spawned 1555
new thread spawned 1556
new thread spawned 1565
--- FAIL: TestVariableFunctionScoping (1.22s)
        proctl_test.go:52: Continue() returned an error : could not get registers could not get registers no such process
new thread spawned 1580
new thread spawned 1581
FAIL
FAIL    github.com/derekparker/delve/proctl     10.588s
```

derekparker:
Yeah, it seems to be a flickering failure but it's definitely there. I'll look into it, I may have an idea why this is happening.

derekparker:
So, for some reason on my laptop (Debian Jessie running kernel 3.17) the tests pass fine 99% of the time, with occasional, but rare flickers. However, on an Ubuntu 14.10 vm I have the tests always seem to fail consistently on master.

I'm going to investigate this more and try and figure out what is causing master to fail on certain machines, and under what circumstances.

epipho:
It seems to always fail on both VMs for me (Ubuntu 14.04 and CoreOS 561)

derekparker:
So, on the ubuntu VM that reproducibly fails, I have `TestNext` and `TestVariableFunctionScoping` failing consistently. I have tracked down why `TestNext` is failing, it's due to `m.procid` not being set correctly for a newly created thread. No idea why this is happening, I'm running go 1.4.1 on both my actual machine and on the VM and the VM fails whereas on my machine all tests pass with no problem.

Haven't dug into the `TestVariableFunctionScoping` failure yet, but it could be related.

pnasrat:
I just tried on GCE on Operating System: CoreOS 575.0.0

@epipho what are you using as a host VM?

PR #49, opened 2015-01-21
Report caller file:line in assert helper fn when test fails

ivarg:
Hi,
I ran into issue #43 and noticed how the error output referenced the assert helper function instead of the actual caller. Here is a suggested fix to make it a bit easier to trace the origin, with the additional cost of importing fmt.

### DIFF
--- a/proctl/proctl_test.go
+++ b/proctl/proctl_test.go
@@ -49,7 +49,9 @@ func dataAtAddr(pid int, addr uint64) ([]byte, error) {
 
 func assertNoError(err error, t *testing.T, s string) {
 	if err != nil {
-		t.Fatal(s, ":", err)
+		_, file, line, _ := runtime.Caller(1)
+		fname := filepath.Base(file)
+		t.Fatalf("failed assertion at %s:%d: %s : %s\n", fname, line, s, err)
 	}
 }
 


proctl/proctl_test.go
```diff
@@ -49,7 +50,10 @@ func dataAtAddr(pid int, addr uint64) ([]byte, error) {
 
 func assertNoError(err error, t *testing.T, s string) {
 	if err != nil {
-		t.Fatal(s, ":", err)
+		_, file, line, _ := runtime.Caller(1)
+		fname := filepath.Base(file)
+		fmt.Printf("\t%s:%d: %s : %s\n", fname, line, s, err)
+		t.FailNow()
```
derekparker: I would change this to `t.Fatalf()` and then use your formatted string you're currently passing into Printf. Also, preface it with something like "failed assertion at:" or something along those lines.

derekparker:
Looks good, just a minor nitpick and then I'll merge it in.

ivarg:
Ok, so I must admit this my first github pull request. I made the requested changes, and to avoid clutter I amended the one commit and forced pushed to my branch. However, I'm not sure this is the preferred workflow. I don't see the change in my github fork, and I begin to suspect force-pushing an amended commit was not the right thing to do.

derekparker:
Hmmm.. that shouldn't cause a problem. Not sure what's happening there, GitHub should pick up the forced push commit.

ivarg:
Now it seems to have been picked up. So I guess I'll stick with the workflow.

derekparker:
Merged, thanks!

PR #50, opened 2015-01-22
proctl: propagate underlying error for failing hardware breakpoint.

pnasrat:
Improves error message on systems that delve is failing on.


derekparker:
Thanks!

PR #52, opened 2015-01-24
Some hw breakpoint fixes

### LINKED ISSUE #51: Running tests: could not set hardware breakpoint: invalid argument
I'm happy to try figure this out.
Steps to reproduce - I'm using GCE to run in on centos 7 image.

1. cd proctl
2. go test -c .
3. ./proctl.test -test.run=TestStep

Output:

```
--- FAIL: TestStep (0.17s)
	proctl_test.go:52: Break() : could not set hardware breakpoint: invalid argument
FAIL
```

```
$ cat /etc/redhat-release
CentOS Linux release 7.0.1406 (Core)
$ rpm -q glibc
glibc-2.17-55.el7_0.3.x86_64
$ uname -a
Linux delve01 3.10.0-123.13.1.el7.x86_64 #1 SMP Tue Dec 9 23:06:09 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
$ go version
go version go1.4.1 linux/amd64
# at commit 46e5cfd39963a489f6e4fc43ad3a9f78b9f5321b
```

```
strace -o trace.out -s 2048   ./proctl.test -test.run=TestStep
grep ptrace trace.out
ptrace(PTRACE_SETOPTIONS, 11858, 0, PTRACE_O_TRACECLONE) = 0
ptrace(PTRACE_POKEUSER, 11858, offsetof(struct user, u_debugreg), 0x400c30) = 0
ptrace(PTRACE_POKEUSER, 11858, offsetof(struct user, u_debugreg) + 56, 0x9) = -1 EINVAL (Invalid argument)
```

Using gdb I can see

```
(gdb) break breakpoints_linux_amd64.go:155
(gdb) run
Breakpoint 1, github.com/derekparker/delve/proctl.setHardwareBreakpoint (reg=0, tid=13764, addr=4197424, ~r3=...)
    at /home/pnasrat/go/src/github.com/derekparker/delve/proctl/breakpoints_linux_amd64.go:155
155		return PtracePokeUser(tid, dr7addr, dr7)
(gdb) p dr7
$1 = 9
(gdb) p dr7addr
$2 = 904
```

ebfe:
Enable usage of dr1-dr3.  Clear control bits when a breakpoint
is disabled. Use DR_LEN_1 instead of DR_LEN_8 so breakpoint work on
unaligned adresses.

Fixes #51.

### DIFF
--- a/proctl/breakpoints_linux_amd64.go
+++ b/proctl/breakpoints_linux_amd64.go
@@ -52,12 +52,12 @@ func PtracePokeUser(tid int, off, addr uintptr) error {
 }
 
 func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
-	var x uintptr // XXX: this should not be necessary
-	ret, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_PEEKUSR, uintptr(tid), uintptr(off), uintptr(unsafe.Pointer(&x)), 0, 0)
+	var val uintptr
+	_, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_PEEKUSR, uintptr(tid), uintptr(off), uintptr(unsafe.Pointer(&val)), 0, 0)
 	if err != syscall.Errno(0) {
 		return 0, err
 	}
-	return ret, nil
+	return val, nil
 }
 
 func (dbp *DebuggedProcess) BreakpointExists(addr uint64) bool {
--- a/proctl/breakpoints_linux_amd64.go
+++ b/proctl/breakpoints_linux_amd64.go
@@ -17,6 +17,7 @@ import "C"
 import (
 	"fmt"
 	"syscall"
+	"unsafe"
 )
 
 // Represents a single breakpoint. Stores information on the break
@@ -48,6 +49,15 @@ func PtracePokeUser(tid int, off, addr uintptr) error {
 		return err
 	}
 	return nil
+}
+
+func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
+	var x uintptr // XXX: this should not be necessary
+	ret, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_PEEKUSR, uintptr(tid), uintptr(off), uintptr(unsafe.Pointer(&x)), 0, 0)
+	if err != syscall.Errno(0) {
+		return 0, err
+	}
+	return ret, nil
 }
 
 func (dbp *DebuggedProcess) BreakpointExists(addr uint64) bool {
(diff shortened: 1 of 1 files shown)


proctl/breakpoints_linux_amd64.go
```diff
@@ -50,6 +51,15 @@ func PtracePokeUser(tid int, off, addr uintptr) error {
 	return nil
 }
 
+func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
+	var x uintptr // XXX: this should not be necessary
+	ret, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_PEEKUSR, uintptr(tid), uintptr(off), uintptr(unsafe.Pointer(&x)), 0, 0)
```
derekparker: `ret` will never be anything of interest to us here. The actual value from this ptrace request will be stored in `x`. The ptrace API is to provide an address and data, and for reads (PEEK) the data stored in the requested address will be placed in the data argument, which here would be `x`.

`ret` will be 0 every time, whereas `x` will hold to value of interest.

proctl/breakpoints_linux_amd64.go
```diff
@@ -133,24 +143,49 @@ func (dbp *DebuggedProcess) newBreakpoint(fn, f string, l int, addr uint64, data
 // that we want to break at. There are only 4 debug registers
 // DR0-DR3. Debug register 7 is the control register.
 func setHardwareBreakpoint(reg, tid int, addr uint64) error {
-	if reg < 0 || reg > 7 {
+	if reg < 0 || reg > 3 {
```
derekparker: Yeah, this is good. Don't know what I was thinking here... out of the 7 debug regs only the 4 are used for actually storing the address.

proctl/breakpoints_linux_amd64.go
ebfe: On Sun, Jan 25, 2015 at 12:32:55PM -0800, Derek Parker wrote:
> > @@ -50,6 +51,15 @@ func PtracePokeUser(tid int, off, addr uintptr) error {
> >  	return nil
> >  }
> >  
> > +func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
> > +	var x uintptr // XXX: this should not be necessary
> > +	ret, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_PEEKUSR, uintptr(tid), uintptr(off), uintptr(unsafe.Pointer(&x)), 0, 0)
> 
> `ret` will never be anything of interest to us here. The actual value
> from this ptrace request will be stored in `x`. The ptrace API is to
> provide an address and data, and for reads (PEEK) the data stored in
> the requested address will be placed in the data argument, which here
> would be `x`.
> 
> `ret` will be 0 every time, whereas `x` will hold to value of
> interest.

Indeed. The man page apparently documents the ABI of the glibc wrapper.

Fixed.

derekparker:
Thanks!

PR #53, opened 2015-01-25
migrate to go.sys subrepository

tones111:
As of go version 1.4 the standard library syscall package is "locked
down" and code outside of the standard library is recommended to migrate
to the go.sys subrepository.

Reference: https://golang.org/s/go1.4-syscall

### DIFF
--- a/proctl/breakpoints_linux_amd64.go
+++ b/proctl/breakpoints_linux_amd64.go
@@ -16,6 +16,7 @@ import "C"
 
 import (
 	"fmt"
+	sys "golang.org/x/sys/unix"
 	"syscall"
 )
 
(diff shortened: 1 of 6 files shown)


proctl/breakpoints_linux_amd64.go
```diff
@@ -43,7 +44,7 @@ func (bpe BreakPointExistsError) Error() string {
 }
 
 func PtracePokeUser(tid int, off, addr uintptr) error {
-	_, _, err := syscall.Syscall6(syscall.SYS_PTRACE, syscall.PTRACE_POKEUSR, uintptr(tid), uintptr(off), uintptr(addr), 0, 0)
+	_, _, err := sys.Syscall6(sys.SYS_PTRACE, sys.PTRACE_POKEUSR, uintptr(tid), uintptr(off), uintptr(addr), 0, 0)
 	if err != syscall.Errno(0) {
```
derekparker: Why not migrate this line as well?

tones111:
Unfortunately, the return type of the last variable of sys.Syscall6 is still a syscall.Errno, so there appear to still be some dependencies on the standard syscall package.
https://godoc.org/golang.org/x/sys/unix#Syscall6

  There was the same problem with use of syscall.SysProcAttr

derekparker:
@tones111 ah, gotcha. That makes sense.

derekparker:
Merged, thanks!

PR #54, opened 2015-01-26
Fixed missing header file "sys/types.h" to fix compiler error

dsouzae:
On my installed system (cento 6.6) The header file "sys/user.h" requires "sys/types.h" to be included first.


derekparker:
Merged, thanks!

PR #57, opened 2015-02-04
[WIP] Websockets support (fixes #22)

### LINKED ISSUE #22: IDE interfaces
Hi @derekparker,

First of all, Thank you for this!

While I haven't played with it yet, I plan to do so this weekend.

If all things work good enough, I'd be interesting in integrating it with the [Go plugin for IntelliJ IDEA](https://github.com/go-lang-plugin-org/go-lang-idea-plugin).

As I haven't seen anything that could help out IDEs do the interfacing part so I'd like to ask if there's any plans for this. I could potentially help out by defining an HTTP interface for it and just pipe all the commands to the debugger via HTTP calls / JSON responses. A side effect of this, then it could allow creating web interfaces for this.

What do you think? Does it make sense to you? If you'd like to chat, please drop me a mention [on Gitter](https://gitter.im/go-lang-plugin-org/go-lang-idea-plugin).

Kind regards

dlsniper:
Hi,

I've had a chat earlier with @tarqe39 and he had the idea that to do something to the first commit in this PR.
I'm very open to discussion and suggestions on how to move this forward.

I guess the next steps are:

- [x] get the output from cmd as currently the API is ```type cmdfunc func(proc *proctl.DebuggedProcess, args ...string) error``` (maybe add a *string/[]byte next to error?)
- [ ] see if we need to wrap the output to json
- [x] see if we need to wrap the input to json
- [ ] add tests? (looking for volunteers :D)
- [ ] have this merged

P.S.
It took a while but I've manage to get some time off and now I can work a bit on getting this up and running. Sorry for delay.
If you want to change something, and have the credits for the change in the git history, send a PR to my fork, I have all the intentions to get this merged when it's in a good state.

[Review on Reviewable](https://reviewable.io/reviews/derekparker/delve/57)

### DIFF
--- a/client/cli/cli.go
+++ b/client/cli/cli.go
@@ -10,6 +10,7 @@ import (
 
 	sys "golang.org/x/sys/unix"
 
+	. "github.com/derekparker/delve/client/internal/common"
 	"github.com/derekparker/delve/command"
 	"github.com/derekparker/delve/goreadline"
 	"github.com/derekparker/delve/proctl"
@@ -29,23 +30,23 @@ func Run(run bool, pid int, args []string) {
 		cmd := exec.Command("go", "build", "-o", debugname, "-gcflags", "-N -l")
 		err := cmd.Run()
 		if err != nil {
-			die(1, "Could not compile program:", err)
+			Die(1, "Could not compile program:", err)
 		}
 		defer os.Remove(debugname)
 
 		dbp, err = proctl.Launch(append([]string{"./" + debugname}, args...))
 		if err != nil {
-			die(1, "Could not launch program:", err)
+			Die(1, "Could not launch program:", err)
 		}
 	case pid != 0:
 		dbp, err = proctl.Attach(pid)
 		if err != nil {
-			die(1, "Could not attach to process:", err)
+			Die(1, "Could not attach to process:", err)
 		}
 	default:
 		dbp, err = proctl.Launch(args)
 		if err != nil {
-			die(1, "Could not launch program:", err)
+			Die(1, "Could not launch program:", err)
 		}
 	}
 
@@ -69,10 +70,10 @@ func Run(run bool, pid int, args []string) {
 			if err == io.EOF {
 				handleExit(dbp, 0)
 			}
-			die(1, "Prompt for input failed.\n")
+			Die(1, "Prompt for input failed.\n")
 		}
 
-		cmdstr, args := parseCommand(cmdstr)
+		cmdstr, args := ParseCommand(cmdstr)
 
 		if cmdstr == "exit" {
 			handleExit(dbp, 0)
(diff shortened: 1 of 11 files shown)


derekparker:
Awesome, I'm excited to see some work on this.

I think it could be interesting for a command to return maybe some sort of result struct and an error. The result struct could then respond to `.String` and `.MarshalJSON`, which would be suitable for any front end we use.

dlsniper:
Cool, I'll do some work on this over the weekend and submit it in this PR as a different commit.

dlsniper:
Based on the ideas from @tarqe39, I've did some more refactoring on the initial PR in order to move it along the way it's described in the above description.

As a small observation, since goreadline is not using ```goimports``` to format their code, it's pretty ugly to always run ```goimports -w ./.. ``` then revert that one file. Shouldn't it be always part of the repository?

The next step is to add a common command output structure to the core then format the output to the web and CLI as discussed here: https://github.com/dlsniper/delve/pull/1

derekparker:
@dlsniper sorry for not getting back to you for awhile, been kind of heads down with OS X support lately.

I definitely think the goreadline sniped package should have correctly formatted imports, I might just push up a patch that fixes that formatting.

I've taken another brief look at your branch and it's looking good - definitely let me know when you'd like me to take another more in depth look at it.

dlsniper:
@derekparker don't worry, I don't really have time during the week anyway.
I've finished the adding support for output collection: 8c8a8fb0c226e636516ef90751d218c995e9e9c9
There are a few todos that I'd like you to have a look at as I'm not sure what's the correct way to solve them (or if they are correct to begin with, might be bugs?).

I've checked that the cli version works but as it is right now the web version doesn't work just yet.

Also, I'm not sure I'm happy with the way that we have the output right now.
I'm thinking to change the output struct from
```go
type cmdOutput struct {
	Out string
	Err error
}
```
to something like
```go
type cmdOutput struct {
	out interface{}
	err error
}

func (output *cmdOutput) JSONOutput() (string, error)
func (output *cmdOutput) Output() (string, error)
func (output *cmdOutput) Err() error
```

And then have the ```cmdOutput.out``` be decoded by the type of output in the respective output functions. What do you think?

dlsniper:
I've fixed the web version as well and rebased on latest master. Aside from this, the concerns from above are still valid. Once delve can distinguish between output on console vs json, I believe this can be merged (ofc, with the post-feedback changes included as well)

derekparker:
@dlsniper I'll be taking a close look at this PR today and work towards helping move it along. Now that OS X support is merged in it's freed me up a lot.

dlsniper:
@derekparker thanks a lot and congrats for the OS X support. I've put this on hold as I wanted to have some feedback with you before doing further changes.
Let me know if I should rebase this before you have a look at it :)

derekparker:
@dlsniper that would be great if you could rebase it.

dlsniper:
Ok, I'll look over the effort I need to but it should be done in a couple of days. Btw, you can always ping me on: https://gitter.im/go-lang-plugin-org/go-lang-idea-plugin (I can't add you to the org so that we can chat in the private chan, but you can also drop me a line on Gitter directly).

dlsniper:
I've tried to rebase the PR for the past 2 hours but I've bumped into lots of internal changes and at this point I'll just close this and let someone else take care of this feature as I don't have the time required for it.

PR #66, opened 2015-02-13
be consistent with messages for "info" command

jackielii:
the message from "help" command for "info" is different from the one with "info" when the "type" doesn't match


derekparker:
Thanks!

PR #68, opened 2015-02-14
Added TravisCI support and badges

dlsniper:
I think it would be useful to have a CI system working for PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+
+go:
+  - 1.4.1
+  - tip
+
+before_install:
+  - apt-get install libreadline-dev
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Delve
 
+[![Build Status](https://travis-ci.org/github.com/derekparker/delve.png?branch=master)](https://travis-ci.org/github.com/derekparker/delve)
+[![GoDoc](https://godoc.org/github.com/derekparker/delve?status.svg)](https://godoc.org/github.com/derekparker/delve)
+
 ### What is Delve?
 
 Delve is a (Beta) Go debugger, written in Go.


derekparker:
I definitely want some sort of CI - and Travis would work for right now. The trick is that in the future we would need more than Travis CI due to needing to test on multiple OSes and architectures.

derekparker:
@dlsniper and anyone else looking at this PR - any thoughts on this, is there an existing CI service that allows testing on multiple OSes/arches? If not I'll likely pull this in and then maybe start work on one (because we will certainly need that type of control for our CI environment for this project and it sounds fun to write).

pnasrat:
I don't think so - at least not that is free for OSS use - although we could (and probably should as a base) reuse the buildlet configurations from Go itself - https://github.com/golang/build/tree/master/env

I'd say get this in as an instant fix and then look at what would be useful - the trigger/build/report cycle is probably less complex than getting good kernel/arch/distro version coverage for warts - see Rob Pike's comment https://groups.google.com/d/msg/golang-nuts/2AiVI0F3yXE/LqWRn3DSJH4J

"All that stands in the way is the writing of a lot of non-portable low-level code talking to buggy undocumented interfaces."

derekparker:
@pnasrat I agree I think this will be a good first step, especially since I think travis covers ubuntu and OS X, so depending on arch differences (should be fine but Delve does use some arch specific features atm) but that should be enough to get up and running right now.

The buildlet stuff from Go seems like it could definitely be a good starting point for a more custom solution.

However, it's exactly the sentiment in that comment that makes me want all of these other custom builders - it's essential to know if something is broken on another OS/arch due to how reliant Delve _must_ be on those buggy/undocumented interfaces.

dlsniper:
Hi @derekparker. I'll look into it but I think if there's a service that does this for money you can sure ask them if they are willing to give you a free version for this project only ;-) If not, maybe ask the Go team itself?

derekparker:
Merged, thanks!

PR #77, opened 2015-03-03
Issue 73: die should restore the terminal state

raff:
I didn't want to add another parameter to die(), that didn't really make sense so instead I created a Terminator object that contains things that need to be cleaned up (for now, only the liner).

Turns out I had to change a bunch of other places, to pass the Terminator instead of liner.State, so in retrospect maybe adding the liner to die wasn't such a bad idea (if you don't like this implementation I can change it)

While there I also added support for command completion and the option to get help for a specific command (I have a bunch of command line tools with similar support and I can't live without it :)

### DIFF
--- a/client/cli/cli.go
+++ b/client/cli/cli.go
@@ -20,11 +20,11 @@ const historyFile string = ".dbg_history"
 
 func Run(run bool, pid int, args []string) {
 	var (
-		dbp  *proctl.DebuggedProcess
-		err  error
-		line = liner.NewLiner()
+		dbp *proctl.DebuggedProcess
+		err error
+		t   = Terminator{line: liner.NewLiner()}
 	)
-	defer line.Close()
+	defer t.line.Close()
 
 	switch {
 	case run:
@@ -32,23 +32,23 @@ func Run(run bool, pid int, args []string) {
 		cmd := exec.Command("go", "build", "-o", debugname, "-gcflags", "-N -l")
 		err := cmd.Run()
 		if err != nil {
-			die(1, "Could not compile program:", err)
+			t.die(1, "Could not compile program:", err)
 		}
 		defer os.Remove(debugname)
 
 		dbp, err = proctl.Launch(append([]string{"./" + debugname}, args...))
 		if err != nil {
-			die(1, "Could not launch program:", err)
+			t.die(1, "Could not launch program:", err)
 		}
 	case pid != 0:
 		dbp, err = proctl.Attach(pid)
 		if err != nil {
-			die(1, "Could not attach to process:", err)
+			t.die(1, "Could not attach to process:", err)
 		}
 	default:
 		dbp, err = proctl.Launch(args)
 		if err != nil {
-			die(1, "Could not launch program:", err)
+			t.die(1, "Could not launch program:", err)
 		}
 	}
 
@@ -67,7 +67,7 @@ func Run(run bool, pid int, args []string) {
 	if err != nil {
 		f, _ = os.Create(historyFile)
 	}
-	line.ReadHistory(f)
+	t.line.ReadHistory(f)
 	f.Close()
 	fmt.Println("Type 'help' for list of commands.")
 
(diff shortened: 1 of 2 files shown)


client/cli/cli.go
```diff
@@ -63,27 +63,52 @@ func Run(run bool, pid int, args []string) {
 	}()
 
 	cmds := command.DebugCommands()
+	names := cmds.Names()
```
derekparker: I think we'd want to just complete the full name of the command instead of including all of its aliases, which include the short, single letter name.

client/cli/cli.go
```diff
@@ -63,27 +63,52 @@ func Run(run bool, pid int, args []string) {
...
+
+	// set command completer
+	t.line.SetWordCompleter(func(line string, pos int) (head string, completions []string, tail string) {
+		start := strings.LastIndex(line[:pos], " ")
```
derekparker: Is this checking to see if we're trying to complete a subcommand, i.e. `info args` or something? I think those are going away (the `info` part of the last command) so I think only the second conditional is relevant. Also, I think we could get rid of the early return as well.

client/cli/cli.go
```diff
@@ -139,10 +164,18 @@ func handleExit(dbp *proctl.DebuggedProcess, line *liner.State, status int) {
 		}
 	}
 
-	die(status, "Hope I was of service hunting your bug!")
+	t.die(status, "Hope I was of service hunting your bug!")
 }
 
-func die(status int, args ...interface{}) {
+type Terminator struct {
```
derekparker: I like this direction, but I think we should rename this type `Term`, and add more responsibility to it. I think that `promptForInput` should be a method off of this type, and we should add a field in the struct for `prompt`. That way, when we initialize it we could do so like: `Term{prompt: "(dlv) ", line: line}`.

raff: And this is exactly where I wanted to go, after the initial implementation (I was going to change Terminator to Terminal). I'll update to Term.

Of course, the more you add, the more you get closer to reimplement my "cmd" package (github.com/gobs/cmd) :)

client/cli/cli.go
raff: This is to support various type of completion:
- if beginning of line you are looking for the command name
- if somewhere in the middle you are trying to do completion on some arguments. Right now it's just for "help" (complete using the command names) but for "info" as today it would complete on the subcommands, or for "break" it could complete on function names and "print" could complete on variable names (of course the completer code will have to be rearranged)

derekparker: I really like the idea of enhanced completion for breakpoint setting, especially auto completing function names, source files, etc..

Would you mind breaking the auto complete stuff out into a separate PR, because I'd like to merge that in by itself and have a place to discuss that feature.

client/cli/cli.go
raff: I have updated command.Names() to do that

raff:
I have updated the pull request with a couple of your suggestions. I am still not convinced about the third :)

derekparker:
The third being breaking out completion into its own PR?

Right now this commit fixes 1 bug and adds 2 features:

* Fix terminal state
* Add help for individual command
* Add tab completion

I'd like to keep this PR focused on fixing the bug, and have separate PRs for adding those features, that way discussion can be focused on one detail at a time.

raff:
No, I am fine moving the the tab completion and all related stuff in a
different PR. I was referring to some note in the comment for tab
completion.

I am going to close this PR (as soon as I get the time) and reopen it as
two separate PRs.

thanks!

-- Raffaele

derekparker:
Ah, gotcha. Sounds good.

derekparker:
@raff will you be able to tease this apart soon into multiple PRs? I'd really like to merge the liner fix.

raff:
Sorry, I got kind of busy.
I'll try to do today or tomorrow.

derekparker:
I'd really like to merge this liner fix in, as it's quite an annoying issue. I definitely understand you're hacking on this in your free time, but I will be closing this PR tomorrow and put in a fix myself unless this gets teased out (this PR is almost 2 weeks old by now).

Even if this gets closed without merge, I'd like PRs for the other stuff if you're still interested in hacking on them. Thanks.

raff:
Check new pull request (#84)

PR #78, opened 2015-03-05
Handle process natural death a bit better

ivarg:
The error message when the process exits normally is a bit annoying, so I figure we could use the ProcessExitedError value to capture this along with the exit status. Also, the wait status, as returned by trapWait isn't actually used anywhere so I removed it. Finally, a small cleanup in the Registers() error reporting causing an unnecessarily long message chain.

### DIFF
--- a/proctl/proctl_darwin.go
+++ b/proctl/proctl_darwin.go
@@ -172,14 +172,14 @@ func (dbp *DebuggedProcess) findExecutable() (*macho.File, error) {
 	return macho.Open(C.GoString(pathptr))
 }
 
-func trapWait(dbp *DebuggedProcess, pid int) (int, *sys.WaitStatus, error) {
+func trapWait(dbp *DebuggedProcess, pid int) (int, error) {
 	port := C.mach_port_wait(dbp.os.exceptionPort)
 	if port == 0 {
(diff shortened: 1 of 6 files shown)


proctl/proctl_darwin.go
```diff
@@ -172,14 +172,14 @@ func (dbp *DebuggedProcess) findExecutable() (*macho.File, error) {
 	return macho.Open(C.GoString(pathptr))
 }
 
-func trapWait(dbp *DebuggedProcess, pid int) (int, *sys.WaitStatus, error) {
+func trapWait(dbp *DebuggedProcess, pid int) (int, error) {
 	port := C.mach_port_wait(dbp.os.exceptionPort)
 	if port == 0 {
-		return -1, nil, ProcessExitedError{}
+		return -1, ProcessExitedError{Pid: pid}
```
derekparker: This should be `dbp.Pid`, since we're waiting on a mach port here, the pid that gets passed in means nothing to us, it's only for parity with the Linux side.

ivarg:
Of course. Fixed now.

derekparker:
Merged, thanks! There's some more work to be done to get the proper exit status for OS X, but that can be a separate PR.

PR #84, opened 2015-03-16
Issue 73: die should restore the terminal state

raff:
Cleaned up pull request that only contains the new Term object.

Added Term object (terminal or terminator) that wraps the liner.State object so that it can be closed properly on exit. Changed a few functions (die, promptForInput, etc.) to be methods for Term.


derekparker:
Merged, thank you!

PR #88, opened 2015-03-20
WIP: Refactor to HTTP client/server model

ironcladlou:
**NOTE** This is a work-in-progress, and the PR is to facilitate ongoing design discussion. This PR is intended to address #22.

### Goals

1. Introduce an HTTP server using `net.Listener` to provide a JSON-based remote API to the debugger
  1. Support RESTful operations for state retrieval and commands
  2. Support a long-polling watch requests to support event notifications
2. Provide a Go client interface and default HTTP implementation
4. Refactor the terminal to be a pure API client consumer

The original PR is based on [this comment](https://github.com/derekparker/delve/issues/22#issuecomment-84154167) on #22.

PR #98, opened 2015-04-04
Moving history to new ~/.dlv config path

### LINKED ISSUE #81: Move readline history file to more sensible spot
Right now delve creates readline history file in cwd, but this is not acceptable going forward.

epipho:
Fixes #81 

This is also a setup for a ~/.dlv/,dlv.conf file to allow commands to be run at startup (i.e. change display modes).

### DIFF
--- a/client/cli/cli.go
+++ b/client/cli/cli.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"os/user"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -18,6 +20,7 @@ import (
 	"github.com/peterh/liner"
 )
 
+const configDir = ".dlv"
 const historyFile string = ".dbg_history"
 
 func Run(args []string) {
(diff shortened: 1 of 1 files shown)


client/cli/cli.go
```diff
@@ -18,6 +20,7 @@ import (
 	"github.com/peterh/liner"
 )
 
+const configDir = ".dlv"
```
pnasrat: Probably want to think about win32 vs posix where you probably want different locations.

pnasrat: Actually I see you're using os/user to get homedir which I think will give USERPROFILE.  https://golang.org/src/os/user/lookup_windows.go#L115

epipho:
It does give USERPROFILE on windows.  

While the official recommendation from MS is to use the LocalAppData system folder pretty much all the cross platform apps on my Win7 box appear to use a dot directory in USERPROFILE including VirtualBox, Vagrant, Maven, Netbeans, Babun, Atom, and CMake.

derekparker:
This looks good but I'm going to wait until #88 lands to pull this in.

derekparker:
@epipho now that #88 has been merged, can you take a look at rebasing this and pushing back up? I'd be happy to pull this in after that.

epipho:
Yep I'll take a look this weekend.

epipho:
Closing this in favor of #124

PR #105, opened 2015-04-19
Variables.go 'info vars' and 'info locals' Fix

bforsyth927:
Fixed the 'index out of range' runtime error by testing for a bogus address and returning an empty string out of ReadString.

### DIFF
--- a/proctl/variables.go
+++ b/proctl/variables.go
@@ -693,7 +693,7 @@ func (thread *ThreadContext) readString(addr uintptr) (string, error) {
 	}
 	addr = uintptr(binary.LittleEndian.Uint64(val))
 	if addr == 0 {
-		return "", err
+		return "", nil
 	}
 
 	val, err = thread.readMemory(addr, strlen)


proctl/variables.go
```diff
@@ -692,6 +692,9 @@ func (thread *ThreadContext) readString(addr uintptr) (string, error) {
 		return "", err
 	}
 	addr = uintptr(binary.LittleEndian.Uint64(val))
+	if addr == 0 {
+		return "", err
```
derekparker: Explicitly return nil here, instead of value of last error.

bforsyth927:
Let me know if there's anything else you'd like me to take care of.

derekparker:
@bforsyth927 can you rebase this onto master, squash the commits and I'll merge this in.

derekparker:
@bforsyth927 any chance you can rebase and squash soon so I can merge this in?

derekparker:
Fix is in master.

PR #107, opened 2015-04-25
Improve the help text for dlv

technosophos:
This patch makes a few small changes to `dlv` to try to make the default help text work a little better.

* It adds support for the `-h` flag.
* It generates flag help text from the `flag.Flag` objects.
* it overrides `flag.Usage` with a custom `help()` function.


derekparker:
Merged, thanks!

PR #113, opened 2015-04-30
Headless support

ironcladlou:
I wanted to start discussion about how to extend the client/server model out to the CLI. Basically, there are a few use cases:

1. Start just the server in the foreground
2. Start just the server as a daemon
3. Connect a debugger Terminal to a server
4. Start both a server and a connected Terminal in the foreground

We should figure out what defaults and flags best facilitate all these. This PR has a small example of the "bare minimum" for IDE integrators, which is a `-headless` mode that runs just the server in the foreground and shuts down on SIGINT.

### DIFF
--- a/cmd/dlv/main.go
+++ b/cmd/dlv/main.go
@@ -3,11 +3,13 @@ package main
 import (
 	"flag"
 	"fmt"
+	sys "golang.org/x/sys/unix"
 	"io/ioutil"
 	"log"
 	"net"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


derekparker:
Starting the server as a daemon is tricky, because that makes it seem we can have a server sitting around, waiting to debug something, and many clients could connect to it, be it terminal/IDE whatever. This would potentially involve some modifications to Delve to support debugging multiple processes simultaneously, or at the least supporting multiple clients connected at once.

The headless support is definitely something we want asap, which would be essential for IDE/Editor integration.

cmd/dlv/main.go
```diff
@@ -120,15 +124,28 @@ func main() {
...
+	if !headless {
+		// Create and start a terminal
+		client := rest.NewClient(listener.Addr().String())
+		term := terminal.New(client)
+		err, status = term.Run()
+	} else {
+		ch := make(chan os.Signal)
+		signal.Notify(ch, sys.SIGINT)
+	runLoop:
+		for {
```
derekparker: Since we break immediately once we get a signal, is there need for the loop? Just block on the signal and stop the server.

ironcladlou: I must have been very tired...

ironcladlou:
I think this is the minimum viable thing for integrations. A reasonable followup PR would be another flag which starts a terminal connected to a specified headless server.

>Starting the server as a daemon is tricky, because that makes it seem we can have a server sitting around, waiting to debug something, and many clients could connect to it, be it terminal/IDE whatever. This would potentially involve some modifications to Delve to support debugging multiple processes simultaneously, or at the least supporting multiple clients connected at once.

I definitely don't want to imply or support multiple client support, although it's technically safe-ish at the moment due to our protection of the DebuggedProcess using the operations channel. Daemon mode probably doesn't make much sense.

derekparker:
Merged.

PR #114, opened 2015-05-02
Suppress output of HTTP error from client

ironcladlou:
Introduce a client error type which preserves HTTP error information
but which doesn't render the HTTP details by default.


ironcladlou:
@derekparker 

Before:

>Command failed: 500 Internal Server Error: could not get registers: no such process

After:

>Command failed: could not get registers: no such process

derekparker:
Merged.

Issue #115, opened 2015-05-03
Share more between proctl/integration tests


derekparker:
I'd like to share more between proctl/integration tests, including the TestMain precompilation step.

@ironcladlou you interested in working on this?

ironcladlou:
Sure. I think I might start this tonight.

PR #125, opened 2015-05-04
Consolidate test support/setup

ironcladlou:
Add a test support package which allows shared test functionality
for both the unit and integration tests.

Tests importing the proctl/test package will gain access to a special
test entrypoint which precompiles fixtures and makes them available
to tests.


ironcladlou:
This is part of #115. Once the basic pattern is in place, we can think about also sharing scenario data (e.g. the location scenarios for the next-tests).

derekparker:
Merged, thanks!

PR #128, opened 2015-05-07
Add default answer hint to kill process prompt

sethwklein:
Before exiting, dlv asks, "Would you like to kill the process? [y/n] ". This provides no hint as to what will happen if one just presses enter, nor is there any output to indicate what happened. A common convention in this situation is to capitalize the choice that will be assumed if just enter is pressed. This patch does so, and also allows answering with a capitalized Y or N and being messy with the space bar.

### DIFF
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -132,11 +132,11 @@ func handleExit(client service.Client, t *Term) (error, int) {
 		}
 	}
 
-	answer, err := t.line.Prompt("Would you like to kill the process? [y/n] ")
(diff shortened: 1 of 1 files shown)


derekparker:
This is good, but while we're on the subject, I'd like to actually reverse this behavior, and for the default to be to kill the process unless otherwise specified.

Could you modify this PR, or close it and submit another with the behavior described above?

sethwklein:
I'm glad you said this because I felt the same way. Update incoming.

sethwklein:
Superseded by #129.

PR #129, opened 2015-05-08
Default to killing process on exit

sethwklein:
This is a rebased version of #128 with an additional commit that changes the default answer to kill as requested.


derekparker:
Thanks!

PR #147, opened 2015-06-19
Implement regs command to print registers values, fixes #62

### LINKED ISSUE #62: Print contents of registers
Should be able to print contents of registers for current thread, and any other arbitrary thread.

dullgiulio:
Having some problems running the tests at the moment (on Linux and with go-tip). I will investigate further.

In the mean time you can have a look at this patch, if it does what is needed.

### DIFF
--- a/proc/registers.go
+++ b/proc/registers.go
@@ -11,6 +11,7 @@ type Registers interface {
 	SP() uint64
 	CX() uint64
 	SetPC(*Thread, uint64) error
+	String() string
 }
 
 // Obtains register values from the debugged process.
--- a/proc/registers_darwin_amd64.go
+++ b/proc/registers_darwin_amd64.go
@@ -8,6 +8,11 @@ type Regs struct {
 	pc, sp, cx uint64
 }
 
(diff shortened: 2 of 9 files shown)


derekparker:
Hey, thanks for this patch.

This is close, however we should print the contents of all the registers, not just the ones Delve cares about. On Linux, the generic Regs struct embeds the full system registers. We will want to do something similar on Darwin, and print the value of every register, using correct names (i.e. RIP instead of PC, etc..). Since this will be a relatively long list we'd want to newline delimit everything instead of comma separated.

derekparker:
Also, I'm working through some fixes for getting Delve to work against go-tip. Should have some patches in this afternoon.

dullgiulio:
Of course, I included all registers now.

I didn't use reflection: code is a bit redundant but explicit. Also, I don't have a Mac where I can try to build at the moment.

derekparker:
Merged, thanks!

I fixed up a few things that we're broken on OSX, and also standardized the formatting a bit for Darwin vs Linux (register ordering mostly).

PR #153, opened 2015-06-26
Linux bugfixes

aarzilli:
I see you already covered my d95f9ca, 
c2a346e is just an optimization,
b7e6e27 is a workaround for weird behaviour of wait on linux

### DIFF
--- a/proc/breakpoints.go
+++ b/proc/breakpoints.go
@@ -123,7 +123,7 @@ func (dbp *Process) setBreakpoint(tid int, addr uint64, temp bool) (*Breakpoint,
 	if _, err := readMemory(thread, uintptr(addr), originalData); err != nil {
 		return nil, err
 	}
-	if _, err := writeMemory(thread, uintptr(addr), dbp.arch.BreakpointInstruction()); err != nil {
+	if err := dbp.writeSoftwareBreakpoint(thread, addr); err != nil {
 		return nil, err
 	}
 	dbp.Breakpoints[addr] = &Breakpoint{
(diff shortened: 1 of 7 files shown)


proc/proc_linux.go
```diff
@@ -270,6 +272,10 @@ func (dbp *Process) trapWait(pid int) (*Thread, error) {
 			}
 			continue
 		}
+		if th == nil {
+			// Sometimes we get an unknown thread, ignore it?
```
derekparker: This doesn't seem right. The only time we should get an unknown thread is when a new one is cloned and it's the first time we're hearing about it. Was this crashing?

proc/proc_linux.go
```diff
@@ -304,14 +310,45 @@ func stopped(pid int) bool {
...
+
+		for {
+			wpid, err := sys.Wait4(pid, &s, sys.WNOHANG|sys.WALL|options, nil)
+			if err != nil {
+				return 0, nil, err
+			}
+			if wpid != 0 {
+				return wpid, &s, err
+			}
+
```
derekparker: Kill whitespace.

proc/breakpoints.go
```diff
@@ -139,6 +139,11 @@ func (dbp *Process) setBreakpoint(tid int, addr uint64, temp bool) (*Breakpoint,
 	return dbp.Breakpoints[addr], nil
 }
 
+func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
```
derekparker: I like this, but rename to setSoftwareBreakpoint to be more consistent with setHardwareBreakpoint.

derekparker:
This is great, thank you. Just some minor feedback.

proc/proc_linux.go
aarzilli: yes it was crashing, a nil pointer dereference at th.running = false later on, yes this is weird

proc/breakpoints.go
aarzilli: I will do that if you insist but they are not symmetric, setHardwareBreakpoint is a complete operation, to set a software breakpoint we have to read the original instructions and then write the breakpoint and writeSoftwareBreakpoint only does the second part.

derekparker: Good point, disregard this comment.

derekparker:
I rebased this for you since it was a bit tricky due to my code changes and merged it in. Thanks for the patch.

PR #162, opened 2015-07-04
Bug in Stacktrace rpc plumbing

aarzilli:
instead of passing the value the user specified for goroutineId the number '1' is always used.


derekparker:
Merged, thanks.

PR #167, opened 2015-07-09
proc.(*Process).CurrentBreakpoint is a function, it's never nil 

aarzilli:
probable refactoring mistake

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -389,7 +389,7 @@ func (dbp *Process) resume() error {
 	if err != nil {
 		return err
 	}
-	if dbp.CurrentBreakpoint != nil || dbp.halt {
+	if dbp.CurrentBreakpoint() != nil || dbp.halt {
 		return dbp.Halt()
 	}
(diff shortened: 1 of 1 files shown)


derekparker:
Good catch, however it should be `thread.CurrentBreakpoint` which is not a function. For some reason, changing it to that breaks every proc test pretty much right after `Launch`. Going to dig into this and see what's going on.

derekparker:
@aarzilli the broken tests was on OSX. Figured out the issue, should be patched soon.

derekparker:
@aarzilli actually I take my first comment back, `dbp.CurrentBreakpoint` is fine since we switch the main thread a few lines above.

derekparker:
Fixed by: https://github.com/derekparker/delve/commit/aa71d787aa2a96247128d5395301c7d4a5d2c7a6

Essentially, there is no reason to do those checks. Simply stop the world once trapWait returns, as it will have done its due diligence.

PR #172, opened 2015-07-10
Continue bug

aarzilli:
Set function breakpoints on the first instruction (instead of the function entry point)

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -172,7 +172,31 @@ func (dbp *Process) FindLocation(str string) (uint64, error) {
 	// Try to lookup by function name
 	fn := dbp.goSymTable.LookupFunc(str)
 	if fn != nil {
-		return fn.Entry, nil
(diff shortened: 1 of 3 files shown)


derekparker:
Sorry for not responding right away on this. I'm somewhat torn. It seems unexpected to set a BP at a function, yet stop on the first line, however we _do_ want to skip the prologue. Maybe there can be an extra step in 'next' that checks if it is at a function entry && there is a breakpoint set there, and if so temporarily clears and defer restores it.

What are your thoughts there?

aarzilli:
IMHO few people care about the distinction between setting the breakpoint on a function entry and a function's first line, if anyone cared about setting a breakpoint on the prologue I imagine it would be because they are interested in what goes on in the prologue and probably wouldn't appreciate delve messing with it.
There should probably be some syntax for setting a breakpoint on a function's line (which would open up syntax to set the breakpoint on function entry when necessary). I was going to propose something after I'm done with the breakpoints file (I've been occupied with other things)

derekparker:
Merged.

PR #173, opened 2015-07-19
Put '> ' in front of the location marker.

benma:
This makes parsing by external tools much easier.
It is consistent with pdb.

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -546,9 +546,9 @@ func printcontext(state *api.DebuggerState) error {
 		for _, arg := range state.CurrentThread.Function.Args {
 			args = append(args, arg.Value)
 		}
-		fmt.Printf("%s(%s) %s:%d\n", fn.Name, strings.Join(args, ", "), state.CurrentThread.File, state.CurrentThread.Line)
+		fmt.Printf("> %s(%s) %s:%d\n", fn.Name, strings.Join(args, ", "), state.CurrentThread.File, state.CurrentThread.Line)
 	} else {
-		fmt.Printf("%s() %s:%d\n", fn.Name, state.CurrentThread.File, state.CurrentThread.Line)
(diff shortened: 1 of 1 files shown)


benma:
On the GopherCon Hackday, I integrated Delve into Emacs: https://github.com/benma/go-dlv.el

It needs to parse the current location. The format of that has changed in https://github.com/derekparker/delve/commit/3cee10d8bc5331bde18f350be3ff33414dba2cc2#diff-8fc81033af09d02d4ed488bfe1f3d909L544, breaking the Emacs integration. Before,`current loc: ` was the constant prefix, which was also easy to parse.

derekparker:
Sorry for the late reply, I'm on vacation with spotty wifi.

So you're driving Delve via the terminal interface from emacs? The RPC
protocol gives machine readable output which includes the current location.
It's under documented at the moment but it is there for this use case.

benma:
Thanks for the response.

Emacs already comes bundled with an interface to gdb, pdb and many more. They are all built the same way, relying on common functionality to interface with the debuggers stdout. I did it the same way (I basically took the pdb integration code and adapted it slightly to match dlv).

It is good to know that there is an RPC protocol, however, I don't have the time to dig into accessing that from Emacs, seeing that the Emacs debugging facility does not support that yet (see [gud.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/gud.el)). Maybe this can be contributed in the future.

If you don't mind too much, I would still like you to merge this PR, so the current way of integrating it works again. What do you think?

derekparker:
Sure thing, that makes sense. I will merge this in Saturday once I have
access to my machine again.

derekparker:
Merged.

benma:
Thanks. The go-dlv debugger is updated and working again: https://github.com/benma/go-dlv.el/commit/62f657fc538660be2bc97550e7ad75c8fd9f380f

PR #175, opened 2015-07-20
Loading breakpoints/tracepoints from a file

aarzilli:
It's implemented on top of the continue-bug branch that I already have a PR open for, I will rebase it as needed.

### DIFF
--- a/_fixtures/bpfile
+++ b/_fixtures/bpfile
@@ -1,0 +1,8 @@
+# comment line
+trace main.main
+
+break main.sayhi
+
+#
+# comment line
+#
--- a/cmd/dlv/main.go
+++ b/cmd/dlv/main.go
@@ -27,6 +27,7 @@ var (
 	Log      bool
 	Headless bool
 	Addr     string
+	InitFile string
 )
 
 func main() {
@@ -48,6 +49,7 @@ The goal of this tool is to provide a simple yet powerful interface for debuggin
 	rootCommand.PersistentFlags().StringVarP(&Addr, "listen", "l", "localhost:0", "Debugging server listen address.")
 	rootCommand.PersistentFlags().BoolVarP(&Log, "log", "", false, "Enable debugging server logging.")
 	rootCommand.PersistentFlags().BoolVarP(&Headless, "headless", "", false, "Run debug server only, in headless mode.")
+	rootCommand.PersistentFlags().StringVar(&InitFile, "init", "", "Init file, executed by the terminal client")
 
 	// 'version' subcommand.
 	versionCommand := &cobra.Command{
@@ -292,6 +294,10 @@ func execute(attachPid int, processArgs []string, conf *config.Config) int {
 	}
 	defer listener.Close()
 
+	if Headless && (InitFile != "") {
+		fmt.Fprintf(os.Stderr, "Warning: init file ignored\n")
+	}
+
 	// Create and start a debugger server
 	server := rpc.NewServer(&service.Config{
 		Listener:    listener,
@@ -309,6 +315,7 @@ func execute(attachPid int, processArgs []string, conf *config.Config) int {
 		var client service.Client
 		client = rpc.NewClient(listener.Addr().String())
 		term := terminal.New(client, conf)
+		term.InitFile = InitFile
 		err, status = term.Run()
 	} else {
 		ch := make(chan os.Signal)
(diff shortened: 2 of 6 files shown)


derekparker:
I'll have a chance to review this weekend.

derekparker:
Can you rebase this on master.

aarzilli:
rebased

derekparker:
I'm not sold on continuing to watch the file for changes after the debugger has started. Seems like we should just parse that file on startup, add the breakpoints / tracepoints and then go about our way.

aarzilli:
I like to be able to change or add a breakpoint in a single place, instead of having to give a command to delve and then manually update the file.

derekparker:
I wonder how common of a use case that is, and whether it's worth complicating the code with file system notifications. @pmorie you had created the original issue of setting breakpoints from a file, would you mind chiming in on this? Essentially whether it's beneficial enough from a users perspective to have Delve watch the input breakpoint file for changes, and then apply those changes (i.e. add breakpoints / tracepoints).

derekparker:
Two things:

1) Can you separate the watch stuff from the initial "read breakpoints from file" patch.
2) I think a more general "source" type command (for evaluating the breakpoint file) would be cool. Because essentially, the breakpoint file is a newline delimited list of valid Delve commands. Delve should be able to essentially source a file and have it take effect.

derekparker:
Lastly, could you rework the commit message to conform to: https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message

pmorie:
I will TAL in the next couple days

aarzilli:
Done. A file that can contain any command is not a 'breakpoints file' however, so I have renamed it to 'init file'.

But now that we have config files I wonder if it wouldn't be better to add a flag to specify a config file and move the breakpoints/tracepoints inside the configuration?

derekparker:
I think this is really good, however there is one issue. When executing `dlv debug --init="./commands.txt" --headless` the init file is ignored because we do not start a terminal in headless mode. How about we attach the InitFile to the server instead, and let the server have access to the commands so it can call `cmds.executeFile`, while still preserving the `source` command in the terminal.

That way this init file will be executed even in headless mode, and allows us to have a generic "startup file" that is sourced immediately on startup by default, and that would work in headless mode as well.

We could even move the command code into its own package, and away from the terminal package. That would make this a bit cleaner, I think.

Thoughts?

aarzilli:
It seems like a lot of trouble without much of a use case at the moment: where's the output from commands executed by the headless server going to go? What happens if the init script ends with a `continue` command? Are client going to start by issuing a `Halt` call? How are they going to know when the `Halt` has taken effect? If someone writes a different client why should the init file still be in this command language?

derekparker:
@aarzilli good points. Let's just provide information in the doc string for the flag that it will not take effect if Delve is run in headless mode.

pmorie:
@derekparker Apologies, I have been in dire straights getting some stuff sorted for Kubernetes 1.1 and OpenShift 3.1.  Still trying to find time to review this.

aarzilli:
done.

derekparker:
@aarzilli please rebase and I will merge.

aarzilli:
rebased

pmorie:
I haven't had a chance to review the code, but this LGTM from the standpoint of how I was expecting to use the feature.

derekparker:
Merged.

aarzilli:
I know you don't like the idea of automatically reloading on change with inotify, but I find it useful, if you end up changing your mind I'm keeping my changes [here](https://github.com/aarzilli/delve/tree/initfile-autoreload).

pmorie:
One issue w/ using inotify is that you can run out of events.

aarzilli:
@pmorie You mean running out of watches or you mean the event queue overflowing? Neither seems like a problem for this application. Is there some other issue I'm not aware of?

pmorie:
I was referring to running out of watches.  I don't think it will be a
problem for short-running sessions but I know of use-cases where people ran
something in a debugger for longer than you would expect, hoping to
recreate a bug, so I just thought I would mention it.

On Tue, Oct 6, 2015 at 2:14 PM, Alessandro Arzilli <notifications@github.com

aarzilli:
I see.

PR #190, opened 2015-07-29
add list command, for displaying source

### LINKED ISSUE #58: Add command to list current location in source
When stopped in a debug session, have a command to show source code around the current location this (or any arbitrary) thread is stopped at. Should have a default (5 lines above/below or similar) but also accept an argument to display more source code context.

joeshaw:
`list` by itself shows code around the current context.  `list
file.go:100` displays code from `file.go` around line 100.

Fixes #58

--

Still some work to be done if you want to match the gdb behavior.
Subsequent `list` calls don't advance the cursor in displaying source,
and there are no options to get greater context beyond the +/- 5 lines
you get normally.

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -523,7 +523,7 @@ func listCommand(client service.Client, args ...string) error {
 		return fmt.Errorf("Ambiguous location: %s", args[0])
 	}
 
-	printfile(locs[0].File, locs[0].Line)
+	printfile(locs[0].File, locs[0].Line, false)
 	return nil
 }
 
@@ -586,10 +586,10 @@ func printcontext(state *api.DebuggerState) error {
 		return nil
 	}
 
-	return printfile(state.CurrentThread.File, state.CurrentThread.Line)
+	return printfile(state.CurrentThread.File, state.CurrentThread.Line, true)
 }
 
-func printfile(filename string, line int) error {
+func printfile(filename string, line int, showArrow bool) error {
 	file, err := os.Open(filename)
 	if err != nil {
 		return err
@@ -624,9 +624,12 @@ func printfile(filename string, line int) error {
 			}
 		}
 
-		arrow := "  "
-		if i == l {
-			arrow = "=>"
+		var arrow string
+		if showArrow {
+			arrow = "  "
+			if i == l {
+				arrow = "=>"
+			}
 		}
 
 		var lineNum string
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -68,6 +68,7 @@ func DebugCommands(client service.Client) *Commands {
 		{aliases: []string{"regs"}, cmdFn: regs, helpMsg: "Print contents of CPU registers."},
 		{aliases: []string{"exit", "quit", "q"}, cmdFn: exitCommand, helpMsg: "Exit the debugger."},
 		{aliases: []string{"stack", "bt"}, cmdFn: stackCommand, helpMsg: "stack [<depth> [<goroutine id>]]. Prints stack."},
+		{aliases: []string{"list"}, cmdFn: listCommand, helpMsg: "list <linespec>.  Show source around current point or provided linespec."},
 	}
 
 	return c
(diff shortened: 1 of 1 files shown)


derekparker:
Thanks for the patch, I have a few thoughts:

* I really don't like the command name "list", but I can't think of anything more descriptive. Ideas?
* I think we need to standardize how we handle anything that describes a location in source, similar to gdb's linespec (https://sourceware.org/gdb/onlinedocs/gdb/Specify-Location.html). The parsing of a linespec should apply to this command, and any similar command (break, trace, etc..)

That last point should happen first, and be applied to this command. I know @kerlyn-bsd3 started down that path, but got caught up with other obligations.

joeshaw:
`list` is what gdb calls it, but other options could be `source` or `show`.  Or we could go with a longer name and a short alias like `list-source` and `ls`.

I'll defer on the linespec thing for a little bit.  I agree it's the way to go but I'm new to the codebase and my hacks have been pretty superficial so far.  Looks like the breakpoint code ultimately uses `Process.FindLocation` for that, but that just returns a program counter.  Would using that be the right way to go, just checking for `err != nil` and then printing the file?

derekparker:
I think it makes sense for `Process.FindLocation` to be renamed to `Process.ParseLocation` and return a `Location` struct (located in `proc/threads.go:34`). Also I think that `Location` struct should be moved to `proc/proc.go`. That way it can be generalized a bit and used by anything that needs to parse a location string. Thoughts?

aarzilli:
I'm currently implementing the syntax for breakpoint locations that I described on the mailing list, I'm almost done.
Currently the code lives in service/debugger, takes a string as input and returns a []api.Breakpoint, the api.Breakpoint objects in the slice only have the Addr field set (I've changed the code so that CreateBreakpoint can set a breakpoint by address).
This is the most convenient way for break and trace, I could switch to returning []api.Location and fill up all its fields.

derekparker:
@aarzilli I like that.

derekparker:
@joeshaw can you rebase this onto master and work off of the work @aarzilli did for parsing a location?

joeshaw:
I force pushed an update that uses the new `api.Location` stuff.  Much nicer, thanks @aarzilli.

joeshaw:
By the way, I still used `list` as the command name, but I'm happy to change it to something else.

joeshaw:
OS X test failures seem unrelated to my patch

terminal/command.go
```diff
@@ -579,7 +607,14 @@ func printcontext(state *api.DebuggerState) error {
 		}
 	}
 
-	for i := l - 5; i <= l+5; i++ {
+	max := func(a, b int) int {
```
derekparker: I think we should just inline this, and move it up above line 603, and instead of `l - 5`, just use this same logic to set `l` appropriately to `l-5` or `0`.

joeshaw: Yeah, I'm not really sure what I was thinking with creating the separate function, I'll fix that.  But I don't think moving it up to line 603 is the right thing.  `l` is the current position in the file and required state for drawing the arrow, and so we don't want to change it.

I could use `line` instead, but it gets shadowed in the inner loop as a string.  I could fix that too, I guess it's just a question of how much cleanup you want me to do there.

Just inlining the function, the diff would be something like this:

```diff
diff --git i/terminal/command.go w/terminal/command.go
index e44d27e..1661e88 100644
--- i/terminal/command.go
+++ w/terminal/command.go
@@ -607,14 +607,12 @@ func printfile(filename string, line int) error {
                }
        }

-       max := func(a, b int) int {
-               if a > b {
-                       return a
-               }
-               return b
+       s := l - 5
+       if s < 0 {
+               s = 0
        }

-       for i := max(0, l-5); i <= l+5; i++ {
+       for i := s; i <= l+5; i++ {
                line, err := buf.ReadString('\n')
                if err != nil {
                        if err != io.EOF {
```

joeshaw: hmm, and maybe the lower bound should be 1 instead of 0...

derekparker: Good point. That makes sense, and yes let's set the lower bound to 1.

joeshaw:
Ok, force-pushed an update which makes that tweak.

terminal/command.go
```diff
@@ -503,6 +504,29 @@ func stackCommand(client service.Client, args ...string) error {
...
+		printcontext(state)
+		return nil
+	}
+
+	locs, err := client.FindLocation(args[0])
+	if err != nil {
+		return err
+	}
+
+	if len(locs) > 0 {
```
derekparker: This should be `> 1`.

terminal/command.go
```diff
@@ -503,6 +504,29 @@ func stackCommand(client service.Client, args ...string) error {
...
+		return fmt.Errorf("Ambiguous location: %s", args[0])
+	}
+
+	printfile(locs[0].File, locs[0].Line)
```
derekparker: It would be nice to pass a bool to printfile to tell it whether or not to print the current location arrow, since when we `list main.main` or whatever we really don't want the arrow printed.

joeshaw:
@derekparker thanks, i've addressed those comments, PTAL.

derekparker:
Merged, thanks!

PR #191, opened 2015-07-29
add bt alias for stack, quit and q for exit

### LINKED ISSUE #182: Command aliases
It'd be helpful if I could specify aliases for commands, preferably in a global and/or local config file.
Like
```
alias bt=stack
alias quit=exit
```

joeshaw:
Fixes #182


derekparker:
Merged, thanks.

PR #194, opened 2015-07-30
FindLocations API (user friendlier way to specify breakpoint locations)

aarzilli:
I developed it on top of the breakpoints file PR, if it needs to be extricated from it I will.

I've decided against the equivalence between package.function and package.function:1, because line 1 of a function could actually be inside the definition of an inline function.
Instead:
- package.function:0 is the prologue of package.function
- package.function:1 is line 1 of package.function, regardless of what that actually means
- package.function calls dbp.ast.NextLines to find out what the real first line of package.function actually is

IMHO this is the most useful semanthics.

I've also added `+<offset>`, `-<offset>` and `*<address>` location specifiers, like in gdb.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"strings"
 	"sync"
 
@@ -146,69 +145,50 @@ func (dbp *Process) LoadInformation(path string) error {
 	return nil
 }
 
-// Find a location by string (file+line, function, breakpoint id, addr)
-func (dbp *Process) FindLocation(str string) (uint64, error) {
-	// File + Line
-	if strings.ContainsRune(str, ':') {
-		fl := strings.Split(str, ":")
-
-		fileName, err := filepath.Abs(fl[0])
-		if err != nil {
-			return 0, err
-		}
-
-		line, err := strconv.Atoi(fl[1])
-		if err != nil {
-			return 0, err
-		}
-
-		pc, _, err := dbp.goSymTable.LineToPC(fileName, line)
-		if err != nil {
-			return 0, err
-		}
-		return pc, nil
-	}
-
-	// Try to lookup by function name
-	fn := dbp.goSymTable.LookupFunc(str)
-	if fn != nil {
-		// fn.Entry, the entry point of the function, is always a prologue, the prologue may call into the scheduler or grow the stack,
-		// this will result in the breakpoint getting hit multiple times without any apparent program progress inbetween.
-		// In order to avoid this confusing behaviour we try to find the first line of the function and set the breakpoint there.
-		filename, lineno, _ := dbp.goSymTable.PCToLine(fn.Entry)
-		var firstLinePC uint64
-		var err error
-		for {
-			lineno++
-			firstLinePC, _, err = dbp.goSymTable.LineToPC(filename, lineno)
-			if err == nil {
-				break
-			}
-			if _, unk := err.(*gosym.UnknownLineError); !unk {
-				return 0, err
-			}
-		}
-
-		var breakAddr uint64
-		if firstLinePC < fn.End {
-			breakAddr = firstLinePC
-		} else {
-			breakAddr = fn.Entry
-		}
-		return breakAddr, nil
-	}
-
-	// Attempt to parse as number for breakpoint id or raw address
-	id, err := strconv.ParseUint(str, 0, 64)
+func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
+	pc, _, err := dbp.goSymTable.LineToPC(fileName, lineno)
 	if err != nil {
-		return 0, fmt.Errorf("unable to find location for %s", str)
+		return 0, err
 	}
-	if bp, ok := dbp.FindBreakpointByID(int(id)); ok {
-		return bp.Addr, nil
+	return pc, nil
+}
+
+// Finds address of a function's line
+// If firstLine == true is passed FindFunctionLocation will attempt to find the first line of the function
+// If lineOffset is passed FindFunctionLocation will return the address of that line
+// Pass lineOffset == 0 and firstLine == false if you want the address for the function's entry point
+// Note that setting breakpoints at that address will cause surprising behavior:
+// https://github.com/derekparker/delve/issues/170
+func (dbp *Process) FindFunctionLocation(funcName string, firstLine bool, lineOffset int) (uint64, error) {
+	fn := dbp.goSymTable.LookupFunc(funcName)
+	if fn == nil {
+		return 0, fmt.Errorf("Could not find function %s\n", funcName)
 	}
 
-	// Last resort, use as raw address
-	return id, nil
+	if firstLine {
+		filename, lineno, _ := dbp.goSymTable.PCToLine(fn.Entry)
+		if filepath.Ext(filename) != ".go" {
+			return fn.Entry, nil
+		}
+
+		lines, err := dbp.ast.NextLines(filename, lineno)
+		if err != nil {
+			return 0, err
+		}
+
+		if len(lines) > 0 {
+			linePC, _, err := dbp.goSymTable.LineToPC(filename, lines[0])
+			return linePC, err
+		} else {
+			return fn.Entry, nil
+		}
+	} else if lineOffset > 0 {
+		filename, lineno, _ := dbp.goSymTable.PCToLine(fn.Entry)
+		breakAddr, _, err := dbp.goSymTable.LineToPC(filename, lineno+lineOffset)
+		return breakAddr, err
+	}
+
+	return fn.Entry, nil
 }
 
 // Sends out a request that the debugged process halt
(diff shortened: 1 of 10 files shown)


derekparker:
Yes, please extract this from the breakpoint file PR.

aarzilli:
done

proc/proc.go
```diff
@@ -146,69 +144,45 @@ func (dbp *Process) LoadInformation(path string) error {
...
-		for {
-			lineno++
-			firstLinePC, _, err = dbp.goSymTable.LineToPC(filename, lineno)
-			if err == nil {
-				break
-			}
-			if _, unk := err.(*gosym.UnknownLineError); !unk {
-				return 0, err
-			}
+		lines, err := dbp.ast.NextLines(filename, lineno)
```
derekparker: This function explicitly only supports Go source files, what happens if the user specifies a cgo func?

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
 
 	// Returns whether we attached to a running process or not
 	AttachedToExistingProcess() bool
+
+	// Returns the address of a location expression
```
derekparker: Change to: "Returns concrete location information described by a location expression"

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
...
+	// loc ::= <filename>:<line> | <function>[:<line>] | /<regex>/
+	// * <filename> can be the full path of a file or just a suffix
+	// * <function> ::= <package>.<receiver type>.<name> | <package>.(*<receiver type>).<name> | <receiver type>.<name> | <package>.<name> | (*<receiver type>).<name> | <name>
+	// * <function> must be unambiguous
+	// * /<regex>/ will return a breakpoint for each function matched by regex
```
derekparker: s/breakpoint/location/

also

This is inconsistent with the info commands. Most of those accept a regexp but do not require /<regexp>/. I understand it's difficult to determine whether it's a regexp or not otherwise, but I don't like this inconsistency with other commands.

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
...
+	// * *<address> sets a breakpoint at a program address
```
derekparker: Change to: "returns location information for given program address"

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
...
+	// NOTE: this function does not actually set breakpoints.
```
derekparker: Remove all references to breakpoints above and then remove this line.

service/debugger/debugger.go
```diff
@@ -443,3 +459,20 @@ func convertStacktrace(rawlocs []proc.Location) []api.Location {
 
 	return locations
 }
+
+func (d *Debugger) FindLocation(locStr string) ([]api.Location, error) {
+
```
derekparker: Kill whitespace.

service/debugger/locations.go
```diff
@@ -0,0 +1,291 @@
...
+import (
+	"debug/gosym"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/derekparker/delve/service/api"
+)
+
+const maxFindLocationCandidates = 5
```
derekparker: This seems to be only used in NormalLocationSpec.Find which returns an error if > 1 locations are found, is this necessary?

service/debugger/locations.go
```diff
@@ -0,0 +1,291 @@
...
+	if spec.PackageOrReceiverName != "" && spec.PackageOrReceiverName != sym.PackageName() && spec.PackageOrReceiverName != recv {
+		return false
+	}
+
+	return true
+}
+
+func (loc *RegexLocationSpec) Find(d *Debugger, locStr string) ([]api.Location, error) {
+	funcs := d.process.Funcs()
+	r := make([]api.Location, 0, len(funcs))
```
derekparker: Allocate this after we get `matches` back and use `len(matches)` instead of `len(funcs)`.

service/debugger/locations.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	return r, nil
+}
+
+func (loc *AddrLocationSpec) Find(d *Debugger, locStr string) ([]api.Location, error) {
+	return []api.Location{{PC: loc.Addr}}, nil
+}
+
+func (loc *NormalLocationSpec) FileMatch(filename string) bool {
```
derekparker: Does this expect a full file path? If so: s/filename/filepath/

service/debugger/locations.go
```diff
@@ -0,0 +1,291 @@
...
+	return strings.HasSuffix(filename, loc.Base) && (filename[len(filename)-len(loc.Base)-1] == '/')
```
derekparker: Why not just `filepath.Base(filename) == loc.Base`?

service/test/integration_test.go
```diff
@@ -526,3 +526,81 @@ func TestClientServer_traceContinue2(t *testing.T) {
...
+
+		someTypeStringFuncAddr := findLocationHelper(t, c, "locationsprog.go:14", false, 1, 0)[0]
+		otherTypeStringFuncAddr := findLocationHelper(t, c, "locationsprog.go:18", false, 1, 0)[0]
+		findLocationHelper(t, c, "SomeType.String", false, 1, someTypeStringFuncAddr)
+		findLocationHelper(t, c, "(*SomeType).String", false, 1, someTypeStringFuncAddr)
+		findLocationHelper(t, c, "main.SomeType.String", false, 1, someTypeStringFuncAddr)
+		findLocationHelper(t, c, "main.(*SomeType).String", false, 1, someTypeStringFuncAddr)
+
+		stringAddrs := findLocationHelper(t, c, "/^main.*Type.*String$/", false, 2, 0)
+		_ = stringAddrs
```
derekparker: Seems leftover?

service/debugger/locations.go
```diff
@@ -0,0 +1,291 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/derekparker/delve/service/api"
+)
+
+const maxFindLocationCandidates = 5
+
+type LocationSpec interface {
+	Find(d *Debugger, locStr string) ([]api.Location, error)
```
derekparker: The implementation of this method should always return fully filled out Location structs. I noticed some only return Addr filled out.

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
...
+	FindLocation(loc string) ([]api.Location, error)
```
derekparker: This should also accept something like `break <line>` to set a breakpoint at <line> of current file.

service/client.go
aarzilli: Should we change the other commands? Because Some.Function could mean both the function Function in package Some and a regex matching Some1Function, Some2Function, etc…

service/client.go
aarzilli: Which line?

service/debugger/locations.go
aarzilli: It's not strictly necessary, I wanted to include in the error message some extra information about the source for the ambiguity but I also didn't want the error message to potentially be very long.

service/debugger/locations.go
aarzilli: I thought it would be useful to accept partial paths as well. Just matching with filepath.Base would make it impossible to disambiguate service/rpc/client.go from service/client.go.

I changed '/' to filepath.Separator.

service/debugger/locations.go
aarzilli: They all only fill PC, all the other fields are filled by debugger.(*Debugger).FindLocation

derekparker:
Sorry for delay, will review again tomorrow morning.

proc/proc.go
```diff
@@ -146,69 +145,50 @@ func (dbp *Process) LoadInformation(path string) error {
...
 
-		fileName, err := filepath.Abs(fl[0])
-		if err != nil {
-			return 0, err
-		}
+// Finds address of a function's line
+// If firstLine == true is passed FindFunctionLocation will attempt to find the first line of the function
+// If lineOffset is passed FindFunctionLocation will return the address of that line
+// Pass lineOffset == 0 and firstLine == false if you want the address for the function's entry point
+// note that setting breakpoints at that address will cause surpirsing behavior:
```
derekparker: s/note/Note/ && s/surpirsing/surprising/

proc/proc.go
```diff
@@ -146,69 +145,50 @@ func (dbp *Process) LoadInformation(path string) error {
...
-		if err != nil {
-			return 0, err
+	if firstLine {
+		filename, lineno, _ := dbp.goSymTable.PCToLine(fn.Entry)
+		if filepath.Ext(filename) != ".go" {
+			return fn.Entry, nil
 		}
 
-		pc, _, err := dbp.goSymTable.LineToPC(fileName, line)
+		lines, err := dbp.ast.NextLines(filename, lineno)
```
derekparker: This is unsafe, this function assumes that `filename` will be a Go source file. This should support finding the first line of a non-go function as well, for CGO support. I think a more generic NextLines that looks at the file extension could be helpful here.

service/client.go
```diff
@@ -72,4 +72,14 @@ type Client interface {
 
 	// Returns whether we attached to a running process or not
 	AttachedToExistingProcess() bool
+
+	// Returns concrete location information described by a location expression
```
derekparker: Need to update documentation for `break <line_number>` support that I saw you added.

service/debugger/locations.go
derekparker: Ah yeah, I see what you're getting at.

proc/proc.go
derekparker: More thoughts: A non-Go function will obviously not have the stack check we're trying to avoid here in function entry. We could either:

a) Pass `filepath.Base(filename) == ".go"` for firstLine, so that we skip this branch altogether for non-Go files.
b) Use a `NextLines` function that is generalized to accept non Go source files, this would enforce more consistency around setting a breakpoint on a function.

Thoughts? I'm really not tied to either solution.

aarzilli: I added a check for filepath.Ext(filename) != ".go" there, I think you missed it. As you say non-go functions do not have the prologue we are trying to skip.

derekparker: Perfect, yeah I did miss that. Also typo'ed `filepath.Base` instead of `filepath.Ext` in my comment. Sigh.

derekparker:
Merged, thanks.

PR #196, opened 2015-08-01
Kill should wait for the process to exit

aarzilli:
TestKill was failing for me.

### DIFF
--- a/proc/proc_linux.go
+++ b/proc/proc_linux.go
@@ -71,9 +71,15 @@ func (dbp *Process) Kill() (err error) {
 	if err != nil {
 		return errors.New("could not deliver signal " + err.Error())
 	}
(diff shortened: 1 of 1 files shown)


proc/proc_linux.go
```diff
@@ -71,9 +71,15 @@ func (dbp *Process) Kill() (err error) {
 	if err != nil {
 		return errors.New("could not deliver signal " + err.Error())
 	}
-	_, _, err = wait(-1, dbp.Pid, 0)
```
derekparker: What about replacing `-1` with `dbp.Pid` to explicitly wait for the process pid, instead of iterating?

derekparker:
Fixed by 8034b7c038a476fd5506324c119af5eeeaf638fe

PR #199, opened 2015-08-07
Three bugs concerning variable evaluation

aarzilli:


proc/variables.go
```diff
@@ -240,6 +240,20 @@ func (thread *Thread) EvalVariable(name string) (*Variable, error) {
 		}
 	}
 
+	// attempt to evaluate name as a package variable
+	if strings.Index(name, ".") >= 0 {
```
derekparker: This check is already happening above. You could just check that `memberName != ""`.

derekparker:
Merged.

PR #205, opened 2015-08-12
Variable evaluation on arbitrary (goroutine, frame) pair.

aarzilli:


proc/proc.go
```diff
@@ -408,6 +408,20 @@ func (dbp *Process) SwitchThread(tid int) error {
 	return fmt.Errorf("thread %d does not exist", tid)
 }
 
+// Change from current thread to the thread running the specified goroutine
+func (dbp *Process) SwitchGoroutine(gid int) error {
+	for i := range dbp.Threads {
+		if dbp.Threads[i].blocked() {
+			continue
+		}
+		g, _ := dbp.Threads[i].GetG()
```
derekparker: Shouldn't ignore error here, we could panic if `g` is nil.

derekparker:
Can you rebase this on master? I'll review once that's done.

derekparker:
Also, for larger changes like this, it would help to provide a description in the commit after the title.

aarzilli:
Sorry, this adds the ability to examine variables on a arbitrary stack frames of an arbitrary goroutines (including ones that aren't running on a thread). 
With this change it's possible to print the value of function arguments (and local variables) of all the frames of a stack trace, like gdb does.

I've copied the gdb model for the UI as well, I added two commands 'frame' and 'goroutine' that select the "current" goroutine and the "current" frame, then print, args and locals evaluate variables on that scope.

proc/proc.go
```diff
@@ -408,6 +408,20 @@ func (dbp *Process) SwitchThread(tid int) error {
 	return fmt.Errorf("thread %d does not exist", tid)
 }
 
+// Change from current thread to the thread running the specified goroutine
+func (dbp *Process) SwitchGoroutine(gid int) error {
+	for i := range dbp.Threads {
```
derekparker: Rename `i` to `tid`.

proc/stack.go
```diff
@@ -49,42 +55,51 @@ func (n NullAddrError) Error() string {
...
 }
 
-func (dbp *Process) stacktrace(pc, sp uint64, depth int) ([]Location, error) {
-	var (
-		ret       = pc
-		btoffset  int64
-		locations []Location
-		retaddr   uintptr
-	)
+func (dbp *Process) curFrameInfo(pc, sp uint64) (Stackframe, error) {
```
derekparker: Rename to just `frameInfo`.

proc/stack.go
```diff
@@ -49,42 +55,51 @@ func (n NullAddrError) Error() string {
...
-		f, l, fn = dbp.goSymTable.PCToLine(ret)
-		if fn == nil {
+		frames = append(frames, frame)
+		if frame.Ret <= 0 {
 			break
 		}
-		locations = append(locations, Location{PC: ret, File: f, Line: l, Fn: fn})
 		// Look for "top of stack" functions.
-		if fn.Name == "runtime.rt0_go" {
+		if frame.Fn.Name == "runtime.goexit" {
```
derekparker: Combine the two branches with `||`.

proc/threads.go
```diff
@@ -107,12 +107,12 @@ func (thread *Thread) Step() (err error) {
 // of the corresponding source code, the function we're in
 // and the current instruction address.
 func (thread *Thread) Location() (*Location, error) {
-	pc, err := thread.PC()
```
derekparker: Why get rid of the shorthand function?

proc/variables.go
```diff
@@ -289,8 +303,10 @@ func (thread *Thread) PackageVariables() ([]*Variable, error) {
 	return vars, nil
 }
 
-func (thread *Thread) EvalPackageVariable(name string) (*Variable, error) {
-	reader := thread.dbp.DwarfReader()
+func (dbp *Process) EvalPackageVariable(name string) (*Variable, error) {
+	reader := dbp.DwarfReader()
+	pc, _ := dbp.CurrentThread.PC()
```
derekparker: Do we need PC here? Can we get away with just passing `0`?

derekparker:
I'm unable to set a breakpoint:

```shell
(dlv) b main.main
Command failed: could not find FDE for PC 0x7fff65886000
(dlv) exit
```

proc/threads.go
aarzilli: no idea, probably went through a revision where I was using another register there for some reason.

aarzilli:
it should be fixed now.

derekparker:
I really like the `EvalScope` abstraction. However, it seems like the client is the wrong level to store this state. To me, the ideal place would be the `Debugger (service/debugger)`. That way the responsibility of managing this state is not the responsibility of each client implementation. Clients should remain dumb and stateless as much as possible. Also, if the state of the current `EvalScope` is being kept around, why pass scope into the functions, as opposed to have the assume the current `EvalScope` is stored in the state of the debugger? The benefit we get from your current approach (passing scope in) is the ability to specify scope without setting state (on client or debugger) however it doesn't seem to be used that way (everything assumes `*c.evalScope()` will be passed in - unless I missed a few things).

What are your thoughts?

aarzilli:
The reason I did it this way is so that an hypotetical UI would be able to print the value of a variable on a goroutine other than the current one without having to switch to it first. I was thinking about a GUI that wants to update a list of variables every time continue/next/step returns.

A simple client does not have to keep track of any state, passing EvalScope{ -1, 0 } will work just as well, I did it in service/rpc/client.go only because it's convenient and it avoids a few roundtrips.

If you insist on having EvalScope be part of debugger I will move it there, in that case will it be acceptable to add an API call to get the current EvalScope, plus a second API call to set the current frame (the current goroutine being always set by SwitchGoroutine)?

derekparker:
Can we query that state instead of (essentially) caching it somewhere? In other words, instead of `c.setEvalState()`, anything that needs an eval state can just query for the current state, and for the cases where the eval state is different from the current, we can just create a new one and pass it along to  the eval methods? Does that make sense?

I want to avoid any state being held on the client, and I'd like to avoid having to maintain state at all if we can get away with it.

aarzilli:
>Can we query that state instead of (essentially) caching it somewhere?

If we move the EvalScope member variable to debugger then yes, otherwise no. Serverside has no concept of "current frame", it doesn't need it. There is also the problem of evaluating variables on goroutines that aren't running, I would have to add a `CurrentGoroutine` separate from `CurrentThread` (and it would have to be reset to CurrentThread.GetG after every next/continue).

> I want to avoid any state being held on the client

Then maybe we shouldn't copy gdb's interface, the commands `goroutine` and `frame` are inherently stateful, imho this state also belongs into the client since it is a UI detail.
We could add flags to pick a goroutine/frame to args, locals and print, I don't like this solution at all since it's cumbersome to always have to type it out and it will create parsing ambiguities if we ever want to handle arbitrary expressions on `print`.

derekparker:
> the commands goroutine and frame are inherently stateful

I agree that the commands are stateful, however I think there's a difference between storing that state and deriving that state (CFA, SP - essentially).

> imho this state also belongs into the client since it is a UI detail

But the json-rpc client that Delve provides isn't the only client. The `Debugger` type/package is going to be common amongst all client/server implementations since it will always be used by the server. That's why it seems like the spot to put this shared state.

I understand this is a stateful interaction, since setting the frame / goroutine only modifies the context which variables are evaluated against. The thing I want to avoid is having to babysit this state in the code (i.e. the `c.setEvalState` calls after the debugger taking an action).

aarzilli:
>When yesterday I was rebasing my other PRs I noticed some test failures in proc that I could also reproduce with master.
Some of the errors were in TestNextConcurrent and looked like this:
    proc_test.go:326: Program did not continue to correct next location expected 10 was sys_linux_amd64.s:289

Today I investigated them a bit and it seems to me that the bug(s?) where introduced by 288e2036f6bb6aa0959cf8edf4dce3768ef26675 in proc.(*Process).next.
In particular:

1. we only check whether one of the running threads has our goroutine but our goroutine running doesn't mean that we hit one of our temp breakpoints, another breakpoint could have been hit

2. we call setNextBreakpoints for all the running threads, I'm not sure I understand why we are interested in setting breakpoints for goroutines other than the one we are next'ing through. Furthermore this exacerbates (1), the more breakpoints we set the likelier it is that we catch our goroutine running on a thread but stopped in a random place

3. calling Continue on some of the threads while we haven't yet determined if we found our temp breakpoint will result on having half the threads running after we call 'next', I'm not sure this is intended and the rest of the code can deal with this state.

This pull request fixes those problems and the test failures seem to be greatly reduced (I can get rid of those other errors by disabling hardware breakpoints, imho those don't work for some fundamental reason and they should probably be disable as they aren't worth the trouble), however I wonder whether I may be missing something given that the comment explicitly mentions setting temp breakpoints for every thread every time trapWait returns.

But the fact that our UI has a stateful implementation of this does not mean that all UI should implement this in a stateful way.

aarzilli:
> But the json-rpc client that Delve provides isn't the only client. The Debugger type/package is going to be common amongst all client/server implementations since it will always be used by the server

But the fact that our UI has a stateful implementation for this does not mean that all UI will implement this in a stateful way, I wouldn't expect a GUI to do it, for example.

derekparker:
Let me loop in @dlsniper and @ignatov since they're working on an IDE client for any thoughts on this.

dlsniper:
Hi, thanks for CC-ing. So, if I understand this right, the PR should add the ability to "watch" a variable that I'm setting from the client, correct?

If so, then I'd say that when the frame / goroutine changes a refresh action should be performed from the client to get the state of the watched variables. I've yet to check how IDEA does this part in particular but I don't expect to be a problem for the IDE to request the state for the watches it has (ideally the API should accept a list to refresh). If I've got this wrong, @ignatov please correct me if so.

Aside from IDEA itself, if I were to have a web ui for example, I don't see a reason why that wouldn't be able to store the list of watched variables and trigger a refresh on those when the user changes the goroutine / frame.

If I'm wrong about the assumption, I'm sorry long day + missing description / example and my mental model is not yet familiar with all the terminology and delve itself. Can you explain it in layman terms?

Thank you.

derekparker:
@dlsniper essentially the conversation is mostly around where the EvalState should be kept, on the Debugger struct or on the client.

Essentially, will the GUI plugin need to keep this state around? In other words, will it just request "let me see the state of the world N frames up" and be finished, or will it request to move the context into a different frame?

If only the former, than we can keep the state on client as it is right now, I just want to prevent each client having to maintain this state, but maybe it's a non-issue.

aarzilli:
@dlsniper this PR lets a delve client extract the value of a variable that isn't on the topmost frame of the stack or is on the stack of a goroutine that isn't currently attached to a thread.

Reading the value of a variable on the topmost frame of a goroutine mapped to a thread remains essentially the same. 

The point we are discussing is whether it's preferrable to do something like this:

```
EvalVariable(EvalScope{ <goroutine id>, <frame number> }, <variable name>)
```

or this:

```
SwitchGoroutine(<goroutine id>)
SwitchFrame(<frame id>)
EvalVariable(<variable name>)
```

service/rpc/client.go
```diff
@@ -15,6 +15,7 @@ type RPCClient struct {
 	addr       string
 	processPid int
 	client     *rpc.Client
+	evalScope  api.EvalScope
```
derekparker: To be specific, this is the piece of state I'd like to see if we can get rid of. I _really_ _really_ like the EvalScope abstraction and I like the way the rest of the code is developed such that they are more generalized.

The issue to me is, we're keeping this state here on `c.evalScope`, and is this state that other clients will have to manage as well? If so, let's move it somewhere higher (i.e. the Debugger struct). If this is state that is only going to be needed by this client, essentially only for the terminal, then fine.

Does this make sense?

aarzilli:
>The issue to me is, we're keeping this state here on c.evalScope, and is this state that other clients will have to manage as well?

In general no, passing EvalScope{ -1, 0 } should work.

ignatov:
As for me, the variant `EvalVariable(EvalScope{ <goroutine id>, <frame number> }, <variable name>)` is preferable.

derekparker:
That makes sense. Alright, I'm going to be traveling today but my hope is
that tomorrow I will be available and we can get this merged in.

terminal/command.go
```diff
@@ -199,12 +201,91 @@ func goroutines(client service.Client, args ...string) error {
...
+
+		for i := range gs {
+			if gs[i].ID == client.EvalScope().GoroutineID {
+				g = gs[i]
+				break
+			}
+		}
+	}
+
+	fmt.Printf("Thread %s\nGoroutine %s Frame %d\n", formatThread(state.CurrentThread), g, client.EvalScope().Frame)
```
derekparker: I think you want `formatGoroutine(g)` here. I'm getting printf formatting errors in the console.

aarzilli: fixed.

terminal/command.go
```diff
@@ -174,7 +176,7 @@ func threads(client service.Client, args ...string) error {
 
 func thread(client service.Client, args ...string) error {
 	if len(args) == 0 {
-		return fmt.Errorf("you must specify a thread")
+		return printscope(client)
```
derekparker: This line break the test in `command_test.go:55` names `TestCommandThread`.

derekparker: Is the reasoning for this change that `thread` should just return current thread info if one is not specified? Because this function allows you to switch threads.

aarzilli: Yes, that's the reason.

derekparker: Would you mind reverting that and submitting it as a separate PR? I'd like to keep this patch focused on the main feature.

aarzilli: done

proc/variables.go
```diff
@@ -520,16 +512,16 @@ func (thread *Thread) extractValueInternal(instructions []byte, addr int64, typ
 	case *dwarf.StructType:
 		switch {
 		case t.StructName == "string":
-			return thread.readString(ptraddress)
+			return scope.Thread.readString(ptraddress)
 		case strings.HasPrefix(t.StructName, "[]"):
-			return thread.readSlice(ptraddress, t, recurseLevel)
+			return scope.readSlice(ptraddress, t, recurseLevel)
```
derekparker: Why have some methods on scope and some still on the thread attached to scope?

aarzilli: readSlice calls extractValue therefore it needs scope, everything I could leave on thread I left there, readIntRaw doesn't need a scope, just an address.

When the code that calculates addresses is separated from the code that reads memory it will be possible to clean the interface further (the methods reading memory can all be on Thread, the ones calculating addresses need EvalScope).

derekparker:
This is looking really good thanks for your great work @aarzilli, I just have one more question / thought to throw out there.

Is there benefit to changing context in a stateful way (i.e. `frame 1` changing your eval scope via frame, or `goroutine N` changing eval scope to goroutine) as opposed to using these commands in a one-off fashion such as: `frame 1 locals` or `goroutine 12 p foo`? That way we could still use the awesome EvalScope abstraction without having to set state on the client. Also, for what it's worth I'm not hung up on the stateful-ness of the client implementation per-se anymore. I realize that a GUI/IDE will be most likely making one-off requests like I've suggested above.

Again, I just wanted to pose that question before merging this in and I'd appreciate thoughts / feedback.

aarzilli:
>Is there benefit to changing context in a stateful way (i.e. frame 1 changing your eval scope via frame, or goroutine N changing eval scope to goroutine) as opposed to using these commands in a one-off fashion

I did it this way because gdb does it this way and people are familiar with it. I think there are pros and cons for both approaches, With the stateful UI you probably end un typing less on most workflows and it's more common, with stateless UI you can't get confused by being on the wrong state.

I think the goroutine command should stay and I think it should always work on `next`, that is:

```
goroutine 10
next
```

should always move to the next instruction in goroutine 10, even if goroutine 10 isn't attached to a thread, with the latests changes you made to proc.(*Process).Next it's not even that big of a change.

For the `frame` command I think it would also be good as a prefix to other commands, as you describe, prehaps better.

>Again, I just wanted to pose that question before merging this in and I'd appreciate thoughts / feedback.

No problem, it's your project.

derekparker:
> I think the goroutine command should stay and I think it should always work on next
> ...
> with the latests changes you made to proc.(*Process).Next it's not even that big of a change

I agree. I think what makes sense is to make `frame` stateless, and provide a useful error for when a command is not provided, either that or just print information for the previous frame by default (maybe file:line + locals?). And the `goroutine` command can take an id _only_ and then switch to that ID, or take an ID + command and simply execute command within that context _without_ actually permanently switching.

Thoughts? I'd really like feedback on usability and what would be the most user friendly / useful approach (without conforming to GDB).

aarzilli:
Sounds good, I'll add a commit for this.

aarzilli:
I left the commits unsquashed for now.

derekparker:
Going to take another look at this today.

proc/proc.go
```diff
@@ -412,13 +412,39 @@ func (dbp *Process) Step() (err error) {
 
 // Change from current thread to the thread specified by `tid`.
 func (dbp *Process) SwitchThread(tid int) error {
+	fmt.Printf("SwitchThread: %d\n", tid)
+	if dbp.CurrentThread != nil && dbp.CurrentThread.Id == tid {
```
derekparker: Let's drop this branch and just let the map lookup below happen, and let `SelectedGoroutine` get set that way.

proc/proc.go
```diff
@@ -412,13 +412,39 @@ func (dbp *Process) Step() (err error) {
...
+	for _, g := range gs {
+		if g.Id == gid {
+			if g.thread != nil {
+				return dbp.SwitchThread(g.thread.Id)
+			}
+			dbp.SelectedGoroutine = g
+			return nil
+		}
+	}
+	return fmt.Errorf("Could not find goroutine %d", gid)
```
derekparker: s/Could/could/

proc/proc.go
```diff
@@ -580,21 +601,26 @@ func initializeDebugProcess(dbp *Process, path string, attach bool) (*Process, e
...
+		return nil, err
+	}
+
 	ver, isextld, err := dbp.getGoInformation()
 	if err != nil {
 		return nil, err
 	}
 
 	dbp.arch.SetGStructOffset(ver, isextld)
+	// SelectedGoroutine can no be set correctly by the call to updateThreadList
```
derekparker: s/can no/can not/
s/without a SetGStructOffset/without calling SetGStructOffset/

proc/proc.go
```diff
@@ -699,3 +725,50 @@ func (dbp *Process) getGoInformation() (ver GoVersion, isextld bool, err error)
...
+func (dbp *Process) ConvertEvalScope(gid, frame int) (*EvalScope, error) {
+	var g *G
+
+	if gid == -1 {
+		g = dbp.SelectedGoroutine
+		if g == nil {
+			return dbp.CurrentThread.Scope()
+		}
+	} else {
+		gs, err := dbp.GoroutinesInfo()
```
derekparker: Probably worth extracting this into a method off of `Process`, something like `FindGoroutine`.

proc/threads.go
```diff
@@ -280,6 +280,12 @@ func (thread *Thread) GetG() (g *G, err error) {
 		return nil, err
 	}
 
+	if thread.dbp.arch.GStructOffset() == 0 {
+		// GetG was called through SwitchThread / updateThreadList during initialization
+		// thread.dbp.arch isn't setup yet (it needs a CurrentThread to read global variables from)
+		return nil, nil
```
derekparker: This should really return an error, otherwise we have to check for `nil` on all calls to `GetG`.

derekparker: Was this put in so that `GetG` calls in `SwitchThread` would work fine, and essentially set `SelectedGoroutine` to nil if we don't have a proper `GStructOffset`?

aarzilli: SetGStructOffset needs a CurrentThread, to have a current thread we need to have called updateThreadList, which calls addThread, which calls SwitchThread, which calls GetG, which needs GStructOffset to be set.

derekparker:
The `frame` command causes Delve to panic.

terminal/command.go
```diff
@@ -202,18 +208,124 @@ func thread(client service.Client, args ...string) error {
...
+		switch fullargs[i] {
+		case "goroutine":
+			n, err := strconv.Atoi(fullargs[i+1])
+			if err != nil {
+				return fmt.Errorf("Invalid argument to goroutine, expected integer")
+			}
+			scope.GoroutineID = int(n)
+			i++
+		case "frame":
+			n, err := strconv.Atoi(fullargs[i+1])
```
derekparker: It's this line that is panic'ing with an "index out of range" error, when you invoke `frame` without any arguments.

proc/proc.go
```diff
@@ -580,21 +594,26 @@ func initializeDebugProcess(dbp *Process, path string, attach bool) (*Process, e
...
 	if err != nil {
 		return nil, err
 	}
 
 	dbp.arch.SetGStructOffset(ver, isextld)
+	// SelectedGoroutine can not be set correctly by the call to updateThreadList
+	// because without calling SetGStructOffset we can not read the G struct of CurrentThread
+	// but without calling updateThreadList we can not examine memory to determine
+	// the offset of g struct inside TLS
+	dbp.SwitchThread(dbp.CurrentThread.Id)
```
derekparker: Instead of calling `SwitchThread` for the side effect of setting `SelectedGoroutine` let's just set it as such:

`dbp.SelectedGoroutine, err = dbp.CurrentThread.GetG()`

That makes the intention of the code more clear and at that point we can probably do without the comments.

aarzilli:
fixed

derekparker:
Squashed and merged, thanks.

ThomasMoinel:
I think you miss a file in the PR, dlv 0.8.0beta does'nt work
```
(dlv) b main.main
Command failed: could not find FDE for PC 0x7fff63302028
```

applying this diff makes it work again

```diff
% git df master eval-scope service/debugger/debugger.go
diff --git a/service/debugger/debugger.go b/service/debugger/debugger.go
index 2c15e1b..c34f9c4 100644
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -502,10 +502,7 @@ func (d *Debugger) FindLocation(scope api.EvalScope, locStr string) ([]api.Locat
                return nil, err
        }

-       s, err := d.process.ConvertEvalScope(scope.GoroutineID, scope.Frame)
-       if err != nil {
-               return nil, err
-       }
+       s, _ := d.process.ConvertEvalScope(scope.GoroutineID, scope.Frame)

        locs, err := loc.Find(d, s, locStr)
        for i := range locs {
```

derekparker:
@ThomasMoinel what OS are you running on? I'm working through some fixes on OSX right now.

derekparker:
@aarzilli I'm getting the same error on OSX, any ideas, I know this had happened before. I tested this patch on Linux and everything was working fine, seems to be an issue on OSX.

derekparker:
@aarzilli `thread.Stacktrace` is returning an error, which is being called by `thread.Scope`.

ThomasMoinel:
Yes I am on OSX 10.9.5 go 1.4.2. and this diff resolve the issue for me.

derekparker:
@ThomasMoinel I'd rather not ignore the error though. @aarzilli, it seems that in `loc.Find` scope is only used for the PC (and only in 1 `Find` out of the several implementations). I think it would make more sense just to simply pass PC in instead of scope, which would prevent us from having to make the call to `ConvertScope`, which would prevent this error from happening in the first place. Any thoughts on that?

derekparker:
@ThomasMoinel can you pull again and try, should be fixed now.

ThomasMoinel:
@derekparker Yep its seems to be fixed, great, thx.

ThomasMoinel:
I spoke too quickly, TestBreakpointInSeperateGoRoutine-4 Fail (seen on travis-ci osx) and on my machine. but sometime it passes the test.

```
=== RUN TestBreakpointInSeperateGoRoutine-4
unexpected fault address 0xc200005b03
fatal error: fault
[signal 0xa code=0x2 addr=0xc200005b03 pc=0x4010b9b]

goroutine 20 [running]:
runtime.gothrow(0x43f0a30, 0x5)
	/Users/thomas/Programmation/go/src/runtime/panic.go:503 +0x8e fp=0xc2084076d0 sp=0xc2084076b8
runtime.sigpanic()
	/Users/thomas/Programmation/go/src/runtime/sigpanic_unix.go:23 +0x19e fp=0xc208407720 sp=0xc2084076d0
runtime.mallocgc(0x10, 0x43b4d80, 0xc200000000, 0xc2080ae240)
	/Users/thomas/Programmation/go/src/runtime/malloc.go:183 +0x8ab fp=0xc2084077d0 sp=0xc208407720
runtime.newobject(0x43b4d80, 0xc2080ae240)
	/Users/thomas/Programmation/go/src/runtime/malloc.go:353 +0x49 fp=0xc2084077f8 sp=0xc2084077d0
github.com/derekparker/delve/dwarf/reader.New(0xc2080b2600, 0x4005338)
	/Users/thomas/goros/src/github.com/derekparker/delve/dwarf/reader/reader.go:18 +0x42 fp=0xc208407820 sp=0xc2084077f8
github.com/derekparker/delve/proc.(*Process).DwarfReader(0xc20803e300, 0x8)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:527 +0x2c fp=0xc208407838 sp=0xc208407820
github.com/derekparker/delve/proc.parseG(0xc2089a2060, 0x2081c65a0, 0x0, 0x2081c65a0, 0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/variables.go:122 +0x515 fp=0xc208407a18 sp=0xc208407838
github.com/derekparker/delve/proc.(*Thread).GetG(0xc2089a2060, 0xc2089a2060, 0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/threads.go:295 +0x292 fp=0xc208407aa8 sp=0xc208407a18
github.com/derekparker/delve/proc.(*Process).SwitchThread(0xc20803e300, 0x5903, 0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:417 +0xad fp=0xc208407b28 sp=0xc208407aa8
github.com/derekparker/delve/proc.func·003(0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:378 +0xde fp=0xc208407ba0 sp=0xc208407b28
github.com/derekparker/delve/proc.(*Process).run(0xc20803e300, 0xc208407c90, 0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:672 +0x12e fp=0xc208407c30 sp=0xc208407ba0
github.com/derekparker/delve/proc.(*Process).Continue(0xc20803e300, 0x0, 0x0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:393 +0x276 fp=0xc208407d30 sp=0xc208407c30
github.com/derekparker/delve/proc.func·015(0xc20803e300, 0x4414c30, 0xb, 0xc2089a6be0, 0x45, 0xc2089a6f00, 0x4d)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc_test.go:200 +0x27d fp=0xc208407df0 sp=0xc208407d30
github.com/derekparker/delve/proc.withTestProcess(0x4414c30, 0xb, 0xc208040510, 0xc208407f60)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc_test.go:39 +0x2d5 fp=0xc208407f40 sp=0xc208407df0
github.com/derekparker/delve/proc.TestBreakpointInSeperateGoRoutine(0xc208040510)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc_test.go:214 +0x5b fp=0xc208407f78 sp=0xc208407f40
testing.tRunner(0xc208040510, 0x45cf500)
	/Users/thomas/Programmation/go/src/testing/testing.go:447 +0xbf fp=0xc208407fd0 sp=0xc208407f78
runtime.goexit()
	/Users/thomas/Programmation/go/src/runtime/asm_amd64.s:2232 +0x1 fp=0xc208407fd8 sp=0xc208407fd0
created by testing.RunTests
	/Users/thomas/Programmation/go/src/testing/testing.go:555 +0xa8b

goroutine 1 [chan receive]:
testing.RunTests(0x44a7bb0, 0x45cf4a0, 0x1f, 0x1f, 0x2aa3fb5d482bf001)
	/Users/thomas/Programmation/go/src/testing/testing.go:556 +0xad6
testing.(*M).Run(0xc20802e320, 0x0)
	/Users/thomas/Programmation/go/src/testing/testing.go:485 +0x6c
github.com/derekparker/delve/proc/test.RunTestsWithFixtures(0xc20802e320, 0xc20802e320)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/test/support.go:59 +0x33
github.com/derekparker/delve/proc.TestMain(0xc20802e320)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc_test.go:24 +0x28
main.main()
	github.com/derekparker/delve/proc/_test/_testmain.go:110 +0x1d1

runtime.goexit()
	/Users/thomas/Programmation/go/src/runtime/asm_amd64.s:2232 +0x1

goroutine 6 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc20803e000)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:684 +0x5d
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:69 +0x218

goroutine 15 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc20803e180)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:684 +0x5d
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:69 +0x218

goroutine 19 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc20803e0c0)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:684 +0x5d
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:69 +0x218

goroutine 37 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc20803e240)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:684 +0x5d
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:69 +0x218

goroutine 21 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc20803e300)
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:684 +0x5d
	/Users/thomas/goros/src/github.com/derekparker/delve/proc/proc.go:69 +0x218
make: *** [test] Error 2
```

And TestNextConcurrent-4 sometimes hang more than 2min.

derekparker:
@ThomasMoinel yeah, those are flickers that I'm trying to fix now. Been trying to track them down for a little while now, think I may have a solution soon.

aarzilli:
Regarding the FindLocation error, sometimes programs on OS X (I think when cgo is involved?) start at an address that can not be resolved to a location.Only offset location specifiers and file-relative location specifiers use the EvalScope, the error returned by ConvertEvalScope should be ignored there because those two code branches will fail anyway if we can not lookup PC and give a more specific error.

aarzilli:
The TestBreakpointInSeperateGoRoutine problem looks like corrupted memory to me, which is weird. I would understand if the memory of the target got corrupted but how does delve corrupt its own memory? Maybe the problem is in the C parts of delve.

derekparker:
That's a good thought. I'll look into that.

PR #207, opened 2015-08-12
stack command, added -full flag

aarzilli:
When -full is passed all arguments and local variables of functions on the stacktrace are printed

### DIFF
--- a/proc/stack.go
+++ b/proc/stack.go
@@ -14,9 +14,16 @@ func (nra NoReturnAddr) Error() string {
 }
 
 type Stackframe struct {
-	Location
-	CFA int64
-	Ret uint64
+	// Address the function above this one on the call stack will return to
+	Current Location
+	// Address of the call instruction for the function above on the call stack.
+	Call Location
+	CFA  int64
+	Ret  uint64
+}
+
+func (frame *Stackframe) Scope(thread *Thread) *EvalScope {
+	return &EvalScope{Thread: thread, PC: frame.Current.PC, CFA: frame.CFA}
 }
 
 // Takes an offset from RSP and returns the address of the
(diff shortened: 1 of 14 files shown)


derekparker:
Can you rebase this PR on master instead of your previous PRs so I can see the changeset in isolation.

aarzilli:
It is not possible to implement this functionality without the previous commit.

derekparker:
Can you rebase this on master so I can review and possibly merge.

aarzilli:
done

derekparker:
Can you rebase this onto master, and fix test failures with the latest patch that was pulled in. I can review today.

proc/stack.go
```diff
@@ -14,9 +14,16 @@ func (nra NoReturnAddr) Error() string {
 }
 
 type Stackframe struct {
-	Location
-	CFA int64
-	Ret uint64
+	// Address the function above this one on the call stack will return to
+	Return Location
```
derekparker: I think Return is overloaded here. How about something like `CurrentLocation`, basically something that conveys "this is the location we will be at once we are back in this frame". I understand the motivation for `Return` since it's the location that the function invoked from this frame will return to, however just reading through the source code you could also think it means that it is the return address of this function, not the one above it. Does that make sense?

proc/stack.go
```diff
@@ -14,9 +14,16 @@ func (nra NoReturnAddr) Error() string {
...
+	// Address of the call instruction for the function above this one on the call stack
```
derekparker: Change to: "Address of the call instruction that invoked the function above on the call stack."

derekparker:
Can you actually split this into 2 PRs, one that handles the `-full` flag, and one that deals only with being more tolerant of uninitialized memory, etc.

proc/stack.go
aarzilli: Sure.

aarzilli:
Done. This PR now tracks only the changes to stacktraces, PR #238 the changes to `variables.go`. Note however that `stack -full` will not work for any sizeable program without PR #238 because of the increasing likelihood of encountering uninitialized memory.

derekparker:
Thanks, this is awesome.

One thing though is I'd like to switch the default numeric argument to `bt` to be the depth, and then allow `bt` as an argument to `goroutine <id>`. I don't like the `-<depth>` approach as much.

PR #210, opened 2015-08-17
Add config file. Add command alias config options.

### LINKED ISSUE #182: Command aliases

tylerb:
Fix #182 

# Overview
This PR adds support for a yaml configuration file. This file, if it does not exist, is created and populated with a default configuration. The default configuration is currently generated by iterating all available commands and outputting a commented line for aliasing the command.

This yaml configuration file is loaded into the new `config` struct, which can be extended in the future to provide additional configuration options through the configuration file. The file is located at the standard path, and is named `config.yaml`.

# Aliasing
The aliases defined in the yaml file are simply appended to the list of aliases for the "primary" command alias. For example: `breakpoints: ["breaks"]` results in `breakpoints, bp, breaks`. There are no checks to detect conflicting aliases.

A different approach I considered was to completely replace the alias array with the alias array provided in the config file, starting with the "primary" command. For example, `exit` has the aliases `quit,q`. If defined as `exit: "stop, s"` in the config file, the resulting set would be `exit,stop,s`. If this is the preferred behavior, let me know and I'll make the change.


terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
...
+aliases:
+  # example: ["alias1", "alias2"]
+
+`)
+
+	for _, cmd := range cmds.cmds {
+		buffer.WriteString(fmt.Sprintf("  # %s: []\n", cmd.aliases[0]))
+	}
+
+	_, err := f.WriteString(buffer.String())
```
derekparker: Could probably just use `buffer.WriteTo(f)` here.

terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
...
+	}
+
+	var c config
+	err = yaml.Unmarshal(data, &c)
+	if err != nil {
+		fmt.Printf("Unable to decode config file: %v.", err)
+		return nil
+	}
+
+	f.Close()
```
derekparker: This should be in a defer after a successful open.

terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
...
+	f, err := os.Open(fullConfigFile)
+	if err != nil {
+		f, err = os.Create(fullConfigFile)
+		if err != nil {
+			fmt.Printf("Unable to create config file: %v.", err)
+		} else {
+			err = writeDefaultConfig(f, cmds)
+			if err != nil {
+				fmt.Printf("Unable to write default configuration: %v.", err)
+				_, err = f.Seek(0, os.SEEK_SET)
```
derekparker: Why seek on error? We should just return and let defer close the file descriptor for us.

terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
...
+		fmt.Printf("Unable to load config file: %v.", err)
+	}
+
+	f, err := os.Open(fullConfigFile)
+	if err != nil {
+		f, err = os.Create(fullConfigFile)
+		if err != nil {
+			fmt.Printf("Unable to create config file: %v.", err)
+		} else {
+			err = writeDefaultConfig(f, cmds)
```
derekparker: I feel like, upon a successful default config creation we should just return. It's just the default config, currently with only commented out values, so we don't need to parse it or anything.

terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
...
+	fullConfigFile, err := getConfigFilePath(configFile)
+	if err != nil {
+		fmt.Printf("Unable to load config file: %v.", err)
+	}
+
+	f, err := os.Open(fullConfigFile)
+	if err != nil {
+		f, err = os.Create(fullConfigFile)
+		if err != nil {
+			fmt.Printf("Unable to create config file: %v.", err)
```
derekparker: We should just bail here as well. If we can't create the config file there isn't much more to do.

terminal/terminal.go
```diff
@@ -166,3 +182,68 @@ func getConfigFilePath(file string) (string, error) {
 	}
 	return path.Join(usr.HomeDir, configDir, file), nil
 }
+
+func loadConfig(cmds *Commands) *config {
+	fullConfigFile, err := getConfigFilePath(configFile)
+	if err != nil {
+		fmt.Printf("Unable to load config file: %v.", err)
```
derekparker: We should bail here if we can't get the path to the config file.

terminal/command.go
```diff
@@ -110,6 +110,17 @@ func (c *Commands) Find(cmdstr string) cmdfunc {
 	return noCmdAvailable
 }
 
+// Merge takes aliases defined in the config struct and merges them with the default aliases
```
derekparker: Period at end of comment.

derekparker:
I'm on board with the append approach.

tylerb:
Thanks for all the line comments. I'll get them taken care of!

derekparker:
Sounds good. Sorry for the delay.

terminal/terminal.go
tylerb: Looks like I put it in the error block when it should have been outside it.

terminal/terminal.go
tylerb: That makes sense. No reason to overcomplicate it since we're not loading anything.

tylerb:
Fixes pushed.

derekparker:
Looks good. The only other thing is I think it makes sense to move the code from `terminal/terminal.go` to `cmd/dlv/main.go`. The reasoning is because eventually there will likely be config options that effect the debugger itself, not just the terminal client. That way, if all that is handled in main, the config can be passed to clients / servers or whoever needs it. Thoughts?

tylerb:
Agreed. Much better place for it. I'll get that done soon.

tylerb:
@derekparker Done. I had to abstract the functionality of the config stuff from terminal.go to its own package that can be used throughout the various components. The config is now created and loaded in `main.go` and passed down to the various components.

Please let me know if anything needs to be cleaned up or changed.

derekparker:
The CI build is failing with:

```
cmd/dlv/main.go:113: not enough arguments in call to execute
make: *** [test] Error 2
```

cmd/dlv/main.go
```diff
@@ -28,6 +30,14 @@ var (
 )
 
 func main() {
+
+	// Config setup and load.
+	err := config.CreateConfigPath()
+	if err != nil {
+		fmt.Printf("Could not create config directory: %v.")
+	}
+	conf := config.LoadConfig()
```
derekparker: Maybe consolidate these steps (create / load) into just LoadConfig. That way, it can try and Stat the config dir or something, if it doesn't exist then create it, otherwise just load from it. Thoughts?

tylerb:
Odd. I will take a look.

cmd/dlv/main.go
tylerb: Sounds great to me. They are separate at the moment because they were in different places before, with CreateConfigPath existing before I added this new stuff. It makes sense to consolidate them now.

tylerb:
CI is failing, but I don't know why. On my local copy of the code, line 109 doesn't even have `execute` on it. Also, my local copy builds fine. Everything is committed and pushed. See https://github.com/tylerb/delve/blob/master/cmd/dlv/main.go#L109

@derekparker any idea?

derekparker:
On my copy, once I pull your branch and rebase on master, line 109 shows a `execute` function call within the `exec` subcommand that does not have config passed in. Try rebasing on master and you should see the same issue.

tylerb:
It's now building.

terminal/command.go
```diff
@@ -110,6 +110,17 @@ func (c *Commands) Find(cmdstr string) cmdfunc {
 	return noCmdAvailable
 }
 
+// Merge takes aliases defined in the config struct and merges them with the default aliases.
+func (c *Commands) Merge(allAliases map[string][]string) {
+	for primary, aliases := range allAliases {
```
derekparker: You can cut out the outermost loop with something like this:

```go
for i := range c.cmds {
        if aliases, ok := allAliases[c.cmds[i].aliases[0]]; ok {
		c.cmds[i].aliases = append(c.cmds[i].aliases, aliases...)
	}
}
```

derekparker:
Merged, thanks!

PR #214, opened 2015-08-21
Fixing proc.(*Process).Next

aarzilli:
When yesterday I was rebasing my other PRs I noticed some test failures in proc that I could also reproduce with master.
Some of the errors were in TestNextConcurrent and looked like this:
```proc_test.go:326: Program did not continue to correct next location expected 10 was sys_linux_amd64.s:289```

Today I investigated them a bit and it seems to me that the bug(s?) where introduced by 288e2036f6bb6aa0959cf8edf4dce3768ef26675 in proc.(*Process).next.
In particular:

1. we only check whether one of the running threads has our goroutine but our goroutine running doesn't mean that we hit one of our temp breakpoints, another breakpoint could have been hit

2. we call setNextBreakpoints for all the running threads, I'm not sure I understand why we are interested in setting breakpoints for goroutines other than the one we are next'ing through. Furthermore this exacerbates (1), the more breakpoints we set the likelier it is that we catch our goroutine running on a thread but stopped in a random place

3. calling Continue on some of the threads while we haven't yet determined if we found our temp breakpoint will result on having half the threads running after we call 'next', I'm not sure this is intended and the rest of the code can deal with this state.

This pull request fixes those problems and the test failures seem to be greatly reduced (I can get rid of those other errors by disabling hardware breakpoints, imho those don't work for some fundamental reason and they should probably be disable as they aren't worth the trouble), however I wonder whether I may be missing something given that the comment explicitly mentions setting temp breakpoints for every thread every time trapWait returns.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -272,18 +272,17 @@ func (dbp *Process) next() (err error) {
 		return err
 	}
 
-	var goroutineExiting bool
-	threadNext := func(thread *Thread) error {
-		if err = thread.setNextBreakpoints(); err != nil {
-			switch t := err.(type) {
-			case ThreadBlockedError, NoReturnAddr: // Noop
-			case GoroutineExitingError:
-				goroutineExiting = t.goid == g.Id
-			default:
-				return err
-			}
+	goroutineExiting := false
+
+	err = dbp.CurrentThread.setNextBreakpoints()
+	if err != nil {
+		switch err.(type) {
+		case ThreadBlockedError, NoReturnAddr: // Noop
+		case GoroutineExitingError:
+			goroutineExiting = true
+		default:
+			return err
 		}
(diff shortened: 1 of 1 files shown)


derekparker:
Thanks. There are some flickering tests (especially on CI right now) and something like this could help fix them. I'm doing some more work on the Next implementation right now, so I'll keep this PR around, however I may just pick from it instead of merging. I think we may be able to throw out the idea of every thread nexting, and simply be fine with an approach that sets breakpoints only on the current thread of execution you care about, and any code that executes in the meantime is acceptable since we halt the entire process once we find our goroutine again.

derekparker:
Closing in favor of: https://github.com/derekparker/delve/commit/eb0b4e93921f07638033403a7cc2e8be09b029f1

PR #229, opened 2015-09-10
Allow using list command with frame

icholy:


icholy:
@aarzilli should I be doing that instead?

aarzilli:
Not necessarily, I was just making the alternative known.

icholy:
@derekparker thoughts?

derekparker:
I'm fine with this approach. Merged.

The only concern I have is not with this patch, but with this code path in general. We're specifically white listing certain commands here, which could be aliased in other ways, specifically through configuration. Any user defined aliases for these commands would then not work. I haven't put much thought into a better solution so I'm open to ideas.

aarzilli:
>The only concern I have is not with this patch, but with this code path in general

We could have every command function get a context object as the first argument and stick EvalScope inside that.

PR #230, opened 2015-09-10
Sort threads and goroutines by ID

icholy:


icholy:
The output of `threads` is in a different order every time you run it. The `goroutines` output order is consistent, but I added sorting to those just to keep it consistent.

derekparker:
Thanks!

PR #236, opened 2015-09-14
Don't display dead goroutines

icholy:
@aarzilli do you think there are any use cases for displaying dead goroutines?

https://groups.google.com/forum/#!topic/delve-dev/b0bLJlmCo1Q

### DIFF
--- a/proc/variables.go
+++ b/proc/variables.go
@@ -37,6 +37,19 @@ type M struct {
 	curg     uintptr // Current G running on this thread.
 }
 
+const (
+	// G status, from: src/runtime/runtime2.go
+	Gidle            uint64 = iota // 0
+	Grunnable                      // 1 runnable and on a run queue
+	Grunning                       // 2
+	Gsyscall                       // 3
+	Gwaiting                       // 4
+	Gmoribund_unused               // 5 currently unused, but hardcoded in gdb scripts
+	Gdead                          // 6
+	Genqueue                       // 7 Only the Gscanenqueue is used.
+	Gcopystack                     // 8 in this state when newstack is moving the stack
+)
+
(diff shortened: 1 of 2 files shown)


aarzilli:
I can't think of any.

icholy:
We could also show the goroutine state as part of the output.

Example:

```
Goroutine 1 [BUSY] - /usr/local/go/src/runtime/asm_amd64.s:116 runtime.rt0_go (0x462473)
Goroutine 2 [IDLE] - /usr/local/go/src/runtime/proc.go:142 runtime.init.4 (0x462473)
Goroutine 3 [RUNNING] - /usr/local/go/src/runtime/mgc.go:207 runtime.gcenable (0x462473)
Goroutine 4 [RUNNING] - /usr/local/go/src/runtime/mfinal.go:137 runtime.createfing (0x462473)
Goroutine 5 [SYSCALL] - /usr/local/go/src/runtime/time.go:118 runtime.addtimerLocked (0x44315e)
```

proc/variables.go
```diff
@@ -37,6 +37,22 @@ type M struct {
 	curg     uintptr // Current G running on this thread.
 }
 
+const (
+	// From: src/runtime/runtime2.go
```
derekparker: Condense this comment to `G status, from: ...`.

Also, let's get rid of the `_` prefix, I don't think it is necessary to use the exact same constants from the Go source.

derekparker:
@icholy I definitely agree with your second point there, that we should display the Goroutine status as part of the output.

icholy:
@derekparker I've already started work on displaying the status. I'll open a PR soon.

icholy:
@derekparker done

derekparker:
Merged, thanks!

PR #237, opened 2015-09-15
What it looks like when building with gb.

jeffallen:
FOR REVIEW ONLY, do not commit yet.

Hey, Derek.

This is what delve looks like when it is gb-ified. Not pretty, I have to say. I don't really understand why we have to have src/github.com/derekparker in there. It seems gross.

Maybe I will ask Dave to confirm this is how it should be done.

gb test -c does not currently work right (there is an issue filed on it) so "make test" does not work yet.


kostya-sh:
Just want to leave comment here saying that using `go install` to install the latest version of delve is **extremely** useful. Please don't break it ;)

jeffallen:
My prototype makes it so that "make install" still does the same thing. But
the bottom line is that when you use "gb", you are not using "go install"
anymore.

derekparker:
@jeffallen sorry for not responding, just finished moving. Anyways, it seems GB integration with this project is blocked on https://github.com/constabulary/gb/issues/233 as you mentioned. We simply cannot break the test workflow on OSX.

More generally, I'm really interested in GB mostly for the vendoring solution. Some of the other conveniences are nice as well, and Delve could be well suited to a more "project" like workflow. My only concern is committing to something outside the normal Go toolchain. Although, with that being said, I could be convinced to lean more towards GB since even with the Go vendor experiment, you still need a tool to actually _do_ the vendoring, and after that is done the go toolchain will support looking at the vendor directory, it will not vendor for you.

The deeply nested `src/` dir does stick out to me, but from looking at other GB projects that seems to be the way things are.

jeffallen:
I need to look into the vendor experiment again. The gb-vendor plugin
probably leaves things in the right format, so maybe it would be better to
use gb vendor to maintain the directory, and "go build" to do the builds.

  -jeff

casimir:
Have you guys looked at [glide](https://github.com/Masterminds/glide)? It only manages vendoring and leaves the rest to the normal Go toolchain when `GO15VENDOREXPERIMENT=1`.

derekparker:
@jeffallen @casimir thank you both for the suggestions. I think the way forward, for now, is to use godep, but with support for `GO15VENDOREXPERIMENT`. That gives us the best of all worlds. Going to close this in favor of that. Thanks @jeffallen for your initial work here, though.

jeffallen:
:+1:

Issue #189, opened 2015-07-29
`info locals` crashes with "len out of range" error


joeshaw:
Running Delve 0.6.0.beta (a506bb7) and go version go1.4.2 darwin/amd64.  I was trying to reproduce this with a simpler program when I ran into #186, but haven't had success so far.  A fairly complicated program crashes when I run `info local` in the `main.main` function:

```
$ dlv run
Type 'help' for list of commands.
(dlv) b main.main
Breakpoint 1 set at 0xaa85 for main.main /Users/joeshaw/src/gosrc/src/<redacted>/main.go:643
(dlv) c
> main.main() /Users/joeshaw/src/gosrc/src/<redacted>/main.go:643

   638:		}
   639:	}
   640:
   641:	func main() {
   642:		// Override cli's hardcoded versioning
=> 643:		var showVersion bool
   644:		flag.BoolVar(&showVersion, "v", showVersion, "")
   645:		flag.BoolVar(&showVersion, "version", showVersion, "")
   646:		flag.Parse()
   647:
   648:		if showVersion {

(dlv) info locals
panic: runtime error: makeslice: len out of range

goroutine 21 [running]:
github.com/derekparker/delve/proc.(*Thread).readMemory(0xc2089fe060, 0x536567617373656d, 0x637365447465, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/joeshaw/src/gosrc/src/github.com/derekparker/delve/proc/variables.go:794 +0xb6
github.com/derekparker/delve/proc.(*Thread).readString(0xc2089fe060, 0x536567617373656d, 0x0, 0x0, 0x0, 0x0)
	/Users/joeshaw/src/gosrc/src/github.com/derekparker/delve/proc/variables.go:580 +0x4d7
github.com/derekparker/delve/proc.(*Thread).extractValueInternal(0xc2089fe060, 0x0, 0x0, 0x0, 0x7e9290, 0x4492060, 0xc208dabd40, 0x41d5b01, 0x1, 0x0, ...)
	/Users/joeshaw/src/gosrc/src/github.com/derekparker/delve/proc/variables.go:509 +0x70a
github.com/derekparker/delve/proc.(*Thread).extractValueInternal(0xc2089fe060, 0x0, 0x0, 0x0, 0x7e9270, 0x4492060, 0xc208daa480, 0x1, 0x0, 0x0, ...)
[...]

PR #238, opened 2015-09-17
Be more tolerant of uninitialized memory

aarzilli:
Attempting to read uninitialized memory does not crash the program, instead an effort is made to report as much readable memory as possible. String length is checked before attempting to read the associated byte array and length of a string is limited to maxArrayValues.

Workaround for Issue #189 (and others)

### DIFF
--- a/proc/variables.go
+++ b/proc/variables.go
@@ -17,6 +17,7 @@ import (
 const (
 	maxVariableRecurse = 1  // How far to recurse when evaluating nested types.
 	maxArrayValues     = 64 // Max value for reading large arrays.
+	maxErrCount        = 3  // Max number of read errors to accept while evaluating slices, arrays and structs
 
 	ChanRecv = "chan receive"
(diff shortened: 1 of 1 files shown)


proc/variables.go
```diff
@@ -595,6 +602,14 @@ func (thread *Thread) readString(addr uintptr) (string, error) {
 		return "", fmt.Errorf("could not read string len %s", err)
 	}
 	strlen := int(binary.LittleEndian.Uint64(val))
+	if strlen < 0 {
+		return "", fmt.Errorf("invalid length: %d", strlen)
+	}
+
+	count := strlen
+	if count > maxArrayValues {
+		count = maxArrayValues - 1
```
derekparker: Why `- 1` here?

aarzilli: mistake.

derekparker:
Merged.

PR #242, opened 2015-09-20
Refactor: use FindGoroutine in SwitchGoroutine and Stacktrace

aarzilli:
FindGoroutine didn't exist when we wrote those.


derekparker:
Merged.

PR #248, opened 2015-09-21
proc/variables: Split address calculations from value extraction

aarzilli:
Refactored variables.go to separate calculation of a variable's address from
reading its value. This change is useful to implement the 'set' command
as well as the evaluation of more complex expressions (in the future).
Introduces the types VarAddr (a typed address in the target program) and
ArrayAddr (a typed sequence of addresses in the target program).


derekparker:
Took a glance through and it looked good, will do a full review this evening.

proc/variables.go
```diff
@@ -80,6 +80,48 @@ type EvalScope struct {
 	CFA    int64
 }
 
+// Represents the address and type information for a variable, the thread field is used to read the variable's value
+type VarAddr struct {
```
derekparker: I had some thoughts on these new types. How about putting this information in `Variable` (as you seemed to hint at in your proposal here (https://groups.google.com/forum/#!topic/delve-dev/Z0lDomkS9LI). I think it is completely reasonable for a variable to know its address, dwarfType, etc.. Also, it seems reasonable for a variable to know what thread it's on. I don't mind paying the cost of evaluating the variable either even if all we want to do is set it (as your other patch would allow), but if it becomes an issue or too expensive then we can investigate passing around flags to prevent actual evaluation, but still return a struct with all other fields filled in, however that seems dangerous (you would assume the `Value` field would be filled in).

That way, in your patch to allow setting variables, you could implement the methods on `Variable` instead of these new types. It also seems very reasonable for a variable to know how to change its value.

Thoughts?

proc/variables.go
```diff
@@ -80,6 +80,48 @@ type EvalScope struct {
...
+func newVarAddr(name string, addr uintptr, dwarfType dwarf.Type, thread *Thread) *VarAddr {
+	return &VarAddr{
+		Name:      name,
+		Addr:      addr,
+		dwarfType: dwarfType,
+		thread:    thread,
+	}
+}
+
+func (addr *VarAddr) toField(field *dwarf.StructField) *VarAddr {
```
derekparker: I really like this.

proc/variables.go
```diff
@@ -733,19 +759,21 @@ func (thread *Thread) readComplex(addr uintptr, size int64) (string, error) {
...
 	}
-	i, err := thread.readFloat(addr+uintptr(fs), fs)
+	imagaddr := *addr
+	imagaddr.Addr += uintptr(fs)
+	i, err := imagaddr.readFloat(fs)
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("(%s, %si)", r, i), nil
+	return fmt.Sprintf("(%s + %si)", r, i), nil
```
derekparker: Nice catch, I should've seen the incorrect notation before.

derekparker:
Can you update this with what was mentioned in the mailing list? It'd be great to see how this would look if everything was on Variable instead of the new types.

derekparker:
And actually, if we go the route of merging these types, do we even need this patch to implement the `set` command? The way I see it, maybe this patch could include a boolean to say whether or not to actually evaluate the variable vs. just gather information about it, but other than that all of the heavy lifting should be implemented by methods on the Variable type (i.e. `variable.SetValue` or something).

aarzilli:
Removed VarAddr and ArrayAddr.

proc/variables.go
```diff
@@ -80,6 +89,46 @@ type EvalScope struct {
...
+
+		if t.Count > 0 {
+			v.stride = t.ByteSize / t.Count
+		}
+	}
+
+	return v, nil
+}
+
+func (addr *Variable) toField(field *dwarf.StructField) (*Variable, error) {
```
derekparker: Should be `func (v *Variable)`.

proc/variables.go
```diff
@@ -241,21 +290,69 @@ func parseG(thread *Thread, gaddr uint64, deref bool) (*G, error) {
 	return g, nil
 }
 
+// Returns the address of the named variable
+func (scope *EvalScope) EvalVariableAddr(name string) (*Variable, error) {
```
derekparker: Maybe rename this to `ExtractVariableInfo` since it returns an initialized variable, with all info except for the actual value.

proc/variables.go
```diff
@@ -340,80 +428,40 @@ func (dbp *Process) EvalPackageVariable(name string) (*Variable, error) {
...
+			return scope.extractVarAddrFromEntry(entry, reader)
 		}
 	}
-
 	return nil, fmt.Errorf("could not find symbol value for %s", name)
 }
 
-func (scope *EvalScope) evaluateStructMember(parentEntry *dwarf.Entry, rdr *reader.Reader, memberName string) (*Variable, error) {
-	parentAddr, err := scope.extractVariableDataAddress(parentEntry, rdr)
+func (addr *Variable) structMember(memberName string) (*Variable, error) {
```
derekparker: Same comment as above.

proc/variables.go
```diff
@@ -340,80 +428,40 @@ func (dbp *Process) EvalPackageVariable(name string) (*Variable, error) {
...
 	}
-
-	return nil, fmt.Errorf("%s has no member %s", parentName, memberName)
 }
 
-// Extracts the name, type, and value of a variable from a dwarf entry
-func (scope *EvalScope) extractVariableFromEntry(entry *dwarf.Entry) (*Variable, error) {
+// Extracts the name and type of a variable from a dwarf entry
+// then executes the instructions given in the  DW_AT_location attribute to grab the variable's address
+func (scope *EvalScope) extractVarAddrFromEntry(entry *dwarf.Entry, rdr *reader.Reader) (*Variable, error) {
```
derekparker: Same as above, rename to `extractVarInfoFromEntry`.

derekparker:
Mostly naming cleanup. Rename variables named `addr` and change the function names.

aarzilli:
done

derekparker:
Thanks!

aarzilli:
I'm going to do the change to Variable I proposed in ml, if you don't have objections.

derekparker:
@aarzilli I'm all for it.

PR #249, opened 2015-09-21
Implements set command for pointers and numerical values

aarzilli:
Changing the value of strings requires an allocation which probably requires the ability to execute arbitrary code in the target program. This PR only implements setting numerical variables and pointers (requires PR #248)


derekparker:
Merged, this is awesome - thanks!

PR #254, opened 2015-09-25
Support for dumb terminal (fixes #252)

### LINKED ISSUE #252: don't print ANSI escape codes when TERM=dumb
For those of us who don't use ANSI terminal codes, it would be nice if the output was plain text when TERM=dumb. See attached image.

untitled

dlsniper:
I'm not sure if this is the best way to fix it, the other way would be to have a flag, `--no-color` for example.
Let me know which version is preferred, this or the flag and I'll adjust accordingly.
Thank you.

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -47,6 +47,8 @@ type Commands struct {
 	client  service.Client
 }
 
+var dumbTerminal = strings.ToLower(os.Getenv("TERM")) == "dumb"
+
 // Returns a Commands struct with default commands defined.
 func DebugCommands(client service.Client) *Commands {
 	c := &Commands{client: client}
@@ -724,7 +726,11 @@ func printcontext(state *api.DebuggerState) error {
 	}
 	if len(state.CurrentThread.File) == 0 {
 		fmt.Printf("Stopped at: 0x%x\n", state.CurrentThread.PC)
-		fmt.Printf("\033[34m=>\033[0m    no source available\n")
+		if dumbTerminal {
+			fmt.Printf("=>    no source available\n")
+		} else {
+			fmt.Printf("\033[34m=>\033[0m    no source available\n")
+		}
 		return nil
 	}
 	var fn *api.Function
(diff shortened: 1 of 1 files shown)


terminal/command.go
```diff
@@ -724,7 +726,11 @@ func printcontext(state *api.DebuggerState) error {
 	}
 	if len(state.CurrentThread.File) == 0 {
 		fmt.Printf("Stopped at: 0x%x\n", state.CurrentThread.PC)
-		fmt.Printf("\033[34m=>\033[0m    no source available\n")
+		if dumbTerminal {
```
derekparker: I have an idea for a refactor here.

Let's change all `fmt.Printf` calls to `term.Println(fmt.Sprintf("..."), true)` where the signature is `Term.Println(format string, currentLine bool)` and let's store the state of whether this is a dumb terminal or not in `Term`. Then, this function will turn into `printcontext(term, state)`. We have to do some work to pass the terminal into commands in `terminal/commands.go`, but I think overall it will be cleaner. Possibly instead of expanding the signature for commands in terminal/commands.go we can create a new type which will be the only argument to those functions, and that way it can be expanded in the future if need be without changing a lot of code.

The argument `currentLine` above will determine whether the arrow is printed. Also, instead of continuing to hard code the escape codes it would be best to at the very least store in a variable, or pluck them from https://github.com/golang/crypto/blob/master/ssh/terminal/terminal.go#L24 by just importing that terminal package and using the pre-defined escape codes there.

Thoughts?

dlsniper: It makes sense, as it will allow for further improvements / usages as well. I'll try to do the refactoring tomorrow. Thanks for feedback.

derekparker:
@dlsniper any chance you can get to this refactor tomorrow? If not I can go ahead and take care of it tomorrow evening.

terminal/command.go
derekparker: One thing about the refactor, in the example function signature I gave above, `Term.Println(format string, currentLine bool)`, `format` should just be named `str` or something, it should be a pre-formatted string.

derekparker:
Merged, went ahead with refactor myself to push this fix through. Thanks for the original patch though!

PR #255, opened 2015-09-26
proc/breakpoint Add breakpoint statistics

### LINKED ISSUE #247: Display thread ID / goroutine ID / hit count when hitting a tracepoint
It would be useful to know some additional information when hitting a tracepoint:
- thread ID
- goroutine ID
- hit count -> it should tell you how many times that tracepoint has been hit (preferably in the context of the goroutine)
- total hit count -> like above but regardless of the goroutine.

dlsniper:
This adds support for breakpoints statistics

Fixes #247

### DIFF
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -42,6 +42,10 @@ type Breakpoint struct {
 	Goroutine bool `json:"goroutine"`
 	// variables to evaluate
 	Variables []string `json:"variables,omitempty"`
+	// number of times a breakpoint has been reached in a certain goroutine
+	HitCount map[string]uint64 `json:"hitCount"`
+	// number of times a breakpoint has been reached
+	TotalHitCount uint64 `json:"totalHitCount"`
 }
 
 // Thread is a thread within the debugged process.
--- a/proc/breakpoints.go
+++ b/proc/breakpoints.go
@@ -17,14 +17,16 @@ type Breakpoint struct {
 	Temp         bool   // Whether this is a temp breakpoint (for next'ing).
 
 	// Breakpoint information
-	Tracepoint bool     // Tracepoint flag
-	Stacktrace int      // Number of stack frames to retrieve
-	Goroutine  bool     // Retrieve goroutine information
-	Variables  []string // Variables to evaluate
+	Tracepoint    bool           // Tracepoint flag
+	Stacktrace    int            // Number of stack frames to retrieve
+	Goroutine     bool           // Retrieve goroutine information
+	Variables     []string       // Variables to evaluate
+	HitCount      map[int]uint64 // Number of times a breakpoint has been reached in a certain goroutine
+	TotalHitCount uint64         // Number of times a breakpoint has been reached
 }
 
 func (bp *Breakpoint) String() string {
-	return fmt.Sprintf("Breakpoint %d at %#v %s:%d", bp.ID, bp.Addr, bp.File, bp.Line)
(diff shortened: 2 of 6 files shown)


dlsniper:
I'd like to work a bit more on this, such as adding the information about how many times a breakpoint was hit in a certain goroutine but before that I'd like to hear about, if it would be a good idea or not.

dlsniper:
@derekparker I have two things about this PR:
- can you please point me to how to create the tests for it?
- in the following sample:
```go
package main

import (
	"fmt"
	"math/rand"
	"sync"
	"time"
)

func demo(id int, wait *sync.WaitGroup) {
	for i := 0; i < 100; i++ {
		sleep := rand.Intn(10) + 1
		fmt.Printf("id: %d step: %d sleeping %d\n", id, i, sleep)
		time.Sleep(time.Duration(sleep) * time.Millisecond)
	}

	wait.Done()
}

func main() {
	wait := new(sync.WaitGroup)
	wait.Add(1)
	wait.Add(1)
	go demo(1, wait)
	go demo(2, wait)

	wait.Wait()
}

```
setting a tracepoint on line 12, `sleep := rand.Intn(10) + 1` would show that the tracepoint doesn't necessarily always show when the line is reached. For example, I have something like this on a test run:
```
id: 1 step: 99 sleeping 7
id: 2 step: 98 sleeping 1
> main.demo() /home/florin/golang/src/github.com/dlsniper/u/dbg/demo4/demo4.go:12 (hits goroutine(6):59 total:112)

id: 2 step: 99 sleeping 7
```

aarzilli:
What OS are you using?
Suppose the two goroutines are running on two separate threads, since the loop in `demo` is so tight and the sleep so relatively it would sometimes happen that both goroutines get to the breakpoint simultaneously. It may seem strange but consider that a lot of code needs to execute inside the kernel between a thread hitting a breakpoint and our `waitpid` call returning and part of that will count as 'simultaneous'.
This is one thing that could be happening.

dlsniper:
I'm on Ubuntu 15.04 x64 `Linux sniper 4.2.0-040200-generic #201508301530 SMP Sun Aug 30 19:31:40 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux`

aarzilli:
In that case could you try to disable hardware breakpoints by putting an `if false` around the for loop at `proc/breakpoints.go:101`? For some reason they are less reliable for me.

dlsniper:
Thanks, I've tried that as well but with the same result.

After thinking a bit on how goroutines work, this might be a problem caused by the fact that the breakpoint is set in the same thread even if multiple goroutines are used. Further investigation is required. What I've said might also be completely useless or dumb.

service/api/types.go
```diff
@@ -42,6 +42,10 @@ type Breakpoint struct {
 	Goroutine bool `json:"goroutine"`
 	// variables to evaluate
 	Variables []string `json:"variables,omitempty"`
+	// number of times a breakpoint has been reached in a certain goroutine
+	HitCount map[string]uint64 `json:"hitCount"`
```
derekparker: Why string key instead of uint, like the map in proc?

dlsniper: Unfortunately when converting to json, if `HitCount` would be `map[int]uint64` then marshaling will not work properly.
See the exemple here: https://play.golang.org/p/-1EMYNSCoa vs the working version: https://play.golang.org/p/MscIV7i306

derekparker: Ah, yeah that makes sense, thanks.

derekparker:
Merged.

PR #256, opened 2015-09-26
Three bug fixed concerning target termination on linux

aarzilli:
See commit messages for a description. I'm not sure about the last one, it's a workaround, I don't know why that happens.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -101,11 +101,13 @@ func (dbp *Process) Detach(kill bool) (err error) {
 		}
 	}
 	dbp.execPtraceFunc(func() {
-		var sig int
-		if kill {
-			sig = int(sys.SIGINT)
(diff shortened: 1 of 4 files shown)


derekparker:
Can you rebase this on master? The changes look good, however.

aarzilli:
done

proc/proc.go
```diff
@@ -101,11 +101,13 @@ func (dbp *Process) Detach(kill bool) (err error) {
 		}
 	}
 	dbp.execPtraceFunc(func() {
-		var sig int
+		err = PtraceDetach(dbp.Pid, 0)
```
derekparker: We should return here upon error, so it doesn't get masked below.

aarzilli: It doesn't, if kill isn't set the func ends, if it is and err is not nil it returns. But it's convoluted, I've changed it.

PR #259, opened 2015-09-27
proc.(*Process).Continue skips breakpoints [WIP]

aarzilli:
This is a fix for the problem unearthed by PR #255, it is submitted for discussion, not definitive.

There are two problems with `(*Process).Continue`

1. The first is that between the time one breakpoint is triggered and the point where we loop over all threads and `Halt` eacho one the second goroutine may hit the breakpoint too, this second hit gets lost
2. The second problem is that the way `Continue` resumes the process has a race condition with the tracee

Let's that thread A hit the breakpoint and thread B was sleeping when `Continue` called `Halt` on it. After we finish processing the breakpoint on A we call `Continue` again, `Continue` then loops through all threads, it calls `resume` on B, then moves onto thread A, sees that it's on a breakpoint so it calls `(*Thread).Step()`.
If thread B happens to get to the breakpoint while `(*Thread).Step` on thread A is happening the breakpoint gets skipped, because we temporarily cleared it.

You can test this by reverting the changes and running `TestBreakpointCounts`. The fact that multiple breakpoints can be hit simultaneously needs to be propagated all the way to the client, either by changing CurrentBreakpoint in State to a slice or by returning one breakpoint at a time in Debugger.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -389,20 +389,8 @@ func (dbp *Process) runBreakpointConditions() error {
 
 // Resume process, does not evaluate breakpoint conditionals
 func (dbp *Process) continueOnce() error {
-	// all threads stopped over a breakpoint are made to step over it
-	for _, thread := range dbp.Threads {
-		if thread.CurrentBreakpoint != nil {
-			if err := thread.Step(); err != nil {
-				return err
-			}
-			thread.CurrentBreakpoint = nil
-		}
-	}
-	// everything is resumed
-	for _, thread := range dbp.Threads {
-		if err := thread.resume(); err != nil {
-			return dbp.exitGuard(err)
-		}
+	if err := dbp.resume(); err != nil {
+		return err
 	}
 	return dbp.run(func() error {
 		thread, err := dbp.trapWait(-1)
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -746,6 +746,10 @@ func (dbp *Process) getGoInformation() (ver GoVersion, isextld bool, err error) 
 		err = fmt.Errorf("Could not determine version number: %v\n", err)
 		return
 	}
+	if vv.Unreadable != nil {
+		err = fmt.Errorf("Unreadable version number: %v\n", vv.Unreadable)
+		return
+	}
 
 	ver, ok := parseVersionString(constant.StringVal(vv.Value))
 	if !ok {
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -388,13 +388,8 @@ func (dbp *Process) Continue() error {
 			return dbp.exitGuard(err)
 		}
 		dbp.SwitchThread(thread.Id)
-		for _, th := range dbp.Threads {
-			if th.CurrentBreakpoint == nil {
-				err := th.SetCurrentBreakpoint()
-				if err != nil {
-					return err
-				}
-			}
+		if err := dbp.setExtraBreakpoints(); err != nil {
+			return err
 		}
 		loc, err := thread.Location()
 		if err != nil {
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -364,10 +364,19 @@ func (dbp *Process) setChanRecvBreakpoints() (int, error) {
 
 // Resume process.
 func (dbp *Process) Continue() error {
+	// all threads stopped over a breakpoint are made to step over it
 	for _, thread := range dbp.Threads {
-		err := thread.Continue()
-		if err != nil {
-			return fmt.Errorf("could not continue thread %d %s", thread.Id, err)
+		if thread.CurrentBreakpoint != nil {
+			if err := thread.Step(); err != nil {
+				return err
+			}
+			thread.CurrentBreakpoint = nil
+		}
+	}
+	// everything is resumed
+	for _, thread := range dbp.Threads {
+		if err := thread.resume(); err != nil {
+			return dbp.exitGuard(err)
 		}
 	}
 	return dbp.run(func() error {
@@ -376,9 +385,17 @@ func (dbp *Process) Continue() error {
 			return err
 		}
 		if err := dbp.Halt(); err != nil {
-			return err
+			return dbp.exitGuard(err)
 		}
 		dbp.SwitchThread(thread.Id)
+		for _, th := range dbp.Threads {
+			if th.CurrentBreakpoint == nil {
+				err := th.SetCurrentBreakpoint()
+				if err != nil {
+					return err
+				}
+			}
+		}
 		loc, err := thread.Location()
 		if err != nil {
 			return err
@@ -642,24 +659,18 @@ func (dbp *Process) handleBreakpointOnThread(id int) (*Thread, error) {
 	if !ok {
 		return nil, fmt.Errorf("could not find thread for %d", id)
 	}
-	pc, err := thread.PC()
+	// Check to see if we have hit a breakpoint.
+	err := thread.SetCurrentBreakpoint()
 	if err != nil {
 		return nil, err
 	}
-	// Check to see if we have hit a breakpoint.
-	if bp, ok := dbp.FindBreakpoint(pc); ok {
-		thread.CurrentBreakpoint = bp
-		if err = thread.SetPC(bp.Addr); err != nil {
-			return nil, err
-		}
-		if g, err := thread.GetG(); err == nil {
-			thread.CurrentBreakpoint.HitCount[g.Id]++
-		}
-		thread.CurrentBreakpoint.TotalHitCount++
+	if (thread.CurrentBreakpoint != nil) || (dbp.halt) {
 		return thread, nil
 	}
-	if dbp.halt {
-		return thread, nil
+
+	pc, err := thread.PC()
+	if err != nil {
+		return nil, err
 	}
 	fn := dbp.goSymTable.PCToFunc(pc)
 	if fn != nil && fn.Name == "runtime.breakpoint" {
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -244,35 +244,11 @@ func (dbp *Process) Status() *sys.WaitStatus {
 }
 
 // Step over function calls.
-func (dbp *Process) Next() error {
-	return dbp.run(dbp.next)
-}
-
-func (dbp *Process) next() (err error) {
-	defer func() {
-		// Always halt process at end of this function.
-		herr := dbp.Halt()
-		// Make sure we clean up the temp breakpoints.
-		cerr := dbp.clearTempBreakpoints()
-		// If we already had an error, return it.
-		if err != nil {
-			return
+func (dbp *Process) Next() (err error) {
+	for i := range dbp.Breakpoints {
+		if dbp.Breakpoints[i].Temp {
+			return fmt.Errorf("next while nexting")
 		}
-		if herr != nil {
-			err = herr
-			return
-		}
-		if cerr != nil {
-			err = cerr
-		}
-	}()
-
-	// Set breakpoints for any goroutine that is currently
-	// blocked trying to read from a channel. This is so that
-	// if control flow switches to that goroutine, we end up
-	// somewhere useful instead of in runtime code.
-	if _, err = dbp.setChanRecvBreakpoints(); err != nil {
-		return
 	}
 
 	// Get the goroutine for the current thread. We will
@@ -283,6 +259,14 @@ func (dbp *Process) next() (err error) {
 		return err
 	}
 
+	// Set breakpoints for any goroutine that is currently
+	// blocked trying to read from a channel. This is so that
+	// if control flow switches to that goroutine, we end up
+	// somewhere useful instead of in runtime code.
+	if _, err = dbp.setChanRecvBreakpoints(); err != nil {
+		return
+	}
+
 	var goroutineExiting bool
 	if err = dbp.CurrentThread.setNextBreakpoints(); err != nil {
 		switch t := err.(type) {
@@ -290,46 +274,20 @@ func (dbp *Process) next() (err error) {
 		case GoroutineExitingError:
 			goroutineExiting = t.goid == g.Id
 		default:
+			dbp.clearTempBreakpoints()
 			return
 		}
 	}
 
-	for _, th := range dbp.Threads {
-		if err = th.Continue(); err != nil {
-			return
+	if !goroutineExiting {
+		for i := range dbp.Breakpoints {
+			if dbp.Breakpoints[i].Temp {
+				dbp.Breakpoints[i].Cond = g.Id
+			}
 		}
 	}
 
-	for {
-		_, err := dbp.trapWait(-1)
-		if err != nil {
-			return err
-		}
-		for _, th := range dbp.Threads {
-			if !th.Stopped() {
-				continue
-			}
-			tg, err := th.GetG()
-			if err != nil {
-				return err
-			}
-			// Make sure we're on the same goroutine, unless it has exited.
-			if tg.Id == g.Id || goroutineExiting {
-				// Check to see if the goroutine has switched to another
-				// thread, if so make it the current thread.
-				if err := dbp.SwitchThread(th.Id); err != nil {
-					return err
-				}
-				return nil
-			}
-			// This thread was not running our goroutine.
-			// We continue it since our goroutine could
-			// potentially be on this threads queue.
-			if err = th.Continue(); err != nil {
-				return err
-			}
-		}
-	}
+	return dbp.Continue()
 }
 
 func (dbp *Process) setChanRecvBreakpoints() (int, error) {
@@ -362,8 +320,75 @@ func (dbp *Process) setChanRecvBreakpoints() (int, error) {
 	return count, nil
 }
 
-// Resume process.
+// Resume process
 func (dbp *Process) Continue() error {
+	for {
+		if err := dbp.continueOnce(); err != nil {
+			return err
+		}
+		// if dbp.CurrentThread.CurrentBreakpoint is nil a manual stop was requested
+		exitAnyway := (dbp.CurrentThread.CurrentBreakpoint == nil)
+		if err := dbp.runBreakpointConditions(); err != nil {
+			return err
+		}
+		if dbp.CurrentThread.onTriggeredBreakpoint() {
+			if dbp.CurrentThread.onTriggeredTempBreakpoint() {
+				if err := dbp.clearTempBreakpoints(); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+		if exitAnyway {
+			return nil
+		}
+	}
+}
+
+func (dbp *Process) runBreakpointConditions() error {
+	// first thread stopped on a breakpoint with true condition
+	var trigth *Thread
+	// first thread stopped on a temp breakpoint with true condition
+	var tempth *Thread
+
+	for _, th := range dbp.Threads {
+		if th.CurrentBreakpoint == nil {
+			continue
+		}
+
+		th.BreakpointConditionMet = th.CurrentBreakpoint.checkCondition(th)
+
+		if th.onTriggeredBreakpoint() {
+			if th.onTriggeredTempBreakpoint() {
+				if tempth == nil {
+					tempth = th
+				}
+			} else {
+				if trigth == nil {
+					trigth = th
+				}
+			}
+		}
+	}
+
+	// If a temp breakpoint was encountered make its thread the CurrenThread
+	// otherwise ensure that CurrentThread is on a triggered breakpoint if there is one
+	cth := dbp.CurrentThread
+	var err error
+	if tempth != nil {
+		if !cth.onTriggeredTempBreakpoint() {
+			err = dbp.SwitchThread(tempth.Id)
+		}
+	} else if trigth != nil {
+		if !cth.onTriggeredBreakpoint() {
+			err = dbp.SwitchThread(trigth.Id)
+		}
+	}
+	return err
+}
+
+// Resume process, does not evaluate breakpoint conditionals
+func (dbp *Process) continueOnce() error {
 	// all threads stopped over a breakpoint are made to step over it
 	for _, thread := range dbp.Threads {
 		if thread.CurrentBreakpoint != nil {
@@ -644,6 +669,11 @@ func (dbp *Process) clearTempBreakpoints() error {
 		}
 		if _, err := dbp.ClearBreakpoint(bp.Addr); err != nil {
 			return err
+		}
+	}
+	for i := range dbp.Threads {
+		if dbp.Threads[i].CurrentBreakpoint != nil && dbp.Threads[i].CurrentBreakpoint.Temp {
+			dbp.Threads[i].CurrentBreakpoint = nil
 		}
 	}
 	return nil
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -320,105 +320,71 @@ func (dbp *Process) setChanRecvBreakpoints() (int, error) {
 	return count, nil
 }
 
-// Resume process
 func (dbp *Process) Continue() error {
 	for {
-		if err := dbp.continueOnce(); err != nil {
+		if err := dbp.resume(); err != nil {
 			return err
 		}
-		// if dbp.CurrentThread.CurrentBreakpoint is nil a manual stop was requested
-		exitAnyway := (dbp.CurrentThread.CurrentBreakpoint == nil)
-		if err := dbp.runBreakpointConditions(); err != nil {
-			return err
-		}
-		if dbp.CurrentThread.onTriggeredBreakpoint() {
-			if dbp.CurrentThread.onTriggeredTempBreakpoint() {
-				if err := dbp.clearTempBreakpoints(); err != nil {
-					return err
-				}
-			}
+
+		var trapthread *Thread
+		var err error
+
+		dbp.run(func() error {
+			trapthread, err = dbp.trapWait(-1)
 			return nil
-		}
-		if exitAnyway {
-			return nil
-		}
-	}
-}
-
-func (dbp *Process) runBreakpointConditions() error {
-	// first thread stopped on a breakpoint with true condition
-	var trigth *Thread
-	// first thread stopped on a temp breakpoint with true condition
-	var tempth *Thread
-
-	for _, th := range dbp.Threads {
-		if th.CurrentBreakpoint == nil {
-			continue
-		}
-
-		th.BreakpointConditionMet = th.CurrentBreakpoint.checkCondition(th)
-
-		if th.onTriggeredBreakpoint() {
-			if th.onTriggeredTempBreakpoint() {
-				if tempth == nil {
-					tempth = th
-				}
-			} else {
-				if trigth == nil {
-					trigth = th
-				}
-			}
-		}
-	}
-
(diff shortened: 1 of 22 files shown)


derekparker:
The code changes look good, but I agree there is some more discussion that needs to happen here.

> The fact that multiple breakpoints can be hit simultaneously needs to be propagated all the way to the client, either by changing CurrentBreakpoint in State to a slice or by returning one breakpoint at a time in Debugger.

There are some UI considerations with something like this. Let me organize some thoughts on this, because it warrants some discussion.

aarzilli:
I've also noticed that the test fails on OSX (ironically, by triggering the breakpoint too many times). Either there is a similar bug or my change broke it.

aarzilli:
Thanks to @ignatov I found a bug with my changes to `Continue` which would cause it to never advance when a breakpoint was set on a 1 byte instruction, it was caused by a misunderstanding on my part of the behaviour of PC over soft-breakpoint instructions (on hindsight it's pretty obvious what happens).

I also found a second race condition in `Continue` when the traced program terminates. I've fixed this for linux on a second commit, I don't know if this also happens on OSX. I have seen this behaviour also in master, I don't think my patch introduced it.

derekparker:
@aarzilli please rebase on master. I'm going to review and also work on top of this branch, and continue the discussion above.

aarzilli:
Done, note that this is still on top of #255.

derekparker:
#255 has been merged. Can you please rebase again, I'd like to work on merging your patches in.

aarzilli:
Rebased.

proc/threads.go
```diff
@@ -91,6 +91,7 @@ func (thread *Thread) Step() (err error) {
 	}
 
 	err = thread.singleStep()
+	pc, _ = thread.PC()
```
derekparker: Seems leftover?

proc/threads.go
```diff
@@ -323,3 +324,26 @@ func (thread *Thread) Scope() (*EvalScope, error) {
 	}
 	return locations[0].Scope(thread), nil
 }
+
+func (thread *Thread) SetCurrentBreakpoint(excluded *Breakpoint) error {
```
derekparker: I don't see the value of passing anything in here, this method should always set the current breakpoint based on the PC. In the only case above where something is passed in, it's after a step which should mean we're past the breakpoint anyway. If `FindBreakpoint` is returning a breakpoint when it should not, then that is a bug with that function and should be dealt with there. I think the issue you're trying to work around is the fact that `FindBreakpoint` will try and resolve a breakpoint at `pc - <bp instruction len>`. That is to compensate for hitting a software breakpoint in which the current PC will be `<bp addr> + <bp instruction len>`. That is likely causing the issue in this case, since we could be single stepping over single byte instruction, after which we would incorrectly resolve the old breakpoint.

proc/proc.go
```diff
@@ -359,10 +359,27 @@ func (dbp *Process) setChanRecvBreakpoints() (int, error) {
...
+	// all threads stopped over a breakpoint are made to step over it
 	for _, thread := range dbp.Threads {
-		err := thread.Continue()
-		if err != nil {
-			return fmt.Errorf("could not continue thread %d %s", thread.Id, err)
+		if thread.CurrentBreakpoint != nil {
+			if err := thread.Step(); err != nil {
+				return err
+			}
+			if err := thread.SetCurrentBreakpoint(thread.CurrentBreakpoint); err != nil {
```
derekparker: Ideally this should be done by any method that continues execution. I.e. let's move this directly into `thread.Step`.

derekparker: On second thought, maybe move this to `dbp.run`, that's the central function that wraps any other function that continues execution.

proc/proc.go
```diff
@@ -359,10 +359,27 @@ func (dbp *Process) setChanRecvBreakpoints() (int, error) {
...
+			}
+		}
+	}
+	// if some threads have found a new breakpoint then we are done continuing
+	for _, thread := range dbp.Threads {
+		if thread.CurrentBreakpoint != nil {
+			return dbp.SwitchThread(thread.Id)
+		}
+	}
+	// othrewise resume everything
```
derekparker: s/othrewise/otherwise/

derekparker:
The test `TestBreakpointCounts` consistently fails for me on OSX, I'll try and dig in a bit further to see why. Could be related to the OSX test flickers. I would love to get rid of those...

proc/proc.go
aarzilli: I don't think moving this code in `dbp.run` is appropriate for `next` and `step`: `next` needs to set its temp breakpoints before continuing and `step` shouldn't call `resume` at all. Also moving this into `thread.Step` would break `step` (because it steps one instruction anyway, not just if the thread is on a breakpoint).

proc/threads.go
aarzilli: That's why that parameter existed. However I realized that I can just singlestep all the threads stopped on a breakpoint and then call resume on all threads indiscriminately, no need check for hitting a breakpoint after the singlestepping phase, it will never happen anyway. I have changed the patch to do this.

aarzilli:
We also need to decide what to do with API and UI, right now we are counting breakpoint hits correctly but only reporting the first one of each group to the user.

ignatov:
Please ping me when it will be merged.

derekparker:
> We also need to decide what to do with API and UI, right now we are counting breakpoint hits correctly but only reporting the first one of each group to the user.

Thinking about this a bit more, I have some thoughts which I would love input on from @ignatov, @ironcladlou and @dlsniper since they are working on actual client integrations.

First let me ask, how you you all like this information presented to you? Specifically, since we have the concept of "current thread" and "current goroutine", is it enough for Delve to decide which "current breakpoint" to present to clients? If not, would it make sense to return a slice of current breakpoints? To me that would be confusing. Would it instead make more sense to continue to return information on the "current thread / goroutine" and additionally provide a slice containing all thread state, which includes current breakpoints. That way, you could easily use the "current breakpoint" of the "current thread", or check out all the other threads and see if they have a breakpoint set.

Would really like feedback on this, and I'd really like to get this merged in asap.

ironcladlou:
Having access to "all" plus an indicator of "current" sounds workable to me.

I have no background in debugger internals or debugger frontend development prior to messing with Delve/Sublime, so I'm winging it and happily defer to those who have already thought through these issues. I've been learning a lot from go-lang-idea-plugin, and since IntelliJ's debugger rocks, I normally assume whatever is needed to make the IntelliJ plugin work is probably more than enough for my little Sublime plugin. :grin:

aarzilli:
I have moved `Breakpoint` and `BreakpointInfo` inside `Thread` and added a `Threads` slice to `DebuggerState`. Current breakpoints can be retrieved by iterating through `Threads` and looking at non nil `Thread.Breakpoint`s, `DebuggerState.CurrentThread.Breakpoint` is the current thread's breakpoint.

Is this acceptable?

derekparker:
@aarzilli that seems fine to me.

derekparker:
The tests here seem to consistently fail on OSX. I'll try and dig into it so we can get this merged.

aarzilli:
Added a fix for #262

aarzilli:
I have fixed `TestBreakpointCounts` on OS X. The problems that I fixed on linux also affected os x but my fix caused delve to become desynchronized with mach's message queue.

Now `TestNextConcurrent` seems to fail more frequently but imho it would be better to reimplement `Next` in terms of conditional breakpoints and `Continue`.

derekparker:
@aarzilli want to expand on your second point a bit?

derekparker:
@aarzilli I agree with both points. Let's never silently ignore a valid breakpoint, even when nexting. Also, I think implementing next this way with conditional breakpoints would be a very elegant approach, I'm all for this.

What do you see as the way forward for this PR? Do you think it's worth exploring a refactor that could help fix `TestNextConcurrent` within this PR, or wait for these other abstractions to be built in, separately.

aarzilli:
I would prefer if work on next was done on a separate PR. For conditional breakpoints we need to discuss their syntax and it doesn't belong here (but doing it before merging this is a problem because changes to breakpoints will conflict).

derekparker:
@aarzilli sorry, didn't mean to imply that the conditional breakpoints would be a part of this PR, but that they could land before this one. Aside from the conditional breakpoints, what are your thoughts on moving forward with the rest of your proposal in this PR? Ideally, once this is merged the TestNextConcurrent flicker is solved, not made worse.

aarzilli:
> Aside from the conditional breakpoints, what are your thoughts on moving forward with the rest of your proposal in this PR?

The rest of my porposal depends on having conditional breakpoints. I could just add the internals of conditional breakpoints to this PR (no API, no UI) and then reimplement `Next` on top of that. Unless I'm overlooking something it shouldn't be too hard.

> Ideally, once this is merged the TestNextConcurrent flicker is solved, not made worse

Ideally, yes. Of course there could be other bugs I didn't think of.

aarzilli:
See the commit message for 77f3985 for a description of the change.

I have run proc's tests continuously for about 1.5hrs on both linux and os x. The issues that remain are:

On linux:

* TestIssue262 sometimes fails with the error "process did not exit after second continue: no such process" this is probably an exitGuard that's missing somewhere or the sys.ESRCH getting wrapped so that exitGuard is unable to convert it
* TestBreakpointCounts sometimes fails with an off by 1 count (201 instead of 200). This is a more serious problem but unfortunately it's also very infrequent (I've seen it twice in 1.5 hr of continuously running tests)

On OS X the only problem I can find is some kind of problem with Launch that manifests in one of two ways:

* getGoInformation is unable to read runtime.buildVersion (the data at that address is nonsensical)
* if I comment out getGoInformation (it's useless on OS X) the target program seems to sometimes escape control of delve. This manifests with a number of different tests failing because of a unexpected ProcessExitedError.

I "solved" this problem by adding a `time.Sleep` inside proc.Launch, this is why I know it's a race condition even though I can't fix it. My theory is that `Launch` sometimes catches the program between `read` and `execve` in `exec_darwin.c` and that's what's causing this problem.

Overall the situation seems to be improved.

aarzilli:
This should fix the two remaining problems on linux.

derekparker:
@aarzilli nice, all the tests are consistently passing on my OSX system. I put some comments on https://github.com/aarzilli/delve/commit/77f3985c1dd00422ac55488b6b832d48f583229a - other than that the only issue was the Sleep work around.

Check out this diff, this should work for preventing the racy-ness:

```diff
diff --git a/proc/exec_darwin.c b/proc/exec_darwin.c
index ac2d9b3..5d339c2 100644
--- a/proc/exec_darwin.c
+++ b/proc/exec_darwin.c
@@ -15,6 +15,12 @@ fork_exec(char *argv0, char **argv, int size,
        int fd[2];
        if (pipe(fd) < 0) return -1;
 
+       // Create another pipe so that we know when we're about to exec. This ensures that control only returns
+       // back to Go-land when we call exec, effectively eliminating a race condition between launching the new
+       // Create another pipe so that we know when we're about to exec. This ensures that control only returns
+       // back to Go-land when we call exec, effectively eliminating a race condition between launching the new
+       // process and trying to read its memory.
+       int wfd[2];
+       if (pipe(wfd) < 0) return -1;
+
        kern_return_t kret;
        pid_t pid = fork();
        if (pid > 0) {
@@ -26,6 +32,11 @@ fork_exec(char *argv0, char **argv, int size,
                char msg = 'c';
                write(fd[1], &msg, 1);
                close(fd[1]);
+
+               char w;
+               read(wfd[0], &w, 1);
+               close(wfd[0]);
+
                return pid;
        }
 
@@ -53,6 +64,10 @@ fork_exec(char *argv0, char **argv, int size,
        pret = ptrace(PT_SIGEXC, 0, 0, 0);
        if (pret != 0 && errno != 0) return -errno;
 
+       char msg = 'd';
+       write(wfd[1], &msg, 1);
+       close(wfd[1]);
+
        // Create the child process.
        execve(argv0, argv, environ);
 
diff --git a/proc/proc_darwin.go b/proc/proc_darwin.go
index 574df21..478a956 100644
--- a/proc/proc_darwin.go
+++ b/proc/proc_darwin.go
@@ -13,7 +13,6 @@ import (
        "os/exec"
        "path/filepath"
        "sync"
-       "time"
        "unsafe"
 
        "github.com/derekparker/delve/dwarf/frame"
@@ -74,7 +73,6 @@ func Launch(cmd []string) (*Process, error) {
        for i := range argvSlice {
                C.free(unsafe.Pointer(argvSlice[i]))
        }
-       time.Sleep(250 * time.Millisecond) // ( ͡° ͜ʖ ͡°)
 
        dbp, err = initializeDebugProcess(dbp, argv0Go, false)
        if err != nil {
```

aarzilli:
I ran the tests on OS X for a while, your fix seems to work: I never saw the unreadable `runtime.buildVersion` problem, I saw the runaway process problem a few times but it's very very rare (1 in 50?), I didn't get it with the `time.Sleep` but maybe I was just lucky.

aarzilli:
Removed unused `gid` argument from `setNextBreakpoints`.

aarzilli:
done.

derekparker:
@aarzilli pulled branch and ran tests several times, getting failures probably 80% of the time on OSX. Are you getting any failures in your OSX VM?

aarzilli:
I see what I was seeing [here](https://github.com/derekparker/delve/pull/259#issuecomment-158369586). A rare failure where the process runs to completion immediately after we start it. 

It did not happen with the `time.Sleep`, if it's still the race condition it would make sense that it's more frequent on real hardware.

derekparker:
@aarzilli update: didn't have time to look into this today. I'll be spending some time tomorrow trying to figure out the issue on OSX, with the goal being to get this merged in right after that's sorted out.

aarzilli:
Rebased on master at ad5097a8dc716e61992d5c7dd6213e58cb0a88b5 (no conflicts)

aarzilli:
It seems that the behaviour of `go/constant` has changed on go-tip, that's what's breaking the tests.

derekparker:
@aarzilli interesting. I'm still trying to track down OSX issues, however I haven't had much time recently to dig into it. I may actually have time today, so let's hope I can figure something out and we can get this merged in. Unrelated, but I've also been working on implementing function calls. That work is sitting on a branch (maybe I'll send a WIP PR just to publicly track the work). I haven't touched it in a week or so (no time) but it was working in a very limited case, but there were things to consider that I wasn't sure about, which maybe would be another good reason to submit a WIP PR and begin discussion around a few things. Anyways, hopefully we can address the changes on tip and the OSX issues with this PR and keep pushing forward.

derekparker:
@aarzilli leaving these traces here in case they jump out to you for any reason:

```
make test-proc-run RUN=Next
go test -exec=/Users/derekparker/code/go/src/github.com/derekparker/delve/scripts/testsign -ldflags="-s -X main.Build=fd1f97d449541403a2fac75019a4df75d5d852be" -test.run="Next" github.com/derekparker/delve/proc
foo
fatal error: unexpected signal during runtime execution
hi 9
[signal 0xa code=0x2 addr=0xc80000331b pc=0x4012e14]

goroutine 50 [running]:
runtime.throw(0x453e1a0, 0x2a)
	/usr/local/go/src/runtime/panic.go:527 +0x90 fp=0xc82034b4f8 sp=0xc82034b4e0
runtime.sigpanic()
	/usr/local/go/src/runtime/sigpanic_unix.go:12 +0x5a fp=0xc82034b548 sp=0xc82034b4f8
runtime.mallocgc(0x9, 0x4379040, 0x1, 0x49672)
	/usr/local/go/src/runtime/malloc.go:585 +0x314 fp=0xc82034b618 sp=0xc82034b548
runtime.newarray(0x4379040, 0x9, 0x9f)
	/usr/local/go/src/runtime/malloc.go:777 +0xc9 fp=0xc82034b658 sp=0xc82034b618
runtime.makeslice(0x4369920, 0x9, 0x9, 0x0, 0x0, 0x0)
	/usr/local/go/src/runtime/slice.go:32 +0x165 fp=0xc82034b6a8 sp=0xc82034b658
github.com/derekparker/delve/proc.parseG(0xc8202c8f40, 0x8201d7200, 0x0, 0xc8203212e0, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/variables.go:320 +0x94 fp=0xc82034b8d8 sp=0xc82034b6a8
github.com/derekparker/delve/proc.(*Thread).GetG(0xc8202c8f40, 0xc8200804d0, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/threads.go:281 +0x347 fp=0xc82034b980 sp=0xc82034b8d8
github.com/derekparker/delve/proc.(*Breakpoint).checkCondition(0xc8200925a0, 0xc8202c8f40, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/breakpoints.go:116 +0x3d fp=0xc82034b9a8 sp=0xc82034b980
github.com/derekparker/delve/proc.(*Thread).SetCurrentBreakpoint(0xc8202c8f40, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/threads.go:330 +0x10c fp=0xc82034ba10 sp=0xc82034b9a8
github.com/derekparker/delve/proc.(*Process).setCurrentBreakpoints(0xc82007a2a0, 0xc8202c8f40, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc_darwin.go:318 +0x31 fp=0xc82034ba50 sp=0xc82034ba10
github.com/derekparker/delve/proc.(*Process).Continue(0xc82007a2a0, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc.go:342 +0x183 fp=0xc82034bb00 sp=0xc82034ba50
github.com/derekparker/delve/proc.(*Process).Next(0xc82007a2a0, 0x0, 0x0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc.go:290 +0x3b1 fp=0xc82034bbf0 sp=0xc82034bb00
github.com/derekparker/delve/proc.TestNextConcurrent.func1(0xc82007a2a0, 0x44c59f0, 0xd, 0xc8200ac820, 0x47, 0xc820527ec0, 0x56)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc_test.go:352 +0x74a fp=0xc82034bdc0 sp=0xc82034bbf0
github.com/derekparker/delve/proc.withTestProcess(0x44c59f0, 0xd, 0xc820340120, 0xc82034bf28)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc_test.go:41 +0x339 fp=0xc82034bed0 sp=0xc82034bdc0
github.com/derekparker/delve/proc.TestNextConcurrent(0xc820340120)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc_test.go:364 +0xb7 fp=0xc82034bf58 sp=0xc82034bed0
testing.tRunner(0xc820340120, 0x46bd4f8)
	/usr/local/go/src/testing/testing.go:456 +0x98 fp=0xc82034bf90 sp=0xc82034bf58
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1696 +0x1 fp=0xc82034bf98 sp=0xc82034bf90
created by testing.RunTests
	/usr/local/go/src/testing/testing.go:561 +0x86d

goroutine 1 [chan receive]:
testing.RunTests(0x4574320, 0x46bd420, 0x25, 0x25, 0x1)
	/usr/local/go/src/testing/testing.go:562 +0x8ad
testing.(*M).Run(0xc820037f08, 0x0)
	/usr/local/go/src/testing/testing.go:494 +0x70
github.com/derekparker/delve/proc/test.RunTestsWithFixtures(0xc820037f08, 0x44bbcd0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/test/support.go:65 +0x30
github.com/derekparker/delve/proc.TestMain(0xc820037f08)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc_test.go:26 +0x21
main.main()
	github.com/derekparker/delve/proc/_test/_testmain.go:124 +0x113

runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1696 +0x1

goroutine 51 [chan receive, locked to thread]:
github.com/derekparker/delve/proc.(*Process).handlePtraceFuncs(0xc82007a2a0)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc.go:656 +0xc3
	/Users/derekparker/code/go/src/github.com/derekparker/delve/proc/proc.go:71 +0x1d4
FAIL	github.com/derekparker/delve/proc	2.064s
make: *** [test-proc-run] Error 1
```

AND

```
make test-proc-run RUN=Next                                                                                                            {2}
go test -exec=/Users/derekparker/code/go/src/github.com/derekparker/delve/scripts/testsign -ldflags="-s -X main.Build=fd1f97d449541403a2fac75019a4df75d5d852be" -test.run="Next" github.com/derekparker/delve/proc
foo
hi 0
hi 4
hi 6
hi 1
hi 8
hi 0
hi 8
hi 5
hi 9
hi 9
hi 7
hi 7
hi 4
[signal 0xb code=0x1 addr=0x8a0 pc=0x2041]

goroutine 7 [running]:
main.sayhi(0x2, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8 +0x1
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 1 [semacquire]:
sync.runtime_Semacquire(0x8201e047c)
	/usr/local/go/src/runtime/sema.go:43 +0x26
sync.(*WaitGroup).Wait(0x8201e0470)
	/usr/local/go/src/sync/waitgroup.go:126 +0xb4
main.main()
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:20 +0xb9

goroutine 5 [runnable]:
main.sayhi(0x0, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 6 [runnable]:
main.sayhi(0x1, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 8 [runnable]:
main.sayhi(0x3, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 9 [runnable]:
main.sayhi(0x4, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 10 [runnable]:
main.sayhi(0x5, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:9 +0xce
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 11 [runnable]:
main.sayhi(0x6, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 12 [runnable]:
main.sayhi(0x7, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 13 [runnable]:
main.sayhi(0x8, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:8
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88

goroutine 14 [runnable]:
fmt.glob.func1(0x0, 0x0)
	/usr/local/go/src/fmt/print.go:128 +0x2f
sync.(*Pool).getSlow(0x191f20, 0x0, 0x0)
	/usr/local/go/src/sync/pool.go:141 +0x188
sync.(*Pool).Get(0x191f20, 0x0, 0x0)
	/usr/local/go/src/sync/pool.go:117 +0x141
fmt.newPrinter(0x1b5830)
	/usr/local/go/src/fmt/print.go:133 +0x27
fmt.Fprintln(0x5f89000, 0x820200010, 0x8201fa788, 0x2, 0x2, 0xca00, 0x0, 0x0)
	/usr/local/go/src/fmt/print.go:253 +0x2d
fmt.Println(0x8201fa788, 0x2, 0x2, 0xc0100, 0x0, 0x0)
	/usr/local/go/src/fmt/print.go:264 +0x73
main.sayhi(0x9, 0x8201e0470)
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:9 +0x1a0
created by main.main
	/Users/derekparker/code/go/src/github.com/derekparker/delve/_fixtures/parallel_next.go:18 +0x88
--- FAIL: TestNextConcurrentVariant2 (0.30s)
	proc_test.go:61: failed assertion at proc_test.go:405: Continue 2 - Process 47797 has exited with status 2
foo
Hello, World!
bye
hi
FAIL
FAIL	github.com/derekparker/delve/proc	5.646s
make: *** [test-proc-run] Error 1
```

Those are the panics I get. I also get errors like "expected to be at line 11, was at line 10", e.g.:

```
--- FAIL: TestNextConcurrentVariant2 (0.27s)
	proc_test.go:410: Program did not continue to correct next location expected 10 was parallel_next.go:11
```

The other failure condition I see is that the test will just run forever. Also, occasionally it passes. Do you see any of these errors running bare metal on Linux? If not, I suspect it may be hidden in the C code (or more concretely the interfacing between Go and C) that is causing the issue, since that would be the major difference there since on Darwin we call into C a lot of syscalls and such.

I'm going to keep debugging this for a bit and see if I get anywhere but I figured I'd leave those traces around for you or anyone else who may want to take a stab at debugging this on OSX.

derekparker:
@aarzilli can you rebase this on master? I'm going to spend some time tomorrow in this branch, trying to fix the OSX issues.

aarzilli:
Done, did you have the time to check if it's the same weird launch race condition that's 'fixed' by the `time.Sleep`?

derekparker:
@aarzilli yeah I did look into that but it didn't seem to fix it. I think I'm running into another issue, but I think there may still be a race somewhere between Delve and the debugged process manifesting itself oddly. Looking at the stack traces above, the panic always happens in the same spot, when we're trying to parse the G struct, I just haven't had much time to actually dig into it, but I'm going to start now and see what I come up with either today or tomorrow.

aarzilli:
added the stuff, rebased on master.

Issue #220, opened 2015-09-04
Support printing nested struct member


ThomasMoinel:
Can't print nested struct member, see error below with example code.
```
% dlv debug 
Type 'help' for list of commands.
(dlv) b main.main:3
Breakpoint 1 set at 0x400c37 for main.main ./nestedstruct.go:15
(dlv) c
> main.main() ./nestedstruct.go:15
	
   10:	}
   11:	
   12:	func main() {
   13:		b := B{A{42}}
   14:	
=> 15:		fmt.Println(b)
   16:	}

(dlv) print b
main.B {a: main.A {val: 42}}
(dlv) print b.a
main.A {val: 42}
(dlv) print b.a.val
Command failed: b has no member a.val
(dlv) 
```

nestedstruct.go
``` go
package main

import "fmt"

type A struct {
	val int
}
type B struct {
	a A
}

func main() {
	b := B{A{42}}

	fmt.Println(b)
}
```

ThomasMoinel:
A more complete example where every print give the same _Command failed: b has no member *_
```go
package main

import "fmt"

type A struct {
	val int
}
type B struct {
	A
	a   A
	ptr *A
}

func main() {
	b := B{A: A{-314}, a: A{42}, ptr: &A{1337}}

	// print b.val
	// print b.A.val
	// print b.a.val
	// print b.ptr.val

	fmt.Println(b)
}
```

I think a correct way is to make a recursive call in proc/variable.go evaluateStructMember.
I'm trying to make it work but I struggle to learn dwarf and delve internals. If I find out a way, I'll make a pull request.

derekparker:
@aarzilli or @ebfe any interest in this?

aarzilli:
The current code uses a special case to extract the address of a member field, so supporting this is not easy. When I refactored variables.go to split address extraction and evaluation I also rewrote this part to be more general. Once I commit that it should be easy to add a for loop and go any number of levels deep, as a provisional measure, until we have a more general expression evaluator.

ThomasMoinel:
From the last example print embedded struct value doesn't work:
```
print b.val
print b.A.val
```

aarzilli:
This should be reopened, we have to figure out how embedded structs are stored in dwarf.

lukehoban:
I've opened a new PR (#261) adds support for embedded structs and a test case that includes the full example from @ThomasMoinel.

PR #260, opened 2015-09-29
Support for nested variables

lukehoban:
Adds support for nested variables in ExtractVariableInfo.  It is a targeted fix for #220.  Ideally this will be further expanded in the future to include an even broader expression subset (in particular array/slice indexing).


derekparker:
Merged, thanks for this patch! In the future, please format you commit messages according to: https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message.

PR #261, opened 2015-10-01
proc/variables: Support for embedded structs

### LINKED ISSUE #220: Support printing nested struct member

lukehoban:
Embedded structs are encoded in DWARF as fields with
package-qualified names.  They define an anonymous field
on the struct with the non-qualified name, as well as
promoted fields for each field of the embedded struct so
long as these are not shadowed by fields of the containing
struct.

Fixes #220.

### DIFF
--- a/proc/variables.go
+++ b/proc/variables.go
@@ -127,7 +127,12 @@ func newVariable(name string, addr uintptr, dwarfType dwarf.Type, thread *Thread
 func (v *Variable) toField(field *dwarf.StructField) (*Variable, error) {
 	name := ""
 	if v.Name != "" {
-		name = fmt.Sprintf("%s.%s", v.Name, field.Name)
+		parts := strings.Split(field.Name, ".")
+		if len(parts) > 1 {
+			name = fmt.Sprintf("%s.%s", v.Name, parts[1])
+		} else {
+			name = fmt.Sprintf("%s.%s", v.Name, field.Name)	
+		}
 	}
 	return newVariable(name, uintptr(int64(v.Addr)+field.ByteOffset), field.Type, v.thread)
 }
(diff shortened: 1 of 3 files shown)


proc/variables.go
```diff
@@ -442,6 +447,13 @@ func (scope *EvalScope) packageVarAddr(name string) (*Variable, error) {
 	return nil, fmt.Errorf("could not find symbol value for %s", name)
 }
 
+func (v *Variable) toFieldNilChecked(field *dwarf.StructField, name string) (*Variable, error) {
+	if v.Addr == 0 {
```
derekparker: Is there any reason not to simply check this by default and fold this code into `toField` itself?

derekparker: @lukehoban thoughts?

lukehoban: Thanks - I've moved the null check to `toField` now.  I don't see any callers who expected this to allow nil receivers, so I think this is okay.

derekparker:
Merged - thanks!

PR #264, opened 2015-10-02
terminal/command: add optional linespec argument to clearall

kostya-sh:
If an optional linespec argument is passed to clearall command then only
breakpoints matching this linespec will be cleared.

Behavior of clearall command without arguments hasn't changed.


derekparker:
Merged, thanks!

Issue #243, opened 2015-09-20
variables: Store struct member info


derekparker:
Structs, when evaluated, should include a slice of variables which contains the member fields.

Issue #186, opened 2015-07-29
`info locals` shows uninitialized value for as-yet undeclared variable


joeshaw:
Running Delve 0.6.0.beta (a506bb7d8e33e85878866152ba738206dbef9209) and go version go1.4.2 darwin/amd64.  With the following test program:

``` go
package main

import "fmt"

func main() {
    fmt.Println("hello world")
    x := 100
    fmt.Println(x)
}
```

```
$ dlv run
Type 'help' for list of commands.
(dlv) b main.main
Breakpoint 1 set at 0x2022 for main.main ./foo.go:6
(dlv) c
> main.main() ./foo.go:6

    1:  package main
    2:
    3:  import "fmt"
    4:
    5:  func main() {
=>  6:      fmt.Println("hello world")
    7:      x := 100
    8:      fmt.Println(x)
    9:  }

(dlv) info locals
x = 8589934592
```

`x` is undeclared at this point.  I'd expect either that it wouldn't display at `info locals` at all or that it'd say something like `x = (undeclared)`.  Once I step past the declaration/assignment, it has the correct value of `x = 100`.

derekparker:
Related to https://github.com/derekparker/delve/issues/106, with similar issues for fixing.

Issue #106, opened 2015-04-21
Variable scoping


derekparker:
Given the following source:

```go
func fn() {
        for i := 0; i < 10: i++ {
                fmt.Println(i)
        }

        for i := 0; i < 10; i++ {
            fmt.Println(i)
        }
}
```

If you break in the second loop, the result of `print i` will be 10 instead of the value of the next `i` that shadows it.

derekparker:
@epipho would you have any interest in taking a look at this?

epipho:
Yeah I will take a look at this.  I think I know what is going on.

derekparker:
@epipho awesome, thank you.

epipho:
So this is what I thought and is unfortunately a related problem to #83 

There are actually two variables with the same name and we print the first one we come across.  You can actually see this if you use info locals, it will show both.

```
   6:   for foobarbaz := 0; foobarbaz < 10; foobarbaz++ {
   7:           fmt.Println(foobarbaz)
   8:   }
   9:
   10:  for foobarbaz := 0; foobarbaz < 10; foobarbaz++ {
=> 11:          fmt.Println(foobarbaz)
   12:  }
   13: }

(dlv) info locals
foobarbaz = 0
foobarbaz = 10
(dlv) print foobarbaz
10
```

As in #83 there is no way to disambiguate them as there is no scoping or .debug_loc data

```
 <2><40>: Abbrev Number: 4 (DW_TAG_variable)
    <41>   DW_AT_name        : foobarbaz
    <4b>   DW_AT_location    : 5 byte block: 9c 11 98 7e 22     (DW_OP_call_frame_cfa; DW_OP_consts: -232; DW_OP_plus)
    <51>   DW_AT_type        : <0x29a43>
 <2><59>: Abbrev Number: 4 (DW_TAG_variable)
    <5a>   DW_AT_name        : foobarbaz
    <64>   DW_AT_location    : 5 byte block: 9c 11 a0 7e 22     (DW_OP_call_frame_cfa; DW_OP_consts: -224; DW_OP_plus)
    <6a>   DW_AT_type        : <0x29a43>
```

For print statements we could print out all the variables with the name, but the order is not going to be guaranteed.

Thoughts? I think we really need to solve this issue for a 1.0, but am not sure where to start other than putting together a proposal for go1.6 as 1.5 is about to lock.

derekparker:
@epipho thanks for looking into this. You're right, this seems to be a Go linker shortcoming. 

[...]

Issue #83, opened 2015-03-11
breakpoints inside multi-line var declaration are unsafe


ivarg:
It is possible to set a breakpoint in a multi-line variable declaration, at which point the variables have not yet been initialized. Calling `info locals` at that point will render undefined behavior and may result in a panic.

derekparker:
Sure enough, thanks for reporting this.

@epipho would you have any interest in looking into this? You've taken an interest in the variable evaluation side of things and had some great commits. If not, I'll start digging into it.

epipho:
I have run into this before.

The bad news is I am not sure what we can do about it as I don't think there is any information in the binary.

 Instead of a location expression for the variable location I would expect a location list (http://www.dwarfstd.org/doc/dwarf-2.0.0.pdf section 2.4.6) that would give a set of instruction ranges that the variable is valid for but the binary does not contain a .debug_loc section.

The same problem exists for code that shadows variables
```
func main() {
   f := "foo"
   {
     f := "bar"
     fmt.Println(f)
   }
   fmt.Println(f)
}
```
This actually generates two variables, and there does not appear to be any way to disambiguate them.
```
<2><40>: Abbrev Number: 4 (DW_TAG_variable)
    <41>   DW_AT_name        : abcde
    <47>   DW_AT_location    : 5 byte block: 9c 11 a8 7f 22     (DW_OP_call_frame_cfa; DW_OP_consts: -88; DW_OP_plus)
    <4d>   DW_AT_type        : <0x2939f>
 <2><55>: Abbrev Number: 4 (DW_TAG_variable)
    <56>   DW_AT_name        : abcde
    <5c>   DW_AT_location    : 5 byte block: 9c 11 b8 7f 22     (DW_OP_call_frame_cfa; DW_OP_consts: -72; DW_OP_plus)
    <62>   DW_AT_type        : <0x2939f>
```
Thoughts?

derekparker:
@epipho that's unfortunate. I don't see a clear fix other than patching the linker(s) to emit the `.debug_loc` section, or at the very least submitting an issue to the Go project to include that information.

epipho:
This weekend I will dig through the go issue tracker to see if this has been brought up. I wonder what the complexity of the change would be to add .debug_log to the linker.

derekparker:
@epipho not sure, but I know there's a lot going on with the compiler / linker rewrites. I'll likely post something on golang-dev about this and another Dwarf tag I'd like the linker to emit in the .debug_info section.

[...]

PR #268, opened 2015-10-05
Changed api.Variable to have a machine readable value

aarzilli:
Implements #243, Workaround for #186, #106, #83 

I have moved conversions.go from api to proc, this way I could import service/api into proc and use the type SimpleType defined there in proc, without having to duplicate it.

api.Variable now has two methods to convert it into a string: `SinglelineString` and `MultilineString`. 
`SinglelineString` is mostly identical to the previous serialization, the two differences are that I'm quoting string values and when variables are cut from the evaluation because we reached the recursion limit instead of printing `{...}` I print `(*<type>)(<address>)` which could be (in the future) be used by a user to further explore the data structure.
`MultilineString` is a simple attempt at pretty printing, I'm sure it can be improved significantly.

I have also included a workaround for Issue #186 (and related issues). When there is more than one definition of a variable in scope I'll return a Variable with SimpleType `multi` and the list of its possible values as Children. Those issues should still get their compiler side fix but in it's not going to happen soon and it's unlikely that it will ever happen for 1.4.x and 1.5.0/1.5.1.

There will probably be conflicts with #261, I will fix them when that's merged.


derekparker:
I'd rather keep `proc` in the dark about the API. What about defining `SimpleType` in `proc` instead, and importing it into `api`? The `api` conversions could, in the future, be tied to a certain version of the api we expose (which we really should consider that we're building `v1` of the api and namespace it as such now).

Fixes for #186, #106, #83 _should_ be landing in Go 1.6, at least that's what I've been told.

Can you separate the workaround code for #186, #106, #83 into a different PR (on top of this one), but keep the new serialization code in this PR (quoted strings, SingleLine, MultiLine).

aarzilli:
> I'd rather keep proc in the dark about the API. What about defining SimpleType in proc instead, and importing it into api?

Done, however I also had to bring prettyprint.go over to proc, the tests in variables_test.go need a way to do serialize proc.Variable.

> Can you separate the workaround code for #186, #106, #83 into a different PR

Done, I have restored the previous behaviour, with EvalVariable returning the first variable with a given name, however I would suggest returning an error instead if we aren't doing the workaround.

>Fixes for #186, #106, #83 should be landing in Go 1.6, at least that's what I've been told.

That would be great news, I checked the issue tracker but found nothing (except for old bugs)

kostya-sh:
This change looks great, I am looking forward to more readable print output for complex objects.

I have a few comment/suggestion about the change:

- Does MultilineString and pretty printing need to be in the server? If a client (e.g. terminal) has a JSON-based representation of a variable it can do pretty printing itself and even apply custom formatting rules without any remote calls
- It would be good to add maxVariableRecurse and maxArrayValues parameters to the EvalVariable API call instead of hard-coding them

aarzilli:
> Does MultilineString and pretty printing need to be in the server?

MultilineString runs in the client.

kostya-sh:
> MultilineString runs in the client.

I am not yet very familiar with the code and its usages but I would have expected MultilineString to be in terminal package then (not exported). Unless delve also aims to provide a library that can be used to write another client app in Go.

proc/prettyprint.go
```diff
@@ -0,0 +1,263 @@
...
+	maxShortStringLen = 7
+	// string used for one indentation level (when printing on multiple lines)
+	indentString = "\t"
+)
+
+// Returns a representation of v on a single line
+func (v *ApiVariable) SinglelineString() string {
+	var buf bytes.Buffer
+	v.writeTo(&buf, false, true, "")
+	return string(buf.Bytes())
```
derekparker: Use `buf.String()` instead.

proc/prettyprint.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// Returns a representation of v on multiple lines
+func (v *ApiVariable) MultilineString(indent string) string {
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "%s", indent)
+	v.writeTo(&buf, true, true, indent)
+	return string(buf.Bytes())
```
derekparker: Same as above.

proc/prettyprint.go
```diff
@@ -0,0 +1,263 @@
...
+func (v *ApiVariable) SinglelineString() string {
+	var buf bytes.Buffer
+	v.writeTo(&buf, false, true, "")
+	return string(buf.Bytes())
+}
+
+// Returns a representation of v on multiple lines
+func (v *ApiVariable) MultilineString(indent string) string {
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "%s", indent)
```
derekparker: Use `buf.WriteString(indent)` instead.

proc/variables.go
```diff
@@ -28,20 +28,81 @@ const (
...
 	fieldType dwarf.Type
+
+	Children []Variable
+
+	loaded     bool
+	Unreadable error
 }
 
+// Variable describes a variable.
+type ApiVariable struct {
```
derekparker: Why not define this in api/types?

derekparker:
@aarzilli I agree with what @kostya-sh is saying. The prettyprint logic should be in `terminal/` or `api/types`. I think I'd prefer `terminal/` since `api/types` should remain relatively generic. I think the `api.Variable` type that the client receives should have all of the necessary information for the terminal client to pretty print.

aarzilli:
So, do you want to move variables_test.go to service/test? Those tests depend on the serialization code (and, in general, using serialization is a convenient way to test these things).

derekparker:
@aarzilli no, we shouldn't have to move all of variables_test.go to service/test, only the specific bits that deal with the stringifying of the data structure, unless I'm missing something. I think it is totally appropriate to test the raw memory -> `Variable` structure at the `proc` level, but anything that is purely concerned with output (pretty printing) should be moved elsewhere.

aarzilli:
Every test in variables_test.go uses stringification to do its job, doing it without stringification seems pretty inconvenient. For example this test case:

```
{"a12", "[]main.FooBar len: 2, cap: 2, [{Baz: 4, Bur: \"d\"},{Baz: 5, Bur: \"e\"}]", "", "struct []main.FooBar", nil},
```
would become something like this:

```
{"a12", Variable{ Len: 2, Cap: 2, Children: []Variable{ { Children: []Variable{ { Name: "Baz", value: 4 }, { Name: "Bur", value: "d" } } }, { Children: []Variable{ { Name: "Baz", value: 5 }, { Name: "Bur", value: "e" } } } } }, "", "struct []main.FooBar", nil }
```

`a13` in TestVariableEvaluation is even worse because of all the pointers involved.

derekparker:
@aarzilli right, that is inconvenient. What if we kept some basic tests in proc to assert about simple evals and small things like your patch to include []Children and then move the rest of the tests over to service/test and test it at the post-serialization level?

aarzilli:
Done.

proc/variables.go
```diff
@@ -28,20 +28,47 @@ const (
...
+
+	Children []Variable
+
+	loaded     bool
+	Unreadable error
 }
 
+type SimpleType string
+
+const (
```
derekparker: On second thought, instead of defining our own simple types, the Go linker actually provides this information in the Dwarf info, we just need to parse it: https://github.com/golang/debug/blob/97a893dc61d8b085786f92514092a127fe25589c/dwarf/const.go#L89. Once we have that value we can use `reflect.Kind` to get a string representing the Go type.

proc/proc_test.go
```diff
@@ -814,3 +815,217 @@ func TestIssue239(t *testing.T) {
...
+			}
+			if v.Cap != cap {
+				t.Fatalf("%s cap: expected: %d got: %d", name, cap, v.Cap)
+			}
+			if len(v.Children) != childrenlen {
+				t.Fatalf("%s children len: expected %d got: %d", name, childrenlen, len(v.Children))
+			}
+		}
+
+		avar("a1", StString, "foofoofoofoofoofoo", 18, 0, 0)
```
derekparker: Any reason to prefer this over a table driver test?

proc/variables.go
aarzilli: I need a dwarf.Entry to do that and when I dereference pointers, access struct members and index arrays I don't have it; I only have a dwarf.Type and there is no easy way to go from that to a type entry. 

To do what you suggest I would have to convert the dwarf.Type into a String and search for its entry in newVariable. Should I do that?

derekparker: We have access to the dwarf.Entry in `scope.extractVarInfoFromEntry` so why not just parse it there and store it as a field in the `Variable` struct?

aarzilli:
`scope.extractVarInfoFromEntry` is not the only thing that makes variables, `toField`, `maybeDereference` and `loadArrayValues` also do. Those do not have entries.

derekparker:
@aarzilli good point. A few ideas:

1) Add method to scope to look up entry for a given name.
2) Fill out the GoType for all []Children at the highest level
3) Store the dwarf.Entry on Variable, that way we can use `Entry.Offset` to seek the reader and parse the children entries for structs, etc..

Thoughts?

aarzilli:
I've started doing this change but tbh I'm not liking it. Tracking both `dwarf.Type` and the type entry is verbose and redundant. Tracking only the type entry wouldn't be redundant but far more verbose. What's the objection to using `golang.org/x/debug/dwarf` that keeps track of AttrGoKind in `dwarf.CommonType`?

derekparker:
So, CommonType has `ReflectKind` and `Offset`, so that should be all we need (likely just `ReflectKind`). Ideally we replace the `SimpleType` stuff with a switch on `ReflectKind` if we can, so we don't have to define all of those types ourselves and just rely on the actual information given in the Dwarf info. Does that make sense or did I miss your point? If so ping me on gitter and we can talk through it.

aarzilli:
All correct, however only CommonType from `golang.org/x/debug/dwarf` has `ReflectKind`, the one from `debug/dwarf` in go's standard library doesn't (yet?).
If we switch we can remove `SimpleType` from `Variable` (and all the St* constants) and just use `Variable.RealType.ReflectKind`.

derekparker:
@aarzilli you're right, that's annoying. I'd much rather rely on the stdlib package, as that is more likely to be actively maintained for the time being. My current thought is, lets vendor debug/dwarf with the patch applied to have offset and reflectkind on commontype, and then submit a patch to get it merged in upstream before the 1.6 code freeze.

Thoughts?

aarzilli:
>Thoughts?

Actually, we can't do this one either. The code in `golang.com/x/debug/dwarf` discards the value of `ReflectKind` in some circumstances (strings and slices at least) so we can't use it.

derekparker:
@aarzilli alright, go ahead and change SimpleType to reflect.Kind and let's leave it at that for now.

aarzilli:
Done

proc/variables.go
```diff
@@ -119,12 +144,48 @@ func newVariable(name string, addr uintptr, dwarfType dwarf.Type, thread *Thread
...
+}
+
+func (v *Variable) clone() *Variable {
+	r := *v
+	return &r
 }
 
 func (v *Variable) toField(field *dwarf.StructField) (*Variable, error) {
+	if v.Unreadable != nil {
+		return v.clone(), nil
```
derekparker: Are we ok to ignore the `Unreadable` error here? Should we return it as an error instead of cloning?

aarzilli:
> Are we ok to ignore the Unreadable error here?

If I didn't miss anything it should be.

> Should we return it as an error instead of cloning?

IMHO errors should be for actual errors: either we can't make sense of the debug information or there is an error in the user supplied expression. Being unable to read memory isn't an error, it just means we are dealing with uninitialized or corrupted memory, it should be normal for a debugger.

derekparker:
@aarzilli it looks like there is still code in this PR for handling poor scoping info, etc.. that should have been removed when submitting #270, i.e. the various `extractVar*` methods returning a slice instead of the var itself.

aarzilli:
> it looks like there is still code in this PR for handling poor scoping info

removed

derekparker:
@ironcladlou would like your input here as well with regard to the work you've done on #278.

ironcladlou:
Still catching up, but in general this approach while more complex for clients seems to be way closer to the end-game than #278 (which is easy to render but lacks all the type information).

One thought regarding https://github.com/derekparker/delve/pull/268#issuecomment-147798767: The example justifying testing against rendered output could perhaps be simplified with test-scoped functions and types to build `Variable`s from more compact commands/structs. I think it's important to have strong tests against the internal types; relying on a particular rendering of the variables for this particular class of test is only going to make it even harder to change the internal types in the future (something we're already getting a taste of.)

Getting variable access solvent for clients is extremely important for any useful tool integration, so I'm wary of pulling the tests up away from `proc`.

ironcladlou:
Quick clarification:

>Getting variable access solvent for clients is extremely important for any useful tool integration, so I'm wary of pulling the tests up away from proc.

I see that the tests are still in `proc` but rely on rendered output, so the overall point still stands.

derekparker:
> Still catching up, but in general this approach while more complex for clients seems to be way closer to the end-game than #278 (which is easy to render but lacks all the type information).

Can you expand on that a bit? What in particular is more complex for clients and do you have any thoughts on what could be simplified?

> I see that the tests are still in proc but rely on rendered output, so the overall point still stands.

That's true, however I think possibly a mixed approach could be useful here. We certainly need to assert on rendered output, however maybe that's best kept in `service/test` and anything that asserts about the actual underlying data structure that Delve uses (i.e. `Variable`) could be left in proc. The latter being type information, value, kind, etc. That would complicate the `proc` tests but I think it makes sense as far as separation of concerns goes.

ironcladlou:
>>Still catching up, but in general this approach while more complex for clients seems to be way closer to the end-game than #278 (which is easy to render but lacks all the type information).

>Can you expand on that a bit? What in particular is more complex for clients and do you have any thoughts on what could be simplified?

Just that with the approach in #278, the client can basically print the raw JSON and get something instantly human readable (which was important to me while developing debugger infrastructure from scratch in Sublime), like:

```
Locals:
fde1 (*struct github.com/derekparker/delve/dwarf/frame.FrameDescriptionEntry): {
  "Instructions": [], 
  "CIE": null, 
  "end": "49", 
  "begin": "0", 
  "Length": "0"
}
foo (struct string): "testing!"
```

That's enough to be useful for me right now, but probably not enough for a more advanced IDE integration. My needs are likely simpler than those of the IntelliJ plugin or LiteIDE which have the ability to do things like render interactive tree views and stuff for drilling down (with Sublime I'm basically working with a static text view for my debugger output, so I'm trying to show only what's absolutely necessary at the moment.)

With the data structures in this PR, I'll have to write a renderer that can walk the children and print stuff according to type, etc, just like the one included here. Which was inevitable anyway. I can't really think of a simpler way to represent the data than you guys have come up with. I don't think it's imposing an unreasonable burden on the client or anything.

>>I see that the tests are still in proc but rely on rendered output, so the overall point still stands.

>That's true, however I think possibly a mixed approach could be useful here. We certainly need to assert on rendered output, however maybe that's best kept in service/test and anything that asserts about the actual underlying data structure that Delve uses (i.e. Variable) could be left in proc. The latter being type information, value, kind, etc. That would complicate the proc tests but I think it makes sense as far as separation of concerns goes.

Agree; I didn't mean to say there's no use in testing the renderers. Just that I think renderer tests are separate from the Variable tests (and that the latter should be tested directly against Variables, not against rendered output, and that there are strategies for making that less annoying).

aarzilli:
I started writing some code for the expression evaluator and realized that it's much easier if `Variable.Value` is a `go/constant.Value` instead of `interface{}`, I've added a commit to change this.

proc/variables.go
```diff
@@ -29,16 +31,24 @@ const (
...
+	DwarfType dwarf.Type
+	RealType  dwarf.Type
+	Kind      reflect.Kind
 	thread    *Thread
 
-	Len       int64
-	Cap       int64
+	Value constant.Value
+
+	Len int64
```
derekparker: These fields are fairly obvious but please add some comments with some docs explaining these fields and when they are valid, etc.

proc/variables.go
```diff
@@ -29,16 +31,24 @@ const (
...
+	Cap int64
+
 	base      uintptr
 	stride    int64
 	fieldType dwarf.Type
+
+	Children []Variable
```
derekparker: Let's add some documentation here to explain this field.

proc/variables.go
```diff
@@ -29,16 +31,24 @@ const (
...
+
+	loaded     bool
```
derekparker: Add some docs to explain this field as well.

derekparker:
I've been really busy the past several days, however I'm going to finish reviewing this tonight, from what I remember the code LGTM but I'll go through everything tonight and put in the final review if anything needs to be changed and get this merged in asap.

aarzilli:
Regarding documentation on `proc.Variable`'s fields, I documented them in `api.Variable`, is it enough to copy those comments over?

proc/variables.go
```diff
@@ -91,24 +101,40 @@ type EvalScope struct {
...
-		dwarfType: dwarfType,
+		DwarfType: dwarfType,
 		thread:    thread,
-		Type:      dwarfType.String(),
 	}
 
-	switch t := dwarfType.(type) {
+	v.RealType = v.DwarfType
+	for {
+		if tt, ok := v.RealType.(*dwarf.TypedefType); ok {
```
ironcladlou: I don't understand the purpose of this for loop. And doesn't it loop forever if `v.RealType` is a `*dwarf.TypedefType`?

aarzilli: It resolves typedefs. For example:

```
type notInt int

func main() {
    var i notInt
}
```

`i`'s `DwarfType` will be a `*dwarf.TypedefType` named `notInt`, after that loop `RealType` will be `int`. It's not an infinite loop because `tt.Type` is the type named by the typedef and:

```
type not legal
type legal not
```

does not compile.

ironcladlou: Oops, now I see that you reassign `v.RealType` which changes the `if` input and can break the loop. Sorry for the noise!

derekparker:
merged

PR #270, opened 2015-10-06
Workaround for ambiguous variable names

aarzilli:
Depends on PR #268 

Workaround for #186, #106, #83


derekparker:
Going to close this out for now, we may revisit once we get closer to announcing a 1.0 release. I'd really rather not implement workarounds, instead I'd favor trying to make sure things get fixed upstream.

Issue #262, opened 2015-10-01
Continue command doesn't work in certain situations


ignatov:
```go
package main
import "fmt"

func typicalFunction() (res int) {
	defer func() {
		res = 2
		return
	}()
	res = 10
	return // setup breakpoint here
}

func main() {
	fmt.Println(typicalFunction())
}
```
Delve revision: af16cfa90b04ffda668dfd98d73247d2e9499a2d

```
> main.typicalFunction() ./defer.go:10

    5:		defer func() {
    6:			res = 2
    7:			return
    8:		}()
    9:		res = 10
=> 10:		return
   11:	}
   12:
   13:	func main() {
   14:		fmt.Println(typicalFunction())

(dlv) c
> main.typicalFunction() ./defer.go:10

    5:		defer func() {
    6:			res = 2
    7:			return
    8:		}()
    9:		res = 10
=> 10:		return
   11:	}
   12:
   13:	func main() {
   14:		fmt.Println(typicalFunction())
```

aarzilli:
Are you sure this happens on that revision? Because I can only reproduce this on my [fix-247 branch](https://github.com/aarzilli/delve/tree/fix-247). I have fixed it there but I can't think of a reason why this behaviour would happen on master.

ignatov:
The same on f2e5214e1b5c on Mac OS X 10.11 El Capitan.

derekparker:
I was just able to reproduce this on master.

aarzilli:
A couple of weeks ago I managed to get an OS X VM working so today I thought I would take a look at this bug.
It turns out that the problem is very similar to my own fuckup in PR #259: `proc.(*Thread).singleStep` calls `t.dbp.trapWait`, and trapWait thinks we hit the breakpoint instead of stepping through a `NOP` instruction, since `NOP` (which is the instruction the breakpoint in question is set to) is one byte long, just like `INT 3`.

One possible solution is to call `mach_port_wait` directly, initially I did this:
```
C.mach_port_wait(C.mach_port_t(t.os.thread_ct))
``` 
thinking it would wait for a single thread, but it didn't work (I only have a vague idea of how this things work). Calling `C.mach_port_wait(t.dbp.os.portSet)` instead works, but my opinion is that waiting for just the specific thread would be better.

Another option is to add an 'enabled' flag to breakpoints and temporarily set it to false when we are stepping over them, then `trapWait` would know that the breakpoint wasn't hit.

derekparker:
[...]

PR #276, opened 2015-10-16
[WIP] Windows support

### LINKED ISSUE #198: Windows support
Add support for Windows.

lukehoban:
Initial work on porting Delve to support Windows.  

There's a lot more to do before this is ready to be merged, but most of the core infrastructure for Windows support is in place, and some basic Delve scenarios are working in simple scenarios (launching, setting breakpoints, continuing, breaking on breakpoints, inspecting registers, seeing sources, inspecting locals, exiting, etc.).

Fixes #198.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"sync"
 
-	sys "golang.org/x/sys/unix"
-
 	"github.com/derekparker/delve/dwarf/frame"
 	"github.com/derekparker/delve/dwarf/line"
 	"github.com/derekparker/delve/dwarf/reader"
@@ -112,7 +110,7 @@ func (dbp *Process) Detach(kill bool) (err error) {
 			return
 		}
 		if kill {
-			err = sys.Kill(dbp.Pid, sys.SIGINT)
+			err = killProcess(dbp.Pid)
 		}
 	})
 	return
@@ -160,6 +158,7 @@ func (dbp *Process) LoadInformation(path string) error {
 
 // FindFileLocation returns the PC for a given file:line.
 func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
+	fileName = filepath.ToSlash(fileName)
 	pc, _, err := dbp.goSymTable.LineToPC(fileName, lineno)
 	if err != nil {
 		return 0, err
@@ -253,8 +252,9 @@ func (dbp *Process) ClearBreakpoint(addr uint64) (*Breakpoint, error) {
 	return bp, nil
 }
 
+
 // Status returns the status of the current main thread context.
-func (dbp *Process) Status() *sys.WaitStatus {
+func (dbp *Process) Status() *WaitStatus {
 	return dbp.CurrentThread.Status
 }
 
@@ -592,7 +592,7 @@ func (dbp *Process) FindBreakpoint(pc uint64) (*Breakpoint, bool) {
 func initializeDebugProcess(dbp *Process, path string, attach bool) (*Process, error) {
 	if attach {
 		var err error
-		dbp.execPtraceFunc(func() { err = sys.PtraceAttach(dbp.Pid) })
+		dbp.execPtraceFunc(func() { err = PtraceAttach(dbp.Pid) })
 		if err != nil {
 			return nil, err
 		}
--- a/proc/proc_windows.go
+++ b/proc/proc_windows.go
@@ -1,0 +1,445 @@
+package proc
+
+// #include "windows.h"
+import "C"
+import (
+	"debug/gosym"
+	"debug/pe"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"syscall"
+	"unsafe"
+
+	sys "golang.org/x/sys/windows"
+
+	"github.com/derekparker/delve/dwarf/frame"
+	"github.com/derekparker/delve/dwarf/line"
+)
+
+const (
+	// DEBUGONLYTHISPROCESS tracks https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx
+	DEBUGONLYTHISPROCESS = 0x00000002
+)
+
+// OSProcessDetails holds Windows specific information.
+type OSProcessDetails struct {
+	hProcess    sys.Handle
+	breakThread int
+}
+
+// Launch creates and begins debugging a new process.
+func Launch(cmd []string) (*Process, error) {
+	argv0Go, err := filepath.Abs(cmd[0])
+	if err != nil {
+		return nil, err
+	}
+	// Make sure the binary exists.
+	if filepath.Base(cmd[0]) == cmd[0] {
+		if _, err := exec.LookPath(cmd[0]); err != nil {
+			return nil, err
+		}
+	}
+	if _, err := os.Stat(argv0Go); err != nil {
+		return nil, err
+	}
+
+	argv0, _ := syscall.UTF16PtrFromString(argv0Go)
+
+	// Duplicate the stdin/stdout/stderr handles
+	files := []uintptr{uintptr(syscall.Stdin), uintptr(syscall.Stdout), uintptr(syscall.Stderr)}
+	p, _ := syscall.GetCurrentProcess()
+	fd := make([]syscall.Handle, len(files))
+	for i := range files {
+		err := syscall.DuplicateHandle(p, syscall.Handle(files[i]), p, &fd[i], 0, true, syscall.DUPLICATE_SAME_ACCESS)
+		if err != nil {
+			return nil, err
+		}
+		defer syscall.CloseHandle(syscall.Handle(fd[i]))
+	}
+
+	// Initialize the startup info and create process
+	si := new(sys.StartupInfo)
+	si.Cb = uint32(unsafe.Sizeof(*si))
+	si.Flags = syscall.STARTF_USESTDHANDLES
+	si.StdInput = sys.Handle(fd[0])
+	si.StdOutput = sys.Handle(fd[1])
+	si.StdErr = sys.Handle(fd[2])
+	pi := new(sys.ProcessInformation)
+	err = sys.CreateProcess(argv0, nil, nil, nil, true, DEBUGONLYTHISPROCESS, nil, nil, si, pi)
+	if err != nil {
+		return nil, err
+	}
+	sys.CloseHandle(sys.Handle(pi.Process))
+	sys.CloseHandle(sys.Handle(pi.Thread))
+
+	dbp := New(int(pi.ProcessId))
+
+	switch runtime.GOARCH {
+	case "amd64":
+		dbp.arch = AMD64Arch()
+	}
+
+	// Note - it should not actually be possible for the
+	// call to waitForDebugEvent to fail, since Windows
+	// will always fire a CreateProcess event immediately
+	// after launching under DEBUGONLYTHISPROCESS.
+	var tid, exitCode int
+	dbp.execPtraceFunc(func() {
+		tid, exitCode, err = dbp.waitForDebugEvent()
+	})
+	if err != nil {
+		return nil, err
+	}
+	if tid == 0 {
+		dbp.postExit()
+		return nil, ProcessExitedError{Pid: dbp.Pid, Status: exitCode}
+	}
+
+	return initializeDebugProcess(dbp, argv0Go, false)
+}
+
+// Attach to an existing process with the given PID.
+func Attach(pid int) (*Process, error) {
+	return nil, fmt.Errorf("not implemented: Attach")
+}
+
+// Kill kills the process.
+func (dbp *Process) Kill() error {
+	if dbp.exited {
+		return nil
+	}
+	if !dbp.Threads[dbp.Pid].Stopped() {
+		return errors.New("process must be stopped in order to kill it")
+	}
+	// TODO: Should not have to ignore failures here,
+	// but some tests appear to Kill twice causing
+	// this to fail on second attempt.
+	_ = C.TerminateProcess(C.HANDLE(dbp.os.hProcess), 1)
+	dbp.exited = true
+	return nil
+}
+
+func (dbp *Process) requestManualStop() error {
+	res := C.DebugBreakProcess(C.HANDLE(dbp.os.hProcess))
+	if res == C.FALSE {
+		return fmt.Errorf("failed to break process %d", dbp.Pid)
+	}
+	return nil
+}
+
+func (dbp *Process) updateThreadList() error {
+	// We ignore this request since threads are being
+	// tracked as they are created/killed in waitForDebugEvent.
+	return nil
+}
+
+func (dbp *Process) addThread(hThread sys.Handle, threadID int, attach bool) (*Thread, error) {
+	if thread, ok := dbp.Threads[threadID]; ok {
+		return thread, nil
+	}
+	thread := &Thread{
+		ID:  threadID,
+		dbp: dbp,
+		os:  new(OSSpecificDetails),
+	}
+	thread.os.hThread = hThread
+	dbp.Threads[threadID] = thread
+	if dbp.CurrentThread == nil {
+		dbp.SwitchThread(thread.ID)
+	}
+	return thread, nil
+}
+
+func (dbp *Process) parseDebugFrame(exe *pe.File, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	if sec := exe.Section(".debug_frame"); sec != nil {
+		debugFrame, err := sec.Data()
+		if err != nil && uint32(len(debugFrame)) < sec.Size {
+			fmt.Println("could not get .debug_frame section", err)
+			os.Exit(1)
+		}
+		if 0 < sec.VirtualSize && sec.VirtualSize < sec.Size {
+			debugFrame = debugFrame[:sec.VirtualSize]
+		}
+		dbp.frameEntries = frame.Parse(debugFrame)
+	} else {
+		fmt.Println("could not find .debug_frame section in binary")
+		os.Exit(1)
+	}
+}
+
+// Borrowed from https://golang.org/src/cmd/internal/objfile/pe.go
+func findPESymbol(f *pe.File, name string) (*pe.Symbol, error) {
+	for _, s := range f.Symbols {
+		if s.Name != name {
+			continue
+		}
+		if s.SectionNumber <= 0 {
+			return nil, fmt.Errorf("symbol %s: invalid section number %d", name, s.SectionNumber)
+		}
+		if len(f.Sections) < int(s.SectionNumber) {
+			return nil, fmt.Errorf("symbol %s: section number %d is larger than max %d", name, s.SectionNumber, len(f.Sections))
+		}
+		return s, nil
+	}
+	return nil, fmt.Errorf("no %s symbol found", name)
+}
+
+// Borrowed from https://golang.org/src/cmd/internal/objfile/pe.go
+func loadPETable(f *pe.File, sname, ename string) ([]byte, error) {
+	ssym, err := findPESymbol(f, sname)
+	if err != nil {
+		return nil, err
+	}
+	esym, err := findPESymbol(f, ename)
+	if err != nil {
+		return nil, err
+	}
+	if ssym.SectionNumber != esym.SectionNumber {
+		return nil, fmt.Errorf("%s and %s symbols must be in the same section", sname, ename)
+	}
+	sect := f.Sections[ssym.SectionNumber-1]
+	data, err := sect.Data()
+	if err != nil {
+		return nil, err
+	}
+	return data[ssym.Value:esym.Value], nil
+}
+
+// Borrowed from https://golang.org/src/cmd/internal/objfile/pe.go
+func pcln(exe *pe.File) (textStart uint64, symtab, pclntab []byte, err error) {
+	var imageBase uint64
+	switch oh := exe.OptionalHeader.(type) {
+	case *pe.OptionalHeader32:
+		imageBase = uint64(oh.ImageBase)
+	case *pe.OptionalHeader64:
+		imageBase = oh.ImageBase
+	default:
+		return 0, nil, nil, fmt.Errorf("pe file format not recognized")
+	}
+	if sect := exe.Section(".text"); sect != nil {
+		textStart = imageBase + uint64(sect.VirtualAddress)
+	}
+	if pclntab, err = loadPETable(exe, "runtime.pclntab", "runtime.epclntab"); err != nil {
+		// We didn't find the symbols, so look for the names used in 1.3 and earlier.
+		// TODO: Remove code looking for the old symbols when we no longer care about 1.3.
+		var err2 error
+		if pclntab, err2 = loadPETable(exe, "pclntab", "epclntab"); err2 != nil {
+			return 0, nil, nil, err
+		}
+	}
+	if symtab, err = loadPETable(exe, "runtime.symtab", "runtime.esymtab"); err != nil {
+		// Same as above.
+		var err2 error
+		if symtab, err2 = loadPETable(exe, "symtab", "esymtab"); err2 != nil {
+			return 0, nil, nil, err
+		}
+	}
+	return textStart, symtab, pclntab, nil
+}
+
+func (dbp *Process) obtainGoSymbols(exe *pe.File, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	_, symdat, pclndat, err := pcln(exe)
+	if err != nil {
+		fmt.Println("could not get Go symbols", err)
+		os.Exit(1)
+	}
+
+	pcln := gosym.NewLineTable(pclndat, uint64(exe.Section(".text").Offset))
+	tab, err := gosym.NewTable(symdat, pcln)
+	if err != nil {
+		fmt.Println("could not get initialize line table", err)
+		os.Exit(1)
+	}
+
+	dbp.goSymTable = tab
+}
+
+func (dbp *Process) parseDebugLineInfo(exe *pe.File, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	if sec := exe.Section(".debug_line"); sec != nil {
+		debugLine, err := sec.Data()
+		if err != nil && uint32(len(debugLine)) < sec.Size {
+			fmt.Println("could not get .debug_line section", err)
+			os.Exit(1)
+		}
+		if 0 < sec.VirtualSize && sec.VirtualSize < sec.Size {
+			debugLine = debugLine[:sec.VirtualSize]
+		}
+		dbp.lineInfo = line.Parse(debugLine)
+	} else {
+		fmt.Println("could not find .debug_line section in binary")
+		os.Exit(1)
+	}
+}
+
+func (dbp *Process) findExecutable(path string) (*pe.File, error) {
+	if path == "" {
+		// TODO: Find executable path from PID/handle on Windows:
+		// https://msdn.microsoft.com/en-us/library/aa366789(VS.85).aspx
+		return nil, fmt.Errorf("not yet implemented")
+	}
+	f, err := os.OpenFile(path, 0, os.ModePerm)
+	if err != nil {
+		return nil, err
+	}
+	peFile, err := pe.NewFile(f)
+	if err != nil {
+		return nil, err
+	}
+	data, err := peFile.DWARF()
+	if err != nil {
+		return nil, err
+	}
+	dbp.dwarf = data
+	return peFile, nil
+}
+
+func (dbp *Process) waitForDebugEvent() (threadID, exitCode int, err error) {
+	var debugEvent C.DEBUG_EVENT
+	for {
+		// Wait for a debug event...
+		res := C.WaitForDebugEvent(&debugEvent, C.INFINITE)
+		if res == C.FALSE {
+			return 0, 0, fmt.Errorf("could not WaitForDebugEvent")
+		}
+
+		// ... handle each event kind ...
+		unionPtr := unsafe.Pointer(&debugEvent.u[0])
+		switch debugEvent.dwDebugEventCode {
+		case C.CREATE_PROCESS_DEBUG_EVENT:
+			debugInfo := (*C.CREATE_PROCESS_DEBUG_INFO)(unionPtr)
+			hFile := debugInfo.hFile
+			if hFile != C.HANDLE(uintptr(0)) /* NULL */ && hFile != C.HANDLE(uintptr(0xFFFFFFFFFFFFFFFF)) /* INVALID_HANDLE_VALUE */ {
+				res = C.CloseHandle(hFile)
+				if res == C.FALSE {
+					return 0, 0, fmt.Errorf("could not close create process file handle")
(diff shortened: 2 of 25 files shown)


ignatov:
Great!

lukehoban:
A quick update on progress:

- All debugger features are generally working (except `attach`).
- Unit tests are clean on Windows.
- Unit tests remain clean on Mac/Linux (modulo known flakiness).
- Appveyor CI support added to guard against regressions.

A few of the known remaining work items:

- The `TestNextNetHTTP` test is disabled and needs to be investigated.
- Attach support is not yet enabled on Windows.
- Thread information is not displayed nicely for most non-active threads because the PC on those threads is in `ntdll` or other dynamically loaded code.  To address this, we would need to load symbols on Windows and merge that information with the DWARF symbols.  This should be possible, but worth discussing before implementing.
- There are a handful of TODOs to investigate and improve on workarounds in the existing implementation.
- Effort was made to have minimal impact on the Mac/Linux codebase so far.  In the future, it likely will make sense to define a more formal platform abstraction layer that is platform independent.

However, at this point it is worth starting to get more eyes on this code and discussing if/when this should be merged into master.

/cc @derekparker

proc/threads_windows.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	
+	// TODO: It's possible that this resume/trapWait will 
+	// land us on another thread. We need to guard for that 
+	// and loop here till we sucessfully land back on our thread.
+	err = t.resume()
+	if err != nil {
+		return err
+	}
+	_, err = t.dbp.trapWait(0)
```
aarzilli: I think this is a bug, calling `trapWait` inside `singleStep` means delve will get stuck on breakpoints set on 1 byte instructions. See #262 and the commit I added today to #259.

proc/proc_windows.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+	}
+}
+
+func (dbp *Process) trapWait(pid int) (*Thread, error) {	
+	for {
+		var err error
+		var tid int
+		var exitCode int
+		dbp.execPtraceFunc(func() {
```
derekparker: Is there a restriction in Windows that these syscalls must originate from the same thread?

proc/test/support.go
```diff
@@ -48,13 +50,23 @@ func BuildFixture(name string) Fixture {
...
+
 	// Build the test binary
-	if err := exec.Command("go", "build", "-gcflags=-N -l", "-o", tmpfile, path).Run(); err != nil {
+	if err := exec.Command("go", buildFlags...).Run(); err != nil {
 		fmt.Printf("Error compiling %s: %s\n", path, err)
 		os.Exit(1)
 	}
 
 	source, _ := filepath.Abs(path)
+	if runtime.GOOS == "windows" {
```
derekparker: Same question from above applies here.

proc/threads_windows.go
```diff
@@ -0,0 +1,142 @@
...
+	res, err := C.GetThreadContext(C.HANDLE(unsafe.Pointer(t.os.hThread)), &context)
+	if res == 0 {
+		return err
+	}
+	
+	//TODO: Is it really okay to not decrement to IP?
+	//context.Rip--;
+	context.EFlags |= 0x100;
+	
+
```
derekparker: nit: cleanup extra newline here.

proc/threads_windows.go
derekparker: That's true. @aarzilli is there a test for that? If not we should add one, because for now let's simply not use trapWait here, but shortly after this and some other PRs land I'd like to look into ensuring that trapWait is safe to call in that situation. Either way I'd like to prevent that as a regression in the future.

proc/threads_windows.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	
+	return nil
+}
+
+func (t *Thread) resume() error {
+	t.running = true
+	var res C.WINBOOL
+	t.dbp.execPtraceFunc(func() {
+		//TODO: Note that we are ignoring the thread we were asked to continue and are continuing the
```
derekparker: Any reason for this? A `thread.resume` call should only act on that that.

proc/threads_windows.go
```diff
@@ -0,0 +1,142 @@
...
+		res = C.ContinueDebugEvent(C.DWORD(t.dbp.Pid), C.DWORD(t.dbp.os.breakThread), C.DBG_CONTINUE)
+	})
+	if res == 0 {
+		return fmt.Errorf("Could not ContinueDebugEvent.")	
+	}
+	return nil
+}
+
+func (thread *Thread) blocked() bool {
+	// TODO: Probably incorrect - what are teh runtime functions that
```
derekparker: I can look into this.

service/debugger/locations.go
```diff
@@ -111,6 +116,10 @@ func parseLocationSpecDefault(locStr, rest string) (LocationSpec, error) {
 	spec := &NormalLocationSpec{}
 
 	spec.Base = v[0]
+	if runtime.GOOS == "windows" {
+		// Normalize paths to "/" to match Go symbol table paths
+		spec.Base = strings.Replace(spec.Base, "\\", "/", -1)
```
derekparker: See above filepath comment.

proc/threads_windows.go
aarzilli: #259 has `TestIssue262` which does catch this on linux and OS X, however it's highly dependent on what the compiler does.

proc/proc.go
```diff
@@ -151,6 +149,9 @@ func (dbp *Process) LoadInformation(path string) error {
 }
 
 func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
+	if runtime.GOOS == "windows" {
```
lukehoban: Unfortunately no - this is doing the opposite of that library function.  We need to turn a platform-specific path into one which always uses '/', since for some reason that is how paths are being encoded into DWARF, even on Windows.  

We could avoid the runtime check by always doing something like this - which will just be a no-op on unix:

```go
fileName = strings.Replace(fileName, string(os.PathSeparator), "/", -1)
```

proc/proc_windows.go
lukehoban: That's right - both WaitForDebugEvent and ContinueDebugEvent must be called from the same thread that created the process.

Details here (https://msdn.microsoft.com/en-us/library/windows/desktop/ms679285(v=vs.85).aspx) and here (https://msdn.microsoft.com/en-us/library/windows/desktop/ms681423(v=vs.85).aspx).

proc/test/support.go
lukehoban: Same answer as above - happy to change to use the code without a platform test if that's preferred.

proc/threads_windows.go
lukehoban: This is a temporary workaround for a problem somewhere else that is causing requests for resuming on a thread that should not be resumed. 

I believe it is related to the issue with the resume call from singleStep.  We should be able to remove this workaround once those other places are fixed.

lukehoban: One other note is that the Windows debugging APIs only allow continuing the same thread on which the previous debug event was raised.  When the debug event is raised all threads are paused and then when ContiuneDebugEvent is called all threads resume.

To correctly singleStep (or continue only a single thread while keeping others stopped), we will need to pause all the threads other than the one we want to step, then continue, then resume all the threads again after the break event is raised for the step.

proc/proc.go
lukehoban: I take it back - `filePath.ToSlash` does what's needed here and in the other two places you called out.  I'll use that.

proc/proc_windows.go
```diff
@@ -0,0 +1,369 @@
...
+	switch runtime.GOARCH {
+	case "amd64":
+		dbp.arch = AMD64Arch()
+	}
+	
+	dbp.execPtraceFunc(func() {
+		// TODO - We're ignoring the results because we assume we'll immediately hit
+		// the default breakpoint that Windows sets at process creation.
+		// Should perhaps be testing that we're not overlooking an exit event or similar?
+		_, _, err = dbp.waitForDebugEvent()
```
lukehoban: I'm not certain it's even possible, but I've left the TODO here and added a note about returning `ProcessExitedError` if it is possible.

derekparker:
@lukehoban sorry for the delay, have gotten really busy lately. I will be taking another look tomorrow. How is the work going to finish implementing features and get all tests to pass?

derekparker:
@lukehoban do you plan on 100% parity with this PR or to get an initial implementation? Because I'm fine with merging a solid implementation that doesn't have 100% parity, as long as we have issues to track the other features that must be implemented.

aarzilli:
My note about the `trapWait` call can be disregarded given the `Continue` refactoring we did on the other branch.

lukehoban:
@derekparker Apologies for the delay getting back to you.

I do think this is close to ready to merge in with a discrete set of known issues we can track in the issue tracker.

There is only one truly "incorrect" thing I am aware of in the current implementation, which is the issue mentioned in https://github.com/derekparker/delve/pull/276#discussion_r45100529.  

Other than that, the remaining points from https://github.com/derekparker/delve/pull/276#issuecomment-155274585 are places where there is not yet 100% parity, but what does work works correctly.  I'm happy to work on these as issues after merging this PR.

Certainly, merging this in with the CI turned on will help to make sure the Windows support doesn't regress going forward.

On the singleStep issue, I'll try to find some time to look at that in the next few days, and to bring this branch up to date with master.  Once I have an update on that, we can decide whether to merge this in.

dprime:
Hey,

I'm just trying to build this to test it out on my windows setup. Having a few problems getting things running. Should I be building on windows or cross compiling from linux to windows? When attempting to build on windows, it moans about lacking a gcc install and when on linux, it complains about `../../service/api/conversions.go:9:2: C source files not allowed when not using cgo or SWIG: threads_windows.c`
Any thoughts?

lukehoban:
Yeah - Delve uses cgo heavily, and cgo on Windows [requires a GCC toolchain](https://github.com/golang/go/wiki/cgo#windows).  I've been compiling on Windows with the MinGW64 tools.  

The [appveyor.yml](https://github.com/lukehoban/delve/blob/windows/appveyor.yml) has one particular toolchain configuration that works and is being used in the CI toolchain: `x86_64-4.9.2-release-win32-seh-rt_v4-rev3`.

Locally I've used a slightly different MinGW version: `x86_64-5.2.0-posix-seh-rt_v4-rev0`.

dprime:
Thanks, giving that a spin now.

dprime:
Builds and works very nicely. Issue now is how to integrate into the windows version of intellij? It just works in linux and I'm trying to figure out how to coax the windows version to realise I have a working delve binary. Sorry to hijack this PR thread for this.

ignatov:
@dprime You should create a small patch and build the plugin from sources.
E.g. `com.goide.dlv.DlvDebugProcess#IS_DLV_DISABLED = SystemInfo.is32Bit;`
and setup a desired `customDlvPath` in `gradle.properties`.

mistbow-git:
Confirmed! Windows version builds and works perfectly!
I need debugging support on windows very much!
Please resolve all conflicts and merge with main branch!
And build idea plugin with support dalve for windows already...
Thanks!

zanven42:
firstly i would like to say thank you for everyone who has been working on windows support and i am super excited to start using it, 
Secondly @mistbow-git these people are doing this for free without being paid (most likely) if you really want to see it in and finished and you are not willing to wait till they have the spare time to do it i suggest you start cracking on finishing it off yourself and give them a hand.

mistbow-git:
@zanven42 Join the words of gratitude!
I builded dalve for windows last nite from souces https://github.com/mistbow-git/delve with last commits from @derekparker master. All working good! But one test failed (
Then I build a custom plugin for IDEA and testing debugging feature. Its working!
Thank you very much for all! Waiting for relise!

zanven42:
yeah after i posted i noticed you were doing things and it got me excited hehe. ill have to get it building tonight and have a look.

kolkov:
If you want, I can share files ready for testing...

zanven42:
that would be handy thanks.

kolkov:
http://kolkov.ru/share/dlv/dlv.rar 
dlv.exe folder d:/Go/bin

derekparker:
@lukehoban can you please rebase this on master, squash commits and push. Could you also open issues for any functionality that this PR does not provide? I'd like to get this branch merged in soon, and add more visibility into what hasn't been finished.

Thanks again for all your work on this port!

shiena:
I also tried the windows version.
There is almost no problem, but there is one question.
Does not display the string in `fmt.Println("hello")` In the Windows version.
Do I have failed anything?

dlv2

lukehoban:
@derekparker Looks like there's some work to do to update this to catch up to the refactorings/changes that were merged into master in the last week or two.  I'm working on that now.

lukehoban:
@shiena Yes - I believe that's a missing feature of the Windows port currently.  I'll make sure that's one of the issues we open before merging this.

aarzilli:
@lukehoban the work that was done in `handleBreakpointOnThread` is now inside `setCurrentBreakpoints` (which should find all threads that are stopped on a breakpoint if there is more than one).

The process is always restarted by calling `(*Process).resume()`, this method is responsible for making threads stopped on breakpoints step over their breakpoints correctly, from what I understand this should be a better fit for windows.

shiena:
@lukehoban Thanks for your reply. I see.
This patch is sufficiently great even while now.

lukehoban:
I've pushed a few new commits which:

1. Brings this branch up to date with `derekparker/master`
2. Addresses some latent bugs that @aarzilli's recent refactorings uncovered
3. Fixes the issue with single stepping potentially stepping the wrong thread (https://github.com/derekparker/delve/pull/276#discussion_r45100529)
4. Adds support for redirecting stdin/stdout/stderr from the debuggee per @shiena's feedback

This appears to be passing tests on both Travis and Appveyor.

I will pull together the list of known issues and cleanup items and will squash these commits down tomorrow.

derekparker:
@lukehoban thank you for your work on this. I'll be looking through the code soon for a final review and we will get this merged in.

lukehoban:
@derekparker Great - thanks.  I'm working on some cleanup and the list of open issues, which I'll have available soon.  Might be worth waiting on that before doing another review pass.

lukehoban:
This branch is now up to date with master and squashed.  

Tests are passing on Appveyor (https://ci.appveyor.com/project/lukehoban/delve) and Travis (https://travis-ci.org/derekparker/delve/builds/103290576).

Code reviews welcomed. @derekparker @aarzilli 

Here's an inventory of remaining work I'm aware of.  I'm okay with merging this branch and opening issues to track each of these.  If there are any that should be fixed before merging, let me know.

- [ ] **TestDebugLinePrologueParser**
Only on some  machines and Go versions.  Appears to be because of buffers in the debugline data which are larger than they should be in Windows binaries?  I suspect there is a bug in the Go compiler on Windows that we'll need to work around.

- [ ] **TestNextConcurrent flakiness**
Occasional occurrences of the following errors:
`proc_test.go:355: Program did not continue to correct next location expected 11 was parallel_next.go:10`
`proc_test.go:355: Program did not continue to correct next location expected 11 was parallel_next.go:8`
`proc_test.go:355: Program did not continue to correct next location expected 10 was parallel_next.go:8`
`proc_test.go:410: Program did not continue to correct next location expected 11 was parallel_next.go:10`
`proc_test.go:403: Initial breakpoint triggered twice for the same goroutine`

- [ ] **Attach is not supported**
Should also add test coverage for attach on all platforms.

- [ ] **TestNextNetHTTP is disabled on Windows**
Temporarily disabled this test, need to investigate the failure.

- [ ] **Should we load symbols on Windows?**
Thread information is not displayed nicely for most non-active threads because the PC on those threads is in ntdll or other dynamically loaded code. To address this, we would need to load symbols on Windows and merge that information with the DWARF symbols. This should be possible, but worth discussing before implementing.

- [ ] **thread.halt() may need to actually halt threads in some cases**
See comment in implementation.

- [ ] **Can we remove canContinue?**
dbp.Step currently relies on checking this, but would be nice to factor it to not need this concept in addition to blocked and stopped.

- [ ] **Infrequent failures in TestBreakPointCount**
Pretty infrequent occurence, and seemingly always with 101 + 99.  So it appears that one of the breakpoint hits is getting recorded as being on the wrong thread. 

Once merged, we will also want to do:

- [ ] **Add Appveyor project tracking derekparker/delve and badge in README**

- [ ] **Update docs with info on building/running on Windows**
Mostly just MinGW requirement which is likely not present for many Windows Go users.

aarzilli:
> Can we remove canContinue?
> dbp.Step currently relies on checking this, but would be nice to factor it to not need this concept in addition to blocked and stopped.

Are you talking about `canStep`? In that case, my opinion is that `proc.(*Process).Step` shouldn't step all threads but just the one associated with `SelectedGoroutine` (if there is one). I don't think a user would expect all threads to step and I'm not sure what use that would have.

lukehoban:
Re: @aarzilli comment on `dbp.resume()`:

```go
	// In case we are now on a different thread, make sure we resume
	// the thread that is broken.
	thread = dbp.Threads[dbp.os.breakThread]
	if err := thread.resume(); err != nil {
		return err
	}
```
> Under what circumstance does this happen? Could it be that when two breakpoints are hit simultaneously the debug event for the second breakpoint gets erroneusly ignored here? This could be an explanation for some of the test failures.

The underlying problem is in `singleStep`.  The assumption there is that if we set the processor trap flag on the current thread, then suspend all the threads other than the current thread, then `ContinueDebugEvent`/`WaitForDebugEvent`, that the debug event we'll get is the single step, and therefore will be on the same thread we were on prior to the call to `step`.

It appears though that in rare cases, the debug event returned for `WaitForDebugEvent` is on a different thread.  The only way I can explain this is that either `SuspendThread` is not actually suspending the thread, or somehow the additional debug event was queued concurrently with the event raised from `WaitForDebugEvent` so that it gets returned before the single step.  

The second of these I am nearly certain should not be possible.  This would be along the lines of your suggestion that perhaps the second breakpoint event is being raised here before the `singleStep` gets returned.  But with the way the Windows debug APIs work, all threads should be synchronized on debug events, with every thread being suspended when a debug event is raised.  So there really should be no notion of two breakpoints being hit simultaneously.  (Though two threads may perhaps be able to have their PC on the breakpoint, with the second ready to raise another breakpoint exception debugevent once it's thread is continued?).  If this is the issue though, it's going to be hard to fix, since there would never be a safe way to singleStep without potentially triggering other debug events that should reenter other parts of Delve's implementation.

The former seems more likely, and [this post](https://blogs.msdn.microsoft.com/oldnewthing/20150205-00/?p=44743) suggests this may be where the problem is, and a potential workaround (calling `GetContext` to force synchronizing the effect of the `SuspendThread` call).

From a brief investigation just now, it seems plausible that all of the other intermittent test failures occur only after this condition occurs (though it happens a little more frequently than actual test failures).  In the last 20 minutes of running tests, in every case where there has been a test failure, this unexpected `singleStep` condition was hit at least once during the test.

I'll look into this further...

lukehoban:
> my opinion is that proc.(*Process).Step shouldn't step all threads but just the one associated with SelectedGoroutine (if there is one). I don't think a user would expect all threads to step and I'm not sure what use that would have.

Yes, that was what I would have expected as well.  I wasn't sure if there was some Linux/OSX specific reason for expecting all threads to step instead of just the SelectedGoroutine's thread.

I quickly tested on OSX changing `proc.(*Process).Step` to only step the SelectedGoroutine's thread, and I don't see any test failures.  Is there a scenario or test case that highlights the need for the current logic?  Or should this be changed across all platforms?

> Other than the go fmt nitpick this looks good to me.

I've fixed the `go fmt` in `threads_windows.go`.

joelday:
Is it reasonable to conclude that this is far enough, but with known issues?

joelday:
Also, @lukehoban, instructions on how to install from your PR/branch?

zanven42:
have mingw installed (64 if u dont know if ur go is 32bit or 64bit install). 

in derekparker's repo add lukehoban as a extra remote and checkout he's windows branch. 

go to the folder cmd -> dlv. open a command prompt. go build

hope that helps.

joelday:
@zanven42 Thanks! Managed to figure that out a few minutes ago.

This is what worked for me:
> git clone -q --branch=windows https://github.com/lukehoban/delve.git %GOPATH%\src\github.com\derekparker\delve
> set GO15VENDOREXPERIMENT=1
> cd %GOPATH%\src\github.com\derekparker\delve
> mingw32-make install

kolkov:
I use go to build dlv.exe
cd D:\Go\src\github.com\derekparker\delve\cmd\dlv
and then run: go build.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	if filepath.Base(cmd[0]) == cmd[0] {
+		if _, err := exec.LookPath(cmd[0]); err != nil {
+			return nil, err
+		}
+	}
+	if _, err := os.Stat(argv0Go); err != nil {
+		return nil, err
+	}
+
+    argv0, _ := syscall.UTF16PtrFromString(argv0Go)
```
derekparker: Looks like there is some strange formatting in this function. Can you make sure gofmt has been run on all source files?

derekparker:
> Yes, that was what I would have expected as well. I wasn't sure if there was some Linux/OSX specific reason for expecting all threads to step instead of just the SelectedGoroutine's thread.
>
> I quickly tested on OSX changing proc.(*Process).Step to only step the SelectedGoroutine's thread, and I don't see any test failures. Is there a scenario or test case that highlights the need for the current logic? Or should this be changed across all platforms?

Per discussion on IRC with @aarzilli we have decided that Step should only step the "current" thread. The concerns I had in mind when I initially implemented this to step all threads is essentially a non-issue, and is causing more problems.

I'm going to submit a patch today to step only the "current" thread, as well as do a final review of this branch. Once that patch lands @lukehoban the last major change to this branch (aside from review comment fixes) should be to match the new behavior of step. After all that is finished this branch will be merged and outstanding issues will have GH issues created to track them. The plan is (depending on how quickly the changes can be made) to have this branch merged into master this week (honestly, the ideal situation would be this evening or tomorrow). From there we can address the outstanding issues individually.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+		dbp.postExit()
+		return nil, ProcessExitedError{Pid: dbp.Pid, Status: exitCode}
+	}
+
+	return initializeDebugProcess(dbp, argv0Go, false)
+}
+
+// Attach to an existing process with the given PID.
+func Attach(pid int) (*Process, error) {
+	fmt.Println("Attach")
```
derekparker: Let's remove this and just return the error.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	return nil, fmt.Errorf("Not implemented: Attach")
+}
+
+// Kill kills the process.
+func (dbp *Process) Kill() (err error) {
```
derekparker: Named return isn't being used, remove it in favor of explicit return arg.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+	// TODO: Should not have to ignore failures here,
+	// but some tests appear to Kill twice causing
+	// this to fail on second attempt.
+	_ = C.TerminateProcess(C.HANDLE(dbp.os.hProcess), 1)
+	dbp.exited = true
+	return
+}
+
+func (dbp *Process) requestManualStop() (err error) {
```
derekparker: Same comment as above re: named return.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	res := C.DebugBreakProcess(C.HANDLE(dbp.os.hProcess))
+	if res == C.FALSE {
+		return fmt.Errorf("Failed to break process %d", dbp.Pid)
```
derekparker: nit: Lowercase `Failed`.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	return peFile, nil
+}
+
+func (dbp *Process) waitForDebugEvent() (threadID, exitCode int, err error) {
+	var debugEvent C.DEBUG_EVENT
+	for {
+		// Wait for a debug event...
+		res := C.WaitForDebugEvent(&debugEvent, C.INFINITE)
+		if res == C.WINBOOL(0) {
+			return 0, 0, fmt.Errorf("Could not WaitForDebugEvent")
```
derekparker: nit: lowercase `Could`.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+			}
+			break
+		case C.CREATE_THREAD_DEBUG_EVENT:
+			debugInfo := (*C.CREATE_THREAD_DEBUG_INFO)(unionPtr)
+			_, err = dbp.addThread(sys.Handle(debugInfo.hThread), int(debugEvent.dwThreadId), false)
+			if err != nil {
+				return 0, 0, err
+			}
+			break
+		case C.LOAD_DLL_DEBUG_EVENT, C.UNLOAD_DLL_DEBUG_EVENT, C.EXIT_THREAD_DEBUG_EVENT, C.OUTPUT_DEBUG_STRING_EVENT, C.RIP_EVENT:
```
derekparker: Why leave a thread exiting unimplemented? Should be pretty simple, just remove it from the list of known threads. That way dead threads won't clutter output of `threads` command.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	th := dbp.Threads[tid]
+	return th, nil
+}
+
+func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
+	wg.Done()
+}
+
+func (dbp *Process) wait(pid, options int) (int, *sys.WaitStatus, error) {
+	fmt.Println("wait")
```
derekparker: Remove this Println, just return error.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	return 0, nil, fmt.Errorf("Not implemented: wait")
```
derekparker: nit: lowercase `Not`.

ignatov:
@lukehoban Could you ping me directly when the PR will be merged, please.

proc/proc_windows.go
```diff
@@ -0,0 +1,423 @@
...
+	// threads that happen to have hit this BP. But doing so leads to periodic
+	// failures in the TestBreakpointsCounts test with hit counts being too high,
+	// which can be traced back to occurences of multiple threads hitting a BP
+	// at the same time.
+
+	// My guess is that Windows will correctly trigger multiple DEBUG_EVENT's
+	// in this case, one for each thread, so we should only handle the BP hit
+	// on the thread that the debugger was evented on.
+
+	err := trapthread.SetCurrentBreakpoint()
```
derekparker: nit: rewrite as `return trapthread.SetCurrentBreakpoint()`

proc/ptrace_windows.go
```diff
@@ -0,0 +1,13 @@
+package proc
+
+import (
+	"fmt"
+)
+
+func PtraceAttach(pid int) error {
+	return fmt.Errorf("Not implemented: PtraceAttach")
```
derekparker: nit: lowercase `Not`.

proc/ptrace_windows.go
```diff
@@ -0,0 +1,13 @@
...
+}
+
+func PtraceDetach(tid, sig int) error {
+	return fmt.Errorf("Not implemented: PtraceDetach")
```
derekparker: nit: lowercase `Not`.

proc/registers_windows_amd64.go
```diff
@@ -0,0 +1,166 @@
...
+
+// TLS returns the value of the register
+// that contains the location of the thread
+// local storage segment.
+func (r *Regs) TLS() uint64 {
+	return r.tls
+}
+
+// SetPC sets the RIP register to the value specified by `pc`.
+func (r *Regs) SetPC(thread *Thread, pc uint64) error {
```
derekparker: General nit: lowercase first letter of all new error messages.

proc/registers_windows_amd64.go
```diff
@@ -0,0 +1,166 @@
...
+		eflags: uint64(context.EFlags),
+		cs:     uint64(context.SegCs),
+		fs:     uint64(context.SegFs),
+		gs:     uint64(context.SegGs),
+		tls:    uint64(tls),
+	}
+	return regs, nil
+}
+
+func (thread *Thread) saveRegisters() (Registers, error) {
```
derekparker: Remove Println's and just return error.

proc/threads_windows.c
```diff
@@ -0,0 +1,16 @@
+#include "threads_windows.h"
+
+typedef NTSTATUS (WINAPI *pNtQIT)(HANDLE, LONG, PVOID, ULONG, PULONG);
+
+WINBOOL thread_basic_information(HANDLE h, THREAD_BASIC_INFORMATION* addr) {
```
derekparker: Looks like some formatting issues here.

proc/threads_windows.go
```diff
@@ -0,0 +1,171 @@
...
+		res = C.ContinueDebugEvent(C.DWORD(t.dbp.Pid), C.DWORD(t.ID), C.DBG_CONTINUE)
+	})
+	if res == C.FALSE {
+		return fmt.Errorf("Could not ContinueDebugEvent.")
+	}
+	return nil
+}
+
+func (t *Thread) blocked() bool {
+	// TODO: Probably incorrect - what are teh runtime functions that
```
derekparker: s/teh/the/

derekparker:
@lukehoban https://github.com/derekparker/delve/pull/351 has landed, fixing the single step behavior.

proc/threads_windows.go
```diff
@@ -0,0 +1,171 @@
...
+	pc, err := t.PC()
+	if err != nil {
+		return false
+	}
+	fn := t.dbp.goSymTable.PCToFunc(pc)
+	if fn == nil {
+		return false
+	}
+	switch fn.Name {
+	case "runtime.kevent", "runtime.mach_semaphore_wait", "runtime.usleep":
```
derekparker: We can definitely remove "runtime.mach_semaphore_wait" from this list.

derekparker:
@lukehoban also, let's just go ahead and add the Appveyor badge to the README in the PR, since CI is already setup there it seems. That way it'll just be there when this lands.

lukehoban:
I've addressed all the code review feedback in the latest commits.

> #351 has landed, fixing the single step behavior.

Great.  I've updated the windows branch to align with this (actually just removing some code).

> let's just go ahead and add the Appveyor badge to the README in the PR, since CI is already setup there it seems.

I think you'll need to set up an AppVeyor project to track `derekparker/delve`.  The one I have at https://ci.appveyor.com/project/lukehoban/delve is pointing at `lukehoban/delve`.

derekparker:
Just merged this branch, will now start creating issues to track outstanding work that needs to be done, and I will get AppVeyor setup as well. Thank you @lukehoban for all your work on this!

lukehoban:
:zap:

kolkov:
This is very cool! 
@ignatov as you asked!

ahmetalpbalkan:
congrats!

techtonik:
@derekparker looks like this article https://blog.gopheracademy.com/advent-2015/debugging-with-delve/ should be updated with new info about Windows support. )

Issue #251, opened 2015-09-25
Support for evaluating expressions


dlsniper:
One of the features of a debugger is to provide support for evaluating expressions.
One example of usage for this is when you have a watch sent on an expression such as:
```
watch a == true
```

```go
package main

func main() {
	a := false
	_ = a // a breakpoint here should evaluate the watch to false
}
```

Setting a breakpoint on line 5 should yield `false` for the evaluation of `a == true`.

I'm pretty sure this is a complex one to do but I haven't done the research yet.

ramigg:
+1

gshilin:
+2
Totally agree!
Please, please implement it!!!

aarzilli:
Expression evaluation was merged a while ago, we have issues #119 and #59 tracking missing features. I'm going to close this, if there is some feature that's needed a new issue specific to that should be opened.

PR #278, opened 2015-10-18
RFC: Introduce structured variable values

ironcladlou:
### :construction: WIP, for discussion only :construction:

Hello again! I'd like to get some early feedback on some changes I've made locally to support [GoTools](https://github.com/ironcladlou/GoTools) integration. This provides an alternative (and/or replacement) to the current string rendering of variable data. Instead of building up a string while loading variable data from the debugged process, I instead build up an `interface{}` which is one of `string` (primitives), `[]interface{}` (arrays), or `map[string]interface{}` (structs). This lets delve generically represent arbitrary Go variable types internally. The variable's value object is exposed via a new union type in the API called `Variable.ValueObject`.

The net result is that clients will get a parseable JSON object graph they can render however they like.

This could be expanded so that nested values (in arrays or maps) have type information (e.g. the types of fields or array elements), but I'm not sure if it's worth the complexity yet.

I'm not going to try and tackle porting unit tests (which could be rewriting cases or writing a backwards-compatible renderer) until I've got some feedback.

### DIFF
--- a/proc/variables.go
+++ b/proc/variables.go
@@ -28,12 +28,13 @@ const (
 
 // Represents a variable.
 type Variable struct {
-	Addr      uintptr
-	Name      string
-	Value     string
-	Type      string
-	dwarfType dwarf.Type
-	thread    *Thread
+	Addr        uintptr
+	Name        string
+	Value       string
+	ValueObject interface{}
+	Type        string
+	dwarfType   dwarf.Type
+	thread      *Thread
 
 	Len       int64
 	Cap       int64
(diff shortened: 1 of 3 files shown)


ironcladlou:
cc @pmorie

ironcladlou:
Oops, I totally forgot to check for existing issues before hacking away. I'll review #268 in detail tomorrow and join the discussion.

proc/variables.go
```diff
@@ -540,31 +541,35 @@ func (v *Variable) resolveTypedefs() *Variable {
...
 		}
 
 		if ptrv.Addr == 0 {
-			return fmt.Sprintf("%s nil", t.String()), nil
+			return nil, nil
 		}
 
 		// Don't increase the recursion level when dereferencing pointers
-		val, err := ptrv.loadValueInternal(printStructName, recurseLevel)
+		derefed, err := ptrv.loadValueInternal(recurseLevel)
```
pmorie: dRefd

pmorie:
This makes sense to me as a first stab. I think this has a touchpoint with #251

ironcladlou:
Closing in favor of the superior #268

Issue #117, opened 2015-05-03
Support printing address of variable


derekparker:
The following syntax should print the address of a variable, if it is not a pointer already:

` print &myVar`.

aarzilli:
Implemented by 43b64ec39e8188cd5b215d15ec0175868c548a8b

Issue #116, opened 2015-05-03
Support array/slice indexing for variable evaluation


derekparker:
Support for the following syntax:

` print arr[5]`.

Should respect array bounds and print proper error messages for out of bounds indexing.

aarzilli:
Implemented by 43b64ec39e8188cd5b215d15ec0175868c548a8b

PR #285, opened 2015-10-30
proc: Implements expression interpreter

aarzilli:
Supported operators:

- All (binary and unary) operators between basic types except <-, ++ and -- (includes & to take the address of an expression)
- Comparison operators between supported compound types
- Typecast of integer constants into pointer types
- struct members
- indexing of arrays, slices and strings
- slicing of arrays, slices and strings
- pointer dereferencing
- true, false and nil constants

Implements #116, #117 and #251 

The big missing things are:
- function calls
- support for some kinds (maps, interfaces, chans)
- support for some builtins that do not require function calls (like `len` for example)
- support for type casts between simple types

The weirdest thing here is that constants in go are untyped so evaluating a constant results in a variable with `Type == ""`. However for constant strings and constant booleans we could create a fake type.

### DIFF
--- a/proc/eval.go
+++ b/proc/eval.go
@@ -1,0 +1,759 @@
+package proc
+
+import (
+	"bytes"
+	"debug/dwarf"
+	"fmt"
+	"go/ast"
+	"go/constant"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"reflect"
+)
+
+// Returns the value of the given expression
+func (scope *EvalScope) EvalExpression(expr string) (*Variable, error) {
+	t, err := parser.ParseExpr(expr)
+	if err != nil {
+		return nil, err
+	}
+
+	ev, err := scope.evalAST(t)
+	if err != nil {
+		return nil, err
+	}
+	ev.loadValue()
+	return ev, nil
+}
+
+func (scope *EvalScope) evalAST(t ast.Expr) (*Variable, error) {
+	switch node := t.(type) {
+	case *ast.CallExpr:
+		if fnnode, ok := node.Fun.(*ast.Ident); ok && len(node.Args) == 2 && (fnnode.Name == "complex64" || fnnode.Name == "complex128") {
+			// implement the special case type casts complex64(f1, f2) and complex128(f1, f2)
+			return scope.evalComplexCast(fnnode.Name, node)
+		}
+		// this must be a type cast because we do not support function calls
+		return scope.evalTypeCast(node)
+
+	case *ast.Ident:
+		return scope.evalIdent(node)
+
+	case *ast.ParenExpr:
+		// otherwise just eval recursively
+		return scope.evalAST(node.X)
+
+	case *ast.SelectorExpr: // <expression>.<identifier>
+		// try to interpret the selector as a package variable
+		if maybePkg, ok := node.X.(*ast.Ident); ok {
+			if v, err := scope.packageVarAddr(maybePkg.Name + "." + node.Sel.Name); err == nil {
+				return v, nil
+			}
+		}
+		// if it's not a package variable then it must be a struct member access
+		return scope.evalStructSelector(node)
+
+	case *ast.IndexExpr:
+		return scope.evalIndex(node)
+
+	case *ast.SliceExpr:
+		if node.Slice3 {
+			return nil, fmt.Errorf("3-index slice expressions not supported")
+		}
+		return scope.evalReslice(node)
+
+	case *ast.StarExpr:
+		// pointer dereferencing *<expression>
+		return scope.evalPointerDeref(node)
+
+	case *ast.UnaryExpr:
+		// The unary operators we support are +, - and & (note that unary * is parsed as ast.StarExpr)
+		switch node.Op {
+		case token.AND:
+			return scope.evalAddrOf(node)
+
+		default:
+			return scope.evalUnary(node)
+		}
+
+	case *ast.BinaryExpr:
+		return scope.evalBinary(node)
+
+	case *ast.BasicLit:
+		return newConstant(constant.MakeFromLiteral(node.Value, node.Kind, 0), scope.Thread), nil
+
+	default:
+		return nil, fmt.Errorf("expression %T not implemented", t)
+
+	}
+}
+
+func exprToString(t ast.Expr) string {
+	var buf bytes.Buffer
+	printer.Fprint(&buf, token.NewFileSet(), t)
+	return buf.String()
+}
+
+// Eval expressions: complex64(<float const>, <float const>) and complex128(<float const>, <float const>)
+func (scope *EvalScope) evalComplexCast(typename string, node *ast.CallExpr) (*Variable, error) {
+	realev, err := scope.evalAST(node.Args[0])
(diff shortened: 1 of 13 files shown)


derekparker:
@aarzilli this looks great. Can you expose it via the `print` command? Print should evaluate a full expression instead of just the name of the variable.

derekparker:
@aarzilli actually, what all did you have in mind for this? I could see some useful applications beyond just improving `print`, but I wanted a way to actually play around with this feature in the debugger CLI.

derekparker:
Applied this simple patch to be able to play around with the expression evaluation:

```diff
diff --git a/terminal/command.go b/terminal/command.go
index 972c835..ba2b0b5 100644
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -625,7 +625,7 @@ func printVar(t *Term, scope api.EvalScope, args ...string) error {
        if len(args) == 0 {
                return fmt.Errorf("not enough arguments")
        }
-       val, err := t.client.EvalVariable(scope, args[0])
+       val, err := t.client.EvalVariable(scope, strings.Join(args, ""))
        if err != nil {
                return err
        }
```

This is really nice, great work!

aarzilli:
As you have noticed, it already is "kind of" attached to print, because I rerouted EvalVariable to use it (perhaps it should be renamed before the API is frozen?)

`set` also uses this through `SetVariable` and has a worse problem than `print` with interpreting its argument because we can't just split on the first space anymore. If we change the syntax of `set` to be:

```
     set <expression> = <expression>
```

we can use `go/ast` (in a rather hacky way) to find out the correct split point.

When map support will be added there will be an additional problem with terminal:

```
print someMap["a  string"]
```

won't work because the two spaces in the strings will get collapsed.

Beyond `print` and `set` I was thinking of hooking up linespecs to this so that when we have code like this:

```
var amap map[string]func(int) = ...
```

we can set breakpoints like this:

```
break *amap["some string"]
```

Finally this could be used to implement conditional breakpoints. That's what I have in mind.

derekparker:
@aarzilli I'm all for changing set syntax to `set <expression> = <expression>` I think that makes total sense, and we can feel it out before the API freeze. I think for now, let's expand print to evaluate these expressions, as I think that's a very powerful feature.

Lastly, as you mentioned this opens up possibilities for conditional breakpoints landing after this.

@aarzilli this is really great, thank you for this patch!

aarzilli:
> @aarzilli this is really great, thank you for this patch!

you're welcome :)

Would it be ok with you if I added a second commit on this PR that changes terminal/command like so:

1. commands get their arguments unsplit
2. the commands that need to split their arguments do it independently (currently I think this is just `stack`, `break` and `trace`)
3. add the split on `=` trick to `set`

derekparker:
@aarzilli that sounds good to me.

aarzilli:
Alright, I'll see if I can do it over the w/e.

aarzilli:
updated

derekparker:
@aarzilli the CI build seems to be legitimately broken, not a flicker failure.

aarzilli:
Forgot about terminal's unit tests.

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+	default:
+		return nil, fmt.Errorf("expression %T not implemented", t)
+
+	}
+}
+
+func exprToString(t ast.Expr) string {
+	var buf bytes.Buffer
+	printer.Fprint(&buf, token.NewFileSet(), t)
+	return string(buf.Bytes())
```
derekparker: Use `buf.String()` here.

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+		if !ok {
+			break
+		}
+		fnnode = p.X
+	}
+
+	var typ dwarf.Type
+
+	if snode, ok := fnnode.(*ast.StarExpr); ok {
+		// Pointer types only appear in the dwarf informations when a pointer to the type is used in the target program, here we create a pointer type on the fly so that the user can specify a pointer to any variable used in the target program
```
derekparker: Break this comment into multiple lines.

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+		return nil, err
+	}
+	argv.loadValue()
+	if argv.Unreadable != nil {
+		return nil, argv.Unreadable
+	}
+
+	fnnode := node.Fun
+
+	// remove all enclosing parenthesis form the type name
```
derekparker: s/form/from

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+	}
+}
+
+func exprToString(t ast.Expr) string {
+	var buf bytes.Buffer
+	printer.Fprint(&buf, token.NewFileSet(), t)
+	return string(buf.Bytes())
+}
+
+// Eval expressions: complex64(<float const>, <float const>) and complexq128(<float const>, <float const>)
```
derekparker: Typo: should just be `complex128`.

derekparker:
Just a few nits, otherwise LGTM.

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+		typ, err = scope.findType(exprToString(fnnode))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// only supports cast of integer constants into pointers
+	ptyp, isptrtyp := typ.(*dwarf.PtrType)
+	if !isptrtyp {
+		return nil, fmt.Errorf("can not convert \"%s\" to %s\n", exprToString(node.Args[0]), typ.String())
```
derekparker: Do not need newline at end of format string.

proc/eval.go
```diff
@@ -0,0 +1,730 @@
...
+	}
+
+	switch argv.Kind {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		// ok
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		// ok
+	default:
+		return nil, fmt.Errorf("can not convert \"%s\" to %s\n", exprToString(node.Args[0]), typ.String())
```
derekparker: Do not need newline at end of format string.

aarzilli:
done

lukehoban:
This is great.  

A few examples that don't seem to be working yet:

```js
// Type conversion (at least from uint32 to int)
(dlv) p a
5
(dlv) p int(a)
Command failed: can not convert "a" to int

// Left and right shifts
(dlv) p 8 >> 1
Command failed: invalid binary operation 8 >> 1

// Package constants
(dlv) p http.StatusInternalServerError
Command failed: could not find symbol value for http

// Calling functions seems to frequently crash instead of error
(dlv) p fmt.Errorf("Hello")
[crashes dlv]
```

aarzilli:
Type casts are documented as not yet supported, I'll look into package constants but I wonder if they exist at all in the executable.

Good catch with functions and shifts, I'll fix them.

aarzilli:
Added shift and fixed function calls, can't find constants if someone knows where they are I'll add them.

aarzilli:
The typechecking for shifts was wrong.

PR #286, opened 2015-11-06
Support for channels, maps and unsafe.Pointers

aarzilli:
Added support for some of the leftover types to proc/variables and proc/eval.

Chan and unsafe.Pointer support is mostly cosmetic, with a few extra rules for comparisons of chans in eval. 

Support for maps is, of course, a lot more involved. My iterator code is derived from `go/src/runtime/hashmap.go`, I use the same code for loading maps that I use to access a map key, which means map access is currently O(n) instead of amortized constant, but this way I did not have to implement the hashing functions.

For maps I only return maxArrayValues at a time, like we already do for slices, arrays and strings. Since there would be no way to get anything but the first 64 pairs I have allowed the slicing syntax to also work on maps:
```
someMap[128:]
```
will return 64 key/value pairs from someMap skipping the first 128 pairs. This looks somewhat unnatural but it's at least consistent with what we do with slices, strings and arrays.
By the way, the iteration order over maps (and therefore the order in which keys will appear when printed) is deterministic in my implementation, I did not copy the randomization that's done by go's map iterators.

### DIFF
--- a/_fixtures/testvariables3.go
+++ b/_fixtures/testvariables3.go
@@ -43,11 +43,13 @@ func main() {
 	var fn2 functype = nil
 	var nilslice []int = nil
 	var nilptr *int = nil
+	ch1 := make(chan int, 2)
+	var chnil chan int = nil
 
 	var amb1 = 1
 	runtime.Breakpoint()
 	for amb1 := 0; amb1 < 10; amb1++ {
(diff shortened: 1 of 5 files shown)


derekparker:
Merged this, works great but filed this bug: https://github.com/derekparker/delve/issues/288

PR #290, opened 2015-11-07
proc/eval: Support type casts between basic types

aarzilli:


proc/eval.go
```diff
@@ -177,33 +178,104 @@ func (scope *EvalScope) evalTypeCast(node *ast.CallExpr) (*Variable, error) {
...
+		typ = resolveTypedef(styp)
 	}
 
-	// only supports cast of integer constants into pointers
-	ptyp, isptrtyp := typ.(*dwarf.PtrType)
-	if !isptrtyp {
-		return nil, fmt.Errorf("can not convert \"%s\" to %s", exprToString(node.Args[0]), typ.String())
+	converr := fmt.Errorf("can not convert \"%s\" to %s", exprToString(node.Args[0]), typ.String())
+
+	fixargv := func(val constant.Value) *Variable {
```
derekparker: Unless I'm missing `stype` or `typ` getting mutated somewhere else, why not just inline those 2 assignments here and use `argv.Value = val` below instead of calling this function?

aarzilli: I realized that I wasn't patching `argv.Kind` either, I rewrote it to return a new Variable.

aarzilli:
Rebased on master at ad5097a8dc716e61992d5c7dd6213e58cb0a88b5

aarzilli:
This commit was also on the interfaces pull request, closing.

PR #291, opened 2015-11-12
Support for interface types and some builtin functions

aarzilli:
Added len, cap, imag and real (complex was already there, I refactored the code).

For interfaces I return a `Variable` with the interface type in `DwarfType` and a single element in `Children` containing the concrete value of the interface variable with its correct type.

To access member fields of the concrete value of an interface variable a type assertion can be used:
```
ifaceVar.(*concreteStructType).someField
```

### DIFF
--- a/proc/eval.go
+++ b/proc/eval.go
@@ -5,6 +5,7 @@ import (
 	"debug/dwarf"
 	"encoding/binary"
 	"fmt"
+	"github.com/derekparker/delve/dwarf/reader"
 	"go/ast"
 	"go/constant"
 	"go/parser"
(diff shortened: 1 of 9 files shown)


proc/eval.go
```diff
@@ -30,12 +32,17 @@ func (scope *EvalScope) EvalExpression(expr string) (*Variable, error) {
...
-		if fnnode, ok := node.Fun.(*ast.Ident); ok && len(node.Args) == 2 && (fnnode.Name == "complex64" || fnnode.Name == "complex128") {
-			// implement the special case type casts complex64(f1, f2) and complex128(f1, f2)
-			return scope.evalComplexCast(fnnode.Name, node)
+		if len(node.Args) == 1 {
+			v, err := scope.evalTypeCast(node)
+			if err == nil {
+				return v, nil
+			}
+			if err != reader.TypeNotFoundErr {
+				fmt.Printf("returning error %v for %s\n", err, exprToString(node))
```
derekparker: Debug statement leftover?

aarzilli: yes.

aarzilli:
Added two bugfixes related to this changeset:

1. Casting to map or channel: `(map[string]int)(0x40fff7)` was not rejected but would cause a crash later on
2. If the concrete type of an interface was inside a package with a long path (`github.com/someuser/somepackage.TheType`) delve was unable to find it because the type stored inside the interface object only contains the package name (`somepackage.TheType`)

aarzilli:
Rebased on master at ad5097a8dc716e61992d5c7dd6213e58cb0a88b5

proc/eval.go
```diff
@@ -95,24 +104,242 @@ func exprToString(t ast.Expr) string {
...
+		fnnode = p.X
+	}
+
+	styp, err := scope.Thread.dbp.findTypeExpr(fnnode)
 	if err != nil {
 		return nil, err
 	}
+	typ := resolveTypedef(styp)
+
+	converr := fmt.Errorf("can not convert \"%s\" to %s", exprToString(node.Args[0]), typ.String())
```
derekparker: Let's use `%q` here instead of escaping strings.

derekparker:
LGTM - I'm just going to make a commit for my nit and then merge.

PR #297, opened 2015-11-20
Dockerfile for running make

mishak87:
Usage:
```
docker run -v $(pwd):/go/src/github.com/derekparker/delve $(docker build -f Dockerfile.make . | tail -n 1 | cut -d' ' -f3) make build
```

### DIFF
--- a/Dockerfile.make
+++ b/Dockerfile.make
@@ -1,0 +1,9 @@
+FROM golang:1.5
+
+RUN go get github.com/tools/godep
+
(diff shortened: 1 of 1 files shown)


derekparker:
Thanks, but I'm going to close this out. I don't think the build process is sufficiently complicated, and we explicitly need to build cross-platform so this would really only Linux users.

mishak87:
@derekparker Cross compilation can be done in same image you just need to add all targets to makefile explicitly.

Issue #296, opened 2015-11-19
Specifying <packagename>.<receiver>.<function> fails while line number works


jmcauley:
I am able to specify a break point via filename:linenumber just fine. When I do so, the break command returns the package.receiver.function specification that should correspond to that line number. However, when I try to do: `break package.receiver.function` using the values returned by the first break call, it is not able to find it.
```
(dlv) break github.com/treetopllc/gonoble/graph/submission/dispatcher_service.go:183
Breakpoint 2 set at 0xb2c8f0 for github.com/treetopllc/gonoble/graph/submission.DispatcherService.update ./graph/submission/dispatcher_service.go:183
(dlv) break submission.DispatcherService.update
Command failed: Location "submission.DispatcherService.update" not found
```

My guess is it is because the receiver.function spec (DispatcherService.update in this case) exists in multiple packages.

aarzilli:
It's because it doesn't support partial package names, and the full package name is: `github.com/treetopllc/gonoble/graph/submission`.

aarzilli:
Closed by b5cf1572f812031801f5d77940aa149c8c840c6a

PR #307, opened 2015-12-12
service/debugger: support function spec. with partial package paths

aarzilli:
`packagename.SomeFunction` should match `github.com/someuser/packagename.SomeFunction` since the former is the familiar syntax.
To disambiguate between `io.SomeFunction` and `github.com/someuser/somepackage/io.SomeFunction` specify one extra slash at the start of the location specifier: `/io.SomeFunction`.

Fixes Issue #296


derekparker:
LGTM

PR #311, opened 2015-12-13
build: Add git sha to version output

derekparker:


derekparker:
Ping @aarzilli for review.

aarzilli:
Good idea, I was thinking of doing something similar (but with go generate) recently.

derekparker:
@aarzilli agree with your comment, pushed new change.

aarzilli:
Merged with 0abc772023bcde7f84a1094787e608220f9d8c0f

PR #313, opened 2015-12-15
terminal/command: bugfix: wrong check in sourceCommand

### LINKED ISSUE #312: source restricted to single-character filenames
In *terminal/command.go*, the test at the top of `sourceCommand()` seems overly restrictive:

```
805 func (cmds *Commands) sourceCommand(t *Term, args string) error {
806         if len(args) != 1 {
807                 return fmt.Errorf("wrong number of arguments: source <filename>")
808         }
809 
810         return cmds.executeFile(t, args)
811 }
```

(Single-character filenames work fine, but anything else returns an error.) I'm guessing the test is either left over from a time when `args` was declared as `[]string`, or foreshadows your intent to make it so.

aarzilli:
Fixes #312

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -803,7 +803,7 @@ func listCommand(t *Term, args string) error {
 }
 
 func (cmds *Commands) sourceCommand(t *Term, args string) error {
-	if len(args) != 1 {
+	if len(args) == 0 {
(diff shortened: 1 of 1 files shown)


derekparker:
LGTM

PR #314, opened 2015-12-15
travis: allow failures on tip

derekparker:
Make Go tip an "allowed failure" in travis.

@aarzilli there is a failing test on master that only fails on tip: `TestEvalExpression`. I think in general however, we should stay on top of changes to `tip` of Go, but let's not mark a build as failed if it doesn't pass against tip.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,7 @@ os:
 go:
   - tip
   - 1.5
+
+matrix:
+  allow_failures:
+    - go: tip


derekparker:
Not sure if this is actually the correct way to express this, might have to include OS or something, doesn't seem like travis is doing the right thing.

derekparker:
Fixed it.

aarzilli:
Agreed.

PR #315, opened 2015-12-16
service/locations: hooked expression evaluator to location specifiers

aarzilli:
Location specifiers starting with '*' can be followed by any
expression supported by the evaluator.
The expression should evaluate to either an integer (which will be
interpreted as an address) or to a function pointer (which will be
dereferenced to get the function's entry point).

### DIFF
--- a/_fixtures/locationsprog2.go
+++ b/_fixtures/locationsprog2.go
@@ -1,0 +1,29 @@
+package main
+
+import (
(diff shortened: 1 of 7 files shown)


aarzilli:
Issue with OSX fixed.

derekparker:
LGTM

PR #322, opened 2015-12-18
proc/variables: bugfix: makes TestEvalExpression more robust

### LINKED ISSUE #320: Test failure due to `map[string]int` not found?
When I pull in the latest from master into the Windows PR, I'm seeing a test failure in `TestEvalExpression` on this case:

```
{"(map[string]int)(0x4000)", false, "", "", "", fmt.Errorf("can not convert \"0x4000\" to *struct hash<string,int>")},
```

Instead of the expected error, I'm seeing "function calls are not supported".

This appears to be because the type `map[string]int` is not found in the call to `SeekToTypeNamed`.  Browsing through the types that are available in the DWARF metadata, I see `map[string]int64` but not `map[string]int`.  

I can't think of any reason this would be Windows-specific, but I'm not sure why this failure is not showing up on the Linux/OSX test runs. 

Note that I'm on Go 1.5.1 on `go version go1.5.1 windows/amd64`.  Perhaps this is an issue with 1.5.1?

aarzilli:
Fixes #320


derekparker:
LGTM

PR #323, opened 2015-12-19
Go 1.6beta1 compatibility

aarzilli:
- Updated TestEvalExpression to work with the new behaviour of `go/constant` in go1.6beta1
- Reverted `runtime.buildVersion` change :(


derekparker:
LGTM

PR #331, opened 2015-12-28
[WIP] Function call

derekparker:
# WIP - DO NOT MERGE (YET)!!!

Still a work in progress. One of the many wip commits on this branch actually passes all the tests, except it's a bit contrived. A lot of the work is done (for the most part), including creating a correct dummy stack frame. The code needs to be cleaned up a lot and we need to account for a few things, please ping me if you have interest and maybe want to hack on this. I will be trying to finish this PR soon.

### DIFF
--- a/_fixtures/testfunctioncall.go
+++ b/_fixtures/testfunctioncall.go
@@ -1,6 +1,9 @@
 package main
 
-import "runtime"
+import (
+	"fmt"
+	"runtime"
+)
 
 func double(i int) int {
 	return i * 2
@@ -8,4 +11,5 @@ func double(i int) int {
 
 func main() {
 	runtime.Breakpoint()
+	fmt.Println(double(4))
 }
--- a/_fixtures/testfunctioncall.go
+++ b/_fixtures/testfunctioncall.go
@@ -6,13 +6,14 @@ import (
 )
 
 func double(i int) int {
-	fmt.Println("i is:", i)
-	i = i * 2
-	fmt.Println("i is now:", i)
-	return i
+	return i * 2
+}
+
+func mult(a, b int) int {
+	return a * b
 }
 
 func main() {
 	runtime.Breakpoint()
-	fmt.Println(double(4))
+	fmt.Println(double(4), mult(2, 4))
 }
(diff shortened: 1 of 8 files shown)


aarzilli:
I did think a bit about calling functions (without writing any code so this is purely philosophical) this is what I thought would be the pain points:

a. panics
b. breakpoints getting hit while evaluating a function
c. our goroutine getting scheduled out -- which makes (b) worse.

Regarding (a), one solution is to set a breakpoint on panic and abort the evaluation as soon as it is called, before it starts unwinding the stack. This is not ideal, the panic could be recovered by a frame below our dummy frame, but given how panics are generally used by go code it should be a big deal.

The other possibility would be to abort the call only when the stack unwind gets to our dummy frame, possibly modifying the `g` struct to slip in our own fake `_defer` object?

Regarding (b) and (c) I have two possible solution

# Orthodox solution

For (b) we disable all user breakpoints. (c) is hard: I'm not sure it's possible to suppress co-operative scheduler calls, maybe abort on them? Even pre-emptive scheduler calls may be hard because they seem to be at least in part intertwined with stack resizes (which we can not suppress).

# Unorthodox solution

We call `Continue` in the middle of `Call`, any breakpoint that gets hit is reported to the user normally, the rest of `Call` is executed when its temp breakpoint (which would have a condition on the goroutine id) gets hit, in a callback executed by `Continue`.

This is easier to implement because we don't have to worry about (b) and especially (c) but it's weird for the user. 

```
> main.somefunction() somefile.go:10
(dlv) print a
3
(dlv) print double(a)
Command failed: function calls are not supported
```

The user must use a different command to enable function calls, this command will have a behaviour similar to `continue` and use a similar API, let's call this command `exec`:

```
(dlv) exec double(a)
> main.someotherfunction() somefile.go:50
```

the function call did not complete, the user got a breakpoint somewhere else in the program, the user can interact with the debugger:

```
(dlv) continue
> main.somefunction() somefile.go:10
    double(a) = 6
``` 

after another continue the original `exec` call completes and the user gets the evaluation result: `double(a) = 6`.

derekparker:
This PR is extremely outdated so I will close. @aarzilli might be worth moving your comment to the issue instead, and we can always potentially reference this code for a starting point, but most of it is pretty ugly and hacky anyways.

PR #335, opened 2015-12-30
 proc/variables: prefetch of target process memory 

aarzilli:
Prefetch the entire memory of structs and arrays and cache it instead
of issuing readMemory calls only when we get down to primitive types.
This reduces the number of system calls to ptrace that variables makes.

Improves performance in general, greatly improving it in some
particular cases (involving large structs).

```
Benchmarks without prefetching:
	BenchmarkArray-4         	      10	 132189944 ns/op	   0.06 MB/s
	BenchmarkArrayPointer-4  	       5	 202538503 ns/op	   0.04 MB/s
	BenchmarkMap-4           	     500	   3804336 ns/op	   0.27 MB/s
	BenchmarkGoroutinesInfo-4	      10	 126397104 ns/op
	BenchmarkLocalVariables-4	     500	   2494846 ns/op

Benchmarks with prefetching:
	BenchmarkArray-4         	     200	  10719087 ns/op	   0.76 MB/s
	BenchmarkArrayPointer-4  	     100	  11931326 ns/op	   0.73 MB/s
	BenchmarkMap-4           	    1000	   1466479 ns/op	   0.70 MB/s
	BenchmarkGoroutinesInfo-4	      10	 103407004 ns/op
	BenchmarkLocalVariables-4	    1000	   1530395 ns/op

Improvement factors:
	BenchmarkArray				12.33x
	BenchmarkArrayPointer		16.97x
	BenchmarkMap				 2.59x
	BenchmarkGoroutinesInfo		 1.22x
	BenchmarkLocalVariables		 1.63x
```

### DIFF
--- a/proc/mem.go
+++ b/proc/mem.go
@@ -1,0 +1,54 @@
+package proc
+
+const cacheEnabled = true
+
+type memoryReadWriter interface {
+	readMemory(addr uintptr, size int) (data []byte, err error)
+	writeMemory(addr uintptr, data []byte) (written int, err error)
+}
+
+type memCache struct {
+	cacheAddr uintptr
+	cache     []byte
+	mem       memoryReadWriter
+}
+
+func (m *memCache) contains(addr uintptr, size int) bool {
+	return addr >= m.cacheAddr && (addr+uintptr(size)) <= (m.cacheAddr+uintptr(len(m.cache)))
+}
+
+func (m *memCache) readMemory(addr uintptr, size int) (data []byte, err error) {
+	if m.contains(addr, size) {
+		d := make([]byte, size)
+		copy(d, m.cache[addr-m.cacheAddr:])
+		return d, nil
+	}
+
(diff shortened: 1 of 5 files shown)


derekparker:
This is great, I'm very excited about this patch. Is there any reason this is on top of #315? That PR has a legitimate failure on OSX, which is causing CI to fail on this PR.

I will review the code tomorrow morning.

aarzilli:
> This is great, I'm very excited about this patch. Is there any reason this is on top of #315? 

No, it's a mistake. I also didn't notice the issue with #315

aarzilli:
Decoupled from #315

proc/mem.go
```diff
@@ -0,0 +1,54 @@
+package proc
+
+const cacheEnabled = true
+
+type memory interface {
```
derekparker: nit: Rename to `memoryReadWriter`.

derekparker:
One small nit, otherwise LGTM.

PR #338, opened 2016-01-03
bugfix: proc/types: loadPackageMap and github.com packages

aarzilli:
Makes loadPackageMap actually work with github.com/... packages

This fixes my 38716dcc26d445ae3f6b081968266e33a8d99f58 that was too conservative.

### DIFF
--- a/proc/types.go
+++ b/proc/types.go
@@ -38,16 +38,10 @@ func (dbp *Process) findTypeExpr(expr ast.Expr) (dwarf.Type, error) {
 }
 
 func complexType(typename string) bool {
-	dot := 0
 	for _, ch := range typename {
 		switch ch {
 		case '*', '[', '<', '{', '(', ' ':
 			return true
-		case '.':
-			dot++
-			if dot > 1 {
-				return true
-			}
 		}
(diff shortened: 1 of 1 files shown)


derekparker:
Seems like this would be difficult to test without adding a dep to the test file (which is a definite 'no'), but can you run through steps to reproduce an error? Just want to verify this fix before merging. Otherwise the code LGTM.

aarzilli:
>Seems like this would be difficult to test without adding a dep to the test file

exactly

>can you list steps to reproduce an error?

1. Import any github.com package
2. Create an interface variable with a concrete type inside the github.com package:
```
import "github.com/.../somepackage"
...
var ivar interface{} = &somepackage.SomeType{}
```
3. Try to evaluate `ivar`

I've been thinking about replacing this code with one that parses the source code for the current file to get the package name → package path mappings. Pro: it would work for code that imports packages with a different name than what would be assigned automatically; Contra: it wouldn't work at all if source isn't available, it would be much slower and therefore require caching to be usable.

derekparker:
LGTM - merged.

PR #339, opened 2016-01-04
workaround: proc/eval: go sometimes inserts &v instead of v

aarzilli:
For example: in testvariables3.go the `main` function instead of having an entry for `i1` has an entry for `&i1`. I guess this could be considered a go bug and I think at some point fixing it was discussed in the issue tracker, however it doesn't seem like it's getting fixed in 1.6 and it happens frequently in ordinary code.

### DIFF
--- a/proc/eval.go
+++ b/proc/eval.go
@@ -432,16 +432,23 @@ func (scope *EvalScope) evalIdent(node *ast.Ident) (*Variable, error) {
 	// try to interpret this as a local variable
 	v, err := scope.extractVarInfo(node.Name)
 	if err != nil {
-		// if it's not a local variable then it could be a package variable w/o explicit package name
 		origErr := err
-		_, _, fn := scope.Thread.dbp.PCToLine(scope.PC)
-		if fn != nil {
-			if v, err := scope.packageVarAddr(fn.PackageName() + "." + node.Name); err == nil {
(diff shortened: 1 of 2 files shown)


proc/eval.go
```diff
@@ -432,16 +432,23 @@ func (scope *EvalScope) evalIdent(node *ast.Ident) (*Variable, error) {
...
 	v, err := scope.extractVarInfo(node.Name)
 	if err != nil {
-		// if it's not a local variable then it could be a package variable w/o explicit package name
 		origErr := err
-		_, _, fn := scope.Thread.dbp.PCToLine(scope.PC)
-		if fn != nil {
-			if v, err := scope.packageVarAddr(fn.PackageName() + "." + node.Name); err == nil {
-				v.Name = node.Name
-				return v, nil
+		// workaround: sometimes go inserts an entry for '&varname' instead of varname
```
derekparker: Is there an issue in the Go tracker you can point to here? If the issue gets fixed it would be nice to have a link in this comment so we can discuss removing the workaround once this gets fixed.

derekparker:
Small documentation nit, otherwise LGTM.

aarzilli:
I remember a discussion about this in gerrit but I can't find it anymore.

PR #345, opened 2016-01-10
proc: bugfix: clearing temp breakpoints

aarzilli:
Temp breakpoints should be cleared even if a non-temp breakpoint is
triggered on the same goroutine that the temp breakpoints are set on.

### DIFF
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1315,6 +1315,25 @@ func TestIssue262(t *testing.T) {
 	})
 }
 
+func TestIssue305(t *testing.T) {
+	// If 'next' hits a breakpoint on the goroutine it's stepping through the temp breakpoints aren't cleared
+	// preventing further use of 'next' command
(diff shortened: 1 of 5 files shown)


proc/proc_test.go
```diff
@@ -1315,6 +1315,24 @@ func TestIssue262(t *testing.T) {
 	})
 }
 
+func TestIssue305(t *testing.T) {
+	// If 'next' hits a breakpoint on the goroutine it's stepping through the temp breakpoints aren't cleared
+	// preventing further use of 'next' command
+	withTestProcess("issue305", t, func(p *Process, fixture protest.Fixture) {
+		addr, _, err := p.goSymTable.LineToPC(fixture.Source, 5)
```
derekparker: This error gets shadowed, let's check it and fail the test if it's not nil.

aarzilli:
check added

PR #346, opened 2016-01-12
Conditional Breakpoints (backend)

aarzilli:
Changed `Breakpoint.Cond` to accept any expression that can be evaluated instead of just an integer.
Added fake variable `runtime.curg` to `eval.go` so that temp breakpoints used by `Next` can be expressed as a comparision on `runtime.curg.goid`.

### DIFF
--- a/proc/breakpoints.go
+++ b/proc/breakpoints.go
@@ -1,6 +1,12 @@
 package proc
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+	"go/ast"
+	"go/constant"
+	"reflect"
+)
 
 // Breakpoint represents a breakpoint. Stores information on the break
 // point including the byte of data that originally was stored at that
@@ -24,7 +30,7 @@ type Breakpoint struct {
 	HitCount      map[int]uint64 // Number of times a breakpoint has been reached in a certain goroutine
 	TotalHitCount uint64         // Number of times a breakpoint has been reached
 
-	Cond int // When Cond is greater than zero this breakpoint will trigger only when the current goroutine id is equal to it
+	Cond ast.Expr // When Cond is not nil the breakpoint will be triggered only if evaluating Cond returns true
 }
 
 func (bp *Breakpoint) String() string {
@@ -78,7 +84,7 @@ func (dbp *Process) setBreakpoint(tid int, addr uint64, temp bool) (*Breakpoint,
 		Line:         l,
 		Addr:         addr,
 		Temp:         temp,
-		Cond:         -1,
+		Cond:         nil,
 		HitCount:     map[int]uint64{},
 	}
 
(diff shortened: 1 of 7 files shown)


proc/breakpoints.go
```diff
@@ -110,14 +115,18 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
...
 func (bp *Breakpoint) checkCondition(thread *Thread) bool {
-	if bp.Cond < 0 {
+	if bp.Cond == nil {
 		return true
 	}
-	g, err := thread.GetG()
+	scope, err := thread.Scope()
 	if err != nil {
-		return false
+		return true
```
derekparker: I don't think we really want to return `true` here. I understand that it would stop the debugger upon an error (essentially) but it could likely be a false positive, and yet there's no indication of that. We should either log/print a warning or modify the method signature to return an error and let the caller deal with it.

proc/breakpoints.go
```diff
@@ -110,14 +115,18 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
...
+	}
+	v, err := scope.evalAST(bp.Cond)
+	if err != nil || v.Kind != reflect.Bool || v.Unreadable != nil {
+		return true
```
derekparker: Same comment as above.

aarzilli:
The problem is that an error there could be caused by the user entering an expression that isn't boolean or refers to a non existing variable. If we propagate the error the easy way `proc.(*Process).setCurrentBreakpoints` exits early, potentially without finding some breakpoints, which would create more serious problems later.

I have added a `BreakpointConditionError` field to `Thread`, I store errors created by `checkCondition` there, then turn them into an error before exiting `Continue`. I believe this way it shouldn't create problems.

proc/breakpoints.go
```diff
@@ -109,15 +114,25 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
 	return err
 }
 
-func (bp *Breakpoint) checkCondition(thread *Thread) bool {
-	if bp.Cond < 0 {
-		return true
+func (bp *Breakpoint) checkCondition(thread *Thread) (met bool, err error) {
```
derekparker: nit: named returns aren't actually being used, so just remove them.

proc/breakpoints.go
```diff
@@ -109,15 +114,25 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
...
+	if bp.Cond == nil {
+		return true, nil
 	}
-	g, err := thread.GetG()
+	scope, err := thread.Scope()
 	if err != nil {
-		return false
+		return true, nil
```
derekparker: Why not return the error here?

proc/breakpoints.go
```diff
@@ -109,15 +114,25 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
...
-	return g.ID == bp.Cond
+	v, err := scope.evalAST(bp.Cond)
+	if err != nil {
+		return true, fmt.Errorf("error evaluating expression: %v", err)
+	}
+	if v.Unreadable != nil {
+		return true, fmt.Errorf("condition expression unreadable: %v", v.Unreadable)
+	}
+	if v.Kind != reflect.Bool {
+		return true, fmt.Errorf("condition expression not boolean")
```
derekparker: nit: `errors.New` for non formatted errors.

proc/breakpoints.go
aarzilli: oversight

aarzilli:
updated

derekparker:
@aarzilli needs a rebase, otherwise LGTM.

aarzilli:
rebased

Issue #240, opened 2015-09-20
Refactor: scope prefixed commands should not be hard coded


derekparker:
Currently, scope prefixed commands are hard coded. In other words, only certain commands can be scope prefixed (i.e. used in `goroutine <id> <cmd>` or `frame <n> <cmd>`. Also, the more pressing issue is that user supplied aliases for those commands will not work.

Scope prefix commands should respect user aliases, and should if possible not be hard coded.

PR #349, opened 2016-01-19
Breakpoint Names and conditions

aarzilli:
I've implemented the mailing list proposal.

This PR builds on top of PR #346 of course.

I've also refactored `terminal/command.go` so that every time we print a location or a breakpoint two new functions, `formatLocation` and `formatBreakpoint`, are used instead of adhoc printf format strings.

This PR introduces a second switch statement for subcommands, similar to the one in `scopePrefix`, maybe this would be a good time to do issue #240 before this gets out of hand.

### DIFF
--- a/service/client.go
+++ b/service/client.go
@@ -36,12 +36,19 @@ type Client interface {
 
 	// GetBreakpoint gets a breakpoint by ID.
 	GetBreakpoint(id int) (*api.Breakpoint, error)
+	// GetBreakpointByName gets a breakpoint by name.
+	GetBreakpointByName(name string) (*api.Breakpoint, error)
 	// CreateBreakpoint creates a new breakpoint.
 	CreateBreakpoint(*api.Breakpoint) (*api.Breakpoint, error)
 	// ListBreakpoints gets all breakpoints.
 	ListBreakpoints() ([]*api.Breakpoint, error)
 	// ClearBreakpoint deletes a breakpoint by ID.
 	ClearBreakpoint(id int) (*api.Breakpoint, error)
+	// ClearBreakpointByName deletes a breakpoint by name
+	ClearBreakpointByName(name string) (*api.Breakpoint, error)
+	// Allows user to update an existing breakpoint for example to change the information
+	// retrieved when the breakpoint is hit or to change, add or remove the break condition
+	AmendBreakpoint(*api.Breakpoint) error
 
 	// ListThreads lists all threads.
 	ListThreads() ([]*api.Thread, error)
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -4,6 +4,7 @@ import (
 	"debug/gosym"
 	"errors"
 	"fmt"
+	"go/parser"
 	"log"
 	"path/filepath"
 	"regexp"
@@ -100,11 +101,12 @@ func (d *Debugger) Restart() error {
 	if err != nil {
 		return fmt.Errorf("could not launch process: %s", err)
 	}
-	for addr, bp := range d.process.Breakpoints {
-		if bp.Temp {
-			continue
+	for _, oldBp := range d.Breakpoints() {
+		newBp, err := p.SetBreakpoint(oldBp.Addr)
+		if err != nil {
+			return err
 		}
-		if _, err := p.SetBreakpoint(addr); err != nil {
+		if err := copyBreakpointInfo(newBp, oldBp); err != nil {
 			return err
 		}
 	}
@@ -150,6 +152,16 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 		addr      uint64
 		err       error
 	)
+
+	if requestedBp.Name != "" {
+		if err = api.ValidBreakpointName(requestedBp.Name); err != nil {
+			return nil, err
+		}
+		if d.FindBreakpointByName(requestedBp.Name) != nil {
+			return nil, errors.New("breakpoint name already exists")
+		}
+	}
+
 	switch {
 	case len(requestedBp.File) > 0:
 		fileName := requestedBp.File
@@ -182,14 +194,39 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	if err != nil {
 		return nil, err
 	}
-	bp.Tracepoint = requestedBp.Tracepoint
-	bp.Goroutine = requestedBp.Goroutine
-	bp.Stacktrace = requestedBp.Stacktrace
-	bp.Variables = requestedBp.Variables
-	bp.Cond = nil
+	if err := copyBreakpointInfo(bp, requestedBp); err != nil {
+		if _, err1 := d.process.ClearBreakpoint(bp.Addr); err1 != nil {
+			err = fmt.Errorf("error while creating breakpoint: %v, additionally the breakpoint could not be properly rolled back: %v", err, err1)
+		}
+		return nil, err
+	}
 	createdBp = api.ConvertBreakpoint(bp)
 	log.Printf("created breakpoint: %#v", createdBp)
 	return createdBp, nil
+}
+
+func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
+	original := d.findBreakpoint(amend.ID)
+	if original == nil {
+		return fmt.Errorf("no breakpoint with ID %d", amend.ID)
+	}
+	if err := api.ValidBreakpointName(amend.Name); err != nil {
+		return err
+	}
+	return copyBreakpointInfo(original, amend)
+}
+
+func copyBreakpointInfo(bp *proc.Breakpoint, requested *api.Breakpoint) (err error) {
+	bp.Name = requested.Name
+	bp.Tracepoint = requested.Tracepoint
+	bp.Goroutine = requested.Goroutine
+	bp.Stacktrace = requested.Stacktrace
+	bp.Variables = requested.Variables
+	bp.Cond = nil
+	if requested.Cond != "" {
+		bp.Cond, err = parser.ParseExpr(requested.Cond)
+	}
+	return err
 }
 
 // ClearBreakpoint clears a breakpoint.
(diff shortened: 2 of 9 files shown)


aarzilli:
rebased and updated.

terminal/command.go
```diff
@@ -83,6 +84,8 @@ func DebugCommands(client service.Client) *Commands {
 		{aliases: []string{"stack", "bt"}, cmdFn: stackCommand, helpMsg: "stack [<depth>] [-full]. Prints stack."},
 		{aliases: []string{"frame"}, cmdFn: frame, helpMsg: "Sets current stack frame (0 is the top of the stack)"},
 		{aliases: []string{"source"}, cmdFn: c.sourceCommand, helpMsg: "Executes a file containing a list of delve commands"},
+		{aliases: []string{"on"}, cmdFn: onCmd, helpMsg: "on <breakpoint name or id> <command>. Executes command when the specified breakpoint is hit (supported commands: print, stack and goroutine without arguments)"},
+		{aliases: []string{ "condition", "cond"}, cmdFn: conditionCmd, helpMsg: "on <breakpoint name or id> <boolean expression>. Specifies that the breakpoint or tracepoint should break only if the boolean expression is true."},
```
derekparker: Cleanup whitespace after slice bracket. Also, helpMsg says `on` instead of `cond`.

terminal/command.go
```diff
@@ -83,6 +84,8 @@ func DebugCommands(client service.Client) *Commands {
 		{aliases: []string{"stack", "bt"}, cmdFn: stackCommand, helpMsg: "stack [<depth>] [-full]. Prints stack."},
 		{aliases: []string{"frame"}, cmdFn: frame, helpMsg: "Sets current stack frame (0 is the top of the stack)"},
 		{aliases: []string{"source"}, cmdFn: c.sourceCommand, helpMsg: "Executes a file containing a list of delve commands"},
+		{aliases: []string{"on"}, cmdFn: onCmd, helpMsg: "on <breakpoint name or id> <command>. Executes command when the specified breakpoint is hit (supported commands: print, stack and goroutine without arguments)"},
```
derekparker: Let's make this more clear, something like `print OR stack / goroutine without arguments` to make it more clear that print will take an arg. Also, why limit the arg to something like stack? Seems reasonable to want to limit stack depth potentially, and since we will need to support an arg for print it doesn't seem out of the question.

service/debugger/debugger.go
```diff
@@ -168,16 +178,66 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
+			break
+		}
+	}
+	if original == nil {
+		return fmt.Errorf("no breakpoint with ID %d", amend.ID)
+	}
+	if err := d.validBreakpointName(amend.Name, false); err != nil {
+		return err
+	}
+	copyBreakpointInfo(original, amend)
```
derekparker: Need to check the error here, especially since a condition is added via amend and could error out.

service/debugger/debugger.go
```diff
@@ -168,16 +178,66 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
+	bp.Stacktrace = requested.Stacktrace
+	bp.Variables = requested.Variables
+	bp.Cond = nil
+	if requested.Cond != "" {
+		bp.Cond, err = parser.ParseExpr(requested.Cond)
+	}
+	return err
+}
+
+func (d *Debugger) validBreakpointName(name string, checkdup bool) error {
```
derekparker: Let's make this a pure function and move the duplicate name check outside of this func.

service/debugger/debugger.go
```diff
@@ -212,6 +272,16 @@ func (d *Debugger) FindBreakpoint(id int) *api.Breakpoint {
 	return nil
 }
 
+// FindBreakpointByName returns the breakpoint specified by 'name'
+func (d *Debugger) FindBreakpointByName(name string) *api.Breakpoint {
```
derekparker: nit: Move this next to the other `FindBreakpoint` method for organization.

terminal/command.go
```diff
@@ -558,7 +563,11 @@ func breakpoints(t *Term, args string) error {
 		if bp.Tracepoint {
 			thing = "Tracepoint"
 		}
-		fmt.Printf("%s %d at %#v %s:%d (%d)\n", thing, bp.ID, bp.Addr, ShortenFilePath(bp.File), bp.Line, bp.TotalHitCount)
+		name := bp.Name
+		if name == "" {
+			name = strconv.Itoa(bp.ID)
+		}
+		fmt.Printf("%s %s at %#v %s:%d (%d)\n", thing, name, bp.Addr, ShortenFilePath(bp.File), bp.Line, bp.TotalHitCount)
```
derekparker: Use the same `formatBreakpoint*` functions here as well.

terminal/command.go
```diff
@@ -578,37 +587,52 @@ func breakpoints(t *Term, args string) error {
 	return nil
 }
 
-func setBreakpoint(t *Term, tracepoint bool, argstr string) error {
-	args := strings.Split(argstr, " ")
-	if len(args) < 1 {
-		return fmt.Errorf("address required, specify either a function name or <file:line>")
+func validBreakpointName(name string) bool {
```
derekparker: We shouldn't have to duplicate this logic here, can't we just rely on this error being caught later on down the line?

terminal/command.go
derekparker: Forget the last part of my comment, but let's make the help text clearer as to what takes args and what does not if we can.

terminal/command.go
aarzilli: This check is to figure out where to split `break`'s arguments, the other is a true validation. I have moved the validation logic in `service/api`. If you want I can change `terminal.setBreakpoint` to first pass its whole argument string to the server and if there is an error do the split.

service/debugger/debugger.go
aarzilli: Which `FindBreakpoint`? Isn't it already next to it?

derekparker: Eh, it is indeed. Sorry.

service/debugger/debugger.go
```diff
@@ -182,16 +193,45 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
+		return nil, err
+	}
 	createdBp = api.ConvertBreakpoint(bp)
 	log.Printf("created breakpoint: %#v", createdBp)
 	return createdBp, nil
 }
 
+func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
+	var original *proc.Breakpoint
+	for i := range d.process.Breakpoints {
```
derekparker: Replace this logic with `original := d.FindBreakpoint(amend.ID)`.

service/client.go
```diff
@@ -36,12 +36,19 @@ type Client interface {
...
+	GetBreakpointByName(name string) (*api.Breakpoint, error)
 	// CreateBreakpoint creates a new breakpoint.
 	CreateBreakpoint(*api.Breakpoint) (*api.Breakpoint, error)
 	// ListBreakpoints gets all breakpoints.
 	ListBreakpoints() ([]*api.Breakpoint, error)
 	// ClearBreakpoint deletes a breakpoint by ID.
 	ClearBreakpoint(id int) (*api.Breakpoint, error)
+	// ClearBreakpointByName deletes a breakpoint by name
+	ClearBreakpointByName(name string) (*api.Breakpoint, error)
+	// Allows user to update an existing breakpoint for example to change the informations
```
derekparker: s/informations/information/

service/client.go
```diff
@@ -36,12 +36,19 @@ type Client interface {
 
 	// GetBreakpoint gets a breakpoint by ID.
 	GetBreakpoint(id int) (*api.Breakpoint, error)
+	// GetBreakpointByName gets a breakpoint by name
```
derekparker: nit: Add period to end of comments for proper punctuation.

service/debugger/debugger.go
```diff
@@ -182,16 +193,45 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	if err != nil {
 		return nil, err
 	}
-	bp.Tracepoint = requestedBp.Tracepoint
-	bp.Goroutine = requestedBp.Goroutine
-	bp.Stacktrace = requestedBp.Stacktrace
-	bp.Variables = requestedBp.Variables
-	bp.Cond = nil
+	if err := copyBreakpointInfo(bp, requestedBp); err != nil {
+		d.process.ClearBreakpoint(bp.Addr)
```
derekparker: Clearing the breakpoint could error out, we should surface that error but provide a meaningful message so the user isn't confused as to why a `ClearBreakpoint` error happened when they were trying to set a breakpoint. Either that, or try and parse the conditional expression earlier and surface an invalid condition error earlier, before the breakpoint is ever set.

service/debugger/debugger.go
aarzilli: `FindBreakpoint` returns an `api.Breakpoint` this needs a `proc.Breakpoint`. I have refactored the code of `FindBreakpoint` and `AmendBreakpoint` to both use a `findBreakpoint` that returns a `proc.Breakpoint`, this also works as a micro-optimizations for `FindBreakpoint`.

derekparker:
Need rebase.

aarzilli:
rebased.

derekparker:
When testing this out I noticed that if I set a condition on a breakpoint, and then used the `on` command to print a variable, the condition is erased. We should not erase an existing condition when amending a breakpoint.

aarzilli:
done

derekparker:
@aarzilli the problem still exists.

derekparker:
@aarzilli [deleted my last comment] but there is still an issue. Using the `restart` command does not preserve the breakpoint name, condition, or amendments.

service/debugger/debugger.go
```diff
@@ -182,16 +193,41 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
 		return nil, err
 	}
-	bp.Tracepoint = requestedBp.Tracepoint
-	bp.Goroutine = requestedBp.Goroutine
-	bp.Stacktrace = requestedBp.Stacktrace
-	bp.Variables = requestedBp.Variables
-	bp.Cond = nil
+	if err := copyBreakpointInfo(bp, requestedBp); err != nil {
+		if _, err1 := d.process.ClearBreakpoint(bp.Addr); err1 != nil {
+			err = fmt.Errorf("Error while creating breakpoint: %v, additionally the breakpoint could not be properly rolled back: %v", err, err1)
```
derekparker: nit: Lowercase first word in error string.

aarzilli:
None of the breakpoint decorations were preserved across restarts. Fixed it.

derekparker:
Merged.

PR #350, opened 2016-01-19
terminal/command: bugfix: goroutines doesn't work without arguments

aarzilli:
Splitting an empty string results in a one element array with a
single empty string element.

See: http://play.golang.org/p/EzVX64Q82A


derekparker:
Thanks, LGTM. We really need some e2e tests to make sure regressions like this don't happen.

aarzilli:
agreed.

PR #351, opened 2016-01-20
proc: Single step only "current" thread

derekparker:
Instead of the `step` command single stepping every thread, instead only
step the "current" thread. This fixes a few issues surrounding single
stepping, and simplifies the logic. The original concerns around only
stepping a single thread (with regard to coordination) are invalid and
generally non-issues.


aarzilli:
approved

PR #353, opened 2016-01-20
Refactor parseg var

derekparker:
Builds on top of https://github.com/derekparker/delve/pull/346, will need to be rebased.

### DIFF
--- a/proc/breakpoints.go
+++ b/proc/breakpoints.go
@@ -1,6 +1,11 @@
 package proc
 
-import "fmt"
+import (
+	"fmt"
+	"go/ast"
+	"go/constant"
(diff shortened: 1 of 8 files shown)


aarzilli:
This is loading and parsing a lot more than `parseG` needs but I ran the benchmarks and it's actually faster than the old `parseG` anyway (probably because we aren't getting type info directly from dwarf) so it doesn't matter.

Looks good.

derekparker:
@aarzilli it should just have parity with the original implementation. There are certainly fields we're parsing we don't need that could be removed.

aarzilli:
Merged.

PR #362, opened 2016-01-23
proc: replace debug/dwarf with golang.org/x/debug/dwarf

### LINKED ISSUE #356: Debugger returns only only first Item from slice
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2266
The following code snippet demonstrates the problem:
```go
package main

import "fmt"

type Item struct {
	Name   string
	Route  string
	Active int
}

type Menu []Item

func main() {
	mainMenu := Menu{
		{Name: "home", Route: "/", Active: 1},
		{Name: "About", Route: "/about", Active: 1},
		{Name: "Login", Route: "/login", Active: 1},
	}

	fmt.Println(mainMenu)
	
}
```
image
image

aarzilli:
Typedefs that resolve to slices are not recorded in DWARF as typedefs
but instead as structs in a way that there is no way to know they
are really slices using `debug/dwarf`.
Using `golang.org/x/debug/dwarf` instead this problem is solved and
as a bonus some types are printed with a nicer names: (`struct string`
→ `string`, `struct []int` → `[]int`, etc)

 Fixes #356


aarzilli:
Looks like `golang.org/x/debug/dwarf` does not work with go 1.6 beta 2.

aarzilli:
Actually it looks like delve has problems with 1.6 in general, it's just that no other test is triggering it, see: https://github.com/aarzilli/delve/tree/oops16

aarzilli:
The problem was that `dwarf/frame` incorrectly assumes that all dwarf sections are big endian. That's not the case, I copied the trick that `debug/dwarf` uses to determine endianness and used it for `dwarf/frame`.

I don't know why this only manifested with go 1.6b2, it could be that only <= 1.5 did not emit any endian-relative opcode in debug_frame or it could be that only 1.6 + the changes to testvariables3 created a frame large enough to coax the compiler to output a `DW_CFA_advance_loc2`.

derekparker:
@aarzilli now that the Windows PR has landed, can you update this branch with any Windows specific files that need to be updated as well.

lukehoban:
Unfortunately, it looks like there is no PE support in golang.org/x/debug yet.  Not clear whether that is a point in time issue or why PE and Plan9Obj support is missing from [golang.org/x/debug](https://godoc.org/golang.org/x/debug) (both were available in [debug](https://golang.org/pkg/debug/)).

 Perhaps in the meantime, something like https://github.com/golang/go/blob/master/src/debug/pe/file.go#L297 can be inlined into `proc_windows.go` to extract the `golang.org/x/debug/dwarf` data from the PE?

aarzilli:
Rebased and updated, we can get around `golang.org/x/debug` not having `pe` easily.

derekparker:
Merged.

Issue #363, opened 2016-01-24
Remaining Windows work


derekparker:
- [x] **TestDebugLinePrologueParser**
  Only on some  machines and Go versions.  Appears to be because of buffers in the debugline data which are larger than they should be in Windows binaries?  I suspect there is a bug in the Go compiler on Windows that we'll need to work around.
- [x] **TestNextConcurrent flakiness**
  Occasional occurrences of the following errors:
  `proc_test.go:355: Program did not continue to correct next location expected 11 was parallel_next.go:10`
  `proc_test.go:355: Program did not continue to correct next location expected 11 was parallel_next.go:8`
  `proc_test.go:355: Program did not continue to correct next location expected 10 was parallel_next.go:8`
  `proc_test.go:410: Program did not continue to correct next location expected 11 was parallel_next.go:10`
  `proc_test.go:403: Initial breakpoint triggered twice for the same goroutine`
- [x] **Attach is not supported**
  Should also add test coverage for attach on all platforms.
- [x] **TestNextNetHTTP is disabled on Windows**
  Temporarily disabled this test, need to investigate the failure.
- [ ] **Should we load symbols on Windows?**
  Thread information is not displayed nicely for most non-active threads because the PC on those threads is in ntdll or other dynamically loaded code. To address this, we would need to load symbols on Windows and merge that information with the DWARF symbols. This should be possible, but worth discussing before implementing.
- [x] **thread.halt() may need to actually halt threads in some cases**
  See comment in implementation.
- [x] **Can we remove canContinue?**
  dbp.Step currently relies on checking this, but would be nice to factor it to not need this concept in addition to blocked and stopped.
- [x] **Infrequent failures in TestBreakPointCount**
  Pretty infrequent occurence, and seemingly always with 101 + 99.  So it appears that one of the breakpoint hits is getting recorded as being on the wrong thread. 

Once merged, we will also want to do:
- [x] **Add Appveyor project tracking derekparker/delve and badge in README**
- [x] **Update docs with info on building/running on Windows**
  Mostly just MinGW requirement which is likely not present for many Windows Go users.

lukehoban:
> **Update docs with info on building/running on Windows.** Mostly just MinGW requirement which is likely not present for many Windows Go users.

This is updated in the wiki now at https://github.com/derekparker/delve/wiki/Building#windows.

uuf6429:
@lukehoban Regarding the wiki update, there are at least two issues:
1. This is not valid for Windows:
[...]

PR #364, opened 2016-01-24
proc:  reenable TestNextNetHTTP on Windows

lukehoban:
Explicitly use `localhost` in HTTP test as required by Windows.

Part of #363.


derekparker:
Merged, thanks.

PR #365, opened 2016-01-24
proc: step now goes to next line, including funcs

### LINKED ISSUE #360: Step into function
We should replace the default behavior of `step` to be step into function, and have another command for stepping a single CPU instruction.

derekparker:
This patch modifies the `step` command to step to the next source line,
stepping into any function encountered along the way.

Fixes #360

### DIFF
--- a/_fixtures/teststep.go
+++ b/_fixtures/teststep.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func callme() {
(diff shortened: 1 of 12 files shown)


derekparker:
Not mentioned in commit msg, but it also introduces the `step-instruction` command, which has parity with how `step` used to work (steps exactly 1 cpu instruction).

cc: @aarzilli

aarzilli:
LGTM

PR #375, opened 2016-01-26
[service/debugger] Case-insensitive paths on Windows

### LINKED ISSUE #370: Breakpoint Locations should be case insensitve on Windows
I'm looking to add support for Delve to the Zeus IDE and came across the following issue.

As a test I create a Windows test Go application here:

    C:\Go\src\github.com\jussij\DelveTest\main.go

When I tried to set a breakpoint in that file at line 15 it failed because of the Windows (i.e non unix) format of the file name.
    
    break C:\Go\src\github.com\jussij\DelveTest\main.go:15

That was easy enough to to fix so I made a change to issue this command instead:

    break C:/Go/src/github.com/jussij/DelveTest/main.go:15

But that also failed as shown below:

    break C:/Go/src/github.com/jussij/DelveTest/main.go:15
    Command failed: Location "C:/Go/src/github.com/jussij/DelveTest/main.go:15" not found

I think the reason it is failing is because at line 315 in this delve file:

    C:\Go\src\github.com\derekparker\delve\service\debugger\locations.go

The file lookup code below is failing to find that file:

    func (loc *NormalLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string) ([]api.Location, error) {
        funcs := d.process.Funcs()
        files := d.process.Sources()

Then after lots of head scratching I found out that this command works:

    break c:/go/src/github.com/jussij/DelveTest/main.go:15
    Breakpoint 1 set at 0x401240 for main.main c:/go/src/github.com/jussij/DelveTest/main.go:15

I suspect this error is somehow related to the Windows file system being case in-sensitive where as the Linix file system is case sensitive and Delve is coded for Linux.

But what has me confused is the file correctly named on the Windows disk is as follows:

    C:\Go\src\github.com\jussij\DelveTest\main.go

Which gives this Unix version:

    C:/Go/src/github.com/jussij/DelveTest/main.go

Yet the break command is expecting this:

    c:/go/src/github.com/jussij/DelveTest/main.go:15

If run the delve sources command I get this:

    sources
    <autogenerated>
    c:/go/src/errors/errors.go
    c:/go/src/fmt/doc.go
    c:/go/src/fmt/format.go
    c:/go/src/fmt/print.go
    c:/go/src/fmt/scan.go
    c:/go/src/github.com/jussij/DelveTest/main.go
    ..............

So where did that lower case version of the c:/go/ folder come from?

If this is a case sensitivity issue, could the Windows file test be made case insensitive?

For example if all the sources where made lower case that would be easy enough to also do that at the IDE side:

    c:/go/src/github.com/jussij/delvetest/main.go

Zeus also talks to Subversion (which is also file case sensitive) and that behavior has cause no amount of trouble, only because users on Windows get into the habit of not worry about the case of files as Windows does not care.

lukehoban:
Fixes #370.

The policy on path normalization after this change is roughly:

* `service/debugger` takes care of normalizing any path inputs from the user before using in comparisons
* `proc` package assumes normalized file names

This feels right, but other options might be available.  For example, normalization could be pushed all the way down into proc and any other code that handles paths.  Unfortunately `(*debug.gosym.Table).LineToPC` doesn't normalize it's input, so there would need to be a wrapper around that to do the normalization.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -160,8 +160,8 @@ func (dbp *Process) LoadInformation(path string) error {
 }
 
 // FindFileLocation returns the PC for a given file:line.
+// Assumes that `file` is normailzed to lower case and '/' on Windows.
 func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
-	fileName = filepath.ToSlash(fileName)
 	pc, _, err := dbp.goSymTable.LineToPC(fileName, lineno)
 	if err != nil {
 		return 0, err
--- a/proc/test/support.go
+++ b/proc/test/support.go
@@ -7,8 +7,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"testing"
 	"runtime"
+	"strings"
+	"testing"
 )
 
 // Fixture is a test binary.
@@ -64,7 +65,10 @@ func BuildFixture(name string) Fixture {
 
 	source, _ := filepath.Abs(path)
 	source = filepath.ToSlash(source)
-	
+	if runtime.GOOS == "windows" {
+		source = strings.ToLower(source)
+	}
+
 	Fixtures[name] = Fixture{Name: name, Path: tmpfile, Source: source}
 	return Fixtures[name]
 }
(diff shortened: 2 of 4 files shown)


jussij:
I merged in these changes and my initial testing shows those changes have fixed this issue.

Good stuff ;)

derekparker:
Merged.

Issue #354, opened 2016-01-21
panic: runtime error: makeslice: len out of range


antonikonovalov:
when debug of gorilla mux

run from idea go plugin 0.11.1026 in idea 16 last  eap
```
/usr/local/go/bin/go build -o "/private/var/folders/rl/lgwhvbs905lg29n026_5b3t00000gn/T/Build main.go and rungo" -gcflags "-N -l" /Users/antoniko/go/src/gitlab.nct/msg/msg/src/cmd/regulus/main.go
"/Users/antoniko/Library/Application Support/IntelliJIdea16/Go/lib/dlv/mac/dlv" --listen=localhost:57925 --headless=true exec "/private/var/folders/rl/lgwhvbs905lg29n026_5b3t00000gn/T/Build main.go and rungo" --
```

GO version 1.5.3
Idea plugin last version 
```
goroutine 29 [running]:
github.com/derekparker/delve/proc.(*Thread).readMemory(0xc820310480, 0xc820016540, 0xfffffffffffffff9, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/builduser/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/threads_darwin.go:108 +0x8d
github.com/derekparker/delve/proc.cacheMemory(0x4a4a178, 0xc820769110, 0xc820016540, 0xfffffffffffffff9, 0x0, 0x0)
	/Users/builduser/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/mem.go:42 +0xf2
github.com/derekparker/delve/proc.(*Variable).loadArrayValues(0xc820112c60, 0x1)
	/Users/builduser/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:996 +0xa6
github.com/derekparker/delve/proc.(*Variable).loadValueInternal(0xc820112c60, 0x1)
	/Users/builduser/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:802 +0x939
github.com/derekparker/delve/proc.(*Variable).loadValueInternal(0xc82009c540, 0x0)
	/Users/builduser/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:816 +0x10de
github.com/derekparker/delve/proc.(*Variable).loadValueInternal(0xc82009c000, 0x0)
[...]

PR #377, opened 2016-01-27
Two bugs reported in issue #354

aarzilli:
This pull request attempts to address the two bugs reported in Issue #354.

@antonikonovalov's bug is probably connected to reading uninitializer memory, I can't be sure because there aren't any exect instructions to reproduce it, but it seems pretty likely.

@OneOfOne's bug happens when the stack contains one frame or less.

### DIFF
--- a/terminal/command_test.go
+++ b/terminal/command_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/derekparker/delve/proc/test"
+	"github.com/derekparker/delve/service/api"
 )
 
 func TestCommandDefault(t *testing.T) {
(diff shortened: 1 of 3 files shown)


aarzilli:
I don't think that test failure has anything to do with this changeset and I don't see it happening locally.

terminal/command_test.go
```diff
@@ -98,3 +99,8 @@ func TestExecuteFile(t *testing.T) {
 		t.Fatalf("Wrong counts break: %d trace: %d\n", breakCount, traceCount)
 	}
 }
+
+func TestIssue354(t *testing.T) {
+	//printStack([]api.Stackframe{ }, "")
```
derekparker: Seems like leftover comment?

aarzilli: I actually want it to be there, I just commented it to check that it crashed on the second one too.

PR #380, opened 2016-01-28
dlv: Flag to print stacktrace on trace subcommand

derekparker:


aarzilli:
LGTM. I just wanted to mention that the trace subcommand should probably be implemented by setting the `Tracepoint` flag and then just calling `terminal.cont`.

derekparker:
>  I just wanted to mention that the trace subcommand should probably be implemented by setting the Tracepoint flag and then just calling terminal.cont.

Good call, I'll refactor this.

derekparker:
@aarzilli refactored.

aarzilli:
Merged with 8a1f36a1ce71d708d1d82afbc2191de9aefba021

PR #381, opened 2016-01-29
service/rpc: clean irrelevant net/rpc error messages

### LINKED ISSUE #378: dlv --log: method Run has wrong number of ins
This works without errors (mac os) in code that contains `http.ListenAndServe`
```
./debug -port=8080
dlv debug -- -port=8080
```

This works but displays the following errors
```
dlv --log debug -- -port=8080
2016/01/27 20:51:59 debugger.go:56: launching process with args: [./debug -port=8080]
2016/01/27 20:51:59 server.go:301: method Run has wrong number of ins: 1
2016/01/27 20:51:59 server.go:301: method Stop has wrong number of ins: 2
(dlv) exit
```
(https://github.com/Microsoft/vscode-go/issues/175)

aarzilli:
Note that this change is purely cosmetic, I tried to do it with the least possible amount of diff noise I could get away with.

Fixes #378

### DIFF
--- a/service/rpc/server.go
+++ b/service/rpc/server.go
@@ -14,6 +14,10 @@ import (
 	"github.com/derekparker/delve/service/debugger"
 )
 
+type ServerImpl struct {
+	s *RPCServer
+}
+
 type RPCServer struct {
 	// config is all the information necessary to start the debugger and server.
 	config *service.Config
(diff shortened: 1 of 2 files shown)


derekparker:
Does this patch actually fix the issue? Won't it still complain about the arity? I'd say, instead of wrapping RPCServer, let's just fix the arity on the methods it's complaining about, with `_` as name of param to indicate clearly that they are ignored.

aarzilli:
The methods are fine with the arity they have, `Run` and `Stop` aren't part of the API, they are called server side. This patch publishes an object without `Run` and `Stop` so there's no complaints from `net/rpc`.

Another way would be to make an interface for `RPCServer` that contains only the API methods and publish the object through that interface. The problem with this solution is that it adds yet another place that we need to change when we add a new command.

derekparker:
Ah, I see now.

derekparker:
Merged.

PR #386, opened 2016-01-30
Fix path lookup logic on Windows.

### LINKED ISSUE #370: Breakpoint Locations should be case insensitve on Windows

lukehoban:
Addresses further issues raised after merging #375.

It turns out that the Go compiler does not actually canonicalize paths in the symbol table on Windows.  

Instead, when looking up user-provided paths in the symbol table, we have to iterate the symbol tables
keys to do a case- and slash- insensitive comparison.

This maintains the policy set in #375 of doing all the path normalization in `service/debugger`.  

Fixes #370.

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"path/filepath"
 	"regexp"
+	"runtime"
+	"strings"
 
 	"github.com/derekparker/delve/proc"
 	"github.com/derekparker/delve/service/api"
@@ -149,7 +152,18 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	)
 	switch {
 	case len(requestedBp.File) > 0:
-		addr, err = d.process.FindFileLocation(normalizePath(requestedBp.File), requestedBp.Line)
+		fileName := requestedBp.File
+		if runtime.GOOS == "windows" {
+			// Accept fileName which is case-insensitive and slash-insensitive match
+			fileNameNormalized := strings.ToLower(filepath.ToSlash(fileName))
+			for symFile := range d.process.Sources() {
+				if fileNameNormalized == strings.ToLower(filepath.ToSlash(symFile)) {
+					fileName = symFile
+					break
+				}
+			}
+		}
+		addr, err = d.process.FindFileLocation(fileName, requestedBp.Line)
 	case len(requestedBp.FunctionName) > 0:
 		if requestedBp.Line >= 0 {
 			addr, err = d.process.FindFunctionLocation(requestedBp.FunctionName, false, requestedBp.Line)
--- a/service/debugger/locations.go
+++ b/service/debugger/locations.go
@@ -50,13 +50,6 @@ type FuncLocationSpec struct {
 	BaseName              string
 }
 
-func normalizePath(path string) string {
-	if runtime.GOOS != "windows" {
-		return path
-	}
-	return strings.ToLower(filepath.ToSlash(path))
-}
-
 func parseLocationSpec(locStr string) (LocationSpec, error) {
 	rest := locStr
 
@@ -291,8 +284,11 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
 }
 
 func partialPathMatch(expr, path string) bool {
-	expr = normalizePath(expr)
-	path = normalizePath(path)
+	if runtime.GOOS == "windows" {
+		// Accept `expr` which is case-insensitive and slash-insensitive match to `path`
+		expr = strings.ToLower(filepath.ToSlash(expr))
+		path = strings.ToLower(filepath.ToSlash(path))
+	}
 	if len(expr) < len(path)-1 {
 		return strings.HasSuffix(path, expr) && (path[len(path)-len(expr)-1] == '/')
 	} else {
--- a/_fixtures/locationsUpperCase.go
+++ b/_fixtures/locationsUpperCase.go
@@ -1,0 +1,7 @@
+package main
+
+import "fmt"
+
+func main() {
+	fmt.Println("test")
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ version: '{build}'
 os: Windows Server 2012 R2
 clone_folder: c:\gopath\src\github.com\derekparker\delve
 environment:
-  GOPATH: c:\gopath
+  GOPATH: C:\gopath
 install:
   - ps: |
       # Install MinGW.
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -161,7 +161,7 @@ func (dbp *Process) LoadInformation(path string) error {
 }
 
 // FindFileLocation returns the PC for a given file:line.
-// Assumes that `file` is normailzed to lower case and '/' on Windows.
+// Assumes that `file` is normailzed to lower case and '/' on Windows. 
 func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
 	pc, _, err := dbp.goSymTable.LineToPC(fileName, lineno)
 	if err != nil {
--- a/proc/test/support.go
+++ b/proc/test/support.go
@@ -8,7 +8,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
 	"testing"
 )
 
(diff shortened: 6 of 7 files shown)


service/debugger/debugger.go
```diff
@@ -149,7 +152,18 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	)
 	switch {
 	case len(requestedBp.File) > 0:
-		addr, err = d.process.FindFileLocation(normalizePath(requestedBp.File), requestedBp.Line)
+		fileName := requestedBp.File
+		if runtime.GOOS == "windows" {
+			// Accept fileName which is case-insensitive and slash-insensitive match
+			fileNameNormalized := filepath.ToSlash(strings.ToLower(fileName))
```
derekparker: `normalizePath` should be used here.

service/debugger/debugger.go
```diff
@@ -149,7 +152,18 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
+			for symFile := range d.process.Sources() {
+				if fileNameNormalized == filepath.ToSlash(strings.ToLower(symFile)) {
```
derekparker: Here as well.

lukehoban:
Updated to use `normalizePath` for all path normalization.

service/debugger/locations.go
```diff
@@ -290,6 +283,13 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
 	return partialPathMatch(loc.Base, path)
 }
 
+func normalizePath(path string) string {
+	if runtime.GOOS != "windows" {
```
ignatov: Invert condition?

aarzilli:
Can we add tests for this? For example check that setting breakpoints works on a program that has upcase letters in the name? Plus a windows-only test that sets breakpoints using the wrong casing?

service/debugger/locations.go
derekparker: Agreed, make non-windows code path the default. E.g.:

```go
        return strings.ToLower(filepath.ToSlash(path))
}
return path
```

lukehoban:
Tests are now included.

ignatov:
@derekparker any updates? Will be really cool to ship the plugin for IntelliJ with Delve on Windows as soon as possible.

service/debugger/debugger.go
```diff
@@ -149,7 +150,18 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	)
 	switch {
 	case len(requestedBp.File) > 0:
-		addr, err = d.process.FindFileLocation(normalizePath(requestedBp.File), requestedBp.Line)
+		fileName := requestedBp.File
+		if runtime.GOOS == "windows" {
```
derekparker: Instead of adding the check here, let's remove the assumption in `FindFileLocation` that it's input has been normalized, let's take that burden off the caller and centralize that logic within that method. That way we're not special casing throughout the codebase.

Additionally, since `normalizePath` already is checking `runtime.GOOS` let's remove that from any logic that calls it. I'm also unsure about the iteration here, we break when `fileNameNormalized == normalizePath(symfile)`, but why can't we just assume that the string returned from `normalizePath` will correctly match a file (if it exists in sources)? Since, if that iteration fails to find anything `fileName` will not be normalized at all, and we don't bail if we can't find anything either, so that check isn't buying us anything it seems.

derekparker:
Sorry for the delay in reviewing this, just finished up a lot of traveling for Fosdem.

service/debugger/debugger.go
lukehoban: > Instead of adding the check here, let's remove the assumption in  FindFileLocation  that its input has been normalized, let's take that burden off the caller and centralize that logic within that method. That way we're not special casing throughout the codebase.

Happy to do this.  Though there is only one caller currently who has to worry about this, so making this change won't reduce the number of places in total, it will just push the boundary of where path normalization further in.  I think that's fine though, and make this a little less brittle.

It does though mean that we will need to duplicate `normalizePath` since we'll have users of that functionality in both `service/debugger` and `proc`, and there's no clean way I can see to share utility code like `normalizePath` between these two.

>> Additionally, since  normalizePath  already is checking  runtime.GOOS  let's remove that from any logic that calls it. 

In that case, we would do this key iteration on all platforms, which will be less efficient on platforms that don't need it, but is otherwise okay.

> why can't we just assume that the string returned from  normalizePath  will correctly match a file (if it exists in sources)?

There's no good way I can tell to correctly guess what normalization the Go compiler will have used.  That's actually exactly the problem that this PR is fixing.

I had previously thought that the Go compiler always did a normalization to lowercase with `/`.  But I was only seeing that because I had my `GOPATH` set to lowercase and happened to be using all lower case files.  

So to do the case-insensitive path comparison we actually have to normalize both the requested filename as well as the name used as a key in the Go symbol table.  The latter is built outside of Delve's control by the Go compiler - potentially even on a different machine.

As an example, imagine that the Go compiler embedded `C:\go\src\foo.go`, but the user requests a bp on `c:\Go\src\foo.go`.  These should match, but neither one normalizes to the other.

lukehoban:
Updated based on some of @derekparker's suggestions.  Thoughts?

derekparker:
@lukehoban thanks for your reply. Let's go ahead and revert to your previous version (sorry :-/), keeping the logic in the caller (essentially in the debugger package). I misunderstood the logic for the iteration, and that was my fault. Ultimately I think it's fine in the debugger for now, my goal with my feedback last night was to attempt to make this normalization fall into a default file lookup path, so we can tackle this problem somewhat permanently even if future code dealing with paths is added, if that makes sense. It may be the case that we end up pushing this further down into proc and have some sort of wrapper around calling any function in the line table, but that's more of a refactor.

Anyways, let's revert to the version prior to my last review and I will merge.

lukehoban:
Got it.  

Updated now to be similar to what was in the previous revision.  I've left `normalizePath` out though, since at least one of the two callers would still need to have a separate `runtime.GOOS == "windows"` check anyway.  Simpler for now to just inline the normalization logic.

derekparker:
@lukehoban the provided test only exercises the logic in `partialPathMatch`, can you please add another test to exercise the new logic in `CreateBreakpoint`.

lukehoban:
Updated tests to cover `CreateBreakpoint` as well.

derekparker:
Merged, thanks!

Issue #383, opened 2016-01-29
Extend Delve backend RPC API to be able to make a Delve REPL front-end console in a IDE


gertcuykens:
Some commands are better in a console interface, to create a console in a IDE it would be much easier if the API has also REPL front-end support.

Example: https://github.com/Microsoft/vscode-go/issues/186

Please discuss with @lukehoban I am just the ticket man :)

lukehoban:
Thanks @gertcuykens. 

To provide some more technical details on the issue here:

IDE integrations make calls through the RPC API, so ensure that they can programmatically invoke commands and read results in a structured way.  

However, in some cases, the IDE integration may want to support also allowing users to drop into a REPL where they can send commands to the current instance of `dlv` using the same front-end commands as the `dlv` binary.

Right now, the structure of Delve doesn't appear to make this possible.  There is no way to invoke the Delve front-end code (`terminal/command.go`) from the RPC API (`server/rpc/client.go`).  This design is nicely decoupled.  But it doesn't support this IDE scenario (unless the IDE effectively re-implements `terminal/command` and stays up to date with it).

aarzilli:
Would it be unacceptable to run a dlv frontend with `dlv connect` and communicate with it using pipes? 
It seems that the biggest problem is that the API is not reentrant, especially `Command`, so you would have to make sure no two calls are ever made simultaneously.

lukehoban:
@aarzilli 

Good suggestion - connecting with `dlv connect` would be an acceptable solution I think.

Unfortunately, it doesn't seem to work when connecting to a headless server that is in the middle of a debug session that has been accepting RPC requests from another client.  I tried this, and I can connect, but any commands I try to execute hang `dlv`.

```
c:\dd\go\src\github.com\lukehoban\hello>dlv connect 127.0.0.1:2345
Type 'help' for list of commands.
(dlv) sources

```

If I then continue the other client to the end of the debug session, the `dlv connect` session unhangs with:

```
Command failed: read tcp 127.0.0.1:4991->127.0.0.1:2345: wsarecv: An existing connection was forcibly closed by the remote host.
(dlv)
```

Is this expected?

aarzilli:
It seems the server only accepts one connection, probably the correct behavior give how the API works. I've added a flag to allow multiple clients in PR #395.

aarzilli:
Closed by c277b2715750a46153b2974f66eec84d438f669c

PR #395, opened 2016-02-01
cmd/dlv: add flag to make headless server accept multiple clients

aarzilli:
See issue #383 for related discussion.

### DIFF
--- a/service/rpc/server.go
+++ b/service/rpc/server.go
@@ -23,6 +23,8 @@ type RPCServer struct {
 	config *service.Config
 	// listener is used to serve HTTP.
 	listener net.Listener
+	// stopChan is used to stop the listener goroutine
+	stopChan chan struct{}
 	// debugger is a debugger service.
 	debugger *debugger.Debugger
 }
@@ -38,13 +40,22 @@ func NewServer(config *service.Config, logEnabled bool) *ServerImpl {
 		&RPCServer{
 			config:   config,
 			listener: config.Listener,
+			stopChan: make(chan struct{}),
 		},
 	}
 }
 
 // Stop detaches from the debugger and waits for it to stop.
 func (s *ServerImpl) Stop(kill bool) error {
-	return s.s.debugger.Detach(kill)
+	if s.s.config.AcceptMulti {
+		close(s.s.stopChan)
+		s.s.listener.Close()
+	}
+	err := s.s.debugger.Detach(kill)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 // Run starts a debugger and exposes it with an HTTP server. The debugger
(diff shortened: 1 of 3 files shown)


service/rpc/server.go
```diff
@@ -60,16 +60,21 @@ func (s *ServerImpl) Run() error {
...
-		if err != nil {
-			panic(err)
-		}
 		defer s.s.listener.Close()
-
-		rpcs := grpc.NewServer()
-		rpcs.Register(s.s)
-		rpcs.ServeCodec(jsonrpc.NewServerCodec(c))
+		for {
+			c, err := s.s.listener.Accept()
```
derekparker: We need to signal this goroutine to bail when the server is told to stop, as it should not continue accepting connections. Maybe when the server is initialized (in `NewServer`) we create 2 channels, `stopChan` and `newConnChan`. In this loop we can use `select { ... }` and do the appropriate task depending on the channel.

aarzilli:
I think we can probably just close `s.s.listener` from somewhere else to kill this goroutine, however what decides that the server should stop?

derekparker:
@aarzilli good point, better solution than mine. How about we close the listener in `ServerImpl.Stop`?

service/rpc/server.go
```diff
@@ -60,16 +67,21 @@ func (s *ServerImpl) Run() error {
...
+			if err != nil {
+				panic(err)
```
derekparker: We should no longer simply panic here. Since the listener could be closed elsewhere, if this goroutine is allowed to execute long enough to panic users will get confused. We need to either detect [this error](https://golang.org/src/net/net.go#L373) within `OpError` -- or come up with a scheme where this goroutine is the one that closes the listener to avoid this issue.

Since that error var isn't exported I feel we would have to resort to string matching which seems kind of brittle if that error message ever changes significantly (which is not likely, but still possible I suppose).

The other route would be, instead of closing the listener elsewhere, we signal this goroutine to close it like so:

```go
go func() {
	conns := make([]net.Conn)
	defer func() {
		for _, conn := range conns {
			conn.Close()
		}
		s.s.listener.Close()
	}()
	for {
		c, err := s.s.listener.Accept()
		if err != nil {
			select {
			case <-s.s.stopChan:
				// We were supposed to exit, do nothing and return
				return
			default:
				panic(err)
			}
		}
		conns = append(conns, con)
		go rpcs.ServeCodec(jsonrpc.NewServerCodec(c))
		if !s.s.config.AcceptMulti {
			break
		}
	}
}()
```

And then `Stop` becomes:

```go
func (s *ServerImpl) Stop(kill bool) error {
	err := s.s.debugger.Detach(kill)
	if err != nil {
		return err
	}
	if s.s.config.AcceptMulti {
		close(s.s.stopChan)
	}
	return nil
}
```

Thoughts?

derekparker: Also, if we go with that model then we should also place the `close(s.s.stopChan)` in a defer in case `Detach` returns an error, we would still want to close the listener and cleanup connections.

Issue #332, opened 2015-12-28
could not find FDE for PC


dlsniper:
Hi,

Using go 1.5.2, delve d4bfd25a28cab4589808471d987e28ae6c6175c6 and the code sample from here: http://play.golang.org/p/GDQCFi8dgn

- set a breakpoint on line 8
- try to use `si` a couple of times
- then use `n` a few times
- eventually `Command failed: could not find FDE for PC 0xc8200601a0` (or some other address).

Please let me know if you need any further details.

(Original issue at https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2185 reported by @PatrickLaban)

Issue #368, opened 2016-01-25
Display disassembly of a function


derekparker:
We should have a command to display the assembly for a given function. Most likely the easiest path forward is to simply have Delve invoke `go tool objdump -s <symname>`. Either that or take a look at that tool and see if we can use the code there as a package and do it all programmatically.

aarzilli:
Implemented in e7a9a3ea9a72d943a9e9012977d28419933f5896, Issue #415 tracks remaining feature request.

PR #399, opened 2016-02-02
proc: bugs setting next breakpoints

aarzilli:
These bugs were found while investigating Issue #332:

1. A running goroutine is by definition not parked waiting for a
chan recv
2. The FDE end address is intended to be exclusive, the code
interpreted as inclusive and sometimes ended up setting a breakpoint
on a function other than the current one.

This however does not fix Issue #332 itself. The problem there is that the FDE information is wrong inside some parts of the prologue for some functions (not all functions and not all parts of the prologue). I have added a test (disabled) for this problem.

The fix for issue #332 is to check if the current instruction is a `CALL` and if it is set a temp breakpoint after the function's prologue and call `proc.(*Process).Continue`.
We would need to implement Issue #368 to do this.

In addition I would also return an error when a breakpoint is set on a function prologue using the `<filename>:<lineno>` syntax, given that in addition to being generally confusing, this problem with FDE means it could break delve.

### DIFF
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -19,6 +19,8 @@ import (
 	protest "github.com/derekparker/delve/proc/test"
 )
 
+const disableFailingTests = true
+
 func init() {
 	runtime.GOMAXPROCS(4)
 	os.Setenv("GOMAXPROCS", "4")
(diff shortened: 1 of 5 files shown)


derekparker:
@aarzilli is the FDE issue explicitly a problem with the frame information the Go linker is generating, or how we're handling it? If it is an issue with the Dwarf information, let's make sure we can file a bug in the Go issue tracker about it, and track it on our end.

aarzilli:
As far as I can tell it's a problem with the linker.

aarzilli:
This is the text of main.changeMe:

```
TEXT main.changeMe(SB) /home/a/n/go/src/github.com/derekparker/delve/_fixtures/issue332.go
        issue332.go:12  0x4012e0        64488b0c25f8ffffff      FS MOVQ FS:0xfffffff8, CX
        issue332.go:12  0x4012e9        488d4424f8              LEAQ -0x8(SP), AX
        issue332.go:12  0x4012ee        483b4110                CMPQ 0x10(CX), AX
        issue332.go:12  0x4012f2        0f8638010000            JBE 0x401430
        issue332.go:12  0x4012f8        4881ec88000000          SUBQ $0x88, SP
…
        issue332.go:14  0x401422        8903                    MOVL AX, 0(BX)
        issue332.go:14  0x401424        e94effffff              JMP 0x401377
        issue332.go:13  0x401429        e892400200              CALL runtime.panicindex(SB)
        issue332.go:13  0x40142e        0f0b                    UD2
        issue332.go:12  0x401430        e82b240500              CALL runtime.morestack_noctxt(SB)
        issue332.go:12  0x401435        e9a6feffff              JMP main.changeMe(SB)
        issue332.go:12  0x40143a        0000                    ADDL AL, 0(AX)
        issue332.go:12  0x40143c        0000                    ADDL AL, 0(AX)
        issue332.go:12  0x40143e        0000                    ADDL AL, 0(AX)
```

This is the FDE table for `main.changeMe` as printed by objdump:

```
00000040 0000000000000028 00000000 FDE cie=00000000 pc=00000000004012e0..0000000000401440
  DW_CFA_def_cfa_offset_sf: 8
  DW_CFA_advance_loc: 31 to 00000000004012ff
  DW_CFA_def_cfa_offset_sf: 144
  DW_CFA_advance_loc2: 270 to 000000000040140d
  DW_CFA_def_cfa_offset_sf: 8
  DW_CFA_advance_loc: 1 to 000000000040140e
  DW_CFA_def_cfa_offset_sf: 144
  DW_CFA_advance_loc: 49 to 000000000040143f
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
```

If I interpret this correctly this says that between 0x40140e and 0x401440 the CFA offset is -144, however it's clear from the text that in the range 0x401430…0x401440 it should be the same offset that is set for the beginning of the function, ie -8.
Am I wrong?

derekparker:
@aarzilli I think you are correct, the stack is not grown until PC `0x4012f8` which should _never_ happen until the prologue has finished and the goroutine has determined it has enough stack (it should always hit the branch one instruction prior). So the remaining portion of the prologue `0x401430` ... `0x401435` should be an offset of 8, since we're guaranteed to not have grown our frame at that point.

Can you post a gist of the entire TEXT of that function?

aarzilli:
https://gist.github.com/aarzilli/6544327c04b14f4a26f5

derekparker:
Alright yeah, the frame entry should look like this instead:

```
00000040 0000000000000028 00000000 FDE cie=00000000 pc=00000000004012e0..0000000000401440
  DW_CFA_def_cfa_offset_sf: 8
  DW_CFA_advance_loc: 31 to 00000000004012ff
  DW_CFA_def_cfa_offset_sf: 144
  DW_CFA_advance_loc2: 270 to 000000000040140d
  DW_CFA_def_cfa_offset_sf: 8
  DW_CFA_advance_loc: 1 to 000000000040140e
  DW_CFA_def_cfa_offset_sf: 144
  DW_CFA_advance_loc: <n> to 0000000000401430
  DW_CFA_def_cfa_offset_sf: 8
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop
```

derekparker:
Basically `0x40140e`...`0x40142e` is reachable via the branch at `0x4013de` so could be run after the frame is setup, but after `0x401430` the rest is the continuation of the prologue code and is run before the frame is setup.

derekparker:
@aarzilli I think this warrants a bug in Go. I can submit the issue, but can you verify the Go compiler version your using, and compare the results against the current 1.6rc to verify this hasn't already been addressed?

derekparker:
@aarzilli just checked this issue out using Go 1.6rc2 and the problem seems to be resolved.

proc/proc_test.go
```diff
@@ -1475,3 +1477,64 @@ func TestIssue384(t *testing.T) {
...
+		}
+	})
+}
+
+func TestIssue332_Part2(t *testing.T) {
+	// Step should skip a function's prologue
+	// In some parts of the prologue, for some functions, the FDE data is incorrect
+	// which leads to 'next' and 'stack' failing with error "could not find FDE for PC: <garbage>"
+	// because the incorrect FDE data leads to reading the wrong stack address as the return address
+	if disableFailingTests {
```
derekparker: Add a TODO here to enable or remove this test once 1.6 lands, since that seems to fix the FDE issue.

aarzilli:
done.

PR #401, opened 2016-02-02
cmd/dlv: bugfix: trace disables terminal echo

### LINKED ISSUE #397: trace subcommand leaves terminal with echo disabled
Probably introduced by the trace refactoring, I didn't notice during the review.

aarzilli:
Fixes #397

### DIFF
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -41,9 +41,13 @@ func New(client service.Client, conf *config.Config) *Term {
 	}
 }
 
+func (t *Term) Close() {
+	t.line.Close()
+}
+
 // Run begins running dlv in the terminal.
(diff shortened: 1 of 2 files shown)


terminal/terminal.go
```diff
@@ -41,6 +41,10 @@ func New(client service.Client, conf *config.Config) *Term {
 	}
 }
 
+func (t *Term) Close() {
+	t.line.Close()
+}
+
 // Run begins running dlv in the terminal.
 func (t *Term) Run() (int, error) {
 	defer t.line.Close()
```
derekparker: nit: let's call `t.Close()` here as well to stay consistent with how we cleanup Term.

Issue #402, opened 2016-02-02
Step Command Not Working


jussij:
This issue is in regard to Delve running on Windows.

Consider this main.go example file:

    // main.go
    package main

    import "fmt"

    type User struct {
        FirstName string
        LastName  string
    }

    func (user User) String() string {
        return fmt.Sprintf("%s %s", user.FirstName, user.LastName)
    }

    func main() {
        user := User{
            FirstName: "John",
            LastName:  "Smith",
        }

        message := FormatMessage(user, "Golang Weekly Newsletter #756")

        for index := 0; index < 3; index++ {
            fmt.Printf("Sending #%d message with %s\n", index, message)
        }
    }

    func FormatMessage(user User, message string) string {
        return fmt.Sprintf("body: %s by %s", message, user.String())
    }

Run the debugger using this file:

    dlv debug main.go

Next set the break point at line 21 using this command:

    break main.go:21

That results in this output:

    (dlv) break main.go:21
    Breakpoint 1 set at 0x4012be for main.main C:/Go/src/github.com/jussij/DelveTest/main.go:21

Next issue the continue command which should hit that break point:

    continue

As you can see it does hits that break point and correctly stops as shown below:

    > main.main() C:/Go/src/github.com/jussij/DelveTest/main.go:21 (hits goroutine(1):1 total:1)
        16:         user := User{
        17:                 FirstName: "John",
        18:                 LastName:  "Smith",
        19:         }
        20:
    =>  21:         message := FormatMessage(user, "Golang Weekly Newsletter #756")
        22:
        23:         for index := 0; index < 3; index++ {
        24:                 fmt.Printf("Sending #%d message with %s\n", index, message)
        25:         }
        26: }

But look what happens if you then issue the next command which gives you this output:

    (dlv) next
    > main.FormatMessage() C:/Go/src/github.com/jussij/DelveTest/main.go:28
        23:         for index := 0; index < 3; index++ {
        24:                 fmt.Printf("Sending #%d message with %s\n", index, message)
        25:         }
        26: }
        27:
    =>  28: func FormatMessage(user User, message string) string {
        29:         return fmt.Sprintf("body: %s by %s", message, user.String())
        30: }

From the commands reference web page: https://github.com/derekparker/delve/wiki/Commands

It says next does this:

    Next - Step over to next source line.

In the example above example the **next** command has not stepped over the FormatMessage function but instead it stepped into that function

aarzilli:
[...]

Issue #396, opened 2016-02-01
`proc.(*Process).FunctionEntryToFirstLine` can hang forever


aarzilli:
For example:

```
dlv trace main.init
```

will hang forever because calling `FunctionEntryToFirstLine` never returns when called on `main.init`. The problem is that main.init doesn't have any lines.

PR #405, opened 2016-02-06
Disassemble command

aarzilli:
Implements #368 

This is more complicated than it could be but I intend to use it to skip the prologue for #332, #396 and #402.


terminal/command.go
```diff
@@ -83,6 +84,7 @@ func DebugCommands(client service.Client) *Commands {
 		{aliases: []string{"stack", "bt"}, cmdFn: stackCommand, helpMsg: "stack [<depth>] [-full]. Prints stack."},
 		{aliases: []string{"frame"}, cmdFn: frame, helpMsg: "Sets current stack frame (0 is the top of the stack)"},
 		{aliases: []string{"source"}, cmdFn: c.sourceCommand, helpMsg: "Executes a file containing a list of delve commands"},
+		{aliases: []string{"diassemble", "disass"}, cmdFn: g0f0(disassCommand), helpMsg: "Disassembles memory"},
```
derekparker: Need to provide flag listing in help message.

derekparker:
Only one nit, which I'll go ahead and fix before merging. Other than that, the only other suggestion I have is that we provide a flag to the command to let the user specify gnu/intel syntax. Otherwise this looks very solid, thanks a lot!

I'll create an issue to track my feedback on the syntax flavor flag.

derekparker:
Merged. I also noticed that `disassemble` was spelled incorrectly, fixed that as well.

derekparker:
Created https://github.com/derekparker/delve/issues/415 to track flavor output.

PR #407, opened 2016-02-07
proc: Bug loading maps

### LINKED ISSUE #406: Printing a variable causes: panic: runtime error: index out of range
Mac OS X 10.11.3
Code that causes panic: [Go Playground](https://play.golang.org/p/qnvETiq85g)
go version go1.6rc2 darwin/amd64

Delve Debugger
Version: 0.11.0-alpha
Build: 06bba9b7a9ad0e06f4d74bf71f7667d24872a838

To reproduce: 

1. set breakpoint: b cfg_parser.go:146
2. Print variable: p cfgtree

Output and stack trace attached
[delve_panic.txt](https://github.com/derekparker/delve/files/120317/delve_panic.txt)

aarzilli:
Past the maximum recursion depth maps shouldn't be loaded at all,
adding map children and not loading them breaks assumptions in
the prettyprinter.

Fixes #406

### DIFF
--- a/_fixtures/issue406.go
+++ b/_fixtures/issue406.go
@@ -1,0 +1,262 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+
+	//	"golang.org/x/crypto/ssh/terminal"
+)
+
+// Blacklist type is a map of Nodes with string keys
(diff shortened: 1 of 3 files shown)


derekparker:
Merged.

PR #413, opened 2016-02-11
Using the disassembler to skip the prologue

aarzilli:
This PR builds on top of PR #405.

I replaced FunctionEntryToFirstLine with SkipPrologue. SkipPrologue disassembles the function, does some really simple pattern matching to detect where the prologue ends (and if exists) and returns the address of the first instruction after the prologue.

This means that SkipPrologue will work correctly on functions that are defined on a single line, it should never hang (#396) and also that we can still call it if the users sets a breakpoint with `<filename>:<lineno>` because the first instruction after the prologue, usually, still refers to the same source line as the prologue.

### DIFF
--- a/proc/disasm_amd64.go
+++ b/proc/disasm_amd64.go
@@ -1,6 +1,7 @@
 package proc
 
 import (
+	"debug/gosym"
 	"encoding/binary"
 	"rsc.io/x86/x86asm"
 )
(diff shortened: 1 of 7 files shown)


derekparker:
Please rebase.

aarzilli:
rebased

proc/disasm_amd64.go
```diff
@@ -111,3 +112,50 @@ func (thread *Thread) resolveCallArg(inst *ArchInst, currentGoroutine bool, regs
...
+
+var windowsPrologue = instrseq{x86asm.MOV, x86asm.MOV, x86asm.LEA, x86asm.CMP, x86asm.JBE}
+var windowsPrologue2 = instrseq{x86asm.MOV, x86asm.MOV, x86asm.CMP, x86asm.JBE}
+var linuxPrologue = instrseq{x86asm.MOV, x86asm.LEA, x86asm.CMP, x86asm.JBE}
+var linuxPrologue2 = instrseq{x86asm.MOV, x86asm.CMP, x86asm.JBE}
+var prologues = []instrseq{windowsPrologue, windowsPrologue2, linuxPrologue, linuxPrologue2}
+
+// Returns the address of the first instruction after the prologue for function fn
+// If sameline is set SkipPrologue will always return an address associated with the same line as fn.Entry
+func (dbp *Process) SkipPrologue(fn *gosym.Func, sameline bool) (uint64, error) {
```
derekparker: nit: Rename to something like `FirstPCAfterPrologue`.

proc/disasm_amd64.go
```diff
@@ -111,3 +112,50 @@ func (thread *Thread) resolveCallArg(inst *ArchInst, currentGoroutine bool, regs
...
+				}
+			}
+			return r.Loc.PC, nil
+		}
+	}
+
+	return fn.Entry, nil
+}
+
+func checkPrologue(s []AsmInstruction, prologue instrseq) bool {
```
derekparker: nit: rename to `prologueMatches` or something similar to better describe the intent.

derekparker:
This makes me think we should go back to setting a breakpoint on `fn.Entry` when a users types `break foo` and then rely on this to skip the prologue when they issue a `next` or `step` command. That way it's less surprising than landing on the first line of the function instead of the entry, and we still get the added benefit of skipping the prologue to avoid the odd behavior.

aarzilli:
I don't think it's possible to implement `Continue` correctly if we set a breakpoint on `fn.Entry`. If we do that `Continue` will land us on `fn.Entry` to `Continue` correctly from that point we would need to disable the breakpoint on `fn.Entry` and set a new one after the prologue, but we can't disable the original breakpoint because another goroutine may hit it in the meantime.

Even if it's possible it would seem like it requires a lot of tricky bookkeeping.

proc/disasm_amd64.go
```diff
@@ -111,3 +112,50 @@ func (thread *Thread) resolveCallArg(inst *ArchInst, currentGoroutine bool, regs
 	}
 	return &Location{PC: pc, File: file, Line: line, Fn: fn}
 }
+
+type instrseq []x86asm.Op
+
+var windowsPrologue = instrseq{x86asm.MOV, x86asm.MOV, x86asm.LEA, x86asm.CMP, x86asm.JBE}
+var windowsPrologue2 = instrseq{x86asm.MOV, x86asm.MOV, x86asm.CMP, x86asm.JBE}
+var linuxPrologue = instrseq{x86asm.MOV, x86asm.LEA, x86asm.CMP, x86asm.JBE}
```
derekparker: Does this apply to OSX as well (I'm assuming so), if so let's just rename this `unixPrologue` for now. Also, I'm somewhat tempted to prefer splitting these declarations out into their own OS specific file, but I fear that would hide the discoverability of `prologues`.

derekparker:
Other than the naming nit, this LGTM. Please rebase and address the naming nit and I'll merge.

derekparker:
Merged.

PR #417, opened 2016-02-12
proc: Step should skip function prologue 

### LINKED ISSUE #332: could not find FDE for PC

aarzilli:
Builds on top of #413

Step disassembles the current instruction, if it is a CALL sets a
temp breakpoint inside the called function, after the prologue and
calls Continue.

Fixes #332

### DIFF
--- a/_fixtures/callme.go
+++ b/_fixtures/callme.go
@@ -1,0 +1,27 @@
+package main
+
+import "fmt"
+
+func callme(i int) {
+	fmt.Println("got:", i)
+}
+
+const nBytes = 10
+var zeroarr [nBytes]byte
+
+func callme2() {
+	for i := 0; i < nBytes; i++ {
+		zeroarr[i] = '0'
+	}
+}
+
+func callme3() {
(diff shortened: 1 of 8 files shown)


aarzilli:
Why aren't Travis and AppVeyor testing this?

derekparker:
Not sure why AppVeyor is not testing this, but closing/reopening caused Travis to pick up on it. Issue with GH hooks maybe?

derekparker:
Merged.

Issue #419, opened 2016-02-14
Panic on signal handling


dlsniper:
Hi,

On the following program:
```go
package main

import (
	"fmt"
	"os"
	"os/signal"
)

func main() {

	fmt.Println("Start")

	sc := make(chan os.Signal, 1)

	//os.Interrupt, os.Kill
	signal.Notify(sc, os.Interrupt, os.Kill)

	quit := <-sc

	fmt.Printf("Receive signal %s \n", quit.String())

	fmt.Println("End")

}
```

Simply compile it to a binary and then run `dlv exec <binary>`. In delve issue the `continue` command, with no breakpoint or anything else set. Then, after the `Start` text appears on screen, simply `CTRL+C` the app and it will give the following crash:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0xc8 pc=0x5f9a2c]

goroutine 68 [running]:
github.com/derekparker/delve/proc.(*Variable).structMember(0x0, 0xa63f88, 0x2, 0x0, 0x0, 0x0)
        /opt/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:593 +0x3c
github.com/derekparker/delve/proc.(*Variable).toFieldNamed(0x0, 0xa63f88, 0x2, 0x0)
        /opt/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:398 +0x39
github.com/derekparker/delve/proc.(*Variable).parseG(0xc8201661c0, 0xc8201661c0, 0x0, 0x0)
        /opt/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/variables.go:369 +0x806
github.com/derekparker/delve/proc.(*Thread).GetG(0xc82013cc80, 0xc82025a120, 0x0, 0x0)
        /opt/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/proc/threads.go:317 +0x69
github.com/derekparker/delve/service/api.ConvertThread(0xc82013cc80, 0xc8201428d0)
        /opt/buildAgent/work/b97f8195591f35ac/gopath/src/github.com/derekparker/delve/service/api/conversions.go:62 +0x105
github.com/derekparker/delve/service/debugger.(*Debugger).State(0xc82013a810, 0x0, 0x0, 0x0)
[...]

PR #421, opened 2016-02-15
Fixes for Issue #419

aarzilli:
There were a couple of problems. The crash was caused (probably) by a race condition between `debugger.(*Debugger).State`, called by `rpc.(*Client).Halt`, and `proc.(*Process).Continue` both reading the type `runtime.g` simultaneously.
They both call `proc.(*Thread).GetG` which calls `debug/dwarf.(*Data).Type` which is not thread safe and can return incomplete values if called simultaneously by multiple threads.

The second problem is that this test example will never terminate when it is run under delve on linux, even if we send it a SIGTERM using `kill(1)` because we intercept and suppress all signals. I have changed `proc.(*Process).trapWait` to propagate any signal we don't handle ourselves to inferior, I don't know if we want to do this or not, it seems like something that could be useful and the OS X version already does it implicitly.

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/derekparker/delve/proc"
 	"github.com/derekparker/delve/service/api"
@@ -25,6 +26,7 @@ import (
 // lower lever packages such as proc.
 type Debugger struct {
 	config  *Config
+	processMutex sync.Mutex
 	process *proc.Process
 }
 
@@ -76,6 +78,13 @@ func (d *Debugger) ProcessPid() int {
 // If `kill` is true we will kill the process after
 // detaching.
 func (d *Debugger) Detach(kill bool) error {
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
+	return d.detach(kill)
+}
+
+func (d *Debugger) detach(kill bool) error {
 	if d.config.AttachPid != 0 {
 		return d.process.Detach(kill)
 	}
@@ -85,6 +94,9 @@ func (d *Debugger) Detach(kill bool) error {
 // Restart will restart the target process, first killing
 // and then exec'ing it again.
 func (d *Debugger) Restart() error {
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
 	if !d.process.Exited() {
 		if d.process.Running() {
 			d.process.Halt()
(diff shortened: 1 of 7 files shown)


proc/proc.go
```diff
@@ -46,6 +46,7 @@ type Process struct {
 	packageMap map[string]string
 
 	allGCache               []*G
+	dwarfLock               sync.Mutex
```
derekparker: I think to make this more comprehensive we should instead put this lock on the dwarf reader in `dwarf/reader.go`, and instead of storing the raw `*dwarf.Data` on the `Process` we should just have the reader. That way this lock/unlock logic can all live in the `dwarf/reader` package instead of the caller having to worry about it.

aarzilli: We create readers in multiple places in proc and the reader itself isn't thread safe. How do you want to do this?

derekparker: Ideally we create a reader on init of the `Process` struct after we've parsed all the dwarf info we need. After that, any area in `proc` that was creating a new reader should just use the one stored in the `Process` struct. And then, any reader method will have a `lock/defer unlock` step before any work is done. Maybe we also have a method on `Process` like `GetDwarfReader` or something as a convenience method that will `Seek` the reader so caller doesn't have to worry about it?

aarzilli: Why not add a `Process`-wide lock and let `Debugger` acquire and release it? That would solve this problem and also make the API thread safe: no function call works while `Continue` is running anyway, besides `RequestManualStop`, right?

derekparker: That's a good idea, I'm up for that. I don't have a strong opinion on the design of that (Process acquires its own lock vs Debugger acquired a lock when it needs to use Process) so I'll leave that up to you.

service/debugger/debugger.go
```diff
@@ -347,15 +406,12 @@ func (d *Debugger) Command(command *api.DebuggerCommand) (*api.DebuggerState, er
 		log.Printf("switching to goroutine %d", command.GoroutineID)
 		err = d.process.SwitchGoroutine(command.GoroutineID)
 	case api.Halt:
-		// RequestManualStop does not invoke any ptrace syscalls, so it's safe to
-		// access the process directly.
-		log.Print("halting")
-		err = d.process.RequestManualStop()
+		// RequestManualStop alread called
```
derekparker: s/alread/already/

derekparker:
The code LGTM, but I'd like to wait on feedback from @dlsniper to ensure this tackles the problem first.

dlsniper:
I can confirm this fixes the issue both from command line and from the IDE.
Thank you for all your hard work on fixing this!

derekparker:
Merged.

PR #424, opened 2016-02-16
dlv: Take package name in certain commands

### LINKED ISSUE #423: Running debugger requires the working directory to be the same as the package / file directory
Hi,

It seems that delve needs the working directory to be same as the directory where the package / file is located in otherwise it produces the following error:
```
could not launch process: fork/exec ./<binary>: no such file or directory
```

derekparker:
This patch allows certain commands (`debug`, `test`, and `trace`) to
take an argument specifying a package name. This allows a user to
specify a package other than that of the current package in the working
directory.

If a package name is not specified, the behavior remains the same, as in
Delve will look in the current working directory for a `main` package to
compile.

Fixes #423

### DIFF
--- a/cmd/dlv/main.go
+++ b/cmd/dlv/main.go
@@ -21,10 +21,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const version string = "0.11.0-alpha"
-
-// Build is the current git hash.
-var Build string
+const (
+	version       = "0.11.0-alpha"
+	debugname     = "debug"
+	testdebugname = "debug.test"
+)
 
 var (
 	// Log is whether to log debug statements.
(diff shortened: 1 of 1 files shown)


cmd/dlv/main.go
```diff
@@ -210,17 +211,28 @@ func traceCmd(cmd *cobra.Command, args []string) {
...
+				if cmd.ArgsLenAtDash() != 1 {
+					pkg = args[0]
+				}
+				regexp = args[1]
+			}
+			if err := gobuild(debugname, pkg); err != nil {
 				return 1
 			}
-			defer os.Remove(fp)
+			defer os.Remove("./" + debugname)
```
dlsniper: I think there will still be a problem with this one. Wouldn't it work better if for example all builds would default to `/tmp` and then use that binary from there with specifying the working directory when executing?

aarzilli:
Also seconding @dlsniper comment.

cmd/dlv/main.go
derekparker: Yeah I agree with that, however I think it's a different issue, different PR. I will send another patch after this lands.

derekparker:
@aarzilli done.

PR #429, opened 2016-02-18
_fixtures: Removed testvariables2 and testvariables4

aarzilli:
Moved the code in testvariables2 and testvariables4 into
testvariables3, renamed testvariables3 into testvariables2

Just some housekeeping.


derekparker:
Merged.

PR #431, opened 2016-02-19
docs: Move wiki docs into Documentation dir

derekparker:
Going forward, all documentation should be placed in the Documentation
directory in the root of the project. This switch allows maintainers to
approve updates to documentation before they are committed, as opposed
to the pre-existing wiki which anybody could modify.

Currently the Documentation directory includes docs on building, usage,
and minimal docs around the API. This is just the initial commit, and
documentation will continue to improve over time.

Some changes have been made (and will continue to be made) to `cmd/dlv` to
ensure we can auto-generate documentation for all commands from the
newly provided script `scripts/gen-usage-docs.go`, which can be invoked
via `go run scripts/gen-usage-docs.go`.

Additionally, version has been split into its own package. This was a
bit of housekeeping related to the changes made the `cmd/dlv`.


aarzilli:
When the docs in the wiki are moved over we should either (1) kill the wiki and link to the documentation from README.md, or (2) write a script to copy the documentation over to the wiki (I think there is a way to set the wiki to contributors only).

PR #432, opened 2016-02-21
proc: Add filter for non-variable global symbols to PackageVariables

aarzilli:
proc.(*EvalScope).PackageVariables shouldn't display unreadable global symbols that are not package variables (as indicated by their name).

### DIFF
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1556,3 +1556,24 @@ func TestIssue396(t *testing.T) {
 		assertNoError(err, t, "FindFunctionLocation()")
 	})
 }
+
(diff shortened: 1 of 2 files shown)


derekparker:
If I'm not mistaken, we should be able to get rid of `packageNonVariable` and simply check if the dwarf type of the symbol is `DW_TAG_unspecified_type` and the `DW_AT_NAME` of that type is `<unspecified>`. That way we don't have to hard code any filtering.

aarzilli:
changed.

derekparker:
Rebased and merged.

PR #433, opened 2016-02-22
New command 'types'

aarzilli:
Lists all the types defined in the debugged program.

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -656,6 +656,25 @@ func (dbp *Process) Funcs() []gosym.Func {
 	return dbp.goSymTable.Funcs
 }
 
+// Types returns list of types present in the debugged program.
+func (dbp *Process) Types() ([]string, error) {
+	reader := dbp.DwarfReader()
+	types := []string{}
+	seen := map[string]struct{}{}
+	for entry, err := reader.NextType(); entry != nil; entry, err = reader.NextType() {
+		if err != nil {
+			return nil, err
+		}
+		if n, ok := entry.Val(dwarf.AttrName).(string); ok {
+			if _, isseen := seen[n]; !isseen {
+				seen[n] = struct{}{}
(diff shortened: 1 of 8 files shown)


proc/proc.go
```diff
@@ -656,6 +656,25 @@ func (dbp *Process) Funcs() []gosym.Func {
 	return dbp.goSymTable.Funcs
 }
 
+// Types returns list of types present in the debugged program
```
derekparker: nit: End with period.

service/debugger/debugger.go
```diff
@@ -436,6 +436,27 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
 	return regexFilterFuncs(filter, d.process.Funcs())
 }
 
+func (d *Debugger) Types(filter string) ([]string, error) {
```
derekparker: Note: This will have to grab the mutex once https://github.com/derekparker/delve/pull/421 lands.

service/test/integration_test.go
```diff
@@ -1070,3 +1070,12 @@ func TestSkipPrologue2(t *testing.T) {
...
 	})
 }
+
+func TestTypesCommand(t *testing.T) {
+	withTestClient("testvariables2", t, func(c service.Client) {
+		state := <-c.Continue()
+		assertNoError(state.Err, t, "Continue()")
+		_, err := c.ListTypes("")
+		assertNoError(err, t, "ListTypes()")
+	})
```
derekparker: Let's assert about the filter working correctly here, possibly by filtering to a custom type and asserting only it comes back.

service/debugger/debugger.go
aarzilli: Indeed, thanks for the reminder.

derekparker:
Merged -- modified the test a bit to assert the filter a bit stronger.

Issue #426, opened 2016-02-17
Print / get the value of unsafe.Pointer


dlsniper:
Hi,

In the following program:
```go
package main

import (
    "fmt"
    "github.com/go-ozzo/ozzo-config"
)

func main() {
    // create a Config object
    c := config.New()

    // load configuration from a JSON string
    c.LoadJSON([]byte(`{
        "Version": "2.0",
        "Author": {
            "Name": "Foo",
            "Email": "bar@example.com"
        }
    }`))

    // get the "Version" value, return "1.0" if it doesn't exist in the config
    version := c.GetString("Version", "1.0")

    var author struct {
        Name, Email string
    }
    // populate the author object from the "Author" configuration
    c.Configure(&author, "Author")

    fmt.Println(version) // set breakpoint here
    fmt.Println(author.Name)
    fmt.Println(author.Email)
    // Output:
    // 2.0
    // Foo
    // bar@example.com
}
```

it is not possible to get the value of `c.data.ptr` as `ptr` is an `unsafe.Pointer` type.
Would it be possible, for a not so near release of delve, to have delve evaluate the pointer and return the information for it?

(related issue: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2347, cc @kolkov)

Thank you.

aarzilli:
Unsafe pointers do not have type information so we can not print nothing more than their address. If the user happens to know what the unsafe.Pointer happens to point to a type cast can be used

```
print *(*someType)(0x...)
```

if we are unable to resolve the type expression into a type name double quotes can be used:

```
print *(*"map[string]interface {}")(0x...)
```

introduced in a7a0cc75e128c696af0d90a811cd7235339074db.

dlsniper:
I wonder if we (the delve "consumers") can automate that given we should know information about the type of pointers. What would be the API call in this case? `Eval`?

aarzilli:
Yes, `Eval`. But in general you don't know the type either, unsafe.Pointer could contain an entirely arbitrary value.

dlsniper:
Fair enough, I haven't read how that lib works until now. Thanks!

PR #434, opened 2016-02-22
Better support for type casts to anonymous types

aarzilli:
This set of commits was prompted by Issue #426. Obviously we can't print the contents of an unsafe.Pointer (because we don't know what it contains), however one should be able to cast it to some other type and read it that way, in that case: `(*map[string]interface{})(0x...)`.

However that doesn't work because of the way `go/printer` formats the expression `map[string]interface{}`: `go/printer` doesn't put a space between `interface` and `{}`, but in debug_info the type appears as `map[string]interface {}`.

I took a look at other anonymous types and found other discrepancies. The first commit attempts to fix these discrepancies in the most common circumstances. It's not a perfect solution, it would probably be better to abandon `go/printer` and just write a serialization that matches the one used in debug_info but I couldn't find the part of the compiler that does this and I'd rather not guess.

The second commit introduces an alternate way to specify the type of a type casts, by putting the type in quotes, like: `(*"map[string]interface {}")(0x...)`, the type `map[string]interface {}` will be searched in `debug_info`, verbatim. This is also not particularly elegant but it can serve as a last resort (along with the `types` command in PR #433) for the user if we get the parsing/formatting wrong. I think lldb also takes this approach.

### DIFF
--- a/proc/eval.go
+++ b/proc/eval.go
@@ -41,7 +41,13 @@ func (scope *EvalScope) evalAST(t ast.Expr) (*Variable, error) {
 			if err == nil {
 				return v, nil
 			}
-			if err != reader.TypeNotFoundErr {
+			_, isident := node.Fun.(*ast.Ident)
+			// we don't support function calls at the moment except for a few
+			// builtin functions so just return the type error here if the function
+			// isn't an identifier.
+			// More sophisticated logic will be required when function calls
+			// are implemented.
+			if err != reader.TypeNotFoundErr || !isident {
 				return v, err
 			}
 		}
--- a/proc/types.go
+++ b/proc/types.go
@@ -3,8 +3,10 @@ package proc
 import (
 	"github.com/derekparker/delve/dwarf/reader"
 	"go/ast"
+	"go/token"
 	"golang.org/x/debug/dwarf"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 )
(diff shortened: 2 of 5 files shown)


derekparker:
@aarzilli just a heads up the DWARF output is going to get a lot of attention from myself and some other folks soon (see https://groups.google.com/forum/#!topic/golang-dev/TBApivWRGsk).

aarzilli:
Even if things are going to improve on the compiler side, I'd still merge the second commit (c18117d) since it's also helpful if we can't get the package name expansion right.

derekparker:
Thinking about this some more, I think we should abandon commit 1, but commit 2 seems reasonable. However, what are your thoughts on printing a message if a user tries to cast to a type we cannot reconcile with DWARF info. In other words, if we can't find it, print a message like (roughly) "could not find type %s, use the 'types' command for valid types".

derekparker:
(To expand) that way they can use the quote feature to type in the exact type they were trying to coerce to.

aarzilli:
done

aarzilli:
Rebased.

proc/eval.go
```diff
@@ -38,7 +38,8 @@ func (scope *EvalScope) evalAST(t ast.Expr) (*Variable, error) {
 			if err == nil {
 				return v, nil
 			}
-			if err != reader.TypeNotFoundErr {
+			_, isident := node.Fun.(*ast.Ident)
```
derekparker: Can you add a quick comment explaining this assignment, and the usage in the branch.

proc/types.go
```diff
@@ -24,13 +26,17 @@ func (dbp *Process) pointerTo(typ dwarf.Type) dwarf.Type {
 
 func (dbp *Process) findTypeExpr(expr ast.Expr) (dwarf.Type, error) {
 	dbp.loadPackageMap()
+	if lit, islit := expr.(*ast.BasicLit); islit && lit.Kind == token.STRING {
```
derekparker: Would be helpful to have some documentation around why we expect a string literal here.

aarzilli:
Done.

derekparker:
Needs rebase.

aarzilli:
Rebased.

PR #435, opened 2016-02-23
terminal: Mechanism to handle command prefixes

aarzilli:
Implements extensible mechanism to specify which commands accept
prefixes (goroutine, frame, on) instead of hardcoding them in
a switch.

Implements #240 

In the process of testing this I also found and fixed a bug in `Debugger`.


derekparker:
Needs rebase.

aarzilli:
rebased.

PR #444, opened 2016-03-05
Improvements to Variable printing

aarzilli:
Most of these changes were proposed on the mailing list.

[dlv output before the changes](https://gist.githubusercontent.com/aarzilli/8a78798bb359d2aa5c08/raw/b88f679252b91437615526bd6d50f832c35fbe92/before)
[dlv output after the changes](https://gist.githubusercontent.com/aarzilli/3503eff5b25952befff9/raw/5c5296c0a02263f8973d20b5da1447bcaf100b57/after)

Commit messages for the changeset:

service/api: Removed unused fields of service/api.Function
proc/eval: Set return variable name to input expression
all: fine-grained control of loadValue for better variable printing
Makes proc.(*Variable).loadValue loading parameters configurable
through one extra argument of type LoadConfig.
This interface is also exposed through the API so clients can control
how much of a variable delve should read.

BREAKS API COMPATIBILITY:

- CreateBreakpoint, AmendBreakpoint
    Breakpoint struct has two new members LoadLocals and LoadArgs that
    control if and how delve should load local variables and arguments
    when the breakpoint is hit
- State objects for breakpoints no longer automatically fill
    BreakpointInfo.Arguments
- Stacktrace (second argument is a LoadConfig not a boolean)
    Second argument is a *LoadConfig not a boolean. Pass nil to not
    load local variables and arguments.
- ListPackageVars, ListLocalVars, ListFunctionArgs, EvalSymbol
    All gain one extra LoadConfig argument.

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -40,6 +40,10 @@ type B struct {
 	ptr *A
 }
 
+type D struct {
+	u1, u2, u3, u4, u5, u6 uint32
+}
+
 func afunc(x int) int {
 	return x + 2
 }
@@ -181,6 +185,7 @@ func main() {
 	aas[0].aas = aas
 	b := B{A: A{-314}, C: &C{"hello"}, a: A{42}, ptr: &A{1337}}
 	b2 := B{A: A{42}, a: A{47}}
+	var sd D
 
 	for i := range benchparr {
 		benchparr[i] = &benchstruct{}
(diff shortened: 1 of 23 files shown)


derekparker:
cc: @lukehoban @dlsniper @ignatov - ping to notify about API changes.

lukehoban:
Re: the breaking API change.  

At least for the VS Code extension, we don't control what version of Delve the user has installed.  So in principal, we have to support working with any build of Delve.  That of course makes these breaking changes more challenging for us.  Now - the right fix is likely to distribute our own fixed version of Delve, but especially due the signing requirement on OS/X, that's not something we're likely to do soon.

With breaking API changes being reasonably infrequent, this is still manageable.

But it would help if there was some version exposed on the Delve server API that tools could use to make decisions about how to consume the API.  

Without that, I think we're going to have to resort to testing whether certain API calls fail, and if so, trying again with different arguments.

Similarly, is there any way to make the need for breaking API changes less common?  For example, making sure that all to the server API functions use a FooArguments struct which can then grow if additional fields are needed?  I believe that layer of indirection would ensure that adding additional arguments to the APIs (which is all that is this PR is doing as far as I can tell, at least from an API shape perspective), would not break the API.

aarzilli:
>But it would help if there was some version exposed on the Delve server API that tools could use to make decisions about how to consume the API.

Not in the API but you could read the output of `dlv version`. Wouldn't work with a remote instance but that's probably an infrequent use case.

>Similarly, is there any way to make the need for breaking API changes less common? For example, making sure that all to the server API functions use a FooArguments struct which can then grow if additional fields are needed?

+1 for this.

ignatov:
Agree with @lukehoban, but we haven't such issues in IntelliJ. We use the bundled Delve version inside the plugin.

derekparker:
@lukehoban definitely agreed. I think we need to take steps towards that now, before a 1.0 release. After 1.0 release, the API will be considered completely stable.

derekparker:
@aarzilli can you please rebase.

aarzilli:
I'm waiting on #460

aarzilli:
Rebased on #460.

aarzilli:
Rebased.

derekparker:
Now that #460 has landed I will be taking another look at this tomorrow (2016/4/21) and will review and move towards merging.

PR #447, opened 2016-03-06
terminal: Auto-advance to main.main

### LINKED ISSUE #374: Explore better starting points for debugged process
Currently, we stop the process as soon as it is `exec`ed, so it really isn't anywhere useful. The reason we do not continue directly to `main.main` is because that would skip package init functions, which users may want to debug. Possible solutions include:

1) Continue directly to `main.main`, but optionally provide a flag to prevent this
2) Continue directly to `runtime.main` which would not be user code, but should (iirc) be before any package initialization, and would provide useful feedback when executing commands in the debugger.
3) Do nothing, and either print a startup message telling user to set a breakpoint, or otherwise consider it a documentation problem instead of a software one.

I'm open to suggestions, I want whatever ultimately makes the user experience better.

aarzilli:
Makes the command line UI set a breakpoint on main.main and
automatically call continue on startup, after executing the init file,
as suggested by @jussij.

Fixes #374

### DIFF
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/derekparker/delve/config"
 	"github.com/derekparker/delve/service"
+	"github.com/derekparker/delve/service/api"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


derekparker:
@aarzilli hmm, we should also provide a flag to opt out of this feature.

aarzilli:
IMHO it's not necessary, if one wants to set other breakpoints they can with the init file. I'll add it if you insist.

aarzilli:
Rebased.

derekparker:
I'm going to close this for now as it seems too opinionated IMO. We should probably wait for improved DWARF info to land which will specify the proper user starting point. Until then, users will just have to set breakpoints.

I am open to further discussion, however, if more people want this now.

PR #448, opened 2016-03-06
terminal,service: auto-continue during next and step

### LINKED ISSUE #387: proc: next while nexting
Running into an issue, testing out 'next'ing with multiple goroutines, and eventually I end up on the wrong goroutine, and when I try and 'next' again, I get the error: 'next while nexting'.

I have been able to repro on my OSX machine and a Linux VM.

Example code: 

```go

func dostuff(id int, wg *sync.WaitGroup) {
	fmt.Println("goroutine:", id)
	fmt.Println("goroutine:", id)
	wg.Done()
}

func main() {
	runtime.Breakpoint()

	var wg sync.WaitGroup
	wg.Add(10)
	for i := 0; i < 10; i++ {
		go dostuff(i, &wg)
	}
	wg.Wait()
}
```

Set a breakpoint on `dostuff` and then print id, next, print id, next and usually the third time it happens.

cc: @aarzilli - might not have time to track this down until later, but we need to get this fixed :).

aarzilli:
Make dlv call continue automatically when a breakpoint is hit on a
different goroutine during a next or step operation.
Added API hooks to implement the other solution to this problem (cancel
the next/step operation if a different breakpoint is hit).

Fixes #387

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -163,6 +163,13 @@ func (d *Debugger) state() (*api.DebuggerState, error) {
 		}
 	}
 
+	for _, bp := range d.process.Breakpoints {
+		if bp.Temp {
+			state.NextInProgress = true
+			break
+		}
+	}
+
 	return state, nil
 }
 
@@ -241,6 +248,10 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 		return err
 	}
 	return copyBreakpointInfo(original, amend)
+}
+
+func (d *Debugger) CancelNext() error {
+	return d.process.ClearTempBreakpoints()
 }
 
 func copyBreakpointInfo(bp *proc.Breakpoint, requested *api.Breakpoint) (err error) {
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -449,14 +449,35 @@ func cont(t *Term, ctx callContext, args string) error {
 	return nil
 }
 
+func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string) error {
+	if !state.NextInProgress {
+		printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
+		return nil
+	}
+	for {
+		stateChan := t.client.Continue()
+		var state *api.DebuggerState
+		for state = range stateChan {
+			if state.Err != nil {
+				return state.Err
+			}
+			printcontext(t, state)
+		}
+		if !state.NextInProgress {
+			printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
+			return nil
+		}
+		fmt.Printf("\tbreakpoint hit during %s, continuing...\n", op)
+	}
+}
+
 func step(t *Term, ctx callContext, args string) error {
 	state, err := t.client.Step()
 	if err != nil {
 		return err
 	}
 	printcontext(t, state)
-	printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
-	return nil
+	return continueUntilCompleteNext(t, state, "step")
 }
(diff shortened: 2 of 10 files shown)


aarzilli:
looks like there's a bug

aarzilli:
fixed it

derekparker:
I'll be taking a closer look and testing this out either today or tomorrow. Sorry for the delay.

aarzilli:
rebased.

service/debugger/debugger.go
```diff
@@ -163,6 +163,12 @@ func (d *Debugger) state() (*api.DebuggerState, error) {
 		}
 	}
 
+	for _, bp := range d.process.Breakpoints {
+		if bp.Temp {
+			state.NextInProgress = true
```
derekparker: Break loop when this condition is met.

service/debugger/debugger.go
```diff
@@ -243,6 +249,10 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 	return copyBreakpointInfo(original, amend)
 }
 
+func (d *Debugger) CancelNext() {
+	d.process.ClearTempBreakpoints()
```
derekparker: Propagate error here.

terminal/command.go
```diff
@@ -443,13 +443,35 @@ func cont(t *Term, ctx callContext, args string) error {
...
+		stateChan := t.client.Continue()
+		var state *api.DebuggerState
+		for state = range stateChan {
+			if state.Err != nil {
+				return state.Err
+			}
+			printcontext(t, state)
+		}
+		if state.NextInProgress {
+			fmt.Printf("\tbreakpoint hit during %s, continuing...\n", op)
```
derekparker: Could be nice to give some details on the breakpoint that we're skipping.

terminal/command.go
```diff
@@ -443,13 +443,35 @@ func cont(t *Term, ctx callContext, args string) error {
...
+	for {
+		stateChan := t.client.Continue()
+		var state *api.DebuggerState
+		for state = range stateChan {
+			if state.Err != nil {
+				return state.Err
+			}
+			printcontext(t, state)
+		}
+		if state.NextInProgress {
```
derekparker: nit: maybe refactor to:

```go
if !state.NextInProgress {
        break
}
fmt.Printf("...")
```

terminal/command.go
```diff
@@ -443,13 +443,35 @@ func cont(t *Term, ctx callContext, args string) error {
...
+}
+
 func step(t *Term, ctx callContext, args string) error {
 	state, err := t.client.Step()
 	if err != nil {
 		return err
 	}
 	printcontext(t, state)
-	return nil
+	if !state.NextInProgress {
```
derekparker: nit: would also recommend refactor to:

```go
if state.NextInProgress {
        return continueUntilCompleteNext(t, "step")
}
return nil
```

terminal/command.go
```diff
@@ -467,7 +489,10 @@ func next(t *Term, ctx callContext, args string) error {
...
```
derekparker: Same refactor nit here.

derekparker:
I like giving an option to the caller of whether they want to acknowledge the breakpoint or not.

terminal/command.go
aarzilli: The call to `printcontext(t, state)` will do that.

aarzilli:
I should have addressed everything.

aarzilli:
Rebased, moved `CancelNext` to rpc2 exclusively.

derekparker:
Needs rebase.

aarzilli:
Rebased.

Issue #317, opened 2015-12-16
Break on panic


derekparker:
There should be a continue mode that will break on panic, or maybe the default behavior should be to stop on a panic and let the user inspect the state of the program.

aarzilli:
Implemented by c4797ea44506e0501376bd7e99ec9f35ffdd0769

PR #449, opened 2016-03-06
proc: Breakpoint to catch unrecovered panics

aarzilli:
Automatically sets a breakpoint on runtime.startpanic, the function
that gets called by runtime.dopanic when a panic is not recovered.

Implements #317

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -114,6 +114,9 @@ func (d *Debugger) Restart() error {
 		return fmt.Errorf("could not launch process: %s", err)
 	}
 	for _, oldBp := range d.breakpoints() {
+		if oldBp.ID < 0 {
+			continue
+		}
 		newBp, err := p.SetBreakpoint(oldBp.Addr)
 		if err != nil {
 			return err
--- a/_fixtures/panic.go
+++ b/_fixtures/panic.go
@@ -1,0 +1,5 @@
+package main
+
+func main() {
+	panic("BOOM!")
+}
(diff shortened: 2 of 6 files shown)


service/debugger/debugger.go
```diff
@@ -114,6 +114,9 @@ func (d *Debugger) Restart() error {
 		return fmt.Errorf("could not launch process: %s", err)
 	}
 	for _, oldBp := range d.breakpoints() {
+		if oldBp.ID < 0 {
```
derekparker: Why not set the panic breakpoint on restart?

aarzilli: Because recreating the `Process` already sets the breakpoint, setting it here would cause an error, which would have to be ignored.

derekparker: Right, makes sense.

derekparker:
Thanks! We might want to improve the UI around this at some point, like print a message saying something about "uncaught panic" and maybe print the panic message.

aarzilli:
The name of the breakpoint is "unrecovered-panic", the output on the sample program is:

```
$ ./dlv exec _fixtures/panic
(dlv) c
> [unrecovered-panic] runtime.startpanic() /usr/local/go/src/runtime/panic.go:507 (hits goroutine(1):1 total:1) (PC: 0x4223a0)
   502:		}
   503:		return nil
   504:	}
   505:	
   506:	//go:nosplit
=> 507:	func startpanic() {
   508:		systemstack(startpanic_m)
   509:	}
   510:	
   511:	//go:nosplit
   512:	func dopanic(unused int) {
(dlv) 
```

PR #450, opened 2016-03-09
dlv: Misc refactors

krhubert:


terminal/terminal.go
```diff
@@ -158,10 +154,10 @@ func (t *Term) promptForInput() (string, error) {
 func (t *Term) handleExit() (int, error) {
 	fullHistoryFile, err := config.GetConfigFilePath(historyFile)
 	if err != nil {
-		fmt.Println("Error saving history file:", err)
+		fmt.Println("Error saving history file: ", err)
```
derekparker: `fmt.Println` automatically inserts a space separator in between arguments.

derekparker:
Sorry for the delay, I've been swamped lately. I scanned this and it seems fine, will take a closer look this evening or tomorrow once I'm caught up on a few things. Saw one minor thing which I pointed out, but as I said I mostly just scanned this.

terminal/terminal.go
krhubert: True, so 5 lines belowe is fmt.Println("readline history error: ", err), which has space, I will fix this

derekparker:
Merged, thanks.

derekparker:
@krhubert I also went ahead and took care of the `Println` formatting.

PR #451, opened 2016-03-10
terminal: do not use escape codes on windows unless they are supported

### LINKED ISSUE #436: delve prints ←[34m in Windows command window
CC @lukehoban

To reproduce do this:
```
C:\dev\src\t>systeminfo | findstr /B /C:"OS Name" /C:"OS Version"
OS Name:                   Microsoft Windows 7 Professional
OS Version:                6.1.7601 Service Pack 1 Build 7601

C:\dev\src\t>go version
go version devel +a4b8339 Mon Feb 22 21:15:50 2016 +0000 windows/amd64

C:\dev\src\t>dlv version
Delve Debugger
Version: 0.11.0-alpha
Build:

C:\dev\src\t>type main.go
package main

import (
        "fmt"
)

func main() {
        for i := 0; i < 10; i++ {
                fmt.Printf("i=%d\n", i)
        }
}

C:\dev\src\t>dlv debug
(dlv) b main.main
Breakpoint 1 set at 0x40105f for main.main() c:/dev/src/t/main.go:7
(dlv) c
> main.main() c:/dev/src/t/main.go:7 (hits goroutine(1):1 total:1) (PC: 0x40105f)
←[34m     2:    ←[0m
←[34m     3:    ←[0mimport (
←[34m     4:    ←[0m    "fmt"
←[34m     5:    ←[0m)
←[34m     6:    ←[0m
←[34m=>   7:    ←[0mfunc main() {
←[34m     8:    ←[0m    for i := 0; i < 10; i++ {
←[34m     9:    ←[0m            fmt.Printf("i=%d\n", i)
←[34m    10:    ←[0m    }
←[34m    11:    ←[0m}
(dlv) q

C:\dev\src\t>
```

I see that delve determines wether to print escape codes or not by checking `strings.ToLower(os.Getenv("TERM")) == "dumb"`. Perhaps it is OK on non Windows OS. But Windows command line output doesn't do escape codes. Windows also normally does not set TERM environment variable. I propose to change windows code to assume terminal is "dumb" also if TERM environment variable is not found or empty. What do you think? Alternatives? Should we change this code for every OS?

Alex

alexbrainman:
Fixes #436

@lukehoban for your review. Thank you.

### DIFF
--- a/terminal/terminal_other.go
+++ b/terminal/terminal_other.go
@@ -1,0 +1,13 @@
+// +build !windows
+
+package terminal
+
+import (
+	"os"
+	"strings"
+)
+
+// supportsEscapeCodes returns true if console handles escape codes.
+func supportsEscapeCodes() bool {
+	return strings.ToLower(os.Getenv("TERM")) != "dumb"
+}
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -42,7 +42,7 @@ func New(client service.Client, conf *config.Config) *Term {
 		line:   liner.NewLiner(),
 		client: client,
 		cmds:   cmds,
-		dumb:   strings.ToLower(os.Getenv("TERM")) == "dumb",
+		dumb:   !supportsEscapeCodes(),
 	}
 }
 
--- a/terminal/terminal_windows.go
+++ b/terminal/terminal_windows.go
@@ -1,0 +1,31 @@
+package terminal
+
+import (
+	"os"
+	"strings"
+	"syscall"
+)
+
+// supportsEscapeCodes returns true if console handles escape codes.
+func supportsEscapeCodes() bool {
+	if strings.ToLower(os.Getenv("TERM")) == "dumb" {
+		return false
+	}
+	if strings.ToLower(os.Getenv("ConEmuANSI")) == "on" {
+		// The ConEmu terminal is installed. Use it.
+		return true
+	}
+
+	const ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+
+	h, err := syscall.GetStdHandle(syscall.STD_OUTPUT_HANDLE)
+	if err != nil {
+		return false
+	}
+	var m uint32
+	err = syscall.GetConsoleMode(h, &m)
+	if err != nil {
+		return false
+	}
+	return m&ENABLE_VIRTUAL_TERMINAL_PROCESSING != 0
+}


aarzilli:
lgtm

terminal/terminal_other.go
```diff
@@ -0,0 +1,13 @@
+// +build !windows
```
derekparker: This file should be `terminal_unix.go` and then you can omit the build tag.

alexbrainman: @derekparker I tried your suggestion, but it does not work:

```
$ git status
On branch fix_console_output
Your branch is up-to-date with 'origin/fix_console_output'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    terminal_other.go -> terminal_unix.go

$ git diff HEAD
diff --git a/terminal/terminal_other.go b/terminal/terminal_other.go
deleted file mode 100644
index 6287e06..0000000
--- a/terminal/terminal_other.go
+++ /dev/null
@@ -1,13 +0,0 @@
-// +build !windows
-
-package terminal
-
-import (
-       "os"
-       "strings"
-)
-
-// supportsEscapeCodes returns true if console handles escape codes.
-func supportsEscapeCodes() bool {
-       return strings.ToLower(os.Getenv("TERM")) != "dumb"
-}
diff --git a/terminal/terminal_unix.go b/terminal/terminal_unix.go
new file mode 100644
index 0000000..a78986f
--- /dev/null
+++ b/terminal/terminal_unix.go
@@ -0,0 +1,11 @@
+package terminal
+
+import (
+       "os"
+       "strings"
+)
+
+// supportsEscapeCodes returns true if console handles escape codes.
+func supportsEscapeCodes() bool {
+       return strings.ToLower(os.Getenv("TERM")) != "dumb"
+}
$ GOOS=windows go build -o /dev/null
# github.com/derekparker/delve/terminal
./terminal_windows.go:10: supportsEscapeCodes redeclared in this block
        previous declaration at ./terminal_unix.go:9
a@spot: terminal $ git reset --hard
HEAD is now at 3fdfa4e terminal: do not use escape codes on windows unless they are supported
$ git status
On branch fix_console_output
Your branch is up-to-date with 'origin/fix_console_output'.

nothing to commit, working directory clean
$ GOOS=windows go build -o /dev/null
$
```

I don't think *_unix.go filename has any special meaning, like _windows.go or _linux.go. I also looked for similar files in GOROOT:
```
$ cat $GOROOT/src/os/file_unix.go | grep build
// +build darwin dragonfly freebsd linux nacl netbsd openbsd solaris
$
```
Where does it say that _unix.go means something?

Also there is plan9 version of Go. :-)

Ale

derekparker: You're right, sorry about that! Must've seen that pattern before and mistook its significance.

joegrasse: You are right, *_unix.go doesn't have any special meaning, but *_linux.go does. See [Build Constraints](https://golang.org/pkg/go/build/). Also, for examples within the delve code look in the /proc dir. 

Instead of ```terminal_unix.go```, it should be ```terminal_linux.go```.

derekparker: @joegrasse it was just a mistake on my end, I had remembered (incorrectly) that `*_unix.go` was a valid build constraint just like `_linux` or `_darwin`, etc, just a shorthand basically. I think I had just seen it before and assumed it behaved as I thought, definitely my bad.

derekparker:
LGTM -- @lukehoban any thoughts on this?

terminal/terminal_other.go
joegrasse: @derekparker, Yeah, I was just pointing out that it didn't work for @alexbrainman, because the file should be ```terminal_linux.go``` instead of ```terminal_other.go```. That is the convention used elsewhere instead of ```// +build !windows```.

joegrasse: I see now though that this is windows vs all other.

lukehoban:
Yep - this looks good.  

I verified the behaviour on Windows 10 where the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` mode is supported per https://github.com/derekparker/delve/issues/436#issuecomment-194578983.

aarzilli:
Merged as ff0ec8ce00fbac7f84859cf420b31bc3a513e43c

Issue #302, opened 2015-11-25
running, locked to thread with hashmap


nati:
I'm trying to debug https://github.com/cloudwan/gohan using "dlv debug".
When I start debbuging I got this error

``` shell
2015/11/25 11:30:46 debugger.go:56: launching process with args: [./debug]
2015/11/25 11:30:46 server.go:301: method Run has wrong number of ins: 1
2015/11/25 11:30:46 server.go:301: method Stop has wrong number of ins: 2
2015/11/25 11:30:46 debugger.go:231: continuing
unexpected fault address 0x49ebe20
fatal error: fault
[signal 0xb code=0x1 addr=0x49ebe20 pc=0x41172ad]
1 [running, locked to thread]:
runtime.throw(0x49d3a78, 0x5)
    /usr/local/go/src/runtime/panic.go:527 +0x90 fp=0xc8200e9d78 sp=0xc8200e9d60
runtime.sigpanic()
    /usr/local/go/src/runtime/sigpanic_unix.go:27 +0x2ba fp=0xc8200e9dc8 sp=0xc8200e9d78
runtime.aeshashbody()
    /usr/local/go/src/runtime/asm_amd64.s:981 +0x38d fp=0xc8200e9dd0 sp=0xc8200e9dc8
runtime.mapassign1(0x4867220, 0xc8201a4150, 0x4bbcbe0, 0x4bbcbf0)
    /usr/local/go/src/runtime/hashmap.go:424 +0xbf fp=0xc8200e9e78 sp=0xc8200e9dd0
github.com/coreos/go-etcd/etcd.init()
    /Users/nati/go/src/github.com/coreos/go-etcd/etcd/options.go:31 +0x295 fp=0xc8200e9eb0 sp=0xc8200e9e78
github.com/cloudwan/gohan/sync/etcd.init()
    /Users/nati/go/src/github.com/cloudwan/gohan/sync/etcd/log.go:24 +0x59 fp=0xc8200e9ee8 sp=0xc8200e9eb0
github.com/cloudwan/gohan/server.init()
    /Users/nati/go/src/github.com/cloudwan/gohan/server/sync.go:597 +0xc2 fp=0xc8200e9f20 sp=0xc8200e9ee8
github.com/cloudwan/gohan/cli.init()
    /Users/nati/go/src/github.com/cloudwan/gohan/cli/cli.go:345 +0x75 fp=0xc8200e9f28 sp=0xc8200e9f20
main.init()
[...]

PR #452, opened 2016-03-10
proc: bugfix: do not fork from cgo [OS X]

### LINKED ISSUE #294: Can't debug github.com/drone/drone
```
/usr/local/Cellar/go/1.5.1/libexec/bin/go build -o "/private/var/folders/p6/tg8pmz4509z5236gz3lprynh0000gq/T/Build drone.go and run7go" -gcflags "-N -l" /Users/ignatov/src/gocode/src/github.com/drone/drone/drone.go
/Users/ignatov/src/go/build/idea-sandbox/plugins/Go/lib/dlv/mac/dlv --listen=localhost:65233 --headless=true exec "/private/var/folders/p6/tg8pmz4509z5236gz3lprynh0000gq/T/Build drone.go and run7go" --
runtime: function symbol table header: 0x0 0x0 0x0 0x0 0x0
fatal error: invalid function symbol table

runtime: panic before malloc heap initialized

runtime stack:
```

Mac OS X 10.11.1, the latest version of delve

aarzilli:
Forking from cgo causes memory corruption in the target process.

Fixes #294, #302 

This should also solve the problem @evandigby reported on gitter. The downside of this solution is that it exacerbates the Launch race condition we were never able to nail down making tests on OS X far flakier.


derekparker:
Does this patch work? I haven't tested, I'm on Linux atm. Have you tested this with softsignal handling? I ask because it doesn't seem like we'd have set the exception port before the child calls `PT_SIGEXC`, which IIRC caused issues. If all tests on your end are passing locally then we should be fine in that regard, which means we could probably take another approach. The reason this was implemented in cgo in the first place was because of the weird relationship with `PT_SIGEXC` and setting exception ports. If that isn't actually an issue, we could implement this purely in Go.

I think what we're going to want to do here is modify Go stdlib, vendor it for now, and un-vendor once the patch lands. To explain, I think we should modify `forkAndExecInChild`, see $GOROOT/src/syscall/exec_linux.go` for an example. What we really need is a modified `SysProcAttr` that has a field like: `PtraceSigexc` on darwin. And then, if that field is set, just like with the `Ptrace` field on that struct, and it will call that for us.

That way we don't have to manage this "pivot" binary on users machines.

Thoughts?

evandigby:
Come Monday I can test whether this resolves the issue I was having in OSX. If I get a chance earlier than that, I will. Can't speak to the other details here!

Thank you!

aarzilli:
I tested it on drone and it fixed the problem there. The tests randomly fail, in the same way that they fail without this patch but more frequently, there's definitely some problems.

I don't think we can vendor this part of the stdlib, it seems like it needs too much help from the runtime.

derekparker:
@aarzilli you may be right about not being able to vendor it.

evandigby:
`func getitab(inter *interfacetype, typ *_type, canfail bool) *itab {
	if len(inter.mhdr) == 0 {
		throw("internal error - misuse of itab")
	}`

Unfortunately with this pull this is throwing (inter.mhdr == 0)

Is there anything I need to be doing other than fetching the pull request and make install?

aarzilli:
That looks like another problem with the target process not having VM initialized properly, I wonder why. I guess I didn't get it right.

As a workaround you could launch your process outside of delve then attach to it using `dlv attach` (I don't know if VSCode lets you do this), I suppose that should work.

derekparker:
@aarzilli I'm working on a patch to the current CGO fork/exec I'm hoping will fix the race condition issue, and hopefully this issue as well. I had to pause on it for a bit, but will hopefully have a patch up tomorrow. If I end up making that work I'll close out this issue, otherwise we can continue the discussion here.

aarzilli:
Forgot to close this

Issue #209, opened 2015-08-12
Linux 2.6.x kernel: Breakpoints hang, program unresponsive


derekparker:
On the Linux 2.6.x kernel, after setting a breakpoint in the traced process and continuing it, Delve will hang instead of registering the breakpoint. Further inspection shows the traced process is not running.

This could be a ptrace(2) bug on that version of the kernel, or a bug with Delve. Requires further investigation.

derekparker:
Broken out from https://github.com/derekparker/delve/issues/203

joegrasse:
@derekparker I created #453 to fix this issue.

PR #453, opened 2016-03-11
Fix for Issue #209

joegrasse:


derekparker:
Thanks! Would you mind squashing your commits and amending the commit message to match https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message, please.

joegrasse:
Done.

derekparker:
I feel like we should detect the kernel version here. While technically both `t` and `T` are both considered "stop" statuses, they mean different things depending on the kernel version. For example, `T` is "stopped by job control signal" on modern kernels.

@aarzilli any thoughts on this?

aarzilli:
IMHO that would only be useful if we were going to actually write support for job control signals. It seems like something that would take a lot of research for little payoff, I think it would only be relevant when we are attaching to a process running under a shell and the user is trying to control the process with the shell.  
Unless we are going to, there is no point in distinguishing them. I would merge the patch as is.

derekparker:
Merged.

derekparker:
@joegrasse thanks for the patch!

Issue #280, opened 2015-10-19
int and int64 fields in JSON API


kostya-sh:
API types from service/api/types.go have many int, int64 and uint64 fields that are represented as numbers in JSON.

However not all languages support int64 types. In particular Javascript has only IEEE-754 8-byte float type that can only represent 53-bit integers without loss of precision.

I suggest to serialize all numbers as strings. E.g.:

```
ID int `json:"id,string"`
```

aarzilli:
Because of address space restrictions on supported hardware real addresses never have more than 53bits. With the advent of regabi in go1.17 we actually started returning addresses with more than 53bits to give unique addresses to registerized variables. Nobody complained about this issue, so I guess it doesn't matter.

PR #460, opened 2016-03-15
service,terminal: APIv2 plus method to select API version

aarzilli:
New API version with better backwards compatibility plus mechanism to
select the API version that a headless instance should use.

Adds service/test/cmd/typecheckrpc.go to type check the RPC interface.

One downside of this implementation is that attaching to a headless instance of delve works only if the user specified '-api=2', since this use case is less common than a plugin attaching to an instance of delve I think it's acceptable.

I've tested this with VSCode and everything seems to work correctly.

If this is ok I will rebase #444 on top of this so that those changes won't break VSCode.


derekparker:
Thought for discussion:

What if all RPC methods had a specific type for it's argument, e.g.:

```golang
type GoroutineArg struct {
        id int
}
```

That way, anybody who is relying on the RPC interface (currently our only API) will never be broken by arity / function signature changes. We can always add new fields to the struct and preserve old ones in a backwards compatible way.

aarzilli:
I did that already for every API entry point with the exception of `CreateBreakpoint` and  `AmendBreakpoint`. I feel that those are specific enough that passing an api.Breakpoint as argument isn't a problem. The command line switch is there for backwards compatibility with the current API, hopefully it will never need to go above 2 because we can just add new fields to input structs in a backwards compatible way.

derekparker:
@aarzilli indeed you did, I was looking at the client instead of the server, my fault.

service/rpc2/server.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+	out.Registers = regs
+	return nil
+}
+
+type ListLocalVarsIn struct {
+	Scope api.EvalScope
+}
+
+func (s *RPCServer) ListLocalVars(arg ListLocalVarsIn, variables *[]api.Variable) error {
```
derekparker: I think for the sake of consistency we should also wrap this return value in a struct as well.

```go
struct ListLocalVarsOut {
        variables *[]api.Variable
}
```

aarzilli:
Done, I've also changed `ListPackageVars` and `ListFunctionArgs`.

derekparker:
@aarzilli we need to be completely consistent across all RPC calls, so let's just have all RPC functions accept an "In" struct and return an "Out" struct, even if it's single argument/single return value.

aarzilli:
Made it more consistent. I still left out these:

`GetBreakpoint(GetBreakpointIn, *api.Breakpoint)`
`State(StateIn, *api.DebuggerState)`
`GetThread(GetThreadIn, *api.Thread)`
`CreateBreakpoint(CreateBreakpointIn, *api.Breakpoint)`
`Eval(EvalIn, *api.Variable)`

they are all functions that ask for one particular object I don't think it will ever make sense to return anything that doesn't belong to that object.

derekparker:
@aarzilli needs rebase. Also, I feel like we should convert those as well. I'd rather cover all our bases, just ensure backwards compat across the board.

aarzilli:
done

cmd/dlv/cmds/commands.go
```diff
@@ -78,6 +81,7 @@ func New() *cobra.Command {
 	RootCommand.PersistentFlags().BoolVarP(&Log, "log", "", false, "Enable debugging server logging.")
 	RootCommand.PersistentFlags().BoolVarP(&Headless, "headless", "", false, "Run debug server only, in headless mode.")
 	RootCommand.PersistentFlags().BoolVarP(&AcceptMulti, "accept-multiclient", "", false, "Allows a headless server to accept multiple client connections. Note that the server API is not reentrant and clients will have to coordinate")
+	RootCommand.PersistentFlags().IntVar(&ApiVersion, "api", 1, "Selects API version when headless")
```
derekparker: nit: Let's change this to `api-version`, and specify in the help output what version numbers are accepted.

derekparker:
Other than nit, the only other thing is docs need to be regenerated. Just run `./scripts/gen-usage-docs.go`.

aarzilli:
done.

service/test/cmd/typecheckrpc.go
```diff
@@ -0,0 +1,201 @@
+package main
```
derekparker: Is this checking that the new structs match the old arguments, or that the new RPC methods all contain struct args? Haven't fully studied the code, but could use some documentation anyways. Also, is this intended to be run as part of a CI check to ensure consistent RPC method types (in/out structs)?

aarzilli: This checks that calls to the API in `service/rpc2/client.go` (using `rpc2.(*Client).call`) match what's expected by `service/rpc2/server.go`, the compiler doesn't because they are RPC calls.

aarzilli:
Added documentato to typecheckrpc.

dlsniper:
Hi, I've missed this. Where can one read the difference between API versions? Also, does it also contain the changes to make the int64 fields strings so that JS can use them? Thanks for all your hard work on this project!

aarzilli:
> Where can one read the difference between API versions?
@dlsniper: work on documenting this API is in issue #507. You can compare `service/rpc1/server.go` with `service/rpc2/server.go` for the differences. Basically, both the argument and the return value of all method calls are wrapped in a struct type specific to that method call. A few method calls were renamed to reflect their changed semanthics (`EvalVariable` → `Eval`), `PackageVarsForThread` was removed for being redundant.

Version 1 is still available and the default for headless instances.

> Also, does it also contain the changes to make the int64 fields strings so that JS can use them? 

No. To be honest I'm not comfortable burdening every implementation with ints as strings because of `JSON.parse` questionable choice of silently corrupting data.

dlsniper:
@aarzilli thanks for the fast reply. I'll cc @ignatov here but I hope this should be a easier contribution to do.

Regarding the ints json issue, a nicer solution could be to provide the `_string` fields where needed. A good pattern could be to do this at just before marshaling happens, if possible (I should have look at the code there). The issue is that all Electron based editors (Atom, VSCode) might have problems with this even if others don't.

aarzilli:
Things like [json-bigint](https://www.npmjs.com/package/json-bigint) also exist and they are the philosophically correct solution, IMHO. 

The `_str` variant, like twitter does, could be an acceptable solution. However before we commit to that I would like to do some benchmarks on tracing performance, it's already pretty bad and it seems that 30% of that badness happens in the layers above proc (after #444 is merged I'm going to look into it).

kostya-sh:
See also #280.

I think using a better format instead of JSON (such as protobuf) would solve most of the problems, including performance, API schema evolution and int types.

Another idea: delve can provide a separate endpoint for JS clients that does int to string transformation on the request and response JSON objects transparently.

dlsniper:
> better format instead of JSON (such as protobuf)

I don't think that protobuf is a good fit for this. There's no evidence it will be much better for the use-case than the current JSON implementation. And for the issue I'm talking about, the user will have the same problem anyway (Javascript doesn't understand int64 correctly).

> separate endpoint for JS clients 

It just makes the work harder for delve maintainers without a need to do so. An extra field is reasonably easy to add, a host of other endpoints just to convert a value is too much imho.

@derekparker should I open a new issue / mail on the dev-list about this and continue there?

PR #463, opened 2016-03-18
proc: bugfix: Truncate stacktrace when FDE of a frame can not be found

### LINKED ISSUE #462: Delve cannot attach and debug a simple http service
## 1. What version of Delve are you using (`dlv version`)?

```
$ dlv version
Delve Debugger
Version: 0.11.0-alpha
Build:
```

Git revision: ff0ec8ce00fbac7f84859cf420b31bc3a513e43c

## 2. What version of Go are you using? (`go version`)?

```
$ go version
go version go1.6 linux/amd64
```

## 3. What operating system and processor architecture are you using?

```
$ uname -a
Linux vimeo-vm 3.10.0-327.10.1.el7.x86_64 #1 SMP Tue Feb 16 17:03:50 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
$ cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core)
```

## 4. What did you do?

```
$ cat test.go
package main

import (
        "fmt"
        "net/http"
)

func main() {
        http.HandleFunc("/bar", func(w http.ResponseWriter, r *http.Request) {
                fmt.Fprintf(w, "Hello")
        })

        http.ListenAndServe(":5555", nil)
}
$ go build
$ ./test &
[1] 23629
$ dlv attach 23629
(dlv) stack
Command failed: could not find FDE for PC 0x91e800
```

I also tested with `go build -gcflags '-N -l'`.

## 5. What did you expect to see?

A backtrace.

## 6. What did you see instead?

An error.

aarzilli:
Instead of returning an error when FDE of a frame can not be found,
just truncate the stack trace.

Fixes #462

I also did a bit of refactoring of the stacktrace code:

- made GoroutineStacktrace a method of struct G
- made stacktrace a method of StackIterator
- renamed StackIterator to stackIterator
- factored out logic to obtain a stackIterator from a goroutine that's
used by both (*G).Stacktrace and by (*G).UserCurrent

### DIFF
--- a/dwarf/frame/entries.go
+++ b/dwarf/frame/entries.go
@@ -65,6 +65,14 @@ func NewFrameIndex() FrameDescriptionEntries {
 	return make(FrameDescriptionEntries, 0, 1000)
 }
 
+type NoFDEForPCError struct {
+	PC uint64
+}
+
+func (err *NoFDEForPCError) Error() string {
+	return fmt.Sprintf("could not find FDE for PC %#v", err.PC)
+}
+
 // Returns the Frame Description Entry for the given PC.
 func (fdes FrameDescriptionEntries) FDEForPC(pc uint64) (*FrameDescriptionEntry, error) {
 	idx := sort.Search(len(fdes), func(i int) bool {
@@ -77,7 +85,7 @@ func (fdes FrameDescriptionEntries) FDEForPC(pc uint64) (*FrameDescriptionEntry,
 		return true
 	})
 	if idx == len(fdes) {
-		return nil, fmt.Errorf("could not find FDE for PC %#v", pc)
+		return nil, &NoFDEForPCError{pc}
 	}
 	return fdes[idx], nil
 }
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -876,7 +876,7 @@ func (dbp *Process) ConvertEvalScope(gid, frame int) (*EvalScope, error) {
 		out.Thread = g.thread
 	}
 
-	locs, err := dbp.GoroutineStacktrace(g, frame)
+	locs, err := g.Stacktrace(frame)
 	if err != nil {
 		return nil, err
 	}
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -758,7 +758,7 @@ func TestStacktraceGoroutine(t *testing.T) {
 		mainCount := 0
 
 		for i, g := range gs {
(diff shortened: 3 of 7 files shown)


derekparker:
@aarzilli what's the underlying cause of this issue? Why do we encounter PCs we cannot reconcile with an FDE? I haven't dug into this, but is it a bug in our FDE lookup, or in the information generated?

aarzilli:
`runtime.mcall` (which happens to be at the bottom of the stack in the related issue) is apparently one of those functions that never returns. In general I think it's better to show a truncated stack trace rather than a error message when we can't decode a stack frame.

derekparker:
@aarzilli makes sense, but in that case we should print a warning that the stack trace has been truncated. Additionally, I think that `runtime.mcall` should be considered a "top of stack" function, which should signify that we can stop collecting frames.

aarzilli:
Ok, I've changed `proc.(*stackIterator).Next` so that it will return an extra frame with `File` "?" and `Line` -1 for all undecodable frames except when the undecodable frame is the top of the stack (in that case we lose no informations in returning the error directly).
A truncated stack trace will look like this:

```
(dlv) stack
0  0x000000000045d6a9 in runtime.epollwait
   at /usr/local/go/src/runtime/sys_linux_amd64.s:440
1  0x0000000000428acc in runtime.netpoll
   at /usr/local/go/src/runtime/netpoll_epoll.go:67
2  0x00000000004323e3 in runtime.findrunnable
   at /usr/local/go/src/runtime/proc.go:1955
3  0x0000000000432b7a in runtime.schedule
   at /usr/local/go/src/runtime/proc.go:2072
4  0x0000000000432e16 in runtime.park_m
   at /usr/local/go/src/runtime/proc.go:2137
5  0x0000000000459b09 in runtime.mcall
   at /usr/local/go/src/runtime/asm_amd64.s:233
6  0x000000000091e800 in (nil)
   at ?:-1
```

Additionally I have added runtime.mcall to the "top of stack" functoins.

aarzilli:
Found a bug while investigating issue #468 (stacktrace returned 0 stack frames and no errors because PCToLine didn't find anything), I've added a commit to address it.

aarzilli:
I've disabled the test on windows since [there are no stack frames to print](https://ci.appveyor.com/project/derekparker/delve/build/254).

derekparker:
Merged.

PR #466, opened 2016-03-19
Adding support to multi-colon GOPATH in codesigning part of Makefile

wlbr:
go install installs always into the first subpath of GOPATH, so we can simply cut the GOPATH at the first colon and find the dlv executable to be able to sign it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,9 @@ endif
 install: check-cert
 	go install $(BUILD_FLAGS) github.com/derekparker/delve/cmd/dlv
 ifdef DARWIN
-	codesign -s "$(CERT)" $(GOPATH)/bin/dlv
+	#go install install alsways into the first subpath of GOPATH, so a cut at 
+	#the first colon is safe.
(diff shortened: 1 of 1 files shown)


derekparker:
Thanks for the patch. Maybe instead of assuming where the go tool will put the binary, we can do something like `codesign -s "$(CERT)" $(which dlv)` and let the system tell us instead.

wlbr:
Sure, this will work as well, for as long as the go install path will be already included in the OS path.
Nevertheless - the behavior of "go install" to install into the first GOPATH segment is strictly defined and well documented, so it is not an assumption.
Probably the "which"-solution will be more robust to future change of Go, e.g. an extension to the vendor things.

derekparker:
> Sure, this will work as well, for as long as the go install path will be already included in the OS path.

I think it's fine to make that assumption. This is all happening during the "install" target anyway, so it should be assumed that we're installing somewhere in $PATH.

wlbr:
Go for it ;-)
I'll close the merge request.
Regards,
Michael

PR #469, opened 2016-03-22
service: Make Halt work during tracing

aarzilli:
While tracing delve may spend most of its time outside of
proc.(*Process).Continue, which renders
service/rpc/client.(*Client).Halt ineffective.
This commit changes the implementation of
service/rpc/client.(*Client).Halt to make it capable of stopping traces.

### DIFF
--- a/service/rpc1/client.go
+++ b/service/rpc1/client.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/derekparker/delve/service"
 	"github.com/derekparker/delve/service/api"
+	"sync"
 )
 
 // Client is a RPC service.Client.
@@ -15,6 +16,8 @@ type RPCClient struct {
 	addr       string
 	processPid int
 	client     *rpc.Client
+	haltMu     sync.Mutex
+	haltReq    bool
 }
 
 // Ensure the implementation satisfies the interface.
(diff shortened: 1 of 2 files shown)


derekparker:
Can you provide some information on how to reproduce the situation that this resolves? Is this during the `trace` subcommand, or while setting tracepoints within a debug session?

aarzilli:
While setting tracepoints on a debug session. For example, with this progam:

```
package main

import (
        "os"
        "strconv"
)

func ack(n, m int64) int64 {
        for n != 0 {
                if m == 0 {
                        m = 1
                } else {
                        m = ack(n, m-1)
                }
                n = n - 1
        }
        return m + 1
}

func main() {
        if len(os.Args) > 2 {
                ia1, _ := strconv.ParseInt(os.Args[1], 10, 0)
                ia2, _ := strconv.ParseInt(os.Args[2], 10, 0)
                r := ack(ia1, ia2)
                _ = r
                //fmt.Println(ack(ia1, ia2))
        }
}
```

If you do this:

```
$ dlv exec _fixtures/ackermann -- 3 8
(dlv) t ack
(dlv) c
```

then press ^C, it will only work about 50% of the time. The percentage depends on how much data the user requests to load from the breakpoint (or the amount of data that needs to be loaded for the function arguments), if the tracepoint prints a lot of stuff stopping the program can be almost impossible.

derekparker:
Alright, so just to clarify, the problem seems to be: for `halt` we are sending `SIGSTOP` to the child process, and we expect to see that in the normal flow and return control back to the user. However, in the case of tracing, we could send the signal, but the child is already stopped and we're simply reading data and so the halt request is essentially ignored. Is that correct?

Also, when testing this patch, halt requests sometimes do not show the current location in source when the Delve prompt returns.

aarzilli:
> we could send the signal, but the child is already stopped and we're simply reading data and so the halt request is essentially ignored. Is that correct?

yes.

>Also, when testing this patch, halt requests sometimes do not show the current location in source when the Delve prompt returns.

Mmmh, weird, I did not notice that.

aarzilli:
If the problem was that it didn't call `printfile` sometimes then I have fixed it. I've moved the call to `printfile` out of `printcontext` and into `cont`, `next`, `step` and `stepInstruction`. I think this is more readable, it gets rid of one of the complex if conditions and will make PR #448 better too.

derekparker:
Sorry for delay, will review tonight or tomorrow again.

aarzilli:
Rebased

PR #470, opened 2016-03-22
debugger: Allow disassembly even if the scope can not be converted

aarzilli:
It is sometimes impossible to convert an EvalScope immediately after
starting the program but disassembly can still proceed.


derekparker:
Merged.

PR #475, opened 2016-03-24
use go-colorable

mattn:
I'm on Windows7, but possible to display colors.


aarzilli:
1. Dependencies need to be vendored
2. This will break dlv for people who are using actual `TERM=dumb` terminals on *nix.

mattn:
done

aarzilli:
By using `supportsEscapeCodes` to set `dumb` you are disabling your library in every situation where it's useful (when `supportsEscapeCodes == true` we are already printing colors correctly).

What you probably want to do is set `dumb` using `os.Getenv("TERM") == "dumb"` and then if `!dumb` enable your library only when stdout is a windows console that doesn't support our ANSI escape sequences.

Also you didn't vendor `go-isattty` (see how the build is failing on windows) and I don't think you did the vendoring correctly, I think they need to be added with `godep`, @derekparker will tell you more.

mattn:
Currently Godep/Godep.json seems strange for me because localtests does't exists.
I did following.

```
rm -rf Godeps vendor
godep save ./...
```

derekparker:
@mattn there shouldn't be anything vendored from localtests, what about that is making godeps act up for you?

mattn:
```
$ godep save ./...
godep: Package (github.com/spf13/cobra) not found
```

derekparker:
@mattn (sorry for late response, was away for a few days). You should do `godep restore` first.

mattn:
@derekparker i did it already. but got erorrs.

derekparker:
@mattn Ok, I think I know what the issue is, there _was_ some localtest stuff that was included in Godeps.json by accident. Hopefully https://github.com/derekparker/delve/pull/485 will help get rid of any errors you are seeing.

derekparker:
@mattn can you please rebase on master and try again?

derekparker:
Ping @mattn

mattn:
what you expect command with godep? `godep save ./...`?

derekparker:
@mattn you should be able to run: `godep save $(go list ./... | grep -v vendor)` without getting an error back. Also this PR needs to be rebased.

mattn:
updated.

terminal/terminal.go
```diff
@@ -37,12 +38,21 @@ func New(client service.Client, conf *config.Config) *Term {
 	if conf != nil && conf.Aliases != nil {
 		cmds.Merge(conf.Aliases)
 	}
+
+	dumb := false
```
derekparker: Let's move the dumb terminal check here, and then based on that we can assign `w` to either `os.Stdout` directly (for the `dumb == true` case) or the result of `getColorableWriter()`. That prevents us from having to do that check in multiple places, and in terminal_other we can just always return `os.Stdout`, and on windows we go through the rest of the logic you've defined.

derekparker:
@mattn Aside from the last comment this looks good. Can you also squash your commits and format the messages according to https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message, please.

mattn:
done

derekparker:
Merged, thanks! I added a commit on top of yours to fix doc strings on `getColorableWriter`. You can check it out on master and let me know if the documentation is incorrect.

PR #476, opened 2016-03-25
terminal: show message if there are no vars/locals/args

ooesili:
When I started using delve, and I ran commands to look at variables, I became uneasy when there were no variables at that breakpoint:
```
(dlv) locals
(dlv)
```
```
(dlv) args
(dlv)
```
I couldn't tell if something went wrong or there were actually no locals or args at that point. This pull request makes the `vars`, `locals`, and `args` commands print messages when there is nothing to show:

```
(dlv) locals
(no locals)
(dlv)
```
```
(dlv) args
(no args)
(dlv)
```

This is completely about user experience and doesn't add any real functionality. I didn't add tests for `args` and `vars`, because I would like to get feedback on whether or not this seems like a good idea before investing the time to do so.

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -682,7 +682,7 @@ func args(t *Term, ctx callContext, filter string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	return filterVariables(vars, filter), nil
+	return describeNoVars("args", filterVariables(vars, filter)), nil
 }
 
 func locals(t *Term, ctx callContext, filter string) ([]string, error) {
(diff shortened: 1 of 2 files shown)


derekparker:
I agree, this makes sense from a UI perspective.

derekparker:
@ooesili can you add in some simple tests for the other cases.

derekparker:
@ooesili ping.

ooesili:
Rebased onto master and added tests.

I used an existing fixture program for the test that is not directly related to it. Do you think it would have made more sense to create a new fixture specifically for this test?

PR #479, opened 2016-03-29
Pass commandline to the Windows syscall when creating the process.

moitias:
Currently, Delve will not pass command-line arguments while on Windows and at least with a cursory glance I saw no reference to any other given argument except the filename being handled - so, I'm not sure if it ever worked? 

Anyhow, the fix is really simple, let me know if you'd prefer some other way of handling it.

ref; [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx) on the Windows syscall in question.

Also included are some very basic sanity tests that validate that args are getting passed correctly.

### DIFF
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1689,7 +1689,7 @@ func TestCmdLineArgs(t *testing.T) {
 
 	}, []string{"test", "-passFlag"})
 	withTestProcessArgs("testargs", t, func(p *Process, fixture protest.Fixture) {
-		err := p.Continue()
+		p.Continue()
 		bp := p.CurrentBreakpoint()
 		if bp == nil || bp.Name != "unrecovered-panic" {
 			t.Fatalf("not on unrecovered-panic breakpoint: %v", p.CurrentBreakpoint)
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1689,18 +1689,14 @@ func TestCmdLineArgs(t *testing.T) {
 
 	}, []string{"test", "-passFlag"})
 	withTestProcessArgs("testargs", t, func(p *Process, fixture protest.Fixture) {
-		err = p.Continue()
+		err := p.Continue()
 		bp := p.CurrentBreakpoint()
 		if bp == nil || bp.Name != "unrecovered-panic" {
 			t.Fatalf("not on unrecovered-panic breakpoint: %v", p.CurrentBreakpoint)
 		}
-		exit, exited := err.(ProcessExitedError)
-		if !exited {
-			t.Fatalf("Process did not exit!", err)
-		} else {
-			if exit.Status != 0 {
-				t.Fatalf("process exited with invalid status", exit.Status)
-			}
+		_, exited := err.(ProcessExitedError)
+		if exited {
+			t.Fatalf("Process exited unexpectedly!", err)
 		}
 	}, []string{"txest", "-pxassFlag"})
 }
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1694,10 +1694,6 @@ func TestCmdLineArgs(t *testing.T) {
 		if bp == nil || bp.Name != "unrecovered-panic" {
 			t.Fatalf("not on unrecovered-panic breakpoint: %v", p.CurrentBreakpoint)
 		}
-		_, exited := err.(ProcessExitedError)
-		if exited {
-			t.Fatalf("Process exited unexpectedly!", err)
-		}
 	}, []string{"txest", "-pxassFlag"})
 }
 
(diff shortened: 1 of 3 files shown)


moitias:
Not an Win32 API expert so can't really comment on what the actual differences between the implementations are (well, rather the implications of them), and reading the documentation doesn't really clarify;

> If both lpApplicationName and lpCommandLine are non-NULL, the null-terminated string pointed to by lpApplicationName specifies the module to execute, and the null-terminated string pointed to by lpCommandLine specifies the command line. The new process can use GetCommandLine to retrieve the entire command line. Console processes written in C can use the argc and argv arguments to parse the command line. Because argv[0] is the module name, C programmers generally repeat the module name as the first token in the command line.

And to clarify; my version uses both lpApplicationName and lpCommandLine, including the module name as the first token on in lpCommandLine while @visualfc sets lpApplicationName to null and only uses lpCommandLine which is also acceptable according to the docs - while also correctly quoting the executable name, which I don't do. 

So, maybe use his and I'll just make a separate PR for the tests? They are super trivial anyway.

derekparker:
@moitias @lukehoban I'd prefer going with this implementation since it is already tested, and just adding the quoting of the executable name to this PR.

moitias:
Looks like this way, it fails on OS X (if I'm reading the Travis output correctly) - which is odd as the tests didn't change (and worked before) and proc_windows.go should not be compiled into OS X binaries, right?

derekparker:
@moitias the tests are flaky on OSX at the moment.

proc/proc_windows.go
```diff
@@ -61,7 +59,9 @@ func Launch(cmd []string) (*Process, error) {
 		}
 		defer syscall.CloseHandle(syscall.Handle(fd[i]))
 	}
-
+	// create suitable command line for CreateProcess - quote the executable name with absolute path
+	// and pass on the rest of the parameters. see https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425.aspx
+	cmdLine, _ := syscall.UTF16PtrFromString(strings.Join(append([]string{"\"" + argv0Go + "\""}, cmd[1:]...), " "))
```
alexbrainman: This will not work in many valid scenarios. You should do what standard library does. See https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L326 for details.

derekparker: Ping @moitias

moitias:
thanks for the pointer @alexbrainman and sorry for the holdup, forgot all about this. the method used for creating the commandline is private to syscall, so adapted it for use here, so the argument handling is pretty much identical to syscall.StartProcess.

moitias:
Are the OS X tests in general flaky? Looks like it's another test failing (testbreakpointcounts, only on 1.6), and none of the code in the commit should touch that on OS X, I don't think? 

Either way, I have no reliable access to an OS X machine (nor any experience with it) so I'm not sure how quickly I'll be able to fix the failing tests.

derekparker:
@moitias yes, I'm working on a patch to clean up the flaky OSX failures at the moment. Until then, I'll kick the build.

alexbrainman:
Thank you @moitias. LGTM

parallaxisjones:
any word on this?  I keep running into the same place in stack.go:265

// intentional nil pointer dereference to trigger sigpanic

proc/proc_test.go
```diff
@@ -43,6 +43,21 @@ func withTestProcess(name string, t testing.TB, fn func(p *Process, fixture prot
 	fn(p, fixture)
 }
 
+func withTestProcessArgs(name string, t testing.TB, fn func(p *Process, fixture protest.Fixture), Args []string) {
```
derekparker: nit: lowercase `Args`.

derekparker:
@moitias LGTM aside from nit. Can you please squash commits and format them according to https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message.

@parallaxisjones are you running into this on this PR, or without it?

parallaxisjones:
PR?

derekparker:
@parallaxisjones Pull Request. In other words, is that error happening on this branch with these changes, or without them? I'm unsure of where you're running into that issue that you mentioned in your comment.

derekparker:
@parallaxisjones actually I'm just confused by your comment. Not sure if you're running into an issue on this branch, or if you're just waiting on this to get merged in.

parallaxisjones:
Oh, sorry.  Yes I figured out what you meant by PR as soon as I commented sorry.  No I have the master branch pulled with go get, not on this branch.

I was asking as far as the status of this issue in general...  just waiting for the merge. Following the discussion threads it would seem like it's close to a merge back into master

moitias:
re-formatted PR in #501

Issue #483, opened 2016-04-03
runtime.mach_semaphore_wait()


Wizermil:
1. What version of Delve are you using (`dlv version`)?
    Delve Debugger
    Version: 0.11.0-alpha
    Build: cfe0e52d040955efc34e7b34ef41313f03fe4864

2. What version of Go are you using? (`go version`)?
    go version go1.6 darwin/amd64

3. What operating system and processor architecture are you using?
    Max OS X 10.11.4
    Darwin Sarge.local 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 x86_64

4. What did you do?
    I compiled this file: test.go
    ```go
    package main

    import (
    	"fmt"
    	"os"
    	"os/signal"
    )

    func main() {

    	c := make(chan os.Signal, 1)
    	signal.Notify(c, os.Interrupt)
    	done := make(chan bool, 1)

    	go func() {
    		select {
    		case <-done:
    			return
    		}
    	}()

    	select {
    	case <-c:
    		done <- true
    	}

    	fmt.Printf("hello, world\n")
    }
    ```
    and then run
    ```
    go build test.go
    dlv exec test
    ```
    Press Ctrl+C for (SIGINT)
    ```
    Type 'help' for list of commands.
    (dlv) c
    > runtime.mach_semaphore_wait() /usr/local/go/src/runtime/sys_darwin_amd64.s:411 (PC: 0x58ddb)
       406:	// func mach_semaphore_wait(sema uint32) int32
       407:	TEXT runtime·mach_semaphore_wait(SB),NOSPLIT,$0
       408:		MOVL	sema+0(FP), DI
       409:		MOVL	$(0x1000000+36), AX	// semaphore_wait_trap
       410:		SYSCALL
    => 411:		MOVL	AX, ret+8(FP)
       412:		RET
       413:
       414:	// func mach_semaphore_timedwait(sema, sec, nsec uint32) int32
       415:	TEXT runtime·mach_semaphore_timedwait(SB),NOSPLIT,$0
       416:		MOVL	sema+0(FP), DI
    (dlv) exit
    ```

5. What did you expect to see?
    I'm not sure that this go code is legit and I just wanted to see "hello world" when I send SIGINT to the process.
[...]

PR #486, opened 2016-04-06
terminal: helpful message in SIGINT handler

aarzilli:
Our handling of SIGINT is confusing to people that expect SIGINT to be
delivered to inferior when they press ^C (see #483 for example).
Let's make an effort to point out what's happening and how one can send
SIGINT to inferior.


derekparker:
Let's consolidate those messages, and omit specific instructions. Let's just say something like: "recieved SIGINT, stopping process (will not forward signal)".

aarzilli:
Done.

PR #487, opened 2016-04-07
Headless mode print listener address info

visualfc:
dlv headless mode print address info if --listen flag default

Example
> dlv debug --headless
> start listener addr 127.0.0.1:49635

> dlv connect 127.0.0.1:49635


aarzilli:
maybe we should only do this if `--log=true`?

visualfc:
This address information is provided to the dlv client to use, This information does not belong to --log information.

cmd/dlv/cmds/commands.go
```diff
@@ -358,6 +358,8 @@ func execute(attachPid int, processArgs []string, conf *config.Config) int {
 
 	var status int
 	if Headless {
+		// Print listener address
+		fmt.Printf("start listener addr %s\n", listener.Addr())
```
derekparker: This should be a `log.Printf`, and lets change the wording to "API server listening at: %s\n".

derekparker:
I agree with @aarzilli that this belongs in the logs, however maybe we should consider logging to be enabled by default when running in headless mode?

visualfc:
I want a dlv connect client to use the address (the dlv server not specified --listen), but do not want to print the log, so as to avoid mixing applications and log output. Perhaps there should be a new flags marker?

aarzilli:
IMHO it's a reasonable use case, maybe only print if `if Headless && Addr == "localhost:0`

visualfc:
@aarzilli , yes it is my intention. I'm changed the pull request, thanks.
```
if Headless {
		if Addr == "localhost:0" {
			// Print listener address
			fmt.Printf("API server listening at: %s\n", listener.Addr())
		}
...
```

cmd/dlv/cmds/commands.go
```diff
@@ -358,6 +358,10 @@ func execute(attachPid int, processArgs []string, conf *config.Config) int {
 
 	var status int
 	if Headless {
+		if Addr == "localhost:0" {
```
derekparker: I think this would be cleaner if we did something like `RootCmd.Flags().Get("listen").HasChanged()` to indicate that we're only printing if user defined value for that flag, and not relying on our default value, which may change, and it would be non obvious to update this piece of code.

derekparker: On second thought, I think it's fine to just print this our irregardless. I don't feel too strongly about this being in log output or normal stdout, but I think it should be printed consistently either way.

derekparker:
@visualfc sorry for the delay on this. I don't think you saw my last comment, but let's just be consistent and always print the listener address, even if the user specified one themselves. After that, please squash into a single commit, and format it as specified by https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message and I will merge.

Sorry for all of the delay and back & forth on this, it has been an especially busy few weeks.

visualfc:
update, headless always print listener address
  
    fmt.Printf("API server listening at: %s\n", listener.Addr())

PR #498, opened 2016-04-14
Make next, step, step-instruction work on parked goroutines, implemented stepout

aarzilli:
Applies on top of #448.

I've refactored the code for `proc.(*Process).Next` to make it work on parked goroutines. Our previous behavior when the user switched to a parked goroutine was to simply step through the goroutine running on `CurrentThread`, which was wrong. 

I've implemented the same changes to `Step` and `StepInstruction` as well, for consistency.

Finally I have implemented `stepout` that continues the execution of the program until the selected goroutine returns from the function in execution. In gdb this command is called `finish` but graphical debuggers usually call it "Step out".

### DIFF
--- a/_fixtures/defercall.go
+++ b/_fixtures/defercall.go
@@ -1,0 +1,29 @@
+package main
+
+var n = 0
+
+func sampleFunction() {
+	n++
+}
+
+func callAndDeferReturn() {
+	defer sampleFunction()
+	sampleFunction()
+	n++
+}
+
+func callAndPanic2() {
+	defer sampleFunction()
+	sampleFunction()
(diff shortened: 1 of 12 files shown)


aarzilli:
The failure on linux/1.5 seems unrelated to this patch.

derekparker:
Would you be able to break out the changes for parked goroutines vs stepout functionality into separate PRs for easier review?

aarzilli:
I will do that after we are done with #448

aarzilli:
Closing this in favor of #563 and #564. I will make a PR for stepout after #563.

PR #500, opened 2016-04-15
Misc osx refactors

derekparker:
This set of patches sets the close-on-exec flag on the pipes used for parent/child coordination. Also includes a small refactor to avoid type coercion for C types.

I should have another PR up soon to clean up a few other issues on OSX.


derekparker:
cc: @aarzilli

PR #501, opened 2016-04-16
Fix command-line arguments on Windows

moitias:
This is a better-formatted PR replacing #479 and fixing the issue of command-line arguments not getting passed to the binary on Windows, including tests.

The nit has been fixed and commits are formatted better. Sorry for not reading the documentation on contributing!

### DIFF
--- a/proc/proc_windows.go
+++ b/proc/proc_windows.go
@@ -47,9 +47,6 @@ func Launch(cmd []string) (*Process, error) {
 	if _, err := os.Stat(argv0Go); err != nil {
 		return nil, err
 	}
-
-	argv0, _ := syscall.UTF16PtrFromString(argv0Go)
-
 	// Duplicate the stdin/stdout/stderr handles
 	files := []uintptr{uintptr(syscall.Stdin), uintptr(syscall.Stdout), uintptr(syscall.Stderr)}
 	p, _ := syscall.GetCurrentProcess()
@@ -62,6 +59,32 @@ func Launch(cmd []string) (*Process, error) {
 		defer syscall.CloseHandle(syscall.Handle(fd[i]))
 	}
 
+	argv0, err := syscall.UTF16PtrFromString(argv0Go)
+	if err != nil {
+		return nil, err
+	}
+
+	// create suitable command line for CreateProcess
+	// see https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L326
+	// adapted from standard library makeCmdLine
+	// see https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L86
+	var cmdLineGo string
+	if len(cmd) > 1 {
+		for _, v := range cmd {
+			if cmdLineGo != "" {
+				cmdLineGo += " "
+			}
+			cmdLineGo += syscall.EscapeArg(v)
+		}
+	}
+
+	var cmdLine *uint16
+	if cmdLineGo != "" {
+		if cmdLine, err = syscall.UTF16PtrFromString(cmdLineGo); err != nil {
+			return nil, err
+		}
+	}
+
 	// Initialize the startup info and create process
 	si := new(sys.StartupInfo)
 	si.Cb = uint32(unsafe.Sizeof(*si))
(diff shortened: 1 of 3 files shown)


proc/proc_windows.go
```diff
@@ -62,6 +59,32 @@ func Launch(cmd []string) (*Process, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// create suitable command line for CreateProcess
+	// see https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L326
+	// adapted from standard library makeCmdLine
+	// see https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L86
+	var cmdLineGo string
+	if len(cmd) > 1 {
```
derekparker: Sorry I didn't catch this in earlier reviews, but this branch seems like an error. If we enter the branch, we call `syscall.EscapeArg` on every element in the `cmd` slice, including index `0`. However if we don't enter that branch, index `0` never gets escaped.

Unless I'm mistaken it seems like we would want to escape everything the `cmd` slice regardless of the number of elements, or at least if `len(cmd) > 0`.

derekparker:
Aside from my last comment everything looks good. Thanks for updating your PR, and for this patch in general.

moitias:
Hey, good catch - that was probably a leftover from the time when it was passing *just* the args as command line to the syscall (which was then deemed to be possibly problematic). 

I also improved the testing to catch this - although it didn't catch it as with 0 args, len(cmd) == 1, the cmdLine was not created. However, cmd[0] (eg. the binary) was still passed on to sys.createprocess due to the previously discussed Windows quirks, so the tests work even with the bug. I'm not even sure how to reproduce as I'm not familiar with the edge cases mentioned by @alexbrainman in #479 .. 

No problems, happy to help out and sorry for the issues.

alexbrainman:
> I'm not familiar with the edge cases mentioned by @alexbrainman in #479 ..

Sorry for confusion. I was talking about having space or " (and similar) inside of Launch parameters. For example:
```
Launch("test.exe", `C:\Documents and Settings\brainman\Desktop\a.txt`)
```
I don't insist you add test for it. But I think it is important that your code works correctly.

Alex

moitias:
Ah, right - this should be good to go then as that indeed happens now.

derekparker:
@moitias I think it would be good to have a test that asserts the escaping is done properly, to avoid regressions.

moitias:
Tests added to make sure the arguments are escaped correctly.

Looks like OS X tests are still flaky (failures in various different tests), but at least on OS X/Go 1.6 the args test gets run and passes.

derekparker:
@moitias thanks! I'm working on fixes for the OSX flakyness, should be fixed up soon. Going to kick travis, make sure we can get those tests green and then merge your code.

dlsniper:
@moitias / @derekparker thanks for this. How does one sends parameters to the executable? Will it work as it does in Linux / OS X with `-- -myExec=1` or in a different way? Thanks.

derekparker:
@dlsniper I believe it should work the same, however I do not have a Windows box I can test on. @moitias can you confirm?

dlsniper:
Yeap, same here, I'll be able to get a Windows machine only in the weekend (at best) I've looked for a test case but it wasn't clear, thus the question.

moitias:
Doesn't seem like there are any tests for dlv (the CLI part), but the flag parsing is handled by [spf13/pflag](https://github.com/spf13/pflag) by way of [spf13/cobra](https://github.com/spf13/cobra). dlv does [split](https://github.com/derekparker/delve/blob/master/cmd/dlv/cmds/commands.go#L191) arguments from the command line, but detects the double dash using pflag. Did not investigate further into pflags innards..

Tested dlv manually on Windows command prompt, and passing arguments both with and without -- worked, for whatever thats worth!

dlsniper:
@moitias thanks for this and quick reply. I'll land the upgraded version in the IntelliJ plugin (might be in the next nightly or one after that). Awesome stuff!

PR #511, opened 2016-04-21
proc: better error message when running dlv debug when there is no main

### LINKED ISSUE #509: command/debug: Unexpected error when 'main' package doesn't exist in current directory

Delve Debugger
Version: 0.11.0-alpha
Build: f454207a340cc6b16b69f5924c647938b3c1a771
go version devel +b57ac33 Wed Apr 20 15:31:42 2016 +0000 darwin/amd64
darwin/amd64
4. What did you do?
cd go/src/fmt; dlv debug
A working debugger.
6. What did you see instead?

Here is the trace. Note that I can do 'dlv exec' if I build the binary, and it works great, but 'dlv debug' gives me the dreaded fork/exec error. This makes me think it might be an actual bug rather than a codesigning issue.

dunnart=% dlv debug
could not launch process: could not fork/exec
dunnart=% dlv exec fmt.test
(dlv) r
Process restarted with PID 85560
(dlv) break main.main
Breakpoint 1 set at 0x21a3 for main.main() fmt/_test/_testmain.go:192
(dlv) continue
> main.main() fmt/_test/_testmain.go:192 (hits goroutine(1):1 total:1) (PC: 0x21a3)
(dlv) stack
0  0x00000000000021a3 in main.main
   at fmt/_test/_testmain.go:192
1  0x0000000000027855 in runtime.main
   at /Users/r/go/src/runtime/proc.go:188
2  0x00000000000504b0 in runtime.goexit
   at /Users/r/go/src/runtime/asm_amd64.s:2013
(dlv) quit
dunnart=%

aarzilli:
Fixes #509

### DIFF
--- a/proc/proc_windows.go
+++ b/proc/proc_windows.go
@@ -37,15 +37,20 @@ func Launch(cmd []string) (*Process, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Make sure the binary exists.
+
+	// Make sure the binary exists and is an executable file
 	if filepath.Base(cmd[0]) == cmd[0] {
 		if _, err := exec.LookPath(cmd[0]); err != nil {
 			return nil, err
 		}
 	}
-	if _, err := os.Stat(argv0Go); err != nil {
-		return nil, err
+
+	peFile, err := openExecutablePath(argv0Go)
+	if err != nil {
+		return nil, NotExecutableErr
 	}
+	peFile.Close()
+
 	// Duplicate the stdin/stdout/stderr handles
 	files := []uintptr{uintptr(syscall.Stdin), uintptr(syscall.Stdout), uintptr(syscall.Stderr)}
 	p, _ := syscall.GetCurrentProcess()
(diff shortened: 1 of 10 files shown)


derekparker:
This doesn't include Windows. Maybe instead we should perform this check in `cmd/dlv/cmds/commands.go`, right after we run `gocommand` and ensure that the binary we expected to have been created exists, or check return code of the `go` command (will it return non-zero if it doesn't produce a binary?).

aarzilli:
The binary is always created, it's just that sometimes it's not an executable (and, of course, it doesn't return an error).

derekparker:
@aarzilli got it. We still need to account for Windows. Also, I think we should still catch this error earlier, and consolidate the logic for checking this consdition into a single location if we can.

derekparker:
@aarzilli any update on this? I'd like to merge this, but we must account for Windows.

aarzilli:
I'm don't know how to solve this on windows (I don't even know how this problem manifests on windows)

aarzilli:
I don't think it's going to be very useful (especially on OS X) but it couldn't hurt.

aarzilli:
@alexbrainman: I've added your changes but rewrote the test to use one of the existing fixtures.

alexbrainman:
LGTM. Thank you.

Alex

proc/proc_windows.go
```diff
@@ -308,19 +313,23 @@ func (dbp *Process) parseDebugLineInfo(exe *pe.File, wg *sync.WaitGroup) {
...
+	if err == nil {
+		dbp.dwarf, err = dwarfFromPE(peFile)
+		if err != nil {
+			return nil, err
+		}
 	}
-	peFile, err := pe.NewFile(f)
-	if err != nil {
-		return nil, err
+	if peFile.Machine != pe.IMAGE_FILE_MACHINE_AMD64 {
```
derekparker: `peFile` could be nil here. We must explicitly check for an error, or check that this pointer is not nil.

derekparker:
nit: Can you switch any message with `${OS}-${ARCH}` to be `${OS]/${ARCH}` for consistency (all other references generally include a slash, not hyphen.

PR #512, opened 2016-04-24
Sort methods in alphabetical order

dlsniper:
This should help out comparing differences between API versions.
I've tried to diff things before but it was simply too much surface to cover and the methods being out of order between commits didn't helped.
As such, I hope this will help others if needed.
Would be helpful to keep this ordered in the future to also ensure easier updates between same version or cross version.
Thanks.

### DIFF
--- a/service/rpc1/server.go
+++ b/service/rpc1/server.go
@@ -14,20 +14,55 @@ import (
 	"github.com/derekparker/delve/service/debugger"
 )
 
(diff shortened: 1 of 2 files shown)


dlsniper:
I'll close this for now as the changes make this unmergeable. Please let me know if there's any interest in having this done and I'll redo it if so. Thank you!

PR #518, opened 2016-04-27
Documentation for command line frontend

aarzilli:
Copied old wiki documentation in `Documentation/cli/`, changed `help` command so it prints only a summary of each command and more detailed documentation when followed by the name of a command, added a script to automatically generate `Documentation/cli/README.md`.

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -72,38 +72,139 @@ func DebugCommands(client service.Client) *Commands {
 	c := &Commands{client: client}
 
 	c.cmds = []command{
-		{aliases: []string{"help", "h"}, cmdFn: c.help, helpMsg: "Prints the help message."},
-		{aliases: []string{"break", "b"}, cmdFn: breakpoint, helpMsg: "break [name] <linespec>"},
-		{aliases: []string{"trace", "t"}, cmdFn: tracepoint, helpMsg: "Set tracepoint, takes the same arguments as break."},
+		{aliases: []string{"help", "h"}, cmdFn: c.help, helpMsg: `Prints the help message.
+
+	help [command]
+	
+Type "help" followed by the name of a command for more information about it.`},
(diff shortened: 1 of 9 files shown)


terminal/command.go
```diff
@@ -72,38 +72,139 @@ func DebugCommands(client service.Client) *Commands {
...
+	-g	displays location of go instruction that created the goroutine
+	
+If no flag is specified the default is -u.`},
+		{aliases: []string{"goroutine"}, allowedPrefixes: onPrefix | scopePrefix, cmdFn: c.goroutine, helpMsg: `Shows or changes current goroutine
+
+	goroutine
+	goroutine <id>
+	goroutine <id> <command>
+
+Called without arguments it will show informations about the current goroutine.
```
derekparker: s/informations/information/

terminal/command.go
```diff
@@ -165,26 +266,53 @@ func (c *Commands) Merge(allAliases map[string][]string) {
...
 	return nil
 }
 
 func (c *Commands) help(t *Term, ctx callContext, args string) error {
+	if args != "" {
+		for _, cmd := range c.cmds {
+			match := false
+			for _, alias := range cmd.aliases {
+				if alias == args {
+					match = true
```
derekparker: Let's just print the help message here and return instead of breaking and printing below.

derekparker:
This is so good, thanks for improving the documentation around the commands.

Regarding the script, I'm absolutely a fan of being able to automatically generate this documentation, but maybe it would be better to extend our command interpreter to produce this documentation for us? Similar to how Cobra can provide CLI documentation. That is not a necessity for this PR, btw. Just something to consider as we continue to improve our command interpreter.

aarzilli:
> Regarding the script, I'm absolutely a fan of being able to automatically generate this documentation, but maybe it would be better to extend our command interpreter to produce this documentation for us?

Yes, that's a better idea, I've done that.

PR #520, opened 2016-04-29
service/rpc: RPC server should serve requests consequently

zolotov:
`rpc.ServeCodec` runs all remote methods via go-routine. So it cannot give any guarantees about methods invokation order. As a result we have following bugs in IDEA<->Delve integration when breakpoints are set after `continue` command invokation: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2360

This PR uses `ServeRequest` method instead, which is synchronously variant of serving request. 

Also I've run gofmt before commit, please tell me if you want to keep formatting changes in the separate commit.

### DIFF
--- a/service/rpc1/server.go
+++ b/service/rpc1/server.go
@@ -13,9 +13,10 @@ import (
 	"github.com/derekparker/delve/service"
 	"github.com/derekparker/delve/service/api"
 	"github.com/derekparker/delve/service/debugger"
+	"io"
 )
 
-var defaultLoadConfig = proc.LoadConfig{ true, 1, 64, 64, -1 }
+var defaultLoadConfig = proc.LoadConfig{true, 1, 64, 64, -1}
 
 type ServerImpl struct {
 	s *RPCServer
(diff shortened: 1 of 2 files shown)


aarzilli:
The tests are failing. I suspect the problem is that with this change it's impossible to interrupt a running process.

zolotov:
@aarzilli indeed, for example in Issue419 test `continue` method waits for some event from debugger. `halt` operation is supposed to lead to this event. Since all requests are synchronized, `halt` will be never started. 

I'm not sure if it's issue of test or production code and I have no enough knowledge of internal delve architecture to fix tests. So feel free to close this PR, in this case I'll just file an issue about race-condition due to requests invokation reordering.

aarzilli:
It's not just the tests, a user can interrupt a running program by 
calling the halt command, if a continue command blocks further commands 
from being executed this doesn't work.

This patch needs a more complex approach.

zolotov:
Filed an issue #523

derekparker:
Closing this as we will likely take another approach.

PR #524, opened 2016-05-07
Go 1.7 compatibility

aarzilli:
There's really only two changes here
1. go1.7 changed a bit how DWARF is organised, this results in sometimes types being printed as 'struct structName' instead of just 'structName' and vice versa. The easiest way to fix this was to make a small change to x/debug/dwarf, which I have temporarily imported. Should we submit the change upstream (it hasn't been very active recently)?

2. there's a big change to how types are represented internally, which impacts our `loadInterface` function. We used to use a `string` field inside `runtime._type` to determine the concrete type and now that field is gone, it has been replaced by a `str` field that needs to be used in concert with the address of its containing `runtime._type` struct to find the area of memory containing the actual string name. New code is mostly adapted from `runtime` and `reflect`.

PS. how do we add go1.7 to appveyor?


derekparker:
@aarzilli yes, let's send a patch in upstream for this fix.

aarzilli:
Done: https://go-review.googlesource.com/#/c/23085/

PR #525, opened 2016-05-11
WIP - OSX/misc refactoring

derekparker:
*This is still a work in progress*

Submitting this to get early feedback / let folks know movement is happening on this refactor. Plan is to fully finish this and have it merged by end of the week, in time to cut a new release.

The goal of this refactor is to improve code readability, testability, and use that to improve stability (especially on OSX).

As mentioned, this is still a work in progress at the moment, still more to be done before it is ready to be merged.


cmd/dlv/cmds/commands.go
```diff
@@ -81,7 +81,7 @@ func New() *cobra.Command {
 	RootCommand.PersistentFlags().BoolVarP(&Log, "log", "", false, "Enable debugging server logging.")
 	RootCommand.PersistentFlags().BoolVarP(&Headless, "headless", "", false, "Run debug server only, in headless mode.")
 	RootCommand.PersistentFlags().BoolVarP(&AcceptMulti, "accept-multiclient", "", false, "Allows a headless server to accept multiple client connections. Note that the server API is not reentrant and clients will have to coordinate")
-	RootCommand.PersistentFlags().IntVar(&ApiVersion, "api-version", 1, "Selects API version when headless")
+	RootCommand.PersistentFlags().IntVar(&ApiVersion, "types.version", 1, "Selects API version when headless")
```
aarzilli: It happened again :)

proc/proc.go
```diff
@@ -768,47 +744,9 @@ func (dbp *Process) ClearTempBreakpoints() error {
...
-func (dbp *Process) run(fn func() error) error {
-	dbp.allGCache = nil
-	if dbp.exited {
-		return fmt.Errorf("process has already exited")
-	}
-	for _, th := range dbp.Threads {
-		th.CurrentBreakpoint = nil
-		th.BreakpointConditionMet = false
-		th.BreakpointConditionError = nil
-	}
```
aarzilli: I don't see these lines in `resume`, what clears the obsolete breakpoints now?

proc/threads.go
```diff
@@ -68,40 +66,55 @@ func (thread *Thread) Continue() error {
...
-	if err != nil {
-		if _, exited := err.(ProcessExitedError); exited {
-			return err
-		}
-		return fmt.Errorf("step failed: %s", err.Error())
+	// Clear state.
+	thread.CurrentBreakpoint = nil
+	thread.BreakpointConditionMet = false
+	thread.BreakpointConditionError = nil
+	thread.running = true
```
aarzilli: Never mind about the previous comment, I see that those lines are here.

proc/threads.go
```diff
@@ -129,27 +142,22 @@ func (thread *Thread) setNextBreakpoints() (err error) {
...
+	fde, err := thread.dbp.Dwarf.Frame.FDEForPC(loc.PC)
 	if err != nil {
 		return err
 	}
 	if filepath.Ext(loc.File) == ".go" {
-		err = thread.next(curpc, fde, loc.File, loc.Line)
+		err = thread.next(loc, fde)
 	} else {
-		err = thread.cnext(curpc, fde, loc.File)
+		err = thread.cnext(loc.PC, fde, loc.File)
```
aarzilli: Should probably take the whole `loc` like `(*Thread).next`.

aarzilli:
I like `pkg/dwarf` and I like that `threadResume` is concentrated in one function.

cmd/dlv/cmds/commands.go
derekparker: Hah! I'll fix that.

derekparker:
Update on this: making good progress, but there is a lot to be done and I often have too little time. Hoping to finish this up by tomorrow or this weekend. Really want to finish this to unblock other PRs.

derekparker:
Another update: Made great progress, looks like still on track to get this merged in very soon. Mostly doing some cleanup now before getting ready for review.

corvino:
Great to hear. I will play with getting the IntelliJ Go plugin to build so I can help with testing when you are ready.

Keep up the good work!

derekparker:
Made a bit more progress, but through the refactoring some new errors have popped up. The code is shaping up nicely, but teasing things apart has broken things (to be expected). In the interest of finishing this refactor as soon as humanly possible, it would be nice to get some more eyes on this.

@aarzilli -- since we're in different time zones maybe you can take a look at a few things while I'm asleep :). The tests are failing on all 3 supported OSes at the moment, but the 2 I'm most interested in tracking down are Linux and OSX. Firstly, on Linux there is a flicker on the BreakpointCounts test where the count is sometimes off by 1-3, and I haven't been able to track down why (haven't dug too deep, been mostly focusing on OSX). Additionally on OSX, `TestNextConcurrent` and `TestNextConcurrentVariant2` consistently fail and I'm unsure why, but also haven't been able to dig too deep. I will be able to spend some time on this tomorrow, so if you wouldn't mind spending a bit of time to track down these errors or get some clues any help would be greatly appreciated to get this over the finish line.

Additionally, if anyone else is interested in digging into the test failures on this branch, any eyes would be greatly appreciated. As I said, I will be digging deeper tomorrow, but fresh eyes always helps.

aarzilli:
I don't have time to look into this in depth today however, here's some things I have noticed:

- you have unified the setCurrentBreakpoints for all three operating systems this will almost certainly never work since windows api works differently (from what I recall it was also broken on OSX but maybe you have reworked that part)

- on OS X there needs to be a call to updateThreadsList after we are certain that all threads have stopped, otherwise we will miss breakpoints (and crash the target)

- the implementation of Step doesn't work, it needs to be reverted to the way it was. Just checking if the current instruction is a call isn't enough since calls are often preceded by a series of stack setup instructions

- `proc.(*Thread).singleStep` in thread_linux used to have a for loop, that wasn't an accident I put it there to fix the very bug with TestBreakpointsCount that you see as a regression

- I don't like logrus all it does can be done more readably and more optimizer friendly with if, log and a constant. Plus I had to disable it anyway because it was barfing to much logs on me and it was impossible to figure out what was actually wrong

I'm wary of all the changes to proc, that code is very fragile due to the unusual nature of its concurrency model, I wish we could take a more gradual approach to refactoring that area.

derekparker:
@aarzilli it's so good to have other eyes on things! Thanks for the feedback.

> the implementation of Step doesn't work, it needs to be reverted to the way it was. Just checking if the current instruction is a call isn't enough since calls are often preceded by a series of stack setup instructions

I've increased the number of instructions it looks at to handle this. Admittedly brittle, but I would honestly rather avoid the cost of instruction stepping through the process. Although, maybe it's not too bad since we're only stepping to the next line so theoretically it should be _horrendous_, but on that same token we already have code that steps to the next line (Next) but doesn't step into function, so maybe we rework that as well?

> proc.(*Thread).singleStep in thread_linux used to have a for loop, that wasn't an accident I put it there to fix the very bug with TestBreakpointsCount that you see as a regression

Got it, I'll put that loop back in and see if it solves the problem.

> I don't like logrus all it does can be done more readably and more optimizer friendly with if, log and a constant. Plus I had to disable it anyway because it was barfing to much logs on me and it was impossible to figure out what was actually wrong

I am not 100% sold on logrus, but adding logging into proc has increased visibility into what is going on there, at least for me, and has helped me clean up some issues. The logging that is there will surely be cut down a bit, and you can select log levels, etc. But I think continuing to have 0 visibility into what proc is doing is not helpful in most cases.

proc/threads.go
```diff
@@ -42,24 +45,25 @@ type Location struct {
...
-	}
-	// Check whether we are stopped at a breakpoint, and
-	// if so, single step over it before continuing.
-	if _, ok := thread.dbp.FindBreakpoint(pc); ok {
-		if err := thread.StepInstruction(); err != nil {
-			return err
-		}
-	}
-	return thread.resume()
+func (t *Thread) Continue() error {
```
aarzilli: inline

aarzilli: should this be a public method?

proc/threads.go
```diff
@@ -42,24 +45,25 @@ type Location struct {
...
+	return threadResume(t, ModeResume, 0)
+}
+
+func (t *Thread) ContinueWithSignal(sig int) error {
+	return threadResume(t, ModeResume, sig)
```
aarzilli: should this be a public method?

proc/threads.go
```diff
@@ -68,51 +72,61 @@ func (thread *Thread) Continue() error {
...
-	thread.singleStepping = true
-	defer func() {
-		thread.singleStepping = false
-		thread.running = false
-	}()
-	pc, err := thread.PC()
-	if err != nil {
-		return err
-	}
+func (t *Thread) StepInstruction() error {
```
aarzilli: should this be a public method?

proc/threads.go
```diff
@@ -312,104 +321,105 @@ func (thread *Thread) newGVariable(gaddr uintptr, deref bool) (*Variable, error)
...
 	}
 	return
 }
 
 // Stopped returns whether the thread is stopped at
 // the operating system level. Actual implementation
 // is OS dependant, look in OS thread file.
-func (thread *Thread) Stopped() bool {
-	return thread.stopped()
+func (t *Thread) Stopped() bool {
```
aarzilli: we should have only one of Stopped and stopped

proc/proc.go
```diff
@@ -142,34 +135,70 @@ func (dbp *Process) Running() bool {
...
 
-	wg.Add(5)
-	go dbp.loadProcessInformation(&wg)
-	go dbp.parseDebugFrame(exe, &wg)
-	go dbp.obtainGoSymbols(exe, &wg)
-	go dbp.parseDebugLineInfo(exe, &wg)
-	go dbp.loadTypeMap(&wg)
-	wg.Wait()
+// CurrentLocation returns the location of the current thread.
+func (p *Process) CurrentLocation() (*Location, error) {
```
aarzilli: this helper is called in a single place (a test), should be inlined.

proc/proc.go
```diff
@@ -180,92 +209,131 @@ func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error
...
 }
 
 // SetTempBreakpoint sets a temp breakpoint. Used during 'next' operations.
-func (dbp *Process) SetTempBreakpoint(addr uint64) (*Breakpoint, error) {
-	return dbp.setBreakpoint(dbp.CurrentThread.ID, addr, true)
+func SetTempBreakpoint(p *Process, addr uint64) (*Breakpoint, error) {
+	return setBreakpoint(p, p.CurrentThread, addr, true, p.arch.BreakpointInstruction())
+}
+
+func setBreakpoint(p *Process, mem memoryReadWriter, addr uint64, temp bool, instr []byte) (*Breakpoint, error) {
```
aarzilli: I'm not sure what's the point of passing memoryReadWriter and instr separately, I don't think there can be a single circumstance where it would be useful for those to not be p.CurrentThread and p.arch.BreakpointInstruction.

proc/proc.go
```diff
@@ -180,92 +209,131 @@ func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error
...
-	}
-	return dbp.setBreakpoint(dbp.CurrentThread.ID, addr, false)
+func SetBreakpoint(p *Process, addr uint64) (*Breakpoint, error) {
+	return setBreakpoint(p, p.CurrentThread, addr, false, p.arch.BreakpointInstruction())
 }
 
 // SetTempBreakpoint sets a temp breakpoint. Used during 'next' operations.
-func (dbp *Process) SetTempBreakpoint(addr uint64) (*Breakpoint, error) {
-	return dbp.setBreakpoint(dbp.CurrentThread.ID, addr, true)
+func SetTempBreakpoint(p *Process, addr uint64) (*Breakpoint, error) {
```
aarzilli: I don't think this should be a public method.

proc/proc.go
```diff
@@ -180,92 +209,131 @@ func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error
...
+	}
 
+	delete(p.Breakpoints, addr)
 	return bp, nil
 }
 
 // Status returns the status of the current main thread context.
-func (dbp *Process) Status() *WaitStatus {
-	return dbp.CurrentThread.Status
+func (p *Process) Status() *WaitStatus {
```
aarzilli: This helper function is never called.

proc/proc.go
```diff
@@ -308,109 +376,207 @@ func (dbp *Process) Next() (err error) {
...
-		default:
-			// not a manual stop, not on runtime.Breakpoint, not on a breakpoint, just repeat
 		}
+		return conditionErrors(p)
 	}
+	return breakpointConditionNotMetError
 }
 
-func (dbp *Process) conditionErrors() error {
+func Wait(p *Process) (*WaitStatus, *Thread, error) {
```
aarzilli: This helper function is never called.

aarzilli:
This `resume` function in proc, I'm not sure about it. In principle it's ok but in practice it's branches share almost zero code, I think it's better to split it back into its components. I made `StepInto` a public function but it should be private, btw.

aarzilli:
I've added a few commits to resolve some semi-spurious test failures (process exits returned as errors).
I've also reverted the changes to step, the explanation why is in the relevant commit.

Unfortunately there's still an infrequent bug in the linux specific backend, the failure is rare but *not* spurious: it will cause a continue to fail with "no such process" message despite the process being still alive.

Given the infrequency of the bug and the fact that it's been introduced by the refactoring I have no desire to track this problem down.

derekparker:
After talking a few things through with @aarzilli I'm going to close this PR in favor of chunking it up into smaller commits, which will be merged over time. This unblocks the main tree, so we can start merging in other PRs, and improve our CI story.

PR #532, opened 2016-05-16
proc: fix could not read process comm name error

### LINKED ISSUE #531: proc: Could not read process comm name when launching debugger

1. What version of Delve are you using (`dlv version`)? b9b3e4ac03c1033fe6a8866813670fe58e96e0fd
2. What version of Go are you using? (`go version`)? 1.6.2
3. What operating system and processor architecture are you using?  linux (2.6.32-279.el6.x86_64)
4. What did you do? debugging
5. What did you expect to see? debugger started successfully
6. What did you see instead? 
Could not read process comm name: open /proc/29019/comm: no such file or directory

I think the cause of this issue is that /proc/[pid]/comm was not introduced before linux 2.6.33.

ggndnn:
fix #531

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/532)

### DIFF
--- a/proc/proc_linux.go
+++ b/proc/proc_linux.go
@@ -1,6 +1,7 @@
 package proc
 
 import (
+	"bytes"
 	"debug/gosym"
 	"errors"
 	"fmt"
@@ -8,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -350,12 +352,30 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
 	defer wg.Done()
 
 	comm, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/comm", dbp.Pid))
-	if err != nil {
-		fmt.Printf("Could not read process comm name: %v\n", err)
-		os.Exit(1)
+	if err == nil {
+		// removes newline character
+		comm = bytes.TrimSuffix(comm, []byte("\n"))
 	}
-	// removes newline character
-	comm = comm[:len(comm)-1]
+
+	if comm == nil || len(comm) <= 0 {
+		stat, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/stat", dbp.Pid))
+		if err != nil {
+			fmt.Printf("Could not read proc stat: %v\n", err)
+			os.Exit(1)
+		}
+		expr := fmt.Sprintf("%d\\s*\\((.*)\\)", dbp.Pid)
+		rexp, err := regexp.Compile(expr)
+		if err != nil {
+			fmt.Printf("Regexp compile error: %v\n", err)
+			os.Exit(1)
+		}
+		match := rexp.FindSubmatch(stat)
+		if match == nil {
+			fmt.Printf("No match found using regexp '%s' in /proc/%d/stat\n", expr, dbp.Pid)
+			os.Exit(1)
+		}
+		comm = match[1]
+	}
 	dbp.os.comm = strings.Replace(string(comm), "%", "%%", -1)
 }
 


proc/proc_linux.go
```diff
@@ -338,12 +339,32 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
 	defer wg.Done()
 
 	comm, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/comm", dbp.Pid))
-	if err != nil {
+	if(err == nil) {
+		// removes newline character
+		comm = comm[:len(comm)-1]
+	}
+	for (comm == nil || len(comm) <= 0) {
```
aarzilli: This should be an if not a for, there is no way this loops more than once, right?

aarzilli:
Commits need to be squashed, also there's a refactoring in progress, I don't know what @derekparker wants to do with this in the short term.

ggndnn:
I get it, no problem, and commit has been squashed.

proc/proc_linux.go
ggndnn: Right, logically it's if. I meant to use break statement to leave current block when error occur.

derekparker:
@ggndnn thanks for the fix! I'm going to hold off on merging this until https://github.com/derekparker/delve/pull/525 lands (should be soon). I'll make sure to ping you once that happens so we can get this merged.

ggndnn:
@derekparker ok, no problem, thanks.

proc/proc_linux.go
```diff
@@ -338,12 +339,32 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
 	defer wg.Done()
 
 	comm, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/comm", dbp.Pid))
-	if err != nil {
+	if err == nil {
+		// removes newline character
+		comm = comm[:len(comm)-1]
```
derekparker: Please use https://golang.org/pkg/bytes/#TrimSuffix for this instead.

proc/proc_linux.go
```diff
@@ -338,12 +339,32 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
...
+	}
+	for comm == nil || len(comm) <= 0 {
```
derekparker: I don't understand the loop here, can we do without it?

derekparker:
@ggndnn I'm going to break that large refactor into chunks so I'd like to merge your code after the feedback is taken care of. Thanks!

ggndnn:
@derekparker I have commited my modifications. Thanks.

proc/proc_linux.go
```diff
@@ -337,12 +339,30 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
...
+		comm = bytes.TrimSuffix(comm, []byte("\n"))
+	}
+
+	if comm == nil || len(comm) <= 0 {
+		stat, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/stat", dbp.Pid))
+		if err != nil {
+			fmt.Printf("Could not read proc stat: %v\n", err)
+			os.Exit(1)
+		}
+		expr := fmt.Sprintf("%d\\s*\\((.*)\\)", dbp.Pid)
```
derekparker: This would be easier to read if we simply used `scan` here instead of using a regexp.

ggndnn: I am not sure how to scan proc name including space character, any suggestion?

derekparker: Sorry for delay on responding, vacation + conferences and such. However, you're right, I misspoke. Scanning is not the way to go here.

derekparker:
@ggndnn if you can squash your commits I will merge. Thanks for the patch!

ggndnn:
@derekparker commits have been squashed. Thanks!

proc/proc_linux.go
```diff
@@ -350,12 +352,30 @@ func (dbp *Process) loadProcessInformation(wg *sync.WaitGroup) {
...
+		rexp, err := regexp.Compile(expr)
+		if err != nil {
+			fmt.Printf("Regexp compile error: %v\n", err)
+			os.Exit(1)
+		}
+		match := rexp.FindSubmatch(stat)
+		if match == nil {
+			fmt.Printf(fmt.Sprintf("No match found using regexp '%s' in /proc/%d/stat", expr, dbp.Pid))
```
derekparker: Should of caught this before, but the `Printf(Sprintf())` is unnecessary, can just format with `Printf` directly. Also, needs newline at end of string.

ggndnn: Yes, it's my fault. Thanks.

ggndnn:
@derekparker modification has been committed.

PR #563, opened 2016-06-02
Implement next, step, step-instruction for parked goroutines

aarzilli:
Also refactors proc.(*Process).Next to make implementing a stepout command easier.

### DIFF
--- a/proc/threads.go
+++ b/proc/threads.go
@@ -11,7 +11,6 @@ import (
 	"runtime"
 
 	"github.com/derekparker/delve/dwarf/debug/dwarf"
-	"github.com/derekparker/delve/dwarf/frame"
 )
 
 // Thread represents a single thread in the traced process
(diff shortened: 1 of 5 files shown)


aarzilli:
I've also added a commit to remove the setChanRecvBreakpoints functionality, the reason for this is in the commit message:

```
This function was broken for a very long time until the changes in this
PR fixed it. However I do not think it is desirable to be catapulted
into some other goroutine by 'next' just because that other goroutine
happened to receive a message on a channel.

This feature should be implemented by a new command, not next.
```

proc/threads.go
```diff
@@ -123,80 +124,65 @@ func (tbe ThreadBlockedError) Error() string {
...
-				// entry point of the deferred function so instead create
-				// a fake breakpoint which will be cleaned up later.
-				thread.dbp.Breakpoints[g.DeferPC] = new(Breakpoint)
-				defer func() { delete(thread.dbp.Breakpoints, g.DeferPC) }()
-				if _, err = thread.dbp.SetTempBreakpoint(dpc); err != nil {
-					return err
-				}
-				break
-			}
+// The first return value is set to true if the goroutine is in the process of exiting
```
derekparker: Erroneous comment?

proc/threads.go
```diff
@@ -123,80 +124,65 @@ func (tbe ThreadBlockedError) Error() string {
...
+	topframe, err := dbp.topframe()
 	if err != nil {
 		return err
 	}
 
-	// Get current file/line.
-	loc, err := thread.Location()
+	// Grab info on our current stack frame. Used to determine
+	// whether we may be stepping outside of the current function.
+	fde, err := dbp.frameEntries.FDEForPC(topframe.Current.PC)
```
derekparker: nit: Might be worth attaching FDE to frame struct itself.

derekparker:
Overall looks good, just a few nits.

proc/threads.go
aarzilli: good idea.

aarzilli:
Updated.

derekparker:
@aarzilli so it seems, unless I'm misunderstanding, that if we step on a parked goroutine, the actual behavior is that we simply end up on the same PC we were on, but with the goroutine now active. We need to ensure that we actually continue execution in a meaningful way once the goroutine has become active, and actually get to the next line.

aarzilli:
I prefer if we merged this with this known bug and tracked it in a separate issue/PR. This problem would be solved by the changes I was proposing to @dr2chase in mailing list.

derekparker:
Created https://github.com/derekparker/delve/issues/581 to track.

proc/threads.go
```diff
@@ -123,122 +123,97 @@ func (tbe ThreadBlockedError) Error() string {
 	return ""
 }
 
-// Set breakpoints for potential next lines.
-func (thread *Thread) setNextBreakpoints() (err error) {
-	if thread.blocked() {
-		return ThreadBlockedError{}
+// returns topmost frame of selected goroutine
+func (dbp *Process) topframe() (Stackframe, error) {
```
derekparker: I want to get away from adding methods to `Process`. Can you make this a function that takes the goroutine and thread, it seems that's all it really needs from `Process`.

derekparker:
One last style nit, otherwise lgtm.

PR #566, opened 2016-06-05
Add support for Appengine; parsing Appengine Go custom version

dbenque:
Appengine run its own version of Go.
The version returned is "go1.6.1 (appengine-1.9.37)".
The function ParseVersionString does not support that customization of the version.
This PR provides a simple patch that uses only the first part of the string (before space) to parse the version.


aarzilli:
lgtm

PR #574, opened 2016-06-15
proc: Fix stepping into runtime.duff*

### LINKED ISSUE #573: Step into runtime.duffzero (and runtime.duffcopy) continues instead

Version: 0.11.0-alpha
ede8801 Add support for Appengine; parsing Appengine Go custom version (#566)

go version devel +af0fc83 Tue Jun 14 15:32:34 2016 +0000 darwin/amd64
(1.7 beta-ish; I work on the Go compiler, this is part of an exercise to be sure that 1.7 works well with delve.)

OS X 10.11.5, amd64

4. What did you do?
Single stepped (attempted) into a call to fmt.Printf debugging this file
https://gist.github.com/dr2chase/acdd448e4fd19fdca6ab8551e55772a9#file-duffsteptest-go

A step into fmt.Printf (or into a runtime support function; getting that right is another RFE)
https://gist.github.com/dr2chase/63fbe3d2275811ddde733438624c462e#file-duffsteptest_expected-txt

6. What did you see instead?
Delve continued to the end.
https://gist.github.com/dr2chase/40de9f72c7d69e125248a56e1fab5f2a#file-duffsteptest_bug-txt

I have a fix for this (that's how I generated "expected output"), I'm trying to figure out how to create a proper delve test for this.

The cause of the bug is that duffzero and duffcopy aren't regular functions and are instead called at arbitrary offsets within their bodies.  Step encounters a call, uses StepInto to get into the called function, and StepInto sets its breakpoint near the beginning and Continue()s.  Since the call itself enters well past the prologue, the breakpoint is never hit, and the process (in this case) runs to exit.  Proposed fix is to check call PC target against function entry point, if they do not match, just step.  Landing in assembly language for duffzero/copy is not best-case, but it's better than not stopping at all.

dr2chase:
Detect calls that do not target a function's entrypoint
(i.e, calls to runtime.duffzero and runtime.duffcopy) and
instead step into them directly.  StepInto sets a breakpoint
past the called function's prologue and expects that continue
will hit that breakpoint, but because the call is into the
interior of the function (well past the prologue) this fails.

Fixes #573

### DIFF
--- a/_fixtures/issue573.go
+++ b/_fixtures/issue573.go
@@ -1,0 +1,25 @@
+package main
+
+// A debugger test.
+//   dlv debug
+//   b main.foo
+//   c
+//   s
+//   s
+// Expect to be stopped in fmt.Printf or runtime.duffzero
+import "fmt"
+
+var v int = 99
+
+func foo(x, y int) (z int) { // c goes here
+	fmt.Printf("x=%d, y=%d, z=%d\n", x, y, z) // s #1 goes here
+	z = x + y
+	return
+}
+
+func main() {
+	x := v
+	y := x * x
+	z := foo(x, y)
+	fmt.Printf("z=%d\n", z)
+}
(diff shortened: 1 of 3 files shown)


dr2chase:
Apologies in advance for process errors -- all my github work up till now has been on the Golang repo, which doesn't do github pull requests.

dr2chase:
Will redo pull request from a branch that's not master, since that seems to cause problems over time.

PR #565, opened 2016-06-04
proc: bugfix: Step doesn't progress on CALL with breakpoint

aarzilli:
If step is called on a CALL instruction where a breakpoint is also set
it will make no progress because information about the breakpoint is
cleared at the beginning of Step, Step calls Continue and Continue just
hits the breakpoint again.

Fixes #561

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/565)

PR #575, opened 2016-06-16
proc: Fix stepping into runtime.duff*

### LINKED ISSUE #573: Step into runtime.duffzero (and runtime.duffcopy) continues instead

dr2chase:
Detect calls that do not target a function's entrypoint
(i.e, calls to runtime.duffzero and runtime.duffcopy) and
instead step into them directly.  StepInto sets a breakpoint
past the called function's prologue and expects that continue
will hit that breakpoint, but because the call is into the
interior of the function (well past the prologue) this fails.

Fixes #573

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/575)

### DIFF
--- a/_fixtures/issue573.go
+++ b/_fixtures/issue573.go
@@ -1,0 +1,28 @@
+package main
+
+// A debugger test.
+//   dlv debug
+//   b main.foo
+//   c
+//   s
+//   s
+// Expect to be stopped in fmt.Printf or runtime.duffzero
+// In bug, s #2 runs to the process exit because the call
+// to duffzero enters duffzero well after the nominal entry
+// and skips the temporary breakpoint placed by StepZero().
+import "fmt"
+
+var v int = 99
+
+func foo(x, y int) (z int) { // c stops here
+	fmt.Printf("x=%d, y=%d, z=%d\n", x, y, z) // s #1 stops here; s #2 is supposed to stop in Printf or duffzero.
+	z = x + y
+	return
+}
+
+func main() {
+	x := v
+	y := x * x
+	z := foo(x, y)
+	fmt.Printf("z=%d\n", z)
+}
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -406,7 +406,13 @@ func (dbp *Process) Step() (err error) {
 			}
 			text, err := dbp.CurrentThread.Disassemble(pc, pc+maxInstructionLength, true)
 			if err == nil && len(text) > 0 && text[0].IsCall() && text[0].DestLoc != nil && text[0].DestLoc.Fn != nil {
-				return dbp.StepInto(text[0].DestLoc.Fn)
+				fn := text[0].DestLoc.Fn
+				// Ensure PC and Entry match, otherwise StepInto is likely to set
+				// its breakpoint before DestLoc.PC and hence run too far ahead.
+				// Calls to runtime.duffzero and duffcopy have this problem.
+				if fn.Entry == text[0].DestLoc.PC {
+					return dbp.StepInto(fn)
+				}
 			}
 
 			err = dbp.CurrentThread.StepInstruction()
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -1881,3 +1881,17 @@ func TestUnsupportedArch(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func Test1Issue573(t *testing.T) {
+	// calls to runtime.duffzero and runtime.duffcopy jump directly into the middle
+	// of the function and the temp breakpoint set by StepInto may be missed.
+	withTestProcess("issue573", t, func(p *Process, fixture protest.Fixture) {
+		f := p.goSymTable.LookupFunc("main.foo")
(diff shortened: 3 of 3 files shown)


dr2chase:
Any way to recheck that failure?  This is 3rd time (at least) I have attempted to do a github pull request properly, the previous two succeeded.

aarzilli:
Delve on OS X has some known problems, the test failure there is clearly unrelated to your changes.

service/test/integration1_test.go
```diff
@@ -945,6 +945,23 @@ func Test1NegativeStackDepthBug(t *testing.T) {
 	})
 }
 
+func Test1Issue573(t *testing.T) {
```
aarzilli: integration1_test exists to insure that we do not break backwards compatibility with the old API, things shouldn't be added to it. This test is better suited for `proc/proc_test.go`, translating it should be easy.

dr2chase:
Oops, left out test description, will do that.

derekparker:
Going to take a look today, was on vacation all last week. Sorry for the delay.

derekparker:
@aarzilli from our IRC discussion I think you're referring to this PR. Can you carry over your thoughts from the mailing list onto this thread so we can discuss here please.

aarzilli:
This is how I propose to reimplement step:

1. change the Temp field of to Kind of type breakpointKind, where breakpointKind is:

```
    type breakpointKind int
    const (
        normalBreakpoint = breakpointKind(iota)
        nextBreakpoint
        stepBreakpoint
    )
```

2. Next will create a bunch of breakpoints of kind == nextBreakpoint

3. Step will work like this:
   - disassemble all instructions belonging to the current source line
   - if we are on a CALL instruction set a breakpoint of kind == nextBreakpoint on the target and call Continue, otherwise:
   - set a breakpoint of kind == stepBreakpoint on all non-hidden CALLs (which also means all CALLs with a non-immediate destination) found in what we disassembled
   - set all the breakpoints that Next would set with kind == nextBreakpoint
   - call Continue

4. Change Continue to do this when a breakpoint of kind == stepBreakpoint is triggered
    - delete all temp breakpoints, set a new breakpoint on the target of current CALL
    - if there are no other triggered breakpoints loop, otherwise pick another CurrentThread and exit

derekparker:
@aarzilli I'm on board with that, as from what I understand you are proposing removing the current `SingleStep` implementation and instead going with the `Next`-style implementation, correct?

> set a breakpoint of kind == stepBreakpoint on all non-hidden CALLs (which also means all CALLs with a non-immediate destination) found in what we disassembled

So this is what would satisfy this bug if I understand correctly, since we're setting a breakpoint on non-immediate destinations such as these computed jumps into duff* functions, correct?

We're still going to need logic to determine if where we are jumping into is equal to the function entrypoint, as we want to skip the looping prologue. 

Correct me if I am misunderstanding anything. If that is the case, I am leaning towards merging this fix in for the time being, with the understanding that Step/StepInto will be updated with the logic you describe.

aarzilli:
> So this is what would satisfy this bug if I understand correctly, since we're setting a breakpoint on non-immediate destinations such as these computed jumps into duff* functions, correct?

duff* stuff are private runtime functions, so I would just skip them, at least by default. But besides that, yes.

> Correct me if I am misunderstanding anything. If that is the case, I am leaning towards merging this fix in for the time being, with the understanding that Step/StepInto will be updated with the logic you describe.

Makes sense. I would have pulled code from this patch anyway.

aarzilli:
BTW, in chat I was referring to #565, my PR concerning Step, not this PR.

Issue #358, opened 2016-01-22
step out of a function


aybabtme:
Heyo!

it would be cool to have the inverse of `step`, something that steps out of the current function.

alvarolm:
first thanks @derekparker @aarzilli and everyone who has contributed to delve, out of curiosity, its there a date or timeline to merge @aarzilli s work ?

lfrdreamman:
thanks all, when the step out can work?

aarzilli:
Implemented by 0f4b5150c370467c6478bc73b7ed64b92d94af56

PR #583, opened 2016-06-26
proc, terminal: stepout command

aarzilli:
Command to step out of the currently executing function.

Implements #358

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/583)

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -9,6 +9,7 @@ import (
 	"go/constant"
 	"go/token"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


aarzilli:
Rebased on #595.

aarzilli:
Since this made use of the `DeferCond` member that was removed during the review of #603 I have rebased it on top of #603. I've also removed this text:

```
Known bug: stepout will step into a deferred function that is called directly or by runtime.deferreturn, but it should only step into deferred functions if they are called by a panic (see #582).
```

from the opening message since it does not apply anymore.

aarzilli:
Rebased.

derekparker:
@aarzilli can you rebase please?

aarzilli:
Rebased.

PR #585, opened 2016-07-04
proc: Back to using vendored golang.org/x/debug/

aarzilli:
Patch https://go-review.googlesource.com/23085 has been merged so we
can go back to using golang.org/x/debug/.


derekparker:
lgtm

PR #586, opened 2016-07-04
debugger/locations: Fixed bug parsing remote package paths

aarzilli:
`parseFuncLocationSpec` did not parse function locations with a full package path correctly, just like go's `debug/gosym` didn't before go1.7. My fault :(


derekparker:
lgtm

PR #591, opened 2016-07-08
prettyprint: Print type of the elements of arrays of interface type

aarzilli:
It's often useful to see the concrete type of the individual elements that comprise an array of interface type.

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/591)

### DIFF
--- a/service/api/prettyprint.go
+++ b/service/api/prettyprint.go
@@ -35,7 +35,11 @@ func (v *Variable) writeTo(buf io.Writer, top, newlines, includeType bool, inden
 	}
 
 	if !top && v.Addr == 0 {
-		fmt.Fprintf(buf, "%s nil", v.Type)
+		if includeType && v.Type != "void" {
+			fmt.Fprintf(buf, "%s nil", v.Type)
+		} else {
+			fmt.Fprintf(buf, "nil")
+		}
 		return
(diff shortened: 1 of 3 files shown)


service/api/prettyprint.go
```diff
@@ -88,10 +92,10 @@ func (v *Variable) writeTo(buf io.Writer, top, newlines, includeType bool, inden
 			} else if data.Children[0].OnlyAddr {
 				fmt.Fprintf(buf, "0x%x", v.Children[0].Addr)
 			} else {
-				v.Children[0].writeTo(buf, false, newlines, false, indent)
+				v.Children[0].writeTo(buf, false, newlines, !includeType, indent)
```
derekparker: It seems odd to negate what was passed in. I think I understand the sentiment, that if we say "don't print type info", if the type happens to be an interface w/ children we automatically print the types of the children, right?

Maybe we should let the user control this, and allow the user to specify via the `print` command whether type information should be included? If types should be included, then if the type is an interface type we also print out the type of each underlying element.

aarzilli: Adding flags to `print` is hard because it accepts any valid expression, also I'm not convinced about its usefulness.

PR #595, opened 2016-07-14
proc: bugfix: Next and normal calls to deferred function

### LINKED ISSUE #582: Next steps into function calls to deferred functions even when they are called normally
Given this program:

```
package main

func f() {
}

func main() {
	defer f()
	f()
}
```

this happens:

```
% ~/n/go/src/github.com/derekparker/delve/dlv exec ./nextdefer
(dlv) break nextdefer.go:8
Breakpoint 1 set at 0x401040 for main.main() ./nextdefer.go:8
(dlv) continue
> main.main() ./nextdefer.go:8 (hits goroutine(1):1 total:1) (PC: 0x401040)
     3:	func f() {
     4:	}
     5:	
     6:	func main() {
     7:		defer f()
=>   8:		f()
     9:	}
(dlv) next
> main.f() ./nextdefer.go:4 (PC: 0x401000)
     1:	package main
     2:	
     3:	func f() {
=>   4:	}
     5:	
     6:	func main() {
     7:		defer f()
     8:		f()
     9:	}
(dlv) q
``` 

expected behaviour: program continues to nextdefer.go:9
what happens: program continues to nextdefer.go:4

The temp breakpoint we set for deferred functions should check that the function call was made either by (a) a panic or (b) runtime.deferreturn.

The (not yet merged) stepout command should behave similarly but only accept calls from panic.

aarzilli:
When a deferred function is called directly Next and StepOut should not
step into it.

Fixes #582

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/595)

### DIFF
--- a/proc/breakpoints.go
+++ b/proc/breakpoints.go
@@ -33,7 +33,18 @@ type Breakpoint struct {
 	HitCount      map[int]uint64 // Number of times a breakpoint has been reached in a certain goroutine
 	TotalHitCount uint64         // Number of times a breakpoint has been reached
 
-	Cond ast.Expr // When Cond is not nil the breakpoint will be triggered only if evaluating Cond returns true
+	// When DeferCond is set the breakpoint will only trigger
+	// if the caller is runtime.gopanic or if the return address
+	// is in the DeferReturns array.
+	// Next sets DeferCond for the breakpoint it sets on the
+	// deferred function, DeferReturns is populated with the
+	// addresses of calls to runtime.deferreturn in the current
+	// function. This insures that the breakpoint on the deferred
+	// function only triggers on panic or on the defer call to
+	// the function, not when the function is called directly
+	DeferCond    bool
+	DeferReturns []uint64
+	Cond         ast.Expr // When Cond is not nil the breakpoint will be triggered only if evaluating Cond returns true
 }
 
 func (bp *Breakpoint) String() string {
@@ -121,6 +132,24 @@ func (dbp *Process) writeSoftwareBreakpoint(thread *Thread, addr uint64) error {
 func (bp *Breakpoint) checkCondition(thread *Thread) (bool, error) {
 	if bp.Cond == nil {
 		return true, nil
+	}
+	if bp.DeferCond {
+		frames, err := thread.Stacktrace(2)
+		if err == nil {
+			ispanic := len(frames) >= 3 && frames[2].Current.Fn != nil && frames[2].Current.Fn.Name == "runtime.gopanic"
+			isdeferreturn := false
+			if len(frames) >= 1 {
+				for _, pc := range bp.DeferReturns {
+					if frames[0].Ret == pc {
+						isdeferreturn = true
+						break
+					}
+				}
+			}
+			if !ispanic && !isdeferreturn {
+				return false, nil
+			}
+		}
 	}
 	scope, err := thread.Scope()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


aarzilli:
The test failure on linux/go1.5 is real.

aarzilli:
I don't think the failure on windows has anything to do with this patch, the other problem is fixed.

aarzilli:
Ok, maybe the failure on windows is real too...

aarzilli:
I need assistance for windows, @lukehoban?

alexbrainman:
> I need assistance for windows,

If you want to debug TestCondBreakpoint and TestOnPrefixLocals failures, neither of them fails for me:
```
c:\Users\Alex\dev\src\github.com\derekparker\delve>git rev-parse HEAD
26f90b06daabe501d445b207ddfc602b9cc3f943

c:\Users\Alex\dev\src\github.com\derekparker\delve>git status
On branch tmp
nothing to commit, working directory clean

c:\Users\Alex\dev\src\github.com\derekparker\delve>go test github.com/derekparker/delve/terminal
ok      github.com/derekparker/delve/terminal   3.240s

c:\Users\Alex\dev\src\github.com\derekparker\delve>go test github.com/derekparker/delve/proc
ok      github.com/derekparker/delve/proc       30.287s

c:\Users\Alex\dev\src\github.com\derekparker\delve>
```

Alex

aarzilli:
Some flakiness in the windows version always existed (#363), I reran a version of #583 and that's also failing (https://ci.appveyor.com/project/derekparker/delve/build/702). Maybe something changed on Appveyor that exacerbated the tests' flakiness?

alexbrainman:
From https://ci.appveyor.com/project/derekparker/delve/build/702 - the PATH has 62 directories in it. Why so many?

Also
```
set PATH=c:\mingw64\bin;%GOPATH%\bin;%PATH%
```
it is better to add your own directories to the end of the PATH, not the start of it.

Alex

proc/breakpoints.go
```diff
@@ -33,7 +33,18 @@ type Breakpoint struct {
...
+	// if the caller is runtime.gopanic or if the return address
+	// is in the DeferReturns array.
+	// Next sets DeferCond for the breakpoint it sets on the
+	// deferred function, DeferReturns is populated with the
+	// addresses of calls to runtime.deferreturn in the current
+	// function. This insures that the breakpoint on the deferred
+	// function only triggers on panic or on the defer call to
+	// the function, not when the function is called directly
+	DeferCond    bool
+	DeferReturns []uint64
```
derekparker: From what I understand from reading https://github.com/golang/go/blob/master/src/runtime/panic.go#L269 is that there is only ever one call to `runtime.deferreturn` and, when jumping into the deferred function, it is called such that it returns not into `runtime.deferreturn` but into the caller of `runtime.deferreturn`, at the point right before it was called, essentially looping until all deferred functions are called.

As such, we shouldn't need a slice, only a single location, correct?

aarzilli: Everything correct but the compiler emits multiple calls to deferreturn, one for each exit point from the function. At least, that's true for go1.7 and go1.6. Try: `go tool objdump -s terminal.printfile ./dlv | grep deferreturn` for example.

derekparker: Got it, thanks.

aarzilli:
Since this commit has been reviewed as part of the review for #603 I'm closing this PR.

Issue #594, opened 2016-07-13
Delve getting stuck on intentional sigpanic trigger (via go-stack package)


producerism:
I'm using the [log15](https://github.com/inconshreveable/log15) package for logging, which uses the [go-stack](https://github.com/go-stack/stack) package for stack tracing.

There is a particular line/method within the go-stack package that intentionally triggers a sigpanic:

https://github.com/go-stack/stack/blob/master/stack.go#L228

Due to this line, the debugger just keeps breaking at this same place over and over, and never allows me to continue or step over it, constantly printing out "continuing" from this line:

https://github.com/derekparker/delve/blob/master/service/debugger/debugger.go#L393

Not sure where to start in troubleshooting this.

aarzilli:
What operating system are you using? Can you give me some step to reproduce this problem?
 I tried with this sample program:

```
package main

import (
	"fmt"
	
	"github.com/go-stack/stack"
)

func main() {
	s := stack.Trace().TrimRuntime()
	fmt.Printf("%v\n", s)
}
```

but I have no problem debugging it.

LeeWong:
I am facing the same issue on windows 10 and here is the output of go env

set GOARCH=amd64
set GOBIN=
set GOEXE=.exe
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=...\go
set GORACE=
set GOROOT=C:\Go
set GOTOOLDIR=C:\Go\pkg\tool\windows_amd64
set CC=gcc
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0 -fdebug-prefix-map=...\AppData\Local\Temp\go-build500025894=/tmp/go-build -gno-record-gcc-switches
set CXX=g++
set CGO_ENABLED=1

the file I try to debug is 
https://github.com/go-kit/kit/blob/master/examples/stringsvc1/main.go

The delve debug output:
 github.com/go-stack/stack.findSigpanic.func1() .../go/src/github.com/go-stack/stack/stack.go:228 (PC: 0x5faf0a)
  223:                                         }
  224:                                 }
  225:                         }
  226:                 }()
  227:                 // intentional nil pointer dereference to trigger sigpanic
> 228:                 return *p
  229:         }()
  230:         return fn
  231: }
  232:
  233: var sigpanic = findSigpanic()

aarzilli:
I'm going to assume the original issue was about windows as well.

BardoBravo:
Hi,
[...]

PR #598, opened 2016-07-20
proc: changed windows backend to deal with simultaneous breakpoints

aarzilli:
This solves most of the test flakiness that the windows back end has on multi core. 

Before this patch when multiple breakpoints were hit simultaneously only one was reported, the other one was processed by (*Thread).singleStep while stepping over the breakpoint which caused the second breakpoint to be missed and the first breakpoint to be processed twice, because (*Thread).singleStep was unable to complete.

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/598)

### DIFF
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -2087,8 +2087,8 @@ func TestStepConcurrentDirect(t *testing.T) {
 }
 
 func nextInProgress(p *Process) bool {
-	for _, th := range p.Threads {
-		if th.CurrentBreakpoint != nil && th.CurrentBreakpoint.Internal() {
+	for _, bp := range p.Breakpoints {
+		if bp.Internal() {
 			return true
 		}
 	}
@@ -2138,13 +2138,13 @@ func TestStepConcurrentPtr(t *testing.T) {
 				assertNoError(p.Continue(), t, "Continue()")
 			}
 
-			f, ln = currentLineNumber(p, t)
-			if ln != 13 {
-				t.Fatalf("Step did not step into function call (13): %s:%d (gid: %d)", f, ln, p.SelectedGoroutine.ID)
+			if p.SelectedGoroutine.ID != gid {
+				t.Fatalf("Step switched goroutines (wanted: %d got: %d)", gid, p.SelectedGoroutine.ID)
 			}
 
-			if p.SelectedGoroutine.ID != gid {
-				t.Fatalf("Step switched goroutines (%d %d)", gid, p.SelectedGoroutine.ID)
+			f, ln = currentLineNumber(p, t)
+			if ln != 13 {
+				t.Fatalf("Step did not step into function call (13): %s:%d", f, ln)
 			}
 
 			count++
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -2200,3 +2200,17 @@ func TestStepOnCallPtrInstr(t *testing.T) {
 		}
 	})
 }
+
+func TestIssue594(t *testing.T) {
+	// Exceptions that aren't caused by breakpoints should be propagated
+	// back to the target.
+	// In particular the target should be able to cause a nil pointer
+	// dereference panic and recover from it.
+	withTestProcess("issue594", t, func(p *Process, fixture protest.Fixture) {
+		assertNoError(p.Continue(), t, "Continue()")
+		f, ln := currentLineNumber(p, t)
+		if ln != 21 {
+			t.Fatalf("Program stopped at %s:%d, expected :21", f, ln)
+		}
+	})
+}
--- a/proc/proc_test.go
+++ b/proc/proc_test.go
@@ -801,7 +801,10 @@ func TestStacktraceGoroutine(t *testing.T) {
 
 		for i, g := range gs {
 			locations, err := g.Stacktrace(40)
-			assertNoError(err, t, "GoroutineStacktrace()")
+			if err != nil {
+				// On windows we do not have frame information for goroutines doing system calls.
+				continue
+			}
 
 			if stackMatch(mainStack, locations, false) {
 				mainCount++
--- a/proc/proc_windows.go
+++ b/proc/proc_windows.go
@@ -440,6 +440,7 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
 	var debugEvent _DEBUG_EVENT
 	shouldExit := false
 	for {
+		continueStatus := uint32(_DBG_CONTINUE)
 		var milliseconds uint32 = 0
 		if flags&waitBlocking != 0 {
 			milliseconds = syscall.INFINITE
@@ -496,9 +497,14 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
 		case _RIP_EVENT:
 			break
 		case _EXCEPTION_DEBUG_EVENT:
-			tid := int(debugEvent.ThreadId)
-			dbp.os.breakThread = tid
-			return tid, 0, nil
+			exception := (*_EXCEPTION_DEBUG_INFO)(unionPtr)
+			if code := exception.ExceptionRecord.ExceptionCode; code == _EXCEPTION_BREAKPOINT || code == _EXCEPTION_SINGLE_STEP {
+				tid := int(debugEvent.ThreadId)
+				dbp.os.breakThread = tid
+				return tid, 0, nil
+			} else {
+				continueStatus = _DBG_EXCEPTION_NOT_HANDLED
+			}
 		case _EXIT_PROCESS_DEBUG_EVENT:
 			debugInfo := (*_EXIT_PROCESS_DEBUG_INFO)(unionPtr)
 			exitCode = int(debugInfo.ExitCode)
@@ -508,7 +514,7 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
 		}
 
 		// .. and then continue unless we received an event that indicated we should break into debugger.
-		err = _ContinueDebugEvent(debugEvent.ProcessId, debugEvent.ThreadId, _DBG_CONTINUE)
+		err = _ContinueDebugEvent(debugEvent.ProcessId, debugEvent.ThreadId, continueStatus)
 		if err != nil {
 			return 0, 0, err
 		}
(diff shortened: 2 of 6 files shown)


proc/proc_windows.go
```diff
@@ -127,6 +127,20 @@ func Launch(cmd []string) (*Process, error) {
 		return nil, ProcessExitedError{Pid: dbp.Pid, Status: exitCode}
 	}
 
+	for _, thread := range dbp.Threads {
+		_, err := _SuspendThread(thread.os.hThread)
```
derekparker: We need to wrap this in `execPtraceFunc` as well, correct? I assume it has the same origin thread restrictions.

proc/proc_windows.go
```diff
@@ -117,7 +117,7 @@ func Launch(cmd []string) (*Process, error) {
 	// after launching under DEBUGONLYTHISPROCESS.
 	var tid, exitCode int
 	dbp.execPtraceFunc(func() {
-		tid, exitCode, err = dbp.waitForDebugEvent()
+		tid, exitCode, err = dbp.waitForDebugEvent(true)
```
alexbrainman: Unrelated to your review, but "Only the thread that created the process being debugged can call WaitForDebugEvent." (see https://msdn.microsoft.com/en-us/library/windows/desktop/ms681423(v=vs.85).aspx). The thread that calls WaitForDebugEvent here (the one running handlePtraceFuncs) is different from the thread that called CreateProcess just few lines above.

proc/proc_windows.go
```diff
@@ -127,6 +127,20 @@ func Launch(cmd []string) (*Process, error) {
 		return nil, ProcessExitedError{Pid: dbp.Pid, Status: exitCode}
 	}
 
+	for _, thread := range dbp.Threads {
```
alexbrainman: If you want to start process with all its threads suspended, you can pass CREATE_SUSPENDED into CreateProcess.

proc/proc_windows.go
```diff
@@ -370,12 +384,16 @@ func dwarfFromPE(f *pe.File) (*dwarf.Data, error) {
...
 	var debugEvent _DEBUG_EVENT
 	shouldExit := false
 	for {
+		var milliseconds uint32 = 0
+		if blocking {
+			milliseconds = syscall.INFINITE
+		}
 		// Wait for a debug event...
-		err := _WaitForDebugEvent(&debugEvent, syscall.INFINITE)
+		err := _WaitForDebugEvent(&debugEvent, milliseconds)
```
alexbrainman: I wander what the err is when milliseconds is set to 0 and you have no debug events pending.

proc/proc_windows.go
aarzilli: @lukehoban code didn't, [their documentation](https://msdn.microsoft.com/en-us/en-en/library/windows/desktop/ms686345.aspx) doesn't mention it, and it doesn't seem to cause problems.

proc/proc_windows.go
aarzilli: I wonder why this doesn't cause problems.

proc/proc_windows.go
aarzilli: It doesn't work, the first WaitForDebugEvent call never returns (I'm not saying there is no way to make it work, just that I don't know it).

proc/proc_windows.go
aarzilli: The semaphore timeout period has expired.

aarzilli:
I've found out that some of the errors that still happen are because somehow we get new threads starting while even after we suspended all existing threads in setCurrentBreakpoints, I changed waitForDebugEvent to suspend new threads in that section of the code and the only errors I still see are in TestIssue414, which isn't a very important test.

I've put those changes in a separate commit (addenda) for easier reviewing, I'll squash before merge.

proc/proc_windows.go
derekparker: It seems like maybe that is unnecessary (the first `WaitForDebugEvent`). I assume it wouldn't trigger an event as the process is simply created, but never executed any instructions. This is distinct from how we handle process creation on UNIX systems where we wait for that first `PTRACE_TRACEME` call, which triggers a signal that we wait for.

proc/proc_windows.go
```diff
@@ -475,39 +503,65 @@ func (dbp *Process) wait(pid, options int) (int, *sys.WaitStatus, error) {
...
+		_, err := _SuspendThread(thread.os.hThread)
+		if err != nil {
+			return err
+		}
+	}
+
+	for {
+		var err error
+		var tid int
+		dbp.execPtraceFunc(func() {
```
derekparker: I don't like the idea of this function having the side effect of potentially causing the inferior to execute instructions. We continue and then immediately execute a non-blocking wait, which will return an error when we do not have any pending events. That means we never stop this thread, correct?

Is is not possible to just loop through `WaitForDebugEvent` in non-blocking mode until we have no events? I assume all events would be queued and we shouldn't have to continue the inferior.

aarzilli: The loop on `dbp.Threads` calls `_SuspendTherad` on every thread so the inferior shouldn't be able to execute any instructions. 

`WaitForDebugEvent` doesn't seem to return anything unless we call `ContinueDebugEvent` in between.

proc/proc_windows.go
alexbrainman: > It doesn't work, the first WaitForDebugEvent call never returns

Fair enough. I don't know much about debugging to be useful here.

aarzilli:
rebased on master, TestIssue414 is flaky but that shouldn't be a problem after #603.

proc/proc_windows.go
```diff
@@ -507,39 +536,65 @@ func (dbp *Process) wait(pid, options int) (int, *sys.WaitStatus, error) {
...
+			return err
+		}
+	}
 
-	return trapthread.SetCurrentBreakpoint()
+	for {
+		var err error
+		var tid int
+		dbp.execPtraceFunc(func() {
+			err = _ContinueDebugEvent(uint32(dbp.Pid), uint32(dbp.os.breakThread), _DBG_CONTINUE)
```
derekparker: This is the only thing that sticks out to me in this patch set. Here, we're continuing a thread and waiting for a debug event, however in the meantime it's possible that the thread will be executing instructions. This function should not have that side effect.

A non-blocking wait for debug event should suffice, yeah? If there's no events we simply move on. I don't understand the need to continue a thread in a function that is setting state on our thread objects.

aarzilli: If we suspend all threads (which we did above this line) `ContinueDebugEvent` doesn't resume execution, it just acknowledges that we have processed the event we received. I couldn't figure out any other way of doing this that would work with the windows API. If we don't do this later on on the Continue procedure while we are singlestepping CurrentThread over its breakpoint instruction we receive events for other threads and we are not equipped to dealing with them there.

derekparker: Ah, so if I understand correctly, it behaves similar to the Darwin API, where you can suspend a thread multiple times, and have to resume an equal number of times for it to actually execute instructions. Is that correct, from your understanding? If so, could you add a comment / link to docs to make it clear we're not actually continuing execution.

aarzilli: Yes, it's like the darwin API, however I don't actually ever suspend the thread more than once, when one of the threads generates a debug event windows will halt the execution of all threads but this state doesn't count as suspended, it's weird.

aarzilli: I added a comment describing what we are doing on that function.

proc/proc_windows.go
```diff
@@ -112,7 +117,7 @@ func newDebugProcess(pid int, exepath string) (*Process, error) {
 	var err error
 	var tid, exitCode int
 	dbp.execPtraceFunc(func() {
-		tid, exitCode, err = dbp.waitForDebugEvent()
+		tid, exitCode, err = dbp.waitForDebugEvent(true, false)
```
derekparker: nit: Passing booleans in as parameters makes it hard to know what they mean when reading at the caller. Would do you think about named constants or using a bitmap flag? E.g. `opts := WAIT_NONBLOCK | SUSPEND_NEW_THREADS`?

proc/proc_windows.go
derekparker: Thanks. I think maybe we should alias `_DBG_CONTINUE` to `_DBG_EXCEPTION_HANDLED`. When reading up on the Windows debug API the other day, IIRC, that existed for 32 bit, but not 64 bit. However, semantically it's more accurate to what we're trying to accomplish, and makes it more explicit from the callers point of view.

derekparker:
Overall looks good. Can you rebase this on master so we can pull in the patch fixing prologue detection and get the AppVeyor CI to pass?

proc/proc_windows.go
aarzilli: Googling DBG_EXCEPTION_HANDLED leads me to [this blog post](https://breakingcode.wordpress.com/2009/08/05/dbg_continue-vs-dbg_exception_handled/) which suggests that at some point DBG_EXCEPTION_HANDLED and DBG_CONTINUE did different things and had different values, I think appropriating the name as a synonym muddles the water but if you insist I'll switch.

aarzilli:
Rebased, the only failures I see (and the one in Appveyor) is connected to the Step code, I will debug it after #603, assuming it persists.

aarzilli:
I added a commit that fixes #594, I can split it out if you want to review it separately, it's pretty simple.

proc/proc_test.go
```diff
@@ -778,7 +778,9 @@ func TestStacktraceGoroutine(t *testing.T) {
 
 		for i, g := range gs {
 			locations, err := g.Stacktrace(40)
-			assertNoError(err, t, "GoroutineStacktrace()")
+			if err != nil {
```
derekparker: This needs a comment / explanation / TODO about why we're ignoring errors.

proc/proc_windows.go
```diff
@@ -121,6 +126,20 @@ func newDebugProcess(pid int, exepath string) (*Process, error) {
 		dbp.postExit()
 		return nil, ProcessExitedError{Pid: dbp.Pid, Status: exitCode}
 	}
+	for _, thread := range dbp.Threads {
```
derekparker: Need to add a comment here as to why we're suspending and then continuing. It's hard to tell from here that we're trying to drain all events.

It's probably worth putting this logic into a well named function so it's apparent to the reader what's going on.

derekparker:
Overall LGTM, just a few comments. After those are addressed I will merge this.

aarzilli:
Done.

aarzilli:
Rebased on master, I can't replicate the failure I see in appveyor, someone that can will have to fix that. I still think this is an improvement and should be merged.

PR #607, opened 2016-08-04
proc: implement attach on windows

alexbrainman:
This change implements attach command on windows

Updates #363

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/607)

### DIFF
--- a/proc/proc_windows.go
+++ b/proc/proc_windows.go
@@ -8,21 +8,15 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"sync"
 	"syscall"
 	"unsafe"
 
 	sys "golang.org/x/sys/windows"
 
-	"golang.org/x/debug/dwarf"
 	"github.com/derekparker/delve/dwarf/frame"
 	"github.com/derekparker/delve/dwarf/line"
-)
-
-const (
-	// DEBUGONLYTHISPROCESS tracks https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx
-	DEBUGONLYTHISPROCESS = 0x00000002
+	"golang.org/x/debug/dwarf"
 )
 
 // OSProcessDetails holds Windows specific information.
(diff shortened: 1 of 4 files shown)


derekparker:
Awesome, thanks for this! I'll do a full review sometime this afternoon, in the meantime @lukehoban mind reading through this?

proc/proc_windows.go
```diff
@@ -126,13 +121,56 @@ func Launch(cmd []string) (*Process, error) {
...
+}
+
+// findExePath searches for process pid, and returns its executable path.
+func findExePath(pid int) (string, error) {
+	// Original code suggested different approach (see below).
+	// Maybe it could be useful in the future.
+	//
+	// Find executable path from PID/handle on Windows:
+	// https://msdn.microsoft.com/en-us/library/aa366789(VS.85).aspx
+	//
```
derekparker: nit: remove this line.

proc/proc_windows.go
```diff
@@ -126,13 +121,56 @@ func Launch(cmd []string) (*Process, error) {
...
+		default:
+			return "", err
+		}
+	}
 }
 
 // Attach to an existing process with the given PID.
 func Attach(pid int) (*Process, error) {
-	return nil, fmt.Errorf("not implemented: Attach")
+	// TODO: Probably should have SeDebugPrivilege before starting here.
```
derekparker: Does this apply to the exec path as well? Is it possible we can check for that, and return an appropriate error?

derekparker:
Overall looks good, just a nit and question. Thanks for this!

proc/proc_windows.go
alexbrainman: I am not clear what you are asking about. If you asking if we need to hold SeDebugPrivilege when we are starting debugged process ourselves (as well as when attaching to external process), then I do not know. CreateProcess Windows API does not mention SeDebugPrivilege, while DebugActiveProcess does.

I played around with new "attach" functionality, and I can attach to external process (small Go program) with code in this PR. So it seems I don't always need to ask for SeDebugPrivilege. But I have no doubt we will find ourselves in a situation when this is required. So I have added a TODO, so we don't forget. I suspect we could try acquire SeDebugPrivilege once at the start of the program, and continue with our job regardless of result - maybe display warning message if SeDebugPrivilege could not be acquired.

proc/proc_windows.go
alexbrainman: I have just done that. I have changed my local commit and pushed it (with -f) into my fork https://github.com/alexbrainman/delve/commit/f09ef23f9ec650f18f2530d7164229ea4a119700 I don't understand why this pull request does not see my change. I don't use PRs. Please help.

PR #611, opened 2016-08-11
docs: Description for cli adding arguments

rdleon:
Added a description for passing command line arguments to the package being debugged.

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/611)


derekparker:
Hey, thanks for the patch, however that document is auto-generated. What you'll want to do instead is update: https://github.com/derekparker/delve/blob/master/cmd/dlv/cmds/commands.go#L125 and then regenerate the docs with: `go run scripts/gen-cli/docs.go`.

rdleon:
Thanks, doing that and reviewing the contribute guidelines.

PR #619, opened 2016-08-23
cmd: Support space separate flags for build-flags

zchee:
Currently, `dlv` not support multiple build flags such as `-v -x`.
If set `--build-flags='-v -x'`, will occur `flag provided but not defined: -v -x` error from go compiler because append the `-v` and `-x` does not separate by space.
Also, `--build-flags=-v --build-flags=-x` is recognizes `-x` only.

This pull request will split `BuildFlags` variable to `[]string` slice uses `string.Fields`.

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/619)

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/derekparker/delve/config"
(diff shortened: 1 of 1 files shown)


aarzilli:
LGTM

zchee:
@derekparker @aarzilli Sorry, just ping comment.

derekparker:
@zchee thanks for the patch, sorry for the delay.

zchee:
@derekparker np! thanks great debug tool :)

BTW, I developping the Neovim plugin that implemented inline debug (like sublime, ms code) uses delve written in pure Go.
I was vendoring and use the delve internal code(not `internal` dir, internal API), it's no problem?

Just in case, that's is, If implemented d lve core debugger, also needs codesign to my binary on macOS.
So I choice, exec dlv binary on background, and connecting it with RPC uses delve sources.

Now still private repository because it's learning Go and vim script for me, but will publish publicly.
If have problem, I will consider to another approach or remove this implementation.
What do you think?

aarzilli:
> So I choice, exec dlv binary on background, and connecting it with RPC uses delve sources.

I think you have made the right choice: you can use `service/rpc2.(RPCClient)` as long as you vendor it, we are trying to keep backwards compatibility, if something breaks it's our fault.

PR #632, opened 2016-09-12
proc: Names of concrete types of interfaces parsing their runtime._type

aarzilli:
Generate names of the concrete types stored inside interface variables
by fully parsing their runtime._type instead of simply using the str
field.

This allows delve to read the contents of an interface variable when
the program imports multiple packages that have the same name. It also
allows delve to correctly interpret some complex anonymous types.

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/632)

### DIFF
--- a/proc/moduledata.go
+++ b/proc/moduledata.go
@@ -2,11 +2,13 @@ package proc
 
 import (
 	"go/constant"
+	"unsafe"
 )
 
 // delve counterpart to runtime.moduledata
 type moduleData struct {
 	types, etypes uintptr
+	typemapVar    *Variable
 }
 
 func (dbp *Process) loadModuleData() (err error) {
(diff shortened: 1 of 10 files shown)


proc/moduledata.go
```diff
@@ -51,37 +56,135 @@ func (dbp *Process) loadModuleData() (err error) {
...
+		if typeAddr >= dbp.moduleData[i].types && typeAddr < dbp.moduleData[i].etypes {
+			md = &dbp.moduleData[i]
+		}
+	}
+
+	rtyp, err := dbp.findType("runtime._type")
+	if err != nil {
+		return nil, err
+	}
+	//prtyp := dbp.pointerTo(rtyp)
```
derekparker: Leftover comment?

proc/moduledata.go
```diff
@@ -51,37 +56,135 @@ func (dbp *Process) loadModuleData() (err error) {
...
+
+	rawstr, err := dbp.CurrentThread.readMemory(off, int(namelen))
+	off += uintptr(namelen)
+	if err != nil {
+		return "", "", 0, err
+	}
+
+	name = string(rawstr)
+
+	const (
```
derekparker: nit: maybe move this right about this function.

aarzilli:
I added a comment describing how the more obscure part of the patch works.

PR #640, opened 2016-09-17
Add substitute-path parameter to configuration file.

rustyrobot:
Allows to rewrite a source path stored in program's debug information, if the sources were moved to a different place between compilation and debugging.

Implements similar logic as `substitute-path` [in gdb](https://sourceware.org/gdb/onlinedocs/gdb/Source-Path.html).

---
This change is [Reviewable](https://reviewable.io/reviews/derekparker/delve/640)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,9 +15,23 @@ const (
 	configFile string = "config.yml"
 )
 
+// Describes a rule for substitution of path to source code file.
+type SubstitutePathRule struct {
+	// Directory path will be substituted if it matches `From`.
+	From string
+	// Path to which substitution is performed.
+	To   string
+}
+
+// Slice of source code path substitution rules.
+type SubstitutePathRules []SubstitutePathRule
+
 // Config defines all configuration options available to be set through the config file.
 type Config struct {
-	Aliases map[string][]string
+	// Commands aliases.
+	Aliases        map[string][]string
+	// Source code path substitution rules.
+	SubstitutePath SubstitutePathRules `yaml:"substitute-path"`
 }
 
 // LoadConfig attempts to populate a Config object from the config.yml file.
@@ -89,6 +103,14 @@ func writeDefaultConfig(f *os.File) error {
 # Provided aliases will be added to the default aliases for a given command.
 aliases:
   # command: ["alias1", "alias2"]
+
+# Define sources path substitution rules. Can be used to rewrite a source path stored
+# in program's debug information, if the sources were moved to a different place
+# between compilation and debugging.
+# Note that substitution rules will not be used for paths passed to "break" and "trace"
+# commands.
+substitute-path:
+  # - {from: path, to: path}
 `)
 	return err
 }
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"runtime"
 	"strings"
 
 	"syscall"
@@ -24,6 +25,7 @@ const (
 // Term represents the terminal running dlv.
 type Term struct {
 	client   service.Client
+	conf     *config.Config
 	prompt   string
 	line     *liner.State
 	cmds     *Commands
@@ -49,9 +51,10 @@ func New(client service.Client, conf *config.Config) *Term {
 	}
 
 	return &Term{
+		client: client,
+		conf:   conf,
 		prompt: "(dlv) ",
 		line:   liner.NewLiner(),
-		client: client,
 		cmds:   cmds,
 		dumb:   dumb,
 		stdout: w,
@@ -148,6 +151,46 @@ func (t *Term) Println(prefix, str string) {
 		prefix = fmt.Sprintf("%s%s%s", terminalBlueEscapeCode, prefix, terminalResetEscapeCode)
 	}
 	fmt.Fprintf(t.stdout, "%s%s\n", prefix, str)
+}
+
+// Substitues directory to source file.
+//
+// Ensures that only directory is substitued, for example:
+// substitute from `/dir/subdir`, substitute to `/new`
+// for file path `/dir/subdir/file` will return file path `/new/file`.
+// for file path `/dir/subdir-2/file` substitution will not be applied.
+//
+// If more than one substitution rule is defined, the rules are applied
+// in the order they are defined, first rule that matches is used for
+// substitution.
+func (t *Term) substitutePath(path string) string {
+	path = crossPlatformPath(path)
+	if t.conf == nil {
+		return path
+	}
+	separator := string(os.PathSeparator)
+	for _, r := range t.conf.SubstitutePath {
+		from := crossPlatformPath(r.From)
+		to := r.To
+
+		if !strings.HasSuffix(from, separator) {
+			from = from + separator
+		}
+		if !strings.HasSuffix(to, separator) {
+			to = to + separator
+		}
+		if strings.HasPrefix(path, from) {
+			return strings.Replace(path, from, to, 1)
+		}
+	}
+	return path
+}
+
+func crossPlatformPath(path string) string {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+		return strings.ToLower(path)
+	}
+	return path
 }
 
 func (t *Term) promptForInput() (string, error) {
--- a/terminal/terminal_test.go
+++ b/terminal/terminal_test.go
@@ -1,0 +1,79 @@
+package terminal
+
+import (
+	"testing"
+	"runtime"
+
+	"github.com/derekparker/delve/config"
+)
+
+type tRule struct {
+	from string
+	to string
+}
+
+type tCase struct {
+	rules []tRule
+	path  string
+	res   string
+}
+
+func platformCases() []tCase {
+	casesUnix := []tCase{
+		// Should not depend on separator at the end of rule path
+		{[]tRule{{"/tmp/path", "/new/path2"}}, "/tmp/path/file.go", "/new/path2/file.go"},
+		{[]tRule{{"/tmp/path/", "/new/path2/"}}, "/tmp/path/file.go", "/new/path2/file.go"},
+		{[]tRule{{"/tmp/path/", "/new/path2"}}, "/tmp/path/file.go", "/new/path2/file.go"},
+		{[]tRule{{"/tmp/path", "/new/path2/"}}, "/tmp/path/file.go", "/new/path2/file.go"},
+		// Should apply only for directory names
+		{[]tRule{{"/tmp/path", "/new/path2"}}, "/tmp/path-2/file.go", "/tmp/path-2/file.go"},
+		// First matched rule should be used
+		{[]tRule{
+			{"/tmp/path1", "/new/path1"},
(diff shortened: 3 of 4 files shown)


terminal/terminal.go
```diff
@@ -150,6 +167,34 @@ func (t *Term) Println(prefix, str string) {
...
+// Ensures that only directory is substitued, for example:
+// substitute from `/dir/subdir`, substitute to `/new`
+// for file path `/dir/subdir/file` will return file path `/new/file`.
+// for file path `/dir/subdir-2/file` substitution will not be applied.
+//
+// If more than one substitution rule is defiend, the rules are applied
+// in the order they are defined, first rule that matches is used for
+// substitution.
+func (t *Term) SubstitutePath(path string) string {
+	separator := string(os.PathSeparator)
```
alexbrainman: This will not work on windows. See path/filepath package for functions to manipulate file paths. I would also add new test to verify your changes actually work now and in the future.

rustyrobot: @alexbrainman Could you please provide a bit more information, on what exactly will not work on windows? (don't have windows to test).

> I would also add new test to verify your changes actually work now and in the future.

Sure, [I will add tests](https://github.com/derekparker/delve/pull/640#issuecomment-247786492).

derekparker:
@rustyrobot thanks for the patch, I'll take a look today, sorry for the delay.

terminal/terminal.go
alexbrainman: > Could you please provide a bit more information, on what exactly will not work on windows?

For example your line 191 uses strings.HasPrefix to compare beginning of two paths. But windows paths are case insensitive. So the user can provide either from="c:\dir\subdir" or from="C:\DIR\SUBDIR" (even from="c:\DiR\SuBdIr" should work) and it should work.

rustyrobot: Thanks, forgot about this difference, will fix that.

rustyrobot: @alexbrainman I've looked at this issue more carefully, case sensitivity depends on file system, it's tricky to do cross-platform detection to determine if file system is case sensitive or not.
Do you have some hints on that? I'm not sure if it's ok to do it based on the type of operating system (but this would be [consistent with gdb](https://github.com/bminor/binutils-gdb/blob/master/libiberty/filename_cmp.c#L127-L130)).

alexbrainman: > case sensitivity depends on file system

I have never used Windows OS that cared about file path case. I do not know what else I can say.

> Do you have some hints on that?

I would always assume that file path is case insensitive on Windows.

> I'm not sure if it's ok to do it based on the type of operating system (but this would be consistent with gdb).

Converting to lower case on Windows before comparing strings sounds good to me.

There is also path/filepath.EvalSymlinks which will convert your file path to the same form (regardless case and everything else). But filepath.EvalSymlinks works with real files only - the file must exist.

Alex

terminal/terminal.go
```diff
@@ -150,6 +157,50 @@ func (t *Term) Println(prefix, str string) {
...
+}
+
+func crossPlatformPath(path string) string {
+	if isCaseSensitive() {
+		return path
+	}
+	return strings.ToLower(path)
+}
+
+func isCaseSensitive() bool {
```
alexbrainman: I would just inline isCaseSensitive. crossPlatformPath is not large enough.

terminal/terminal.go
```diff
@@ -150,6 +157,50 @@ func (t *Term) Println(prefix, str string) {
...
+	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
```
alexbrainman: Are you sure "darwin" is like "windows"?

terminal/terminal_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+	"runtime"
+
+	"github.com/derekparker/delve/config"
+)
+
+func TestSubstitutePath(t *testing.T) {
+	type cases []struct {
+		rules [][]string
```
alexbrainman: Why not
rules []struct {
  from, to string
}
?

terminal/terminal_test.go
```diff
@@ -0,0 +1,73 @@
...
+		{[][]string{{"c:\\tmp\\path\\", "d:\\new\\path2\\"}}, "c:\\tmp\\path\\file.go", "d:\\new\\path2\\file.go"},
+		{[][]string{{"c:\\tmp\\path", "d:\\new\\path2\\"}}, "c:\\tmp\\path\\file.go", "d:\\new\\path2\\file.go"},
+		{[][]string{{"c:\\tmp\\path\\", "d:\\new\\path2"}}, "c:\\tmp\\path\\file.go", "d:\\new\\path2\\file.go"},
+		// Should apply only for directory names
+		{[][]string{{"c:\\tmp\\path", "d:\\new\\path2"}}, "c:\\tmp\\path-2\\file.go", "c:\\tmp\\path-2\\file.go"},
+		// Should be case-insensitive
+		{[][]string{{"c:\\tmp\\path", "d:\\new\\path2"}}, "C:\\TmP\\PaTh\\file.go", "d:\\new\\path2\\file.go"},
+	}
+
+	var platformCases cases
```
alexbrainman: I would create a separate function that returns platformCases. You could use "return" statement to return for every OS. You would not need to have this multilevel if statement, and will keep all code in one straight line.

terminal/terminal_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestSubstitutePath(t *testing.T) {
+	type cases []struct {
+		rules [][]string
+		path  string
+		res   string
+	}
+
+	casesUnix := cases{
```
alexbrainman: Maybe do
... := []case{
instead. Code reader can see that it is just an array.

terminal/terminal_test.go
```diff
@@ -0,0 +1,73 @@
...
+		// Should be case-sensitive
+		{[][]string{{"/tmp/path", "/new/path2"}}, "/TmP/path/file.go", "/TmP/path/file.go"},
+	}
+	casesDarwin := cases{
+		// Should be case-insensitive
+		{[][]string{{"/tmp/path", "/new/path2"}}, "/TmP/PaTh/file.go", "/new/path2/file.go"},
+	}
+	casesWindows := cases{
+		// Should not depend on separator at the end of rule path
+		{[][]string{{"c:\\tmp\\path", "d:\\new\\path2"}}, "c:\\tmp\\path\\file.go", "d:\\new\\path2\\file.go"},
```
alexbrainman: I would
s/"c:\\tmp\\path"/`c:\tmp\path`/
So there are no double back slashes.

terminal/terminal.go
rustyrobot: Fixed.

terminal/terminal.go
rustyrobot: Fixed.

terminal/terminal.go
rustyrobot: Yes, Mac's HFS is case insensitive by default, and it's configured this way in [gdb](https://github.com/T-J-Teru/binutils-gdb/blob/fdbd04a8adef3a48eeac7702e9243c7c31e47256/include/filenames.h#L45-L50).

terminal/terminal_test.go
rustyrobot: Fixed, I'm not completely sure if it's what you wanted, feel free to comment new code.

terminal/terminal_test.go
rustyrobot: Fixed.

terminal/terminal_test.go
rustyrobot: Fixed.

terminal/terminal_test.go
rustyrobot: Fixed.

terminal/terminal_test.go
alexbrainman: LGTM. Your test also passes on my Windows computer.

Alex

config/config.go
```diff
@@ -15,9 +15,17 @@ const (
 	configFile string = "config.yml"
 )
 
+type SubstitutePathRule struct {
```
derekparker: Exported types need doc comment.

config/config.go
```diff
@@ -15,9 +15,17 @@ const (
...
+	From string
```
derekparker: Same with exported fields.

config/config.go
```diff
@@ -15,9 +15,17 @@ const (
...
+	To   string
+}
+
+type SubstitutePathRules []SubstitutePathRule
```
derekparker: Needs doc comment.

config/config.go
```diff
@@ -15,9 +15,17 @@ const (
...
+
 // Config defines all configuration options available to be set through the config file.
 type Config struct {
-	Aliases map[string][]string
+	Aliases        map[string][]string
```
derekparker: Exported fields need doc comments.

terminal/terminal.go
```diff
@@ -150,6 +157,43 @@ func (t *Term) Println(prefix, str string) {
 	fmt.Fprintf(t.stdout, "%s%s\n", prefix, str)
 }
 
+// SubstitutePath substitues directory to soruce file.
```
derekparker: s/soruce/source/

terminal/terminal.go
```diff
@@ -150,6 +157,43 @@ func (t *Term) Println(prefix, str string) {
...
+//
+// Ensures that only directory is substitued, for example:
+// substitute from `/dir/subdir`, substitute to `/new`
+// for file path `/dir/subdir/file` will return file path `/new/file`.
+// for file path `/dir/subdir-2/file` substitution will not be applied.
+//
+// If more than one substitution rule is defiend, the rules are applied
```
derekparker: s/defiend/defined/

terminal/terminal.go
```diff
@@ -150,6 +157,43 @@ func (t *Term) Println(prefix, str string) {
...
+// in the order they are defined, first rule that matches is used for
+// substitution.
+func (t *Term) SubstitutePath(path string) string {
```
derekparker: nit: Doesn't need to be exported.

terminal/terminal.go
```diff
@@ -23,20 +24,25 @@ const (
...
 }
 
 // New returns a new Term.
 func New(client service.Client, conf *config.Config) *Term {
+	var substitutePathRules config.SubstitutePathRules
 	cmds := DebugCommands(client)
-	if conf != nil && conf.Aliases != nil {
-		cmds.Merge(conf.Aliases)
+	if conf != nil {
+		substitutePathRules = conf.SubstitutePath
```
aarzilli: IMHO we should stick the entire conf object inside cmds.

rustyrobot: Could you please elaborate on that? As far as I understand it's not really related to `conf.SubstitutePath` but to the way `cmds` is created, do you want `conf` to be passed as an argument for `DebugCommands` and to do `Merge` inside?

rustyrobot: Done.

derekparker:
Sorry for the delay in merging this, has been a busy few weeks. Thanks for this patch!

PR #521, opened 2016-04-29
[WIP] ability to set the working directory

ctaggart:
This is work in progress. These are the change from @yuntan in #295 to make it work on Linux.

Issue #295, opened 2015-11-19
flag to specify workdir of debug binary


yuntan:
I want command line flag to specify working directory of the program, for example

```
dlv --wd=/path/to/workdir debug
```

runs debug binary in /path/to/workdir.

yuntan:
This changes works for me (on linux).
https://github.com/yuntan/delve/compare/master...yuntan:90ec527
But still work in progress. Remaining tasks

- [ ] work for darwin
- [ ] work for exec, test subcommands
- [ ] write some tests

ctaggart:
@yuntan What do we need to do to move this forward? It would be super helpful to have when using Visual Studio Code with Go. Can you create a pull request and mark it as work in progress. I can test it in Mac on my iMac.

PR #650, opened 2016-10-07
Flag to set working directory

rustyrobot:
Continuation of pull request #521, adds darwin support and test.

Fixes #521 #295


proc/proc_windows.go
```diff
@@ -26,7 +26,7 @@ type OSProcessDetails struct {
 }
 
 // Launch creates and begins debugging a new process.
-func Launch(cmd []string) (*Process, error) {
+func Launch(cmd []string, wd string) (*Process, error) {
```
alexbrainman: It looks like we are rewriting os/exec.Cmd. Can we just use os/exec.Cmd here instead?

proc/proc_test.go
```diff
@@ -2197,3 +2201,25 @@ func TestStepOnCallPtrInstr(t *testing.T) {
...
 }
+
+func TestWorkDir(t *testing.T) {
+	wd := "/tmp"
+	if runtime.GOOS == "windows" {
+		wd = `C:\Windows\TEMP`
+	}
+	if runtime.GOOS == "darwin" {
+		wd = "/private/tmp"
+	}
```
alexbrainman: Why not

wd := os.TempDir()

?

rustyrobot: Fixed for `windows` and `linux`, but it's a bit trickier for `darwin` added a comment.

proc/proc_windows.go
alexbrainman: Sorry @rustyrobot, I was talking to myself.

proc/proc_test.go
```diff
@@ -32,8 +32,12 @@ func TestMain(m *testing.M) {
 }
 
 func withTestProcess(name string, t testing.TB, fn func(p *Process, fixture protest.Fixture)) {
+	withTestProcessLaunchArgs(name, t, ".", fn)
+}
+
+func withTestProcessLaunchArgs(name string, t testing.TB, wd string, fn func(p *Process, fixture protest.Fixture)) {
```
aarzilli: I think it would be better to add a `wd string` argument to `withTestProcessArgs` instead of making another helper function.

proc/proc_windows.go
aarzilli: @alexbrainman A quick grep through go source suggest no way to pass DEBUG_ONLY_THIS_PROCESS to a newly created process, if os/exec used syscall.SysProcAttr.Ptrace to set DEBUG_ONLY_THIS_PROCESS on windows we could probably use it. We do use it on linux.

service/config.go
```diff
@@ -13,6 +13,10 @@ type Config struct {
 	Listener net.Listener
 	// ProcessArgs are the arguments to launch a new process.
 	ProcessArgs []string
+	// Wd is working directory of the new process. This field is used only when
+	// ProcessArgs is specified.
```
aarzilli: Technically this comment is correct but it sounds pretty weird, I'd suggest: `s/only when ProcessArgs is specified/only when launching a new process./`

service/debugger/debugger.go
```diff
@@ -38,6 +38,10 @@ type Debugger struct {
 type Config struct {
 	// ProcessArgs are the arguments to launch a new process.
 	ProcessArgs []string
+	// Wd is working directory of the new process. This field is used only when
+	// ProcessArgs is specified.
```
aarzilli: Same as in service/config.go

service/config.go
alexbrainman: @aarzilli, if I am not wrong, you should be able to pass DEBUG_ONLY_THIS_PROCESS in syscall.SysProcAttr.CreationFlags. Would you like me to try and change code once this PR is submitted?

aarzilli: @alexbrainman: interesting. Yes, we should try that after this PR.

service/debugger/debugger.go
```diff
@@ -38,6 +38,10 @@ type Debugger struct {
 type Config struct {
 	// ProcessArgs are the arguments to launch a new process.
 	ProcessArgs []string
+	// Wd is working directory of the new process. This field is used
+	// only when launching a new process.
+	Wd string
```
derekparker: nit: Change to `WorkingDir`

service/config.go
```diff
@@ -13,6 +13,10 @@ type Config struct {
 	Listener net.Listener
 	// ProcessArgs are the arguments to launch a new process.
 	ProcessArgs []string
+	// Wd is working directory of the new process. This field is used
+	// only when launching a new process.
+	Wd string
```
derekparker: nit: Change to `WorkingDir`

cmd/dlv/cmds/commands.go
```diff
@@ -39,6 +39,8 @@ var (
 	InitFile string
 	// BuildFlags is the flags passed during compiler invocation.
 	BuildFlags string
+	// Wd is the working directory for running the program.
+	Wd string
```
derekparker: nit: Change to `WorkingDir`

derekparker:
Also needs rebase. Thanks @rustyrobot for picking this up!

PR #661, opened 2016-10-28
terminal/commands/bt: don't panic on goroutines with only one stackframe

pborzenkov:
If a goroutine has only one stack frame, like this:

(dlv) bt
0  0x00000000000d0621 in runtime.goexit
   at /usr/local/Cellar/go/1.7.3/libexec/src/runtime/asm_amd64.s:2087

digits() is called with n == 0 and math.Log10 returns -Inf, which later,
after conversion to int, causes panic() in strings.Repeat (because it
tries to allocate huge memory area).

Signed-off-by: Pavel Borzenkov <pavel.borzenkov@gmail.com>

### DIFF
--- a/terminal/command.go
+++ b/terminal/command.go
@@ -819,6 +819,9 @@ func (c *Commands) sourceCommand(t *Term, args string) error {
 }
 
 func digits(n int) int {
+	if n <= 0 {
+		return 1
+	}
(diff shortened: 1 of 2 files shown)


pborzenkov:
Oops, looks like it's already fixed on master. And I had v0.11.0-alpha checked-out.
Sorry for the noise.

PR #667, opened 2016-11-06
Go 1.8 compatibility (part 2)

aarzilli:
Do not commit / Do not review.

### DIFF
--- a/dwarf/line/line_parser_test.go
+++ b/dwarf/line/line_parser_test.go
@@ -38,6 +38,13 @@ func grabDebugLineSection(p string, t *testing.T) []byte {
 	return data
 }
 
+const (
+	lineBaseGo14  int8  = -1
+	lineBaseGo18  int8  = -4
+	lineRangeGo14 uint8 = 4
+	lineRangeGo18 uint8 = 10
+)
+
 func TestDebugLinePrologueParser(t *testing.T) {
 	// Test against known good values, from readelf --debug-dump=rawline _fixtures/testnextprog
 	p, err := filepath.Abs("../../_fixtures/testnextprog")
@@ -67,11 +74,15 @@ func TestDebugLinePrologueParser(t *testing.T) {
 		t.Fatal("Initial value of 'is_stmt' not parsed correctly", prologue.InitialIsStmt)
 	}
 
-	if prologue.LineBase != int8(-1) {
+	if prologue.LineBase != lineBaseGo14 && prologue.LineBase != lineBaseGo18 {
+		// go < 1.8 uses -1
+		// go >= 1.8 uses -4
 		t.Fatal("Line base not parsed correctly", prologue.LineBase)
 	}
 
-	if prologue.LineRange != uint8(4) {
+	if prologue.LineRange != lineRangeGo14 && prologue.LineRange != lineRangeGo18 {
+		// go < 1.8 uses 4
+		// go >= 1.8 uses 10
 		t.Fatal("Line Range not parsed correctly", prologue.LineRange)
 	}
 
@@ -90,8 +101,15 @@ func TestDebugLinePrologueParser(t *testing.T) {
 		t.Fatal("Include dirs not parsed correctly")
 	}
 
-	if !strings.Contains(dbl.FileNames[0].Name, "/delve/_fixtures/testnextprog.go") {
-		t.Fatal("First file entry not parsed correctly")
+	ok := false
+	for _, n := range dbl.FileNames {
+		if strings.Contains(n.Name, "/delve/_fixtures/testnextprog.go") {
+			ok = true
+			break
+		}
+	}
+	if !ok {
+		t.Fatal("File names table not parsed correctly")
 	}
 }
 
--- a/dwarf/line/state_machine.go
+++ b/dwarf/line/state_machine.go
@@ -27,6 +27,11 @@ type StateMachine struct {
 	endSeq          bool
 	lastWasStandard bool
 	lastDelta       int
+	// if valid is true the current value of the state machine is the address of
+	// an instruction (using the terminology used by DWARF spec the current
+	// value of the state machine should be appended to the matrix representing
+	// the compilation unit)
+	valid bool
 }
 
 type opcodefn func(*StateMachine, *bytes.Buffer)
@@ -91,7 +96,9 @@ func (dbl *DebugLines) AllPCsForFileLine(f string, l int) (pcs []uint64) {
 		}
 		if sm.line == l && sm.file == f && sm.address != lastAddr {
 			foundFile = true
-			pcs = append(pcs, sm.address)
+			if sm.valid {
+				pcs = append(pcs, sm.address)
+			}
 			line := sm.line
 			// Keep going until we're on a different line. We only care about
 			// when a line comes back around (i.e. for loop) so get to next line,
@@ -123,6 +130,9 @@ func (dbl *DebugLines) AllPCsBetween(begin, end uint64, filename string) ([]uint
 
 	for b, err := buf.ReadByte(); err == nil; b, err = buf.ReadByte() {
 		findAndExecOpcode(sm, buf, b)
+		if !sm.valid {
+			continue
+		}
 		if sm.address > end {
 			break
 		}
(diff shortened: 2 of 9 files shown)


aarzilli:
Partially blocked by https://github.com/golang/go/issues/17830

aarzilli:
The go issue has been fixed, this can be reviewed now.

dwarf/line/state_machine.go
```diff
@@ -151,9 +161,10 @@ func execSpecialOpcode(sm *StateMachine, instr byte) {
 
 	sm.lastDelta = int(sm.dbl.Prologue.LineBase + int8(decoded%sm.dbl.Prologue.LineRange))
 	sm.line += sm.lastDelta
-	sm.address += uint64(decoded / sm.dbl.Prologue.LineRange)
+	sm.address += uint64(decoded/sm.dbl.Prologue.LineRange) * uint64(sm.dbl.Prologue.MinInstrLength)
```
dr2chase: I'm doubtful about this change.  I ended up with this, instead:
```
                sm.isStmt = true
        }
 
-       sm.lastDelta = int(sm.dbl.Prologue.LineBase + int8(decoded%sm.dbl.Prologue.LineRange))
+       var deltapc uint64
+       var deltaline int
+
+       switch opcode {
+       case 249:
+               deltaline = 5
+               deltapc = 23
+       case 255:
+               deltaline = 1
+               deltapc = 24
+       default:
+               deltaline = int(sm.dbl.Prologue.LineBase + int8(decoded%sm.dbl.Prologue.LineRange))
+               deltapc = uint64(decoded / sm.dbl.Prologue.LineRange)
+       }
+
+       sm.lastDelta = deltaline
        sm.line += sm.lastDelta
-       sm.address += uint64(decoded / sm.dbl.Prologue.LineRange)
+       sm.address += deltapc
        sm.basicBlock = false
        sm.lastWasStandard = false
```
I got this from https://go.googlesource.com/go/+/f9feaffdf51967bc5cf3c9363db9ddee98c9b3a0/src/cmd/link/internal/ld/dwarf.go#918

```
		} else {
			// Use opcode 249 (pc+=23, lc+=5) or 255 (pc+=24, lc+=1).
			//
```
and I don't see any way that the change proposed here takes us from a state where the PC increments were not 23 and 24 to one where they are (because 23 and 24 have no common factors other than 1).

aarzilli: This change is just a formality MinInstrLen, at least on amd64, is always 1. AFAIK delve always interpreted special opcodes correctly:

```
objdump --dwarf ./testvariables2 | grep -A11 'debug_line' 
Raw dump of debug contents of section .debug_line:

  Offset:                      0x0
  Length:                      76002
  DWARF Version:               2
  Prologue Length:             7461
  Minimum Instruction Length:  1
  Initial value of 'is_stmt':  1
  Line Base:                   -4
  Line Range:                  10
  Opcode Base:                 10

```

plugging these values into that code we get https://play.golang.org/p/PhTBbfJ8GX :

```249 = addr 23 line 5
255 = addr 24 line 1
```

I don't know why link would need a special branch to handle 249 and 255. What we interpreted wrong was the standard opcode DW_LNS_const_add_pc, which wasn't emitted (or was only emitted rarely) by previous versions of link.

dwarf/line/state_machine.go
```diff
@@ -26,6 +26,11 @@ type StateMachine struct {
 	endSeq          bool
 	lastWasStandard bool
 	lastDelta       int
+	// if valid is true the current value of the state machine is the addres of
```
derekparker: nit: s/addres/address/

aarzilli: Done.

proc/proc_test.go
```diff
@@ -2374,8 +2373,25 @@ func TestIssue683(t *testing.T) {
 				break
 			}
 		}
-		if !goterror {
-			t.Fatal("expeceted an error we didn't get")
+		ver, _ := ParseVersionString(runtime.Version())
+
+		if !(ver.Major < 0 || ver.AfterOrEqual(GoVersion{1, 8, -1, 0, 0})) {
```
derekparker: Would be nice to have consts that represent the different Go versions for simplicity.

dwarf/line/line_parser_test.go
```diff
@@ -67,11 +67,15 @@ func TestDebugLinePrologueParser(t *testing.T) {
 		t.Fatal("Initial value of 'is_stmt' not parsed correctly", prologue.InitialIsStmt)
 	}
 
-	if prologue.LineBase != int8(-1) {
+	if prologue.LineBase != int8(-1) && prologue.LineBase != int8(-4) {
```
derekparker: nit: would be nice to use consts for these values `int8(-1)` and `int8(-4)`.

dwarf/line/line_parser_test.go
```diff
@@ -67,11 +67,15 @@ func TestDebugLinePrologueParser(t *testing.T) {
...
+		// go < 1.8 uses -1
+		// go >= 1.8 uses -4
 		t.Fatal("Line base not parsed correctly", prologue.LineBase)
 	}
 
-	if prologue.LineRange != uint8(4) {
+	if prologue.LineRange != uint8(4) && prologue.LineRange != uint8(10) {
```
derekparker: nit: same as above.

dwarf/line/state_machine.go
```diff
@@ -27,6 +27,11 @@ type StateMachine struct {
 	endSeq          bool
 	lastWasStandard bool
 	lastDelta       int
+	// if valid is true the current value of the state machine is the address of
```
derekparker: nit: s/if valid is true/valid is true if/

service/test/integration2_test.go
```diff
@@ -1048,8 +1048,16 @@ func TestSkipPrologue2(t *testing.T) {
 
 		callme3 := findLocationHelper(t, c, "main.callme3", false, 1, 0)[0]
 		callme3Z := findLocationHelper(t, c, "main.callme3:0", false, 1, 0)[0]
-		// callme3 does not have local variables therefore the first line of the function is immediately after the prologue
-		findLocationHelper(t, c, "callme.go:18", false, 1, callme3Z)
+		ver, _ := proc.ParseVersionString(runtime.Version())
```
derekparker: Fail test on error here.

service/test/integration2_test.go
```diff
@@ -1048,8 +1048,16 @@ func TestSkipPrologue2(t *testing.T) {
...
+		if ver.Major < 0 || ver.AfterOrEqual(proc.GoVersion{1, 8, -1, 0, 0}) {
```
derekparker: nit: Use const to represent this version (`proc.GoVersion{1, 8, -1, 0, 0}`).

proc/proc.go
```diff
@@ -359,10 +359,18 @@ func (dbp *Process) Continue() error {
 		case dbp.CurrentThread.CurrentBreakpoint == nil:
 			// runtime.Breakpoint or manual stop
 			if dbp.CurrentThread.onRuntimeBreakpoint() {
-				for i := 0; i < 2; i++ {
+				// Single-step current thread until we exit runtime.breakpoint and
+				// runtime.Breakpoint.
+				// On go < 1.8 it was sufficient to single-step twice on go1.8 a change
+				// to the compiler requires 4 steps
```
derekparker: nit: Add punctuation at end of sentence (sorry).

proc/proc.go
```diff
@@ -359,10 +359,18 @@ func (dbp *Process) Continue() error {
...
+				for {
 					if err = dbp.CurrentThread.StepInstruction(); err != nil {
 						return err
 					}
+					loc, err := dbp.CurrentThread.Location()
+					if err != nil || loc.Fn == nil || (loc.Fn.Name != "runtime.breakpoint" && loc.Fn.Name != "runtime.Breakpoint") {
```
derekparker: nit: (don't feel strongly, can ignore) might be nice to wrap this logic in a function like `isInRuntimeBreakpoint`.

proc/proc_test.go
```diff
@@ -2374,8 +2373,25 @@ func TestIssue683(t *testing.T) {
...
+			// in go1.8 the debug_line entry is correct for this piece of code
+			if !goterror {
```
aarzilli: I think it's better if I remove the error check completely, whether or not we get the error is irrelevant and depends on what the compiler does, what's important is that the test doesn't panic.

proc/proc_test.go
aarzilli: Will do in a separate PR since it doesn't have anything to do with 1.8 compatibility

service/test/integration2_test.go
aarzilli: Second value is not an error, it's a bool indicating whether the version string is parsable. In particular we can't parse the version string of development versions of go (because they contain the sha-1 abbrev of HEAD), this condition is covered by the `ver.Major < 0` check.

proc/proc.go
aarzilli: it's short and only used here, imho it's better inline.

aarzilli:
Done.

ignatov:
@derekparker Do you have any estimates for that fix?

service/test/integration2_test.go
derekparker: SGTM

PR #672, opened 2016-11-18
tests: fix TestBuild on macOS, windows

aarzilli:


derekparker:
@aarzilli this is failing on OSX in CI.

aarzilli:
fixed.

cmd/dlv/dlv_test.go
```diff
@@ -22,13 +22,22 @@ func assertNoError(err error, t testing.TB, s string) {
 
 func TestBuild(t *testing.T) {
 	const listenAddr = "localhost:40573"
-	cmd := exec.Command("go", "build", "github.com/derekparker/delve/cmd/dlv")
+	args := []string{"build"}
```
derekparker: Thinking about this, all of our CI platforms have `make`, we should just invoke `make build` here instead of trying to duplicate platform specific building rules here.

aarzilli:
How about this?

derekparker:
@aarzilli seems AppVeyor isn't happy, I really thought we had make available in our Window CI environment.

PR #682, opened 2016-12-13
Small clarification on how to use dlv debug

VictorDegliame:
Took me some time to figure this out. Can we add this in the right file?


derekparker:
I think this might be better served as a startup message, e.g. "use `break` to set a breakpoint and `continue` to continue execution of your program, for other available commands use the `help` command" or something along those lines.

PR #690, opened 2016-12-23
proc: Improved prologue detection for big stackframes

aarzilli:
Improved prologue detection for big stackframes and added references to
the function of the compiler that inserts the prologue.


derekparker:
Been on vacation, will take a look at this today or tomorrow.

PR #694, opened 2017-01-04
terminal: fixed typo.

aarzilli:
one `Println` was supposed to be `Printf`, thanks @zchee.


aarzilli:
Edit: I've added a commit to fix a more serious bug with the same patch, by using the old process instead of the new one to call FindFileLocation the prologue wouldn't be skipped.

PR #695, opened 2017-01-04
Trace optimizations

aarzilli:
Following up from a mailing list discussion I took some measurements on where the time is spent on a simple Continue()+FunctionArguments() loop and made some optimizations taking the time spent on a single loop iteration from ~40ms to under 1ms.

The bulk of the improvement comes from skipping a time.Sleep in the linux implementation of singleStep, this sleep is needed to work around a quirk of the linux kernel but that particular quirk can not happen when we are stepping over a breakpoint, so it's unnecessary in that case.

The rest of the optimizations have a much smaller impact but cumulate to ~50% of the remaining time.

```
* proc: Added trace benchmark

Results:

BenchmarkTrace-4   	    5000	  36195899 ns/op

* proc/linux: faster single step implementation.

BenchmarkTrace-4   	    5000	   2093271 ns/op

* proc: Cache function debug_info entries to speed up variable lookup.

BenchmarkTrace-4   	    5000	   1864846 ns/op

* proc/variables: Optimize FunctionArguments by prefetching frame

BenchmarkTrace-4   	    5000	   1815795 ns/op

* proc/variables: optimized parseG

BenchmarkTrace-4   	   10000	    712767 ns/op
```

### DIFF
--- a/proc/types.go
+++ b/proc/types.go
@@ -8,6 +8,7 @@ import (
 	"go/constant"
 	"go/token"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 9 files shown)


proc/types.go
```diff
@@ -118,22 +119,61 @@ func (dbp *Process) loadPackageMap() error {
...
-		if _, exists := dbp.types[name]; !exists {
-			dbp.types[name] = entry.Offset
+	}
+	sort.Sort(sortFunctionsDebugInfoByLowpc(dbp.functions))
+}
+
+func (dbp *Process) findFunctionDebugInfo(pc uint64) (dwarf.Offset, error) {
+	i := sort.Search(len(dbp.functions), func(i int) bool {
+		fn := dbp.functions[i]
+		return pc <= fn.lowpc || (fn.lowpc <= pc && pc < fn.highpc)
```
derekparker: The logic here seems wrong, wouldn't we want `pc >= fn.lowpc`? E.g. we want to return `true` if the `pc` we're looking up is within the bounds of `(lowpc...highpc)`, and this initial branch (before the OR) would return true if the argument `pc` is less than `fn.lowpc` -- wouldn't that cause a shortcircuit and give us back the wrong function potentially? Why not remove that first condition and use the range check?

proc/variables.go
```diff
@@ -1567,6 +1606,34 @@ func (scope *EvalScope) variablesByTag(tag dwarf.Tag, cfg LoadConfig) ([]*Variab
 			vars = append(vars, val)
 		}
 	}
+	if len(vars) <= 0 {
+		return vars, nil
+	}
+
+	// prefetch the whole chunk of memory relative to this variables
```
derekparker: s/this/these/

derekparker: This unconditional caching of all this memory makes me a bit nervous. I understand for function args, as they should be passed in contiguously, but if we're dealing with local variables we could have addresses on the stack and addresses on the heap. We should check and only cache contiguous memory, or modify how caching works internally to only read when necessary (lazy memory evaluation).

proc/types.go
aarzilli: sort.Search requires its predicate f to respect the rule that if f(i) is true then f(i+1) is also true. This means all functions located after the pc we are searching need to evaluate to true (`pc <= fn.lowpc`) as well as the one function we want (`fn.lowpc <= pc && pc < fn.highpc`).

But the `||` here is unnecessary.

proc/variables.go
aarzilli: Done.

aarzilli: AFAIK this is impossible. Escaping locals are replaced by `&v` entries that are stored on the stack, the compiler does not produce DW_OP_deref instructions (and we don't support them). But I'll add a check anyway, as all this could change.

PR #724, opened 2017-02-08
Target interface refactor and directory layout cleanup

derekparker:
This patch does two things, so could be broken up, but since it's disruptive (file moves) I wanted to get it in quickly to prevent blocking other work.

Essentially this patch introduces a new interface `target.Interface` which represents the debuggers target, which currently is a system process but could eventually be a core file, etc. The main goal is to continue to abstract, isolate, and cleanup the `proc` package, whose usage going forward is strictly to a be an extremely low-level package for dealing with and manipulation of system processes.

Additionally, I've moved all scattered top level packages into `pkg/` for consistency.


pkg/target/target.go
```diff
@@ -0,0 +1,84 @@
...
+	PCToLine(uint64) (string, int, *gosym.Func)
+	FirstPCAfterPrologue(fn *gosym.Func, sameline bool) (uint64, error)
+}
+
+// ThreadInfo is an interface for getting information on active threads
+// in the process.
+type ThreadInfo interface {
+	Threads() map[int]*proc.Thread
+	CurrentThread() *proc.Thread
+	SwitchThread(int) error
```
aarzilli: SwitchThread seems more at home in ProcessManipulation

pkg/target/target.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// GoroutineInfo is an interface for getting information on running goroutines.
+type GoroutineInfo interface {
+	GoroutinesInfo() ([]*proc.G, error)
+	SelectedGoroutine() *proc.G
+	SwitchGoroutine(int) error
```
aarzilli: SwitchGoroutine seems more at home in ProcessManipulation, like SwitchThread.

pkg/target/target.go
derekparker: Agreed.

pkg/target/target.go
derekparker: Agreed.

derekparker:
@aarzilli updated.

aarzilli:
LGTM

PR #725, opened 2017-02-08
proc/variables: fix infinite recursion with pointer loop

aarzilli:
```
proc/variables: fix infinite recursion with pointer loop

loadValue didn't react correctly to pointer loops going through
slice -> interface{} -> slice or pointer -> interface{} -> pointer.

```


derekparker:
Looks fine, will wait for at least Linux build to finish before merge.

aarzilli:
Rebased.

PR #726, opened 2017-02-08
dlv: add core file support for linux_amd64

heschik:
Core files contain a variety of memory mappings either to files or
anonymous regions stored in the core file. These regions can overlap, so
figuring out what exactly to read can be tricky. This commit contains
a data structure, SplicedMemory, which accumulates mappings and reads
from the correct sources.


derekparker:
Thanks for this patch!

We just merged a refactor PR this should go on top of (https://github.com/derekparker/delve/pull/724) -- but the refactor helps enable this support. Essentially, `proc -> pkg/proc` and added `target.Interface`.

heschik:
Oh, awesome, that should make things a lot less ugly in the next patch. Rebased.

heschik:
I had the next commit ready, so you can certainly take a look at that one.

A working implementation might be a little tricky in the current state of the repo. The target interfaces are great, but to actually use the parsed core file you need a bunch of the code that currently lives in Process, basically everything in initializeDebugProcess and down, plus most of moduledata.go, stack.go, and types.go. In my prototype, I had just hacked the core support into Process which was a little ugly.

I don't know if I'll have the time to do the refactoring necessary to do this cleanly -- I want to start looking at some compiler stuff. I can clean up the Process hack a little bit and send it (it's not *that* bad, a dozen or so if core {return error}), or we could merge this even though it isn't hooked up, or we could put this on hold and see if the refactoring happens?

derekparker:
@heschik makes sense, I will likely merge this in today then and create an issue or something to track remaining work that needs to be done.

heschik:
Cool! Since I already have it to some extent, I'll clean up the commit that adds it to Process and send it as a separate PR, and we can see how you feel about it.

heschik:
Uh, actually, hang on. I found a typo in the note parsing which the tests didn't catch, so I'm going to have to improve them and fix up some error handling while I'm at it.

heschik:
OK, fixed the dumb bugs. Since I would be a little suspicious in your shoes, the 4th commit is the Process and command implementation so you can try it out for yourself. No tests for it yet though. Let me know how you feel and I can add tests or revert it.

derekparker:
@heschik let's add tests for it and work to get this PR merged, following that I can tackle the refactor to then separate this out into its own target and clean up the `Target` abstraction.

heschik:
Added a test, PTAL.

aarzilli:
Two problems:
* TestCore [seems to fail on travis](https://travis-ci.org/derekparker/delve/jobs/202451685)
* io.SeekCurrent is not in go1.6, use `os.SEEK_CUR` instead (at this point we could drop support for go1.6 but it's such a small thing...)

heschik:
Changed to SEEK_CUR.

For the travis tests: it looks like travis needs some tweaks to generate core files because /proc/sys/kernel/core_pattern wants to run apport, but apport isn't installed in the default travis image, and judging by the latest comments on https://github.com/travis-ci/travis-ci/issues/3754, the version you can install is now too old to work. So I changed to sudo: required and modified the core pattern in the install section. Sudo support does cost something like 30 seconds, though, because it requires a VM rather than containers.

(As I write this Travis is not scheduling the most recent run, I think I might have DOSed it a little, but I'm pretty sure I finally got it right.)

aarzilli:
You got it right.

heschik:
@derekparker : anything else to do before this is merged? I see that the core test is failing on tip. Given that other things are still failing I suspect it's not a problem with core files specifically, but I can look into it more if you'd like.

pkg/proc/core_linux_amd64.go
```diff
@@ -0,0 +1,248 @@
...
+	Pid     int
+}
+
+// Note is a note from the PT_NOTE prog.
+// Relevant types:
+// - NT_FILE: File mapping information, e.g. program text mappings. Desc is a LinuxNTFile.
+// - NT_PRPSINFO: Information about a process, including PID and signal. Desc is a LinuxPrPsInfo.
+// - NT_PRSTATUS: Information about a thread, including base registers, state, etc. Desc is a LinuxPrStatus.
+// - NT_FPREGSET (Not implemented): x87 floating point registers.
+// - NT_X86_XSTATE (Not implemented): Other registers, including AVX and such.
```
hyangah: Make it godoc-friendly

pkg/proc/core_linux_amd64.go
```diff
@@ -0,0 +1,248 @@
...
+			break
+		}
+	}
+
+	r := notesProg.Open()
+	notes := []*Note{}
+	for {
+		note, err := readNote(r)
+		if err != nil {
+			// Assume that any error is because we ran out of notes.
```
hyangah: I am confused - is it true that readNote returns an error only when we run out of notes? EOF? No possibility of corruption or bug in the api?

pkg/proc/core_linux_amd64.go
```diff
@@ -0,0 +1,248 @@
...
+		note.Desc = data
+	}
+
+	return note, nil
+}
+
+func buildMemory(core *elf.File, exe io.ReaderAt, notes []*Note) MemoryReader {
+	memory := &SplicedMemory{}
+
+	// For now, assume all file mappings are to the exe.
```
hyangah: Does delve code base use "TODO" convention? If so, TODO will be better for search.

pkg/proc/core_linux_amd64.go
```diff
@@ -0,0 +1,248 @@
...
+				reader: prog.ReaderAt,
+				offset: uintptr(prog.Vaddr),
+			}
+			memory.Add(r, uintptr(prog.Vaddr), uintptr(prog.Filesz))
+		}
+	}
+	return memory
+}
+
+// Various structures from the ELF spec and the Linux kernel.
```
hyangah: FYI, if grouping was used, the comment will be used to all grouped types. 

type (
   LinuxPrPsInfo struct {
     ...
   }
   LinuxPrStatus struct {
     ...
   }
)

hyangah: But you have doc for ELFNotesHdr , so maybe doesn't matter.

pkg/proc/core_linux_amd64_test.go
```diff
@@ -112,5 +120,26 @@ func TestSplicedReader(t *testing.T) {
...
+	tempDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fix := test.BuildFixture("panic")
+	bashCmd := fmt.Sprintf("cd %v && ulimit -c unlimited && GOTRACEBACK=crash %v", tempDir, fix.Path)
+	exec.Command("bash", "-c", bashCmd).Run()
+	corePath := path.Join(tempDir, "core")
 
+	core, err := readCore(corePath, fix.Path)
```
hyangah: How about checking if the core was produced first, and skipping the test if the core is not produced?
This test should test whether the core can be read, not whether the core is produced.

pkg/proc/core.go
```diff
@@ -0,0 +1,146 @@
...
+// can address all of 64-bit memory.
+// Redundant with memoryReadWriter but more easily suited to working with
+// the standard io package.
+type MemoryReader interface {
+	// ReadMemory is just like io.ReaderAt.ReadAt.
+	ReadMemory(buf []byte, addr uintptr) (n int, err error)
+}
+
+// A SplicedMemory represents a memory space formed from multiple regions,
+// each of which may override previously regions. For example, in the following
```
heschik: Done.

pkg/proc/core.go
```diff
@@ -0,0 +1,146 @@
...
+// core, the program text was loaded at 0x400000:
+// Start               End                 Page Offset
+// 0x0000000000400000  0x000000000044f000  0x0000000000000000
+// but then it's partially overwritten with an RW mapping whose data is stored
+// in the core file:
+// Type           Offset             VirtAddr           PhysAddr
+//                FileSiz            MemSiz              Flags  Align
+// LOAD           0x0000000000004000 0x000000000049a000 0x0000000000000000
+//                0x0000000000002000 0x0000000000002000  RW     1000
```
heschik: Done.

pkg/proc/core.go
```diff
@@ -0,0 +1,146 @@
...
+
+// ReadMemory implements MemoryReader.ReadMemory.
+func (r *SplicedMemory) ReadMemory(buf []byte, addr uintptr) (n int, err error) {
+	started := false
+	for _, entry := range r.readers {
+		if entry.offset+entry.length < addr {
+			if !started {
+				continue
+			}
+			return n, fmt.Errorf("hit unmapped area at %v after %v bytes", addr, n)
```
heschik: Done.

pkg/proc/core_linux_amd64.go
heschik: Done.

pkg/proc/core_linux_amd64.go
heschik: Can't hurt, added a TODO.

pkg/proc/core_linux_amd64.go
heschik: I've never seen that syntax used for types before so I think I'll pass.

pkg/proc/core_linux_amd64.go
heschik: I fixed this in a later commit.

derekparker:
@heschik sorry for the delay on this, have been busy unfortunately. I will take another look at this, and likely merge tomorrow.

derekparker:
@heschik my apologies again for the delay. Will have certainly have to take take a final look at this today, want to also consider whether this might sit better on some refactoring @aarzilli is doing at the moment as well.

heschik:
Ping; I'm not in any particular rush here but it has been quite a while.

derekparker:
Your work will likely be merged via: https://github.com/derekparker/delve/pull/782, which carries your commit.

derekparker:
Done via https://github.com/derekparker/delve/pull/782 -- thanks!

PR #728, opened 2017-02-08
pkg/proc: fix struct methods to unexported name & format by goimports

zchee:
Currently, can't compile on darwin.
Fixed some variable to unexported name.


zchee:
@aarzilli Ah, sorry. I got it. I'll take care.

zchee:
@aarzilli Thanks!

PR #731, opened 2017-02-11
Fix variable evaluation in multiple packages and display of strings on Windows

dlsniper:
Hi,

I've picked up this issue from a user of Gogland and reproduced locally, see: https://youtrack.jetbrains.com/issue/GO-3260
Fairly sure it needs some improvements as it's a major hack but I would need some guidance on the preferred way to continue with this.
Also, after fixing the issue in `variables.go` I've realized that the return value display was more than it needed and thus showing some empty memory. As such, I've added a change to `eval.go` but not sure about it either.

I'll try to improve my contributions in the future, still learning :)
Please let me know how to improve / get this merged.

Thank you.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -847,6 +847,9 @@ func (scope *EvalScope) evalBinary(node *ast.BinaryExpr) (*Variable, error) {
 
 		r := xv.newVariable("", 0, typ)
 		r.Value = rc
+		if r.Kind == reflect.String {
+			r.Len = xv.Len+yv.Len
+		}
 		return r, nil
 	}
 }
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -680,7 +680,7 @@ func (scope *EvalScope) packageVarAddr(name string) (*Variable, error) {
 			continue
 		}
 
-		if n == name {
+		if n == name || strings.HasSuffix(n, "/"+name) {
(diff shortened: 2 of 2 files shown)


dlsniper:
Also, I'm still setting up my dev environment so I apologize for the lack of testing, I still can't figure out how to run them for now (I'm on Windows)

aarzilli:
This should really be two different commits and at least the one about string lengths should have a test (for the other one writing a test is going to be hard).

There are already multiple tests for expression evaluation, just look in proc_test.go, anything mentioning testvariables2 (which you should also use).

pkg/proc/eval.go
```diff
@@ -847,6 +847,9 @@ func (scope *EvalScope) evalBinary(node *ast.BinaryExpr) (*Variable, error) {
 
 		r := xv.newVariable("", 0, typ)
 		r.Value = rc
+		if _, ok := r.DwarfType.(*dwarf.StringType); ok {
```
aarzilli: I think this looks better as `r.Kind == reflect.String`.

derekparker:
Will TAL at this today, sorry for the delay.

dlsniper:
Sorry for delay as well, I've refactored this and I was about to add a test to it (still not sure how). I'll push it in a few hours)

derekparker:
Closed via https://github.com/derekparker/delve/pull/802

PR #735, opened 2017-02-13
proc_test: TestStacktraceWithBarriers failing on go1.8

aarzilli:
```
proc_test: TestStacktraceWithBarriers failing on go1.8

Changes to the GC in 1.8 make this test fail because the GC isn't
inserting stack barriers into our test program anymore. This isn't the
same thing as being unable to print stacktraces in presence of stack
barriers so the test shouldn't fail.

```

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -2436,13 +2436,27 @@ func TestStacktraceWithBarriers(t *testing.T) {
 	// stack frame with the address of runtime.stackBarrier.
 	// The original return address is saved into the stkbar slice inside the G
 	// struct.
+
+	// In Go 1.8 stack barriers are not inserted by default, this enables them.
+	godebugOld := os.Getenv("GODEBUG")
+	defer os.Setenv("GODEBUG", godebugOld)
+	os.Setenv("GODEBUG", "gcrescanstacks=1")
+
+	// TODO(aarzilli): in Go 1.9 stack barriers will be removed completely, therefore
+	// this test will have to be disabled
+
 	withTestProcess("binarytrees", t, func(p *Process, fixture protest.Fixture) {
 		// We want to get a user goroutine with a stack barrier, to get that we execute the program until runtime.gcInstallStackBarrier is executed AND the goroutine it was executed onto contains a call to main.bottomUpTree
 		_, err := setFunctionBreakpoint(p, "runtime.gcInstallStackBarrier")
 		assertNoError(err, t, "setFunctionBreakpoint()")
 		stackBarrierGoids := []int{}
 		for len(stackBarrierGoids) == 0 {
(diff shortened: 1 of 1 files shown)


aarzilli:
I'm going to try to summon @aclements here, see if he knows a reliable way for a debugger to catch the GC putting up stack barriers.

aclements:
I'm not sure what you want to catch, exactly. In Go 1.8, as you said, stack barriers aren't used by default, but all of the support structures are still there, so I don't see why a debugger would get confused. It just shouldn't see any stack barriers. Can you elaborate on what the test is doing and why it's failing?

In terms of detecting stack barrier pre-Go 1.8, I would first look at the `stackLock` field in `runtime.g`. If that's non-zero, there's probably just nothing you can do since GC is in the middle of installing stack barriers and a lot of the state is hiding in local variables in the GC. If that's zero, you should be able to read `runtime.g.stkbar` and unwind through stack barriers the same way that `runtime.gentraceback` does. In Go 1.8, `runtime.g.stkbar` will simply always be an empty slice. (And I'm about to push a commit to master that rips out the stack barrier code entirely.)

aarzilli:
@aclements All I needed to know was:

> In Go 1.8, as you said, stack barriers aren't used by default, but all of the support structures are still there

I enabled `gcrescanstacks` for this test and now it's passing again. It was failing because it couldn't find a stack with a stack barrier (the test is for testing that we scan a stack with stack barriers correctly).

aclements:
> I enabled gcrescanstacks for this test and now it's passing again. It was failing because it couldn't find a stack with a stack barrier (the test is for testing that we scan a stack with stack barriers correctly).

Ah, if you're trying to make sure you see a stack barrier, then it makes sense you wouldn't see them if they are never there. :) Turning on gcrescanstacks seems reasonable for now, though I just pushed the change to master that rips out stack barriers entirely. I'm not sure what the tests are capable of, but perhaps it could check if the runtime support structures are there in the debug info (e.g., `runtime.gcInstallStackBarrier`) and just skip the test if they aren't.

derekparker:
@aarzilli @aclements I think that's the way to go, we should skip this test when not applicable.

aarzilli:
@derekparker we could wait until we do 1.9 support for that.

pkg/proc/proc_test.go
```diff
@@ -2436,13 +2436,24 @@ func TestStacktraceWithBarriers(t *testing.T) {
 	// stack frame with the address of runtime.stackBarrier.
 	// The original return address is saved into the stkbar slice inside the G
 	// struct.
+
+	// In Go 1.8 stack barriers are not inserted by default, this enables them.
```
derekparker: Can you add a `TODO` for ignoring this test on 1.9 when it is no longer applicable.

aarzilli:
Done.

PR #739, opened 2017-02-14
Miscellaneous bug fixes on windows

aarzilli:
## proc/windows: handle delayed events
Sometimes windows will send us events about breakpoints we have
already removed from the code despite the fact that we go to great
lengths to avoid this already.

Change waitForDebugEvent to check that when we receive a breakpoint
event the corresponding memory actually contains an INT 3
instruction, if it doesn't ignore the event and restart the thread.

## proc/stack: use BP when FDE is not available
On Windows we can sometimes encounter threads stopped in locations for
which we do not have entries in debug_frame.

These cases seem to be due to calls to Windows API in the go runtime,
we can still produce a (partial) stack trace in this circumstance by
following frame pointers (starting with BP).

We still prefer debug_frame entries when available since go functions
do not have frame pointers before go1.8.

## proc/test: improve TestStepParked
It was possible for TestStepParked to pick a runtime goroutine and
attempt to step it. Nothing guarantees that a goroutine other than the
ones we are using to run the code would actually ever resume before the
end of the program.

This makes the test more discerning in its choice of goroutines.

## terminal/command_test: improved TestIssue387
The test in question tries to 'next' over a call to wg.Done, this is
not guaranteed to succeed, if the goroutine gets suspended after
wg.Done has notified the waiting group but before returning to
main.dostuff the program could quit before the goroutine is resumed.

*NOTE* this was developed on top of PR #735.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -46,9 +46,12 @@ func TestBuild(t *testing.T) {
 	assertNoError(err, t, "stdout pipe")
 	cmd.Start()
 	defer func() {
-		cmd.Process.Signal(os.Interrupt)
 		if runtime.GOOS != "windows" {
+			cmd.Process.Signal(os.Interrupt)
 			cmd.Wait()
+		} else {
+			// sending os.Interrupt on windows is not supported
+			cmd.Process.Kill()
 		}
 	}()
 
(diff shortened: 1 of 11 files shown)


derekparker:
Needs rebase.

PR #745, opened 2017-02-20
proc refactor: split out BinaryInfo implementation

aarzilli:
```
proc: compile support for all executable formats unconditionally

So far we only compiled in support for loading the executable format
supported by the host operating system.
Once support for core files is introduced it is however useful to
support loading in all executable formats, there is no reason why it
shouldn't be possible to examine a linux coredump on windows, or
viceversa.

proc: refactor BinaryInfo part of proc.Process to own type

The data structures and associated code used by proc.Process
to implement target.BinaryInfo will also be useful to support a
backend for examining core dumps, split this part of proc.Process
to a different type.

```

### DIFF
--- a/pkg/proc/moduledata.go
+++ b/pkg/proc/moduledata.go
@@ -11,9 +11,9 @@ type moduleData struct {
 	typemapVar    *Variable
 }
 
-func (dbp *Process) loadModuleData() (err error) {
+func (dbp *BinaryInfo) loadModuleData(thread *Thread) (err error) {
 	dbp.loadModuleDataOnce.Do(func() {
-		scope := &EvalScope{Thread: dbp.currentThread, PC: 0, CFA: 0}
+		scope, _ := thread.Scope()
 		var md *Variable
 		md, err = scope.packageVarAddr("runtime.firstmoduledata")
 		if err != nil {
@@ -56,9 +56,10 @@ func (dbp *Process) loadModuleData() (err error) {
 	return
 }
 
-func (dbp *Process) resolveTypeOff(typeAddr uintptr, off uintptr) (*Variable, error) {
+func (dbp *BinaryInfo) resolveTypeOff(typeAddr uintptr, off uintptr, thread *Thread) (*Variable, error) {
+	var mem memoryReadWriter = thread
 	// See runtime.(*_type).typeOff in $GOROOT/src/runtime/type.go
-	if err := dbp.loadModuleData(); err != nil {
+	if err := dbp.loadModuleData(thread); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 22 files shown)


derekparker:
@aarzilli is this off of master, including the Windows test fixes? If not can you rebase so we can at least get a clean run in AppVeyor.

aarzilli:
done.

derekparker:
Will TAL today.

pkg/proc/proc.go
```diff
@@ -192,21 +173,11 @@ func (dbp *Process) Breakpoints() map[uint64]*Breakpoint {
...
-		return err
-	}
-	fi, err := os.Stat(path)
-	if err == nil {
-		dbp.lastModified = fi.ModTime()
-	}
+	path = findExecutable(path, dbp.pid)
 
-	wg.Add(5)
+	wg.Add(1)
```
derekparker: We can remove the WaitGroup here since there's only a single operation, doesn't need to be done concurrently.

pkg/proc/moduledata.go
```diff
@@ -11,9 +11,9 @@ type moduleData struct {
 	typemapVar    *Variable
 }
 
-func (dbp *Process) loadModuleData() (err error) {
+func (dbp *BinaryInfo) loadModuleData(thread *Thread) (err error) {
```
derekparker: For all of these: change the receiver name to `bin` or `bi` or something.

pkg/proc/proc.go
```diff
@@ -160,7 +141,7 @@ func (dbp *Process) Running() bool {
 	return false
 }
 
-func (dbp *Process) LastModified() time.Time {
+func (dbp *BinaryInfo) LastModified() time.Time {
```
derekparker: I think it would be clearer to move these to another file.

pkg/proc/proc.go
```diff
@@ -140,6 +120,7 @@ func (dbp *Process) Detach(kill bool) (err error) {
 			err = killProcess(dbp.pid)
 		}
 	})
+	dbp.Close()
```
derekparker: Seems odd to call close on `Process` -- makes me think of not embedding `BinaryInfo`, but could be convinced otherwise.

aarzilli: We could change target.Interface to have a method `BinaryInfo() *proc.BinaryInfo` instead, since this describes the executable it should be the same for all implementation and could be represented by a concrete type.

pkg/proc/proc.go
aarzilli: moved to bininfo.go

pkg/proc/proc.go
aarzilli: `LoadBinaryInfo` actually spawns further goroutines and adds them to the WaitGroup.

aarzilli:
Done.

aarzilli:
Appveyor was failing intermittently, I think it was caused by the detach fix altough I can't reproduce it locally, I've added a patch for that as a separate commit.

pkg/proc/proc.go
derekparker: SGTM

pkg/proc/proc.go
derekparker: Got it.

aarzilli:
Done.

Issue #758, opened 2017-03-07
Add a bash script to create a certificate and sign delve's executable on macOS


aarzilli:
Some people do not want to use homebrew to install go (#753 and https://github.com/go-delve/homebrew-delve/issues/9). We should make a bash script that can be run with sudo after `go get` that:

1. creates a new certificate
2. installs it
3. signs delve's executable.

justinclift:
As a curiosity thought, which kind of certificate are you meaning?

Are you meaning an OSX developer cert, used for signing releases so [Gatekeeper](https://en.wikipedia.org/wiki/Gatekeeper_(macOS)) doesn't complain?

aarzilli:
Yes. Our homebrew formula creates a self signed certificate using openssl, adds it to the keychain then uses it to sign dlv with codesign. The last bit is easy but creating the certificate isn't. There should be a script do to that without having to use homebrew.

See: https://github.com/go-delve/homebrew-delve/blob/master/Formula/delve.rb

justinclift:
k, so converting the commands in the Homebrew formula gives this:

```
#!/bin/bash

CERT="dlv-cert"
TEMPLATE="$CERT.tmpl"

echo '[ req ]' > $TEMPLATE
echo 'default_bits            = 2048                  # RSA key size' >> $TEMPLATE
echo 'encrypt_key             = no                    # Protect private key' >> $TEMPLATE
echo 'default_md              = sha512                # MD to use' >> $TEMPLATE
echo 'prompt                  = no                    # Prompt for DN' >> $TEMPLATE
echo 'distinguished_name      = codesign_dn           # DN template' >> $TEMPLATE
echo '' >> $TEMPLATE
echo '[ codesign_dn ]' >> $TEMPLATE
echo 'commonName              = "dlv-cert"' >> $TEMPLATE
echo '' >> $TEMPLATE
echo '[ codesign_reqext ]' >> $TEMPLATE
echo 'keyUsage                = critical,digitalSignature' >> $TEMPLATE
echo 'extendedKeyUsage        = critical,codeSigning' >> $TEMPLATE

echo "Generating ${CERT}.crt.pem"
openssl req -new -newkey rsa:2048 -x509 -days 3650 -nodes -config ${TEMPLATE} -extensions codesign_reqext -batch -out ${CERT}.crt.pem -keyout $CERT.key

echo "Installing ${CERT}.crt.pem as root"
[...]

PR #760, opened 2017-03-07
Bash script to generate & install signing certificate for delve

justinclift:
For #758


justinclift:
Looking at the Travis build failures, it seems to be something socket related for delve itself.

Reasonably sure it's nothing to do with this PR. :wink:

derekparker:
Thanks for this patch, and you're correct, the Travis failures are unrelated, and something we're actively working on improving. I will take a look at this and test on an OSX system shortly.

Issue #754, opened 2017-03-03
Command failed: Internal debugger error: runtime error: invalid memory address or nil pointer dereference


alexbrainman:
1. What version of Delve are you using (`dlv version`)?

I am using current tip bb07fc36

2. What version of Go are you using? (`go version`)?

go version devel +4b22485 Sun Feb 26 23:36:09 2017 +0000 windows/amd64

3. What operating system and processor architecture are you using?

Windows 7
amd64

4. What did you do?

I modified delve:
```
diff --git a/pkg/proc/stack.go b/pkg/proc/stack.go
index 096cf7a..7c01969 100644
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -130,6 +130,7 @@ type savedLR struct {
 }
 
 func newStackIterator(dbp *Process, pc, sp, bp uint64, stkbar []savedLR, stkbarPos int) *stackIterator {
+	fmt.Printf("%v %v\n", runtimeStackBarrier, dbp.goSymTable.LookupFunc(runtimeStackBarrier))
 	stackBarrierPC := dbp.goSymTable.LookupFunc(runtimeStackBarrier).Entry
 	if stkbar != nil {
 		fn := dbp.goSymTable.PCToFunc(pc)
```
to help debugging, then built delve and did this
```
c:\dev\src\t>type main.go
package main

import (
        "fmt"
)

func main() {
        fmt.Printf("hello\n")
}

c:\dev\src\t>c:\dlv.exe debug main.go
Type 'help' for list of commands.
(dlv) b main.go
runtime.stackBarrier <nil>
Command failed: Internal debugger error: runtime error: invalid memory address or nil pointer dereference
runtime.call32 (0x450a42)
        /root/go/src/runtime/asm_amd64.s:492
runtime.gopanic (0x428a18)
        /root/go/src/runtime/panic.go:489
runtime.panicmem (0x4278d5)
        /root/go/src/runtime/panic.go:63
runtime.sigpanic (0x43cbcb)
        /root/go/src/runtime/signal_windows.go:161
github.com/derekparker/delve/pkg/proc.newStackIterator (0x666a18)
        /root/src/github.com/derekparker/delve/pkg/proc/stack.go:134
github.com/derekparker/delve/pkg/proc.(*Thread).stackIterator (0x66644e)
        /root/src/github.com/derekparker/delve/pkg/proc/stack.go:64
[...]

PR #762, opened 2017-03-10
pkg/proc: tolerate absence of stack barriers in Go 1.9

dr2chase:
Stack barriers were removed in Go 1.9, and thus code that
expected various stack-barrier-related symbols to exist
does not find them.  Check for their absence and do not
crash when they are missing.


alexbrainman:
Fixes #754 ?

Alex

dr2chase:
Yes, fixes #754.

pkg/proc/variables.go
```diff
@@ -498,6 +504,9 @@ func (g *G) Go() Location {
 
 // Returns the list of saved return addresses used by stack barriers
 func (g *G) stkbar() ([]savedLR, error) {
+	if g.stkbarVar == nil { // stack barriers were removed in Go 1.9
+		return make([]savedLR, 0), nil
```
aarzilli: isn't returning nil ok?

PR #763, opened 2017-03-13
bugfix: wrong register used for CH

aarzilli:
```
bugfix: wrong register used for CH

We only use Registers.Get besides for evaluating the argument of a CALL
instruction so this doesn't matter in practice, but it's still wrong.

```


aarzilli:
The error on appveyor seems to be a problem with appveyor, there's no way that file doesn't compile on windows and this PR didn't touch it at all.

derekparker:
@aarzilli kicked over AppVeyor tests, we'll see how the next run goes.

PR #771, opened 2017-03-18
proc/tests: make TestIssue419 more conscientious

aarzilli:
```
proc/tests: make TestIssue419 more conscientious

Do not send SIGINT if the value returned by (*Process).Pid doesn't look
right
.

```

### DIFF
--- a/pkg/proc/proc_unix_test.go
+++ b/pkg/proc/proc_unix_test.go
@@ -3,6 +3,12 @@
 package proc
 
 import (
+	"bufio"
+	"bytes"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/proc/proc_unix_test.go
```diff
@@ -3,27 +3,70 @@
...
 	"testing"
 	"time"
 
 	protest "github.com/derekparker/delve/pkg/proc/test"
 )
 
+// Checks that p.Pid() is the pid of fixture.
+func checkPid(t *testing.T, p *Process, fixture protest.Fixture) error {
+	if p.Pid() <= 0 {
+		return fmt.Errorf("zero or negative: %d", p.Pid())
```
derekparker: We're passing in `*testing.T` here, let's just fail the test instead of returning an error.

aarzilli: Done

aarzilli:
> Overall I'm more curious why we get invalid pids in this test?

I'm experimenting with adding the lldb-server/debugserver backend and it turns out that on linux (maybe also macOS, I haven't tested that far yet) qProcessInfo returns the wrong PID. I killed my desktop session twice.

derekparker:
@aarzilli we discussed this a bit offline, can we close it?

Issue #772, opened 2017-03-19
debugged process is left suspended after exiting delve


alexbrainman:
1. What version of Delve are you using (`dlv version`)?

Delve Debugger
Version: 0.12.1
Build:

2. What version of Go are you using? (`go version`)?

go version go1.7 windows/amd64

3. What operating system and processor architecture are you using?

windows/amd64

4. What did you do?

```
c:\Users\Alex\dev\src\a>type main.go
package main

import (
        "fmt"
        "os"
        "time"
)

func main() {
        fmt.Printf("my pid is %v\n", os.Getpid())
        time.Sleep(time.Minute)
}

c:\Users\Alex\dev\src\a>go build -o a.exe main.go && start a.exe

c:\Users\Alex\dev\src\a>dlv attach 3784
Type 'help' for list of commands.
(dlv) q
Would you like to kill the process? [Y/n] n

c:\Users\Alex\dev\src\a>
```

5. What did you expect to see?

I expect a.exe to continue running after I exit from dlv.

6. What did you see instead?

a.exe process is suspended after I exit from dlv.

Alex

derekparker:
@aarzilli I believe this has been fixed, let me know if I'm wrong about that.

PR #773, opened 2017-03-22
proc: add test for attach/detach, fix detach

aarzilli:
Detach did not work for processes we attach to via PID.

Linux: we were only detaching from the main thread, all threads are
detached independently

Windows: we must resume all threads before detaching.

macOS: still broken.

Updates #772

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -6,6 +6,7 @@ import (
 	"go/ast"
 	"go/constant"
 	"go/token"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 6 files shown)


pkg/proc/proc_windows.go
```diff
@@ -663,6 +663,20 @@ func (dbp *Process) resume() error {
...
 }
 
+func (dbp *Process) detach() error {
+	for _, thread := range dbp.threads {
+		_, err := _ResumeThread(thread.os.hThread)
+		if err != nil {
+			return err
+		}
+	}
+	err := PtraceDetach(dbp.pid, 0)
```
alexbrainman: s/err :=/return /

pkg/proc/proc_test.go
```diff
@@ -2581,3 +2581,46 @@ func TestStacktraceWithBarriers(t *testing.T) {
...
+	}
+	fixture := protest.BuildFixture("testnextnethttp")
+	cmd := exec.Command(fixture.Path)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	assertNoError(cmd.Start(), t, "starting fixture")
+
+	// wait for testnextnethttp to start listening
+	for {
+		conn, err := net.Dial("tcp", "localhost:9191")
```
alexbrainman: What happens if net.Dial will never return err == nil?

pkg/proc/proc_test.go
```diff
@@ -2581,3 +2581,46 @@ func TestStacktraceWithBarriers(t *testing.T) {
...
+			break
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	p, err := Attach(cmd.Process.Pid)
+	assertNoError(err, t, "Attach")
+	go func() {
+		time.Sleep(1 * time.Second)
+		http.Get("http://localhost:9191")
```
alexbrainman: Should not you check for http.Get failure?

pkg/proc/proc_test.go
```diff
@@ -2581,3 +2581,46 @@ func TestStacktraceWithBarriers(t *testing.T) {
...
+	assertNoError(p.Continue(), t, "Continue")
+
+	f, ln := currentLineNumber(p, t)
+	if ln != 11 {
+		t.Fatalf("Expected line :11 got %s:%d", f, ln)
+	}
+
+	assertNoError(p.Detach(false), t, "Detach")
+
+	_, err = http.Get("http://localhost:9191/nobp")
```
alexbrainman: Maybe do not throw away page contents, but check it has "hello, world!" string on it.

pkg/proc/proc_windows.go
```diff
@@ -663,6 +663,20 @@ func (dbp *Process) resume() error {
 	return nil
 }
 
+func (dbp *Process) detach() error {
```
alexbrainman: If you are introducing detach function, I would get rid of ptrace_windows.go file altogether. PtraceAttach and PtraceDetach function names do not make sense on windows. I was confused when reading code. Maybe not in this PR.

pkg/proc/proc_test.go
aarzilli: This is just to make the p.Continue below hit a breakpoint, what happens afterwards doesn't really matter.
If the breakpoint isn't hit the test won't terminate because p.Continue doesn't return, FailNow called on a different goroutine doesn't work anyway.

aarzilli:
Done.

derekparker:
@aarzilli CI failing.

aarzilli:
Pushed the same commit again and it passed, maybe something with appveyor?

PR #775, opened 2017-03-22
Add an entry for Gogland

justinclift:
Not sure if it's better to wait for the official release of Gogland, or point it out now so more people test it.

In near daily use for the past few weeks, it's been better than using the IntelliJ plugin version.  Which itself was pretty good.


justinclift:
Not sure if it's better to wait for the official release of Gogland, or point it out now so more people test it.

In near daily use for the past few weeks, it's been better than using the IntelliJ plugin version.  Which itself was pretty good.

PR #782, opened 2017-04-03
More refactoring + core integration + debugserver backend

aarzilli:
This is a continuation of the refactoring work I started in #745 progressing it to the point where it's actually possible to write multiple implementations of `target.Interface`.

To help me with the refactoring I also produced two new implementations of `target.Interface`:

1. `proc.CoreProcess` properly integrates the changes in #726
2. `proc.GdbserverProcesss` provides an implementation of `target.Interface` using an instance of `lldb-server` or `debugserver` as backend. I've written a longer explanation of how this work in a comment at the top of gdbserver.go.

I've also changed `dlv` to use `proc.GdbserverProcess` by default on macOS since there it's more reliable, as you can see this is the first time we get a green checkmark on macOS in a long while.
This also makes delve go gettable on macOS.

I don't consider the refactoring work finished with this, IMHO `proc` should be split to separate each backend into its own package, we would therefore have:

- `proc/native` for our old native `target.Interface` implementation
- `proc/core` for CoreProcess
- `proc/gdbserver` for GdbserverProcess
- `proc/test` for `proc_test.go` (because of cyclic imports)
- `proc` for all the stuff common to all backends (stacktraces, disassembly, Continue/Next/Step logic, etc.)

I haven't done this now because reviewing this will take a long time and splitting proc would be to disruptive.

I think the individual commits in this branch should be reviewed one by one, separately, I just posted this as a whole so that we can have a general idea of where we go.

### DIFF
--- a/pkg/proc/core.go
+++ b/pkg/proc/core.go
@@ -9,15 +9,6 @@ import (
 	"sync"
 )
 
-// MemoryReader is like io.ReaderAt, but the offset is a uintptr so that it
-// can address all of 64-bit memory.
-// Redundant with memoryReadWriter but more easily suited to working with
-// the standard io package.
-type MemoryReader interface {
-	// ReadMemory is just like io.ReaderAt.ReadAt.
-	ReadMemory(buf []byte, addr uintptr) (n int, err error)
-}
-
 // A SplicedMemory represents a memory space formed from multiple regions,
 // each of which may override previously regions. For example, in the following
 // core, the program text was loaded at 0x400000:
@@ -203,13 +194,12 @@ func (p *CoreProcess) BinInfo() *BinaryInfo {
 	return &p.bi
 }
 
-func (thread *CoreThread) readMemory(addr uintptr, size int) (data []byte, err error) {
-	data = make([]byte, size)
-	n, err := thread.p.core.ReadMemory(data, addr)
+func (thread *CoreThread) ReadMemory(data []byte, addr uintptr) (n int, err error) {
+	n, err = thread.p.core.ReadMemory(data, addr)
 	if err == nil && n != len(data) {
 		err = ErrShortRead
 	}
-	return data, err
+	return n, err
 }
 
 func (thread *CoreThread) writeMemory(addr uintptr, data []byte) (int, error) {
--- a/pkg/proc/core.go
+++ b/pkg/proc/core.go
@@ -1,0 +1,146 @@
+package proc
+
+import (
+	"fmt"
+	"io"
+)
+
+// MemoryReader is like io.ReaderAt, but the offset is a uintptr so that it
+// can address all of 64-bit memory.
+// Redundant with memoryReadWriter but more easily suited to working with
+// the standard io package.
+type MemoryReader interface {
+	// ReadMemory is just like io.ReaderAt.ReadAt.
+	ReadMemory(buf []byte, addr uintptr) (n int, err error)
+}
+
+// A SplicedMemory represents a memory space formed from multiple regions,
+// each of which may override previously regions. For example, in the following
+// core, the program text was loaded at 0x400000:
+// Start               End                 Page Offset
+// 0x0000000000400000  0x000000000044f000  0x0000000000000000
+// but then it's partially overwritten with an RW mapping whose data is stored
+// in the core file:
+// Type           Offset             VirtAddr           PhysAddr
+//                FileSiz            MemSiz              Flags  Align
+// LOAD           0x0000000000004000 0x000000000049a000 0x0000000000000000
+//                0x0000000000002000 0x0000000000002000  RW     1000
+// This can be represented in a SplicedMemory by adding the original region,
+// then putting the RW mapping on top of it.
+type SplicedMemory struct {
+	readers []readerEntry
+}
+
+type readerEntry struct {
+	offset uintptr
+	length uintptr
+	reader MemoryReader
+}
+
+// Add adds a new region to the SplicedMemory, which may override existing regions.
+func (r *SplicedMemory) Add(reader MemoryReader, off, length uintptr) {
+	if length == 0 {
+		return
+	}
+	end := off + length - 1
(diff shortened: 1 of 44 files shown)


derekparker:
This is great, I will TAL at this today, thanks for this!

derekparker:
@aarzilli can you rebase this.

aarzilli:
Rebased.

There are two major ugly things in this:

1. I had to make an interface for threads. I couldn't put it in pkg/target because of cyclic imports and I couldn't call it proc.Thread because it's already taken. I called it proc.IThread but it's not very go-like. I think it would be better if we renamed proc.Process to proc.NativeProcess, proc.Thread to proc.NativeThread and then used proc.Thread for the interface type.

2. I had to introduce a number of interfaces in proc that are essentially subsets of target.Interface. This includes a proc.IProcess, a copy of target.Interface, inside proc_test so that I could make the tests work on any implementation of target.Interface. IMHO it would be better if target.Inteface was moved inside proc as proc.Process.

There are also two shortcomings of using debugserver on macOS:

1. There is no way to find out the path to the executable we are debugging, this means the user has to specify it manually on attach.

2. There is no way to send a SIGSEGV back to the target process, this isn't just a problem with debugserver, lldb also can't do that.

pkg/proc/linux_amd64_core.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+	return &Core{
+		MemoryReader: memory,
+		Threads:      threads,
+		Pid:          pid,
+	}, nil
+}
+
+type Core struct {
```
derekparker: Doc comment on exported type.

pkg/proc/linux_amd64_core.go
```diff
@@ -0,0 +1,307 @@
...
+	Nice                 int8
+	_                    [4]uint8
+	Flag                 uint64
+	Uid, Gid             uint32
+	Pid, Ppid, Pgrp, Sid int32
+	Fname                [16]uint8
+	Args                 [80]uint8
+}
+
+type LinuxPrStatus struct {
```
derekparker: For all the below: Need doc comments on exported types.

pkg/proc/proc.go
```diff
@@ -329,91 +339,117 @@ func (dbp *Process) Next() (err error) {
...
+	if err := dbp.Halt(); err != nil {
+		return nil, dbp.exitGuard(err)
+	}
+	if err := dbp.setCurrentBreakpoints(trapthread); err != nil {
+		return nil, err
+	}
+	return trapthread, err
+}
+
+type Continuable interface {
```
derekparker: Needs doc comment.

pkg/proc/proc.go
```diff
@@ -618,43 +656,54 @@ func (dbp *Process) SwitchGoroutine(gid int) error {
...
 	}
 	if g.thread != nil {
-		return dbp.SwitchThread(g.thread.ID)
+		return dbp.SwitchThread(g.thread.ThreadID())
 	}
 	dbp.selectedGoroutine = g
 	return nil
 }
 
+type AllGCache interface {
```
derekparker: Doc comment.

pkg/proc/proc.go
```diff
@@ -894,23 +947,29 @@ func (dbp *Process) FindGoroutine(gid int) (*G, error) {
 	return nil, fmt.Errorf("Unknown goroutine %d", gid)
 }
 
+type EvalScopeConvertible interface {
```
derekparker: Doc comment.

pkg/proc/threads.go
```diff
@@ -32,6 +31,23 @@ type Thread struct {
 	os             *OSSpecificDetails
 }
 
+type IThread interface {
```
derekparker: Doc comment.

service/debugger/debugger.go
```diff
@@ -78,6 +97,46 @@ func New(config *Config) (*Debugger, error) {
...
+		if runtime.GOOS == "darwin" {
+			return proc.LLDBLaunch(processArgs, wd)
+		}
+		return proc.Launch(processArgs, wd)
+	default:
+		return nil, fmt.Errorf("unknown backend %q", d.config.Backend)
+	}
+}
+
+var ErrNoAttachPath = errors.New("must specify executable path on macOS")
```
derekparker: nit: Doc comment.

pkg/proc/core.go
```diff
@@ -0,0 +1,587 @@
...
+		newReaders = append(newReaders, e)
+	}
+	inserted := false
+	// Walk through the list of regions, fixing up any that overlap and inserting the new one.
+	for _, entry := range r.readers {
+		entryEnd := entry.offset + entry.length - 1
+		switch {
+		case entryEnd < off:
+			// Entry is completely before the new region.
+			newReaders = append(newReaders, entry)
```
aarzilli: Maybe? cc @heschik

pkg/proc/core.go
```diff
@@ -0,0 +1,587 @@
...
+			pb = pb[:entry.offset+entry.length-addr]
+		}
+		pn, err := entry.reader.ReadMemory(pb, addr)
+		n += pn
+		if err != nil || pn != len(pb) {
+			return n, err
+		}
+		buf = buf[pn:]
+		addr += uintptr(pn)
+		if len(buf) == 0 {
```
aarzilli: cc @heschik

pkg/proc/core.go
heschik: Yeah, the entry will never be 0 length but it's easier to read and more consistent.

pkg/proc/core.go
heschik: We just shrank it on 116. A read could go across multiple mapped regions, and we want to track how much is left to read.

aarzilli:
Done and rebased.

pkg/proc/core.go
derekparker: Got it, SGTM.

PR #786, opened 2017-04-07
proc refactoring: make stack, disassemble and eval independent of proc.Process

aarzilli:
Second step on the way to #782

### DIFF
--- a/pkg/proc/disasm.go
+++ b/pkg/proc/disasm.go
@@ -16,14 +16,27 @@ const (
 	IntelFlavour
 )
 
+func (dbp *Process) Disassemble(g *G, startPC, endPC uint64) ([]AsmInstruction, error) {
+	if g == nil {
+		regs, _ := dbp.currentThread.Registers(false)
+		return Disassemble(dbp.currentThread, regs, dbp.breakpoints, &dbp.bi, startPC, endPC)
+	}
+
+	var regs Registers
+	thread := dbp.currentThread
+	if g.thread != nil {
+		thread = g.thread
+		regs, _ = g.thread.Registers(false)
+	}
+
+	return Disassemble(thread, regs, dbp.breakpoints, &dbp.bi, startPC, endPC)
+}
+
 // Disassemble disassembles target memory between startPC and endPC
 // If currentGoroutine is set and thread is stopped at a CALL instruction Disassemble will evaluate the argument of the CALL instruction using the thread's registers
 // Be aware that the Bytes field of each returned instruction is a slice of a larger array of size endPC - startPC
-func (thread *Thread) Disassemble(startPC, endPC uint64, currentGoroutine bool) ([]AsmInstruction, error) {
-	if thread.dbp.exited {
-		return nil, &ProcessExitedError{}
-	}
-	mem, err := thread.readMemory(uintptr(startPC), int(endPC-startPC))
+func Disassemble(memrw memoryReadWriter, regs Registers, breakpoints map[uint64]*Breakpoint, bi *BinaryInfo, startPC, endPC uint64) ([]AsmInstruction, error) {
+	mem, err := memrw.readMemory(uintptr(startPC), int(endPC-startPC))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


pkg/proc/disasm.go
```diff
@@ -16,14 +16,27 @@ const (
 	IntelFlavour
 )
 
+func (dbp *Process) Disassemble(g *G, startPC, endPC uint64) ([]AsmInstruction, error) {
```
derekparker: Needs doc comment.

pkg/proc/disasm.go
```diff
@@ -16,14 +16,27 @@ const (
...
+
 // Disassemble disassembles target memory between startPC and endPC
 // If currentGoroutine is set and thread is stopped at a CALL instruction Disassemble will evaluate the argument of the CALL instruction using the thread's registers
 // Be aware that the Bytes field of each returned instruction is a slice of a larger array of size endPC - startPC
-func (thread *Thread) Disassemble(startPC, endPC uint64, currentGoroutine bool) ([]AsmInstruction, error) {
-	if thread.dbp.exited {
-		return nil, &ProcessExitedError{}
-	}
-	mem, err := thread.readMemory(uintptr(startPC), int(endPC-startPC))
+func Disassemble(memrw memoryReadWriter, regs Registers, breakpoints map[uint64]*Breakpoint, bi *BinaryInfo, startPC, endPC uint64) ([]AsmInstruction, error) {
```
derekparker: Seems like a cleaner signature would be to simply pass in a `[]byte` representing the read memory, and let this function focus just on the actual parsing, disassembling.

pkg/proc/disasm_amd64.go
```diff
@@ -63,7 +63,7 @@ func (inst *AsmInstruction) IsCall() bool {
 	return inst.Inst.Op == x86asm.CALL || inst.Inst.Op == x86asm.LCALL
 }
 
-func (thread *Thread) resolveCallArg(inst *ArchInst, currentGoroutine bool, regs Registers) *Location {
+func resolveCallArg(inst *ArchInst, currentGoroutine bool, regs Registers, mem memoryReadWriter, bininfo *BinaryInfo) *Location {
```
derekparker: Big fan of taking these ugly unexported methods off of their callers! :+1:

pkg/proc/disasm_amd64.go
```diff
@@ -143,10 +139,14 @@ func init() {
 	}
 }
 
+func (dbp *Process) FirstPCAfterPrologue(fn *gosym.Func, sameline bool) (uint64, error) {
```
derekparker: Needs doc comment.

pkg/proc/eval.go
```diff
@@ -67,7 +67,11 @@ func (scope *EvalScope) evalAST(t ast.Expr) (*Variable, error) {
 		// try to interpret the selector as a package variable
 		if maybePkg, ok := node.X.(*ast.Ident); ok {
 			if maybePkg.Name == "runtime" && node.Sel.Name == "curg" {
-				return scope.Thread.getGVariable()
+				if scope.gvar == nil {
+					return nilVariable, nil
+				} else {
```
derekparker: Don't need this `else` clause since we're returning.

pkg/proc/moduledata.go
```diff
@@ -11,9 +11,9 @@ type moduleData struct {
 	typemapVar    *Variable
 }
 
-func (bi *BinaryInfo) loadModuleData(thread *Thread) (err error) {
+func (bi *BinaryInfo) loadModuleData(mem memoryReadWriter) (err error) {
```
derekparker: Would kind of like to take these unexported methods and turn them into just functions that accept `bi *BinaryInfo` ... just seems cleaner to me.

pkg/proc/proc.go
```diff
@@ -179,36 +179,44 @@ func (dbp *Process) LoadInformation(path string) error {
 	return nil
 }
 
+func (dbp *Process) FindFileLocation(fileName string, lineno int) (uint64, error) {
```
derekparker: Needs doc comment.

pkg/proc/proc.go
```diff
@@ -179,36 +179,44 @@ func (dbp *Process) LoadInformation(path string) error {
...
 		return 0, err
 	}
 	if fn.Entry == pc {
-		pc, _ = dbp.FirstPCAfterPrologue(fn, true)
+		pc, _ = FirstPCAfterPrologue(mem, breakpoints, bi, fn, true)
 	}
 	return pc, nil
 }
 
+func (dbp *Process) FindFunctionLocation(funcName string, firstLine bool, lineOffset int) (uint64, error) {
```
derekparker: Needs doc comment.

pkg/proc/variables.go
```diff
@@ -132,15 +132,16 @@ type G struct {
...
-	dbp      *Process
 }
 
 // EvalScope is the scope for variable evaluation. Contains the thread,
 // current location (PC), and canonical frame address.
 type EvalScope struct {
-	Thread *Thread
-	PC     uint64
-	CFA    int64
+	PC   uint64
```
derekparker: nit: Can you add comments to exported struct members.

pkg/proc/disasm.go
aarzilli: For the most barebones functionality, yes. But we are also marking breakpoint and PC positions and resolving the CALL argument whenever possible. In the future I would also like to resolve SP relative addresses to local variable names.

pkg/proc/disasm_amd64.go
aarzilli: Both for this and Disassemble I've added a doc comment to target.Interface.

aarzilli:
Done.

PR #793, opened 2017-04-19
service/test: change test package name to service_test

hyangah:
According to https://golang.org/cmd/go/#hdr-Test_packages
service_test is more appropriate because this directory contains
no non-test code and the intention is to compile these *_test.go
files as a separate package and link/run with the main test package.

### DIFF
--- a/service/test/common_test.go
+++ b/service/test/common_test.go
@@ -1,4 +1,4 @@
-package servicetest
+package service_test
 
 import (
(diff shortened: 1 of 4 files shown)


hyangah:
This looks like a cosmetic change which doesn't affect the test functionality but following the convention mentioned in the document makes it easy for us to use some code analysis & automation tools.

PR #798, opened 2017-04-21
Fixed delve's version extraction to allow propsals

allada:
Go recently introduced proposal tags to their version tags, we
are simply allowing delve to handle it appropriately.

See:
https://github.com/golang/go/commit/0954fdd51e390e209317d4652d7effc15512fd50


derekparker:
Thanks for the patch!

PR #804, opened 2017-04-26
proc/gdbserial: mozilla rr support

aarzilli:
I've only tested on our test set but it should work. 

To try this mozilla rr needs to be compiled from source and installed. RR itself needs a recent-ish linux kernel (>= 3.5) on a recent-ish Intel CPU (nehalem or later). No macOS, no Windows.
To record a program and start debugging:

```
dlv --backend=rr exec ...
dlv --backend=rr debug ...
dlv --backend=rr test ...
```

Or record using `rr record` and then use `dlv replay`:

```
dlv replay path-to-trace-dir
```

New commands:
* `checkpoint` creates a checkpoint at the current position
* `checkpoints` lists checkpoints
* `clear-checkpoint c1` removes checkpoint 1
* `restart c1` restarts from checkpoint 1

Caveats:
* Can't change the value of variables or registers
* Can't send signals to inferior, value of proc.Process.Pid is wrong
* `runtime.Breakpoint` doesn't work as one would naively expect, with live processes we catch the signal and exit from runtime.Breakpoint, here we can't because the process is recorded as crashing (in other words, hard-coded breakpoints and debugger inserted breakpoints behave differently)
* Continuing backwards is slower


cmd/dlv/cmds/commands.go
```diff
@@ -240,6 +242,24 @@ core dump was taken.`,
 	}
 	RootCommand.AddCommand(versionCommand)
 
+	if path, _ := exec.LookPath("rr"); path != "" {
```
aarzilli: I've added a link to mozilla rr's github.

aarzilli:
I'd also forgotten to regenerate usage documents. BTW cobra/doc isn't vendored.

justinclift:
Hmmm, Go rr support is added to a good IDE (Gogland or whatever), that would almost sway me to move coding to Linux (for the rr support).

Well, maybe for tricky bugs nothing else works for solving. :wink:

@zolotov @ignatov @dlsniper Any chance? 😁

dlsniper:
I've added a request for this in the tracker https://youtrack.jetbrains.com/issue/GO-3831

aarzilli:
I should point out that this is all experimental, rr needs to be compiled from source (all releases lack a feature that delve needs) and there is little field testing of rr on go binaries. There is no fundamental reason why it shouldn't work, but you never know.

justinclift:
@aarzilli Sure.  Doesn't hurt to try things out the file bugs for the rough bits though.  If it turns out to be reasonably ok, it would be a fairly killer feature. :smile:

PR #813, opened 2017-05-02
Add instructions to clone delve in case of error

thomaslangston:
Not sure if `go get` should be doing this clone automatically, but here's a workaround if it fails to clone delve.

### DIFF
--- a/Documentation/installation/windows/install.md
+++ b/Documentation/installation/windows/install.md
@@ -7,3 +7,9 @@ go get github.com/derekparker/delve/cmd/dlv
 ```
 
 Note: If you are using Go 1.5 you must set `GO15VENDOREXPERIMENT=1` before continuing. The `GO15VENDOREXPERIMENT` env var simply opts into the [Go 1.5 Vendor Experiment](https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/).
+
+If you encounter an error saying "Not a git repository", clone delve first to the /src/github.com/dererkparker/delve directory.
+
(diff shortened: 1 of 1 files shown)


aarzilli:
> If you encounter an error saying "Not a git repository",

why would this happen? can you make this happen reliably?

thomaslangston:
I'll try tonight to reproduce on another machine. I got this error after
installing VS Code with the Go plugin, which appears to try to install
delve automatically and failed. I installed delve manually afterwards.

I'm assuming by your response that 'go get' should clone by itself.

On May 2, 2017 9:53 AM, "Alessandro Arzilli" <notifications@github.com>
wrote:

If you encounter an error saying "Not a git repository",

why would this happen? can you make this happen reliably?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/derekparker/delve/pull/813#issuecomment-298642045>, or mute
the thread

aarzilli:
> I'm assuming by your response that 'go get' should clone by itself.

yes.

thomaslangston:
Worked fine on my other machine. Not sure what's different about the first
one, other than it is Win 7 instead of Win 10.

On Tue, May 2, 2017 at 2:19 PM, Alessandro Arzilli <notifications@github.com

aarzilli:
I'm voting against this. If there really was a circumstance where `go get` failed and `git clone` didn't it would be either a problem with the documentation above or a bug in `go get`. But it's more likely that @thomaslangston just experienced a transitory network problem, or something like that.

derekparker:
Closed due to this seeming to be more of a transient problem.

PR #831, opened 2017-05-09
proc: next, stepout should work on recursive goroutines

### LINKED ISSUE #828: Support next/step/stepout with recursive functions

 dlv version
Delve Debugger
Version: 1.0.0-rc.1
Build:

go version go1.8 windows/amd64

Windows 10 x64

4. What did you do?
Recursive algorithm in Go 
https://gist.github.com/tylerlrhodes/150ec1d1341515e8c0ba291927614bdd

Ability to step through the code, ie use next

6. What did you see instead?
API server listening at: 127.0.0.1:3411
8
Failed to next.

This happened when it began to unwind the stack and go back up.  

I'm using Visual Studio Code

aarzilli:
This is a substantial change to our implementation of next/step/stepout so maybe we should hold it until after 1.0.0. OTOH next/step/stepout won't work on recursive functions because currently we are only checking that we stay on the same goroutine, where we should be checking that we stay on the same *call frame* too. 

```
proc: next, stepout should work on recursive goroutines

Before this commit our temp breakpoints only checked that we would stay
on the same goroutine.
However this isn't enough for recursive functions we must check that we
stay on the same goroutine AND on the same stack frame (or, in the case
of the StepOut breakpoint, the previous stack frame).

This commit:
1. adds a new synthetic variable runtime.frameoff that returns the
offset of the current frame from the base of the call stack.
This is similar to runtime.curg
2. Changes the condition used for breakpoints on the lines of the
current function to check that runtime.frameoff hasn't changed.
3. Changes the condition used for breakpoints on the return address to
check that runtime.frameoff corresponds to the previous frame in the
stack.
4. All other temporary breakpoints (the step-into breakpoints and defer
breakpoints) remain unchanged.

Fixes #828

```

### DIFF
--- a/pkg/proc/threads.go
+++ b/pkg/proc/threads.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"go/ast"
+	"go/token"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -31,6 +32,8 @@ type Thread interface {
 	StepInstruction() error
 	// Blocked returns true if the thread is blocked
 	Blocked() bool
+	// SetCurrentBreakpoint updates the current breakpoint of this thread
+	SetCurrentBreakpoint() error
 }
 
 // Location represents the location of a thread.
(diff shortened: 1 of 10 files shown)


derekparker:
I'd like to get this merged during this RC window. I'll get this reviewed.

pkg/proc/threads.go
```diff
@@ -52,25 +55,29 @@ func (tbe ThreadBlockedError) Error() string {
 }
 
 // returns topmost frame of g or thread if g is nil
```
derekparker: Can you update this comment now that the return has changed.

aarzilli:
Done.

PR #832, opened 2017-05-09
proc/variables: protect form invalid G struct in parseG

aarzilli:
This problem came up while I was implementing #831, it's pretty rare and I can only seem to reproduce it with the lldb backend, it's caused by the length of G.waitreason containing garbage.

```
proc/variables: protect form invalid G struct in parseG

A waitreason string that has invalid length (because the G struct is
corrupted or being modified) could cause a crash.

```


derekparker:
Will merge after rebase.

Issue #838, opened 2017-05-12
Need documentation for lldb-server


joeshaw:
1. What version of Delve are you using (`dlv version`)?

1.0.0-rc.1

2. What version of Go are you using? (`go version`)?

go version go1.8.1 darwin/amd64

3. What operating system and processor architecture are you using?

macOS Sierra (10.12.3) on an Intel i7 (64-bit).

4. What did you do?

`delve exec myprogram`

5. What did you expect to see?

The delve interactive prompt.

6. What did you see instead?

`could not launch process: exec: "lldb-server": executable file not found in $PATH`

I think this is mostly a documentation issue, since the backend changed between 0.12.2 and 1.0.0-rc.1.  For me, `lldb-server` is `/Applications/Xcode.app/Contents/SharedFrameworks/LLDB.framework/Versions/A/Resources/lldb-server`.

(Things also don't work for me once I add that directory to my `$PATH` -- I get `could not launch process: dial tcp :49754: getsockopt: connection refused` -- so I am unsure what additional setup is needed on my system.)

tylerb:
I am having this same issue.

tylerb:
A bit more info: `dlv test` and `dlv exec` work for me on the command line. However, in VSCode, I receive the `dial tcp connection refused` error.

aarzilli:
What version of xcode do you have installed? What happens when you run your lldb-server?

tylerb:
@aarzilli

Xcode: Version 8.3.2 (8E2002)
lldb version: lldb-370.0.42 Swift-3.1

Running `lldb-server platform --listen *:12343` just sits and waits. I don't know what to do at this point to further debug it.

joeshaw:
For me:

```
$ xcodebuild -version
Xcode 8.3.2
Build version 8E2002

$ lldb --version
lldb-370.0.42
  Swift-3.1
```

aarzilli:
Just `lldb-server`, what's the output? Also the output of `which lldb-server` and `find / -name debugserver`.

If it works from the command line but not from inside VSCode it's probably just a PATH issue at this point.

joeshaw:
`which lldb-server` doesn't return anything for me.

```
$ find / -name lldb-server 2>/dev/null
/Applications/Xcode.app/Contents/SharedFrameworks/LLDB.framework/Versions/A/Resources/lldb-server
[...]

PR #846, opened 2017-05-20
proc/gdbserial: wait for a connection from stub as long as stub lives

aarzilli:
I think most of the problems in issue #838 may actually be due to people not noticing debugserver's authorization prompt, since debugserver won't connect back to us until user authorization is given we should wait forever for a connection as long as the debugserver process remains alive.

Waiting for a fixed amount of time was an ugly solution anyway.

PS. 20% of the code causes 80% of the bugs.

```
proc/gdbserial: wait for a connection from stub as long as stub lives

The authorization prompt on macOS can take a long time to be
acknowledged by the user, we should keep waiting for a connection as
long as the debugserver instance we launched remains alive.

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -86,7 +86,6 @@ const (
 	showLldbServerOutput     = false
 	logGdbWireMaxLen         = 120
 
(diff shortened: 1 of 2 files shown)


derekparker:
> PS. 20% of the code causes 80% of the bugs.

:joy:

PR #856, opened 2017-05-31
cmd/dlv: do not delete built executable when using rr backend

aarzilli:
```
cmd/dlv: do not delete built executable when using rr backend

Mozilla RR will create a trace directory that can be reused with the
replay verb, however if we delete the executable file the trace
directory will become useless.

The other acceptable behavior would be to also clean the trace
directory at the end, but I think this behavior is more useful and
easier to implement.

```


derekparker:
Hmmm, I think we should be consistent and cleanup after ourselves after the run, including removing the trace dir.

I'm open to discussing the best approach for the user though.

aarzilli:
I've changed it to delete the trace directory.

PR #858, opened 2017-05-31
debugger/locations: if locspec isn't found try interpreting it as expr

aarzilli:
```
debugger/locations: if locspec isn't found try interpreting it as expr

If the user tries to list the contents of a function pointer but
forgets the '*' operator the location lookup will fail and result in a
unhelpful "location not found" error.

Instead if the location lookup fails we should try interpreting the
locspec as if it was preceded by '*'.

```

### DIFF
--- a/service/debugger/locations.go
+++ b/service/debugger/locations.go
@@ -362,7 +362,15 @@ func (loc *NormalLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr s
 	}
 
 	if matching := len(candidateFiles) + len(candidateFuncs); matching == 0 {
-		return nil, fmt.Errorf("Location %q not found", locStr)
+		// if no result was found treat this locations string could be an
+		// expression that the user forgot to prefix with '*', try treating it as
+		// such.
+		addrSpec := &AddrLocationSpec{locStr}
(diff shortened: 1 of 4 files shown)


service/debugger/locations.go
```diff
@@ -380,7 +380,13 @@ func (loc *NormalLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr s
 		return []api.Location{{PC: addr}}, nil
 
 	case 0:
-		return nil, fmt.Errorf("Location \"%s\" not found", locStr)
+		addrSpec := &AddrLocationSpec{locStr}
```
derekparker: nit: Add more comments explaining why we're doing this in this case.

service/debugger/locations.go
```diff
@@ -380,7 +380,13 @@ func (loc *NormalLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr s
...
+		locs, err := addrSpec.Find(d, scope, locStr)
+		if err != nil {
+			return nil, fmt.Errorf("Location \"%s\" not found", locStr)
+		} else {
```
derekparker: nit: returning, can remove the `else` clause.

aarzilli:
Done.

PR #870, opened 2017-06-09
Return breakpoints when they already exist

dlsniper:
See https://youtrack.jetbrains.com/issue/GO-3931#comment=27-2224179 for more details


dlsniper:
The other option is not to shadow ` bp ` in this line: https://github.com/derekparker/delve/blob/354055836ad8b31bc7547482ac2487c679e4ac85/pkg/proc/threads.go#L244 

Please let me know which solution do you prefer.
Thank you.

PR #875, opened 2017-06-13
proc: auto-dereference local variables that escape to the heap

aarzilli:
```
proc: auto-dereference local variables that escape to the heap

The compiler a variable 'v' that escapes to the heap with a '&v' entry.
Auto dereference those local variables.

Fixe #871

```

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -45,6 +45,11 @@ type VariableFlags uint16
 
 const (
 	// VariableEscaped is set for local variables that escaped to the heap
+	//
+	// The compiler performs escape analysis on local variables, the variables
+	// that may outlive the stack frame are allocated on the heap instead and
+	// only the address is recorded on the stack. These variables will be
+	// marked with this flag.
 	VariableEscaped VariableFlags = (1 << iota)
 	// VariableShadowed is set for local variables that are shadowed by a
 	// variable with the same name in another scope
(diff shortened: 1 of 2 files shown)


pkg/proc/variables.go
```diff
@@ -39,6 +39,13 @@ const (
 	FloatIsNegInf
 )
 
+type VariableFlags uint16
+
+const (
+	// VariableEscaped is set for local variables that escaped to the heap
```
derekparker: nit: more explanation, what change in behavior does this flag cause.

pkg/proc/variables.go
```diff
@@ -1718,7 +1727,13 @@ func (scope *EvalScope) variablesByTag(tag dwarf.Tag, cfg LoadConfig) ([]*Variab
 		}
 	}
 
-	for _, v := range vars {
+	for i, v := range vars {
+		if name := v.Name; len(name) > 1 && name[0] == '&' {
```
derekparker: nit: `strings.HasPrefix("&")`?

derekparker: nit: `strings.HasPrefix("&")`?

aarzilli:
Done.

derekparker:
Needs rebase.

aarzilli:
Done.

derekparker:
Needs another rebase.

aarzilli:
An older version of this commit was inside the go1.9 support branch because I needed the Flags field there, too. Now this is just a couple of changes to comments.
Oops.

PR #876, opened 2017-06-14
proc: fix interaction of RequestManualStop and conditional breakpoints

aarzilli:
```
proc: fix interaction of RequestManualStop and conditional breakpoints

A conditional breakpoint that is hit but has the condition evaluate to
false can block a RequestManualStop from working. If the conditional
breakpoint is set on an instruction that is executed very frequently by
multiple goroutines (or many conditional breakpoints are set) it could
prevent all calls to RequestManualStop from working.

This commit fixes the problem by changing proc.Continue to exit
unconditionally after a RequestManualStop is called.

```

### DIFF
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -269,6 +269,10 @@ func (p *Process) RequestManualStop() error {
 	return nil
 }
 
+func (p *Process) ManualStopRequested() bool {
+	return false
+}
+
 func (p *Process) CurrentThread() proc.Thread {
 	return &Thread{p.currentThread, p}
 }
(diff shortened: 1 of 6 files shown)


aarzilli:
This is causing TestIssue462 to randomly hang in Travis. I can't reproduce it locally, I'll need to look into this.

aarzilli:
I think TestIssue462 getting stuck is unrelated to this change, it just made it slightly more likely. Unfortunately it doesn't happen on my VM and most of the time it doesn't even happen in travis so I'm not sure that I fixed it. I think the problem is that we are trying to load G for a thread that is blocked, as a temporary workaround if singlestep doesn't return within a time limit we fail the G loading. I also disabled G loading for all threads that are stopped on an unknown file/line.

derekparker:
Looks good, but a refactor I'm thinking is having getter/setter methods that make use of their respective mutexes so that we don't have to remember to pepper them throughout the code base.

aarzilli:
I've done it for gdbserial.(*Process).ctrlC but for the most part these variables need to be changed atomically with other operations so it can't be done with getter/setters. Do you have other refactors in mind?

PR #880, opened 2017-06-15
Fix various issues detected by megacheck

dlsniper:
I've ran [megacheck](http://honnef.co/go/tools/cmd/megacheck) and fixed a few of the things that came up there.

### DIFF
--- a/service/test/integration1_test.go
+++ b/service/test/integration1_test.go
@@ -740,7 +740,7 @@ func Test1ClientServer_FullStacktrace(t *testing.T) {
 					if arg.Name != "i" {
 						continue
 					}
-					t.Logf("frame %d, variable i is %v\n", arg)
+					t.Logf("frame %d, variable i is %v\n", i,  arg)
 					argn, err := strconv.Atoi(arg.Value)
 					if err == nil {
 						found[argn] = true
--- a/service/test/integration1_test.go
+++ b/service/test/integration1_test.go
@@ -172,7 +172,7 @@ func Test1ClientServer_exit(t *testing.T) {
 		if !state.Exited {
 			t.Fatalf("Expected exit after continue: %v", state)
 		}
-		state, err = c.GetState()
+		_, err = c.GetState()
 		if err == nil {
 			t.Fatal("Expected error on querying state from exited process")
 		}
@@ -699,6 +699,9 @@ func Test1ClientServer_SetVariable(t *testing.T) {
 		assertNoError(c.SetVariable(api.EvalScope{-1, 0}, "a2", "8"), t, "SetVariable()")
 
 		a2, err := c.EvalVariable(api.EvalScope{-1, 0}, "a2")
+		if err != nil {
+			t.Fatalf("Could not evaluate variable: %v", err)
+		}
 
 		t.Logf("a2: %v", a2)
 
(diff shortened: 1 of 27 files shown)


dlsniper:
One of the few other things that I've left out are things like this one: https://github.com/derekparker/delve/blob/cbf3e1c6764b0a9823f2331cea5369a17762edf1/pkg/proc/core/core.go#L98-L101
Basically there ` started ` is always false which means that the loop will always continue. I guess it should just be ` continue ` instead of everything else in the ` if ` statement?

service/test/integration1_test.go
```diff
@@ -699,6 +699,9 @@ func Test1ClientServer_SetVariable(t *testing.T) {
 		assertNoError(c.SetVariable(api.EvalScope{-1, 0}, "a2", "8"), t, "SetVariable()")
 
 		a2, err := c.EvalVariable(api.EvalScope{-1, 0}, "a2")
+		if err != nil {
+			t.Fatalf("Could not evaluate variable: %v", a2)
```
aarzilli: should be `err` not `a2`

service/test/integration2_test.go
```diff
@@ -762,6 +762,9 @@ func TestClientServer_SetVariable(t *testing.T) {
 		assertNoError(c.SetVariable(api.EvalScope{-1, 0}, "a2", "8"), t, "SetVariable()")
 
 		a2, err := c.EvalVariable(api.EvalScope{-1, 0}, "a2", normalLoadConfig)
+		if err != nil {
+			t.Fatalf("Could not evaluate variable: %v", a2)
```
aarzilli: should be `err` not `a2`

service/test/variables_test.go
```diff
@@ -81,6 +81,9 @@ func evalVariable(p proc.Process, symbol string, cfg proc.LoadConfig) (*proc.Var
 		}
 	} else {
 		scope, err = proc.GoroutineScope(p.CurrentThread())
+		if err != nil {
```
aarzilli: should be outside the else.

dlsniper: Good catch!

dlsniper:
@aarzilli done, thanks for the review. I'd say, let's merge this first then I can do more small cleanups like that or the one pointed here:  https://github.com/derekparker/delve/pull/880#issuecomment-308893857 What do you think?

aarzilli:
TestIssue419 hangs on linux, nothing is reading from errChan when the output of Continue is written to it.

dlsniper:
I've rebased this against the latest master but for some reason tests like ` TestIssue411 ` just keep timing out even on master branch, see: https://gist.github.com/dlsniper/591dfbd3ee468e668f6d1a4ce18cce9f.

I'm not sure how to debug this as it seems that the executable gets created correctly, see the attachment
[delve_411.zip](https://github.com/derekparker/delve/files/1107056/delve_411.zip).

aarzilli:
TestIssue411 works fine for me, TestIssue419 still deadlocks. You are writing to errChan inside the goroutine, right after the kill, but the result of proc.Continue is also written to errChan and it can't terminate because nothing is draining errChan.

dlsniper:
@aarzilli updated, sorry for the delay. I've had to run this on a proper Ubuntu VM rather than WSL.

aarzilli:
lgtm

Issue #468, opened 2016-03-22
Command failed: no G executing on thread 0


npcode:
1. What version of Delve are you using (`dlv version`)?
   
   Delve Debugger
   Version: 0.11.0-alpha
   Build: 
2. What version of Go are you using? (`go version`)?
   
   go version go1.6 linux/amd64
3. What operating system and processor architecture are you using?
   
   OS: Linux 4.4.5-1-ARCH
   processor arch: x86_64
4. What did you do?
   1. Run `dlv attach 1` to debug a running [docker](https://github.com/docker/docker) daemon process. 
   2. Set a breakpoint
   3. Run 'n' command to step over next source line.
5. What did you expect to see?
   
   Step over to next source line.
6. What did you see instead?
   
   "Command failed: no G executing on thread 0" error as follows:
   
   ```
   > github.com/docker/libnetwork.(*network).CreateEndpoint() ./vendor/src/github.com/docker/libnetwork/network.go:740 (hits total:3) (PC: 0x8a6e60)
   735:                 if mac, ok := opt.(net.HardwareAddr); ok {
   736:                         ep.iface.mac = mac
   737:                 }
   738:         }
   739:
   => 740:         ipam, err := n.getController().getIPAM(n.ipamType)
   741:         if err != nil {
   742:                 return nil, err
   743:         }
   744:
   745:         if ipam.capability.RequiresMACAddress {
   (dlv) n
   Command failed: no G executing on thread 0
   ```

aarzilli:
Apparently in the docker binary the G offset is completely different

```
	docker.go:17	0x40ca00	64488b0c25c8ffffff		mov rcx, qword ptr fs:[0xffffffc8]
```

just compiling something with cgo still produces the good old:

```
	cgotest.go:12	0x4014c0	64488b0c25f8ffffff		mov rcx, qword ptr fs:[0xfffffff8]
```

Hard to say what causes this since docker build process involves downloading a whole operating system.

The `thread 0` thing is just cosmetic and I will push a fix for that.

joegrasse:
@aarzilli just curious, how did you produce the output to show the G offset in https://github.com/derekparker/delve/issues/468#issuecomment-200294856?

aarzilli:
`disasm -l main.main`, but it's the first instruction of almost any function.

heschik:
This should be fixed by #883 if someone wants to test and close it.

PR #883, opened 2017-06-19
proc: read G struct offset from runtime.tlsg if possible

heschik:
When a Go program is externally linked, the external linker is
responsible for picking the TLS offset. It records its decision in the
runtime.tlsg symbol. Read the offset from that rather than guessing -16.

This implementation causes a regression: pre-1.5.2 (I think?) versions
will no longer work, since they apparently had a different default TLS
offset, and reading the version during initialization is quite awkward.
If you still want to support versions that old I can go back to the
previous initialization order. Maybe we should add a DW_AT_producer
attribute so that reading the version is easier.

Fixes issue #468.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -28,13 +28,14 @@ type BinaryInfo struct {
 	// Maps package names to package paths, needed to lookup types inside DWARF info
 	packageMap map[string]string
 
-	Arch         Arch
-	dwarf        *dwarf.Data
-	frameEntries frame.FrameDescriptionEntries
-	lineInfo     line.DebugLines
-	goSymTable   *gosym.Table
-	types        map[string]dwarf.Offset
-	functions    []functionDebugInfo
+	Arch          Arch
+	dwarf         *dwarf.Data
+	frameEntries  frame.FrameDescriptionEntries
+	lineInfo      line.DebugLines
+	goSymTable    *gosym.Table
+	types         map[string]dwarf.Offset
+	functions     []functionDebugInfo
+	gStructOffset uint64
 
 	loadModuleDataOnce sync.Once
 	moduleData         []moduleData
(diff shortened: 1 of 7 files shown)


aarzilli:
work in progress I presume?

heschik:
I didn't think so, since all the Linux tests passed on my machine :( I'm not sure what the problem is elsewhere since I didn't think I'd made any substantive changes there, but I'll look harder.

aarzilli:
You are only loading runtime.tlsg for elf executables, which are only used on linux, not for PE (windows) and Mach-o (macOS).

heschik:
Oh, I thought you were referring to the test failures, which I think were because I got the Mac and Windows constants flipped.

I don't think any change is needed except on Linux. On Linux, the runtime cooperates with the standard libc TLS mechanism, so the location is variable. On Windows (https://golang.org/src/runtime/cgo/gcc_windows_amd64.c) and Mac(https://golang.org/src/runtime/cgo/gcc_darwin_amd64.c), there's assembly that stashes the G pointer, so it can be stored at a fixed location. Maybe you know something I don't?

aarzilli:
> Maybe you know something I don't?

Nope, I just didn't notice that you were initializing gStructOffset with constants for windows and macOS.

pkg/proc/bininfo.go
```diff
@@ -224,6 +232,44 @@ func (bi *BinaryInfo) parseDebugLineInfoElf(exe *elf.File, wg *sync.WaitGroup) {
 	}
 }
 
+func (bi *BinaryInfo) setGStructOffsetElf(exe *elf.File, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	// This is a bit arcane. Essentially:
+	// - If the program is pure Go, it can whatever it wants, and puts the G
```
aarzilli: > it can whatever it wants

do?

heschik:
Thanks, that fixed it. I put an error in to try to confirm the theory, but it never showed up anywhere. I'm hoping the latest failure on Linux 1.8 is a flake.

Any opinion on losing 1.5?

derekparker:
Kicked over the Linux 1.8 failed job.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -269,17 +278,11 @@ func (p *Process) Connect(conn net.Conn, path string, pid int) error {
 	if addr, err := p.conn.allocMemory(256); err == nil {
 		if _, err := p.conn.writeMemory(uintptr(addr), p.loadGInstr()); err == nil {
 			p.loadGInstrAddr = addr
+		} else {
```
aarzilli: I'd revert this. We have other ways to load g if this fails and this wouldn't have detected your bug anyway (the problem wasn't writing the instruction, it was in executing it later on).

heschik: Done.

aarzilli:
LGTM

PR #910, opened 2017-07-05
Pass LD_/DYLD_ env vars to debugserver (fixes #877)

### LINKED ISSUE #877: DYLD_ / LD_ environment variables will not be forwarded correctly on OS X
Hi,

Here's an interesting issue: https://youtrack.jetbrains.com/issue/GO-4001#comment=27-2237730

According to the information available here: https://stackoverflow.com/a/33589760/951414 it seems that OS X doesn't like forwarding any of the DYLD_ or LD_ to the applications anymore, for "security" reasons.

As such, when using the lldb backend on OS X, some applications will fail to launch. The IDE will forward the said environment variable when configured manually but I'm not sure if Delve itself needs to then do anything special to pick it up and then forward it itself to the lldb backend.

I'll try and work on a patch for this soon, but given my 0 experience with this stuff, any pointers / solutions would be welcomed.

(and yes, I know that the users can disable the "security" feature but I'm not sure what effects that will have)

dlsniper:
Unfortunately this is something I can't verify on my machine so I'll let the CI / users to do it for me.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -2934,3 +2934,23 @@ func TestRecursiveNext(t *testing.T) {
 		}
 	})
 }
+
+// TestIssue877 ensures that the environment variables starting with DYLD_ and LD_
+// are passed when executing the binary on OSX via debugserver
+func TestIssue877(t *testing.T) {
+	if runtime.GOOS != "darwin" && testBackend == "lldb" {
+		return
+	}
+	const envval = "/usr/local/lib"
+	os.Setenv("DYLD_LIBRARY_PATH", envval)
+	withTestProcess("issue877", t, func(p proc.Process, fixture protest.Fixture) {
+		assertNoError(proc.Continue(p), t, "Continue()")
+		v, err := evalVariable(p, "dyldenv")
+		assertNoError(err, t, "EvalVariable()")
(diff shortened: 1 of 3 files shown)


aarzilli:
Can you add a test for this? You could make a fixture that does:

```
dyldenv := os.GetEnv("DYLD_LIBRARY_PATH")
runtime.Breakpoint()
```

then in a test (in proc_test) set DYLD_LIBRARY_PATH to something and when the target process hits the breakpoint eval dyldenv and check that the value passed through.

dlsniper:
Will do. Sorry for being lazy.

dlsniper:
@aarzilli I've added a test. Thank you.

pkg/proc/proc_test.go
```diff
@@ -2934,3 +2934,21 @@ func TestRecursiveNext(t *testing.T) {
 		}
 	})
 }
+
+func TestIssue877(t *testing.T) {
+	if runtime.GOOS != "darwin" && testBackend == "lldb" {
```
aarzilli: Can you add a comment describing what the test does "Tests that environment variables starting with DYLD_" are passed to the target on macOS". Thank you.

dlsniper:
Done

PR #912, opened 2017-07-06
proc/core: support floating point registers

aarzilli:
```
proc/core: support floating point registers

Updates #794

```

### DIFF
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -145,14 +145,15 @@ type Process struct {
 	bi                proc.BinaryInfo
 	core              *Core
 	breakpoints       map[uint64]*proc.Breakpoint
-	currentThread     *LinuxPrStatus
+	currentThread     *Thread
 	selectedGoroutine *proc.G
 	allGCache         []*proc.G
 }
 
 type Thread struct {
-	th *LinuxPrStatus
-	p  *Process
+	th     *LinuxPrStatus
+	fpregs []proc.Register
+	p      *Process
 }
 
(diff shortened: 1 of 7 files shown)


pkg/proc/core/core.go
```diff
@@ -145,14 +145,15 @@ type Process struct {
...
+	currentThread     *Thread
 	selectedGoroutine *proc.G
 	allGCache         []*proc.G
 }
 
 type Thread struct {
-	th *LinuxPrStatus
-	p  *Process
+	th     *LinuxPrStatus
+	fpregs []proc.Register
```
derekparker: Why hold floating point registers as part of the `Thread` struct?

pkg/proc/core/linux_amd64_core.go
```diff
@@ -55,6 +55,7 @@ type LinuxCoreTimeval struct {
 }
 
 const NT_FILE elf.NType = 0x46494c45 // "FILE".
+const NT_X86_XSTATE elf.NType = 0x202
```
derekparker: Comment after as above?

pkg/proc/core/core.go
aarzilli: Because they are an attribute of a thread. LinuxPrStatus already contains the other CPU registers inside LinuxCoreRegisters, but both those structures are deserialized directly, so they can't be changed.

pkg/proc/core/linux_amd64_core.go
aarzilli: Done.

PR #923, opened 2017-07-13
service/debugger: improve NomalLocationSpec.Find's heuristics

hyangah:
When location spec is given and the base can be interpreted either in
source file name or function name, NomalLocationSpec searches both
the source file list and the function symbol list, and selects matching
candidates. Previously, all the matching candidates were added to one
single list regardless whether the candidate was from the source file
list or not. Then, later, Find tries to guess whether the candiate
was a function or a file based on a heuristic, i.e, whether the
candidate is an absolute file path. The heuristic is fragile - since
there is no guarantee that the included source file name is an absolute
path.

Instead, this CL preserves where the candidate was found; file list or
function symbol list. Then, use that info to determine whether the
candidate is a source file name or not.

### DIFF
--- a/service/debugger/locations.go
+++ b/service/debugger/locations.go
@@ -327,63 +327,66 @@ func (ale AmbiguousLocationError) Error() string {
 }
 
 func (loc *NormalLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string) ([]api.Location, error) {
-	funcs := d.target.BinInfo().Funcs()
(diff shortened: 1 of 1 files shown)


hyangah:
I have no idea how to interpret the appveyor build failure logs.

aarzilli:
Sometimes appveyor fails like this, we don't know why but it's unlikely to be our fault (since it's a compiler error).

Issue #823, opened 2017-05-05
OSX homebrew install failing to populate build version


SteelPhase:
Attempted to update delve using brew `brew reinstall --HEAD go-delve/delve/delve`
It appears as though this [commit](https://github.com/derekparker/delve/commit/24b20099aa3c464cc1f3cb245e82bd836a69c877) may have caused the issue. brew is still attempting to pass in BUILD_SHA

Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
```
==> Summary
🍺  /usr/local/Cellar/delve/HEAD-f29b9ed: 6 files, 10.2MB, built in 12 seconds
user@host:~$dlv version
Delve Debugger
Version: 0.12.2
Build: 
```
2. What version of Go are you using? (`go version`)?
```
go version go1.8.1 darwin/amd64
```
3. What operating system and processor architecture are you using?
`osx 10.12.4`
4. What did you do?
`dlv version`
5. What did you expect to see?
```
user@host:~$dlv version
Delve Debugger
Version: 0.12.2
Build: f29b9edad20b105f0bede65a6da15cbdd744ff77
```
6. What did you see instead?
```
user@host:~$dlv version
Delve Debugger
Version: 0.12.2
Build: 
```

renannprado:
I've just compiled today (without homebrew) and I also don't have the Build number.

SteelPhase:
This looks to be resolved thanks to the above referenced PR

PR #926, opened 2017-07-19
cmd/dlv: bugfix: make version command actually print revision number

aarzilli:
```
cmd/dlv: bugfix: make version command actually print revision number

```

In the meeting @hyangah pointed out that it's hard to tell which version of delve one is actually running.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,13 @@
 UNAME=$(shell uname)
 PREFIX=github.com/derekparker/delve
 GOVERSION=$(shell go version)
+BUILD_SHA=$(shell git rev-parse HEAD)
 LLDB_SERVER=$(shell which lldb-server)
 
 ifeq "$(UNAME)" "Darwin"
-    BUILD_FLAGS=-ldflags="-s"
+    BUILD_FLAGS=-ldflags="-s -X main.Build=$(BUILD_SHA)"
+else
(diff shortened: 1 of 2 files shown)


pkg/version/version.go
```diff
@@ -19,10 +22,24 @@ var (
...
+	return build
+}
+
 func (v Version) String() string {
 	ver := fmt.Sprintf("Version: %s.%s.%s", v.Major, v.Minor, v.Patch)
 	if v.Metadata != "" {
 		ver += "-" + v.Metadata
 	}
-	return fmt.Sprintf("%s\nBuild: %s", ver, v.Build)
+	return fmt.Sprintf("%s\nBuild: %s", ver, cleanBuild(v.Build))
```
hyangah: Thanks for the change.

Is it possible to change Makefile to include ldflags -X trick to set main.Build by default?
https://stackoverflow.com/questions/11354518/golang-application-auto-build-versioning

aarzilli: What's the advantage? This is changed automatically by git so it works even if the user doesn't use make to build.

aarzilli: Never mind, `$Id$` is actually useless in go :(

SteelPhase:
Fixes #823

PR #933, opened 2017-07-30
terminal: add config command

### LINKED ISSUE #927: Feature request: display full strings
When displaying long strings in the debugger, it will say "...+<num chars remaining> more",
eg "some really long string that is probably a url...+123 more". This makes it impossible to get the actual value out of the debugger because watches, debug console and trying to copy value all return this ellipsis version of the string.

I assume there was some reason for not displaying full strings which could understandably get big, but adding an option or way to enable verbatim strings would be nice.

aarzilli:
```
terminal: add config command

Fixes #927, #644

```


derekparker:
We should include aliases as well, since they are exposed via the config.

aarzilli:
Done.

pkg/terminal/command.go
```diff
@@ -218,6 +219,29 @@ Supported commands: print, stack and goroutine)`},
...
+	config <parameter> <value>
+	
+Changes the value of a configuration parameter.
+
+	config subistitute-path <from> <to>
+	config subistitute-path <from>
+	
+Adds or removes a path subistitution rule.
+
+	config alias <alias> <command>
```
derekparker: nit: I think this UX is kind of odd, mind switching it to `<command> <alias>`? Seems more intuitive to me (as in "alias 'foo' to 'bar'").

aarzilli:
Done.

PR #941, opened 2017-08-07
Better cgo support

aarzilli:
What the title says. Do not merge.

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -15,7 +15,7 @@ type Arch interface {
 	BreakpointInstruction() []byte
 	BreakpointSize() int
 	DerefTLS() bool
-	FixFrameUnwindContext(*frame.FrameContext) *frame.FrameContext
+	FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *BinaryInfo) *frame.FrameContext
 	RegSize(uint64) int
 	RegistersToDwarfRegisters(Registers) op.DwarfRegisters
 	GoroutineToDwarfRegisters(*G) op.DwarfRegisters
@@ -29,6 +29,12 @@ type AMD64 struct {
 	gStructOffset           uint64
 	hardwareBreakpointUsage []bool
 	goos                    string
+
+	// crosscall2fn is the DIE of crosscall2, a function used by the go runtime
+	// to call C functions. This function in go 1.9 (and previous versions) had
+	// a bad frame descriptor which needs to be fixed to generate good stack
+	// traces.
+	crosscall2fn *Function
 }
 
 const (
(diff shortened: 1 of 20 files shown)


aarzilli:
@heschik the last commit on this branch adds support for the location lists emitted with -dwarflocationlists. I thought you'd be interested.

BTW -dwarflocationlists looks broken to me, in multiple ways, 
1. if I compile [this](https://github.com/aarzilli/gdlv/blob/master/examples/vars.go) I get 3 DIE for each slice typed variable in main.main (`go version devel +180bfc4 Fri Aug 25 08:18:58 2017 +0000 linux/amd64`)
2. their loclistptr seem all bad (for example the first one is 0x11 which is in the middle of a location list entry, objdump can't even decode .debug_loc correctly).

heschik:
Thanks, mailed golang.org/cl/59070.

aarzilli:
I've split this into two. This part deals with making stacktraces involving cgo calls better, the other one (#958) has all the changes related to location expressions. I did this to make it easier to review, I can split further if it's still too big.

This still isn't enough to close the cgo issue but I think it makes things better and it's self contained enough that it can be reviewed and merged by itself. 

There seems to be a problem with cgo in go1.7 on macOS, I don't think it's a regression and since it's an old version of go and it works on 1.8 and 1.9 I think it's not worth fixing it.

derekparker:
This build seems to be failing now. You push more changes @aarzilli ?

aarzilli:
No, the test is disabled on go1.8 on windows, that's why it wasn't failing before. Now that appveyor is running go1.9 it's failing. I suspect a go bug here but I haven't investigated it yet.

aarzilli:
Problem was that pkg/dwarf/frame is too incomplete to handle the output of a C compiler, however just using the frame pointer is enough for now. Using the frame pointer is also what was happening on linux and macOS, it seems that there the go linker does not emit debug_frame entries for C functions.

aarzilli:
If https://github.com/golang/go/issues/21569 is fixed we can simplify this.

derekparker:
Needs rebase.

aarzilli:
Rebased.

pkg/proc/stack.go
```diff
@@ -164,6 +203,103 @@ func (it *stackIterator) Next() bool {
 	return true
 }
 
+// switchStack will use the current frame to determine if it's time to
+// switch between the system stack and the goroutine stack or vice versa.
+func (it *stackIterator) switchStack() bool {
```
derekparker: nit: Could be named better, also checks for "top of stack".

pkg/proc/arch.go
```diff
@@ -29,6 +29,8 @@ type AMD64 struct {
 	gStructOffset           uint64
 	hardwareBreakpointUsage []bool
 	goos                    string
+
+	crosscall2fn *Function
```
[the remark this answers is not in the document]
aarzilli: Done, however if we are ok with cgo stacktraces not working in go1.9 we could just remove this. I'm really tempted.

pkg/proc/stack.go
aarzilli: Suggestions for a better name?

PR #945, opened 2017-08-09
terminal: implement whatis command

aarzilli:
```
terminal: implement whatis command

Implements #596

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"math"
 	"os"
+	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


pkg/terminal/command.go
```diff
@@ -141,6 +142,9 @@ Called with more arguments it will execute a command on the specified goroutine.
 	[goroutine <n>] [frame <m>] print <expression>
 
 See $GOPATH/src/github.com/derekparker/delve/Documentation/cli/expr.md for a description of supported expressions.`},
+		{aliases: []string{"whatis"}, allowedPrefixes: scopePrefix, cmdFn: whatisCommand, helpMsg: `Prints type of an expression.
+		
+		whatis <expression>.`},
```
derekparker: Technically this won't eval an entire expression, let's just put `<var name>` or something like that instead.

aarzilli: it's actually the same as print, it will evaluate any expression.

derekparker: Ah, I see that now, seemed when I looked (and tested) before other args were getting ignored. Either way, this seems fine.

derekparker:
@aarzilli can you regen the docs for this PR and I'll merge.

aarzilli:
Done.

PR #947, opened 2017-08-13
proc/tests: fix intermittent test stall on windows

aarzilli:
```
proc/tests: fix intermittent test stall on windows

While we are waiting for the process to exit in native.(*Process).Kill
we could receive queued exception events, those must be continued or
the wait will never finish.

```

### DIFF
--- a/pkg/proc/native/proc_windows.go
+++ b/pkg/proc/native/proc_windows.go
@@ -186,7 +186,7 @@ func (dbp *Process) Kill() error {
 	_ = syscall.TerminateProcess(dbp.os.hProcess, 1)
 
 	dbp.execPtraceFunc(func() {
-		dbp.waitForDebugEvent(waitBlocking)
+		dbp.waitForDebugEvent(waitBlocking | waitDontHandleExceptions)
 	})
 
 	p.Wait()
@@ -237,6 +237,7 @@ type waitForDebugEventFlags int
 const (
 	waitBlocking waitForDebugEventFlags = 1 << iota
 	waitSuspendNewThreads
+	waitDontHandleExceptions
 )
 
 func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, exitCode int, err error) {
@@ -300,6 +301,10 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
 		case _RIP_EVENT:
 			break
 		case _EXCEPTION_DEBUG_EVENT:
+			if flags&waitDontHandleExceptions != 0 {
+				continueStatus = _DBG_EXCEPTION_NOT_HANDLED
+				break
+			}
 			exception := (*_EXCEPTION_DEBUG_INFO)(unionPtr)
 			tid := int(debugEvent.ThreadId)
 
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ endif
 else
 	go test -p 1 $(TEST_FLAGS) $(BUILD_FLAGS) $(ALL_PACKAGES)
 endif
-ifneq "$(shell which lldb-server)" ""
+ifneq "$(shell which lldb-server 2>/dev/null)" ""
 	@echo
 	@echo 'Testing LLDB backend (proc)'
 	go test $(TEST_FLAGS) $(BUILD_FLAGS) $(PREFIX)/pkg/proc -backend=lldb
@@ -76,7 +76,7 @@ ifneq "$(shell which lldb-server)" ""
 	@echo 'Testing LLDB backend (terminal)'
 	go test $(TEST_FLAGS) $(BUILD_FLAGS) $(PREFIX)/pkg/terminal -backend=lldb
 endif
-ifneq "$(shell which rr)" ""
+ifneq "$(shell which rr 2>/dev/null)" ""
 	@echo
 	@echo 'Testing Mozilla RR backend (proc)'
 	go test $(TEST_FLAGS) $(BUILD_FLAGS) $(PREFIX)/pkg/proc -backend=rr


alexbrainman:
@aarzilli how did you discover that? Can I test it here?

Alex

aarzilli:
I suspected native.(*Process).Kill so I added a bunch of printfs to it and ran it repeatedly on both appveyor and on a local VM. The local VM needs to be set to use multiple cores and you need to have the system under CPU load (I was watching a video). 
Appveyor was the first to fail: https://ci.appveyor.com/project/aarzilli/delve/build/8, which told me the call to os.(*Process).Wait was hanging, I decided to take a closer look to the waitForDebugEvent call immediately preceding, I added a bunch more printfs to it, conditional on a variable that I set to true only for that call. 
This time the VM testing was the first one to fail and it told me that waitForDebugEvent exited because it encountered an exception. That's all.

pkg/proc/native/proc_windows.go
```diff
@@ -300,6 +301,10 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
 		case _RIP_EVENT:
 			break
 		case _EXCEPTION_DEBUG_EVENT:
+			if flags&waitContinueAllExceptions != 0 {
+				continueStatus = _DBG_CONTINUE
```
alexbrainman: maybe

s/_DBG_CONTINUE/_DBG_EXCEPTION_NOT_HANDLED /

because _DBG_EXCEPTION_NOT_HANDLED will exit process regardless of situation.

pkg/proc/native/proc_windows.go
```diff
@@ -300,6 +301,10 @@ func (dbp *Process) waitForDebugEvent(flags waitForDebugEventFlags) (threadID, e
...
+				break
```
alexbrainman: Are you sure you want to exit from waitForDebugEvent? I would continue with the loop until you see EXIT_PROCESS_DEBUG_EVENT.

aarzilli:
> Your Launch function creates process and then calls newDebugProcess which calls waitForDebugEvent(waitBlocking) immediately, and that waitForDebugEvent(waitBlocking) call "handles" EXCEPTION_BREAKPOINT by doing nothing - your waitForDebugEvent code just returns tid to the caller. Perhaps it is OK in this situation.

I didn't write that code, but DebugActiveProcess is documents that the first thing returned by WaitForDebugEvent will be a CREATE_PROCESS_DEBUG_EVENT, followed by a series of CREATE_THREAD_DEBUG_EVENT, followed by a single EXCEPTION_DEBUG_EVENT. I imagine that CreateProcess with DEBUG_PROCESS will do something similar (otherwise that code wouldn't work, ever).

> But what might happens is that original Launch call might not get a chance to see EXCEPTION_BREAKPOINT. (Is that possible?).

I don't think it's possible and it's not what's happening here, if you [see the test where we got stuck here](https://ci.appveyor.com/project/aarzilli/delve/build/8) there's a lot of breakpoints happening between Launch and Kill.

> we want (I think) waitForDebugEvent to drain debug event queue until there is nothing left there

Yes, that's what we want, unfortunately we can't have it. When appveyor switched to dual core and windows builds started failing last year I rewrote `setCurrentBreakpoints` to drain the event queue, note how the waitForDebugEvent call there is non-blocking. The problem is, it doesn't always work.

Now, it *should* work, after WaitForDebugEvent returns windows has already "parked" all the threads in the process, so any exception event should be queued when we try to drain the queue in setCurrentBreakpoints, but apparently sometimes they aren't. Maybe I'm missing something. My inclination is to blame imperfect CPU emulation, did you manage to reproduce the bug on real hardware?

pkg/proc/native/proc_windows.go
aarzilli: This is breaking out of the switch not out of the loop: https://play.golang.org/p/vsmUbcdqba

pkg/proc/native/proc_windows.go
aarzilli: Done.

alexbrainman:
> I didn't write that code, but DebugActiveProcess is documents that the first thing returned by WaitForDebugEvent will be a CREATE_PROCESS_DEBUG_EVENT, followed by a series of CREATE_THREAD_DEBUG_EVENT, followed by a single EXCEPTION_DEBUG_EVENT

That is what I see once I print debugEvent.DebugEventCode straight after WaitForDebugEvent call. You can see it yourself.

> I don't think it's possible and it's not what's happening here, if you see the test where we got stuck here there's a lot of breakpoints happening between Launch and Kill.

I was not looking at this log. I assumed that TestIssue398 was hanging, but TestIssue398 has PASS in that log. But I think what is important is that we agreed that we should handle EXCEPTION_DEBUG_EVENT differently once TerminateProcess was called.

> did you manage to reproduce the bug on real hardware?

I am not sure which bug you are taking about. If you are talking about "*** Test killed: ran too long (10m0s)." on https://ci.appveyor.com/project/aarzilli/delve/build/8, then I could not reproduce it (I have Windows 10 with 4 cpus). But I can easily hang TestIssue398, if I insert dbp.Kill() into Launch just before it calls newDebugProcess (as I mentioned before). But that hang is fixed by this PR.

> 've also changed waitContinueAllExceptions to waitDontHandleExceptions.

SGTM

> This is breaking out of the switch not out of the loop

I don't even feel ashamed. :-)
This code is way to tricky to follow.

I think this PR is fine to submit. It won't make things worse.
LGTM

Thank you

Alex

aarzilli:
> I am not sure which bug you are taking about. If you are talking about "*** Test killed: ran too long (10m0s)." on https://ci.appveyor.com/project/aarzilli/delve/build/8, then I could not reproduce it

Yes, that's what I was talking about

PR #949, opened 2017-08-15
proc/gdbserial: automatically retrieve exe path on attach on macOS

aarzilli:
```
proc/gdbserial: automatically retrieve exe path on attach on macOS

debugserver doesn't support qXfer:exec-file:read, and it doesn't return
the executable path in the response to qProcessInfoPID, however we can
find out the executable path by using jGetLoadedDynamicLibrariesInfos.

```


derekparker:
Closing in favor of: https://github.com/derekparker/delve/pull/953

PR #958, opened 2017-09-01
Support location lists and more location expression opcodes

aarzilli:
Includes PR #941


pkg/proc/bininfo.go
```diff
@@ -224,6 +287,97 @@ func (bi *BinaryInfo) LoadError() error {
...
+
+	if debugFrameBytes != nil {
+		bi.frameEntries = frame.Parse(debugFrameBytes, frame.DwarfEndian(debugFrameBytes))
+	}
+
+	bi.loclistInit(debugLocBytes)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go bi.loadDebugInfoMaps(debugLineBytes, &wg)
```
derekparker: Maybe we modify this method to check if `wait group != nil` before calling done. That way here we don't have to go through the trouble of creating a wait group and spawning a goroutine.

pkg/proc/bininfo.go
```diff
@@ -224,6 +287,97 @@ func (bi *BinaryInfo) LoadError() error {
...
+}
+
+func (bi *BinaryInfo) loclistInit(data []byte) {
+	bi.loclist.data = data
+	bi.loclist.ptrSz = bi.Arch.PtrSize()
+}
+
+// Location returns the location described by attribute attr of entry.
+// This will either be an int64 address or a slice of bytes for locations
+// that don't correspond to a memory address (registers, composite locations)
```
derekparker: nit: punctuation.

pkg/proc/bininfo.go
```diff
@@ -224,6 +287,97 @@ func (bi *BinaryInfo) LoadError() error {
...
+	wg.Wait()
+}
+
+func (bi *BinaryInfo) loclistInit(data []byte) {
+	bi.loclist.data = data
+	bi.loclist.ptrSz = bi.Arch.PtrSize()
+}
+
+// Location returns the location described by attribute attr of entry.
+// This will either be an int64 address or a slice of bytes for locations
```
derekparker: Same nit as above similar comment for `ExecuteStackProgram`.

pkg/proc/bininfo.go
```diff
@@ -224,6 +287,97 @@ func (bi *BinaryInfo) LoadError() error {
...
+		return 0, nil, "", fmt.Errorf("could not find loclist entry at %#x for address %#x", off, pc)
+	}
+	var descr bytes.Buffer
+	fmt.Fprintf(&descr, "[%#x:%#x] ", off, pc)
+	op.PrettyPrint(&descr, instr)
+	addr, pieces, err := op.ExecuteStackProgram(regs, instr)
+	return addr, pieces, descr.String(), err
+}
+
+// loclistEntry returns the loclist entry in the loclist starting at off, for address pc
```
derekparker: nit: punctuation.

pkg/proc/bininfo.go
```diff
@@ -224,6 +287,97 @@ func (bi *BinaryInfo) LoadError() error {
...
+		}
+		if pc >= e.lowpc+base && pc < e.highpc+base {
+			return e.instr
+		}
+	}
+
+	return nil
+}
+
+// findCompileUnit returns the compile unit containing address pc
```
derekparker: nit: punctuation.

pkg/proc/mem.go
```diff
@@ -65,3 +74,62 @@ func cacheMemory(mem MemoryReadWriter, addr uintptr, size int) MemoryReadWriter
...
+	addr -= fakeAddress
+	if addr >= uintptr(len(mem.data)) || addr+uintptr(len(data)) > uintptr(len(mem.data)) {
+		return 0, errors.New("read out of bounds")
+	}
+	copy(data, mem.data[addr:addr+uintptr(len(data))])
+	return len(data), nil
+}
+
+func (mem *compositeMemory) WriteMemory(addr uintptr, data []byte) (int, error) {
+	//TODO: implement
```
derekparker: nit: `// TODO(aarzilli): ...`

pkg/dwarf/dwarfbuilder/builder.go
```diff
@@ -0,0 +1,51 @@
+// Package dwarfbuilder provides a way to build DWARF sections with
+// arbitrary contents.
+package dwarfbuilder
```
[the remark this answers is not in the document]
derekparker: I know this is only used in tests, but still would be helpful in coming back to this code.

pkg/config/config.go
```diff
@@ -39,6 +39,10 @@ type Config struct {
 	// MaxArrayValues is the maximum number of array items that the commands
 	// print, locals, args and vars should read (in verbose mode).
 	MaxArrayValues *int `yaml:"max-array-values,omitempty"`
+
+	// If ShowLocationExpr is true whatis will print the DWARF location
+	// expression for its argument.
+	ShowLocationExpr bool `yaml:"show-location-expr"`
```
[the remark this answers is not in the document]
aarzilli: It was useful to debug problems in the compiler side. I think knowing where a variable is stored is useful in general.

aarzilli:
Done.

pkg/config/config.go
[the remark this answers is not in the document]
derekparker: I'm hesitant to add features to Delve that are useful only for us an potentially not so much for the end user. This same feature could be done w/ `readelf` and some grepping from anyone hacking on Delve that needs to actually see this information.

How much of a pain would it be to split this out into a separate PR and have a discussion there, so we can pull in the rest of these changes?

Also, if we add this, it might be worth adding a flag to `whatis` to optionally print this information during a live debug session w/o having to set global (or session) config.

aarzilli:
show-location-expr removed

PR #978, opened 2017-10-04
proc: next should not skip lines with conditional bps

### LINKED ISSUE #844: Conditional breakpoints will cause next/step/stepout to miss lines
If a next/step/stepout command should stop on a line where a conditional breakpoint is already set it will only happen if the breakpoint's condition is also met.

aarzilli:
```
proc: next should not skip lines with conditional bps

Conditional breakpoints with unmet conditions would cause next and step
to skip the line.

This breakpoint changes the Kind field of proc.Breakpoint from a single
value to a bit field, each breakpoint object can represent
simultaneously a user breakpoint and one internal breakpoint (of which
we have several different kinds).

The breakpoint condition for internal breakpoints is stored in the new
internalCond field of proc.Breakpoint so that it will not conflict with
user specified conditions.

The breakpoint setting code is changed to allow overlapping one
internal breakpoint on a user breakpoint, or a user breakpoint on an
existing internal breakpoint. All other combinations are rejected. The
breakpoint clearing code is changed to clear the UserBreakpoint bit and
only remove the phisical breakpoint if no other bits are set in the
Kind field. ClearInternalBreakpoints does the same thing but clearing
all bits that aren't the UserBreakpoint bit.

Fixes #844

proc: breakpoints refactoring

Move some duplicate code, related to breakpoints, that was in both
backends into a single place.
This is in preparation to solve issue #844 (conditional breakpoints
make step and next fail) which will make this common breakpoint code
more complicated.

```

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -152,3 +152,104 @@ type NoBreakpointError struct {
 func (nbp NoBreakpointError) Error() string {
 	return fmt.Sprintf("no breakpoint at %#v", nbp.Addr)
 }
+
+type BreakpointMap struct {
+	M map[uint64]*Breakpoint
+
+	breakpointIDCounter         int
+	internalBreakpointIDCounter int
+}
+
+func NewBreakpointMap() BreakpointMap {
+	return BreakpointMap{
+		M: make(map[uint64]*Breakpoint),
+	}
+}
+
+func (bpmap *BreakpointMap) ResetBreakpointIDCounter() {
+	bpmap.breakpointIDCounter = 0
+}
+
+type writeBreakpointFn func(addr uint64) (file string, line int, fn *Function, originalData []byte, err error)
+type clearBreakpointFn func(*Breakpoint) error
+
+// Set creates a breakpoint at addr calling writeBreakpoint. Do not call this
+// function, call proc.Process.SetBreakpoint instead, this function exists
+// to implement proc.Process.SetBreakpoint.
+func (bpmap *BreakpointMap) Set(addr uint64, kind BreakpointKind, cond ast.Expr, writeBreakpoint writeBreakpointFn) (*Breakpoint, error) {
+	if bp, ok := bpmap.M[addr]; ok {
+		return bp, BreakpointExistsError{bp.File, bp.Line, bp.Addr}
+	}
+
+	f, l, fn, originalData, err := writeBreakpoint(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	newBreakpoint := &Breakpoint{
+		FunctionName: fn.Name,
+		File:         f,
+		Line:         l,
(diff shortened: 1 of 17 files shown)


pkg/proc/native/proc.go
```diff
@@ -193,71 +191,37 @@ func (dbp *Process) ManualStopRequested() bool {
...
 	}
-	newBreakpoint.OriginalData = originalData
-	dbp.breakpoints[addr] = newBreakpoint
 
-	return newBreakpoint, nil
+	return f, l, fn, originalData, nil
+}
+
+// SetBreakpoint sets a breakpoint at addr, and stores it in the process wide
+// break point table. Setting a break point must be thread specific due to
```
derekparker: nit: The fact that it must be thread specific is simply an implementation detail at this (the process) level. Since we're not allowing specification of a thread there, this comment seems more confusing to the programmer than enlightening.

service/debugger/debugger.go
```diff
@@ -398,11 +393,10 @@ func (d *Debugger) Breakpoints() []*api.Breakpoint {
 
 func (d *Debugger) breakpoints() []*api.Breakpoint {
 	bps := []*api.Breakpoint{}
-	for _, bp := range d.target.Breakpoints() {
-		if bp.Internal() {
-			continue
+	for _, bp := range d.target.Breakpoints().M {
+		if bp.Kind&proc.UserBreakpoint != 0 {
```
derekparker: Maybe have predicate methods such as `IsInternal()` & `IsUser()`?

pkg/proc/breakpoints.go
```diff
@@ -152,3 +173,165 @@ type NoBreakpointError struct {
 func (nbp NoBreakpointError) Error() string {
 	return fmt.Sprintf("no breakpoint at %#v", nbp.Addr)
 }
+
+type BreakpointMap struct {
```
derekparker: Comment new exported type.

pkg/proc/breakpoints.go
```diff
@@ -152,3 +173,165 @@ type NoBreakpointError struct {
...
+	M map[uint64]*Breakpoint
+
+	breakpointIDCounter         int
+	internalBreakpointIDCounter int
+}
+
+func NewBreakpointMap() BreakpointMap {
```
derekparker: Comment exported package function.

pkg/proc/breakpoints.go
```diff
@@ -152,3 +173,165 @@ type NoBreakpointError struct {
...
+	return BreakpointMap{
+		M: make(map[uint64]*Breakpoint),
+	}
+}
+
+func (bpmap *BreakpointMap) ResetBreakpointIDCounter() {
```
derekparker: Comment public method.

aarzilli:
Done.

PR #979, opened 2017-10-05
Replace all uses of gosymtab/gopclntab with debug_info/debug_line

aarzilli:
Gosymtab and gopclntab only contain informations about go code, linked C code isn't there, we should use debug_line instead to also cover C.

Also contains a few improvements to the code in `pkg/dwarf/line` to make it as fast as access to gosymtab/gopclntab was.


pkg/proc/types.go
```diff
@@ -129,62 +131,128 @@ func (bi *BinaryInfo) loadPackageMap() error {
...
-func (bi *BinaryInfo) findFunctionDebugInfo(pc uint64) (dwarf.Offset, error) {
-	i := sort.Search(len(bi.functions), func(i int) bool {
-		fn := bi.functions[i]
-		return pc <= fn.lowpc || (fn.lowpc <= pc && pc < fn.highpc)
-	})
-	if i != len(bi.functions) {
-		fn := bi.functions[i]
-		if fn.lowpc <= pc && pc < fn.highpc {
-			return fn.offset, nil
+func uniq(s []string) []string {
```
derekparker: Kind of hard for me to follow this algorithm here. Can you explain a bit what's going on? It seems you're trying to return a unique list of strings? Why not use a map`[string]struct{}` as a set instead?

aarzilli: It's the same as the unix command `uniq`:

```
       Filter  adjacent matching lines from INPUT (or standard input), writing
       to OUTPUT (or standard output).
```

so, yes, it's the same as using a set, only it doesn't allocate anything. I did it this way because I always do it this way, I can change it if you want.

derekparker: This will only filter successive duplicates, is that going to be an issue?

aarzilli: It's called after a sort so all duplicates are consecutive.

derekparker: SGTM

PR #985, opened 2017-10-10
proc: refactor stack.go to use DWARF registers

aarzilli:
```
proc: refactor stack.go to use DWARF registers

Instead of only tracking a few cherrypicked registers in stack.go track
all DWARF registers.

This is needed for cgo code and for the locationlists emitted by go in
1.10:
* The debug_frame sections emitted by C compilers can not be used
without tracking all registers
* the loclists emitted by go1.10 need all registers of a frame to be
interpreted.
```

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -1,5 +1,13 @@
 package proc
 
+import (
+	"encoding/binary"
+
+	"github.com/derekparker/delve/pkg/dwarf/frame"
+	"github.com/derekparker/delve/pkg/dwarf/op"
+	"golang.org/x/arch/x86/x86asm"
+)
+
 // Arch defines an interface for representing a
 // CPU architecture.
 type Arch interface {
@@ -7,6 +15,10 @@ type Arch interface {
 	BreakpointInstruction() []byte
 	BreakpointSize() int
 	DerefTLS() bool
+	FixFrameUnwindContext(*frame.FrameContext) *frame.FrameContext
+	RegSize(uint64) int
+	RegistersToDwarfRegisters(Registers) op.DwarfRegisters
+	GoroutineToDwarfRegisters(*G) op.DwarfRegisters
 }
 
 // AMD64 represents the AMD64 CPU architecture.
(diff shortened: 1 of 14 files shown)


derekparker:
@aarzilli needs rebase.

aarzilli:
Rebased

pkg/proc/arch.go
```diff
@@ -56,3 +74,182 @@ func (a *AMD64) BreakpointSize() int {
...
+	}
+	return 8
+}
+
+// The mapping between hardware registers and DWARF registers is specified
+// in the System V ABI AMD64 Architecture Processor Supplement page 57,
+// figure 3.36
+// https://www.uclibc.org/docs/psABI-x86_64.pdf
+
+var amd64DwarfToAsm = map[int]x86asm.Reg{
```
derekparker: nit: I think this would be clearer as `asm64DwarfToHardware`.

pkg/proc/arch.go
```diff
@@ -56,3 +74,182 @@ func (a *AMD64) BreakpointSize() int {
...
+	}
+	return max
+}
+
+// RegistersToDwarfRegisters converts hardware registers to the format used
+// by the DWARF expression interpreter.
+func (a *AMD64) RegistersToDwarfRegisters(regs Registers) op.DwarfRegisters {
+	dregs := make([]*op.DwarfRegister, maxAmd64DwarfRegister()+1)
+
+	dregs[amd64DwarfIPRegNum] = op.DwarfRegisterFromUint64(regs.PC())
```
derekparker: Instead of special casing these, why not just add them to the map above. We can still preserve the constants as they are used elsewhere.

pkg/proc/stack.go
```diff
@@ -87,62 +88,56 @@ func (n NullAddrError) Error() string {
...
+	frame Stackframe
+	bi    *BinaryInfo
+	mem   MemoryReadWriter
+	err   error
 
 	stackhi        uint64
 	stackBarrierPC uint64
 	stkbar         []savedLR
+
+	regs, callFrameRegs op.DwarfRegisters
```
derekparker: nit: comment explaining the difference between these two.

aarzilli:
Done

pkg/proc/arch.go
derekparker: Discussed offline.

PR #1002, opened 2017-10-30
travis: update go versions

dvrkps:


.travis.yml
```diff
@@ -7,8 +7,9 @@ os:
 
 go:
   - tip
-  - 1.8
-  - 1.7
+  - 1.9.x
```
aarzilli: should be: tip, 1.9 and 1.8.

dvrkps: @aarzilli done.

PR #1020, opened 2017-11-17
pkg/proc: remove callFrameRegs from stackIterator

aarzilli:
```
pkg/proc: remove callFrameRegs from stackIterator

Since it's just a scratchpad to calculate the new set of registers it
makes more sense to have it as a local variable in Next and
advanceRegs.

```

### DIFF
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -100,9 +100,8 @@ type stackIterator struct {
 	stackBarrierPC uint64
 	stkbar         []savedLR
 
-	// regs is the register set for the next frame, callFrameRegs is the
-	// register set for the call frame of the next frame.
-	regs, callFrameRegs op.DwarfRegisters
+	// regs is the register set for the current frame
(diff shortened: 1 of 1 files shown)


pkg/proc/stack.go
```diff
@@ -100,9 +100,8 @@ type stackIterator struct {
 	stackBarrierPC uint64
 	stkbar         []savedLR
 
-	// regs is the register set for the next frame, callFrameRegs is the
-	// register set for the call frame of the next frame.
-	regs, callFrameRegs op.DwarfRegisters
+	// regs is the register set for the next frame
```
derekparker: nit: Since we set regs after we iterate, wouldn't it be more accurate to say that it's the register set for the current frame?

aarzilli: Done

derekparker:
OSX builds seem to be taking a long time. Going to merge this as it isn't OS specific and other OSes have passed.

PR #1024, opened 2017-11-21
pkg/terminal: print DWARF location expression with whatis

aarzilli:
Split out of https://github.com/derekparker/delve/pull/958.

Another possibility is to make &x return a string with the location expression of x when it doesn't have a phisical address.

```
pkg/terminal: print DWARF location expression with whatis

Adds a configuration option (show-location-expr) that when activated
will cause the whatis command to also print the DWARF location
expression for a variable.
```


derekparker:
@aarzilli can you rebase, and then I will review.

aarzilli:
Done.

derekparker:
Overall LGTM -- but we should add this to the default config that gets written out.

aarzilli:
Done.

PR #1025, opened 2017-11-21
proc/native/linux: ignore ENODEV when retrieving fp registers

### LINKED ISSUE #1022: Multiple commands always fail with "Command failed: no such device"
Delve Debugger
Version: 1.0.0-rc.2
Build: $Id: 5c7676404d10d52e16be57fea744a7312b831f0c

go version go1.9.2 linux/amd64

Debian GNU/Linux 9.2 (stretch) - Intel Xeon E5520 (64 bit)

4. What did you do?
Ran `dlv debug hello.go`
(dlv) b hello.go:8
`Breakpoint 1 set at 0x49eecd for main.main() ./hello.go:8`
(dlv) c
```
> main.main() ./hello.go:8 (hits goroutine(1):1 total:1) (PC: 0x49eecd)
     3: import "fmt"
     4:
     5: func main() {
     6:         var s []byte
     7:         s = append(s, []byte("Hello, Debugging!")...)
=>   8:         fmt.Println(string(s))
     9: }
```
(dlv) n
```
n
     3: import "fmt"
     4:
     5: func main() {
     6:         var s []byte
     7:         s = append(s, []byte("Hello, Debugging!")...)
=>   8:         fmt.Println(string(s))
     9: }
```
After pressing 'n' I would expect the debugger should go ahead and print the value of the string.

6. What did you see instead?
**Command failed: no such device**

Also, the same error message is reported whenever I use the ``print`` command on a variable or
a data structure.

I don't understand from where this error comes up. What possibly prevents the debugger to 
execute such a simple program? Other commands from the debugger works as expected 
(e.g. ``goroutines,threads,funcs,etc.``).

I have also checked if the Yama security module is involved but I believe the setting is ok:
```
~$ cat /proc/sys/kernel/yama/ptrace_scope
0
```

The source of the Go program is extremely simple:
```
package main

import "fmt"

func main() {
        var s []byte
        s = append(s, []byte("Hello, Debugging!")...)
        fmt.Println(string(s))
}
```

aarzilli:
```
proc/native/linux: ignore ENODEV when retrieving fp registers

Either the CPU or the kernel may not support the calls we do when
retrieving floating point registers, this isn't an error we should
propagate.
Also improve the error reporint of pkg/proc/native.fpRegisters.

Fixes #1022

```

### DIFF
--- a/pkg/proc/native/ptrace_linux.go
+++ b/pkg/proc/native/ptrace_linux.go
@@ -59,7 +59,8 @@ func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
 // and Section 13.1 (and following) of Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 1: Basic Architecture
 func PtraceGetRegset(tid int) (regset proc.LinuxX86Xstate, err error) {
 	_, _, err = syscall.Syscall6(syscall.SYS_PTRACE, sys.PTRACE_GETFPREGS, uintptr(tid), uintptr(0), uintptr(unsafe.Pointer(&regset.PtraceFpRegs)), 0, 0)
-	if err == syscall.Errno(0) {
+	if err == syscall.Errno(0) || err == syscall.ENODEV {
+		// ignore ENODEV, it just means this CPU doesn't have X87 registers (??)
 		err = nil
 	}
 
@@ -67,6 +68,10 @@ func PtraceGetRegset(tid int) (regset proc.LinuxX86Xstate, err error) {
 	iov := sys.Iovec{Base: &xstateargs[0], Len: _X86_XSTATE_MAX_SIZE}
 	_, _, err = syscall.Syscall6(syscall.SYS_PTRACE, sys.PTRACE_GETREGSET, uintptr(tid), _NT_X86_XSTATE, uintptr(unsafe.Pointer(&iov)), 0, 0)
 	if err != syscall.Errno(0) {
+		if err == syscall.ENODEV {
+			// ignore ENODEV, it just means this CPU or kernel doesn't support XSTATE, see https://github.com/derekparker/delve/issues/1022
+			err = nil
+		}
 		return
 	} else {
 		err = nil
--- a/pkg/proc/native/registers_linux_amd64.go
+++ b/pkg/proc/native/registers_linux_amd64.go
@@ -1,6 +1,8 @@
 package native
 
 import (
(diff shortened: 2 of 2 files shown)


derekparker:
I think we're running into https://github.com/mozilla/rr/issues/1282. 

@mtds can you `cat /proc/cpuinfo` on the affected machine? I suspect the hardware does not support `xsave` as in the RR issue linked above. If so, we can implement a similar workaround and not use `PT_GETREGSET`.

derekparker:
@aarzilli I'm fine with that. I'm still going to open an issue to track if we ever want to implement the workaround.

mtds:
@derekparker: I am not sure if it still useful but the following is a printout of ``cat /proc/cpuinfo``
on my Linux workstation (8 cores). As you suspected, ``xsave`` is not supported (at least is
nowhere to be found on the ``flags`` line).

My second attempt was on a Linux VM (on KVM) where the CPU is marked as Intel and the model
is as follows: ``QEMU Virtual CPU version 2.5+``. Also in this case, ``xsave`` does not appear in
the ``flags`` line of cpuinfo. Anyway, this pull request from @aarzilli solved the problem in both
cases.

```
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 26
model name      : Intel(R) Xeon(R) CPU           E5520  @ 2.27GHz
stepping        : 5
microcode       : 0x11
cpu MHz         : 1596.000
cache size      : 8192 KB
physical id     : 0
siblings        : 8
core id         : 0
cpu cores       : 4
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 11
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx 
rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf eagerfpu pni dtes64 monitor ds_cpl vmx est tm2 ssse3 cx16
 xtpr pdcm dca sse4_1 sse4_2 popcnt lahf_lm tpr_shadow vnmi flexpriority ept vpid dtherm ida
bugs            :
bogomips        : 4533.46
clflush size    : 64
cache_alignment : 64
address sizes   : 40 bits physical, 48 bits virtual
power management:
```

PR #1027, opened 2017-11-23
Go 1.10 support

### LINKED ISSUE #186: `info locals` shows uninitialized value for as-yet undeclared variable

aarzilli:
- Changes command line options of `go build` and `go test -c` to take advantage of go 1.10 new caching
- Use DW_AT_producer to warn user about debugging optimized functions
- Use DW_TAG_constants to describe variable values
- Use DW_AT_decl_line to determine variable visibility (fixes #186, #83)
- Miscellaneous test fixes for go1.10.

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -54,6 +54,8 @@ const (
 	// VariableShadowed is set for local variables that are shadowed by a
 	// variable with the same name in another scope
 	VariableShadowed
+	// VariableConstant means this variable is a constant value
+	VariableConstant
 )
 
 // Variable represents a variable. It contains the address, name,
@@ -291,6 +293,7 @@ func newConstant(val constant.Value, mem MemoryReadWriter) *Variable {
 		v.Kind = reflect.String
 		v.Len = int64(len(constant.StringVal(val)))
 	}
+	v.Flags |= VariableConstant
 	return v
 }
 
@@ -642,7 +645,7 @@ func (scope *EvalScope) PackageVariables(cfg LoadConfig) ([]*Variable, error) {
 	return vars, nil
 }
 
-func (scope *EvalScope) packageVarAddr(name string) (*Variable, error) {
+func (scope *EvalScope) findGlobal(name string) (*Variable, error) {
 	for n, off := range scope.BinInfo.packageVars {
 		if n == name || strings.HasSuffix(n, "/"+name) {
 			reader := scope.DwarfReader()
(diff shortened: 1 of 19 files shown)


derekparker:
Needs rebase.

aarzilli:
Rebased

service/test/variables_test.go
```diff
@@ -949,3 +949,30 @@ func TestPackageRenames(t *testing.T) {
...
+		{"d", true, "33", "", "main.BitFieldType", nil},
+		{"e", true, "10", "", "main.ConstType", nil},
+		{"f", true, "0", "", "main.BitFieldType", nil},
+		{"bitZero", true, "1", "", "main.BitFieldType", nil},
+		{"bitOne", true, "2", "", "main.BitFieldType", nil},
+		{"constTwo", true, "2", "", "main.ConstType", nil},
+	}
+	ver, _ := goversion.Parse(runtime.Version())
+	if ver.Major > 0 && !ver.AfterOrEqual(goversion.GoVersion{1, 10, -1, 0, 0, ""}) {
+		// Not supported on 1.9 or earlier
```
derekparker: nit: `t.Skip("...")`

pkg/proc/variables.go
```diff
@@ -1724,6 +1749,100 @@ func (v *Variable) loadInterface(recurseLevel int, loadData bool, cfg LoadConfig
...
+func (v *Variable) ConstDescr() string {
+	if v.bi == nil || (v.Flags&VariableConstant != 0) {
+		return ""
+	}
+	ctyp := v.bi.consts.Get(v.DwarfType)
+	if ctyp == nil {
+		return ""
+	}
+	if typename := v.DwarfType.Common().Name; strings.Index(typename, ".") < 0 || strings.HasPrefix(typename, "C.") {
+		// only attempt to use constants for user defined type, otherwise every
```
derekparker: I don't understand this comment.

This patch seems to index constants by type, and then use the value & type information to evaluate the constant. I'm curious why we don't simply treat constants like other variables and evaluate them by name to avoid this confusion.

derekparker: Is this to support nicer describes by being able to represent constants as bitwise OR and such?

aarzilli: Yes. Take for example the type `proc.FloatSpecial`, I think when we print variables of types like that (that are essentially enums) the name of the constant should be printed instead.

PR #1028, opened 2017-11-23
Support --output for debug, trace, and test commands

Carpetsmoker:
With the `--output` parameter you can configure the output binary. For
example:

    dlv debug --output /tmp/xxx

Will build the binary to `/tmp/xxx`, instead of always putting it as
`debug` in the current directory.

This ensures that the command always works (even if there is already a
file or directory named `debug`) and doesn't write to the source
directory. Especially for things like Delve/Vim integration this is a
good thing to have, I think.


cmd/dlv/cmds/commands.go
```diff
@@ -272,36 +271,46 @@ https://github.com/mozilla/rr
 
 func debugCmd(cmd *cobra.Command, args []string) {
 	status := func() int {
-		var pkg string
-		dlvArgs, targetArgs := splitArgs(cmd, args)
-
-		if len(dlvArgs) > 0 {
-			pkg = args[0]
+		debugname := cmd.Flag("output").Value.String()
+		if !strings.HasPrefix(debugname, "/") {
```
aarzilli: This needs to be filepath.IsAbs for windows.

cmd/dlv/cmds/commands.go
```diff
@@ -272,36 +271,46 @@ https://github.com/mozilla/rr
...
+			debugname = "./" + debugname
+		}
+		debugname, err := filepath.Abs(debugname)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			return 1
+		}
+
```
aarzilli: This block of code is the same in debug, test and trace. I think it should be its own function.

aarzilli:
remember that windows paths do contain colons.

Carpetsmoker:
> remember that windows paths do contain colons.

Ah, very good, thanks! It's been so long that I had to develop anything on Windows that I sort-of forgot about all the small differences that can trip you up :-)

cmd/dlv/cmds/commands.go
Carpetsmoker: Actually the extra `filepath.IsAbs()` and such isn't needed, and can just be reduces to `filepath.Abs()` plus error check – I think it's okay to keep this inline like this?

Carpetsmoker:
For some reason Travis isn't running macOS tests today, but your feedback should be addressed now @aarzilli. Let me know if you'd like to see any other changes.

Issue #1031, opened 2017-11-28
Internal debugger error: <nil> not numeric on vars command


shevchenkodenis:
1. What version of Delve are you using (`dlv version`)?
```
~ % dlv version
Delve Debugger
Version: 1.0.0-rc.2
Build: $Id: 5c7676404d10d52e16be57fea744a7312b831f0c $
```
2. What version of Go are you using? (`go version`)?
```
~ % go version
go version go1.9.1 linux/amd64
```
3. What operating system and processor architecture are you using?
```
~ % uname -a
Linux 4.7.0-1-amd64 #1 SMP Debian 4.7.5-1 (2016-09-26) x86_64 GNU/Linux
```
4. What did you do?
Executed command _vars_
5. What did you expect to see?
Not error
6. What did you see instead?
```
(dlv) vars
Command failed: Internal debugger error: <nil> not numeric
runtime.call32 (0x4579ba)
	/usr/local/go/src/runtime/asm_amd64.s:509
runtime.gopanic (0x42b492)
	/usr/local/go/src/runtime/panic.go:491
go/constant.Real (0x5ec383)
	/usr/local/go/src/go/constant/value.go:656
github.com/derekparker/delve/service/api.ConvertVar (0x69b4e9)
	/home/denis/go/src/github.com/derekparker/delve/service/api/conversions.go:165
github.com/derekparker/delve/service/api.ConvertVar (0x69b3c3)
	/home/denis/go/src/github.com/derekparker/delve/service/api/conversions.go:180
github.com/derekparker/delve/service/debugger.(*Debugger).PackageVariables (0x708622)
	/home/denis/go/src/github.com/derekparker/delve/service/debugger/debugger.go:714
github.com/derekparker/delve/service/rpc2.(*RPCServer).ListPackageVars (0x8a3647)
	/home/denis/go/src/github.com/derekparker/delve/service/rpc2/server.go:324
runtime.call128 (0x457ad1)
	/usr/local/go/src/runtime/asm_amd64.s:511
reflect.Value.call (0x4b31e5)
	/usr/local/go/src/reflect/value.go:434
reflect.Value.Call (0x4b27c3)
	/usr/local/go/src/reflect/value.go:302
github.com/derekparker/delve/service/rpccommon.(*ServerImpl).serveJSONCodec.func1 (0x8ab1cd)
	/home/denis/go/src/github.com/derekparker/delve/service/rpccommon/server.go:299
[...]

PR #1032, opened 2017-11-28
api/conversions: fix converting of nil complex numbers

shevchenkodenis:
It will help on #1031

### DIFF
--- a/service/api/conversions.go
+++ b/service/api/conversions.go
@@ -148,30 +148,43 @@ func ConvertVar(v *proc.Variable) *Variable {
 		r.Children = make([]Variable, 2)
 		r.Len = 2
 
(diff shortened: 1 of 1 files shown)


derekparker:
Yeah, I'm also curious how a complex var has a nil Value, I suspect the actual issue lies somewhere else. @shevchenkodenis do you have any code that can repro this? Didn't see anything in the issue you created.

aarzilli:
I think I know how this happens, if a variable with a complex variable escapes it's replaced by a pointer, that pointer can have a 0 value if it's out of scope. If we get to a point where we try to evaluate it anyway (because of incomplete .debug_info, corrupted memory or whatever), because of the autodereferencing of escaped variables, we get a complex variable with address 0, which won't have childrens.

I'm ok with patching it like this.

PR #1037, opened 2017-12-04
proc: handle DW_TAG_subprogram with a nochildren abbrev

### LINKED ISSUE #1034: osx: performance issue when using cgo
99cad1044b695e4bd75b0431c4b25982636e4a71
go 1.9.2
osx 10.12.6 / amd64
4. What did you do?
debug app with cgo present
same network traffic between debug server and delve as an app without cgo
6. What did you see instead?
a lot more network traffic (around 30 times more)

Using the attached project, set a breakpoint on the indicated line and then step to the next line while monitoring the traffic between delve and debugserver.
Then, follow the same steps but disable the import cgo line.

[awesomeProject.zip](https://github.com/derekparker/delve/files/1525340/awesomeProject.zip)

When running from the CLI, the application crashes in my case, with a panic saying that there was a signal while nexting (or some similar message but I believe it might be because of how I run the debugging setup, if you want that exact error, please let me know).

I've attached the IDE logs that are relevant for delve, the IDE always sends the same requests, but in the CGO case, about 7MB worth of logs are generated while in the non-CGO case about 90KB.

[idea logs.zip](https://github.com/derekparker/delve/files/1525390/idea.logs.zip)

I'll dig more around this in the coming days as now I'll have more time to check this issue out but in case you have some ideas about this, that would be awesome.

Thank you for your time and help.

Original issue: https://youtrack.jetbrains.com/issue/GO-4923

aarzilli:
On macOS, externally linked programs will have an abbrev for
DW_TAG_subprogram without the haschildren flag set. We should handle
this case instead of expecting all DW_TAG_subprogram entries to have
list of children.

Fixes #1034

### DIFF
--- a/pkg/dwarf/reader/variables.go
+++ b/pkg/dwarf/reader/variables.go
@@ -55,7 +55,7 @@ func (vrdr *VariableReader) Next() bool {
 				}
 			}
 
-			if recur {
+			if recur && vrdr.entry.Children {
 				vrdr.depth++
 			} else {
 				if vrdr.depth == 0 {
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3337,3 +3337,21 @@ func TestSystemstackStacktrace(t *testing.T) {
 		}
 	})
 }
+
+func TestIssue1034(t *testing.T) {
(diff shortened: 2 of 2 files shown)


dlsniper:
As always, thank you!

PR #1044, opened 2017-12-14
Makefile: replace hardcoded paths with computed paths

dr2chase:
This makes it easier to build in forks for pull requests; no need to modify the Makefile to build and test in fork.

This works for people developing Go in the usual style; it's not clear to me what the test environment is, but I guess I'll find out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL=test
 UNAME=$(shell uname)
-PREFIX=github.com/derekparker/delve
+PREFIX=$(shell echo "${PWD}" | sed -e "s?${GOPATH}/src/??" )
 GOVERSION=$(shell go version)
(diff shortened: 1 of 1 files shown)


dlsniper:
@dr2chase I don't think that this would be enough of a change, since every other import path in delve would need to be changed as well to point to the fork.

The way I work is I ` go get ` delve, then add a new remote to my fork but keep the sources under ` github.com/derekparker/delve ` and I haven't had any issues so far. Do you think it could work for your setup as well? (I'm not sure if there are any differences between environments).

Hope it helps.

dr2chase:
Ugh, you're right, what was I thinking?
Never mind.

dlsniper:
Maybe you were thinking of a future without the GOPATH? :D

PR #1047, opened 2017-12-17
Makefile: handle empty GOPATH on darwin

zchee:
As of Go 1.8, allows empty GOPATH environment variable.
Fall back `make install` when GOPATH is not set on darwin OS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .DEFAULT_GOAL=test
 UNAME=$(shell uname)
 PREFIX=github.com/derekparker/delve
+GOPATH=$(shell go env GOPATH)
 GOVERSION=$(shell go version)
 BUILD_SHA=$(shell git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


aarzilli:
Maybe we should always use `go env GOPATH`?

zchee:
@aarzilli Yeah, I think yes. If you agreed, I’ll rewrite this pull request.

I intentionally have not set GOPATH to find a bug in the project(such as spf13/cobra).

aarzilli:
I think it would be better.

zchee:
@aarzilli Fixed. PTAL.

derekparker:
Can you reword the commit message, this change affects all platforms not just Darwin.

zchee:
@derekparker Okay, Will do.

zchee:
@derekparker @aarzilli Remove `darwin` from commit message. PTAL.

zchee:
@derekparker Thanks :D

PR #1049, opened 2017-12-17
dlv_test: disable one of the checks of TestOutput on windows

aarzilli:
```
dlv_test: disable one of the checks of TestOutput on windows

Sometimes delve on Appveyor can't remove the built binary before
exiting and gets an "Access is denied" error when trying.
See: https://ci.appveyor.com/project/derekparker/delve/build/1527

```

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -156,7 +156,12 @@ func testOutput(t *testing.T, dlvbin, output string, delveCmds []string) (stdout
 
 	_, err = os.Stat(debugbin)
 	if err == nil {
-		t.Errorf("running %q: file %v was not deleted\nstdout is %q, stderr is %q", delveCmds, debugbin, stdout, stderr)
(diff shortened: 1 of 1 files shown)


cmd/dlv/dlv_test.go
```diff
@@ -156,7 +156,12 @@ func testOutput(t *testing.T, dlvbin, output string, delveCmds []string) (stdout
 
 	_, err = os.Stat(debugbin)
 	if err == nil {
-		t.Errorf("running %q: file %v was not deleted\nstdout is %q, stderr is %q", delveCmds, debugbin, stdout, stderr)
+		if runtime.GOOS != "windows" {
```
derekparker: We should actually just check if we're running in AppVeyor via env vars it sets, we should still run this test otherwise (if people are running tests on their machines, VMs, etc). https://www.appveyor.com/docs/environment-variables/

Also, I wonder what change in AppVeyor led to this suddenly failing.

PR #1055, opened 2017-12-19
Remove limitation of exit notification only for specific API calls

dlsniper:
I've noticed that Delve is inconsistent in marking when a process exits based on the API call which triggered the exit to happen.

In my case, I've noticed this happening for using ` next ` (via RPC).

To have an example to reproduce this

```go
package main

import (
	"fmt"
	"go/ast"
	"os"
)

func main() {
	a := &ast.CompositeLit{}
	fmt.Println("demo", a) // set breakpoint here and use next twice 
	os.Exit(2)
}
```

This allows clients to have a consistent way to handle cases when the process exits.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -1387,3 +1387,32 @@ func TestClientServer_collectBreakpointInfoError(t *testing.T) {
 		assertNoError(state.Err, t, "Continue()")
 	})
 }
+
+func TestClientServerConsistentExit(t *testing.T) {
+	protest.AllowRecording(t)
+	// This test is useful because it ensures that Next and Continue operations both
+	// exit with the same exit status and details when the target application terminates.
+	// Other program execution API calls should also behave in the same way.
+	// An error should be pressent in state.Err.
+	withTestClient2("pr1055", t, func(c service.Client) {
+		fp := testProgPath(t, "pr1055")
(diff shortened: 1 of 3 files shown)


aarzilli:
Broke TestIssue355 on macOS, fixing is easy (I would just exclude SwitchGoroutine and SwitchThread from the if you are changing). You should also add a test, see service/test/integration2_test.go for some examples of how to test this.

dlsniper:
I've updated the PR as requested but sadly I don't have an OSX to test the patch on so we'll have to wait for Travis to finish the execution.

service/test/integration2_test.go
```diff
@@ -1387,3 +1387,28 @@ func TestClientServer_collectBreakpointInfoError(t *testing.T) {
 		assertNoError(state.Err, t, "Continue()")
 	})
 }
+
+func TestClientServer_PR1055(t *testing.T) {
+	protest.AllowRecording(t)
```
aarzilli: There should be a comment here describing what this is actually testing, as a future reminder, something like "Next should behave like continue when the program exits and return the error in state.Err".

Other than that looks good.

dlsniper: I've updated the PR accordingly. Thank you for your help!

dlsniper:
Done, please have a look.

dlsniper:
And to test what we were just talking about: Ping, please merge this when convenient. Thank you!

PR #1060, opened 2017-12-28
command/terminal: allow restart to change process args

yasushi-saito:
Add -args flag to "restart" command. For example, "restart -args a b c" will
pass args a b c to the new process.

Add "-c" flag to pass the checkpoint name. This is needed to disambiguate the
checkpoint name and arglist.

Reverted unnecessary changes.


pkg/terminal/command.go
```diff
@@ -18,6 +19,7 @@ import (
 	"strings"
 	"text/tabwriter"
 
+	"github.com/cosiner/argv"
```
aarzilli: New libraries need to be vendored.

pkg/terminal/command.go
```diff
@@ -97,11 +99,18 @@ See also: "help on", "help cond" and "help clear"`},
...
 	trace [name] <linespec>
-	
+
 A tracepoint is a breakpoint that does not stop the execution of the program, instead when the tracepoint is hit a notification is displayed. See $GOPATH/src/github.com/derekparker/delve/Documentation/cli/locspec.md for the syntax of linespec.
 
 See also: "help on", "help cond" and "help clear"`},
-		{aliases: []string{"restart", "r"}, cmdFn: restart, helpMsg: "Restart process."},
+		{aliases: []string{"restart", "r"}, cmdFn: restart, helpMsg: `Restart process.
+
+  restart [-c checkpoint] [-args newargv...]
```
aarzilli: Recordings can't get new arguments and live programs can't be restarted at a checkpoint, IMHO the -c/-args distinction is unnecessary and the arguments to `restart` can be interpreted contextually depending on what `client.Recorded()` returns.

service/debugger/debugger.go
```diff
@@ -29,6 +29,8 @@ import (
 // lower lever packages such as proc.
 type Debugger struct {
 	config *Config
+	// arguments to launch a new process.
```
aarzilli: What's the reason for moving this here from config?

service/rpccommon/server.go
```diff
@@ -89,7 +89,7 @@ func (s *ServerImpl) Stop(kill bool) error {
 }
 
 // Restart restarts the debugger.
-func (s *ServerImpl) Restart() error {
+func (s *ServerImpl) Restart(args []string) error {
```
aarzilli: There's no point in adding this argument here since you are not using it below.

service/debugger/debugger.go
yasushi-saito: I just thought it'd be strange to modify a config. I can move it back to config if you wish.

service/rpccommon/server.go
yasushi-saito: Ok, reverted.

pkg/terminal/command.go
yasushi-saito: done.

pkg/terminal/command.go
yasushi-saito: Ok, I changed restart to interpret the arg as the checkpoint when Recorded(). We still need to distinguish the case where (1) the user wants to restart with the current setting, and (2) the user wants to restart with an empty arg list. For (2) only, I retained the -args flag.

pkg/terminal/command.go
```diff
@@ -662,36 +666,45 @@ func writeGoroutineLong(w io.Writer, g *api.Goroutine, prefix string) {
 		prefix, formatLocation(g.GoStatementLoc))
 }
 
-func parseArgs(args string, flags *flag.FlagSet) error {
+func parseArgs(args string) ([]string, error) {
```
aarzilli: rename this parseRestartArgs

pkg/terminal/command.go
```diff
@@ -105,11 +104,16 @@ A tracepoint is a breakpoint that does not stop the execution of the program, in
 See also: "help on", "help cond" and "help clear"`},
 		{aliases: []string{"restart", "r"}, cmdFn: restart, helpMsg: `Restart process.
 
-  restart [-c checkpoint] [-args newargv...]
+  restart
```
aarzilli: cc @derekparker 

Personally I think this would be better like this:

```
       restart [-args newargv...]
       restart [checkpoint]

For recorded processes restarts from the start or from the specified checkpoint.
For normal processes restarts the process, optionally changing the arguments.
```

service/debugger/debugger.go
aarzilli: IMHO it's fine but let's hear what @derekparker thinks.

aarzilli:
PS. you need to commit _fixtures/restartargs.go

pkg/terminal/command.go
```diff
@@ -97,11 +98,23 @@ See also: "help on", "help cond" and "help clear"`},
...
-	
+
 A tracepoint is a breakpoint that does not stop the execution of the program, instead when the tracepoint is hit a notification is displayed. See $GOPATH/src/github.com/derekparker/delve/Documentation/cli/locspec.md for the syntax of linespec.
 
 See also: "help on", "help cond" and "help clear"`},
-		{aliases: []string{"restart", "r"}, cmdFn: restart, helpMsg: "Restart process."},
+		{aliases: []string{"restart", "r"}, cmdFn: restart, helpMsg: `Restart process.
+
+  restart
+  restart checkpoint
```
derekparker: nit: Mention this only works for `rr` backend.

pkg/terminal/command.go
```diff
@@ -97,11 +98,23 @@ See also: "help on", "help cond" and "help clear"`},
...
+  restart newargv...
+  restart -args
```
derekparker: nit: Would be clearer as `-noargs`.

pkg/terminal/command.go
```diff
@@ -97,11 +98,23 @@ See also: "help on", "help cond" and "help clear"`},
...
+
+  The first form restarts the process with the same commandline args or the checkpoint.
+  The second form is used to restarts the recorded program from the given checkpoint.
```
derekparker: s/restarts/restart/

pkg/terminal/command.go
```diff
@@ -653,8 +666,45 @@ func writeGoroutineLong(w io.Writer, g *api.Goroutine, prefix string) {
 		prefix, formatLocation(g.GoStatementLoc))
 }
 
+func parseArgs(args string) ([]string, error) {
+	if args == "" {
+		return nil, nil
+	}
+	v, err := argv.Argv([]rune(args), argv.ParseEnv(os.Environ()),
```
derekparker: Do we gain a lot by introducing this new library? If so, we should refactor other argument parsing to take advantage of it.

service/debugger/debugger.go
derekparker: I'm fine with it, but let's keep it named `processArgs`.

service/rpc2/server.go
```diff
@@ -62,6 +62,12 @@ type RestartIn struct {
 	// Position to restart from, if it starts with 'c' it's a checkpoint ID,
 	// otherwise it's an event number. Only valid for recorded targets.
 	Position string
+
+	// ResetArgs tell whether NewArgs should take effect.
+	ResetArgs bool
```
derekparker: Do we need this flag, can't we infer by the state of `NewArgs`?

yasushi-saito: Without this flag, we can't distinguish between "reuse the current args" and "use an empty args list".

pkg/terminal/command.go
yasushi-saito: For this particular use case, we need a non-trivial argv parser to handle things like string quoting and escaping. This package is the simplest one I could find. I'm not sure if there could be other uses of this package.

PR #1062, opened 2017-12-30
Added key `-u` for `go get` in linux installation readme 

duch94:
`go get` needs key `-u` to use network. Mentioned `go get github.com/dereparker/delve/cmd/dlv` without `-u` didn't work for me on debian 9.


aarzilli:
I guess it should also be changed in the windows documentation.

PR #1069, opened 2018-01-04
service/test: disable TestClientServerConsistentExit for rr backend

### LINKED ISSUE #1067: TestClientServerConsistentExit fails with rr backend
what the title says.

aarzilli:
```
service/test: disable TestClientServerConsistentExit for rr backend

The rr backend doesn't report the exit status (the argument of the W
packet seems to always be 0).

Fixes #1067

```

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -1389,7 +1389,6 @@ func TestClientServer_collectBreakpointInfoError(t *testing.T) {
 }
 
 func TestClientServerConsistentExit(t *testing.T) {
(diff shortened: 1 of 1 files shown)


derekparker:
Is this documented anywhere for RR? Seems like this might be a bug for us or them.

aarzilli:
https://github.com/mozilla/rr/blob/master/src/GdbServer.cc#L1104 but it's in the released version of rr, so if we want the tests to pass on rr we have to disable this.

PR #1080, opened 2018-01-11
Add "l" alias for list command

whit537:
This would help out those of us habituated to [pdb](https://docs.python.org/3/library/pdb.html). 😊

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -191,7 +191,7 @@ If regex is specified only package variables with a name matching it will be ret
 	
 Argument -a shows more registers.`},
 		{aliases: []string{"exit", "quit", "q"}, cmdFn: exitCommand, helpMsg: "Exit the debugger."},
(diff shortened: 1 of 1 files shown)


whit537:
Pdb inherits this behavior from [GDB](https://sourceware.org/gdb/current/onlinedocs/gdb/List.html#List), fwiw. Also, in both cases, additional `(l)ist` commands page further through the source code rather than relisting the same code snippet multiple times. Supporting that in Delve would take a bigger diff, though. ;)

whit537:
No apology necessary! Open source is voluntary! :)

Travis > Linux builds could use a restart? Timed out ...

derekparker:
>Travis > Linux builds could use a restart? Timed out ...

Yeah that's been happening on other PRs too, investigating now.

whit537:
> Yeah that's been happening on other PRs too, investigating now.

If chronic might deserve a ticket of its own?

derekparker:
Travis seems to be having a hard time with the Linux builds, tested locally and merged. Thanks!

PR #1091, opened 2018-01-24
[WIP] Documentation contribution guidelines

schomatis:
This is a placeholder for documenting the best practices of how to contribute documentation. It will be developed as different issues come up while writing documentation, particularly I'll be trying to [contribute internal documentation](https://github.com/derekparker/delve/issues/1074) that will have special considerations which might leave useful lessons for future contributions.

- [ ] How to comment functions. Use [godoc](https://blog.golang.org/godoc-documenting-go-code) style?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,3 +59,7 @@ The first line is the subject and should be no longer than 70 characters, the
 second line is always blank, and other lines should be wrapped at 80 characters.
 This allows the message to be easier to read on GitHub as well as in various
 git tools.
+
+## Contributing documentation
+
+The documentation is part of the repository and thus follows the same guidelines as the code, but there are also special considerations as the style of the writing and the format used.


schomatis:
@derekparker Regarding the inclusion of links to source code in the documentation.

Although they are not strictly necessary (an entity can be unambiguously qualified with its name and import path) they are very useful (specially to someone new to the code base) to be able to easily locate the entity the documentation is talking about. The downside is they are hard to maintain so should be used as least as possible and as less specific as possible (e.g., not including line numbers or specific commit versions).

I'm looking for options to establish some simple criteria of how to add them, to leave here in the guidelines (and use it myself for the internal documentation), some could be:

1. Link to master, pointing to the file that contains it.

2. Link to a specific major version release (e.g., when Delve hits v1.0.0) and adding line number (as it would be a reference to unmodifiable content).

3. Is there some GitHub functionality that can resolve a Go name, e.g., `debugger.Debugger`, to a link to the source code that defines it? Something like [sourcegraph](https://sourcegraph.com/github.com/derekparker/delve/-/blob/service/debugger/debugger.go#L30:6) does.

derekparker:
> Link to a specific major version release (e.g., when Delve hits v1.0.0) and adding line number (as it would be a reference to unmodifiable content).

I think this is a pretty safe option.

Issue #1074, opened 2018-01-09
Internal Documentation


schomatis:
I've noticed the [Internal Documentation](https://github.com/derekparker/delve/tree/master/Documentation/internal) remains empty. I'd like to contribute to it, as I see it as a great opportunity to know more about Delve and also Go internals.

I'm new here and I realize that understanding Delve's internals isn't an easy task (the *TODO* is assigned to its creator) but I just aim at getting the ball rolling contributing some minor PRs with whatever useful information I can provide.

I'm raising this issue to check if you think this is a doable task and see if any of the maintainers would be willing to provide me with some guidance of where to start, and answer some questions along the way (which I'll try to make as concrete as I can to take as little time from you as possible).

schomatis:
@derekparker Would it be OK to submit a small PR to give this a try?

derekparker:
@schomatis you can absolutely submit a small PR to provide this documentation, and I appreciate your interest in taking this on.

I would definitely be available for any questions you have regarding understanding the internals of Delve.

schomatis:
@derekparker Great, thanks. I'll start poking around and see what part of the code seems more accessible to me, unless you think there's a particularly good starting point.

derekparker:
@schomatis I suppose it depends a bit on what piques your interest the most and what part of the internals you'd like to document. If you are doing this also as a learning exercise for yourself it may help to start at the beginning and document how Delve builds, launches, and attaches to a process (and what that means). From there, there are any number of interesting things to document, including various Delve specific low-level features / functionality.

The double edged sword here is that the more detailed and specific the documentation is the harder it will be to maintain as the project is developed. I think there is a middle ground that will produce great, in depth documentation.

schomatis:
> start at the beginning and document how Delve builds, launches, and attaches to a process (and what that means).

I'll try to start there and see where it takes me.

> The double edged sword here is that the more detailed and specific the documentation is the harder it will be to maintain as the project is developed. I think there is a middle ground that will produce great, in depth documentation.

This is for me, by far, the most challenging aspect of writing this kind of documentation, which goes beyond the basic user guide, but aims at helping someone new (like me) delve into a project's internal workings (which change more rapidly than the user API).

I too believe there is a middle ground worth striving for (I don't think the best documentation is just commenting code, although that is the main part of it). I don't know a priori what is the correct balance, but it's worth thinking about it along the development of this documentation (I'm leaving PR #1091 open to  discuss this and document the reached conclusions).

PR #1093, opened 2018-01-24
[WIP] Refactor getting started guide

### LINKED ISSUE #1092: Add code sample for getting started doc
The [getting started guide](https://github.com/derekparker/delve/blob/master/Documentation/cli/getting_started.md) references a dummy `github.com/me/foo/cmd/foo` code. It would be useful to add a concrete (albeit dummy) code, that would have it's own project and `cmd/program` structure, to expand the guide and be able to have reproducible examples.

This issue really aims at using that same sample code for the internal documentation (#1074), and "fit" those two guides together (the getting started guide as a basic introduction and the internal documentation as the advanced part), providing some continuity for the reader.

My question is where should this code be added?

I'd rather not have it embedded in the markdown forcing the reader to copy/paste it, but somewhere it would be accessible through a command (e.g., `go get`) and thus provide the reader with the commands to follow the documentation (I'd rather copy/paste single commands rather than big pieces of code). I think it would be nice to have some sort of `example` project inside `github.com/derekparker` (or maybe just a directory inside this same repository if the previous solution is too complicated).

schomatis:
This PR aims at refactoring the getting started guide to use it as a stepping stone for the internal documentation (#1074).

Some useful steps would be (WIP):

- [x] Bump the guide to the front as to make it very visible to the new user.
- [x] Add its own directory inside Documentation.
- [ ] Add concrete code samples (#1092).
- [ ] Add more commands (e.g. `print`, `goroutines`, `next`, `step`, `stepout`). What would be the minimal useful commands to get started?
- [ ] Describe the used commands more in depth (maybe linking in the future to the internal documentation for more available information about them).
- [ ] Expand the concept of the getting started beyond the CLI, meaning, the first time I'm using Delve, I'm just reaching for the Installation instructions and the CLI API? Or would it be better to have it in the Usage part?
- [ ] Make a clear distinction between the commands that Delve accepts, e.g., `delve debug`, and the commands during the interactive debug session (CLI), e.g., `next`.
- [ ] Is the path `github.com/derekparker/delve/Documentation/getting_started/examples/cmd/hello_world` too long? Running delve inside `getting_started/examples/cmd/hello_world` even with the full qualified package will result inthe print statements from the CLI to just print relative to the CWD, e.g., `./main.go:8`.
- [ ] Why is the warning `Warning: debugging optimized function` showing for the dummy example of `hello_world`?
- [ ] Specify why it's necessary to use `break main.main` instead of just `break main`.
- Clarify that the `dlv debug` command is building a `debug` binary in the CWD and that could potentially clash with other files with the same name with the error `/debug" already exists and is a directory`
- [ ] Document (maybe not here but in the same `dlv debug`) that there is a separation between delve's arguments and the program's (being debugged) through a `--` split.

Closes #1092.


derekparker:
Must be rebased due to: https://github.com/derekparker/delve/pull/1096.

schomatis:
> Must be rebased due to: #1096.

Thanks for the heads-up.

Keep in mind that this is a work in progress that will take me some time so it will become repeatedly outdated. I don't mind rebasing it all at the end, although it's a healthy practice to rebase every so often to not fall that far behind.

derekparker:
Closing for now, feel free to reopen / resubmit when more work can be done.

PR #1094, opened 2018-01-24
[WIP] Internal documentation

schomatis:
An attempt to contribute to Delve internal's documentation (#1074). This documentation is anchored to version [v1.0.0](https://github.com/derekparker/delve/releases/tag/v1.0.0).

The scope now is including the build process, the creation of the server and the debugger, the initialization of the program, the CLI and the interaction between RPC server and client. None of it is described in an exhaustive form.

As much as possible the documentation should reside with the code itself in the form of comments rather than in a separate documents. This PR modifies existing go source files, but only adding new (separate) lines with comments and never modifying functionality. 

Right now the focus is on providing the new user the knowledge to navigate Delve's source code from a CLI command of interest into the functionality that supports it. The downside of this approach is that the documentation is being presented in the form of a narration, e.g., "this command is typed, which calls this function, which send that RPC to that other server, which forwards it to the debugger". That results in a rather cumbersome text, oriented to much for the new user and not so much for a more advanced one which wants to know the architecture of a specific component inside Delve. Furthermore, the description of many components get rather entangled with each other (because of their natural interaction) and may not result in a clear distinction from each other. The description of the process (or rather the flow of information) is being prioritized over the constituent components of the system. In time part of the descriptions can be refactored out into its own documents.


schomatis:
@derekparker Regarding the RPC client and the calls it sends from the terminal, there is a distinction in how some of the commands from the CLI are processed: some are directed translated to the corresponding function, e.g., [`CreateBreakpoint`](https://github.com/derekparker/delve/blob/v1.0.0/service/rpc2/client.go#L178), but others are abstracted trough the `"Command"` string, e.g., [`Next`](https://github.com/derekparker/delve/blob/v1.0.0/service/rpc2/client.go#L116), handled by its corresponding [`Command`](https://github.com/derekparker/delve/blob/v1.0.0/service/rpc2/server.go#L109) function.

Which is the criteria for this distinction? According to the comments in the [server](https://github.com/derekparker/delve/blob/v1.0.0/service/rpc2/server.go#L108), the [debugger](https://github.com/derekparker/delve/blob/v1.0.0/service/debugger/debugger.go#L479) and the [DebuggerCommand](https://github.com/derekparker/delve/blob/v1.0.0/service/api/types.go#L250)  structure these are commands that have a more direct influence over the debugger, but those comments are somewhat heterogeneous,  could you provide me with a more unified explanation of why these commands have another layer of abstraction through the `"Command"` prefix?

derekparker:
@schomatis essentially it allows for a single place to detect process death and update state of the process back to the client.

schomatis:
@derekparker I'm drafting a simple overview of Delve's architecture focusing on the client/server interaction (I'm not expanding into the debugger itself), the most important components I could identify are:

```
                                   Process

Terminal                           Debugger

RPC client  <------------------->  RPC server
               (JSON over TCP)
```

Is there any other relevant component to add (to this higher level model)?

derekparker:
@schomatis that seems like a good layout for a high level architectural overview. I think one thing that would be worth mentioning is that all of the actual logic is in the `debugger` package, whose purpose is essentially to abstract all the low-level details of the `proc` package. But the benefit of the `debugger` package holding all the logic is that other APIs can be created, for instance REST, GRPC, and so on. The client / server implementations are dumb on purpose.

aarzilli:
Personally I like to think of debuggers has having three layers:

1. UI Layer
2. Symbolic Layer
3. Target Layer

(these names are made up, I don't think official names exist).

This is what delve has but also matches broadly the architecture of gdb and also what I've seen in the (limited) literature I've read on the topic. The target layer is responsible for controlling the target process but doesn't know anything about the source code. The symbolic layer bridges the UI with the target layer, translating variable names to addresses and interpreting bytes into high level types.

In delve the target layer is in pkg/proc/native, pkg/proc/gdbserial and pkg/proc/core (we have 5 implementations at the moment) and it's represented by the pkg/proc.Process and pkg/proc.Thread interfaces. The symbolic layer is mostly in pkg/proc, with some pieces in service/debugger and a lot of supporting code in pkg/dwarf. 90% of the symbolic layer is just about decoding and applying the stuff that's in DWARF, but for example the next/step logic also belongs in this layer and isn't derived from DWARF. For delve in particular we also know how to interpret a bunch of data structures that are specific to the go runtime.

The UI layer is in pkg/terminal. Delve has one extra layer between UI and symbolic, the RPC layer, that lets UI and the rest of the debugger to run in different processes.

schomatis:
@aarzilli Thank you for the detailed description. It's a bit advanced to add it to the (basic) document that I'm preparing now, but it helps me personally (and anyone reading this) to get a better idea of what Delve (and a debugger in general) does and how it does it. I'll try to add the path references to the source code organization document.

schomatis:
@derekparker Regarding the RPC client/server and its different versions, what is the main difference between v1 and v2? Could you please check that the following comments are correct (and please complete/correct where appropriate).

The terminal uses the RPCv2 client (RPC client v1 is deprecated, there's a `rpc1/readme.txtr` that is not completely clear on that) and the server by default has RPCv1 running. Both server version methods are stored in `methodMaps` but only one is accessed per run (depending on the `config.APIVersion` set in `dlv` arguments).

Common methods to both versions RPC servers are in `rpccommon/server.go` (there is no equivalent file for the RPC client, but its interface is completely defined in the `service/client.go`).

Both versions of the RPC server interact with the same debugger interface.

derekparker:
@schomatis sorry for the very long delay in responding to you on this. For more info on the differences see this comment: https://github.com/derekparker/delve/pull/460#issuecomment-212784829. 

Part of the actual response is misformatted at a quote, so don't ignore that part.

derekparker:
@schomatis any more movement on this? Do you need anything from myself or @aarzilli ?

schomatis:
No, you answered all of my questions, thanks for asking, I need to move on this, sorry for the delay.

derekparker:
Closing for now, feel free to reopen / resubmit when more work can be done.

hitzhangjie:
One year ago, I tried to study the delve internals and learn delve from scratch by reviewing every commit from the beginning. As @aarzilli mentioned, delve has several layers.
Dwarf is complex, I tried to introduce dwarf standard, then try to describe the delve implementation, but ... there's no enough time ... I abort, sorry about this, hope I can, or we can help to finish this.

https://github.com/hitzhangjie/golang-debugger/tree/master/doc
filename: develop-golang-debugger.docx

I still feel Dwarf is complex, understanding .line section takes me too much time.

PR #1095, opened 2018-01-25
Update usage documentation

schomatis:
Particularly to make visible the fact that `dlv debug` generates an [output binary](https://github.com/derekparker/delve/pull/1095/files#diff-4f9e418d20cb47783cea629aa425160cR22) named `debug` by default.

Questions:

1. Is the official way to update the usage documentation
to run `go run ./scripts/gen-usage-docs.go`? I'm asking to reflect this in the documentation contribution guidelines (and maybe also a clarification in the Usage `README` to avoid [manual updates](https://github.com/derekparker/delve/pull/848)).

2. Is there a way to do this update manually with every commit? (as come kind of CI)


derekparker:
> Is the official way to update the usage documentation
to run go run ./scripts/gen-usage-docs.go?

Yes.

> Is there a way to do this update manually with every commit? (as come kind of CI)

You mean non-manually? I think what _should_ be added is a check that runs via CI that ensures the documentation is up to date (e.g. no diff when regenerating it). I could of sworn that had already been done but it seems not, and it's low hanging fruit for sure.

derekparker:
Thanks for the patch! For future reference, please structure your commit messages as outlined here: https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message. E.g. for these documentation commits, please prefix with `Documentation: ...`.

schomatis:
> Thanks for the patch! For future reference, please structure your commit messages as outlined here: https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message. E.g. for these documentation commits, please prefix with Documentation: ....

Apologies, I'll take that into account.

> You mean non-manually? I think what should be added is a check that runs via CI that ensures the documentation is up to date (e.g. no diff when regenerating it). I could of sworn that had already been done but it seems not, and it's low hanging fruit for sure.

Yes, I've meant to say automatically. Great, I'll research how to integrate it into the CI.

PR #1098, opened 2018-01-26
terminal: do not crash when process exits during next

### LINKED ISSUE #1090: 480fc02d507c96f1ff939254d687e946a2642290 causes dlv to crash
Version: the head as of 2017-01-22 (bec6a65b15ed9bdc1cdbeb1740a4c82a7ee8f573)
Platform: Ubuntu 16.04
Go: 1.9.1

Repro: Debug any program, say

package main
import (
	"fmt"
)
func main() {
	fmt.Println("Main")
}

then set a breakpoint at "fmt.Println". Once you hit the breakpoint, continue issuing "n" command until the program terminates. After exit(0) call, dlv crashes with the following stacktrace. This problem happens at or after commit 480fc02d507c96f1ff939254d687e946a2642290, but not before that. @dlsniper 

(dlv) 
No current thread available
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x72c0db]

goroutine 1 [running]:
github.com/derekparker/delve/pkg/terminal.continueUntilCompleteNext(0xc4204562a0, 0xc42046c300, 0xa07b52, 0x4, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/pkg/terminal/command.go:700 +0x20b
github.com/derekparker/delve/pkg/terminal.next(0xc4204562a0, 0x0, 0xffffffffffffffff, 0x0, 0x0, 0x0, 0x0, 0x2, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/pkg/terminal/command.go:774 +0x116
github.com/derekparker/delve/pkg/terminal.(*Commands).CallWithContext(0xc420359aa0, 0x0, 0x0, 0xc4204562a0, 0x0, 0xffffffffffffffff, 0x0, 0x0, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/pkg/terminal/command.go:339 +0x12a
github.com/derekparker/delve/pkg/terminal.(*Commands).Call(0xc420359aa0, 0x0, 0x0, 0xc4204562a0, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/pkg/terminal/command.go:344 +0x71
github.com/derekparker/delve/pkg/terminal.(*Term).Run(0xc4204562a0, 0x0, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/pkg/terminal/terminal.go:132 +0x33c
github.com/derekparker/delve/cmd/dlv/cmds.execute(0x0, 0xc420160010, 0x1, 0x1, 0xc420063800, 0x0, 0x0, 0x1, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:537 +0x8e3
github.com/derekparker/delve/cmd/dlv/cmds.debugCmd.func1(0xc420154000, 0xd40108, 0x0, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:300 +0x405
github.com/derekparker/delve/cmd/dlv/cmds.debugCmd(0xc420154000, 0xd40108, 0x0, 0x0)
	/home/ysaito/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:301 +0x49
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).execute(0xc420154000, 0xd40108, 0x0, 0x0, 0xc420154000, 0xd40108)
	/home/ysaito/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:647 +0x237
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc4200d98c0, 0xc420154000, 0xc4200d9d40, 0xc4200d9b00)
	/home/ysaito/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:733 +0x2fe
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).Execute(0xc4200d98c0, 0xc4200d98c0, 0xa07fb8)
	/home/ysaito/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:692 +0x2b
main.main()
	/home/ysaito/go/src/github.com/derekparker/delve/cmd/dlv/main.go:18 +0x91

Debugger exited abnormally with code 2

aarzilli:
```
terminal: do not crash when process exits during next

Fixes #1090

```


dlsniper:
Thank you for this @aarzilli, you've beat me to it :D

pkg/terminal/command.go
```diff
@@ -783,11 +783,18 @@ func scopePrefixSwitch(t *Term, ctx callContext) error {
 	return nil
 }
 
+func exitedToError(state *api.DebuggerState, err error) (*api.DebuggerState, error) {
```
derekparker: Thinking through this, we already have a specific error type for this in `pkg/proc`, does that get swallowed before making it out to the client? Can we propagate that error through to the client instead of detecting based off of state?

aarzilli: After 480fc02d507c96f1ff939254d687e946a2642290 in `service/debugger` almost all calls to Command will convert a ProcessExitedError into a DebuggerState with Exited set to true. Before that we only did that for `continue`.

derekparker: Ah, yes, you're right.

@dlsniper from an API consumer point of view, is it more convenient if delve returns an error in this case, or is it fine on your end to detect the `Exited` state for any calls that manipulate process execution?

dlsniper: From a consumer point of view, it's simpler to handle the exit logic of the process regardless of the way it terminated, either by normal means, Exit(0), via an unexpected Exit(2), or via a panic(). I'm sorry the burden has to be on the server in this case.

aarzilli: We've always returned `Exited=true` instead of an error for `continue`, we can't change that because it would be a major backwards compatibility violation. 
`next`, `step` and `stepout` used to return an error instead and 480fc02 changed them to match the behavior of `continue`. Maybe we shouldn't have done that, but we did. At this point, I would prefer to keep them consistent with `continue` than to change them back.

dlsniper: Sorry, I wasn't very clear. I definitely think that all operations should behave the same, regardless of operation performed, for the exit behavior, which is what that PR attempted to do. Unfortunately I haven't tested it enough and introduced this bug.

aarzilli: > Sorry, I wasn't very clear

You were clear, I was agreeing with you.

derekparker: > At this point, I would prefer to keep them consistent with continue than to change them back.

Agreed.

PR #1108, opened 2018-01-29
proc/native: fix race condition between Halt and process death (linux)

### LINKED ISSUE #1101: Problems when following a non-main goroutine with 'next' command: Delve doesn't detect the end of the program
Playing with a basic [goroutine example](https://gobyexample.com/goroutines) to add it to the getting started guide I came across a series of panics and hangs (that I believe are all related) when switching to a non-`main` goroutine and following it with `next`, because (I'm guessing) Delve at some point misses the end of the program and pointers like `(*Thread)` in the server or `DebuggerState.CurrentThread` in the client are `nil`. I'll be looking deeper into this issue but I document it here to get some feedback about it.

The problem, as it deals with threads and goroutines, is not completely deterministic but comes up often enough (at least in my machine) that after a couple of tries it will manifest itself (in different ways). In it most basic form the test code is an empty main function that spawns a goroutine (also empty) and waits for it (the `sync.Wait` was added to the cited example):

```go
package main

import (
    "sync"
)

var wg sync.WaitGroup

func f(from string) {
    defer wg.Done()
    return
}

func main() {
    wg.Add(1)
    go f("goroutine")
    wg.Wait()
}
```

Building the binary

```
$ dlv debug --output debug_binary
```

and running the following CLI commands that break inside the function executed by the goroutine (with an implicit goroutine switch)

```
r
b main.main
c
b f
c
n # repeated three times to exit f() goroutine
n
n
```

leaves it inside the created goroutine

```
     8:	
=>   9:	func f(from string) {
    10:	    defer wg.Done()
    11:	    return
    12:	}
    13:	
    14:	func main() {
(dlv) goroutines
[4 goroutines]
  Goroutine 1 - User: /usr/local/go/src/runtime/sema.go:56 sync.runtime_Semacquire (0x4350b9)
  Goroutine 2 - User: /usr/local/go/src/runtime/proc.go:288 runtime.gopark (0x427b2d)
  Goroutine 17 - User: /usr/local/go/src/runtime/proc.go:288 runtime.gopark (0x427b2d)
* Goroutine 18 - User: ./main.go:9 main.f (0x455f1f) (thread 21588)
(dlv) 
```

From this point, following the execution with the `next` command sometimes ends correctly with the process termination, but in other cases it results in panics or hangs. Alternatively, using the command `continue` always ends correctly without issues.

Problem 1: panic in the terminal client.

```
No current thread available
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x7be5fb]

goroutine 1 [running]:
github.com/derekparker/delve/pkg/terminal.continueUntilCompleteNext(0xc420061e60, 0xc4204362a0, 0xaecfb2, 0x4, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/pkg/terminal/command.go:749 +0x20b
github.com/derekparker/delve/pkg/terminal.next(0xc420061e60, 0x0, 0xffffffffffffffff, 0x0, 0x0, 0x0, 0x0, 0x2, 0x0)
	/home/user/go/src/github.com/derekparker/delve/pkg/terminal/command.go:823 +0x116
github.com/derekparker/delve/pkg/terminal.(*Commands).CallWithContext(0xc420171d40, 0xc420293bec, 0x0, 0xc420061e60, 0x0, 0xffffffffffffffff, 0x0, 0x0, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/pkg/terminal/command.go:348 +0x12a
github.com/derekparker/delve/pkg/terminal.(*Commands).Call(0xc420171d40, 0xc420293bec, 0x0, 0xc420061e60, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/pkg/terminal/command.go:353 +0x71
github.com/derekparker/delve/pkg/terminal.(*Term).Run(0xc420061e60, 0x0, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/pkg/terminal/terminal.go:132 +0x33c
github.com/derekparker/delve/cmd/dlv/cmds.execute(0x0, 0xc42014e3b0, 0x1, 0x1, 0xc420156240, 0x0, 0x0, 0x1, 0x0)
	/home/user/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:537 +0x8e3
github.com/derekparker/delve/cmd/dlv/cmds.debugCmd.func1(0xc420172900, 0xc42014c480, 0x0, 0x2, 0x0)
	/home/user/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:300 +0x405
github.com/derekparker/delve/cmd/dlv/cmds.debugCmd(0xc420172900, 0xc42014c480, 0x0, 0x2)
	/home/user/go/src/github.com/derekparker/delve/cmd/dlv/cmds/commands.go:301 +0x49
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).execute(0xc420172900, 0xc42014c300, 0x2, 0x2, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:647 +0xa4e
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc420172240, 0xc420172900, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:733 +0x68c
github.com/derekparker/delve/vendor/github.com/spf13/cobra.(*Command).Execute(0xc420172240, 0x0, 0x0)
	/home/user/go/src/github.com/derekparker/delve/vendor/github.com/spf13/cobra/command.go:692 +0x4f
main.main()
	/home/user/go/src/github.com/derekparker/delve/cmd/dlv/main.go:18 +0x91

```

Problem 2: panic in the debugger server when restarting what appeared to be a correct end in the execution

```
(dlv) r
Command failed: Internal debugger error: runtime error: invalid memory address or nil pointer dereference
runtime.call32 (0x45ceba)
	/usr/local/go/src/runtime/asm_amd64.s:509
runtime.gopanic (0x42f541)
	/usr/local/go/src/runtime/panic.go:491
runtime.panicmem (0x42e41d)
	/usr/local/go/src/runtime/panic.go:63
runtime.sigpanic (0x44649b)
	/usr/local/go/src/runtime/signal_unix.go:367
github.com/derekparker/delve/pkg/proc/native.(*Thread).stopped (0x797392)
	/home/user/go/src/github.com/derekparker/delve/pkg/proc/native/threads_linux.go:34
github.com/derekparker/delve/pkg/proc/native.(*Thread).Stopped (0x796a8a)
	/home/user/go/src/github.com/derekparker/delve/pkg/proc/native/threads.go:120
github.com/derekparker/delve/pkg/proc/native.(*Process).Kill (0x7916fd)
	/home/user/go/src/github.com/derekparker/delve/pkg/proc/native/proc_linux.go:106
github.com/derekparker/delve/pkg/proc/native.(*Process).Detach (0x78ee2e)
	/home/user/go/src/github.com/derekparker/delve/pkg/proc/native/proc.go:82
github.com/derekparker/delve/service/debugger.(*Debugger).detach (0x79a61b)
	/home/user/go/src/github.com/derekparker/delve/service/debugger/debugger.go:177
github.com/derekparker/delve/service/debugger.(*Debugger).Restart (0x79a78d)
	/home/user/go/src/github.com/derekparker/delve/service/debugger/debugger.go:203
github.com/derekparker/delve/service/rpc2.(*RPCServer).Restart (0x977f87)
	/home/user/go/src/github.com/derekparker/delve/service/rpc2/server.go:83
runtime.call128 (0x45cfd1)
	/usr/local/go/src/runtime/asm_amd64.s:511
reflect.Value.call (0x4e21b5)
	/usr/local/go/src/reflect/value.go:434
reflect.Value.Call (0x4e1366)
	/usr/local/go/src/reflect/value.go:302
github.com/derekparker/delve/service/rpccommon.(*ServerImpl).serveJSONCodec.func1 (0x98095d)
	/home/user/go/src/github.com/derekparker/delve/service/rpccommon/server.go:299
github.com/derekparker/delve/service/rpccommon.(*ServerImpl).serveJSONCodec (0x97f52d)
	/home/user/go/src/github.com/derekparker/delve/service/rpccommon/server.go:301
runtime.goexit (0x45f6e0)
	/usr/local/go/src/runtime/asm_amd64.s:2337
```

Problem 3: Delve hangs with a zombie process.

```
$ command ps aux -T | grep debug_bin
user     15052 15052  0.0  0.0      0     0 pts/18   Zl   08:58   0:00 [debug_binary] <defunct>
user     15052 15053  0.0  0.0      0     0 pts/18   Zl   08:58   0:00 [debug_binary] <defunct>
user     15052 15054  0.0  0.0      0     0 pts/18   Zl   08:58   0:00 [debug_binary] <defunct>
```

Removing the `Wait` call in `main`, although it makes it less probable that Delve will hit the breakpoint inside the goroutine (because `main` ends before that), when it does, it always (or very likely) causes one of the above described problems because (I'm guessing again) Delve is not detecting the end of the `main` goroutine while stepping through the (now dead) non-`main` goroutine. (To increase the probability of hitting the breakpoint the removed `Println` statement from the cited example should be added again to the function executed by the goroutine.)

Development environment:

```
go version go1.9.2 linux/amd64
Delve Debugger
Version: 1.0.0-rc.2
Build: 74d330a013a5412649aaa069d932b93889c2dbd2
```

aarzilli:
```
proc/native: fix race condition between Halt and process death (linux)

If a breakpoint is hit close to process death on a thread that isn't
the group leader the process could die while we are trying to stop it.

This can be easily reproduced by having the goroutine that's executing
main.main (which will almost always run on the thread group leader)
wait for a second goroutine before exiting, then setting a breakpoint
on the second goroutine and stepping through it (see TestIssue1101 in
proc_test.go).

When stepping over the return instruction of main.f the deferred
wg.Done() call will be executed which will cause the main goroutine to
resume and proceed to exit. Both the temporary breakpoint on wg.Done
and the temporary breakpoint on the return address of main.f will be in
close proximity to main.main calling os.Exit() and causing the death of
the thread group leader.

Under these circumstances the call to native.(*Thread).waitFast in
native.(*Thread).halt can hang forever due to a bug similar to
https://sourceware.org/bugzilla/show_bug.cgi?id=12702 (see comment in
native.(*Thread).wait for an explanation).

Replacing waitFast with a normal wait work in most circumstances,
however, besides the performance hit, it looks like in this
circumstances trapWait sometimes receives a spurious SIGTRAP on the
dying group leader which would cause the subsequent call to wait in
halt to accidentally reap the process without noting that it did exit.

Instead this patch removes the call to wait from halt and instead calls
trapWait in a loop in setCurrentBreakpoints until all threads are set
to running=false. This is also a better fix than the workaround to
ESRCH error while setting current breakpoints implemented in 94b50d.

Fixes #1101

```

### DIFF
--- a/pkg/proc/native/proc_linux.go
+++ b/pkg/proc/native/proc_linux.go
@@ -230,9 +230,19 @@ func (dbp *Process) trapWait(pid int) (*Thread, error) {
 			if err != nil {
 				if err == sys.ESRCH {
 					// thread died while we were adding it
+					delete(dbp.threads, int(cloned))
 					continue
 				}
 				return nil, err
+			}
+			dbp.haltMu.Lock()
+			halt := dbp.halt
+			dbp.haltMu.Unlock()
+			if halt {
+				dbp.halt = false
+				th.running = false
+				dbp.threads[int(wpid)].running = false
+				return nil, nil
 			}
 			if err = th.Continue(); err != nil {
 				if err == sys.ESRCH {
@@ -256,7 +266,7 @@ func (dbp *Process) trapWait(pid int) (*Thread, error) {
 		dbp.haltMu.Lock()
 		halt := dbp.halt
 		dbp.haltMu.Unlock()
-		if status.StopSignal() == sys.SIGTRAP && halt {
+		if halt && (status.StopSignal() == sys.SIGTRAP || status.StopSignal() == sys.SIGSTOP) {
 			th.running = false
 			dbp.halt = false
 			return th, nil
@@ -368,16 +378,28 @@ func (dbp *Process) wait(pid, options int) (int, *sys.WaitStatus, error) {
 }
 
 func (dbp *Process) setCurrentBreakpoints(trapthread *Thread) error {
+	// wait for all threads to stop
+	for {
+		allstopped := true
+		for _, th := range dbp.threads {
+			if th.running {
+				allstopped = false
+				break
+			}
+		}
+		if allstopped {
+			break
+		}
+		dbp.halt = true
+		_, err := dbp.trapWait(-1)
+		if err != nil {
+			return err
+		}
+	}
 	for _, th := range dbp.threads {
 		if th.CurrentBreakpoint.Breakpoint == nil {
 			if err := th.SetCurrentBreakpoint(); err != nil {
-				if err == sys.ESRCH {
-					// This thread quit between the point where we received the breakpoint and
-					// the stop signal.
-					delete(dbp.threads, th.ID)
-				} else {
-					return err
-				}
+				return err
 			}
 		}
 	}
--- a/pkg/proc/native/threads.go
+++ b/pkg/proc/native/threads.go
@@ -120,22 +120,6 @@ func (thread *Thread) Stopped() bool {
 	return thread.stopped()
 }
 
-// Halt stops this thread from executing. Actual
-// implementation is OS dependant. Look in OS
-// thread file.
-func (thread *Thread) Halt() (err error) {
-	defer func() {
-		if err == nil {
-			thread.running = false
-		}
-	}()
-	if thread.Stopped() {
-		return
-	}
-	err = thread.halt()
-	return
-}
-
 // SetCurrentBreakpoint sets the current breakpoint that this
 // thread is stopped at as CurrentBreakpoint on the thread struct.
 func (thread *Thread) SetCurrentBreakpoint() error {
--- a/_fixtures/issue1101.go
+++ b/_fixtures/issue1101.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"os"
+	"sync"
+)
+
+var wg sync.WaitGroup
+
+func f(from string) {
+	defer wg.Done()
+	return
+}
+
+func main() {
+	wg.Add(1)
+	go f("goroutine")
+	wg.Wait()
+	os.Exit(2)
+}
--- a/pkg/proc/native/threads_darwin.go
+++ b/pkg/proc/native/threads_darwin.go
@@ -27,6 +27,20 @@ type OSSpecificDetails struct {
 // be continued.
 var ErrContinueThread = fmt.Errorf("could not continue thread")
 
+// Halt stops this thread from executing.
+func (thread *Thread) Halt() (err error) {
+	defer func() {
+		if err == nil {
+			thread.running = false
+		}
+	}()
+	if thread.Stopped() {
+		return
+	}
+	err = thread.halt()
+	return
+}
+
 func (t *Thread) halt() (err error) {
 	kret := C.thread_suspend(t.os.threadAct)
 	if kret != C.KERN_SUCCESS {
--- a/pkg/proc/native/threads_linux.go
+++ b/pkg/proc/native/threads_linux.go
@@ -16,15 +16,19 @@ type OSSpecificDetails struct {
 	registers sys.PtraceRegs
 }
 
+// Halt stops this thread from executing.
+func (thread *Thread) Halt() (err error) {
+	if thread.Stopped() {
+		return
+	}
+	err = thread.halt()
+	return
+}
+
 func (t *Thread) halt() (err error) {
 	err = sys.Tgkill(t.dbp.pid, t.ID, sys.SIGSTOP)
 	if err != nil {
 		err = fmt.Errorf("halt err %s on thread %d", err, t.ID)
-		return
-	}
-	_, _, err = t.dbp.waitFast(t.ID)
-	if err != nil {
-		err = fmt.Errorf("wait err %s on thread %d", err, t.ID)
 		return
 	}
 	return
--- a/pkg/proc/native/threads_windows.go
+++ b/pkg/proc/native/threads_windows.go
@@ -18,6 +18,20 @@ type OSSpecificDetails struct {
 	hThread syscall.Handle
 }
 
+// Halt stops this thread from executing.
+func (thread *Thread) Halt() (err error) {
+	defer func() {
+		if err == nil {
+			thread.running = false
+		}
+	}()
+	if thread.Stopped() {
+		return
+	}
+	err = thread.halt()
+	return
+}
+
 func (t *Thread) halt() (err error) {
 	// Ignore the request to halt. On Windows, all threads are halted
 	// on return from WaitForDebugEvent.
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3400,3 +3400,45 @@ func TestDeclLine(t *testing.T) {
 		}
 	})
 }
+
+func TestIssue1101(t *testing.T) {
+	// If a breakpoint is hit close to process death on a thread that isn't the
+	// group leader the process could die while we are trying to stop it.
+	//
+	// This can be easily reproduced by having the goroutine that's executing
+	// main.main (which will almost always run on the thread group leader) wait
+	// for a second goroutine before exiting, then setting a breakpoint on the
+	// second goroutine and stepping through it (see TestIssue1101 in
+	// proc_test.go).
+	//
+	// When stepping over the return instruction of main.f the deferred
+	// wg.Done() call will be executed which will cause the main goroutine to
+	// resume and proceed to exit. Both the temporary breakpoint on wg.Done and
+	// the temporary breakpoint on the return address of main.f will be in
+	// close proximity to main.main calling os.Exit() and causing the death of
+	// the thread group leader.
+
+	withTestProcess("issue1101", t, func(p proc.Process, fixture protest.Fixture) {
+		_, err := setFunctionBreakpoint(p, "main.f")
+		assertNoError(err, t, "setFunctionBreakpoint()")
+		assertNoError(proc.Continue(p), t, "Continue()")
+		assertNoError(proc.Next(p), t, "Next() 1")
+		assertNoError(proc.Next(p), t, "Next() 2")
+		lastCmd := "Next() 3"
+		exitErr := proc.Next(p)
+		if exitErr == nil {
+			lastCmd = "final Continue()"
+			exitErr = proc.Continue(p)
+		}
+		if pexit, exited := exitErr.(proc.ProcessExitedError); exited {
+			if pexit.Status != 2 && testBackend != "lldb" {
+				// looks like there's a bug with debugserver on macOS that sometimes
+				// will report exit status 0 instead of the proper exit status.
+				t.Fatalf("process exited status %d (expected 2)", pexit.Status)
+			}
+		} else {
+			assertNoError(exitErr, t, lastCmd)
+			t.Fatalf("process did not exit after %s", lastCmd)
+		}
+	})
+}


schomatis:
I've re-tested #1101 and it's now working as expected.

pkg/proc/native/proc_linux.go
```diff
@@ -230,10 +231,20 @@ func (dbp *Process) trapWait(pid int) (*Thread, error) {
 			if err != nil {
 				if err == sys.ESRCH {
 					// thread died while we were adding it
+					delete(dbp.threads, int(cloned))
 					continue
 				}
 				return nil, err
 			}
+			dbp.haltMu.Lock()
```
derekparker: We discussed briefly on IRC about this, but to bring it to this review:

I think we should be more careful of how we use this lock. Also, I'm not sure the value of `dbp.halt` anymore. It used to make sense because of the various implementations of `Halt` (which has changed a bit over time) because we needed to have state to signify that the inferior receiving `SIGSTOP` was actually sent via the debugger in attempt to stop the thread.

As a result of various changes I think we have 2 implementations of essentially the same thing, and their usage internally tends to be confusing and misleading. From an API / client perspective the only thing that gets called is `RequestManualStop`. The API command `halt` actually just calls `RequestManualStop`. We do not need to have two separate methods with 2 separate pieces of state (`halt` & `manualStopRequested`). It seems `(*Process) Halt()` is only used internally throughout proc and in tests.

Additionally, the comment for `RequestManualStop` is wrong, and the implementation, at least on Linux, could be improved. The comment mentions that we send `SIGSTOP` to all threads, when in actuality we only send `SIGTRAP` to the group leader. This may not be the best approach, as we essentially detect that in `trapWait` and then force all other threads to stop (by calling `thread.halt()`, which sends `SIGSTOP` to the thread and then immediately waits, so even there the `halt` flag is unnecessary. This is also inconsistent because I feel all waits should be filtered through the same code path, because what if the result of that wait is not the signal that we sent to the thread, but other important process state (we do specifically wait on the thread ID, but if that's the group leader, we could get other info).

To expand on the above paragraph, I feel we should combine the two methods and have on definitive way of stopping a running process. Also, on Linux the implementation could be improved (on OSX it's more difficult because of how ptrace / mach works). Currently we are sending a signal to the group leader putting it into a `signal-delivery-stop`, throwing away that signal (never delivering to the inferior) and then stopping all other threads within `trapWait`. What we could actually do is send `SIGSTOP`, and actually deliver it to the inferior which would cause all threads being traced to stop in a `group-stop`. I believe we would have to explicitly continue the process with `SIGCONT` as well on the flip side, but the benefit is all threads stop immediately.

One last thing: digging back through the code during this review I found something I'm really not a fan of and wish to change. I do not like that `ManualStopRequested` has the side effect of clearing `(*Process).manualStopRequested`. Predicate methods should not have side effects.

I'd like to have any discussion around the above comments in this PR, however any resulting action to be taken can happen in follow up PR(s).

pkg/proc/native/threads.go
```diff
@@ -120,22 +120,6 @@ func (thread *Thread) Stopped() bool {
 	return thread.stopped()
 }
 
-// Halt stops this thread from executing. Actual
-// implementation is OS dependant. Look in OS
-// thread file.
-func (thread *Thread) Halt() (err error) {
```
derekparker: Why remove this generic implementation in favor of copying everywhere? Also, if it must be copied, the comments need to be updated.

aarzilli: Because I had to remove the defer on the linux version.

pkg/proc/native/proc_linux.go
aarzilli: > As a result of various changes I think we have 2 implementations of essentially the same thing, and their usage internally tends to be confusing and misleading

Agree with this, I think `proc/native.(*Process).Halt` should be combined with `proc/native.(*Process).setCurrentBreakpoints` and just be some private method of proc/native with an OS specific implementation. The call to Halt in the tests can probably be removed entirely, I don't think we ever get to it with the process running.

> This is also inconsistent because I feel all waits should be filtered through the same code path, because what if the result of that wait is not the signal that we sent to the thread, but other important process state (we do specifically wait on the thread ID, but if that's the group leader, we could get other info).

This patch is doing that by calling trapWait after sending SIGSTOP. There's only one other place where we wait, that I can think of, and is in singlestep but that's probably fine given that everything else is stopped and the thread is in single step mode.

> What we could actually do is send SIGSTOP, and actually deliver it to the inferior which would cause all threads being traced to stop in a group-stop. I believe we would have to explicitly continue the process with SIGCONT as well on the flip side, but the benefit is all threads stop immediately.

I've checked the gdb native backend recently and as far as I can tell they are doing the signal delivery stop thing too, I'm not feeling too confident straying off the beaten path here.

> I do not like that ManualStopRequested has the side effect of clearing (*Process).manualStopRequested. Predicate methods should not have side effects.

Do you think CheckAndClearManualStopRequested would be a better name? Suggestions?

derekparker: > Do you think CheckAndClearManualStopRequested would be a better name? 

I'm fine with that, I don't verbosity for clarity.

> Agree with this, I think proc/native.(*Process).Halt should be combined with proc/native.(*Process).setCurrentBreakpoints and just be some private method of proc/native with an OS specific implementation. The call to Halt in the tests can probably be removed entirely, I don't think we ever get to it with the process running.

I think instead of combining we should just merge to a single implementation and call that where needed.

-------

I think overall we can merge this after the function rename, and then send a follow up PR and address the other concerns outlined about. What do you think @aarzilli?

aarzilli: > I think instead of combining we should just merge to a single implementation and call that where needed.

what do you mean?

derekparker: > what do you mean?

Sorry, I wasn't very clear with that. What I meant was: it seemed you were suggesting moving the implementation of `halt` into `(*Process).setCurrentBreakpoints`, but to me that's only hiding the issue. We should have only one code path that knows how to stop a running process consistently, and simply call that routine where needed. Apologies if I misunderstood you.

pkg/proc/native/proc_linux.go
```diff
@@ -222,6 +222,7 @@ func (dbp *Process) trapWait(pid int) (*Thread, error) {
 			if err != nil {
 				if err == sys.ESRCH {
 					// thread died while we were adding it
+					delete(dbp.threads, int(wpid))
```
derekparker: I don't think we need this `delete` as we haven't added the new thread to the list yet.

pkg/proc/native/proc_linux.go
```diff
@@ -368,16 +379,28 @@ func (dbp *Process) wait(pid, options int) (int, *sys.WaitStatus, error) {
 }
 
 func (dbp *Process) setCurrentBreakpoints(trapthread *Thread) error {
+	// wait for all threads to stop
+	for {
```
derekparker: We should try and ensure every thread is stopped before calling this to prevent it from having to do any more work. Can we move this logic elsewhere that would still be centralized?

aarzilli: it's mirroring what we're doing in proc_darwin and proc_windows. So it's ugly, but at least it's consistent. That's why I was saying that IMHO we should fuse Halt and setCurrentBreakpoints for the native backend. I'd rather do it after 1.0, however.

derekparker: Ah, yeah that makes sense. There is so much room more refactoring and cleanup post-1.0 anyways, especially throughout `pkg/proc`. 

I think I would be happy to merge this, with the caveat that it will be refactored later, if we can not spread out the implementation of `Halt` while we keep it around. All the defer is doing is setting the flag on the thread struct that it is no longer running, why would we need to remove this? That should be an invariant that if we call `Halt` the thread should not be running.

aarzilli: > All the defer is doing is setting the flag on the thread struct that it is no longer running, why would we need to remove this?

I'm using it to keep track of which threads haven't received sigstop yet.

aarzilli:
Merged as f32ce1b21d2a08e21e3018bfcab8d1df6e673f11

Issue #645, opened 2016-09-29
OS X: cannot get thread count


scaba:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.
1. What version of Delve are you using (`dlv version`)?
   Delve Debugger
   Version: 0.11.0-alpha
   Build: fb8388ab2e50ca28ef9a83c73c329f8d0b6234d9
2. What version of Go are you using? (`go version`)?
   go version go1.7.1 darwin/amd64
3. What operating system and processor architecture are you using?
   OS X 10.12.1 Beta (16B2333a) (developer beta 2 Sept 28)
4. What did you do?
   Tried to debug my app from within Visual Studio Code and from the command line with the same results
5. What did you expect to see?
   Debugging goodness
6. What did you see instead?
   From VS Code:
   2016/09/29 06:26:35 server.go:71: Using API v1
   2016/09/29 06:26:35 debugger.go:61: launching process with args: [./debug -race -tags debug]
   could not launch process: could not get thread count
   
   From terminal:
   joefrancia@iMac [6:17:20]: dlv debug
   could not launch process: could not get thread count

I believe it may be related to the upgrade from Sierra beta 1 to beta 2. I was debuggin a'plenty earlier in the afternoon on my MacBook which was still on beta 1, but when I tried on the iMac I got the above failure. The iMac is beta 2. I've since upgraded the MacBook to beta 2 (this was before I realized there may be a problem) and now I get the same error on the MacBook.

ftong2010:
Exactly the same here.Same set-up,same issue.The only difference here is that my golang environment is newly installed after I installed Mac OS 10.12 Beta2

zchee:
This problem from after the Developer beta.(`16.1.0: xnu-3789.21.3~20`)
It does not occur Sierra stable(`16.0.0: xnu-3789.1.32~3`)

https://github.com/derekparker/delve/blob/master/proc/proc_darwin.c#L108-L109
`count` is almost(why?) 0. sometimes got `28672` or etc.
But `kret` is not `KERN_SUCCESS`, In my case, always got `268435459`.
I'll try to debug it.

zchee:
@aarzilli I did a clean install Sierra stable(`10.12.0 - 16.0.0: xnu-3789.1.32~3`) for debug to this problem. That certainly no problem on stable.

BTW, Apple has been released `10.12.1`(today?). I do not yet test. I'll test on the VMWare.
[...]

PR #1117, opened 2018-02-07
Handle race between fork and task_for_pid

mattbauer:
pkg/proc: Retry task_for_pid if handed parent task instead of child

On macOS a call to fork and a subsequent call to task_for_pid will race each other. This is because the macOS kernel assigns a new proc_t structure early but the new task, thread and uthread come much later. The function exec_mach_imgact in the XNU sources contains this logic.

In a system under load or one with delays in fork processing (i.e. various security software), task_for_pid as currently called by Delve often returns the parent task. This can be seen by printing out the task number around line 86. In a normal system we would see three calls:

-> ~/go/bin/dlv --listen=localhost:59115 --headless=true --api-version=2 --backend=native exec ./___main_go -- 
Task: 9731
Task: 9731
Task: 9731
API server listening at: 127.0.0.1:59115

This is the result on a system where the race is lost:

-> ~/go/bin/dlv --listen=localhost:59115 --headless=true --api-version=2 --backend=native exec ./___main_go -- 
Task: 8707
Task: 10499
Task: 10499
could not launch process: could not get thread count

In this latter case, task 8707 is the parent task. The child task of 10499 was desired and hence the error.

This code change checks to make sure the returned task is not that of the parent. If it is, it retries. It's possible other macOS reported Delve issues are the result of this failed race.

Fixes: #645 #714

### DIFF
--- a/pkg/proc/native/proc_darwin.go
+++ b/pkg/proc/native/proc_darwin.go
@@ -84,12 +84,17 @@ func Launch(cmd []string, wd string) (*Process, error) {
 	// trapWait to wait until the child process calls execve.
 
 	for {
-		err = dbp.updateThreadListForTask(C.get_task_for_pid(C.int(dbp.pid)))
-		if err == nil {
-			break
-		}
-		if err != couldNotGetThreadCount && err != couldNotGetThreadList {
-			return nil, err
+		task := C.get_task_for_pid(C.int(dbp.pid))
+		// The task_for_pid call races with the fork call. This can
+		// result in the parent task being returned instead of the child.
+                if task != dbp.os.task {
+			err = dbp.updateThreadListForTask(task)
+			if err == nil {
+				break
+			}
+			if err != couldNotGetThreadCount && err != couldNotGetThreadList {
+				return nil, err
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


dlsniper:
That looks great, thank you!

dlsniper:
Actually typed a premature submit there. For my understanding, this could potentially fix issues like: https://github.com/derekparker/delve/issues/1112 ?

pkg/proc/native/proc_darwin.go
```diff
@@ -84,12 +84,17 @@ func Launch(cmd []string, wd string) (*Process, error) {
...
-		err = dbp.updateThreadListForTask(C.get_task_for_pid(C.int(dbp.pid)))
-		if err == nil {
-			break
-		}
-		if err != couldNotGetThreadCount && err != couldNotGetThreadList {
-			return nil, err
+		task := C.get_task_for_pid(C.int(dbp.pid))
+		// The task_for_pid call races with the fork call. This can
+		// result in the parent task being returned instead of the child.
+                if task != dbp.os.task {
```
derekparker: nit: looks like the formatting may be off on this line, can you run gofmt.

PR #1131, opened 2018-02-22
[WIP] Improve trace subcommand to show returns

derekparker:
This patch aims to improve the `dlv trace` command by printing the
return values. This is still a work in progress at the moment.
Improvements to be made are:

1. Detect true function return. Just like there is a function prologue
there is code at the end of a function after the RET instruction. We do
not want to set the function end breakpoint there as it won't hit when
we expect it to (when the function returns and return args are set).
Possible solutions are to do instruction level comparisons like we do
currently for prologue detection, and in the long term improve the DWARF
information to tell the actual address of the RET instruction via
`epilogue_begin`.

2. Discard target stdout / stderr. This will clean up output and only
show trace output from the debugger, making it all easier to read and
consume for the user.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -304,7 +304,6 @@ func debugCmd(cmd *cobra.Command, args []string) {
 
 func traceCmd(cmd *cobra.Command, args []string) {
 	status := func() int {
-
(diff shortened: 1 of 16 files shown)


derekparker:
Closing this for now as the PR will need to be reworked and updated for other changes, and I'd rather just open another PR when it's ready than let this sit here.

PR #1142, opened 2018-03-03
Documentation: remove homebrew as an install method

aarzilli:
```
Documentation: remove homebrew as an install method

The formula is broken and produces an endless stream of duplicate bug
reports yet nobody steps up to fix it. Using the formula isn't
necessary and hasn't been in almost a year, the maintainers of delve
aren't using it and the original maintainer of the formula vacated.

```

### DIFF
--- a/Documentation/installation/osx/install.md
+++ b/Documentation/installation/osx/install.md
@@ -2,14 +2,6 @@
 
 Please use the following steps to build and install Delve on OSX.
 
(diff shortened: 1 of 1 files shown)


AlekSi:
This way people will install latest development version from `master` branch, not the latest release. And `dlv version` currently says `Version: 1.0.0` despite it not being that release. Are you okay with that?

PR #1144, opened 2018-03-06
travis-ci: Update to go 1.10

aarzilli:
```
travis-ci: Update to go 1.10

```


.travis.yml
```diff
@@ -7,8 +7,8 @@ os:
 
 go:
   - tip
+  - "1.10.x"
```
derekparker: nit: doesn't need quotes.

.travis.yml
```diff
@@ -7,8 +7,8 @@ os:
...
   - 1.9.x
-  - 1.8.x
```
derekparker: Let's keep 1.8.x in the matrix for now for a few months. There are still distros that ship 1.8.x by default (sigh).

PR #1146, opened 2018-03-07
proc: manual stop requests should clear internal breakpoints

### LINKED ISSUE #1145: Delve hangs
Delve Debugger
Version: 1.0.0-rc.2
Build: $Id: 5c7676404d10d52e16be57fea744a7312b831f0c $

go version go1.10rc2 linux/amd64

Ubuntu 16.04

4. What did you do?
I executed my code with delve with the purpose of debugging it

I expected delve to debug the code

6. What did you see instead?
After few `step`s delve entered infinite loop, which does not happen in the program when running without delve. Interrupting it showed this message:

received SIGINT, stopping process (will not forward signal)> runtime.futex() /usr/local/go/src/runtime/sys_linux_amd64.s:527 (PC: 0x462723)
	breakpoint hit during next, continuing...
received SIGINT, stopping process (will not forward signal)> runtime.futex() /usr/local/go/src/runtime/sys_linux_amd64.s:527 (PC: 0x462723)
	breakpoint hit during next, continuing...
received SIGINT, stopping process (will not forward signal)> runtime.futex() /usr/local/go/src/runtime/sys_linux_amd64.s:527 (PC: 0x462723)
	breakpoint hit during next, continuing...

So, basically, it just hangs in the middle of the debug session and it happens randomly.

aarzilli:
```
proc: manual stop requests should clear internal breakpoints

Fixes #1145

```

### DIFF
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -88,8 +88,16 @@ func Continue(dbp Process) error {
 		return &ProcessExitedError{Pid: dbp.Pid()}
 	}
 	dbp.CheckAndClearManualStopRequest()
+	defer func() {
+		// Make sure we clear internal breakpoints if we simultaneously receive a
+		// manual stop request and hit a breakpoint.
+		if dbp.CheckAndClearManualStopRequest() {
+			dbp.ClearInternalBreakpoints()
+		}
+	}()
 	for {
 		if dbp.CheckAndClearManualStopRequest() {
+			dbp.ClearInternalBreakpoints()
 			return nil
 		}
 		trapthread, err := dbp.ContinueOnce()
--- a/_fixtures/issue1145.go
+++ b/_fixtures/issue1145.go
@@ -1,0 +1,13 @@
+package main
+
+import "time"
+
+func f() {
+	for {
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
+func main() {
+	f()
+}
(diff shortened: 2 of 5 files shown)


pkg/proc/proc.go
```diff
@@ -88,8 +88,17 @@ func Continue(dbp Process) error {
 		return &ProcessExitedError{Pid: dbp.Pid()}
 	}
 	dbp.CheckAndClearManualStopRequest()
+	defer func() {
+		// Make sure we clear internal breakpoints if we simultaneously receive a
+		// manual stop request and hit a breakpoint.
+		if dbp.CheckAndClearManualStopRequest() {
+
```
derekparker: nit: formatting.

pkg/proc/proc.go
```diff
@@ -88,8 +88,17 @@ func Continue(dbp Process) error {
 		return &ProcessExitedError{Pid: dbp.Pid()}
 	}
 	dbp.CheckAndClearManualStopRequest()
```
derekparker: I'm trying to remember the rationale behind calling this directly before the loop, is this simply to clear state before we enter the loop? 

Is there any way we can avoid this in favor of simply calling this at the top of the loop and not directly before entering.

aarzilli: Yes, just to clear the flag in case the manual stop request was received while the process was stopped. We could use a flag to detect that we are on the first iteration of the loop, IMHO it's clearer this way.

PR #1147, opened 2018-03-08
proc: fix reading of empty strings in core files

aarzilli:
```
proc: fix reading of empty strings in core files

Every time we read an empty string we accidentally issue a read for 0
bytes at address 0, this is fine for real memory but the core file
reader doesn't like it.

Fixes an issue reported on the mailing list.

```


lafolle:
:+1:

PR #1154, opened 2018-03-15
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -694,7 +694,7 @@ func nameOfStructRuntimeType(_type *Variable, kind, tflag int64) (string, error)
 			case "offsetAnon":
 				// The offsetAnon field of runtime.structfield combines the offset of
 				// the struct field from the base address of the struct with a flag
-				// determining whether the field is anonimous (i.e. an embedded struct).
+				// determining whether the field is anonymous (i.e. an embedded struct).
 				//
 				//  offsetAnon = (offset<<1) | (anonFlag)
 				//
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -789,7 +789,7 @@ func typeForKind(kind int64, bi *BinaryInfo) (*godwarf.StructType, error) {
 	return constructTypeForKind(kind, bi)
 }
 
-// constructTypeForKind synthesizes a *dwarf.StrucType for the specified kind.
+// constructTypeForKind synthesizes a *dwarf.StructType for the specified kind.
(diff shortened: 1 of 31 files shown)


service/api/conversions.go
```diff
@@ -210,7 +210,7 @@ func ConvertFunction(fn *proc.Function) *Function {
 	}
 
 	// fn here used to be a *gosym.Func, the fields Type and GoType below
-	// corresponded to the omonymous field of gosym.Func. Since the contents of
+	// corresponded to the anonymous field of gosym.Func. Since the contents of
```
jsoref: I'm not at all sure about this

pkg/proc/types.go
```diff
@@ -799,7 +799,7 @@ func constructTypeForKind(kind int64, bi *BinaryInfo) (*godwarf.StructType, erro
 	}
 	prtyp := pointerTo(rtyp, bi.Arch)
 
-	uintptrtyp, err := bi.findType("uintptr")
+	uintptrtype, err := bi.findType("uintptr")
```
jsoref: I'm pretty sure this is a reasonable change

pkg/proc/variables.go
```diff
@@ -625,18 +625,18 @@ func (scope *EvalScope) PackageVariables(cfg LoadConfig) ([]*Variable, error) {
 	var vars []*Variable
 	reader := scope.DwarfReader()
 
-	var utypoff dwarf.Offset
+	var utypeoff dwarf.Offset
```
jsoref: I'm pretty sure this is a reasonable change

derekparker:
LGTM but please remove any changes to vendored files.

jsoref:
@derekparker: sorry about that, I had removed the vendored files when I did my spelling work, and then prepared to publish, but then I went back and reviewed the codebase for `it's` -- which isn't actually a spelling thing, but I wasn't paying attention to the file paths....

service/api/conversions.go
aarzilli: it's homonymous

pkg/proc/variables.go
aarzilli: There's a lot of places where we use typ instead of type, I would leave those as they are for consistency.

derekparker:
I'm good with this after @aarzilli suggestions.

jsoref:
@derekparker: dropped (anonymous was split) + rebased

aarzilli:
Thank you.

PR #1156, opened 2018-03-17
proc,vendor: show global variables in disassembly

aarzilli:
```
proc,vendor: show global variables in disassembly

updates vendored version of x86asm, adds a symbol lookup function to
pass to the disassembler.

This will show global symbol names in the disassembly like go tool
objdump does.

```


pkg/proc/bininfo.go
```diff
@@ -149,6 +149,15 @@ type constantValue struct {
 	singleBit bool
 }
 
+// packageVar represents a package-level variable (or a C global variable).
+// If a global variable does not have an addres (for example it's stored in
```
derekparker: s/addres/address/

pkg/proc/disasm.go
```diff
@@ -83,3 +86,31 @@ func disassemble(memrw MemoryReadWriter, regs Registers, breakpoints *Breakpoint
 	}
 	return r, nil
 }
+
+// Looks up symbol (either functions or global variables) at address addr.
+// Used by disassembly formatter.
+func (bi *BinaryInfo) symLookup(addr uint64) (string, uint64) {
+	fn := bi.PCToFunc(addr)
+	if fn != nil && fn.Entry == addr {
+		if fn.Entry == addr {
```
derekparker: This check is redundant.

aarzilli:
Done.

PR #1166, opened 2018-03-31
proc/gdbserial,dlv/cmds: make logGdbWire command line configurable

aarzilli:
```
proc/gdbserial,dlv/cmds: make logGdbWire command line configurable

```

### DIFF
--- a/pkg/logflags/logflags.go
+++ b/pkg/logflags/logflags.go
@@ -1,0 +1,43 @@
+package logflags
+
+import "strings"
+
+var debugger = false
+var gdbWire = false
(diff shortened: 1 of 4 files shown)


pkg/logflags/logflags.go
```diff
@@ -0,0 +1,25 @@
+package logflags
+
+import "strings"
+
+var Debugger = false
```
derekparker: nit: Would prefer not to export these and have package level getter methods and only mutate them during setup.

aarzilli: Like this?

PR #1174, opened 2018-04-08
terminal: make printcontext use SelectedGoroutine

aarzilli:
```
terminal: make printcontext use SelectedGoroutine

printcontext should use SelectedGoroutine instead of trusting that the
goroutine running on current thread matches the SelectedGoroutine.

When the user switches to a parked goroutine CurrentThread and
SelectedGoroutine will diverge.

Almost all calls to printcontext are safe, they happen after a continue
command returns when SelectedGoroutine and CurrentThread always agree,
but the calls in frameCommand and listCommand are wrong.

Additionally we should stop reporting an error when the debugger is
stopped on an unknown PC address.

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -73,10 +73,10 @@ func (c command) match(cmdstr string) bool {
 
 // Commands represents the commands for Delve terminal process.
 type Commands struct {
-	cmds         []command
-	lastCmd      cmdfunc
-	client       service.Client
-	frame int // Current frame as set by frame/up/down commands.
+	cmds    []command
+	lastCmd cmdfunc
+	client  service.Client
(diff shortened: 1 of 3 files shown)


pkg/terminal/command.go
```diff
@@ -1462,13 +1465,30 @@ func printcontext(t *Term, state *api.DebuggerState) error {
...
+				break
+			}
+		}
+		if th == nil {
+			printcontextLocation(state.SelectedGoroutine.CurrentLoc)
+			return nil
+		}
+	}
+
+	if len(th.File) == 0 {
```
derekparker: nit: I know this is carry-over but let's be consistent and check against empty string instead of length.

aarzilli:
Done.

PR #1184, opened 2018-04-15
proc: Refactor proc.(*EvalScope).variablesByTag

### LINKED ISSUE #951: VariableShadowed flag not applied to func arguments

```
> dlv version
Delve Debugger
Version: 1.0.0-rc.1
Build: $Id: b6077a6a35c48a31716aad2ad3fdf77bbdf70abd $
```

```
> go1.9rc2 version
go version go1.9rc2 windows/amd64
```
go1.9rc2 has been built as described in here https://golang.org/dl/#go1.9rc2
(= `go get golang.org/x/build/version/go1.9rc2`)

Windows 10 Pro (10.0.15063 Build 15063)
x64 based architecture

4. What did you do?

Given the following code:
```go
package main

import "fmt"

func main() {
	shadow(42)
}

func shadow(i int) {
	fmt.Println("arg", i)
	for i := 0; i < 5; i++ {
		fmt.Println("shadow", i)
		for i := 20; i < 25; i++ {
			fmt.Println("another shadow", i)
		}
	}
}
```

Now run `dlv debug`
Add breakpoint at the inner most `fmt.Println` (line 14) and continue
```
(dlv) list
> main.shadow() [redacted]/main.go:14 (PC: 0x49660a)
     9: func shadow(i int) {
    10:         fmt.Println("arg", i)
    11:         for i := 0; i < 5; i++ {
    12:                 fmt.Println("shadow", i)
    13:                 for i := 20; i < 25; i++ {
=>  14:                         fmt.Println("another shadow", i)
    15:                 }
    16:         }
    17: }
(dlv) locals
(i) = 0
i = 20
(dlv) args
i = 42
```

Both the `i` argument (from `args`) and the `i` with value `0` are marked as shadowed (surrounded by brackets)

6. What did you see instead?

The `i` from `args` is missing the brackets

Is this a bug or has the user of these APIs to figure it out themselves? Like if a variable with the same name occurs in `locals` and `args` then skip the on from `args`?

-----

If it helps here is the JSON representation of the combined calls to `ListLocalVars` and `ListFunctionArgs`:
```js
[
  // locals:
  {
    "name": "i",
    "addr": 825741434416,
    "onlyAddr": false,
    "type": "int",
    "realType": "int",
    "flags": 2, // <-- as you can see the outer "i" has the flag
    "kind": 2,
    "value": "0",
    "len": 0,
    "cap": 0,
    "children": [],
    "unreadable": ""
  },
  {
    "name": "i",
    "addr": 825741434424,
    "onlyAddr": false,
    "type": "int",
    "realType": "int",
    "flags": 0, // <-- inner "i" variable, no flag ✓
    "kind": 2,
    "value": "20",
    "len": 0,
    "cap": 0,
    "children": [],
    "unreadable": ""
  },
  // args:
  {
    "name": "i",
    "addr": 825741434728,
    "onlyAddr": false,
    "type": "int",
    "realType": "int",
    "flags": 0, // <-- the argument "i" has no flag
    "kind": 2,
    "value": "42",
    "len": 0,
    "cap": 0,
    "children": [],
    "unreadable": ""
  }
]
```

### LINKED ISSUE #1106: Change frame caching strategy in proc.(*EvalScope).variablesByTag to support compositeMemory
The simple strategy we are employing now can't deal with variables that are not stored contiguously in memory or are stored (partially) into CPU registers.

See https://github.com/derekparker/delve/pull/1103

aarzilli:
```
proc: Flag shadowed arguments as shadowed

Fixes #951

proc: cache entire frame in FrameToScope instead of variablesByTag

Caching the frame in variablesByTag is problematic:

1. accounting for variables that are (partially) stored in registers is
complicated (see issue #1106)
2. for some types (strings, interfaces...) simply creating the Variable
object reads memory, which therefore happens before we can do any
caching.

Instead cache the entire frame when the EvalScope object is created.
The cached range is between the SP value of the current frame and the
CFA of the preceeding frame, if available, or the CFA of the current
frame otherwise.

Fixes #1106

proc: change memCache to delay reading

Change memCache so that the preloaded memory is not read immediately
but only after the actual read to the preloaded range.

This allows us to request caching the entire stack frame every time we
create an eval scope and no unnecessary reads will be made even if the
user is just trying to evaluate a global variable.

```

### DIFF
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -499,7 +499,7 @@ func ConvertEvalScope(dbp Process, gid, frame int) (*EvalScope, error) {
 		thread = g.Thread
 	}
 
-	locs, err := g.Stacktrace(frame)
+	locs, err := g.Stacktrace(frame + 1)
 	if err != nil {
 		return nil, err
 	}
@@ -508,7 +508,7 @@ func ConvertEvalScope(dbp Process, gid, frame int) (*EvalScope, error) {
 		return nil, fmt.Errorf("Frame %d does not exist in goroutine %d", frame, gid)
 	}
 
-	return FrameToScope(dbp.BinInfo(), thread, g, locs[frame]), nil
+	return FrameToScope(dbp.BinInfo(), thread, g, locs[frame:]...), nil
 }
 
 // FrameToScope returns a new EvalScope for frames[0].
@@ -516,12 +516,27 @@ func ConvertEvalScope(dbp Process, gid, frame int) (*EvalScope, error) {
 // frames[0].Regs.SP() and frames[1].Regs.CFA will be cached.
 // Otherwise all memory between frames[0].Regs.SP() and frames[0].Regs.CFA
 // will be cached.
-func FrameToScope(bi *BinaryInfo, thread MemoryReadWriter, g *G, frame Stackframe) *EvalScope {
+func FrameToScope(bi *BinaryInfo, thread MemoryReadWriter, g *G, frames ...Stackframe) *EvalScope {
 	var gvar *Variable
 	if g != nil {
 		gvar = g.variable
 	}
-	s := &EvalScope{Location: frame.Call, Regs: frame.Regs, Mem: thread, Gvar: gvar, BinInfo: bi, frameOffset: frame.FrameOffset()}
-	s.PC = frame.lastpc
+
+	// Creates a cacheMem that will preload the entire stack frame the first
+	// time any local variable is read.
+	// Remember that the stack grows from downward in memory.
+	minaddr := frames[0].Regs.SP()
+	var maxaddr uint64
+	if len(frames) > 1 && frames[0].SystemStack == frames[1].SystemStack {
+		maxaddr = uint64(frames[1].Regs.CFA)
+	} else {
+		maxaddr = uint64(frames[0].Regs.CFA)
+	}
+	if maxaddr > minaddr && maxaddr-minaddr < maxFramePrefetchSize {
+		thread = cacheMemory(thread, uintptr(minaddr), int(maxaddr-minaddr))
+	}
+
+	s := &EvalScope{Location: frames[0].Call, Regs: frames[0].Regs, Mem: thread, Gvar: gvar, BinInfo: bi, frameOffset: frames[0].FrameOffset()}
+	s.PC = frames[0].lastpc
 	return s
 }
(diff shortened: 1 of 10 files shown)


aarzilli:
The aim of this PR is to fix two bugs caused by variablesByTag: #1106 and #951
	
	#1106 refers to the difficulty that variablesByTag has in dealing with variables that aren't stored completely on stack due to its caching strategy. 
	
	There is another, subtler, problem with caching and variablesByTag, the caching interval is calculated after variablesByTag has created Variable objects for all the variables it wants to load, since calling newVariable does sometimes read memory (for example for strings and interfaces) these reads happen before the caching.
	
	My solution here is to move the caching to before variablesByTag runs, where the EvalScope object is created and to use the same stacktrace that is used to create the EvalScope to determine how much of the stack to preload. Since an EvalScope is also needed when the user evaluates a global variable, and in that case it would be wasteful to load anything from the stack, I changed the caching strategy of memCache to only load the cache the first time something from the cached range is actually read.
	
	#951 refers to the fact that shadowed arguments are not flagged as such. To fix this problem I changed variablesByTag to load unconditionally all local variables (and renamed it to Locals). With this done arguments that are shadowed can be easily flagged as such. Filtering between arguments and variables is done at a later stage.

pkg/proc/proc.go
```diff
@@ -508,16 +508,35 @@ func ConvertEvalScope(dbp Process, gid, frame int) (*EvalScope, error) {
...
 	var gvar *Variable
 	if g != nil {
 		gvar = g.variable
 	}
-	s := &EvalScope{Location: frame.Call, Regs: frame.Regs, Mem: thread, Gvar: gvar, BinInfo: bi, frameOffset: frame.FrameOffset()}
-	s.PC = frame.lastpc
+
+	// Creates a cacheMem that will preload the entire stack frame the first
+	// time any local variable is read.
+	// Remember that the stack grows from downward in memory.
```
derekparker: nit: remove "from".

pkg/proc/variables.go
```diff
@@ -614,12 +620,38 @@ func (scope *EvalScope) SetVariable(name, value string) error {
...
 
 // FunctionArguments returns the name, value, and type of all current function arguments.
 func (scope *EvalScope) FunctionArguments(cfg LoadConfig) ([]*Variable, error) {
-	return scope.variablesByTag(dwarf.TagFormalParameter, &cfg)
+	vars, err := scope.Locals()
+	if err != nil {
+		return nil, err
+	}
+	vars = filterVariables(vars, func(v *Variable) bool {
+		return (v.Flags & (VariableArgument | VariableReturnArgument)) != 0
```
derekparker: I think we should exclude return args here since we can distinguish now and have a new method that is `FunctionReturnVals`.

aarzilli: That wouldn't be backwards compatible for the API (unless we concatenate this and FunctionReturnVals in debugger.go or rpc2/server)

derekparker: That's true, however unfortunate.

PR #1187, opened 2018-04-18
proc/native,proc/gdbserial: let target access terminal

### LINKED ISSUE #65: Debugged process can't read from stdin
A traced program will read its stdin from os.DevNull instead of os.Stdin, since in proctl.Launch() it is only stderr and stdout. I can't figure out the reason for this, and simply adding `proc.Stdin = os.Stdin` in the Launch() function will make a simple test program behave as expected.

aarzilli:
```
proc/native,proc/gdbserial: let target access terminal

Change the linux verison of proc/native and proc/gdbserial (with
debugserver) so that they let the target process use the terminal when
delve is launched in headless mode.

Windows already worked, proc/gdbserial (with rr) already worked.
I couldn't find a way to make proc/gdbserial (with lldb-server) work.

No tests are added because I can't think of a way to test for
foregroundness of a process.

Fixes #65

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -191,7 +191,7 @@ func New(process *os.Process) *Process {
 }
 
 // Listen waits for a connection from the stub.
-func (p *Process) Listen(listener net.Listener, path string, pid int) error {
+func (p *Process) Listen(listener net.Listener, path string, pid int, foreground bool) error {
 	acceptChan := make(chan net.Conn)
 
 	go func() {
@@ -205,7 +205,7 @@ func (p *Process) Listen(listener net.Listener, path string, pid int) error {
 		if conn == nil {
 			return errors.New("could not connect")
 		}
-		return p.Connect(conn, path, pid)
+		return p.Connect(conn, path, pid, foreground)
 	case status := <-p.waitChan:
 		listener.Close()
 		return fmt.Errorf("stub exited while waiting for connection: %v", status)
@@ -217,7 +217,7 @@ func (p *Process) Dial(addr string, path string, pid int) error {
 	for {
 		conn, err := net.Dial("tcp", addr)
 		if err == nil {
-			return p.Connect(conn, path, pid)
+			return p.Connect(conn, path, pid, false)
 		}
 		select {
 		case status := <-p.waitChan:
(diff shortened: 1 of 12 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -423,9 +434,12 @@ func LLDBLaunch(cmd []string, wd string) (*Process, error) {
 		proc = exec.Command("lldb-server", args...)
 	}
 
-	if logflags.LLDBServerOutput() || logflags.GdbWire() {
+	if logflags.LLDBServerOutput() || logflags.GdbWire() || foreground {
 		proc.Stdout = os.Stdout
 		proc.Stderr = os.Stderr
+		if foreground {
```
derekparker: Let's just move this out of the enclosing branch and have it stand on it's own so we're not checking the boolean twice, the short circuit logic doesn't buy us anything and makes the code more verbose.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -380,7 +388,7 @@ func getLdEnvVars() []string {
 // LLDBLaunch starts an instance of lldb-server and connects to it, asking
 // it to launch the specified target program with the specified arguments
 // (cmd) on the specified directory wd.
-func LLDBLaunch(cmd []string, wd string) (*Process, error) {
+func LLDBLaunch(cmd []string, wd string, foreground bool) (*Process, error) {
```
derekparker: You mention not being able figure out LLDB but seem to have implemented it here. If for whatever reason it truly is not supported we should return an error here and mention in the docs.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -495,7 +509,7 @@ func LLDBAttach(pid int, path string) (*Process, error) {
 	p.conn.isDebugserver = isDebugserver
 
 	if listener != nil {
-		err = p.Listen(listener, path, pid)
+		err = p.Listen(listener, path, pid, false)
```
derekparker: Is this because the implementation is not complete, so we default to never foregrounding the process?

pkg/proc/gdbserial/gdbserver_windows.go
```diff
@@ -5,3 +5,7 @@ import "syscall"
 func backgroundSysProcAttr() *syscall.SysProcAttr {
 	return nil
 }
+
+func moveToForeground(pid int) {
+	panic("not on windows")
```
derekparker: nit: Give a better error here.

pkg/proc/gdbserial/gdbserver.go
aarzilli: LLDBLaunch launches either lldb or debugserver depending on the one it finds.
Returning an error would prevent connecting to lldb-server in headless mode, which still works fine except for the fact that it won't get stdin.

pkg/proc/gdbserial/gdbserver.go
aarzilli: It just doesn't make sense to bring the stub process to the foreground if we are attaching to a different process.

derekparker: Ah, didn't notice this was in attach not launch, my fault.

PR #1192, opened 2018-04-19
cmd/dlv: Revert 4e177b, use new flag name for debug logging options

### LINKED ISSUE #1188: Non backwards-compatible change to --log
1. What version of Delve are you using (`dlv version`)? 1.0.0
2. What version of Go are you using? (`go version`)? go1.10.1
3. What operating system and processor architecture are you using? MacOS 10.13.4
4. What did you do?

A week ago I was using Delve successfully to remote debug my Go app in Docker.  The Delve binary is pulled into the Docker image on each build using `go get` so it tracks the latest version at all times.  Recently (the last 2 days or so), it has stopped working with the error referenced in the title.

I ran a local test (outside Docker) where I still had the old Delve binary from a week ago.  I first attempted to run my debugging with the old binary (it worked), I then did `go get -u` to get latest Delve & attempted to run it again immediately - it failed.  Here's the shell output (which includes the Delve build IDs).  Note I've elided my company's package names.

```
➜  go build github.com/...<elided>.../cmd/server
➜  pwd
/Users/rory/code/go/src/github.com/.....<elided>
➜  dlv debug --headless --listen=:2345 --log github.com/...<elided>...
2018/04/18 18:37:14 server.go:73: Using API v1
2018/04/18 18:37:14 debugger.go:98: launching process with args: [/Users/rory/code/go/src/github.com/...<elided>..../debug]
API server listening at: [::]:2345
^C%                                                                                                                                                                                         ➜  dlv version
Delve Debugger
Version: 1.0.0
Build: a708d00e78bc9e6d344d66b4428820c061d35d83
➜  which dlv
/Users/rory/code/go/bin/dlv
```

Then I updated & tried again: 
```
➜  go get -u github.com/derekparker/delve/cmd/dlv
➜ dlv version
Delve Debugger
Version: 1.0.0
➜  dlv debug --headless --listen=:2345 --log github.com/...<elided>.../cmd/server
can't load package: package github.com/...<elided>.... no Go files in /Users/rory/code/go/src/github.com/...<elided>....
exit status 1
```

aarzilli:
```
cmd/dlv: Revert 4e177b, use new flag name for debug logging options

Previously to 4e177bb99acc511897f9cdbfc6cbc50d92ae4725 it was possible
to use --log without arguments to enable logging, this commit reenables
that use case.

The new functionality of the --log flag moved to a new flag name
--logx.

Fixes #1188

```

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -26,7 +26,9 @@ import (
 
 var (
 	// Log is whether to log debug statements.
-	Log string
+	Log bool
+	// LogOutput is a comma separated list of components that should produce debug output.
+	LogOutput string
 	// Headless is whether to run without terminal.
 	Headless bool
 	// APIVersion is the requested API version while running headless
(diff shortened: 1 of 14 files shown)


derekparker:
My only nit is the name `logx`. I'm fine with something more verbose like `--log-output`.

aarzilli:
Done.

derekparker:
@aarzilli we should also note in the docs what the default log output is without this other flag specified.

aarzilli:
Done.

cmd/dlv/cmds/commands.go
```diff
@@ -86,11 +88,12 @@ func New(docCall bool) *cobra.Command {
 	}
 
 	RootCommand.PersistentFlags().StringVarP(&Addr, "listen", "l", "localhost:0", "Debugging server listen address.")
-	RootCommand.PersistentFlags().StringVarP(&Log, "log", "", "false", `Comma separated list of components that should produce debug output, possible values:
+	RootCommand.PersistentFlags().BoolVarP(&Log, "log", "", false, "Enable debugging server logging.")
+	RootCommand.PersistentFlags().StringVarP(&LogOutput, "log-output", "", "false", `Comma separated list of components that should produce debug output, possible values:
```
derekparker: Any reason to default to "false" instead of just empty string? Also, since we have both flags we should force users to provide both `--log` to enable logging and then use the other flag to then control the output. It seems that's how it works implicitly now, but would be good to print an error or something if the flags are used incorrectly.

aarzilli: It was like this because of the backwards compatibility when it was still called `--log`, you are right and I have changed it.

PR #1199, opened 2018-05-03
check that t.conf is not nil in loadConfig

yuval-k:
fixes a SIGSEGV I had when debugger with no tty and trying to print a variable.


derekparker:
@aarzilli agreed, we can do that in a follow up.

PR #1208, opened 2018-05-11
proc,service: display return values when stepping out of a function

aarzilli:
```
proc,service: display return values when stepping out of a function

Displays the return values of the current function when we step out of
it after executing a step, next or stepout command.

Implementation of this feature is tricky: when the function has
returned the return variables are not in scope anymore. Implementing
this feature requires evaluating variables that are out of scope, using
a stack frame that doesn't exist anymore.

We can't calculate the address of these variables when the
next/step/stepout command is initiated either, because between that
point and the time where the stepout breakpoint is actually hit the
goroutine stack could grow and be moved to a different memory address.
```
```
proc: move AllGCache to a common struct

Add a new method "Common" to proc.Process that returns a pointer to a
struct that pkg/proc can use to store its things, independently of the
backend.

This is used here to replace the AllGCache typecasts, it will also be
used to store the return values of the stepout breakpoint and the state
for injected function calls.

```

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -1,6 +1,7 @@
 package proc
 
 import (
+	"debug/dwarf"
 	"errors"
 	"fmt"
 	"go/ast"
@@ -53,6 +54,10 @@ type Breakpoint struct {
 	Cond ast.Expr
 	// internalCond is the same as Cond but used for the condition of internal breakpoints
 	internalCond ast.Expr
+
+	// ReturnInfo describes how to collect return variables when this
+	// breakpoint is hit as a return breakpoint.
+	returnInfo *returnBreakpointInfo
 }
 
 // Breakpoint Kind determines the behavior of delve when the
@@ -100,6 +105,13 @@ type InvalidAddressError struct {
 
 func (iae InvalidAddressError) Error() string {
 	return fmt.Sprintf("Invalid address %#v\n", iae.Address)
+}
+
+type returnBreakpointInfo struct {
+	retFrameCond ast.Expr
+	fn           *Function
+	frameOffset  int64
+	spOffset     int64
 }
 
 // CheckCondition evaluates bp's condition on thread.
@@ -302,6 +314,7 @@ func (bpmap *BreakpointMap) ClearInternalBreakpoints(clearBreakpoint clearBreakp
 	for addr, bp := range bpmap.M {
 		bp.Kind = bp.Kind & UserBreakpoint
 		bp.internalCond = nil
+		bp.returnInfo = nil
 		if bp.Kind != 0 {
 			continue
 		}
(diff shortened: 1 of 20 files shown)


derekparker:
Will review this tomorrow.

pkg/proc/breakpoints.go
```diff
@@ -354,3 +367,82 @@ func (bpstate *BreakpointState) String() string {
...
+	scope, err := GoroutineScope(thread)
+	if err != nil {
+		return returnInfoError("could not get scope", err, thread)
+	}
+	v, err := scope.evalAST(rbpi.retFrameCond)
+	if err != nil || v.Unreadable != nil || v.Kind != reflect.Bool {
+		// This condition was evaluated as part as the breakpoint condition
+		// evaluation, if the errors happen they will be reported as part of the
+		// condition errors.
+		return nil
```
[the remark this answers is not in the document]
aarzilli: Yes, in evalBreakpointCondition

aarzilli:
Done

pkg/proc/threads.go
```diff
@@ -50,6 +52,17 @@ func (tbe ThreadBlockedError) Error() string {
 	return ""
 }
 
+// CommonThread contains fields used by this package, common to all
+// implementations of the Thread interface.
+type CommonThread struct {
+	returnValues []*Variable
```
derekparker: I don't think we should store this extra state at all, it can be derived. The thread already knows the breakpoint it is stopped at, and we can ask that breakpoint to collect return values. This just gives us one more thing to clear when we continue the process.

derekparker: As for optimizations, the reads of these vars should already be cached in our memory layer.

aarzilli: There some problems in doing this:

1. we clear all internal breakpoints before exiting Continue, changing this is pretty complicated since they would still have to be cleared before the next call to Next/Step/StepOut
2. from the point of view of backwards compatibility: we never propagated internal breakpoints to clients, it's probably not a big deal, but this is safer
3. I'd like to use this same interface for the return values of a function call where there is no breakpoint

aarzilli: We could put a fake breakpoint on the thread, with a special negative ID (like unrecovered-panic) and marked as UserBreakpoint and save the return values there.

pkg/proc/breakpoints.go
```diff
@@ -354,3 +367,82 @@ func (bpstate *BreakpointState) String() string {
...
+
+func configureReturnBreakpoint(bp *Breakpoint, topframe *Stackframe, retFrameCond ast.Expr) {
+	if topframe.Current.Fn == nil {
+		return
+	}
+	bp.returnInfo = &returnBreakpointInfo{
+		retFrameCond: retFrameCond,
+		fn:           topframe.Current.Fn,
+		frameOffset:  topframe.FrameOffset(),
+		spOffset:     topframe.FrameOffset() - 0x8, // must be the value that SP had at the entry point of the function
```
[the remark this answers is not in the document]
derekparker: Instead of hard coding this to `0x8` and repeating sins of the past, we should pass in the pointer size.

pkg/proc/proc.go
```diff
@@ -158,6 +161,7 @@ func Continue(dbp Process) error {
 					return err
 				}
 			} else {
+				curthread.Common().returnValues = curbp.Breakpoint.returnInfo.Collect(curthread)
```
derekparker: `curbp.Breakpoint.returnInfo` is a pointer and _could_ be nil here. I suppose we could always assume an internal non-step breakpoint must always have `returnInfo` set, however we should be more intentional about this and check for the pointer being nil and panic with a better error message, since this would be a programmer error.

Alternatively, we could ignore this if there is no `returnInfo` set.

pkg/proc/threads.go
derekparker: That's a good point, I was mistakenly thinking of other breakpoints loading vars that are user set.

Just thinking out loud, tell me if this sounds insane: what if instead of storing _any_ kind of state that we have to manage on the thread or process, we return some sort of global state struct from any function in `proc` which manipulates process state? This would prevent us from ever having to manage or clear any mutable state on any long-lived structure we keep in memory, and enforces boundaries that anything that calls `proc` functions just get state back and can ignore any lower level details that `proc` should encapsulate and hide. Essentially we move `(*Debugger) State()` down a level into proc. the `Debugger` could still handle anything higher-level, but instead of it asking the process for its state when it needs it, the lower layer could just return the information directly for any call that changes process state.

I haven't thought through the full implications of this, but seemed like an interesting enough idea to propose.

derekparker:
Just a few more comments. Once this merges I'm going to implement the trace improvements on top of this since we'll have a reliable way to get return vals.

pkg/proc/proc.go
aarzilli: `(*returnBreakpointInfo).Collect` works fine on nil pointers.

pkg/proc/threads.go
aarzilli: It would work for this, I think, because it's just something extra that we're returning. But ee're also going to need some place to store the state of ongoing function calls, it can't be a local variable of proc.Continue because proc.Continue can return while an injected function call is still in progress, so it needs to be somewhere in Process/Thread.

pkg/proc/breakpoints.go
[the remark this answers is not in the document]
aarzilli: Agreed.

pkg/proc/threads.go
derekparker: That's a good point. What do you think of a compromise to do both, or do you think the latter negates the former? I'm just trying to remove as much state to manage as possible.

aarzilli: The semantics of Continue/Next/Step/StepOut is "resume the process, wait for a stop, mutate Process to reflect the new state", they do all their communication by mutating Process,  I'd rather not add a second return value just for this special thing. It's going to make the use of Continue more verbose, including in our tests.

aarzilli: I'd rather go with either extra state or the fake breakpoint route.

derekparker: I don't want to add a separate value and spread the state around either. I'll keep thinking on this problem and see if there is a sensible refactor here, until then this is fine.

derekparker:
@aarzilli needs rebase.

PR #1210, opened 2018-05-17
Function Calls

aarzilli:


aarzilli:
Depends on: https://go-review.googlesource.com/c/go/+/109699

aarzilli:
The corresponding CL has been merged, this now only depends on the go1.11 support branch (#1157) and on #1208 (which needs some decisions to be made).

This implements the function call injection protocol on *linux only* and doesn't support a number of features that one would reasonably expect (see the TODO comments in pkg/proc/fncall.go for a list).

pkg/proc/fncall.go
```diff
@@ -0,0 +1,545 @@
...
+			break
+		}
+		scope, err := ThreadScope(thread)
+		if err != nil {
+			fncall.err = fmt.Errorf("could not get return values: %v", err)
+			break
+		}
+
+		// pretend we are still inside the function we called
+		fakeFunctionEntryScope(scope, fncall.fn, int64(regs.SP()), regs.SP()-0x8)
```
derekparker: Shouldn't assume address size here.

pkg/proc/fncall.go
```diff
@@ -0,0 +1,545 @@
...
+		fncall.panicvar.Name = "~panic"
+		fncall.panicvar.loadValue(*fncall.loadCfg)
+		if fncall.panicvar.Unreadable != nil {
+			fncall.err = fmt.Errorf("could not get panic: %v", fncall.panicvar.Unreadable)
+			break
+		}
+
+	default:
+		// Got an unknown AX value, this is probably bad but the safest thing
+		// possible is to ignore it and hope it didn't matter.
```
derekparker: Since we have conditional logging we should log the value of AX.

pkg/proc/fncall.go
```diff
@@ -0,0 +1,545 @@
...
+		return nil, err
+	}
+	typ, err := bi.findType(typename)
+	if err != nil {
+		return nil, err
+	}
+	v := scope.newVariable("", uintptr(regs.SP()), typ, scope.Mem)
+	v.loadValue(loadCfg)
+	if v.Unreadable != nil {
+		return nil, err
```
derekparker: This should be `return nil, v.Unreadable`.

pkg/proc/native/registers_linux_amd64.go
```diff
@@ -90,13 +91,31 @@ func (r *Regs) GAddr() (uint64, bool) {
 }
 
 // SetPC sets RIP to the value specified by 'pc'.
-func (r *Regs) SetPC(t proc.Thread, pc uint64) (err error) {
-	thread := t.(*Thread)
+func (thread *Thread) SetPC(pc uint64) (err error) {
```
derekparker: Can do away with named return and pre-declaration of `ir`.

pkg/proc/registers.go
```diff
@@ -33,6 +34,9 @@ type Register struct {
 	Value string
 }
 
+type SavedRegisters interface {
```
derekparker: Instead of defining a new type can we just use the existing `proc.Registers` interface?

derekparker: Ah got it, I like that solution then.

derekparker:
I think this is a great start, but we should make a note in the docs about current shortcomings.

* Can't pass in new vars (e.g. `call myfn(1,2)`)
* Can't call methods
* Arch / OS limitations (this only applies to native IIRC, correct me if I'm wrong)
* etc...

aarzilli:
Where do you think is the appropriate place for this? The docstring of call?

aarzilli:
Done.

PR #1211, opened 2018-05-17
[WIP] Add capability to jump to location in order skip execution of some instructions

### LINKED ISSUE #1045: Support for instruction reordering / jumping
This was received on GoLands issue tracker, see issue [GO-5109](https://youtrack.jetbrains.com/issue/GO-5109).

The original request is:

> a Visual Studio/C#-like "Set Next Statement" debug feature would be amazing. 
> It basically allows you to move the pointer to change the execution flow. 
> See: https://msdn.microsoft.com/en-us/library/y740d9d3.aspx

From my understanding of the page documentation:

> While the debugger is paused, you can move the instruction pointer to set the next statement of code to be executed. A yellow arrowhead in the margin of a source or Disassembly window marks the location of the next statement to be executed. By moving this arrowhead, you can skip over a portion of code or return to a line previously executed. You can use this for situations such as skipping a section of code that contains a known bug.

I believe this effectively means that we'd need to insert jumps in the source code to allow the execution to skip over code portions.

Now, given Go's nature of things, I don't expect this to be 100% bulletproof in the first version, so basically no validations would be required (such as making sure that there's no new memory allocation done by creating a variable and so on). Later on we could think about that.

As I don't know how complex this is to do, in general or in Go in particular, I'll leave this with you. Please let me know if there's anything that I can help.

Thank you.

dlsniper:
This currently is a more or less work in progress, it depends on the feature set we want to support this feature. In my opinion, this could be flagged as an experimental feature and if the user has an issue then it's their fault. Maybe it's not a 100% good statement to have. There are cases where this could be useful to have and a user that understands how to use it, will be able to use it to the full extent.

I've tested this only on Windows, I'll let the integration test do the job on the other systems.

Cross function jumping works as well, but only in the toy example below, I have no clue how the runtime will be affected by this. 

Fixes #1045

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -128,6 +128,11 @@ See also: "help on", "help cond" and "help clear"`},
 		{aliases: []string{"step-instruction", "si"}, cmdFn: c.stepInstruction, helpMsg: "Single step a single cpu instruction."},
 		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: "Step over to next source line."},
 		{aliases: []string{"stepout"}, cmdFn: c.stepout, helpMsg: "Step out of the current function."},
+		{aliases: []string{"jump", "j"}, cmdFn: executionPoint, helpMsg: `Jump execution to a new location. This will NOT execute the lines that are skipped.
+
+	break <linespec>
+
+See $GOPATH/src/github.com/derekparker/delve/Documentation/cli/locspec.md for the syntax of linespec. This works only with filenames and line numbers, or address numbers.`},
 		{aliases: []string{"threads"}, cmdFn: threads, helpMsg: "Print out info for every traced thread."},
 		{aliases: []string{"thread", "tr"}, cmdFn: thread, helpMsg: `Switch to the specified thread.
 
@@ -1063,8 +1068,47 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
 	return nil
 }
 
+func setExecutionPoint(t *Term, ctx callContext, argstr string) error {
+	args := strings.SplitN(argstr, " ", 2)
+
+	requestedEp := &api.ExecutionPoint{}
+	locspec := ""
+	if len(args) != 1 {
+		return fmt.Errorf("address required")
+	} else {
+		locspec = argstr
+	}
+
+	locs, err := t.client.FindLocation(ctx.Scope, locspec)
+	if err != nil {
+		locspec = argstr
+		var err2 error
+		locs, err2 = t.client.FindLocation(ctx.Scope, locspec)
+		if err2 != nil {
+			return err
+		}
+	}
+	for _, loc := range locs {
+		requestedEp.Addr = loc.PC
+
+		err := t.client.SetExecutionPoint(requestedEp)
+		if err != nil {
+			return err
+		}
+
+		requestedEp.File = loc.File
+		requestedEp.Line = loc.Line
+		fmt.Printf("execution point jumped to %v\n", formatExecutionPointLocation(requestedEp))
+	}
+	return nil
+}
+
 func breakpoint(t *Term, ctx callContext, args string) error {
 	return setBreakpoint(t, ctx, false, args)
+}
+
+func executionPoint(t *Term, ctx callContext, args string) error {
+	return setExecutionPoint(t, ctx, args)
 }
 
 func tracepoint(t *Term, ctx callContext, args string) error {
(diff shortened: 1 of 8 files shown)


dlsniper:
I added the ` [WIP] ` tag for now, just in case.

pkg/terminal/command.go
```diff
@@ -128,6 +128,11 @@ See also: "help on", "help cond" and "help clear"`},
 		{aliases: []string{"step-instruction", "si"}, cmdFn: c.stepInstruction, helpMsg: "Single step a single cpu instruction."},
 		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: "Step over to next source line."},
 		{aliases: []string{"stepout"}, cmdFn: c.stepout, helpMsg: "Step out of the current function."},
+		{aliases: []string{"jump", "j"}, cmdFn: executionPoint, helpMsg: `Jump execution to a new location. This will NOT execute the lines that are skipped.
+
+	break <linespec>
```
aarzilli: s/break/jump/

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
...
+	}
+	return nil
+}
+
 func breakpoint(t *Term, ctx callContext, args string) error {
 	return setBreakpoint(t, ctx, false, args)
 }
 
+func executionPoint(t *Term, ctx callContext, args string) error {
+	return setExecutionPoint(t, ctx, args)
```
aarzilli: I think setExecutionPoint should just be inlined here, since it's not called anywhere else.

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
 	return nil
 }
 
+func setExecutionPoint(t *Term, ctx callContext, argstr string) error {
+	args := strings.SplitN(argstr, " ", 2)
```
aarzilli: locspecs can contain spaces the better thing to do is to use all the argstr as locspec and check that it's `!= ""`

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
...
+
+	requestedEp := &api.ExecutionPoint{}
+	locspec := ""
+	if len(args) != 1 {
+		return fmt.Errorf("address required")
```
aarzilli: Other commands return `errors.New("not enough arguments")` in this situation, this should to for uniformity's sake.

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
...
+	} else {
+		locspec = argstr
```
aarzilli: The if above returns, this doesn't need to be inside an else body.

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
...
+	}
+
+	locs, err := t.client.FindLocation(ctx.Scope, locspec)
+	if err != nil {
+		locspec = argstr
```
aarzilli: I don't understand what's going on.

pkg/terminal/command.go
```diff
@@ -1063,10 +1068,49 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) err
...
+		var err2 error
+		locs, err2 = t.client.FindLocation(ctx.Scope, locspec)
+		if err2 != nil {
+			return err
+		}
+	}
+	for _, loc := range locs {
```
aarzilli: Jumping multiple times is idempotent to just doing the last jump, this should either reject the case where there are multiple locations, jump to the first returned location or jump to the last returned location (I think for now the appropriate thing is rejecting).

service/api/types.go
```diff
@@ -73,6 +73,16 @@ type Breakpoint struct {
 	TotalHitCount uint64 `json:"totalHitCount"`
 }
 
+// ExecutionPoint addresses a location at which process execution will jump to.
+type ExecutionPoint struct {
```
aarzilli: This type is basically the same thing as an api.Location. Given that it's constructed starting from an api.Location I think we should pass in the api.Location directly.

Alternatively just pass the address.

service/debugger/debugger.go
```diff
@@ -354,6 +344,49 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 	return createdBp, nil
 }
 
+// SetExecutionPoint will jump the execution point.
+func (d *Debugger) SetExecutionPoint(requestedExecPoint *api.ExecutionPoint) error {
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
+	if len(requestedExecPoint.File) > 0 {
```
aarzilli: I know this is what we are doing for breakpoints, but that's because of backwards compatibility. Given that the intended use of this API is to always call FindLocations first I would lean towards not repeating the sins of the past.

derekparker:
Closing this due to lack of response.

@dlsniper please rebase and reopen when you have the bandwidth to continue working on this.

PR #1212, opened 2018-05-18
    command/terminal: Printing stack trace of all the goroutines

acshekhara1:
* command/terminal: Printing stack trace of all the goroutines

    Add -t falg to "goroutines" command. For example, "goroutines -t" will
    print all the goroutines along with the stack trace.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -142,13 +142,14 @@ See also: "help on", "help cond" and "help clear"`},
 If called with the linespec argument it will delete all the breakpoints matching the linespec. If linespec is omitted all breakpoints are deleted.`},
 		{aliases: []string{"goroutines"}, cmdFn: goroutines, helpMsg: `List program goroutines.
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)]
+	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location) ] [ -t (stack trace)]
 
 Print out info for every goroutine. The flag controls what information is shown along with each goroutine:
 
 	-u	displays location of topmost stackframe in user code
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
+	-t	displyes stack trace of goroutine
 
 If no flag is specified the default is -u.`},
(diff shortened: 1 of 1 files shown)


acshekhara1:
I have run the tests. Now they are running.

pkg/terminal/command.go
```diff
@@ -561,6 +567,11 @@ func goroutines(t *Term, ctx callContext, argstr string) error {
 			prefix = "* "
 		}
 		fmt.Printf("%sGoroutine %s\n", prefix, formatGoroutine(g, fgl))
+		if printStack == true {
```
aarzilli: just `if printStack {`

pkg/terminal/command.go
```diff
@@ -561,6 +567,11 @@ func goroutines(t *Term, ctx callContext, argstr string) error {
...
+			if _, err := t.client.SwitchGoroutine(g.ID); err == nil {
+				stackCommand(t, ctx, "")
```
aarzilli: Replace this if with:

```
	stack, err := t.client.Stacktrace(g.ID, 10, nil)
	if err != nil {
		return err
	}
	printStack(stack, "\t", false)
```

That way the stack trace is indented and we don't call `SwitchGoroutine` which changes the state of the program.

acshekhara1:
Feedback addressed.

acshekhara1:
Merged from master of derekparker.

pkg/terminal/command.go
```diff
@@ -142,13 +142,14 @@ See also: "help on", "help cond" and "help clear"`},
...
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)]
+	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location) ] [ -t (stack trace)]
 
 Print out info for every goroutine. The flag controls what information is shown along with each goroutine:
 
 	-u	displays location of topmost stackframe in user code
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
+	-t	displyes stack trace of goroutine
```
derekparker: s/displyes/displays/

acshekhara1:
Addressed comments from DerekParker.

PR #1214, opened 2018-05-18
proc: add support for dwz compressed DWARF

slp:
'dwz' is a tool that reduces the size of DWARF sections by deduplicating
symbols. The deduplicated symbols are moved from their original 'compile
unit' to a 'partial unit', which is then referenced from its original
location with an 'imported unit' tag.

In the case of Go binaries, all symbols are located in a single 'compile
unit', and the name of each symbol contains a reference to its package,
so 'dwz' is not able to deduplicate them. But still, some C symbols
included in the binary are deduplicated, which also alters the structure
of the DWARF sections, making delve unable to parse them (crashing in
the attempt).

While it would've been possible to simply ignore the C symbols, or
blindly loading all then into BinaryInfo members (packageVars,
Functions...), for correctness sake this change tries to do the right
thing, staging symbols into temporary partialUnit objects, moving them
to BinaryInfo when they are actually requested  by a 'imported unit'
tag.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3697,3 +3697,20 @@ func TestIssue951(t *testing.T) {
 		}
 	})
 }
+	
+func TestDWZCompression(t *testing.T) {
+	// If dwz is not available in the system, skip this test
+	if _, err := exec.LookPath("dwz"); err != nil {
+		t.Skip("dwz not installed")
+	}
+
+	withTestProcessArgs("dwzcompression", t, ".", []string{}, protest.EnableDWZCompression, func(p proc.Process, fixture protest.Fixture) {
+		_, err := setFunctionBreakpoint(p, "C.fortytwo")
+		assertNoError(err, t, "setFunctionBreakpoint()")
+		assertNoError(proc.Continue(p), t, "first Continue()")
+		val := evalVariable(p, t, "stdin")
+		if val.RealType == nil {
+			t.Errorf("Can't find type for \"stdin\" global variable")
+		}
+	})
+}
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -188,12 +188,18 @@ func (bi *BinaryInfo) loadDebugInfoMaps(debugLineBytes []byte, wg *sync.WaitGrou
 	bi.consts = make(map[dwarf.Offset]*constantType)
 	reader := bi.DwarfReader()
 	var cu *compileUnit = nil
+	var pu *partialUnit = nil
+	var partialUnits = make(map[dwarf.Offset]*partialUnit)
 	for entry, err := reader.Next(); entry != nil; entry, err = reader.Next() {
 		if err != nil {
 			break
 		}
 		switch entry.Tag {
 		case dwarf.TagCompileUnit:
+			if pu != nil {
+				partialUnits[pu.entry.Offset] = pu
+				pu = nil
+			}
 			cu = &compileUnit{}
 			cu.entry = entry
 			if lang, _ := entry.Val(dwarf.AttrLanguage).(int64); lang == dwarfGoLanguage {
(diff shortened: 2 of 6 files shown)


pkg/proc/types.go
```diff
@@ -218,45 +224,124 @@ func (bi *BinaryInfo) loadDebugInfoMaps(debugLineBytes []byte, wg *sync.WaitGrou
...
 
+		case dwarf.TagPartialUnit:
+			if pu != nil {
+				partialUnits[pu.entry.Offset] = pu
+				pu = nil
+			}
+			pu = &partialUnit{}
+			pu.entry = entry
+			pu.types = make(map[string]dwarf.Offset)
+			pu.variables = []packageVar{}
```
aarzilli: This line and the two following are unnecessary.

slp:
> This is not true anymore in go1.10 (although there still are no duplicate symbols for dwz to compress)

Thanks, I wasn't aware, I've been working with 1.9.4.

> Is this tool documented anywhere? If it is it would be nice to have a link to the spec here.

The only documentation I'm aware of is its manpage. This is the contents of the DESCRIPTION section:

```
dwz  is  a program that attempts to optimize DWARF debugging information contained in ELF shared
libraries and ELF executables for size,  by  replacing  DWARF  information  representation  with
equivalent smaller representation where possible and by reducing the amount of duplication using
techniques from DWARF standard appendix E - creating DW_TAG_partial_unit compilation units (CUs)
for  duplicated  information and using DW_TAG_imported_unit to import it into each CU that needs
it.

The tool handles DWARF 32-bit format debugging sections of versions 2, 3 and 4  and  GNU  exten‐
sions on top of those, though using DWARF 4 or worst case DWARF 3 is strongly recommended.

The  tool  has  two main modes of operation, without the -m option it attempts to optimize DWARF
debugging information in each given object (executable or shared library) individually, with the
-m  option  it  afterwards  attempts to optimize even more by moving DWARF debugging information
entries (DIEs), strings and macro descriptions duplicated in more than one object into  a  newly
created  ELF ET_REL object whose filename is given as -m option argument.  The debug sections in
the executables and shared libraries specified on the command  line  are  then  modified  again,
referring to the entities in the newly created object.
```

> This also needs at least a small test.

Any suggestion about what this test should cover?

Thanks!

aarzilli:
> Any suggestion about what this test should cover?

Compile a program, run dwz on it then load the executable with LoadBinaryInfo and check that things that you know dwz deduplicated are accessible. Ideally.

We keep test programs in _fixtures, you can make subdirectories for programs composed of multiple files. The function BuildFixture in pkg/proc/test/support.go takes care of compiling a program saved inside _fixtures for you.

slp:
> Compile a program, run dwz on it then load the executable with LoadBinaryInfo and check that things that you know dwz deduplicated are accessible. Ideally.

Just extended the PR with a test for this, which only runs if 'dwz' binary is present. Please let me know if a different approach would be preferred. Thanks!

pkg/proc/proc_test.go
```diff
@@ -3697,3 +3697,20 @@ func TestIssue951(t *testing.T) {
 		}
 	})
 }
+	
+func TestDWZCompression(t *testing.T) {
+	// If dwz is not available in the system, just ignore this test
+	if _, err := exec.LookPath("dwz"); err != nil {
+		return
```
aarzilli: We are moving towards using Skip instead of returning, so this should be `t.Skip("dwz not installed")`

aarzilli:
Actually it would probably have to be:

```
before_install:
  - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get -qq update; sudo apt-get install -y dwz; fi
```

slp:
I've just took a look at travis output, and it seems like your suggestion to automatically install dwz works. Cool!

aarzilli:
LGTM

PR #1220, opened 2018-05-25
pkg/proc: respect specified load configuration after reslicing a map

aarzilli:
```
pkg/proc: respect specified load configuration after reslicing a map

Maps were always loaded with using the default configuration during a
reslice. This is probably a remnant from when we didn't let clients
configure the load parameters.

```

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3697,7 +3697,7 @@ func TestIssue951(t *testing.T) {
 		}
 	})
 }
-	
+
 func TestDWZCompression(t *testing.T) {
 	// If dwz is not available in the system, skip this test
 	if _, err := exec.LookPath("dwz"); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/proc/proc_test.go
```diff
@@ -3714,3 +3714,28 @@ func TestDWZCompression(t *testing.T) {
...
+		assertNoError(err, t, "EvalVariable")
+		t.Logf("m1 returned children %d (%d)", len(m1.Children)/2, m1.Len)
+
+		expr := fmt.Sprintf("(*(*%q)(%d))[10:]", m1.DwarfType.String(), m1.Addr)
+		t.Logf("expr %q\n", expr)
+
+		m1cont, err := scope.EvalExpression(expr, zolotovLoadCfg)
+		assertNoError(err, t, "EvalVariable")
+
+		t.Logf("m1cont returned children %d", len(m1cont.Children)/2)
```
derekparker: Why log the length of the children divided by 2 here (and above) but assert the full value below?

aarzilli: The full value would be 10 (the value of MaxArrayValues). But I forgot to divide by 2 in the call to Fatal.

PR #1224, opened 2018-06-05
proc: add support for separate debug info

### LINKED ISSUE #1206: Allow loading of separate debuginfo files
Right now Delve expects all DWARF debug info to be in the binary being executed. That isn't always the case as many Linux distributions ship the debug info files in their own packages which are installed separately.

slp:
To save disk space, some distributions strip the debugging information
from the binaries, putting it in separate files, usually distributed in
separate packages.

To locate the file containing the debug information for a certain
binary, an ELF note named ".note.gnu.build-id" is added to the latter,
which contains a header and a build identification. This identification
can be used to compose a path with this form:

/usr/lib/debug/.build-id/BUILDID[:2]/BUILDID[2:].debug

With this patch, if Delve can't find the debug information in the main
binary, it'll try to locate and parse ".note.gnu.build-id", to compose
and attempt to open a path with the format described above.

Fixes #1206

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -7,6 +7,7 @@ import (
 	"debug/macho"
 	"debug/pe"
 	"encoding/binary"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -27,8 +28,9 @@ import (
 type BinaryInfo struct {
 	lastModified time.Time // Time the executable of this process was last modified
 
-	GOOS   string
-	closer io.Closer
+	GOOS           string
+	closer         io.Closer
+	sepDebugCloser io.Closer
 
 	// Maps package names to package paths, needed to lookup types inside DWARF info
 	packageMap map[string]string
@@ -244,6 +246,12 @@ func (e *loclistEntry) BaseAddressSelection() bool {
 	return e.lowpc == ^uint64(0)
 }
 
+type buildIdHeader struct {
+	Namesz uint32
+	Descsz uint32
+	Type   uint32
+}
+
 func NewBinaryInfo(goos, goarch string) BinaryInfo {
 	r := BinaryInfo{GOOS: goos, nameOfRuntimeType: make(map[uintptr]nameOfRuntimeTypeEntry), typeCache: make(map[dwarf.Offset]godwarf.Type)}
 
@@ -349,6 +357,9 @@ func (bi *BinaryInfo) PCToFunc(pc uint64) *Function {
 }
 
 func (bi *BinaryInfo) Close() error {
+	if bi.sepDebugCloser != nil {
+		bi.sepDebugCloser.Close()
+	}
 	return bi.closer.Close()
 }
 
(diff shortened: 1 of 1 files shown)


slp:
As the Go compiler doesn't add a ".note.gnu.build-id" note by itself, writing a test for this for would require specialized linking, stripping, and composing of the path. I don't think that's feasible without important changes in the test framework, and that probably is not worth for such a small change.

pkg/proc/bininfo.go
```diff
@@ -442,6 +454,53 @@ func (bi *BinaryInfo) findCompileUnit(pc uint64) *compileUnit {
...
+	descBinary := make([]byte, bh.Descsz)
+	if err := binary.Read(br, binary.LittleEndian, descBinary); err != nil {
+		return nil, nil, errors.New("can't read build-id desc: " + err.Error())
+	}
+	desc := hex.EncodeToString(descBinary)
+
+	debugPath := fmt.Sprintf("/usr/lib/debug/.build-id/%s/%s.debug", desc[:2], desc[2:])
+	sepFile, err := os.OpenFile(debugPath, 0, os.ModePerm)
+	if err != nil {
+		return nil, nil, errors.New("can't open separate debug file: " + err.Error())
```
aarzilli: This error and all the ones that follow should include the path of the debug file (printed as `%q`), so that we can tell for sure which file we are trying to read.

pkg/proc/bininfo.go
```diff
@@ -455,23 +514,33 @@ func (bi *BinaryInfo) LoadBinaryInfoElf(path string, wg *sync.WaitGroup) error {
 	if elfFile.Machine != elf.EM_X86_64 {
 		return UnsupportedLinuxArchErr
 	}
+	debugInfo := elfFile
```
aarzilli: dwarfFile or debugFile instead of debugInfo.

derekparker:
@slp any chance you'll be able to take care of the rebase / feedback?

slp:
@derekparker  Sure, I'll do that ASAP. Probably tomorrow or Monday.

pkg/proc/bininfo.go
```diff
@@ -455,23 +514,33 @@ func (bi *BinaryInfo) LoadBinaryInfoElf(path string, wg *sync.WaitGroup) error {
...
 	bi.dwarf, err = elfFile.DWARF()
 	if err != nil {
-		return err
+		var sepFile *os.File
+		sepFile, debugInfo, err = bi.LoadSeparateDebugInfo(elfFile)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
slp: Hm... I'm a bit lost here, openSeparateDebugInfo already returns a variety of errors here depending on the actual issue dealing with the file.

pkg/proc/bininfo.go
slp: The error coming from os.OpenFile already includes the path. I've added it to the other two.

pkg/proc/bininfo.go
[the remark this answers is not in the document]
aarzilli: Suppose you have an executable with embedded DWARF (no build-id note). Also suppose that calling elfFile.DWARF() returns an error for this file. What you want to return to the user is the error returned from elfFile.DWARF() not the "can't find build-id note" error, the former is informative, the latter is misleading (we only looked for build-id because of the error returned by elfFile.DWARF()).

pkg/proc/bininfo.go
aarzilli: Thanks.

pkg/proc/bininfo.go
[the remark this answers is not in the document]
slp: Now I get it, thanks a lot for the explanation. I've just fixed this one too.

PR #1233, opened 2018-06-14
*: Implement structured logging

derekparker:
Implements structured logging via Logrus. This gives us a logger per
boundry that we care about, allowing for easier parsing of logs if users
have more than one log option enabled. Also, cleans up a lot of
conditionals in the code by simply silencing the logger at creation as
opposed to conditionally logging everywhere.


pkg/dwarf/line/line_parser.go
```diff
@@ -33,8 +36,7 @@ type DebugLineInfo struct {
 	// lastMachineCache[pc] is a state machine stopped at an address after pc
 	lastMachineCache map[uint64]*StateMachine
 
-	// logSuppressedErrors enables logging of otherwise suppressed errors
-	logSuppressedErrors bool
+	log *logrus.Entry
```
aarzilli: IMHO a reusable package like this shouldn't depend on a specific logging library that's not part of the standard library. Especially given the talk of contributing it to x/debug/dwarf.

service/rpccommon/server.go
```diff
@@ -307,10 +296,8 @@ func (s *ServerImpl) serveJSONCodec(conn io.ReadWriteCloser) {
 		}
 
 		if mtype.Synchronous {
-			if s.log {
-				argvbytes, _ := json.Marshal(argv.Interface())
-				log.Printf("-> %s(%T%s)\n", req.ServiceMethod, argv.Interface(), argvbytes)
-			}
+			argvbytes, _ := json.Marshal(argv.Interface())
```
aarzilli: This is a fairly expensive operation, I don't think we should do it unconditionally given that rpc logging will mostly be disabled. Same goes for the other json.Marshal calls in this function and RPCCallback.Return.

pkg/dwarf/line/line_parser.go
```diff
@@ -3,9 +3,12 @@ package line
 import (
 	"bytes"
 	"encoding/binary"
+	"io/ioutil"
 	"path/filepath"
 
 	"github.com/derekparker/delve/pkg/dwarf/util"
+	"github.com/derekparker/delve/pkg/logflags"
+	"github.com/sirupsen/logrus"
```
aarzilli: IMHO a reusable package like this shouldn't depend on a specific logging library that's not part of the standard library. Especially given the talk of contributing it to x/debug/dwarf.

Same goes for depending on logflags.

derekparker: All good points mentioned above, maybe we can change the package to accept a logger function, which will not add any imports.

aarzilli: I think that would be good.

derekparker:
@aarzilli updated.

PR #1234, opened 2018-06-15
*: Show return values of traced functions

derekparker:
This patch enables the trace command to take advantage of the return
value distinction present in the debug info of binaries produced with a
sufficiently new Go compiler.

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -79,6 +79,9 @@ const (
 	// Continue will set a new breakpoint (of NextBreakpoint kind) on the
 	// destination of CALL, delete this breakpoint and then continue again
 	StepBreakpoint
+	// TraceReturnBreakpoint is an internal breakpoint set to get the return
+	// values of a traced function.
+	TraceReturnBreakpoint
 )
 
 func (bp *Breakpoint) String() string {
@@ -372,6 +375,7 @@ func configureReturnBreakpoint(bi *BinaryInfo, bp *Breakpoint, topframe *Stackfr
 	if topframe.Current.Fn == nil {
 		return
 	}
+	bp.Name = fmt.Sprintf("[return values] %d", bp.ID)
 	bp.returnInfo = &returnBreakpointInfo{
 		retFrameCond: retFrameCond,
 		fn:           topframe.Current.Fn,
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -147,6 +147,39 @@ func Continue(dbp Process) error {
 				}
 			}
 			return conditionErrors(threads)
+		case curbp.Active && curbp.Tracepoint:
+			if curbp.returnInfo != nil {
+				curthread.Common().returnValues = curbp.Breakpoint.returnInfo.Collect(curthread)
+				return conditionErrors(threads)
+			}
+			// We have hit a function entry breakpoint, set a
+			// breakpoint at the return address and load return
+			// values.
+			selg := dbp.SelectedGoroutine()
+			topframe, retframe, err := topframe(selg, curthread)
+			if err != nil {
+				return err
+			}
+
+			if topframe.Inlined {
+				if err := next(dbp, false, true); err != nil {
+					return err
+				}
+				return Continue(dbp)
+			}
+
+			sameGCond := SameGoroutineCondition(selg)
+			retFrameCond := andFrameoffCondition(sameGCond, retframe.FrameOffset())
+			bp, err := dbp.SetBreakpoint(topframe.Ret, UserBreakpoint, retFrameCond)
+			if err != nil {
+				if _, isexists := err.(BreakpointExistsError); !isexists {
+					return err
+				}
+			} else {
+				bp.Tracepoint = true
+				configureReturnBreakpoint(dbp.BinInfo(), bp, &topframe, retFrameCond)
+			}
+			return conditionErrors(threads)
 		case curbp.Active && curbp.Internal:
 			if curbp.Kind == StepBreakpoint {
 				// See description of proc.(*Process).next for the meaning of StepBreakpoints
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -645,7 +645,7 @@ func (scope *EvalScope) FunctionArguments(cfg LoadConfig) ([]*Variable, error) {
 		return nil, err
 	}
 	vars = filterVariables(vars, func(v *Variable) bool {
-		return (v.Flags & (VariableArgument | VariableReturnArgument)) != 0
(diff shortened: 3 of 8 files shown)


pkg/proc/breakpoints.go
```diff
@@ -79,6 +79,9 @@ const (
 	// Continue will set a new breakpoint (of NextBreakpoint kind) on the
 	// destination of CALL, delete this breakpoint and then continue again
 	StepBreakpoint
+	// TraceReturnBreakpoint is an internal breakpoint set to get the return
+	// values of a traced function.
+	TraceReturnBreakpoint
```
aarzilli: I don't think you are using this anywhere.

pkg/proc/proc.go
```diff
@@ -147,6 +147,39 @@ func Continue(dbp Process) error {
 				}
 			}
 			return conditionErrors(threads)
+		case curbp.Active && curbp.Tracepoint:
```
aarzilli: This changes the behavior of tracepoints in a backwards incompatible way. It also breaks the `trace` command (not `dlv trace`, the one at the `(dlv)` prompt).

pkg/proc/proc.go
```diff
@@ -147,6 +147,39 @@ func Continue(dbp Process) error {
...
+			topframe, retframe, err := topframe(selg, curthread)
+			if err != nil {
+				return err
+			}
+
+			if topframe.Inlined {
+				if err := next(dbp, false, true); err != nil {
+					return err
+				}
+				return Continue(dbp)
```
aarzilli: I think it would be better to just let Continue repeat its loop again, instead of calling it recursively. It's what every other case is doing.

pkg/proc/proc.go
```diff
@@ -147,6 +147,39 @@ func Continue(dbp Process) error {
...
+			}
+
+			sameGCond := SameGoroutineCondition(selg)
```
aarzilli: I think there is a problem with concurrency here. Two goroutines can call the same function simultaneously (or one goroutine calls a function, then a second goroutine calls the same function before the first goroutine has returned from it). In that case only the return value from the first function would be output.

Removing this condition doesn't work either, the CALL that entered the function call could be a dynamic dispatch.

IMHO the best to do this is to disassemble every function we want traced and set breakpoints on every `RET` and `CALL runtime.deferreturn` instruction. It's also simpler and takes care of the problem with redefining the meaning of  Breakpoint.Tracepoint and having a recursive call inside Continue.

pkg/proc/variables.go
```diff
@@ -645,7 +645,7 @@ func (scope *EvalScope) FunctionArguments(cfg LoadConfig) ([]*Variable, error) {
 		return nil, err
 	}
 	vars = filterVariables(vars, func(v *Variable) bool {
-		return (v.Flags & (VariableArgument | VariableReturnArgument)) != 0
+		return (v.Flags & VariableArgument) != 0
```
aarzilli: We can't do this because of backwards compatibility: FunctionArguments + Locals == all local variables.

pkg/proc/breakpoints.go
derekparker: Meant to remove this before pushing.

pkg/proc/proc.go
derekparker: We may have to do a major version bump in order to fit this in. I want these improvements to land, but we can't sacrifice backwards compat. 

Also, what breaks at the prompt? Just tested and everything seems to be working.

pkg/proc/proc.go
derekparker: Agreed.

pkg/proc/proc.go
aarzilli: > Also, what breaks at the prompt?

you can set a tracepoint anywhere, not just the header of the function and the expected behavior is to just print something when the breakpoint is hit, no automatic return tracepoint.

pkg/proc/proc.go
aarzilli: > We would still have a similar concurrency issue

I don't think you would, those can be persistent and don't have any condition associated. Matching between entry/exit can be done in a higher layer (although that's very complicated in general, since some calls never exit because of panics).

> If we allowed multiple conditions for a single breakpoint, we could solve the issue of only one return breakpoint triggering, but we'd still have to be smarter about grouping the entry / exit breakpoints.

You'll need some complicated logic to handle recursive functions: you'd have to keep, for each goroutine, a stack of ReturnBreakpointInfo and keep it sync'd with the target process (removing multiple entries when there's a panic). I'm not even sure it's possible. 

Consider this scenario:
Let's say we are tracing a function f and it's called through a function like this:

```
func dynamicdispatch(fn func() int) {
   fn()
}

dynamicdispatch(f)
```

Let's say there's also a second function g that is also called through dynamicdispatch.

```
- target calls dynamicdispatch(f)
- dynamicdispatch calls f
-> we set a return breakpoint in dynamicdispatch
- f calls some other function
- there's a panic
- target unwinds the stack to well before dynamicdispatch
- time passes
- target calls dynamicdispatch(g), it just so happens that the stack depth is exactly the same as before, since we are not tracing g we don't notice
- g returns
-> we are now at the return breakpoint we set before, on the right goroutine, with the right stack depth, so we read the return values of g as if they were the return values of f, since they have the same signature the result even looks plausible.
```

By setting the return breakpoint on the RET/CALL runtime.deferreturn instruction you don't have this problem: the only way to get at them is to go through `f`. To associate entry with exit you can just compare stack depth by reading the FrameOffset.

I understand that using the disassembly is not exactly tasteful, especially now that we're finally getting rid of the disassembly based prologue detector, but this particular case is not as bad as the prologue detector: normally there's only one way functions exit, the RET instructions, go has another way (CALL runtime.deferreturn) but there is no reason to believe that this will change and a third way will appear.

pkg/proc/variables.go
aarzilli: I actually have a list of things I would like to change about the API, including this. I don't think they are compelling enough reasons to warrant a new API version. Goland took months to convert to APIv2, VSCode-go converted last week, after 2 years, and they are still using v1 by default. And the change from v1 to v2 was minimal. If we bump API we're looking at supporting 3 versions of the API for years to come.

The same effect here can be had by adding a single optional flag to ListFunctionArguments, it's not worth upgrading the API for this.

derekparker: I agree with this, let's wait for now.

pkg/proc/proc.go
derekparker: That's true, I'll think on a way to preserve functionality of this while improving the CLI subcommand.

derekparker:
Closing this for now, I will have a new implementation of this up by early next week.

PR #1237, opened 2018-06-18
proc: support GNU compressed debug sections (go1.11 support)

aarzilli:
```
proc: support GNU compressed debug sections (go1.11 support)

Go1.11 switched to the zlib-gnu compression format for debug sections.
Change proc and and a test in dwarf/line to support this change.

Also deletes some dead code from pkg/proc/bininfo.go that hadn't been
used in a long time.

```


aarzilli:
Rebased

PR #1259, opened 2018-07-02
*: Fix log level setting in logrus

aarzilli:
```
*: Fix log level setting in logrus

Setting the Level field of a logrus logger doesn't actually do anything
since the Level field simply reports the log level of the last log
message emitted on the logger.
The right way to do that is to set logger.Logger.Level.
Also cleans up newline characters from log messages emitted through
logrus and fixes the direction of the arrows in the messages emitted by
rpccommon, which was inconsistent with the arrows of gdbserial.

```


aarzilli:
Amazing trap set up by logrus authors. Truly devious.

PR #1276, opened 2018-07-23
dwarf/line: remove foundFile "optimization" from LineToPC

aarzilli:
```
dwarf/line: remove foundFile "optimization" from LineToPC

There is no guarantee that files will end up stored contiguously in the
debug_line section which makes this optimization wrong in the general
case.
In particular with recent versions of go1.11 and a go.mod file present
the go compiler seems to sometimes produce executables that actually
violate this assumption.

```


aarzilli:
It's actually breaking go1.8 :( I'll look into it

aarzilli:
Fixed now.

PR #1283, opened 2018-07-27
cmd/dlv: do not override user's CGO_CFLAGS

### LINKED ISSUE #1278: Cannot use custom CGO_CFLAGS variables when building debug binaries
Version: 1.0.0

My project has an include directory for cgo files set with CGO_CFLAGS. This flag gets blown away when trying to run `dlv debug` and the debug binary build fails.

aarzilli:
```
cmd/dlv: do not override user's CGO_CFLAGS

Fixes #1278

```

### DIFF
--- a/cmd/dlv/main.go
+++ b/cmd/dlv/main.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/derekparker/delve/cmd/dlv/cmds"
 	"github.com/derekparker/delve/pkg/version"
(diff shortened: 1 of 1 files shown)


derekparker:
Looks good, but I think it might also be good to print a warning notifying the user that they have this env set and mention that it might affect the debug experience.

aarzilli:
Good idea.

PR #1284, opened 2018-07-27
Add an edit command

benc153:
Which I miss from gdb-- it's nice to be able to open where you are in
the editor you're already working on the project in.


benc153:
Found my patch was NBG anyway because if you went "up" a few times and then did an "edit" it didn't open you in the frame you're actually now in.

Fixed this by looking a bit more closely at listCommand. Maybe would be better to refactor a bit and share this code. But listCommand needs a few other intermediate results so it's not so easy.

Also maybe it would be better if a Thread was a "subclass" of Location like a Stackframe is. It has all the same stuff in it (and I'm just copying it over at line 1109)

benc153:
Latest version didn't actually build (the params to Stacktrace have changed). This should be OK.

benc153:
Build didn't work. Should be OK now.

pkg/terminal/command.go
```diff
@@ -1132,6 +1135,56 @@ func tracepoint(t *Term, ctx callContext, args string) error {
...
+	}
+}
+
+func edit(t *Term, ctx callContext, args string) error {
+	loc, err := getLocation(t, ctx)
+	if err != nil { return err }
+
+	var editor string
+
+	editor = os.Getenv("DELVE_EDITOR")
```
derekparker: nit: Just use an `:=` here instead of declaring the variable above.

derekparker: Either that or declare and cleanup below with something like:

```
var editor string
if editor = os.Getenv("DELVE_EDITOR"); editor == "" {
        if editor = os.Getenv("EDITOR"); editor == "" {
                return <err>
        }
}
```

derekparker:
Thanks for the patch, just a small refactor nit, otherwise lgtm as well.

pkg/terminal/command.go
benc153: Yes good point. I started with a := but had another one in the if block which wasn't right and got confused. The declare and cleanup thing looks quite Gothonic (?) I'll have to remember that one :)

PR #1287, opened 2018-07-31
Add an edit command

benc153:
Which I miss from gdb-- it's nice to be able to open where you are in
your editor where you're already working on the project you're
debugging.


aarzilli:
I imagine this was opened accidentally.

benc153:
Yes. I didn't know whether I was supposed to open a pull request every time I amended my patch or not... So I did, which means there might be a few more bogus pull requests lying around.

PR #1289, opened 2018-08-01
terminal: refactor edit/list to share getLocation code

aarzilli:
```
terminal: refactor edit/list to share getLocation code

Refactors edit and list so that they use the same code to get the
current location and also both accept a locspec as an argument.

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -321,7 +321,11 @@ Adds or removes a path substitution rule.
 
 Defines <alias> as an alias to <command> or removes an alias.`},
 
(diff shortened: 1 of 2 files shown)


pkg/terminal/command.go
```diff
@@ -1408,53 +1382,69 @@ func parseStackArgs(argstr string) (stackArgs, error) {
 	return r, nil
 }
 
-func listCommand(t *Term, ctx callContext, args string) error {
+// getLocation returns the current location or the locations specified by the argument.
+// getLocation is used to process the argument of the list and edit.
```
derekparker: nit: Either add "commands" to the end of the sentence or remove "the".

aarzilli:
Done.

PR #1293, opened 2018-08-06
Use setValue in CallFunction

aarzilli:
```
proc: use (*Variable).setValue in fncall

proc: change (*Variable).setValue for use in CallFunction

Changes (*Variable).setValue so that it can be used in CallFunction to
set up the argument frame for the function call, adding the ability to:
- nil nillable types
- set strings to the empty string
- copy from one structure to another (including strings and slices)
- convert any interface type to interface{}
- convert pointer shaped types (map, chan, pointers, and structs
consisting of a single pointer field) to interface{}

This covers all cases where an assignment statement can be evaluated
without allocating memory or calling functions in the target process.

```

### DIFF
--- a/pkg/proc/mem.go
+++ b/pkg/proc/mem.go
@@ -136,3 +136,18 @@ func DereferenceMemory(mem MemoryReadWriter) MemoryReadWriter {
 	}
 	return mem
 }
+
+// bufferMemoryReadWriter is dummy a MemoryReadWriter backed by a []byte.
+type bufferMemoryReadWriter struct {
+	buf []byte
+}
+
+func (mem *bufferMemoryReadWriter) ReadMemory(buf []byte, addr uintptr) (n int, err error) {
+	copy(buf, mem.buf[addr-fakeAddress:][:len(buf)])
+	return len(buf), nil
+}
+
+func (mem *bufferMemoryReadWriter) WriteMemory(addr uintptr, data []byte) (written int, err error) {
+	copy(mem.buf[addr-fakeAddress:], data)
+	return len(data), nil
+}
--- a/pkg/proc/test/support.go
+++ b/pkg/proc/test/support.go
@@ -243,3 +243,14 @@ func SafeRemoveAll(dir string) {
 	}
 	os.Remove(dir)
 }
+
+// MustSupportFunctionCalls skips this test if function calls are
+// unsupported on this backend/architecture pair.
+func MustSupportFunctionCalls(t *testing.T, testBackend string) {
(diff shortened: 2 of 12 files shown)


pkg/proc/mem.go
```diff
@@ -136,3 +136,18 @@ func DereferenceMemory(mem MemoryReadWriter) MemoryReadWriter {
 	}
 	return mem
 }
+
+// bufferMemoryReadWriter is dummy a MemoryReadWriter backed by a []byte.
+type bufferMemoryReadWriter struct {
```
derekparker: This is fine for now, but this makes me think this would be better conforming to the interfaces in the `io` package so we don't have to re-invent `bytes.Buffer`.

pkg/proc/mem.go
```diff
@@ -136,3 +136,18 @@ func DereferenceMemory(mem MemoryReadWriter) MemoryReadWriter {
...
+	buf []byte
+}
+
+func (mem *bufferMemoryReadWriter) ReadMemory(buf []byte, addr uintptr) (n int, err error) {
+	copy(buf, mem.buf[addr-fakeAddress:][:len(buf)])
```
derekparker: Couldn't this be written as `mem.buf[addr-fakeAddress:len(buf)+1]`?

pkg/proc/test/support.go
```diff
@@ -243,3 +243,11 @@ func SafeRemoveAll(dir string) {
 	}
 	os.Remove(dir)
 }
+
+// MustSupportFunctionCalls skips this test if function calls are
+// unsupported on this backend/architecture pair.
+func MustSupportFunctionCalls(t *testing.T, testBackend string) {
```
derekparker: Should also check for Go runtime version.

pkg/proc/variables.go
```diff
@@ -998,33 +988,109 @@ func (v *Variable) loadValueInternal(recurseLevel int, cfg LoadConfig) {
...
+	if srcv.Addr != 0 {
+		return dstv.writeCopy(srcv)
+	}
+
+	return fmt.Errorf("can not set variables of type %s (not implemented)", dstv.Kind.String())
+}
+
+// convertToEface converts srcv into an "interface {}" and writes it to
+// dstv.
+// Dstv must be a variable of type "inteface {}" and srcv must either by an
```
derekparker: s/by/be/

derekparker:
Done with first pass.

pkg/proc/mem.go
aarzilli: That would be wrong, it would have to be `mem.buf [addr-fakeAddress:addr-fakeAddres+len(buf)]`

PR #1296, opened 2018-08-07
proc: Increase inline function support

### LINKED ISSUE #1261: unable to process inlined functions (go1.11)
<pre>
package main

import "fmt"

func main() {
        x := &X{Foo: 10}          // line 6
        y := InlineThis(x, "bar") // line 7
        fmt.Printf("Y = %+v\n", y)
}

type X struct {
        Foo int
}

func InlineThis(x *X, s string) *X {
        xx := &X{}
        xx.Foo = x.Foo + len(s) // line 17
        return xx
}
</pre>

I expect dlv to do the right thing with optimized binaries when compiled with go1.11
but currently it is still unhappy about inlined functions.

 - funcs command doesn't recognize the inlined functions
 - can't set break point to the line where the inlined function is called (line 7).
 
I could set break points inside the inlined function though (line 16 or line 17).
'bt' from there (inside the inlined func) seemed reasonable.

<pre>
$ go build main.go
$ dlv exec ./main
(dlv) funcs InlineThis
(dlv) b main.go:6
Breakpoint 1 set at 0x485f21 for main.main() ./main.go:6
(dlv) b main.go:7
Command failed: could not find /usr/local/google/home/hakim/go/src/s/main.go:7
(dlv) b main.go:8
Breakpoint 2 set at 0x485f4b for main.main() ./main.go:8
(dlv) b main.go:17
Breakpoint 3 set at 0x485f3f for main.main() ./main.go:17
</pre>

The versions of go, dlv I used.

<pre>
$ go version
go version devel +23ce272bb1 Mon Jul 2 17:50:00 2018 +0000 linux/amd64
$ dlv version
Delve Debugger
Version: 1.0.0
Build: bdcbd8a846eb7d4953449725ef6a23959198acf1
</pre>

derekparker:
This patch makes it so inlined functions are returned in the
function
list, and also allows users to set breakpoints on the call site of
inlined functions.

Fixes #1261

### DIFF
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -195,6 +195,8 @@ func (bi *BinaryInfo) loadDebugInfoMaps(debugLineBytes []byte, wg *sync.WaitGrou
 	var cu *compileUnit = nil
 	var pu *partialUnit = nil
 	var partialUnits = make(map[dwarf.Offset]*partialUnit)
+	abstractOriginNameTable := make(map[dwarf.Offset]string)
+outer:
 	for entry, err := reader.Next(); entry != nil; entry, err = reader.Next() {
 		if err != nil {
 			break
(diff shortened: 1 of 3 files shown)


pkg/proc/types.go
```diff
@@ -365,35 +367,75 @@ func (bi *BinaryInfo) loadDebugInfoMaps(debugLineBytes []byte, wg *sync.WaitGrou
...
+					entry, err = reader.Next()
+					if err != nil {
+						break outer
+					}
+					if entry.Tag == 0 {
+						break
+					}
+					if entry.Tag == dwarf.TagInlinedSubroutine {
+						originOffset := entry.Val(dwarf.AttrAbstractOrigin).(dwarf.Offset)
+						name := abstractOriginNameTable[originOffset]
```
aarzilli: I'm worried that an inlined function will appear before its abstract entry and this won't find anything. I don't think there's anything preventing this in the standard.

pkg/proc/types.go
```diff
@@ -365,35 +367,75 @@ func (bi *BinaryInfo) loadDebugInfoMaps(debugLineBytes []byte, wg *sync.WaitGrou
...
+						if ranges, _ := bi.dwarf.Ranges(entry); len(ranges) == 1 {
+							ok1 = true
+							lowpc = ranges[0][0]
+							highpc = ranges[0][1]
+						}
+						callfileidx := entry.Val(dwarf.AttrCallFile).(int64)
```
aarzilli: both DW_AT_call_file and DW_AT_call_line are optional, go always emits them but I think we should check if they are there and skip this entry if they aren't.

pkg/proc/types.go
derekparker: I understand that concern but it seems unlikely. The concrete instance refers to the offset of it's abstract root, meaning that offset must already be known. Seems inefficient to make another pass to fill in the offset after if the abstract root is written later.

derekparker:
Updated.

PR #1312, opened 2018-08-17
proc: fix type of some struct global variables

aarzilli:
```
proc: fix type of some struct global variables

Normally variables that have a named struct as a type will get a
typedef entry as their type, sometimes however the Go linker will
decide to use the DW_TAG_structure_type entry instead.

For consistency always wrap a struct type into a typedef when we are
creating a new variables (see comment in newVariable for exceptions).

This fixes a bug where it would be impossible to call methods on a
global variable.

```

### DIFF
--- a/_fixtures/fncall.go
+++ b/_fixtures/fncall.go
@@ -65,6 +65,8 @@ func makeclos(pa *astruct) func(int) string {
 	}
 }
 
+var ga astruct
+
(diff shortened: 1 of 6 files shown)


aarzilli:
cc @heschik this should fix the bug that you found about calling methods on global variables.

PR #1313, opened 2018-08-18
debugger: make yama error more verbose

aarzilli:
```
debugger: make yama error more verbose

Explain why we are asking the user to write to a /proc file.

```


rogpeppe:
Much clearer, thank you!

PR #1326, opened 2018-08-29
proc: replace SavedRegisters interface with a Copy method

aarzilli:
```
proc: replace SavedRegisters interface with a Copy method

Fncall.go was written with the assumption that the object returned by
proc.Thread.Registers does not change after we call
proc.Thread.SetPC/etc.

This is true for the native backend but not for gdbserial. I had
anticipated this problem and introduced the Save/SavedRegisters
mechanism during the first implementation of fncall.go but that's
insufficient.

Instead:

1. clarify that the object returned by proc.Thread.Registers could
change when the CPU registers are modified.
2. add a Copy method to Registers that returns a copy of the registers
that are guaranteed not to change when the CPU registers change.
3. remove the Save/SavedRegisters mechanism.

This solution leaves us the option, in the future, to cache the output
of proc.(Thread).Registers, avoiding a system call every time it's
called.

[WIP] Add function call support for OSX

Implements missing functionality in gdbserial to enable function calls
on OSX.

```

### DIFF
--- a/pkg/proc/native/registers_linux_amd64.go
+++ b/pkg/proc/native/registers_linux_amd64.go
@@ -324,14 +324,6 @@ func (thread *Thread) fpRegisters() (regs []proc.Register, fpregs proc.LinuxX86X
 	return
 }
 
-type savedRegisters struct {
-	regs   sys.PtraceRegs
-	fpregs proc.LinuxX86Xstate
-}
-
-func (r *Regs) Save() proc.SavedRegisters {
-	savedRegs := &savedRegisters{}
-	savedRegs.regs = *r.regs
-	savedRegs.fpregs = *r.fpregset
-	return savedRegs
+func (r *Regs) Copy() proc.Registers {
+	return r
 }
--- a/pkg/proc/native/registers_linux_amd64.go
+++ b/pkg/proc/native/registers_linux_amd64.go
@@ -325,5 +325,16 @@ func (thread *Thread) fpRegisters() (regs []proc.Register, fpregs proc.LinuxX86X
 }
 
 func (r *Regs) Copy() proc.Registers {
-	return r
+	var rr Regs
+	rr.regs = &sys.PtraceRegs{}
+	rr.fpregset = &proc.LinuxX86Xstate{}
+	*(rr.regs) = *(r.regs)
+	if r.fpregset != nil {
+		*(rr.fpregset) = *(r.fpregset)
+	}
+	if r.fpregs != nil {
+		rr.fpregs = make([]proc.Register, len(r.fpregs))
+		copy(rr.fpregs, r.fpregs)
+	}
+	return &rr
 }
--- a/pkg/proc/native/registers_windows_amd64.go
+++ b/pkg/proc/native/registers_windows_amd64.go
@@ -378,10 +378,7 @@ func registers(thread *Thread, floatingPoint bool) (proc.Registers, error) {
 	return regs, nil
 }
 
-type savedRegisters struct {
-}
-
-func (r *Regs) Save() proc.SavedRegisters {
+func (r *Regs) Copy() proc.Registers {
 	//TODO(aarzilli): implement this to support function calls
 	return nil
 }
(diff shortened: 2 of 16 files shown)


pkg/proc/registers.go
```diff
@@ -24,8 +24,9 @@ type Registers interface {
 	GAddr() (uint64, bool)
 	Get(int) (uint64, error)
 	Slice() []Register
-	// Save saves a copy of this object that will survive restarts
-	Save() SavedRegisters
+	// Copy returns a copy of this registers that is guaranteed not to change
```
derekparker: s/copy of this/copy of the/

pkg/proc/native/registers_linux_amd64.go
```diff
@@ -324,14 +324,6 @@ func (thread *Thread) fpRegisters() (regs []proc.Register, fpregs proc.LinuxX86X
...
-	regs   sys.PtraceRegs
-	fpregs proc.LinuxX86Xstate
-}
-
-func (r *Regs) Save() proc.SavedRegisters {
-	savedRegs := &savedRegisters{}
-	savedRegs.regs = *r.regs
-	savedRegs.fpregs = *r.fpregset
-	return savedRegs
+func (r *Regs) Copy() proc.Registers {
```
derekparker: We're just returning the same pointer to the receiver, this breaks the guarantee and we should return an actual copy.

pkg/proc/native/registers_windows_amd64.go
```diff
@@ -378,10 +378,7 @@ func registers(thread *Thread, floatingPoint bool) (proc.Registers, error) {
 	return regs, nil
 }
 
-type savedRegisters struct {
-}
-
-func (r *Regs) Save() proc.SavedRegisters {
+func (r *Regs) Copy() proc.Registers {
```
derekparker: We should just go ahead and implement this for Windows as well, get it all done in this pass.

pkg/proc/native/threads_windows.go
```diff
@@ -150,6 +150,6 @@ func (t *Thread) ReadMemory(buf []byte, addr uintptr) (int, error) {
 	return int(count), err
 }
 
-func (t *Thread) restoreRegisters(sr *savedRegisters) error {
+func (t *Thread) restoreRegisters(savedRegs proc.Registers) error {
```
derekparker: We should also just implement this in the PR and have full support for function calls across platforms.

derekparker:
Ping @dr2chase @heschik since we discussed this yesterday in person.

I started work on this yesterday after the conf and @aarzilli finished it up. This implements function calls on OSX via the gdbserial protocol. 

@aarzilli can you amend my commit to remove [WIP] (but leave author and rest of commit info the same).

pkg/proc/native/registers_linux_amd64.go
aarzilli: It's immutable.

derekparker: Then why have a copy in the first place in the initial `Save()` implementation? Also, how is this immutable, we're returning the receiver pointer directly.

aarzilli: > Then why have a copy in the first place
For (1) compatibility with gdbserial where the instance does change and (2) so that in the future we can cache the Regs object

> how is this immutable

The proc.Registers interface doesn't have any methods that mutate the object.

derekparker: Right, but the whole point of `Copy` is that the object won't get modified at all including out from underneath you. Just because the interface cannot mutate doesn't mean anything holding onto the original won't mutate it.

aarzilli: I don't think it matters at this stage since we aren't mutating the object from inside proc/native either, but I don't want to fight this so I changed it to make a copy.

pkg/proc/test/support.go
```diff
@@ -250,7 +250,19 @@ func MustSupportFunctionCalls(t *testing.T, testBackend string) {
...
+	if testBackend == "rr" {
+		t.Skip(unsupported)
+	}
+
+	switch runtime.GOOS {
+	case "darwin":
+		if testBackend == "native" {
+			t.Skip(unsupported)
+		}
+	case "windows":
```
derekparker: Since we support Windows now we can remove this.

PR #1329, opened 2018-09-01
*: Misc refactors, and doc additions

derekparker:
Refactors some code, adds a bunch of docstrings and just generally fixes
a bunch of linter complaints.

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -34,16 +34,20 @@ const (
 	maxFramePrefetchSize = 1 * 1024 * 1024 // Maximum prefetch size for a stack frame
 )
 
-type FloatSpecial uint8
+type floatSpecial uint8
 
 const (
-	FloatIsNormal FloatSpecial = iota
+	// FloatIsNormal means the value is a normal float.
+	FloatIsNormal floatSpecial = iota
+	// FloatIsNaN means the float is a special NaN value.
 	FloatIsNaN
+	// FloatIsPosInf means the float is a special positive inifitiy value.
 	FloatIsPosInf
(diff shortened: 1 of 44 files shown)


pkg/proc/variables.go
```diff
@@ -1026,13 +1031,13 @@ func (v *Variable) loadValueInternal(recurseLevel int, cfg LoadConfig) {
 //   is performed.
 // * If srcv and dstv have the same type and are both addressable then the
 //   contents of srcv are copied byte-by-byte into dstv
-func (dstv *Variable) setValue(srcv *Variable, srcExpr string) error {
+func (v *Variable) setValue(srcv *Variable, srcExpr string) error {
```
aarzilli: I would prefer that this remained `dstv` for symmetry with `srcv`.

service/debugger/debugger.go
```diff
@@ -402,6 +405,7 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 	return copyBreakpointInfo(original, amend)
 }
 
+// CancelNext will clear internal breakpoints, thus cancelling the 'next' operation.
```
aarzilli: I'd say "cancelling the 'next', 'step' or 'stepout' operation"

pkg/proc/variables.go
derekparker: I'll change if you feel strongly, but I like the consistency of receiver names for these methods.

aarzilli: ok, leave it

derekparker:
@aarzilli done and updated.

PR #1333, opened 2018-09-03
cmd,service: in non-headless mode use an in-memory connection

### LINKED ISSUE #1332: Unauthenticated TCP API
The Delve debugger exposes an API to fully control the debugged process. This API is used both by the headless mode and the interactive mode. The API is exposed as an unauthenticated TCP server. This implies that anybody that can connect to the TCP port of the API has full control of the debugger and the process, and can possibly execute arbitrary code.

All this is probably clear to somebody running remote debugging, but a user running `dlv debug` or presses F5 in VS code, may be unaware that other users and processes on the system can hijack the debugger.

#861 discusses the issue for remote debugging, but the same issue and mitigation could apply to local debugging as well.

---

1. What version of Delve are you using (`dlv version`)?: Version 1.1.0
2. What version of Go are you using? (`go version`)?: go1.11
3. What operating system and processor architecture are you using? Windows 10, CentOS, Ubuntu
4. What did you do? `dlv debug`
5. What did you expect to see? `(dlv)` prompt
6. What did you see instead? 

```client.go:29: dialing:dial tcp 127.0.0.1:23456: connectex: No connection could be made because the target machine actively refused it.```

and then somebody took control of my program.

### PR DESCRIPTION
**Pull request by aarzilli**
```
cmd,service: in non-headless mode use an in-memory connection

Replace the socket connection with an in-memory connection (created by net.Pipe) for non-headless uses of delve.
This is faster and more secure.

Fixes #1332

```

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -317,6 +317,13 @@ func traceCmd(cmd *cobra.Command, args []string) {
 			return 1
 		}
 
+		if Headless {
+			fmt.Fprintf(os.Stderr, "Warning: headless mode not supported with trace\n")
+		}
+		if AcceptMulti {
+			fmt.Fprintf(os.Stderr, "Warning: accept multiclient mode not supported with trace")
+		}
+
 		debugname, err := filepath.Abs(cmd.Flag("output").Value.String())
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "%v\n", err)
@@ -343,12 +350,9 @@ func traceCmd(cmd *cobra.Command, args []string) {
 
 			processArgs = append([]string{debugname}, targetArgs...)
 		}
-		// Make a TCP listener
-		listener, err := net.Listen("tcp", Addr)
-		if err != nil {
-			fmt.Printf("couldn't start listener: %s\n", err)
-			return 1
-		}
+
+		// Make a local in-memory connection that client and server use to communicate
+		listener, clientConn := service.ListenerPipe()
 		defer listener.Close()
 
 		// Create and start a debug server
@@ -364,7 +368,7 @@ func traceCmd(cmd *cobra.Command, args []string) {
 			fmt.Fprintln(os.Stderr, err)
 			return 1
 		}
-		client := rpc2.NewClient(listener.Addr().String())
+		client := rpc2.NewClientFromConn(clientConn)
 		funcs, err := client.ListFunctions(regexp)
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
@@ -430,7 +434,7 @@ func connectCmd(cmd *cobra.Command, args []string) {
 		fmt.Fprint(os.Stderr, "An empty address was provided. You must provide an address as the first argument.\n")
 		os.Exit(1)
 	}
-	os.Exit(connect(addr, conf, executingOther))
+	os.Exit(connect(addr, nil, conf, executingOther))
 }
 
 func splitArgs(cmd *cobra.Command, args []string) ([]string, []string) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: service/listenerpipe.go [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"errors"
+	"net"
+	"sync"
+)
+
+// ListenerPipe, like net.Pipe, returns a full-duplex in-memory connection.
+// One end of the connection is returned as a net.Conn interface.
+// The other end is returned as a listener that will return it  when Accept
```
derekparker: nit: This comment needs a little rewording.

aarzilli: Is this better or do you have something else in mind?

cmd/dlv/cmds/commands.go
```diff
@@ -493,6 +494,23 @@ func execute(attachPid int, processArgs []string, conf *config.Config, coreFile
...
+	var listener net.Listener
+	var clientConn net.Conn
+	var err error
+
+	// Make a TCP listener
+	if Headless {
+		listener, err = net.Listen("tcp", Addr)
+	} else {
+		listener, clientConn = service.ListenerPipe()
+		err = nil
```
derekparker: This is the default value, no need to set explicitly here.

service/listenerpipe.go
```diff
@@ -0,0 +1,53 @@
...
+	closeMu sync.Mutex
+}
+
+func (l *preconnectedListener) Accept() (net.Conn, error) {
+	if !l.done {
+		l.done = true
+		return l.conn, nil
+	}
+	<-l.closech
+	return nil, errors.New("closed")
```
derekparker: Let's return a more descriptive error here. Something like: `accept failed: listener closed`.

service/listenerpipe.go
```diff
@@ -0,0 +1,53 @@
+package service
```
derekparker: Add godoc comments to all exported symbols, please.

service/listenerpipe.go
derekparker: Alternatively we could create and return an `OpError` which is used in the net package: https://golang.org/pkg/net/#OpError. It may not be as useful since this is all local anyways, but would be consistent.

service/listenerpipe.go
```diff
@@ -0,0 +1,53 @@
...
+// will block until the listener is closed.
+type preconnectedListener struct {
+	done    bool
+	conn    net.Conn
+	closech chan struct{}
+	closeMu sync.Mutex
+}
+
+func (l *preconnectedListener) Accept() (net.Conn, error) {
+	if !l.done {
```
derekparker: We should enclose this in a mutex to prevent a race as well. Since we block anyways we can still just defer the unlock.

service/listenerpipe.go
```diff
@@ -0,0 +1,53 @@
...
+	conn0, conn1 := net.Pipe()
+	return &preconnectedListener{conn: conn0, closech: make(chan struct{})}, conn1
+}
+
+// preconnectedListener satisfies the net.Listener interface by accepting a
+// single pre-established connection.
+// The first call to Accept will return the conn field, any subsequent call
+// will block until the listener is closed.
+type preconnectedListener struct {
+	done    bool
```
derekparker: nit: let's change this to `accepted` or `hasExcepted`.

service/listenerpipe.go
aarzilli: I think we would need to fake too many fields, there's no network type and no address.

service/listenerpipe.go
```diff
@@ -0,0 +1,61 @@
...
+// Accept returns the pre-established connection the first time it's called,
+// it blocks until the listener is closed on every subsequent call.
+func (l *preconnectedListener) Accept() (net.Conn, error) {
+	l.acceptMu.Lock()
+	if !l.accepted {
+		l.accepted = true
+		l.acceptMu.Unlock()
+		return l.conn, nil
+	}
+	l.acceptMu.Unlock()
```
derekparker: I'd say let's just defer the mutex unlock right after the lock. We block in the function anyways so blocking on a mutex instead of a channel read shouldn't make any difference, and will clean up the code a bit.

aarzilli:
Done.

PR #1335, opened 2018-09-03
proc: add flag to disable escape checking in function calls

aarzilli:
```
proc: add flag to disable escape checking in function calls

Fix escape checking in function calls  and add a flag to disable it.

```

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -73,6 +73,8 @@ Aliases: bp
 ## call
 Resumes process, injecting a function call (EXPERIMENTAL!!!)
 	
+	call [-unsafe] <function call expression>
+	
 Current limitations:
 - only pointers to stack-allocated objects can be passed as argument.
 - only some automatic type conversions are supported.
@@ -174,8 +176,8 @@ Aliases: disass
 ## down
 Move the current frame down.
 
-  down [<m>]
-  down [<m>] <command>
+	down [<m>]
+	down [<m>] <command>
 
 Move the current frame down by <m>. The second form runs the command on the given frame.
 
@@ -201,8 +203,8 @@ Aliases: quit q
 ## frame
 Set the current frame, or execute command on a different frame.
 
-  frame <m>
-  frame <m> <command>
+	frame <m>
+	frame <m> <command>
 
 The first form sets frame used by subsequent commands such as "print" or "set".
 The second form runs the command on the given frame.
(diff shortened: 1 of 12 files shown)


Documentation/cli/README.md
```diff
@@ -73,6 +73,8 @@ Aliases: bp
 ## call
 Resumes process, injecting a function call (EXPERIMENTAL!!!)
 	
+  call [-unsafe] <function call expression>
```
chainhelen: Typo
image

image

pkg/proc/eval.go
```diff
@@ -1183,7 +1183,8 @@ func (err *typeConvErr) Error() string {
 
 func (v *Variable) isType(typ godwarf.Type, kind reflect.Kind) error {
 	if v.DwarfType != nil {
-		if typ != nil && typ.String() != v.RealType.String() {
+		if typ == nil || !sameType(typ, v.RealType) {
+			fmt.Printf("%q %q\n", typ.String(), v.RealType.String()) //XXX
```
derekparker: What's the comment here?

pkg/proc/fncall.go
```diff
@@ -278,9 +278,11 @@ func funcCallArgFrame(fn *Function, actualArgs []*Variable, g *G, bi *BinaryInfo
 		formalArg := &formalArgs[i]
 		actualArg := actualArgs[i]
 
-		//TODO(aarzilli): only apply the escapeCheck to leaking parameters.
-		if err := escapeCheck(actualArg, formalArg.name, g); err != nil {
-			return nil, fmt.Errorf("can not pass %s to %s: %v", actualArg.Name, formalArg.name, err)
+		if checkEscape {
+			//TODO(aarzilli): only apply the escapeCheck to leaking parameters.
+			if err := escapeCheck(actualArg, formalArg.name, g); err != nil {
+				return nil, fmt.Errorf("can not pass %s to %s: %v", actualArg.Name, formalArg.name, err)
```
derekparker: Since we're here, let's make this error a bit more explicit. Something like: `("cannot use %s as argument %s in function %s: %v", ...)`

derekparker: Or specify the argument number.

service/api/types.go
```diff
@@ -295,6 +295,8 @@ type DebuggerCommand struct {
 	ReturnInfoLoadConfig *LoadConfig
 	// Expr is the expression argument for a Call command
 	Expr string `json:"expr,omitempty"`
+	// If UnsafeCall is set the function call will not be checked for escapes
```
derekparker: nit: start doc string with name of exported symbol, e.g. "UnsafeCall disabled parameter escape checking for function calls".

pkg/proc/eval.go
aarzilli: It means I wanted to delete that line before committing.

aarzilli:
Done.

PR #1348, opened 2018-09-20
pkg/proc/core: add an new error named ErrChangeRegisterCore

chainhelen:
This commit just add an new error when trying to change register
values for core files (output the same infomation).


pkg/proc/core/core.go
```diff
@@ -160,15 +160,20 @@ type Thread struct {
...
+	// ErrShortRead is returned on a short read.
+	ErrShortRead = errors.New("short read")
 
-// ErrContinueCore is returned when trying to continue execution of a core process.
-var ErrContinueCore = errors.New("can not continue execution of core process")
+	// ErrContinueCore is returned when trying to continue execution of a core process.
+	ErrContinueCore = errors.New("can not continue execution of core process")
+
+	// ErrChangeRegisterCore is returned when trying to change register values for core files.
+	ErrChangeRegisterCore = errors.New("not supported")
```
derekparker: While we're at it let's add a more descriptive error message, specifically mentioning this action is not supported in a core process.

pkg/proc/core/core.go
```diff
@@ -160,15 +160,20 @@ type Thread struct {
...
+	// ErrShortRead is returned on a short read.
+	ErrShortRead = errors.New("short read")
 
-// ErrContinueCore is returned when trying to continue execution of a core process.
-var ErrContinueCore = errors.New("can not continue execution of core process")
+	// ErrContinueCore is returned when trying to continue execution of a core process.
+	ErrContinueCore = errors.New("can not continue execution of core process")
+
+	// ErrChangeRegisterCore is returned when trying to change register values for core files.
+	ErrChangeRegisterCore = errors.New("not supported to change register values of core process")
```
derekparker: s/not supported to/can not/

Other than that, looks great!

chainhelen:
Done. Fix and rebase.Thx.

PR #1360, opened 2018-10-01
proc/native,Makefile: allow compiling on macOS without native backend

### LINKED ISSUE #1359: Add macOS_SDK_headers_for_macOS_10.14.pkg to osx install

master branch
1.11
osx 10.14
4. What did you do?
$ go get -u github.com/derekparker/delve/cmd/dlv
No error
6. What did you see instead?
```sh
# github.com/derekparker/delve/pkg/proc/native
In file included from Devel/Devel_go/src/github.com/derekparker/delve/pkg/proc/native/proc_darwin.go:3:
./proc_darwin.h:1:10: fatal error: 'sys/types.h' file not found
#include <sys/types.h>
         ^~~~~~~~~~~~~
1 error generated.
```

To resolve it, i needed to install /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg manually. (See https://forums.developer.apple.com/thread/104296)

aarzilli:
```
On macOS 10.14 Apple changed the command line tools so that system
headers now need to be manually installed.

Instead of adding one extra install step to the install procedure add a
build tag to allow compilation of delve without the native backend on
macOS. By default (i.e. when using `go get`) this is how delve will be
compiled on macOS, the make script is changed to enable compiling the
native backend if the required dependencies have been installed.

Insure that both configuration still build correctly on Travis CI and
change the documentation to describe how to compile the native backend
and that it isn't normally needed.

Fixes #1359
```

### DIFF
--- a/Documentation/installation/osx/install.md
+++ b/Documentation/installation/osx/install.md
@@ -1,9 +1,5 @@
 # Installation on OSX
 
-Please use the following steps to build and install Delve on OSX.
-
-## Manual install
-
 Ensure you have a proper compilation toolchain.
 
 This should be as simple as:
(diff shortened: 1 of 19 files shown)


scripts/make.go
```diff
@@ -46,9 +50,13 @@ func NewMakeCommands() *cobra.Command {
 		Use:   "install",
 		Short: "Installs delve",
 		Run: func(cmd *cobra.Command, args []string) {
-			checkCertCmd(nil, nil)
-			execute("go", "install", buildFlags(), DelveMainPackagePath)
-			if runtime.GOOS == "darwin" {
+			tagFlag := ""
```
derekparker: Let's try and remove code duplication here.

pkg/proc/native/nonative_darwin.go
```diff
@@ -0,0 +1,108 @@
+//+build darwin,!macnative
```
derekparker: It's tedious, I know, but please add a godoc comment for all exported symbols.

Documentation/installation/osx/install.md
```diff
@@ -16,15 +12,15 @@ Now you can install delve using `go get`:
 $ go get -u github.com/derekparker/delve/cmd/dlv
 ```
 
-With this method you will not be able to use delve's native backend.
+With this method you will not be able to use delve's native backend, *but you don't need it anyway*.
```
derekparker: We should mention that the native backend is unsupported and no longer really maintained.

aarzilli:
Done.

PR #1362, opened 2018-10-04
pkg/config, terminal/terminal.go: Add config (bool) 'source-list-line-bright'

Russtopia:
This change adds a config flag to make the line numbers in the list command either
'dark' (false: blue) or 'bright' (true: bright yellow) as older eyes may have
difficulty with the dark blue ANSI on a Linux white-on-black xterm.
** No new unit tests were added as the feature seemed too trivial to warrant one
  (it depends just on the YAML parser's builtin capabilities)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -43,6 +43,10 @@ type Config struct {
 	// If ShowLocationExpr is true whatis will print the DWARF location
 	// expression for its argument.
 	ShowLocationExpr bool `yaml:"show-location-expr"`
+	
+	// Source list line-number preference: dark (default, blue) or
+	// light (bright yellow)
+	SrcListLineColor bool `yaml:"source-list-line-bright"`
 }
 
 // LoadConfig attempts to populate a Config object from the config.yml file.
@@ -128,6 +132,10 @@ func writeDefaultConfig(f *os.File) error {
 
 # This is the default configuration file. Available options are provided, but disabled.
 # Delete the leading hash mark to enable an item.
+
+# Uncomment the following line to make line numbers in (list) command show bright yellow
+# rather than dark blue.. much better for old eyes in a black terminal.
+# source-list-line-bright: true
 
 # Provided aliases will be added to the default aliases for a given command.
 aliases:
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -19,9 +19,15 @@ import (
 )
 
 const (
-	historyFile             string = ".dbg_history"
-	terminalBlueEscapeCode  string = "\033[34m"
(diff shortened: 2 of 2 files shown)


pkg/config/config.go
```diff
@@ -43,6 +43,10 @@ type Config struct {
 	// If ShowLocationExpr is true whatis will print the DWARF location
 	// expression for its argument.
 	ShowLocationExpr bool `yaml:"show-location-expr"`
+	
+	// Source list line-number preference: dark (default, blue) or
+	// light (bright yellow)
+	SrcListLineColor bool `yaml:"source-list-line-bright"`
```
Russtopia: I didn't spend time thinking of a better name; feel free to change to something more appropriate.

pkg/terminal/terminal.go
```diff
@@ -158,6 +164,10 @@ func (t *Term) Run() (int, error) {
 		return
 	})
 
+	if t.conf.SrcListLineColor {
```
Russtopia: Wasn't sure if this should be in Run() or New(), or somewhere even better...

pkg/config/config.go
aarzilli: At least make the name of the field match the yaml name: `SrcListLineColor` → `SrcListLineBright`.

pkg/config/config.go
```diff
@@ -129,6 +133,10 @@ func writeDefaultConfig(f *os.File) error {
 # This is the default configuration file. Available options are provided, but disabled.
 # Delete the leading hash mark to enable an item.
 
+# Uncomment the following line to make line numbers in (list) command show bright yellow
+# rather than dark blue.. much better for old eyes in a black terminal.
```
aarzilli: remove `. much better for old eyes in a black terminal.`, too much editorializing.

pkg/terminal/terminal.go
```diff
@@ -19,9 +19,15 @@ import (
...
-	terminalResetEscapeCode string = "\033[0m"
+	historyFile                    string = ".dbg_history"
+	terminalBlueEscapeCode         string = "\033[34m"
+	terminalBrightYellowEscapeCode string = "\033[93m"
+	terminalResetEscapeCode        string = "\033[0m"
+)
+
+var (
+	// Controlled by config 'source-list-line-bright' pref
+	terminalHighlightColor = terminalBlueEscapeCode
```
aarzilli: This needs to be a field of the struct `Term`.

pkg/config/config.go
Russtopia: OK! Thanks for the feedback. I will address all those & resubmit ASAP.

PR #1364, opened 2018-10-06
config,terminal: Add source-list-line-color configuration option

Russtopia:
Attempted to address issues in 1st review.

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -19,9 +19,28 @@ import (
 )
 
 const (
-	historyFile             string = ".dbg_history"
-	terminalBlueEscapeCode  string = "\033[34m"
-	terminalResetEscapeCode string = "\033[0m"
+	historyFile                 string = ".dbg_history"
+	terminalHighlightEscapeCode string = "\033[%2dm"
+	terminalResetEscapeCode     string = "\033[0m"
+)
+
+const (
+	ansiBlack     = 30
+	ansiRed       = 31
+	ansiGreen     = 32
+	ansiYellow    = 33
+	ansiBlue      = 34
+	ansiMagenta   = 35
+	ansiCyan      = 36
+	ansiWhite     = 37
+	ansiBrBlack   = 90
+	ansiBrRed     = 91
+	ansiBrGreen   = 92
+	ansiBrYellow  = 93
+	ansiBrBlue    = 94
+	ansiBrMagenta = 95
+	ansiBrCyan    = 96
+	ansiBrWhite   = 97
 )
 
 // Term represents the terminal running dlv.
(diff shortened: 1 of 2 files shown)


pkg/terminal/terminal.go
```diff
@@ -81,14 +100,20 @@ func New(client service.Client, conf *config.Config) *Term {
 		client.SetReturnValuesLoadConfig(&LongLoadConfig)
 	}
 
+	if conf.SourceListLineColor < ansiBlack ||
+	conf.SourceListLineColor > ansiBrWhite {
+		fmt.Fprintf(os.Stderr, "invalid source-list-line-color, defaulting to ansiBlue(34)\n")
```
aarzilli: Remove this Fprintf, so that people who don't set the variable don't get an error message.

pkg/terminal/terminal.go
```diff
@@ -81,14 +100,20 @@ func New(client service.Client, conf *config.Config) *Term {
 		client.SetReturnValuesLoadConfig(&LongLoadConfig)
 	}
 
+	if conf.SourceListLineColor < ansiBlack ||
+	conf.SourceListLineColor > ansiBrWhite {
```
aarzilli: You need to run `go fmt pkg/terminal/terminal.go`, also I'd only check `if conf.SourceListLineColor == 0`.

Russtopia: OK, thanks forgot a final fmt check there.
Are you certain we should only check against 0 though? With no range check, a user could set

source-list-line-color: -1

resulting in an invalid ANSI escape sequence in source list, such as

[-1m    22:		MOVQ	$0, R10
[-1m    23:		MOVQ	$0, R8

..dumb for the user to do that, but I just tested and that's what would occur.

pkg/terminal/terminal.go
```diff
@@ -101,19 +101,18 @@ func New(client service.Client, conf *config.Config) *Term {
 	}
 
 	if conf.SourceListLineColor < ansiBlack ||
-	conf.SourceListLineColor > ansiBrWhite {
-		fmt.Fprintf(os.Stderr, "invalid source-list-line-color, defaulting to ansiBlue(34)\n")
```
Russtopia: Removed fmt.Fprintf() per review comments
Range check against lower/upper valid ansi color values was left, as only testing against 0 will handle user leaving the pref commented out in config.yml, but not a user setting out of range values, eg

source-list-line-color: -1

yields list output
...
[-1m    22:		MOVQ	$0, R10
[-1m    23:		MOVQ	$0, R8
...

pkg/terminal/terminal.go
```diff
@@ -81,6 +100,11 @@ func New(client service.Client, conf *config.Config) *Term {
 		client.SetReturnValuesLoadConfig(&LongLoadConfig)
 	}
 
+	if conf.SourceListLineColor < ansiBlack ||
```
derekparker: This conditional missed possible incorrect configuration values (38..89).

derekparker:
Everything looks good, please just squash to a single, properly formatted commit and we'll get this merged. Thanks!

PR #1372, opened 2018-10-12
pkg/proc: extend conversion about `string` for array/str

chainhelen:
Extend `string()`
1.convert byte/rune array(not only slice) to string.
2.convert string to string(itself), just like `string(str)` in go language.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -114,15 +114,25 @@ func (scope *EvalScope) evalToplevelTypeCast(t ast.Expr, cfg LoadConfig) (*Varia
 
 	case "string":
 		switch argv.Kind {
+		case reflect.String:
+			s := constant.StringVal(argv.Value)
+			v.Value = constant.MakeString(s)
+			v.Len = int64(len(s))
+			return v, nil
 		case reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Int, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uint, reflect.Uintptr:
 			b, _ := constant.Int64Val(argv.Value)
 			s := string(b)
 			v.Value = constant.MakeString(s)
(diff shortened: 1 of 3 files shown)


pkg/proc/eval.go
```diff
@@ -114,15 +114,25 @@ func (scope *EvalScope) evalToplevelTypeCast(t ast.Expr, cfg LoadConfig) (*Varia
...
 			b, _ := constant.Int64Val(argv.Value)
 			s := string(b)
 			v.Value = constant.MakeString(s)
 			v.Len = int64(len(s))
 			return v, nil
-
-		case reflect.Slice:
-			switch elemType := argv.RealType.(*godwarf.SliceType).ElemType.(type) {
+		case reflect.Slice, reflect.Array:
+			var elem interface{}
```
aarzilli: You should be able to use `godwarf.Type` instead of `interface{}` here. I would prefer it because it's more specific.

service/test/variables_test.go
```diff
@@ -784,6 +784,9 @@ func TestEvalExpression(t *testing.T) {
 		{"string(runeslice)", false, `"tèst"`, `""`, "string", nil},
 		{"[]byte(string(runeslice))", false, `[]uint8 len: 5, cap: 5, [116,195,168,115,116]`, `[]uint8 len: 0, cap: 0, nil`, "[]uint8", nil},
 		{"*(*[5]byte)(uintptr(&byteslice[0]))", false, `[5]uint8 [116,195,168,115,116]`, `[5]uint8 [...]`, "[5]uint8", nil},
+		{"string(bytearray)", false, `"tèst"`, `""`, "string", nil},
+		{"string(bytearray)", false, `"tèst"`, `""`, "string", nil},
```
aarzilli: I imagine this was supposed to say `runearray`.

chainhelen:
@aarzilli Appreciate your review. Has fixed and rebase.

PR #1375, opened 2018-10-15
updated the gcflags that needs to be used [documentation]

altimac:
as described here: https://github.com/derekparker/delve/issues/1124#issuecomment-366500430
-gcflags="-N -l" is not enough. -gcflags="all=-N -l" is necessary.

### DIFF
--- a/Documentation/usage/dlv_exec.md
+++ b/Documentation/usage/dlv_exec.md
@@ -10,7 +10,7 @@ Execute a precompiled binary and begin a debug session.
 This command will cause Delve to exec the binary and immediately attach to it to
 begin a new debug session. Please note that if the binary was not compiled with
 optimizations disabled, it may be difficult to properly debug it. Please
-consider compiling debugging binaries with -gcflags="-N -l".
+consider compiling debugging binaries with -gcflags="all=-N -l".
 
 ```
 dlv exec <path/to/binary>


chainhelen:
@altimac  

1. Maybe like below code for version
```
func optflags(args []string) []string {
    // after go1.9 building with -gcflags='-N -l' and -a simultaneously works.
    // after go1.10 specifying -a is unnecessary because of the new caching strategy, but we should pass -gcflags=all=-N -l to have it applied to all packages
    // see https://github.com/golang/go/commit/5993251c015dfa1e905bdf44bdb41572387edf90

    ver, _ := goversion.Installed()
    switch {
    case ver.Major < 0 || ver.AfterOrEqual(goversion.GoVersion{1, 10, -1, 0, 0, ""}):
        args = append(args, "-gcflags", "all=-N -l")
    case ver.AfterOrEqual(goversion.GoVersion{1, 9, -1, 0, 0, ""}):
        args = append(args, "-gcflags", "-N -l", "-a")
    default:
        args = append(args, "-gcflags", "-N -l")
    }   
    return args
}
```   
2. And the docs  `needs to be regenerated; run scripts/gen-usage-docs.go`. 
3. Please follow [contributing guidelines](https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#contributing-code)
4. Please `make test` before you pull request.

derekparker:
Thanks for the patch!

@altimac as mentioned by @chainhelen please re-format your commit message as described in https://github.com/derekparker/delve/blob/master/CONTRIBUTING.md#format-of-the-commit-message and regenerate the docs using the command mentioned. The doc you are trying to update is actually located at https://github.com/derekparker/delve/blob/master/cmd/dlv/cmds/commands.go#L172. If you update that and regenerate the docs I will happily merge this.

aarzilli:
Merged in a86711f547fdfef6a2c1bc5755569f9b25b64323

PR #1379, opened 2018-10-17
*: Show return values on CLI trace

### LINKED ISSUE #388: Print return values of traced functions
We need to parse and display the return values of traced functions.

derekparker:
This patch allows the `trace` CLI subcommand to display return values of
a function. Additionally, it will also display information on where the
function exited, which could also be helpful in determining the path
taken during function execution.

Fixes #388 
(Note: There is still some work to be done to improve trace both as CLI subcommand and within a debug session. Most notably with the addition of the return value printing, we should visually show recursive function calls better to match call with return.)

### DIFF
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -59,8 +59,11 @@ type Breakpoint struct {
 	// Breakpoint condition
 	Cond string
 
-	// tracepoint flag
+	// Tracepoint flag, signifying this is a tracepoint.
 	Tracepoint bool `json:"continue"`
+	// TraceReturn flag signifying this is a breakpoint set at a return
+	// statement in a traced function.
+	TraceReturn bool `json:"traceReturn"`
 	// retrieve goroutine information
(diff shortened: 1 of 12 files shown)


service/api/types.go
```diff
@@ -59,8 +59,11 @@ type Breakpoint struct {
 	// Breakpoint condition
 	Cond string
 
-	// tracepoint flag
+	// Tracepoint flag, signifying this is a tracepoint.
 	Tracepoint bool `json:"continue"`
+	// TraceReturn will set a breakpoint on all possible return
```
aarzilli: This makes it sound like setting this flag will automatically create a bunch of breakpoints at all return addresses, but it just flags a single breakpoint.

derekparker: Doc leftover from a previous implementation idea, will change.

derekparker:
@aarzilli pushed changes based on our IRC convo and above comment.

PR #1386, opened 2018-10-22
proc/native,proc/core: deduplicate linux register handling code

aarzilli:
```
proc/native,proc/core: deduplicate linux register handling code

The linux version of proc/native and proc/core contained largely
overlapping implementations of the register handling code, deduplicate
it by moving it into proc/linutil.

```


pkg/proc/core/core.go
```diff
@@ -154,8 +155,8 @@ type Process struct {
 
 // Thread represents a thread in the core file being debugged.
 type Thread struct {
+	regs   linutil.AMD64Registers
```
derekparker: We shouldn't use an arch specific register set here, instead prefer the interface.

aarzilli: You are right, but I'd like to change it in a later commit.
Backstory: I'm going to add minidump support to proc/core (minidumps are windows coredump equivalent), the next PR will make proc/core OS and architecture independent.

derekparker: SGTM

PR #1388, opened 2018-10-23
Update to reflect -gcflags changes in go 1.10

guidopola:
Starting from golang 1.10 the -cflags changed in order to build all the files with debug flags you need to use ```-gcflags="all=-N -l"```
see:
https://golang.org/doc/go1.10#build


chainhelen:
#1375

PR #1395, opened 2018-10-31
pkg/dwarf: Remove exec bit from reader.go

derekparker:
For some reason the Go file pkg/dwarf/reader/reader.go had the exec bit
set on it. This patch removes that.


aarzilli:
weird

PR #1405, opened 2018-11-07
Fix 1353

### LINKED ISSUE #1353: Delve should support more ways of specifying debuginfo location

1.1.0
1.10.4
Fedora 28 x64
4. What did you do?
Tried to load external debug info in Delve.
Expected to be able to either place the debuginfo in the same directory as the binary or, better yet, to be able to specify the directory to read debuginfo out of. 

6. What did you see instead?
GDB allows searching for debuginfo in multiple locations (https://www-zeuthen.desy.de/unix/unixguide/infohtml/gdb/Separate-Debug-Files.html). Currently Delve only looks in /usr/lib/debug/.build.id/*/*. The ability to specify a debug file directory would allow delve to read debuginfo even when the user doesn't have permission to write to /usr/lib/debug. It could also be useful to support the other debug info locations, for example, to have a consistent location to find a binary regardless of the build-id.

derekparker:
Adds a config file option to allow specifying a list of directories to
search in when looking for seperate external debug info files.

Fixes #1353

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -205,7 +205,7 @@ func New(process *os.Process) *Process {
 }
 
 // Listen waits for a connection from the stub.
-func (p *Process) Listen(listener net.Listener, path string, pid int) error {
+func (p *Process) Listen(listener net.Listener, path string, pid int, debugInfoDirs []string) error {
 	acceptChan := make(chan net.Conn)
 
 	go func() {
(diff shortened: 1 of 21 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -512,32 +512,32 @@ func LLDBAttach(pid int, path string) (*Process, error) {
...
+	process.SysProcAttr = sysProcAttr(false)
 
-	err := proc.Start()
+	err := process.Start()
 	if err != nil {
 		return nil, err
 	}
 
-	p := New(proc.Process)
+	p := &Process{}
```
aarzilli: The call to `New` is needed here, it initializes a bunch of things. Should be `p := New(process.Process)` after the rename.

derekparker: Fixed... the pattern matching in my brain confused that with the builtin `new`, heh.

PR #1407, opened 2018-11-13
cmd/dlv: Fix tests for use with modules

derekparker:
When we updated to use the new Go 1.11 modules, this test was not updated, and only working because an old copy of Delve in the old GOPATH.


aarzilli:
Failing tests.

derekparker:
@aarzilli fixed.

PR #1408, opened 2018-11-13
proc/proc: Extend GoroutinesInfo to allow specifying a range

### LINKED ISSUE #1403: Excesive memory usage while executing "goroutines" on a core dump with >100000 goroutines

```
Delve Debugger
Version: 1.1.0
```

```
go version go1.11.1 linux/amd64
```

```
RHEL 7.5

Linux 0d422c7c42e7 3.10.0-862.11.6.el7.x86_64 #1 SMP Fri Aug 10 16:55:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

4. What did you do?

Execute ```goroutines``` while debugging a core dump with > 100000 goroutines.

dlv using a reasonable amount of RAM

6. What did you see instead?

dlv needing a couple of GBs to finish the operation

----

I'd like to add a new RPC to obtain the number of goroutines and extend ```ListGoroutines``` to accept a "start" and "end" arguments. Those values will be then passed to ```proc.GoroutinesInfo``` (through ```debugger.Goroutines```) to limit the range of ```runtime.allgs``` that will be processed.

This way, the client can request the goroutines information on smaller batches.

- Does this sound reasonable?

- What's the policy for client/server API changes? Can I safely extend ```ListGouroutines``` in both ```rpc1``` and ```rpc2```, or do I need to a new RPC?

Thanks.

slp:
Instead of unconditionally returning all present goroutines,
GoroutinesInfo now allows specifying a range (start and count). In
addition to the array of goroutines and the error, it now also returns
the next goroutine to be processed, to be used as 'start' argument on
the next call, or 0 if all present goroutines have already been
processed.

This way clients can avoid eating large amounts of RAM while debugging
core dumps and processes with a exceptionally high amount of goroutines.

Fixes #1403

### DIFF
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -464,14 +464,22 @@ func StepOut(dbp Process) error {
 	return Continue(dbp)
 }
 
-// GoroutinesInfo returns an array of G structures representing the information
-// Delve cares about from the internal runtime G structure.
-func GoroutinesInfo(dbp Process) ([]*G, error) {
+// GoroutinesInfo searches for goroutines starting at index 'start', and
+// returns an array of up to 'count' (or all found elements, if 'count' is 0)
+// G structures representing the information Delve care about from the internal
+// runtime G structure.
+// GoroutinesInfo also returns the next index to be used as 'start' argument
+// while scanning for all available goroutines, or -1 if there was an error
+// or if the index already reached the last possible value.
+func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
 	if _, err := dbp.Valid(); err != nil {
-		return nil, err
+		return nil, -1, err
 	}
 	if dbp.Common().allGCache != nil {
-		return dbp.Common().allGCache, nil
+		// We can't use the cached array to fulfill a subrange request
+		if start == 0 && (count == 0 || count >= len(dbp.Common().allGCache)) {
+			return dbp.Common().allGCache, -1, nil
+		}
 	}
 
 	var (
@@ -493,12 +501,12 @@ func GoroutinesInfo(dbp Process) ([]*G, error) {
 
 	addr, err := rdr.AddrFor("runtime.allglen", dbp.BinInfo().staticBase)
 	if err != nil {
-		return nil, err
+		return nil, -1, err
 	}
 	allglenBytes := make([]byte, 8)
 	_, err = dbp.CurrentThread().ReadMemory(allglenBytes, uintptr(addr))
 	if err != nil {
-		return nil, err
+		return nil, -1, err
 	}
 	allglen := binary.LittleEndian.Uint64(allglenBytes)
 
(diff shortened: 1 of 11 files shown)


pkg/proc/proc.go
```diff
@@ -508,31 +512,44 @@ func GoroutinesInfo(dbp Process) ([]*G, error) {
...
 	faddr := make([]byte, dbp.BinInfo().Arch.PtrSize())
 	_, err = dbp.CurrentThread().ReadMemory(faddr, uintptr(allgentryaddr))
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 	allgptr := binary.LittleEndian.Uint64(faddr)
 
-	for i := uint64(0); i < allglen; i++ {
+	glen := 0
```
aarzilli: glen is just tracking `len(allg)` use that instead.

pkg/proc/proc.go
```diff
@@ -508,31 +512,44 @@ func GoroutinesInfo(dbp Process) ([]*G, error) {
...
+	nextg := 0
+	if count == 0 {
+		limit = int(allglen)
+	} else {
+		limit = count
+	}
+	for i := uint64(start); i < allglen; i++ {
+		if glen == limit {
```
aarzilli: Remove limit and check `if count != 0 && len(allg) >= count`

pkg/proc/proc.go
```diff
@@ -466,18 +466,22 @@ func StepOut(dbp Process) error {
 
 // GoroutinesInfo returns an array of G structures representing the information
 // Delve cares about from the internal runtime G structure.
-func GoroutinesInfo(dbp Process) ([]*G, error) {
+func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
 	if _, err := dbp.Valid(); err != nil {
-		return nil, err
+		return nil, 0, err
```
aarzilli: I think returning `nextg==0` on failure and when reading is finished is inconvenient. Seeing the loop in command.go it would be more natural if the sentinel value for "you are done" was -1. Then the goroutines command could be a single loop with `for start >= 0`, and assign the `nextg` returned by ListGoroutines to `start`.

pkg/proc/proc.go
```diff
@@ -543,9 +551,12 @@ func GoroutinesInfo(dbp Process) ([]*G, error) {
 			allg = append(allg, g)
 		}
 	}
-	dbp.Common().allGCache = allg
+	// Save the set for later only if we've proceessed all goroutines
+	if start == 0 && nextg == 0 {
```
aarzilli: This will never cache anything because nextg now is never 0. It's probably better to return directly in the `if count != 0 && len(allg) >= count` and cache unconditionally here. I think.

pkg/terminal/command.go
```diff
@@ -604,25 +625,34 @@ func goroutines(t *Term, ctx callContext, argstr string) error {
...
+		gs, nextg, err := t.client.ListGoroutines(start, count)
+		if err != nil {
+			return err
 		}
-		fmt.Printf("%sGoroutine %s\n", prefix, formatGoroutine(g, fgl))
-		if bPrintStack {
-			stack, err := t.client.Stacktrace(g.ID, 10, false, nil)
+		if start == 0 && nextg == -1 {
+			sort.Sort(byGoroutineID(gs))
+			fmt.Printf("[%d goroutines]\n", len(gs))
```
aarzilli: We are not beholden by backwards compatibility to keep this output unchanged so let's write the number of goroutines at the end of the end regardless. Also sort every time.

slp:
@aarzilli Thanks, I've applied your suggestions and command.go code looks cleaner now. Also, please excuse the obvious bugs. As I'm jumping from one project to the other (switching languages and coding styles along with them) I'm not able to gain enough fluency to catch them earlier.

pkg/proc/proc.go
```diff
@@ -466,12 +466,15 @@ func StepOut(dbp Process) error {
 
 // GoroutinesInfo returns an array of G structures representing the information
 // Delve cares about from the internal runtime G structure.
```
derekparker: Update documentation to cover new arguments and return values.

slp: Sure, how does something like this look to you?

```
// GoroutinesInfo searches for goroutines starting at index 'start', and
// returns an array of up to 'count' (or all found elements, if 'count' is 0)
// G structures representing the information Delve care about from the internal
// runtime G structure.
// GoroutinesInfo also returns the next index to be used as 'start' argument
// while scanning for all available goroutines, or -1 if there was an error
// or if the index already reached the last possible value.
```

derekparker: LGTM

slp:
@derekparker @aarzilli I've just updated the doc. Thanks!

PR #1410, opened 2018-11-13
pkg/proc: Refactor process post initialization

derekparker:
This patch is a slight refactor to share more code used for generic
process initialization. There will always be OS/backend specific
initialization, but as much as can be shared should be to prevent
duplicating of any logic (setting internal breakpoints, loading bininfo,
etc).

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -75,7 +75,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"golang.org/x/arch/x86/x86asm"
@@ -83,7 +82,7 @@ import (
 	"github.com/derekparker/delve/pkg/logflags"
 	"github.com/derekparker/delve/pkg/proc"
 	"github.com/derekparker/delve/pkg/proc/linutil"
-	"github.com/mattn/go-isatty"
+	isatty "github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 )
 
@@ -130,7 +129,7 @@ type Process struct {
 	common proc.CommonProcess
 }
 
-// Thread is a thread.
+// Thread represents an operating system thread.
 type Thread struct {
 	ID                int
 	strID             string
@@ -142,8 +141,7 @@ type Thread struct {
 }
 
 // ErrBackendUnavailable is returned when the stub program can not be found.
-type ErrBackendUnavailable struct {
-}
+type ErrBackendUnavailable struct{}
 
 func (err *ErrBackendUnavailable) Error() string {
 	return "backend unavailable"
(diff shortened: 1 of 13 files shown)


pkg/proc/interface.go
```diff
@@ -105,13 +108,17 @@ type ProcessManipulation interface {
...
+	WriteBreakpointFn() WriteBreakpointFn
 	ClearBreakpoint(addr uint64) (*Breakpoint, error)
 	ClearInternalBreakpoints() error
 }
 
 // CommonProcess contains fields used by this package, common to all
 // implementations of the Process interface.
 type CommonProcess struct {
+	// Path is the path on disk to the executable being debugged.
+	Path string
```
aarzilli: This seems redundant with the same field in BinaryInfo. Given that it only exists to pass this information to PostInitializationSetup I think it's fine if it's a parameter to that function.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -545,10 +448,89 @@ func LLDBAttach(pid int, path string, debugInfoDirs []string) (*Process, error)
...
+	// overwrite some existing memory to store the MOV.
+	if addr, err := p.conn.allocMemory(256); err == nil {
+		if _, err := p.conn.writeMemory(uintptr(addr), p.loadGInstr()); err == nil {
+			p.loadGInstrAddr = addr
+		}
+	}
+
+	err = p.updateThreadList(&threadUpdater{p: p})
+	if err != nil {
+		p.bi.Close()
```
aarzilli: Not sure why you have removed the calls to conn.Close here and elsewhere and strongly suspect it's wrong.

pkg/proc/interface.go
```diff
@@ -105,13 +108,17 @@ type ProcessManipulation interface {
 type BreakpointManipulation interface {
 	Breakpoints() *BreakpointMap
 	SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr) (*Breakpoint, error)
+	WriteBreakpointFn() WriteBreakpointFn
```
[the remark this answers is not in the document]
derekparker: Changed.

pkg/proc/interface.go
```diff
@@ -86,10 +87,12 @@ type ThreadInfo interface {
 // GoroutineInfo is an interface for getting information on running goroutines.
 type GoroutineInfo interface {
 	SelectedGoroutine() *G
+	SetSelectedGoroutine(*G)
 }
 
 // ProcessManipulation is an interface for changing the execution state of a process.
 type ProcessManipulation interface {
+	Initialize([]string) error
```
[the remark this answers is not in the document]
derekparker: Agreed, changed.

pkg/proc/interface.go
derekparker: Agreed, changed.

pkg/proc/gdbserial/gdbserver.go
derekparker: Fixed.

pkg/proc/proc.go
```diff
@@ -41,6 +41,38 @@ func (pe ProcessDetachedError) Error() string {
...
+	if err != nil {
+		return err
+	}
+
+	g, _ := GetG(p.CurrentThread())
+	p.SetSelectedGoroutine(g)
+
+	createUnrecoveredPanicBreakpoint(p, writeBreakpoint)
+
+	panicpc, err := FindFunctionLocation(p, "runtime.startpanic", true, 0)
```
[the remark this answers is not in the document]
derekparker: Fixed.

PR #1412, opened 2018-11-15
Early Go 1.12 support

aarzilli:
```
proc: correctly extract name of functions that have been inlined

If a function can be inlined it will appear as two entries in
debug_info. A DW_TAG_subprogram entry with DW_AT_inlined = true (that
will be used as the abstract origin) and a second DW_TAG_subprogram
entry with an abstract origin.
To retrieve the name of this second entry we must load its abstract
origin.

proc/*: allow stepping into functions without debug_info symbols

If proc.Step encounters a CALL instruction that points to an address
that isn't associated with any function it should still follow the
CALL.

The circumstances creating this problem do not normally occur, it was
encountered in the process of fixing a bug created by Go1.12.

tests: minor fixes for Go 1.12

proc: remove inversion of return variables from returnBreakpointInfo

It was never true that return variables were in the inverse order.
Instead in Go1.11 return variables are saved in debug_info in an
arbitrary order and inverting them just happened to work for this
specific example.

This bug was fixed in Go 1.12, regardless we should attempt to
rearrange return variables anyway.

```

### DIFF
--- a/pkg/proc/threads.go
+++ b/pkg/proc/threads.go
@@ -379,21 +379,16 @@ func setStepIntoBreakpoint(dbp Process, text []AsmInstruction, cond ast.Expr) er
 
 	instr := text[0]
 
-	if instr.DestLoc == nil || instr.DestLoc.Fn == nil {
+	if instr.DestLoc == nil {
+		// Call destination couldn't be resolved because this was not the
+		// current instruction, therefore the step-into breakpoint can not be set.
 		return nil
 	}
 
 	fn := instr.DestLoc.Fn
 
-	// Ensure PC and Entry match, otherwise StepInto is likely to set
-	// its breakpoint before DestLoc.PC and hence run too far ahead.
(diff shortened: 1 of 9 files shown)


aarzilli:
I'm posting this now because I need this for the plugin support changes. There isn't much here. The change concerning inlined functions is just to make a test pass again. Eventually we'll need a better solution for inlined function, I'll propose a solution on the mailing list soon.

pkg/proc/threads.go
```diff
@@ -379,30 +379,36 @@ func setStepIntoBreakpoint(dbp Process, text []AsmInstruction, cond ast.Expr) er
 
 	instr := text[0]
 
-	if instr.DestLoc == nil || instr.DestLoc.Fn == nil {
+	if instr.DestLoc == nil {
+		// Call destination couldn't be resolved because this is was not the
```
derekparker: s/this is was not/ this was not/

aarzilli: Done.

PR #1413, opened 2018-11-15
Plugin support (part 1)

aarzilli:
```
proc,proc/native,proc/gdbserial: initial plugin support

Adds initial support for plugins, this is only the code needed to keep
track of loaded plugins on linux (both native and gdbserial backend).

It does not actually implement support for debugging plugins on linux.

Updates #865

```

### DIFF
--- a/pkg/proc/linutil/dynamic.go
+++ b/pkg/proc/linutil/dynamic.go
@@ -1,0 +1,172 @@
+package linutil
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+
+	"github.com/go-delve/delve/pkg/proc"
+)
+
+const (
+	maxNumLibraries      = 1000000 // maximum number of loaded libraries, to avoid loading forever on corrupted memory
+	maxLibraryPathLength = 1000000 // maximum length for the path of a library, to avoid loading forever on corrupted memory
+)
+
+var ErrTooManyLibraries = errors.New("number of loaded libraries exceeds maximum")
+
+const (
+	_DT_NULL  = 0  // DT_NULL as defined by SysV ABI specification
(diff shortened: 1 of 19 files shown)


webern:
Thank you for working on support of Go plugins. I need to use that feature in my project and I'm hoping that Delve can support them sometime soon.

aarzilli:
Rebased

pkg/proc/linutil/dynamic.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/go-delve/delve/pkg/proc"
+)
+
+const (
+	maxNumLibraries      = 1000000 // maximum number of loaded libraries, to avoid loading forever on corrupted memory
+	maxLibraryPathLength = 1000000 // maximum length for the path of a library, to avoid loading forever on corrupted memory
+)
+
+var ErrTooManyLibraries = errors.New("too many libraries")
```
derekparker: nit: I think we can make this error a little more descriptive, e.g. "maximum number of shared libraries loaded" or something.

pkg/proc/linutil/dynamic.go
```diff
@@ -0,0 +1,172 @@
...
+var ErrStringTooLong = errors.New("string too long")
```
derekparker: I think we can also make this more descriptive, unless there is context coming from a wrapping error.

pkg/proc/linutil/dynamic.go
```diff
@@ -0,0 +1,172 @@
...
+			break
+		}
+		r = append(r, buf[0])
+		addr++
+	}
+	return string(r), nil
+}
+
+// ElfUpdateSharedObjects reads the list of dynamic libraries loaded by the
+// dynamic linker from the .dynamic section and uses it to updated p.BinInfo().
```
derekparker: s/updated/update/

PR #1414, opened 2018-11-16
Plugin support (part 2)

aarzilli:
```
proc: support debugging plugins

This change splits the BinaryInfo object into a slice of Image objects
containing information about the base executable and each loaded shared
library (note: go plugins are shared libraries).

Delve backens are supposed to call BinaryInfo.AddImage whenever they
detect that a new shared library has been loaded.

Member fields of BinaryInfo that are used to speed up access to dwarf
(Functions, packageVars, consts, etc...) remain part of BinaryInfo and
are updated to reference the correct image object. This simplifies this
change.

This approach has a few shortcomings:

1. Multiple shared libraries can define functions or globals with the
same name and we have no way to disambiguate between them.

2. We don't have a way to handle library unloading.

Both of those affect C shared libraries much more than they affect go
plugins. Go plugins can't be unloaded at all and a lot of name
collisions are prevented by import paths.

There's only one problem that is concerning: if two plugins both import
the same package they will end up with multiple definition for the same
function.
For example if two plugins use fmt.Printf the final in-memory image
(and therefore our BinaryInfo object) will end up with two copies of
fmt.Printf at different memory addresses. If a user types
break fmt.Printf
a breakpoint should be created at *both* locations.
Allowing this is a relatively complex change that should be done in a
different PR than this.

For this reason I consider this approach an acceptable and sustainable
stopgap.

Updates #865
```

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -17,7 +17,7 @@ type Arch interface {
 	DerefTLS() bool
 	FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *BinaryInfo) *frame.FrameContext
 	RegSize(uint64) int
-	RegistersToDwarfRegisters(regs Registers, staticBase uint64) op.DwarfRegisters
+	RegistersToDwarfRegisters(bi *BinaryInfo, regs Registers) op.DwarfRegisters
 	GoroutineToDwarfRegisters(*G) op.DwarfRegisters
 }
 
@@ -270,7 +270,7 @@ func maxAmd64DwarfRegister() int {
 
 // RegistersToDwarfRegisters converts hardware registers to the format used
 // by the DWARF expression interpreter.
-func (a *AMD64) RegistersToDwarfRegisters(regs Registers, staticBase uint64) op.DwarfRegisters {
+func (a *AMD64) RegistersToDwarfRegisters(bi *BinaryInfo, regs Registers) op.DwarfRegisters {
 	dregs := make([]*op.DwarfRegister, maxAmd64DwarfRegister()+1)
 
 	dregs[amd64DwarfIPRegNum] = op.DwarfRegisterFromUint64(regs.PC())
@@ -292,7 +292,9 @@ func (a *AMD64) RegistersToDwarfRegisters(regs Registers, staticBase uint64) op.
 		}
 	}
 
-	return op.DwarfRegisters{StaticBase: staticBase, Regs: dregs, ByteOrder: binary.LittleEndian, PCRegNum: amd64DwarfIPRegNum, SPRegNum: amd64DwarfSPRegNum, BPRegNum: amd64DwarfBPRegNum}
+	so := bi.pcToImage(regs.PC())
+
+	return op.DwarfRegisters{StaticBase: so.StaticBase, Regs: dregs, ByteOrder: binary.LittleEndian, PCRegNum: amd64DwarfIPRegNum, SPRegNum: amd64DwarfSPRegNum, BPRegNum: amd64DwarfBPRegNum}
 }
 
 // GoroutineToDwarfRegisters extract the saved DWARF registers from a parked
@@ -302,5 +304,8 @@ func (a *AMD64) GoroutineToDwarfRegisters(g *G) op.DwarfRegisters {
 	dregs[amd64DwarfIPRegNum] = op.DwarfRegisterFromUint64(g.PC)
 	dregs[amd64DwarfSPRegNum] = op.DwarfRegisterFromUint64(g.SP)
 	dregs[amd64DwarfBPRegNum] = op.DwarfRegisterFromUint64(g.BP)
-	return op.DwarfRegisters{StaticBase: g.variable.bi.staticBase, Regs: dregs, ByteOrder: binary.LittleEndian, PCRegNum: amd64DwarfIPRegNum, SPRegNum: amd64DwarfSPRegNum, BPRegNum: amd64DwarfBPRegNum}
+
+	so := g.variable.bi.pcToImage(g.PC)
+
+	return op.DwarfRegisters{StaticBase: so.StaticBase, Regs: dregs, ByteOrder: binary.LittleEndian, PCRegNum: amd64DwarfIPRegNum, SPRegNum: amd64DwarfSPRegNum, BPRegNum: amd64DwarfBPRegNum}
 }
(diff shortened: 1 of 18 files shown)


pkg/proc/bininfo.go
```diff
@@ -427,44 +435,128 @@ func (bi *BinaryInfo) PCToFunc(pc uint64) *Function {
...
-		bi.sepDebugCloser.Close()
+	var errs []error
+	for _, image := range bi.Images {
+		if image.sepDebugCloser != nil {
+			err := image.sepDebugCloser.Close()
+			if err != nil {
+				errs = append(errs, fmt.Errorf("closing shared object %q (split dwarf): %v", image.Path, err))
+			}
+		}
+		if image.closer != nil {
```
derekparker: nit: A refactor could possibly be to embed the closer in image so we can just call `image.Close()`.

pkg/proc/bininfo.go
```diff
@@ -427,44 +435,128 @@ func (bi *BinaryInfo) PCToFunc(pc uint64) *Function {
...
-	bi.loadErr = fmt.Errorf(fmtstr, args...)
-	bi.loadErrMu.Unlock()
+func (image *Image) setLoadError(fmtstr string, args ...interface{}) {
+	image.loadErrMu.Lock()
+	image.loadErr = fmt.Errorf(fmtstr, args...)
+	image.loadErrMu.Unlock()
 }
 
-// LoadError returns any internal load error.
+// LoadError returns any internal error incurred while loading the last shared object.
```
derekparker: Why return just the last error specifically?

pkg/proc/bininfo.go
```diff
@@ -474,22 +566,26 @@ func (c *nilCloser) Close() error { return nil }
 // LoadFromData creates a new BinaryInfo object using the specified data.
 // This is used for debugging BinaryInfo, you should use LoadBinary instead.
 func (bi *BinaryInfo) LoadFromData(dwdata *dwarf.Data, debugFrameBytes, debugLineBytes, debugLocBytes []byte) {
```
derekparker: We should rename this to something like `LoadImageFromData` and update the comment above to reflect the intent of the new changes.

pkg/proc/bininfo.go
```diff
@@ -696,70 +803,82 @@ func (bi *BinaryInfo) LoadBinaryInfoElf(path string, entryPoint uint64, debugInf
...
-		bi.staticBase = entryPoint - elfFile.Entry
-	} else {
-		if elfFile.Type == elf.ET_DYN {
-			return ErrCouldNotDetermineRelocation
+	if image.index == 0 {
+		// adding executable file:
+		// - addr is entryPoint therefore staticBase needs to be calculated by
+		//   subtracting it from the entry point specified in the executable file
+		// - memory address of the .dynamic section needs to be recorded in
+		//   BinaryInfo so that we can find loadded libraries.
```
derekparker: nit: s/loadded/loaded/

pkg/proc/bininfo.go
```diff
@@ -696,70 +803,82 @@ func (bi *BinaryInfo) LoadBinaryInfoElf(path string, entryPoint uint64, debugInf
...
 
-	if entryPoint != 0 {
-		bi.staticBase = entryPoint - elfFile.Entry
-	} else {
-		if elfFile.Type == elf.ET_DYN {
-			return ErrCouldNotDetermineRelocation
+	if image.index == 0 {
+		// adding executable file:
+		// - addr is entryPoint therefore staticBase needs to be calculated by
+		//   subtracting it from the entry point specified in the executable file
```
derekparker: nit: technically you're subtracting the entry point from addr.

pkg/proc/bininfo.go
```diff
@@ -696,70 +803,82 @@ func (bi *BinaryInfo) LoadBinaryInfoElf(path string, entryPoint uint64, debugInf
...
+		// - memory address of the .dynamic section needs to be recorded in
+		//   BinaryInfo so that we can find loadded libraries.
+		if addr != 0 {
+			image.StaticBase = addr - elfFile.Entry
+		} else {
```
derekparker: nit: else if

pkg/proc/bininfo.go
aarzilli: IIRC it was the easiest way to do it at the time, because it was called from the backends. Given that now it's only called in proc.PostInitializationSetup, I've replaced it there with a loop through all the images.

pkg/proc/arch.go
```diff
@@ -302,5 +304,10 @@ func (a *AMD64) GoroutineToDwarfRegisters(g *G) op.DwarfRegisters {
 	dregs[amd64DwarfIPRegNum] = op.DwarfRegisterFromUint64(g.PC)
 	dregs[amd64DwarfSPRegNum] = op.DwarfRegisterFromUint64(g.SP)
 	dregs[amd64DwarfBPRegNum] = op.DwarfRegisterFromUint64(g.BP)
-	return op.DwarfRegisters{StaticBase: g.variable.bi.staticBase, Regs: dregs, ByteOrder: binary.LittleEndian, PCRegNum: amd64DwarfIPRegNum, SPRegNum: amd64DwarfSPRegNum, BPRegNum: amd64DwarfBPRegNum}
+
+	bi := g.variable.bi
```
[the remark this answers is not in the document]
derekparker: I agree but I think this is alright for now. When (and if) we ever run into a situation where a linear search through loaded shared objects becomes a problem we can tackle it.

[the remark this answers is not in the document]
aarzilli: Added the helper function. 

>This should overall be more efficient than simply using the PC to find the compilation unit and thus the image because we can binary search function entries and we're doing a linear search of compilation unit PC ranges, correct?

I think so. Ideally we'd just do a binary search on images but I haven't figured out how to find the end address.

pkg/proc/bininfo.go
```diff
@@ -427,44 +435,128 @@ func (bi *BinaryInfo) PCToFunc(pc uint64) *Function {
...
+	loadErrMu sync.Mutex
+	loadErr   error
 }
 
-// AddImage adds the specified image to bi.
-func (bi *BinaryInfo) AddImage(path string, addr uint64) {
-	if !strings.HasPrefix(path, "/") {
-		return
+// hasImage returns true if the shared object has already been added.
+func (bi *BinaryInfo) hasImage(path string, addr uint64) bool {
```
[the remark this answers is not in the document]
aarzilli: Inlined is good.

scottlepp:
@aarzilli @derekparker sorry a noob here.  How would I set a breakpoint in the plugin if I have a setup like [this](https://github.com/scottlepp/reload/tree/master/pkg) where main loads the plugin?

PR #1418, opened 2018-11-21
substitute-path in config on Windows was not working

zavla:
Because of windows os.PathSeparator always ='\\' and headless server always returning source files paths as c:/dir/file.


derekparker:
Are you saying that in Windows, paths returned from the server are "/" instead of "\"? If that's true, that's the root cause of the issue that should be fixed, not this.

aarzilli:
> Are you saying that in Windows, paths returned from the server are "/" instead of ""? If that's true, that's the root cause of the issue that should be fixed, not this.

We are probably returning whatever the linker writes there and '/' is a valid path separator on windows too. I think substitutePath should work with both.

zavla:
As @aarzilli  have said headless server returns the paths written into exe.

And for example for the precompiled "go.exe" downloaded from golang.org and for the "dlv.exe" compiled by myself on windows -  the hardcoded debug source paths are all with "/" as a separator.
And terminal_test.go fails because of waiting for the "\" separator on windows.
The easy way is to adapt your non exported terminal.substitutePath() , not the compiler itself debug info generation .

PR #1429, opened 2018-11-28
pkg/proc: align memory size of tls arena to pointer-sized boundary (fix for #1428)

### LINKED ISSUE #1428: Incorrectly detected goroutine ID with cgo tests (linux/amd64)
1. `go get -u github.com/lfittl/pg_query_go`
2. `dlv test`
3. `b parse_test.go:707`
4. `c`

Expected: code is running a current goroutine
What happens: no current goroutine

Additionally we should treat `GoroutineID == 0` in api.EvalScope as selecting the current thread if the current thread is not running a goroutine.

Bug originally reported as: https://youtrack.jetbrains.com/issue/GO-6497

aarzilli:
```
pkg/proc: align memory size of tls arena to pointer sized boundary

The size of the TLS memory arena needs to be aligned to pointer sized
boundaries on 86x64 architectures, otherwise some programs using cgo
will not have the correct offset for the g struct.

No tests because reproducing this problem depends on behavior of the
GNU ld linker caused by unclear influences.

Fixes #1428.

proc: handle gid == 0 in FindGoroutine

Goroutine id == 0 is special (there can be many goroutines with id 0).
If the caller of FindGoroutine asks for gid==0 and current thread is
running a goroutine 0 (i.e. either no goroutine or a special
goroutine) return whatever goroutine is running on the current thread.

Updates #1428

```

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -753,9 +753,12 @@ func (bi *BinaryInfo) setGStructOffsetElf(exe *elf.File, wg *sync.WaitGroup) {
 			break
 		}
 	}
+	memsz := tls.Memsz
+
+	memsz = (memsz + uint64(bi.Arch.PtrSize()) - 1) & ^uint64(bi.Arch.PtrSize()-1) // align to pointer-sized-boundary
 	// The TLS register points to the end of the TLS block, which is
 	// tls.Memsz long. runtime.tlsg is an offset from the beginning of that block.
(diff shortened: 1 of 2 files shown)


aarzilli:
cc @zolotov: any chance the affected user could check this out?

zolotov:
@aarzilli the issue is reproducible on my machine, so I can check this tomorrow. Thanks!

zolotov:
Works fine with this fix. Goroutine=5 comes to the client instead of Goroutine=0

aarzilli:
@zolotov thank you.

PR #1440, opened 2018-12-06
Improvements to ListGoroutines

aarzilli:
```
service: make ListGoroutines return all running goroutines when
Start==0

Change ListGoroutines API call so that it tries to fit all running
goroutines in the ListGoroutines(Start=0) results.

Users are more likely to be interested in running goroutines, with this
change a frontend that just calls ListGoroutines(Start=0, Count=n) is
guaranteed to show all running goroutines to its users (as long as
there aren't more than n running goroutines).

proc: improve performance of FindGoroutine in normal circumstances

FindGoroutine can be slow when there are many goroutines running. This
can not be fixed in the general case, however:

1. Instead of getting the entire list of goroutines at once just get a
few at a time and return as soon as we find the one we want.

2. Since FindGoroutine is mostly called by ConvertEvalScope and users
are more likely to request informations about a goroutine running on a
thread, look within the threads first.

proc: fix GoroutinesInfo cache

The allg cache was corrupted when the count parameter was actually
reached.

Fix the bug and add a test for this.

```

### DIFF
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -604,20 +604,37 @@ func FindGoroutine(dbp Process, gid int) (*G, error) {
 		if g == nil || g.ID == 0 {
 			return g, nil
 		}
+		return nil, fmt.Errorf("Unknown goroutine %d", gid)
 	}
 
-	gs, _, err := GoroutinesInfo(dbp, 0, 0)
-	if err != nil {
-		return nil, err
-	}
-	for i := range gs {
-		if gs[i].ID == gid {
-			if gs[i].Unreadable != nil {
-				return nil, gs[i].Unreadable
-			}
-			return gs[i], nil
+	// Calling GoroutinesInfo could be slow if there are many goroutines
+	// running, check if a running goroutine has been requested first.
+	for _, thread := range dbp.ThreadList() {
+		g, _ := GetG(thread)
+		if g != nil && g.ID == gid {
+			return g, nil
 		}
 	}
+
+	const goroutinesInfoLimit = 10
+	nextg := 0
+	for nextg >= 0 {
+		var gs []*G
+		var err error
+		gs, nextg, err = GoroutinesInfo(dbp, nextg, goroutinesInfoLimit)
+		if err != nil {
+			return nil, err
+		}
+		for i := range gs {
+			if gs[i].ID == gid {
+				if gs[i].Unreadable != nil {
+					return nil, gs[i].Unreadable
+				}
+				return gs[i], nil
+			}
+		}
+	}
+
 	return nil, fmt.Errorf("Unknown goroutine %d", gid)
 }
 
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -578,7 +578,9 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
 			allg = append(allg, g)
 		}
 	}
-	dbp.Common().allGCache = allg
+	if start == 0 {
+		dbp.Common().allGCache = allg
+	}
 
 	return allg, -1, nil
 }
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -4135,3 +4135,32 @@ func TestIssue1432(t *testing.T) {
 		assertNoError(err, t, "SetVariable")
 	})
 }
+
+func TestGoroutinesInfoLimit(t *testing.T) {
+	withTestProcess("teststepconcurrent", t, func(p proc.Process, fixture protest.Fixture) {
+		setFileBreakpoint(p, t, fixture, 37)
+		assertNoError(proc.Continue(p), t, "Continue()")
+
+		gcount := 0
+		nextg := 0
+		const goroutinesInfoLimit = 10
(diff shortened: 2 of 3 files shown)


aarzilli:
This is a series of improvements to GoroutinesInfo/ListGoroutines connected to https://github.com/Microsoft/vscode-go/issues/2133:

1. Fix a bug in GoroutinesInfo where the wrong list would be cached in allgCache
2. Improve performance of FindGoroutine, especially when it's called with the gid of a running goroutine
3. Change API call ListGoroutines to always return running goroutines in the first batch (if `Count != 0` is specified), This is a quality of life improvement, it means that a front end can just call `ListGoroutines(0, n)` and its guaranteed to get all running goroutines (unless n is very small or there is a pathologically large number of threads).

service/debugger/debugger.go
```diff
@@ -916,7 +917,44 @@ func (d *Debugger) Goroutines(start, count int) ([]*api.Goroutine, int, error) {
 	if err != nil {
 		return nil, 0, err
 	}
+	if start > 0 || nextg >= 0 {
+		// Ensure that the first batch of goroutines always contains all running
```
derekparker: nit: I think we should reword this comment, as we don't make any explicit guarentee that we are returning *all* running goroutines in the first batch, as that's limited by the count the user specifies. It would suffice, I think, to simply say that we sort and prioritize running goroutines first.

service/debugger/debugger.go
```diff
@@ -916,7 +917,44 @@ func (d *Debugger) Goroutines(start, count int) ([]*api.Goroutine, int, error) {
 	if err != nil {
 		return nil, 0, err
 	}
+	if start > 0 || nextg >= 0 {
```
derekparker: I'm confused, here we're checking if `start > 0` and then trying to return all running goroutines, however in the commit and documentation it's specified that we only want to prioritize and return only running goroutines when `start == 0`. This leads us to having a bunch of conditionals below, I think we can clean this up by having this check first if `start ==0`, no?

aarzilli: There are two paths here: `start == 0 && nextg >= 0` where we add running goroutines and `start > 0` where we remove them (because we returned them before when we had `start == 0`). It is a bit confusing as it is, I rewrote it by factoring out the common code and moving the conditions outside, hopefully it's clearer.

service/debugger/debugger.go
```diff
@@ -906,6 +907,25 @@ func (d *Debugger) SetVariableInScope(scope api.EvalScope, symbol, value string)
 	return s.SetVariable(symbol, value)
 }
 
+func (d *Debugger) getRunningGoroutines(count int) map[int]*proc.G {
+	threads := d.target.ThreadList()
+	sort.Slice(threads, func(i, j int) bool { return threads[i].ThreadID() < threads[j].ThreadID() })
```
derekparker: What does sorting buy us here?

aarzilli: The output of ThreadList is not guaranteed to always be in the same order (because the native backend is converting from a map).

service/debugger/debugger.go
```diff
@@ -916,7 +936,31 @@ func (d *Debugger) Goroutines(start, count int) ([]*api.Goroutine, int, error) {
 	if err != nil {
 		return nil, 0, err
 	}
+
+	// Prioritize running goroutines in the first batch we return (adds up to
+	// 'count' running goroutines to the first batch and insures that they
```
derekparker: s/insures/ensures/

service/debugger/debugger.go
derekparker: Right, but we just iterate them and them stuff them into another map and return it, I don't see why we should sort them first, order doesn't really matter here unless I'm missing something obvious.

derekparker: Is this to make the call idempotent when called with a same `count` value?

service/debugger/debugger.go
```diff
@@ -916,7 +936,31 @@ func (d *Debugger) Goroutines(start, count int) ([]*api.Goroutine, int, error) {
...
+	// aren't returned a second time in the batches after the first one).
+	if start > 0 {
```
derekparker: I still don't like the semantics here, and I think it would be better to have a parameter like `onlyRunning` if a caller is only concerned with running Goroutines.

Consider a caller specifies `start == 0, count == 5`, then (and I don't know why they would, but just consider) they call it again with `start == 0, count ==5`. All running goroutines would be removed from the list, which just kind of seems like odd behavior. Lastly, I don't think we should ever return more than `count` items, that's just very unexpected behavior.

service/debugger/debugger.go
aarzilli: Yes.

service/debugger/debugger.go
aarzilli: It doesn't make sense to call `proc.GoroutinesInfo` with arbitrary values of `start`, it will skip an arbitrary number n (with `0 <= n < start`).

You've convinced me that this feature (returning running goroutines first) isn't worth it. I don't think we can implement it without the weird behavior and since running goroutines are returned as part of GetState it isn't that useful anyway.

I've removed this code and added an exetended explanation of Start and Count.

PR #1442, opened 2018-12-14
tests: fix typecheckrpc.go and convert it to a test

aarzilli:
```
tests: fix typecheckrpc.go and convert it to a test

We forgot to run typecheckrpc.go periodically and it didn't work
anymore and there were some minor errors in service/rpc2/client.go.

Rewrite typecheckrpc.go using go/packages, so that it works with go1.11
and go.mod, and fix the issues in client.go

```


derekparker:
There's a lot of cruft from vendoring in here, that should be cleaned out with the `make vendor` command.

aarzilli:
The vendoring cruft happens because I ran `go mod vendor` for the first time and modules and glide have different ideas about what should go in vendor. I guess this raises the question of what we should do with glide's configuration and lockfile.

derekparker:
@aarzilli let's just run our `make vendor` instead, which cleans all this out.

aarzilli:
I've tried running `make vendor` but glide is just crashing with a sigsegv. I think it's probably because I'm not using GOPATH anymore. Rerunning glide would actually cause far more vendor churn because it will try to update logrus, blackfriday, go-isatty and argv to their latest version (we didn't pin them down). And then we would have to regenerate go.mod.

I'd prefer to eat the difference now and move to `go mod vendor` for our vendor directory.

PR #1449, opened 2018-12-29
Update install.md to reflect the use of go modules

farnasirim:
GO111MODULE=on needs to be set for go modules to be enabled during installation.


aarzilli:
When does this make a difference?

derekparker:
I don't think this is strictly necessary right? As long as a user is running 1.11 and there is a go.mod file in the directory modules are used automatically, if I'm not mistaken.

farnasirim:
I believe you are right. `GO111MODULE=on` in go 1.11 does the job. Closing.

PR #1455, opened 2019-01-04
config/config.go

### LINKED ISSUE #1454: Respect XDG_CONFIG_HOME when deciding where to store config file
It'd be great if delve respected `$XDG_CONFIG_HOME` when defined instead of storing the config in the user's home directory.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables

1.1.0

go version go1.11 linux/amd64

Ubuntu amd64

4. What did you do?
Run any `delve` command

Configuration file stored at `/home/greg/.config/dlv`

6. What did you see instead?
Configuration file stored at `/home/greg/.dlv`

GregorioMartinez:
This change checks for the environmental variable XDG_CONFIG_HOME
before storing the configuration file in the home directory. If found 
it stores the config at $XDG_CONFIG_HOME/dlv

Closes #1454

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -71,19 +71,25 @@ func LoadConfig() *Config {
 
 	h, err := hasOldConfig()
 	if err != nil {
-		fmt.Printf("Unable to determine if old config exists: %s", err)
+		if _, err := fmt.Fprintf(os.Stderr, "Unable to determine if old config exists: %v\n", err); err != nil {
+			fmt.Printf("Unable to write to stderr: %v\n", err)
+		}
 	}
 
 	if h {
 		userHomeDir := getUserHomeDir()
 		oldLocation := path.Join(userHomeDir, configDirHidden)
 		if err := moveOldConfig(); err != nil {
-			fmt.Printf("Unable to move old config: %v.\n", err)
+			if _, err := fmt.Fprintf(os.Stderr, "Unable to move old config: %v\n", err); err != nil {
+				fmt.Printf("Unable to write to stderr: %v\n", err)
+			}
 			return &Config{}
 		}
 
 		if err := os.RemoveAll(oldLocation); err != nil {
-			fmt.Printf("Unable to remove old config location: %s", err)
+			if _, err := fmt.Fprintf(os.Stderr, "Unable to remove old config location: %v\n", err); err != nil {
+				fmt.Printf("Unable to write to stderr: %v\n", err)
+			}
 			return &Config{}
 		}
 		fmt.Printf("Successfully moved config from: %s to: %s\n", oldLocation, fullConfigFile)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -69,12 +69,12 @@ func LoadConfig() *Config {
 		return &Config{}
 	}
 
-	h, err := hasOldConfig()
+	hasOldConfig, err := hasOldConfig()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Unable to determine if old config exists: %v\n", err)
 	}
 
-	if h {
+	if hasOldConfig {
 		userHomeDir := getUserHomeDir()
 		oldLocation := path.Join(userHomeDir, configDirHidden)
 		if err := moveOldConfig(); err != nil {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/user"
 	"path"
+	"runtime"
 
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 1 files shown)


GregorioMartinez:
You're right @aarzilli. 
I'll update this to default to `$HOME/.config` on Linux machines when `XDG_CONFIG_HOME` is not set.

For backwards compatibility what do you think would be best? Check for the existence of `~/.dlv` and move it to `XDG_CONFIG_HOME` if found? 

RE:  `XDG_CONFIG_DIRS` - This is used to search for more config files on top of `XDG_CONFIG_HOME`. I don't believe there would be a case where `/usr/local/share/:/usr/share/ ` or `/etc/xdg` would have an additional config for delve, but I can add it in if you think it's necessary.

> $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory.

aarzilli:
As I said before, as far as I'm concerned this is fine as it is.

GregorioMartinez:
Alright, I can add the backwards compatibility flow in.
Should we also delete the current `$HOME/.dlv` directory or just print a message so users know they can remove it themselves?

GregorioMartinez:
I made the requested changes, but not entirely sure how we should write the tests for this.
If anyone has suggestions on how to do this and could point me to the right direction that'd be helpful.

pkg/config/config.go
```diff
@@ -68,6 +69,32 @@ func LoadConfig() *Config {
...
+		fmt.Printf("Successfully moved config location\n")
+		fmt.Printf("Attempting to remove old config directory: %s\n", oldLocation)
+
+		if err := os.RemoveAll(oldLocation); err != nil {
+			fmt.Printf("Unable to remove old config location: %s", err)
+			return &Config{}
+		}
+
+		fmt.Printf("Successfully removed old config directory\n")
+
```
aarzilli: Remove empty line.

pkg/config/config.go
```diff
@@ -68,6 +69,32 @@ func LoadConfig() *Config {
...
 
+	h, err := hasOldConfig()
+	if err != nil {
+		fmt.Printf("Unable to determine if old config exists: %s", err)
+	}
+
+	if h {
+		userHomeDir := getUserHomeDir()
+		oldLocation := path.Join(userHomeDir, configDirHidden)
+		fmt.Printf("Attempting to move config from: %s/%s to %s\n", oldLocation, configFile, fullConfigFile)
```
aarzilli: I would prefer if this only print a maximum of one message (either one saying that the file was moved successfully or one saying that there was an error).

There's also the problem that this runs before the server starts listening which means that this messages being print before the "API server listening" message, which will break frontends.

GregorioMartinez: @aarzilli I consolidated the messages into one and removed a few other extra empty lines to better match the style that was already present in the file.

Can you elaborate on the API server listening message problem?
When I test it locally I get the message after the "API server listening at:" message and the current file also prints messages out in the same way that I am doing now.

aarzilli: The problem is that clients will look for the "API server listening at:" and expect it to be the first thing on stdout. I think those messages should go to stderr `fmt.Fprintf(os.Stderr, ...` so they don't interfere with that.

pkg/config/config.go
```diff
@@ -71,19 +71,25 @@ func LoadConfig() *Config {
 
 	h, err := hasOldConfig()
 	if err != nil {
-		fmt.Printf("Unable to determine if old config exists: %s", err)
+		if _, err := fmt.Fprintf(os.Stderr, "Unable to determine if old config exists: %v\n", err); err != nil {
```
aarzilli: why? if stderr is closed, it's closed.

pkg/config/config.go
```diff
@@ -71,25 +71,19 @@ func LoadConfig() *Config {
...
 		}
 
 		if err := os.RemoveAll(oldLocation); err != nil {
-			if _, err := fmt.Fprintf(os.Stderr, "Unable to remove old config location: %v\n", err); err != nil {
-				fmt.Printf("Unable to write to stderr: %v\n", err)
-			}
+			fmt.Fprintf(os.Stderr, "Unable to remove old config location: %v\n", err)
 			return &Config{}
 		}
 		fmt.Printf("Successfully moved config from: %s to: %s\n", oldLocation, fullConfigFile)
```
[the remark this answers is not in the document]
GregorioMartinez: @aarzilli Done. Thanks for catching that one.

aarzilli:
LGTM

pkg/config/config.go
```diff
@@ -67,6 +69,26 @@ func LoadConfig() *Config {
 		return &Config{}
 	}
 
+	h, err := hasOldConfig()
```
derekparker: nit: Let's use a more descriptive var name here than just `h`.

GregorioMartinez: @derekparker Done!

PR #1458, opened 2019-01-08
Documentation: update recommended compiler flags for 'exec' command

aarzilli:
```
Documentation: update recommended compiler flags for 'exec' command

On Go 1.10 -gcflags='all=-N -l' should be preferred, update the
documentation of 'exec' command to reflect this.

```


aarzilli:
Cleaned up version of #1375 since Go 1.12rc1 (and therefore delve 1.2.0) is approaching.

PR #1466, opened 2019-01-18
terminal: adds embedded scripting language

aarzilli:
```
terminal: adds embedded scripting language

If the argument of 'source' ends in '.star' it will be interpreted as a
starlark script.
If the argument of 'source' is '-' an interactive starlark repl will be
started.

For documentation on how the starlark execution environment works see
Documentation/cli/starlark.md.

The starlark API is autogenerated from the JSON-RPC API by
script/gen-starlark-bindings.go.
In general for each JSON-RPC API a single global starlark function is
created.
When one of those functions is called (through a starlark script) the
arguments are converted to go structs using reflection. See
unmarshalStarlarkValue in pkg/terminal/starbind/conv.go.
If there are no type conversion errors the JSON-RPC call is executed.
The return value of the JSON-RPC call is converted back into a starlark
value by interfaceToStarlarkValue (same file):

* primitive types (such as integers, floats or strings) are converted
by creating the corresponding starlark value.
* compound types (such as structs and slices) are converted by wrapping
their reflect.Value object into a type that implements the relevant
starlark interfaces.
* api.Variables are treated specially so that their Value field can be
of the proper type instead of always being a string.

Implements #1415

```

### DIFF
--- a/Documentation/cli/starlark.md
+++ b/Documentation/cli/starlark.md
@@ -1,0 +1,263 @@
+# Introduction
+
+Passing a file with the .star extension to the `source` command will cause delve to interpret it as a starlark script.
+
+Starlark is a dialect of python, a [specification of its syntax can be found here](https://github.com/google/starlark-go/blob/master/doc/spec.md).
+
+In addition to the normal starlark built-ins delve defines [a number of global functions](#Starlark-built-ins) that can be used to interact with the debugger.
+
+After the file has been evaluated delve will bind any function starting with `command_` to a command-line command: for example `command_goroutines_wait_reason` will be bound to `goroutines_wait_reason`. 
+
+Then if a function named `main` exists it will be executed.
+
+Global functions with a name that begins with a capital letter will be available to other scripts.
+
+# Starlark built-ins
+
+<!-- BEGIN MAPPING TABLE -->
+Function | API Call
+---------|---------
+amend_breakpoint(Breakpoint) | Equivalent to API call [AmendBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AmendBreakpoint)
+ancestors(GoroutineID, NumAncestors, Depth) | Equivalent to API call [Ancestors](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Ancestors)
+attached_to_existing_process() | Equivalent to API call [AttachedToExistingProcess](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AttachedToExistingProcess)
+cancel_next() | Equivalent to API call [CancelNext](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.CancelNext)
+checkpoint(Where) | Equivalent to API call [Checkpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Checkpoint)
+clear_breakpoint(Id, Name) | Equivalent to API call [ClearBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.ClearBreakpoint)
+clear_checkpoint(ID) | Equivalent to API call [ClearCheckpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.ClearCheckpoint)
+raw_command(Name, ThreadID, GoroutineID, ReturnInfoLoadConfig, Expr, UnsafeCall) | Equivalent to API call [Command](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Command)
+create_breakpoint(Breakpoint) | Equivalent to API call [CreateBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.CreateBreakpoint)
+detach(Kill) | Equivalent to API call [Detach](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Detach)
+disassemble(Scope, StartPC, EndPC, Flavour) | Equivalent to API call [Disassemble](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Disassemble)
+eval(Scope, Expr, Cfg) | Equivalent to API call [Eval](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Eval)
+find_location(Scope, Loc) | Equivalent to API call [FindLocation](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.FindLocation)
+function_return_locations(FnName) | Equivalent to API call [FunctionReturnLocations](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.FunctionReturnLocations)
+get_breakpoint(Id, Name) | Equivalent to API call [GetBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.GetBreakpoint)
+get_thread(Id) | Equivalent to API call [GetThread](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.GetThread)
+is_multiclient() | Equivalent to API call [IsMulticlient](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.IsMulticlient)
+last_modified() | Equivalent to API call [LastModified](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.LastModified)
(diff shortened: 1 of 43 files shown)


alandonovan:
If you haven't seen it already, take a look at stargo, which is a pending change to go.starlark.net that I hope to commit soon. It, like Nate Finch's Starlight-go, provides a Starlark API for Go values, variables, functions, and types, but it more cleanly handles some of the trickier aliasing aspects arising from Go variables of composite (array/struct) type.  Under the hood it uses reflection, but you could make a variant that uses whatever Delve's representation is for values and variables in the target Go program.

The key point is this: stargo adds some changes to the Starlark compiler (see starlark.Variable) to make Go expressions such as a[i].m() work correctly, so that if a is a variable of kind array, print(a[i]) prints the value of the array element a[i], which is itself a variable, but a[i].m() calls a method m on the _address_ of that variable. Ditto for struct fields x.f.  Examples here:
https://github.com/google/starlark-go/blob/wip-stargo/stargo/testdata/bytes.star
https://github.com/google/starlark-go/blob/wip-stargo/stargo/testdata/http.star
https://github.com/google/starlark-go/blob/wip-stargo/stargo/testdata/parser.star

Stargo also adds special syntax for &x[i] and *ptr and *T, but nice though they are I think I'm going to have to withdraw them and use intrinsic functions addr(x[i]) and deref(ptr), to avoid fragmenting the language syntax unduly.  You can see an example of a Stargo program that uses these features at:
https://github.com/google/starlark-go/blob/wip-stargo/staranise/iterator_leak.star

Currently Stargo doesn't allow Starlark globals to be of type Variable, but that's an easy fix that would allow you to expose the Go variables of your target program as toplevel variables in the Starlark REPL, and for their var/value/ptr/method operations to work in the usual way, correctly.

aarzilli:
> If you the script a function that takes the starlark value and returns the go value you want, that can work with either stargo or starlight.

I don't understand. What would this look like?

> Note that in starlight at least, if you pass in a type that implements starlark.Value, starlight will just pass it through as-is, so the value itself determines how it interacts with starlark code. I think stargo probably does something similar.

Yes, but I don't want to manually implement starlark.Value for every type returned by all JSON-RPC APIs, hence conv.go.

derekparker:
Haven't been ignoring this, just been incredibly busy recently. Been looking over this PR however and started a review, I expect to have a full review done today or tomorrow.

aarzilli:
Removing the WIP tag, I'm satisfied with the way the interface looks now and starlark.Variable stuff can probably be added later with minimal disruption.

dgryski:
This is just waiting code review from @derekparker then?

derekparker:
Sorry for the delay everyone, this is a large PR and I've been heads down on some big projects at $WORK. I will review this early next week as I've finally gotten through some of the other PR backlog.

(cc @aarzilli @dgryski )

Documentation/cli/starlark.md
```diff
@@ -0,0 +1,265 @@
...
+Variables of the target program can be accessed using `local_vars`, `function_args` or the `eval` functions. Each variable will be returned as a [Variable](https://godoc.org/github.com/go-delve/delve/service/api#Variable) struct, with one special field: `Value`.
+
+## Variable.Value
+
+The `Value` field will return the value of the target variable converted to a starlark value:
+
+* integers, floating point numbers and strings are represented by equivalent starlark values
+* structs are represented as starlark dictionaries
+* slices and arrays are represented by starlark lists
+* maps are represented by starlark lists where even entries are keys and odd entries are values
```
derekparker: Why not convert to a dictionary instead to preserve the same semantics?

Documentation/cli/starlark.md
```diff
@@ -0,0 +1,265 @@
...
+	A int
+	B int
+}
+
+s2 := []astruct{{1, 2}, {3, 4}, {5, 6}, {7, 8}, {9, 10}, {11, 12}, {13, 14}, {15, 16}}
+```
+
+The following is possible:
+
+<!-- TODO: rewrite all of this -->
```
derekparker: When / why?

pkg/terminal/starlark.go
```diff
@@ -0,0 +1,59 @@
+package terminal
+
+import (
+	"github.com/go-delve/delve/pkg/terminal/starbind"
+	"github.com/go-delve/delve/service"
+	"github.com/go-delve/delve/service/api"
+)
+
+type starlarkContext struct {
+	t *Term
```
derekparker: nit: prefer `term` over just `t`.

pkg/terminal/starlark.go
```diff
@@ -0,0 +1,59 @@
...
+	return ctx.t.client
+}
+
+func (ctx starlarkContext) RegisterCallback(name, helpMsg string, fn func(args string) error) {
+	cmdfn := func(t *Term, ctx callContext, args string) error {
+		return fn(args)
+	}
+
+	found := false
+	for i := range ctx.t.cmds.cmds {
```
derekparker: It bothers me how redundant `cmds.cmds` is...

pkg/terminal/starbind/starlark.go
```diff
@@ -0,0 +1,286 @@
...
+	resolve.AllowBitwise = true
+	resolve.AllowRecursion = true
+	resolve.AllowGlobalReassign = true
+}
+
+// Context is the context in which starlark scripts are evaluated.
+// It contains methods to call API functions, command line commands, etc.
+type Context interface {
+	Client() service.Client
+	RegisterCallback(name, helpMsg string, cmdfn func(args string) error)
```
derekparker: nit: This would be more clear as `RegisterCommand`.

Documentation/cli/starlark.md
aarzilli: Already did, I just forgot to remove the comment.

pkg/terminal/starlark.go
aarzilli: Done.

pkg/terminal/starbind/starlark.go
aarzilli: Done.

Documentation/cli/starlark.md
aarzilli: We can do it, I thought this was good enough.

pkg/terminal/starbind/starlark.go
```diff
@@ -0,0 +1,286 @@
...
+// Source can be either a []byte, a string or a io.Reader. If source is nil
+// Execute will execute the file specified by 'path'.
+// After the file is executed if a function named mainFnName exists it will be called, passing args to it.
+func (env *Env) Execute(path string, source interface{}, mainFnName string, args []interface{}) (starlark.Value, error) {
+	defer func() {
+		err := recover()
+		if err == nil {
+			return
+		}
+		fmt.Printf("panic executing starlark script: %v\n", err)
```
derekparker: We should use our logging setup for this message.

pkg/terminal/starbind/starlark.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	env.contextMu.Lock()
+	var ctx context.Context
+	ctx, env.cancelfn = context.WithCancel(context.Background())
+	env.contextMu.Unlock()
+	thread.SetLocal(dlvContextName, ctx)
+	return thread
+}
+
+func (env *Env) createCallback(name string, val starlark.Value) error {
```
derekparker: This should be changed for consistency with `RegisterCommand`.

aarzilli: Done

Documentation/cli/starlark.md
derekparker: I think it just breaks expectations. If you're using the starlark API and you want to interact with a map, you'd expect to use normal map operations, not necessarily dealing with an array.

aarzilli: Ok, I'll implement it.

aarzilli: Done.

scripts/gen-starlark-bindings.go
```diff
@@ -0,0 +1,351 @@
...
+			fmt.Fprintf(buf, "}\n")
+		}
+
+		fmt.Fprintf(buf, "err := env.ctx.Client().CallAPI(%q, &rpcArgs, &rpcRet)\n", binding.fn.Name())
+		fmt.Fprintf(buf, "if err != nil { return starlark.None, err }\n")
+		fmt.Fprintf(buf, "return env.interfaceToStarlarkValue(rpcRet), nil\n")
+
+		fmt.Fprintf(buf, "})\n")
+
+		/*
```
derekparker: nit: cleanup commented code

derekparker:
@aarzilli per our discussion on IRC the only outstanding item left on this PR is enabled commands defined in the REPL to be used during the active debug session.

Other than that this LGTM and should be merged tomorrow.

aarzilli:
I did the thing.

PR #1478, opened 2019-02-10
service/debugger: support relative paths in location expressions

### LINKED ISSUE #1474: Breakpoint filename resolution does not respect relative paths
When attempting to set a breakpoint within a file that has the same name as a standard library file, I'm met with the following.

```
(dlv) break reader.go:176
Command failed: Location "reader.go:176" ambiguous: /home/nick/Documents/code/lolv/logread/reader.go, /usr/lib64/go/1.11/src/bytes/reader.go, /usr/lib64/go/1.11/src/net/textproto/reader.go, /usr/lib64/go/1.11/src/strings/reader.go…
```
I was hoping I could just do `break ./reader.go:176`, but this does not work.

```
(dlv) break ./reader.go:176
Command failed: Location "./reader.go:176" not found
```

However, if I give the absolute path, it's clear that delve can somehow resolve these relative paths

```
(dlv) break /home/nick/Documents/code/lolv/logread/reader.go:176
Breakpoint 1 set at 0x54a544 for github.com/ollien/lolv/logread.(*LogReader).ReadBackwardsToStartOfObject.func1() ./reader.go:176
```

Note that the output above quite clearly lists `./reader.go`. It would be nice if `break ./reader.go:176` worked as expected. Pasting the full path when I want a breakpoint can be quite frustrating.

```
Delve Debugger
Version: 1.1.0
```
```
go version go1.11.4 linux/amd64
```
OpenSUSE Tumbleweed on x86_64

chainhelen:
This change try to match relative path according to
the directory of ptraced executable program when loc.FileMatch fail.

Fixes #1474  

En, If you don't support this pr, I will close it.

### DIFF
--- a/service/debugger/locations.go
+++ b/service/debugger/locations.go
@@ -3,6 +3,7 @@ package debugger
 import (
 	"fmt"
 	"go/constant"
+	"path"
 	"path/filepath"
 	"reflect"
 	"runtime"
@@ -293,6 +294,10 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
 	return partialPathMatch(loc.Base, path)
 }
 
+func tryMatchRelativePathByProc(expr, debugname, file string) bool {
+	return len(expr) > 0 && expr[0] == '.' && file == path.Join(path.Dir(debugname), expr)
+}
+
 func partialPathMatch(expr, path string) bool {
 	if runtime.GOOS == "windows" {
 		// Accept `expr` which is case-insensitive and slash-insensitive match to `path`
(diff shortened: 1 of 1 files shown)


service/debugger/locations.go
```diff
@@ -293,6 +294,10 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
 	return partialPathMatch(loc.Base, path)
 }
 
+func tryMatchRelativePathByProc(expr, debugname, path string) bool {
+	return path == Path.Join(Path.Dir(debugname), expr)
```
aarzilli: Rename `path` to `file` and then `return file == path.Join(filepath.Dir(debugname), expr)`. So we don't have to rename the import of "path".

Also I think this should only be done if `expr[0] == '.'`, it doesn't do anything useful otherwise.

chainhelen:
Done, thx for your guidance.

Issue #213, opened 2015-08-19
Add support for FreeBSD


tsgan:
```
go get -u github.com/derekparker/delve/cmd/dlv
# github.com/derekparker/delve/proc
../../../github.com/derekparker/delve/proc/proc.go:43: undefined: OSProcessDetails
```

derekparker:
Unfortunately FreeBSD is not a supported platform at the moment.

derekparker:
Actually, let me rename and reopen for adding FreeBSD support.

tpltnt:
For completeness sake, retrieving the dependencies fails like this:
```
% go get -v -u github.com/peterh/liner github.com/derekparker/delve/cmd/dlv
github.com/peterh/liner (download)
github.com/derekparker/delve (download)
Fetching https://gopkg.in/yaml.v2?go-get=1
Parsing meta tags from https://gopkg.in/yaml.v2?go-get=1 (status code 200)
get "gopkg.in/yaml.v2": found meta tag main.metaImport{Prefix:"gopkg.in/yaml.v2", VCS:"git", RepoRoot:"https://gopkg.in/yaml.v2"} at https://gopkg.in/yaml.v2?go-get=1
gopkg.in/yaml.v2 (download)
Fetching https://golang.org/x/debug/dwarf?go-get=1
Parsing meta tags from https://golang.org/x/debug/dwarf?go-get=1 (status code 200)
get "golang.org/x/debug/dwarf": found meta tag main.metaImport{Prefix:"golang.org/x/debug", VCS:"git", RepoRoot:"https://go.googlesource.com/debug"} at https://golang.org/x/debug/dwarf?go-get=1
get "golang.org/x/debug/dwarf": verifying non-authoritative meta tag
Fetching https://golang.org/x/debug?go-get=1
Parsing meta tags from https://golang.org/x/debug?go-get=1 (status code 200)
golang.org/x/debug (download)
Fetching https://rsc.io/x86/x86asm?go-get=1
Parsing meta tags from https://rsc.io/x86/x86asm?go-get=1 (status code 200)
get "rsc.io/x86/x86asm": found meta tag main.metaImport{Prefix:"rsc.io/x86", VCS:"git", RepoRoot:"https://github.com/rsc/x86"} at https://rsc.io/x86/x86asm?go-get=1
get "rsc.io/x86/x86asm": verifying non-authoritative meta tag
Fetching https://rsc.io/x86?go-get=1
[...]

PR #1480, opened 2019-02-15
FreeBSD initial support

rayrapetyan:
This pull request adds initial support for FreeBSD.
It's primarily based on fbsd branch from: github.com:asomers/delve.git, just fixes few critical bugs and adds few new functions.

It passes almost all tests (281 of 288), except few "Concurrency" ones, but I believe it's not a bug, but a natural behavior for FreeBSD and these tests should be slightly changed.

E.g. on "TestBreakpointCounts" two threads are expected to hit same breakpoint 100 times each. In FreeBSD, when main thread exits, all "child" threads are not triggering TRAP signals anymore, process just finishes. That's why when main thread reaches 100 breakpoint hits and exits, child thread will never reach 100 and stops somewhere around 95. Sometimes, when child thread is faster and completes BP race first, we can see this test as PASSED. 

Other than that it works fine.

ref: https://github.com/go-delve/delve/issues/213

Thanks.


aarzilli:
In bpcountstest.go (the fixture used by TestBreakpointCounts) the main goroutine specifically waits for the two child goroutines before exiting. If TestBreakpointCounts doesn't pass either there's a bug in your backend implementation or in Go. I think it's more likely that there's a bug here.

TestBreakpointCounts actually checks that the backend correctly detects when the same breakpoint is hit simultaneously on two different threads. Not doing this correctly is a common mistake, it was made by different people on the darwin, linux and windows native backend. It's likely that you have the same problem.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+	}
+	// all threads are resumed
+	PtraceCont(dbp.pid, 0)
+	return nil
+}
+
+// Used by ContinueOnce
+// stop stops all running threads and sets breakpoints
+func (dbp *Process) stop(trapthread *Thread) (err error) {
```
aarzilli: How do threads other than the trapthread get stopped? Since you aren't callling Thread.stop here.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,368 @@
...
+	StatusStopped  = 4
+	StatusZombie   = 5
+	StatusWaiting  = 6
+	StatusLocked   = 7
+)
+
+// OSProcessDetails contains FreeBSD specific
+// process details.
+type OSProcessDetails struct {
+	comm string
```
aarzilli: I can't find where this is used.

pkg/proc/native/proc_freebsd.c
```diff
@@ -0,0 +1,93 @@
...
+		command_name = malloc(COMMLEN + 1);
+		if (command_name != NULL)
+			strlcpy(command_name, kinfo->ki_comm, COMMLEN + 1);
+		free(kinfo);
+	}
+
+	return (command_name);
+}
+
+int find_status(int pid){
```
aarzilli: Looking at the [procfs manpage](https://www.freebsd.org/cgi/man.cgi?query=procfs&apropos=0&sektion=0&manpath=FreeBSD+10.4-RELEASE&arch=default&format=html) it seems to me that this and find_executable could be implemented using that. Am I wrong?

pkg/proc/native/proc_freebsd.go
rayrapetyan: In FreeBSD, whenever one thread is trapped, the entire process (all threads) are stopped. Since we call "stop" only after "trapWait" (in ContinueOnce), all threads are always stopped, so no need to explicitly check for that.

pkg/proc/native/proc_freebsd.c
rayrapetyan: It is not recommended to use procfs in FreeBSD as it requires some additional explicit changes in /etc/fstab (https://www.freebsd.org/doc/en/articles/linux-users/procfs.html), not all users have this Linux-compatibility option enabled. Using procstat calls is a recommended way to get info.

pkg/proc/native/proc_freebsd.go
rayrapetyan: Seems it's not used anywhere. I've tried to use tid (main process thread id) in singleStep mimicking comparison in line 59 from linux_threads.go:
`if (status == nil || status.Exited()) && wpid == t.dbp.pid`
But later replaced all logic in singleStep with a single call to trapWait(). Btw can we do the same for Linux? The problem with Linux's singleStep implementation is that it doesn't handle threads add\remove properly. I see that we are waiting for a TRAP from "this" thread ignoring signals from other threads. Maybe we can just pass tid into trapWait as additional param to simulate the same?

aarzilli: > Btw can we do the same for Linux?

see the comment in (*Process).wait in proc_linux.go

pkg/proc/native/proc_freebsd.c
aarzilli: Is it possible to do this without cgo?

rayrapetyan: I believe no (except procfs which requires additional explicit mount).

pkg/proc/native/proc_freebsd.go
rayrapetyan: The only comment there is about a workaround for lock when main thread exits and leaves zombies. Could you clarify how does that rely to replacing logic in singleStep() with a call to waitTrap()?

aarzilli: The full wait is slow because of the workaround. Also all threads are stopped and we are only executing one instruction of threads stopped on a breakpoint, so it's fine.

rayrapetyan: On FreeBSD, even when all threads are stopped, wait in cycle may return multiple queued signals from multiple stopped threads.
 
I'm preparing a detailed step by step explanation of why "TestBreakpointCounts" fails, hopefully someone can help with a fix.

rayrapetyan:
So far I was able to reproduce "TRAP events disappear" issue with a pure C code, without go stuff involved. Under some circumstances, app is exiting without delivering queued TRAP events to debugger. Preparing a message to freebsd-hackers group.

rayrapetyan:
FreeBSD bug created: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236035

aarzilli:
Thank you, let me know how it goes.

derekparker:
Finished a cursory review this morning, will be doing a more in depth review tomorrow or the following day. In the meantime, I saw some commented out code, I can just add a blanket statement now to clean all that up.

By the way, thanks for submitting this patch!

aarzilli:
Regarding the discussion around TestBreakpointCounts I'm ok with disabling that test on freebsd and merging this now to prevent this PR from coderotting. There's also a free CI service that has FreeBSD servers: https://cirrus-ci.org/guide/FreeBSD/.

pkg/proc/native/threads_freebsd.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (t *Thread) singleStep() (err error) {
+	for {
+		t.dbp.execPtraceFunc(func() { err = PtraceSingleStep(t.ID) })
+		if err != nil {
+			return err
+		}
+
+		t.dbp.trapWait(t.dbp.pid)
+		/*_, status, err := t.dbp.waitFast(t.dbp.pid)
```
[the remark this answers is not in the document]
rayrapetyan: The problem here is that this logic seems to be broken in any existing implementation, e.g. in Linux waitFast is used, then exiting conditions checked, but there is no proper handling of thread creation\death in this place. trapWait handles all these cases properly.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,368 @@
...
+		if err := th.resumeWithSig(int(status.StopSignal())); err != nil {
+			if err == sys.ESRCH {
+				return nil, proc.ErrProcessExited{Pid: dbp.pid}
+			}
+			return nil, err
+		}
+	}
+}
+
+func (dbp *Process) loadProcessInformation() {
```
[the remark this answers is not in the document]
rayrapetyan: Removed from here and Linux implementation.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,368 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Used by Detach
+func (dbp *Process) detach(kill bool) error {
+	err := PtraceDetach(dbp.pid, 0)
```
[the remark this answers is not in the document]
rayrapetyan: Fixed.

pkg/proc/native/ptrace_freebsd.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Get the lwpid_t of the thread that caused wpid's process to stop, if any.
+// Return also the full pl_flags variable, which indicates why the process
+// stopped.
+func ptraceGetLwpInfo(wpid int) (tid int, flags int, si_code int, err error) {
+	var info C.struct_ptrace_lwpinfo
+	_, err = C.ptrace_lwp_info(C.int(wpid), &info)
+
+	/*
```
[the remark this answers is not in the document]
rayrapetyan: Removed.

pkg/proc/native/threads_freebsd.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		if tid == t.ID && status.StopSignal() == sys.SIGTRAP {
+			return nil
+		}*/
+		return nil
+	}
+}
+
+func (t *Thread) Blocked() bool {
+	regs, err := t.Registers(false)
```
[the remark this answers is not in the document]
rayrapetyan: Changed.

pkg/proc/native/threads_freebsd.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (t *Thread) singleStep() (err error) {
+	for {
+		t.dbp.execPtraceFunc(func() { err = PtraceSingleStep(t.ID) })
+		if err != nil {
+			return err
+		}
+
+		t.dbp.trapWait(t.dbp.pid)
```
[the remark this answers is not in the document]
rayrapetyan: Fixed.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,358 @@
...
+func (dbp *Process) kill() (err error) {
+	if dbp.exited {
+		return nil
+	}
+
+	if err = sys.Kill(-dbp.pid, sys.SIGKILL); err != nil {
+		return errors.New("could not deliver signal " + err.Error())
+	}
+	// If the process is stopped, we must continue it so it can receive the signal
+	PtraceCont(dbp.pid, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,358 @@
...
+	if dbp.currentThread == nil {
+		dbp.SwitchThread(tid)
+	}
+
+	return dbp.threads[tid], nil
+}
+
+// Used by initialize
+func (dbp *Process) updateThreadList() error {
+	tids := PtraceGetLwpList(dbp.pid)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,358 @@
...
+	for _, thread := range dbp.threads {
+		if thread.CurrentBreakpoint.Breakpoint != nil {
+			if err := thread.StepInstruction(); err != nil {
+				return err
+			}
+			thread.CurrentBreakpoint.Clear()
+		}
+	}
+	// all threads are resumed
+	PtraceCont(dbp.pid, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/threads_freebsd.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (t *Thread) stop() (err error) {
+	_, err = C.thr_kill2(C.pid_t(t.dbp.pid), C.long(t.ID), C.int(sys.SIGSTOP))
+	if err != nil {
+		err = fmt.Errorf("stop err %s on thread %d", err, t.ID)
+		return
+	}
+	// If the process is stopped, we must continue it so it can receive the
+	// signal
+	PtraceCont(t.dbp.pid, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/threads_freebsd.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	for {
+		th, err := t.dbp.trapWait(t.dbp.pid)
+		if err != nil {
+			return err
+		}
+		if th.ID == t.ID {
+			break
+		}
+		PtraceCont(th.ID, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+		th, ok := dbp.threads[tid]
+		if ok {
+			th.Status = (*WaitStatus)(status)
+		}
+
+		if status.StopSignal() == sys.SIGTRAP {
+			if pl_flags&sys.PL_FLAG_EXITED != 0 {
+				delete(dbp.threads, tid)
+				PtraceCont(tid, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,358 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Used by Detach
+func (dbp *Process) detach(kill bool) error {
+	return PtraceDetach(dbp.pid, 0)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

pkg/proc/native/registers_freebsd_amd64.go
```diff
@@ -0,0 +1,92 @@
...
+	var (
+		regs fbsdutil.AMD64PtraceRegs
+		err  error
+	)
+	thread.dbp.execPtraceFunc(func() { err = sys.PtraceGetRegs(thread.ID, (*sys.Reg)(&regs)) })
+	if err != nil {
+		return nil, err
+	}
+	var fsbase int64
+	err = sys.PtraceGetFsBase(thread.ID, &fsbase)
```
derekparker: Must be called via our PTRACE thread (e.g. `execPtraceFunc()`).

rayrapetyan:
@derekparker, wrapped all places except PtraceDetach - it doesn't work that way. Thanks.

pkg/proc/native/proc_freebsd.go
rayrapetyan: Doesn't work that way.

pkg/proc/native/threads_freebsd.go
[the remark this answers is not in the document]
derekparker: Ack, understood and makes sense.

derekparker:
@rayrapetyan thanks for the prompt changes! The last thing I have is to disable any concurrent tests that are failing due to the FreeBSD issue with an informative `t.Skip(...)` and then we should be able to merge this and get FreeBSD CI set up.

rayrapetyan:
@derekparker, there are few weird test cases. When running: 
go test
from pkg/proc folder, two tests always fail:
```
--- FAIL: TestAttachDetach (0.08s)
    proc_test.go:136: Did not continue to correct location, expected line :11 got /usr/local/go/src/runtime/panic.go:681
        	at /ara/devel/blockchain/go/src/github.com/go-delve/delve/pkg/proc/proc_test.go:2887
--- FAIL: TestIssue844 (0.14s)
    proc_test.go:113: failed assertion at proc_test.go:3204: Continue - ptraceGetLwpInfo err no such process -1
i: 0 -> equalsTwo: false 
i: 1 -> equalsTwo: false 
0
fatal error: all goroutines are asleep - deadlock!
```
 
But whenever you run each of these two individually (e.g. go test -run TestAttachDetach) - they always PASS. Any thoughts of why "go test" may perform like that? Thanks.

aarzilli:
The TestAttachDetach could happen if you are running tests in parallel (like `go test ./...`), the other one I don't know, it needs to be debugged.

rayrapetyan:
I've tried `cd pkg/proc; go test -parallel 1` - it also fails...

aarzilli:
It could be that Detach(true) isn't actually killing the target process, one instance of testnextnethttp stays running and makes TestAttachDetach fail.

rayrapetyan:
"Skipped" concurrent tests. Please let me know if I can help to setup FreeBSD CI in Travis.
@aarzilli, sorry, just found your "go mod" related question. Please let me know how to perform vendor code update procedure properly and I'll commit missed fixes. Thanks.

derekparker:
@rayrapetyan thanks! You should just have to run `go mod vendor`.

rayrapetyan:
@derekparker, hm, `go mod vendor` just re-creates original vendor directory vanishing all my changes. Does that mean I should commit everything into original repo first (e.g. into golang.org/x/sys)? Thanks!

derekparker:
@rayrapetyan are you making manual changes to the vendor directory? Yeah, that won't work for the reason you described.

If you've made any changes to any dependency of Delve, you must commit that change upstream and then re-vendor it in Delve (from upstream). I didn't take a close look at what changes were made in the vendor dir, but is it possible to implement without any code changes to our dependencies?

rayrapetyan:
@derekparker, @aarzilli, I've cleared vendor-related changes, the only artifacts left are constants for ptrace and registers-related structs, I believe they should stay in vendors. Please suggest how to proceed. I've never committed into golang core, it requires a special procedure (https://golang.org/doc/contribute.html), so if you have accounts there could you commit these changes? Thanks.

aarzilli:
I'm taking a look at the tests you have disabled. It's quite a lot of tests, if they all need to be disabled it seems that there is something that's pretty thoroughly broken here.

As for the changes to golang.org/x/sys/unix, I don't have a freebsd install and I don't know pretty much anything about it, I can't make them for you. 

For the constants it's easier to copy them to our side. The changes to Stat_t, Dirent, etc are worrying however, that's an ABI change, I don't know that's handled on golang.org/x/sys/unix.

It's not that hard, you just have to follow steps 0 through 4 at https://golang.org/doc/contribute.html. Ignore the rest of that page, then you make a branch for your changes and mail them with `git codereview mail`.

rayrapetyan:
@aarzilli, these are all parallel_next and concurrentstep tests, there are also two other tests I will investigate. Will also proceed with commits into sys/unix, thanks!

aarzilli:
> these are all parallel_next and concurrentstep tests, there are also two other tests I will investigate

My guess is that when a thread executes an INT 3 freebsd will send us immediately a signal: if other threads are running on different cores they will keep running until their quantum exipres, but the signal will be sent immediately. Which means that when (*Process).stop is called some of the threads could still be running and that's why we are missing some breakpoints in all those tests.

rayrapetyan:
golang.org: commit for ptrace support on FreeBSD: https://go-review.googlesource.com/c/go/+/166423

derekparker:
@rayrapetyan what's the status of the above patch? Looks like there might be some outstanding review comments to address before it's considered for merging.

rayrapetyan:
@derekparker, I was stuck with committing changes into golang - along with i386 changes which I provided they require changes for CPU architectures I don't own (ARM) and therefore can't test...

derekparker:
@rayrapetyan It looks like all you need to do is generate the files for ARM. I think if you do that and submit it to the patch, the burden for review and testing will be on them. They can use their own internal builders for testing.

rayrapetyan:
Pushed "generated" ARM files. Awaiting merge.
https://github.com/golang/go/pull/30704

derekparker:
@rayrapetyan looks like that PR was closed because that stdlib package is locked, shoulda caught that earlier. In any event, can you open a PR against `x/sys/unix` if you haven't already? We are already vendoring that so it would be easier to vendor your changes once merged.

rayrapetyan:
@derekparker, awaiting approval on https://github.com/golang/sys/pull/34.

derekparker:
@rayrapetyan ack, thanks.

derekparker:
@rayrapetyan looks like it's been stagnant for a while. I just pinged some folks to get that pushed through soon hopefully.

rayrapetyan:
Yeah, thanks!

derekparker:
@rayrapetyan looks like it was merged! Can you update this PR by vendoring these new changes please.

rayrapetyan:
Done!
Wondering why do we need to carry “vendor” folder in git repo? Can’t we
just make ‘go mod vendor’ an installation step? Thanks!

aarzilli:
>Wondering why do we need to carry “vendor” folder in git repo?

the vendor folder makes delve compile in GOPATH mode and on pre-modules version of go. It also makes this project self-contained, preventing left-pad style accidents.

derekparker:
@rayrapetyan looks like the last update broke everything lol. How did you vendor it?

rayrapetyan:
Oh... weird. I did "go mod vendor", it updated go.mod and go.sum and vendor folder, I've committed these changes, that's it. What had I missed?
Now it fails in sys\windows (some missed definitions which I can see defined in sources), linux: x/crypto (misses definitions which are there, x/crypto separately passes all tests) and something in Mac...

derekparker:
@rayrapetyan does it work when you build and test locally?

rayrapetyan:
Yes, it passes all tests locally in FreeBSD. I can reproduce issue with x/crypto in Ubuntu VM. Do you know how to make x/crypto build correctly?

derekparker:
@rayrapetyan the vendor has been updated with your changes via #1600 thanks to @aarzilli. You should be able to just remove your last commit and rebase on master.

rayrapetyan:
@derekparker, @aarzilli, thanks guys, I've spent several hours trying to make latest x/crypto to work with x/sys with no luck. Hopefully we can merge my branch now.

pkg/proc/fbsdutil/regs.go
```diff
@@ -0,0 +1,416 @@
...
+// GAddr returns the address of the G variable if it is known, 0 and false
+// otherwise.
+func (r *AMD64Registers) GAddr() (uint64, bool) {
+	return 0, false
+}
+
+// Get returns the value of the n-th register (in x86asm order).
+func (r *AMD64Registers) Get(n int) (uint64, error) {
+	reg := x86asm.Reg(n)
+	const (
```
derekparker: These bitmasks are incorrect, see: https://github.com/go-delve/delve/pull/1583.

pkg/proc/fbsdutil/regs.go
```diff
@@ -0,0 +1,416 @@
...
+		regs = proc.AppendSSEReg(regs, fmt.Sprintf("XMM%d", i/16), xsave.XmmSpace[i:i+16])
+		if xsave.AvxState {
+			regs = proc.AppendSSEReg(regs, fmt.Sprintf("YMM%d", i/16), xsave.YmmSpace[i:i+16])
+		}
+	}
+
+	return
+}
+
+const (
```
derekparker: Instead of duplicating this, let's reuse what already exists in `linutil`.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1327,6 +1327,9 @@ func (p *Process) loadGInstr() []byte {
 	case "darwin":
 		// mov rcx,QWORD PTR gs:{uint32(off)}
 		op = []byte{0x65, 0x48, 0x8B, 0x0C, 0x25}
+	case "freebsd":
```
derekparker: Let's clean this up a bit by just merging this into the case above, e.g. `case "darwin", "freebsd":`

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,367 @@
...
+// known threads.
+func (dbp *Process) addThread(tid int, attach bool) (*Thread, error) {
+	if thread, ok := dbp.threads[tid]; ok {
+		return thread, nil
+	}
+
+	var err error
+	dbp.execPtraceFunc(func() { err = sys.PtraceLwpEvents(dbp.pid, 1) })
+	if err == syscall.ESRCH {
+		// XXX why do we wait here?
```
derekparker: Is this still an open question?

rayrapetyan:
@derekparker, updated. An interesting finding - with fixed bitmasks another concurrent test is constantly failing, so I've added it to skipped for now. Definitely bitmasks affect other concurrent tests also. I will look into all skipped tests after July 20.

derekparker:
@rayrapetyan hmm, that's odd. That change didn't break any tests on any currently supported OS.

I have one last request before merging. The `t.Skip(...)` message isn't exactly correct. The tests are valid, just broken on the OS. Please reword and link to relevant issues (like the upstream freebsd issue you created) in a comment. After that I will merge promptly and create some follow up issues to track fixing the broken tests.

derekparker:
@aarzilli I'm planning on merging this by EOD today even if the above comment isn't addressed (and just do my own follow up commit if need be). Is there any concern you have with this PR or are you good with it at this point? 

I have questions on why some of the tests are failing and we need to chase those down (@rayrapetyan has mentioned starting after July 20). I'm eager to get this merged, but as we are also _really_ close to a release and I want to make sure we're not releasing something big like a new OS without very much vetting. The upside of merging soon however is we can start getting early adopters to test and provide feedback.

pkg/proc/native/proc_freebsd.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// Used by stop and trapWait
+func (dbp *Process) trapWaitInternal(pid int, halt bool) (*Thread, error) {
+	for {
+		wpid, status, err := dbp.wait(pid, 0)
+		if err != nil {
+			return nil, fmt.Errorf("wait err %s %d", err, pid)
+		}
+		if status.Killed() {
```
derekparker: I wonder if this has anything to do with the failing concurrent tests, since we're effectively ignoring `status.Killed`, yes we send `SIGKILL` to the process in `process.kill()`. I have a freebsd VM I can test this theory out on tomorrow.

derekparker:
@rayrapetyan a huge thanks to you for this port! Now looking forward to getting the concurrency issues sorted out.

PR #1498, opened 2019-02-26
travis.yml: add 1.12 to test matrix

aarzilli:
```
travis.yml: add 1.12 to test matrix

```


aarzilli:
Should we still keep 1.9?

derekparker:
Merged, I don't think we need to keep test matrix support for 1.9.

PR #1502, opened 2019-02-27
proc: catch fatal runtime errors

aarzilli:
```
proc: catch fatal runtime errors

Like we do with unrecovered panics, create a default breakpoint to
catch runtime errors that will cause the program to terminate.
Primarily intended to give users the opportunity to examine the state
of a deadlocked process.

```


aarzilli:
As it turns out this used to work in <=1.10 so it's more of a fix for a regression than a new feature.

PR #1504, opened 2019-02-27
More Function Calls, parts 2

### LINKED ISSUE #826: Setting a string value is not currently supported
1.0.0 RC1
1.8.1
Windows 10 / amd64
4. What did you do?
Tried to set a variable of type string
The new value.
6. What did you see instead?
```
(dlv) set myEnvVar = "some val"
Command failed: can not set variables of type string (not implemented)
```

I understand that this is currently not implemented but because there's also no issue to keep track of it I thought it might be a good idea to at least raise one.

Thank you.

aarzilli:
```
proc: bugfix: closure addr was wrong for non-closure functions

proc: support nested function calls

Changes the code in fncall.go to support nested function calls.

This changes delays argument evaluation until after we have used
the call injection protocol to allocate an argument frame. When
evaluating the parse tree of an expression we'll initiate each
function call we find on the way down and then complete the function
call on the way up.

For example. in:

f(g(x))

we will:

1. initiate the call injection protocol for f(...)
2. progress it until the point where we have space for the arguments
of 'f' (i.e. when we receive the debugCallAXCompleteCall message
from the target runtime)
3. inititate the call injection protocol for g(...)
4. progress it until the point where we have space for the arguments
of 'g'
5. copy the value of x into the argument frame of 'g'
6. finish the call to g(...)
7. copy the return value of g(x) into the argument frame of 'f'
8. finish the call to f(...)

Updates #119

proc: allow function calls to appear inside an expression

The initial implementation of the 'call' command required the
function call to be the root expression, i.e. something like:

double(3) + 1

was not allowed, because the root expression was the binary operator
'+', not the function call.

With this change expressions like the one above and others are
allowed.

This is the first step necessary to implement nested function calls
(where the result of a function call is used as argument to another
function call).

This is implemented by replacing proc.CallFunction with
proc.EvalExpressionWithCalls. EvalExpressionWithCalls will run
proc.(*EvalScope).EvalExpression in a different goroutine. This
goroutine, the 'eval' goroutine, will communicate with the main
goroutine of the debugger by means of two channels: continueRequest
and continueCompleted.

The eval goroutine evaluates the expression recursively, when
a function call is encountered it takes care of setting up the
function call on the target program and writes a request to the
continueRequest channel, this causes the 'main' goroutine to restart
the target program by calling proc.Continue.

Whenever Continue encounters a breakpoint that belongs to the
function call injection protocol (runtime.debugCallV1 and associated
functions) it writes to continueCompleted which resumes the 'eval'
goroutine.

The 'eval' goroutine takes care of implementing the function call
injection protocol.

When the expression is fully evaluated the 'eval' goroutine will
write a special message to 'continueRequest' signaling that the
expression evaluation is terminated which will cause Continue to
return to the user.

Updates #119

```

### DIFF
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -423,7 +423,7 @@ func funcCallEvalFuncExpr(scope *EvalScope, fncall *functionCallState, allowCall
 	if !fncall.fn.cu.isgo {
 		return errNotAGoFunction
 	}
-	fncall.closureAddr = fnvar.funcvalAddr()
+	fncall.closureAddr = fnvar.closureAddr
 
 	fncall.argFrameSize, fncall.formalArgs, err = funcCallArgs(fncall.fn, bi, false)
 	if err != nil {
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -39,6 +39,7 @@ const (
 	debugCallFunctionNamePrefix1 = "debugCall"
 	debugCallFunctionNamePrefix2 = "runtime.debugCall"
 	debugCallFunctionName        = "runtime.debugCallV1"
+	maxArgFrameSize              = 65535
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/proc/fncall.go
```diff
@@ -391,48 +383,69 @@ func callOP(bi *BinaryInfo, thread Thread, regs Registers, callAddr uint64) erro
...
-func (scope *EvalScope) funcCallEvalExpr(callexpr *ast.CallExpr) (fn *Function, closureAddr uint64, argvars []*Variable, err error) {
+// funcCallEvalFuncExpr evaluates expr.Fun and returns the function that we're trying to call.
+// If allowCalls is false function calls will be disabled even if scope.callCtx != nil
+func funcCallEvalFuncExpr(scope *EvalScope, fncall *functionCallState, allowCalls bool) error {
 	bi := scope.BinInfo
 
-	fnvar, err := scope.evalAST(callexpr.Fun)
-	if err != nil {
-		return nil, 0, nil, err
+	if !allowCalls {
```
derekparker: Why not just bail here instead of continuing if function calls are disabled?

aarzilli: It can still succeed and it will in most cases. It will only fail for something like `f()()` where the left side of a function call expression requires itself a nested function call, for those we have to evaluate the nested call (to `f()`) after we started the outer call (to the return value of `f`) so that we have space on the stack to save the return value of `f`.

PR #1506, opened 2019-03-01
Add trimpath flags and support LDFLAGS

Foxboron:
Reproducible builds requires any unique paths to be trimmed out. So we
add the trimpaths for gcflags and asmflags to achieve this.

The second issue is the lack of support for `LDFLAGS` a lot of
distributions and build systems export to enable important compile
settings such as RELRO.

### DIFF
--- a/scripts/make.go
+++ b/scripts/make.go
@@ -236,11 +236,17 @@ func buildFlags() []string {
 	if err != nil {
 		log.Fatal(err)
 	}
+	gctrimpath := "-gcflags=all=-trimpath=${PWD}"
+	asmtrimpath := "-asmflags=all=-trimpath=${PWD}"
 	ldFlags := "-X main.Build=" + strings.TrimSpace(string(buildSHA))
 	if runtime.GOOS == "darwin" {
 		ldFlags = "-s " + ldFlags
 	}
-	return []string{fmt.Sprintf("-ldflags=%s", ldFlags)}
+	extldFlags := os.Getenv("LDFLAGS")
+	if extldFlags != "" {
+		ldFlags = ldFlags + fmt.Sprintf(" -extldflags %s", extldFlags)
+	}
+	return []string{gctrimpath, asmtrimpath, fmt.Sprintf("-ldflags='%s'", ldFlags)}
 }
 
 func testFlags() []string {


scripts/make.go
```diff
@@ -236,11 +236,17 @@ func buildFlags() []string {
 	if err != nil {
 		log.Fatal(err)
 	}
+	gctrimpath := "-gcflags=all=-trimpath=${PWD}"
```
aarzilli: I don't want this always on (or even by default).

scripts/make.go
```diff
@@ -236,11 +236,17 @@ func buildFlags() []string {
...
+	asmtrimpath := "-asmflags=all=-trimpath=${PWD}"
 	ldFlags := "-X main.Build=" + strings.TrimSpace(string(buildSHA))
 	if runtime.GOOS == "darwin" {
 		ldFlags = "-s " + ldFlags
 	}
-	return []string{fmt.Sprintf("-ldflags=%s", ldFlags)}
+	extldFlags := os.Getenv("LDFLAGS")
+	if extldFlags != "" {
+		ldFlags = ldFlags + fmt.Sprintf(" -extldflags %s", extldFlags)
```
aarzilli: I think delve is usually linked internally and this won't do anything (usually).

Foxboron: You'd think so, but we have been trying to compile all go packages on Arch with full RELRO, and funny enough this is needed for this.

```
λ delve foxboron/fix-script» go build -i -v -buildmode=pie -o dlv ./cmd/dlv 
λ delve foxboron/fix-script» checksec -f dlv                               
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable  FILE
Partial RELRO   Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   19406 Symbols	^C
λ delve foxboron/fix-script» go build -i -v -buildmode=pie -ldflags "-extldflags $LDFLAGS" -o dlv ./cmd/dlv 
λ delve foxboron/fix-script» checksec -f dlv                                                               
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable  FILE
Full RELRO      Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   19405 Symbols	
```

scripts/make.go
Foxboron: How do you want this implemented in that case? There is no good reason to keep compilation paths in the binary.

https://reproducible-builds.org/docs/build-path/

scripts/make.go
aarzilli: Sure, if you pass buildmode=pie, but make.go isn't.

scripts/make.go
aarzilli: I'm ok with having a `-trimpath` flag.

scripts/make.go
Foxboron: Yes. Arch personally hacks this by patching `go-pie`. However, extra parsing is required  to utilize `GOFLAGS`, so it's easier for any distribution involved to just pass `LDFLAGS` directly.

aarzilli: That's weird, but ok.

Foxboron: `export GOFLAGS="-ldflags=-extldflags=-zrelro -ldflags=-extldflags=-znow"` Is essentially the format.

`export GOFLAGS="-ldflags=-extldflags=$LDFLAGS"` will result in the `GOFLAGS` parser breaking on you. So things get tedious as they are not following the CLI format.

derekparker:
I think that instead of enforcing a particular flag, we should allow for callers to provide their own flags and make the build step a little more extensible.

Foxboron:
`-trimpath` has finally(!) been added to `go` 1.13, so I guess we can block this PR and fix up the code when the next release is out.

However, building with default flags and let developers picks other ones is something `Makefiles` has been doing well for years. I'm not sure why projects decide it's better to write custom scripts for this.

PR #1533, opened 2019-04-02
cmd/dlv: add Go version check

aarzilli:
```
cmd/dlv: add Go version check

Before doing anything check that the version of Go is compatible with
the current version of Delve.
This will improve the error message in the case that  another change as
disruptive as Go1.11 dwarf compression, happens.

```

### DIFF
--- a/pkg/goversion/compat.go
+++ b/pkg/goversion/compat.go
@@ -1,0 +1,28 @@
+package goversion
+
+import (
+	"fmt"
+)
+
+var (
+	minSupportedVersionOfGoMinor = 10
+	maxSupportedVersionOfGoMinor = 12
+	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version 1.%d, suppress this error with --check-go-version=false)", minSupportedVersionOfGoMinor)
+	dlvTooOldErr                 = fmt.Errorf("Version of Delve is too old for this version of Go (maximum supported version 1.%d, suppress this error with --check-go-version=false)", maxSupportedVersionOfGoMinor)
+)
+
+// Compatible checks that the version specified in the producer string is compatible with
+// this version of delve.
+func Compatible(producer string) error {
+	ver := parseProducer(producer)
+	if ver.IsDevel() {
+		return nil
+	}
+	if !ver.AfterOrEqual(GoVersion{1, minSupportedVersionOfGoMinor, -1, 0, 0, ""}) {
+		return goTooOldErr
+	}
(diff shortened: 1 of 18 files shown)


pkg/goversion/compat.go
```diff
@@ -0,0 +1,28 @@
...
+// this version of delve.
+func Compatible() error {
+	ver, ok := Installed()
+	if ver.IsDevel() || !ok {
+		return nil
+	}
+	if !ver.AfterOrEqual(GoVersion{1, minSupportedVersionOfGoMinor, -1, 0, 0, ""}) {
+		return goTooOldErr
+	}
+	if ver.AfterOrEqual(GoVersion{1, maxSupportedVersionOfGoMinor + 1, -1, 0, 0, ""}) {
```
dlsniper: Wouldn't this not catch issues like the Go 1.11 one?
E.g. if a breaking change happens in Go 1.13 then this check would still happily pass, when instead it should fail because we know nothing about it at the time of compilation for this version of delve.
Also, would this change check against the Go version that compiled the target process than the Go version that compiled delve?

aarzilli: The check only passes for 1.10 through 1.12 which we know we support at this time. 1.13 would be rejected, until we make a version of delve that specifically supports it.

dlsniper: 🤦‍♂️ sorry, misread the condition.
Finally, does it take the go version of the compiled binary or of the system where delve is running?

aarzilli: it takes the version returned by the go command, if it exists. I could see maybe only doing this for debug and test, or adding an option to override it.

dlsniper: If I remember correctly, the go version used to compile the binary is stored in the binary itself, so perhaps it would be doable to read it from there instead and have it as a single check, regardless of how delve debugs the target (via exec, debug, or test)?

aarzilli:
Moved the check in debugger/service and added an override flag, just in case. I didn't do it on attach, if we distrubed a process enough to attach I think it's best to just keep going.

derekparker:
Since this results in a fatal error I suggest we put some more info in the error message such as `ignore with --check-go-version=false`.

aarzilli:
Done.

nd:
@aarzilli I wonder how unsafe is to suppress the check? Asking because some GoLand users still on go1.9. Maybe we can disable some features in UI and provide at least some way to debug programs built by older go versions.

aarzilli:
@nd it is not unsafe: going backwards we just don't test it anymore so we don't know if it works, going forward we'd like to nudge users towards updating delve when they update go.

Issue #1540, opened 2019-04-24
Enable non-invasive debugging


ajwerner:
Using `dlv attach` to connect to a running process can be a powerful debugging tool when you catch unexpected conditions in the wild. it can be convenient to run an init script which attempts to be minimally invasive inside the host process. Consider the following init script:

```
break getPartLocked
continue
locals
clearall
exit
```

The hope of such an init script is to get in and out of the process as quickly as possible in an attempt to not disrupt the interesting state that brought you to attach the debugger in the first place. I'd love some mode where if delve finds itself with the process paused for some amount of time it just detaches gracefully. This can happen due now to a syntax error, or other error I suppose. Maybe just an error handling mode would go a long way. 

Right now however the biggest problem with the above script is the need to furiously type `n` when the exit is reached in order to not tear down the attached process. One might expect that the `-c` flag for exit would allow a user to bypass the teardown prompt.

aarzilli:
I think this was implemented by 7151c6cbcc555c8d15a7f959adc37565d9aed6f2

PR #1541, opened 2019-04-24
terminal/command: permit '-c' flag in 'exit' when attached to a process

ajwerner:
This change permits the '-c' flag to be used in the 'exit' command which enables --init scripts which are run in a terminal to exit without prompting the user.

Informs #1540.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1919,8 +1919,8 @@ func (ere ExitRequestError) Error() string {
 
 func exitCommand(t *Term, ctx callContext, args string) error {
 	if args == "-c" {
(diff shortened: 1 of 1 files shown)


ajwerner:
There didn't seem to be any handy infrastructure lying around so I didn't write a test. I did manually verify it works with great delight.

I could imagine something where which execs some go program that it has lying around (the dlv binary maybe?) that will block on a read call on a file descriptor and then I could execute dlv to attach to itself with another init file that tells it to `exit -c`.

derekparker:
Closing this as there hasn't been any response to @aarzilli comment in a while. Please reopen when comment is addressed.

PR #1546, opened 2019-05-07
Go 1.13 support

aarzilli:
```
goversion: update maximum supported version

proc: allow signed integers as shift counts

Go1.13 allows signed integers to be used as the right hand side of a
shift operator, change eval to match.

tests: fix tests for Go 1.13

- Go 1.13 doesn't autogenerate init functions anymore, tests that
expected that now fail and should be skipped.
- Plugin tests now need -gcflags'all=-N -l' now, we were probably
getting lucky with -gcflags='-N -l' before.

```


aarzilli:
With beta1 released I think it's safe to merge this.

PR #1548, opened 2019-05-10
proc: support setting string values when it requires an allocation

### LINKED ISSUE #826: Setting a string value is not currently supported

aarzilli:
```
proc: support setting string values when it requires an allocation

Allow changing the value of a string variable to a new literal string,
which requires calling runtime.mallocgc to allocate the string into the
target process.

This means that a command like:

call f("some string")

is now supported.

Additionally the command:

call s = "some string"

is also supported.

Fixes #826

```

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -9,6 +9,7 @@ import (
 	"go/constant"
 	"go/parser"
 	"go/printer"
+	"go/scanner"
 	"go/token"
 	"reflect"
 	"strconv"
@@ -28,6 +29,13 @@ func (scope *EvalScope) EvalExpression(expr string, cfg LoadConfig) (*Variable, 
 		defer close(scope.callCtx.continueRequest)
 	}
 	t, err := parser.ParseExpr(expr)
+	if eqOff, isAs := isAssignment(err); scope.callCtx != nil && isAs {
+		lexpr := expr[:eqOff]
+		rexpr := expr[eqOff+1:]
+		err := scope.SetVariable(lexpr, rexpr)
+		scope.callCtx.doReturn(nil, err)
+		return nil, err
+	}
 	if err != nil {
 		scope.callCtx.doReturn(nil, err)
 		return nil, err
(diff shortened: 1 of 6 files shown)


aarzilli:
Previously part of https://github.com/go-delve/delve/pull/1504 (which it depends on)

derekparker:
Needs rebase then I'll review.

aarzilli:
Rebased

pkg/proc/eval.go
```diff
@@ -28,6 +29,13 @@ func (scope *EvalScope) EvalExpression(expr string, cfg LoadConfig) (*Variable,
 		defer close(scope.callCtx.continueRequest)
 	}
 	t, err := parser.ParseExpr(expr)
+	if el, isScannerErr := err.(scanner.ErrorList); scope.callCtx != nil && isScannerErr && el[0].Msg == "expected '==', found '='" {
```
derekparker: nit: we should put this logic into a function that's a bit more descriptive, such as `isAssignment`.

pkg/proc/fncall.go
```diff
@@ -731,6 +746,14 @@ func funcCallStep(callScope *EvalScope, fncall *functionCallState) bool {
 		fncall.retvars = filterVariables(fncall.retvars, func(v *Variable) bool {
 			return (v.Flags & VariableReturnArgument) != 0
 		})
+		if fncall.fn.Name == "runtime.mallocgc" && fncall.retvars[0].Unreadable != nil {
+			// return values never have a location for optimized functions and the
+			// runtime is always optimized. However we want to call runtime.mallocgc,
+			// so we fix the address of the return value manually.
+			fncall.retvars[0].Unreadable = nil
+			lastArg := fncall.formalArgs[len(fncall.formalArgs)-1]
+			fncall.retvars[0].Addr = uintptr(retScope.Regs.CFA + lastArg.off + 8)
```
derekparker: s/8/arch.PtrSize/

aarzilli:
Done.

PR #1556, opened 2019-05-22
Less confusing error message for requesting break on blank line.

tschundler:
The current wording is confusing - the file exists and the line exists, so what is the problem? I suspect this ambiguity is behind go-delve#1496 and likely others. It has lead to confusion both for myself and today a co-worker.

Also I updated the style to return values like the rest of the code in the file, which is also more readable (IMO and per https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -381,28 +381,30 @@ func (bi *BinaryInfo) PCToLine(pc uint64) (string, int, *Function) {
 
 // LineToPC converts a file:line into a memory address.
 func (bi *BinaryInfo) LineToPC(filename string, lineno int) (pc uint64, fn *Function, err error) {
+	fileFound := false
 	for _, cu := range bi.compileUnits {
 		if cu.lineInfo.Lookup[filename] != nil {
-			pc = cu.lineInfo.LineToPC(filename, lineno)
+			fileFound = true
+			pc := cu.lineInfo.LineToPC(filename, lineno)
 			if pc == 0 {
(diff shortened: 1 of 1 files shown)


pkg/proc/bininfo.go
```diff
@@ -381,28 +381,30 @@ func (bi *BinaryInfo) PCToLine(pc uint64) (string, int, *Function) {
...
+				return pc, fn, nil
 			}
 		}
 	}
-	err = fmt.Errorf("could not find %s:%d", filename, lineno)
-	return
+	if fileFound {
+		return 0, nil, fmt.Errorf("could not find %s:%d", filename, lineno)
+	} else {
+		return 0, nil, fmt.Errorf("could not find statement at %s:%d, please use a line with a statement", filename, lineno)
```
aarzilli: Wouldn't `"could not find file %s"` be clearer?

derekparker: It feels like the two messages should be flipped, right? If the file is found, then it is likely the user specified a location that didn't correlate to any actual code, thus this error message makes more sense for that case.

tschundler: Yes, that was a mistake - I"m sorry about the hasty commit, editing in github editor. I properly checked out the code and fixed the logic and and validated that I got the correct.

derekparker: I don't think the commit got applied appropriately, I'm still seeing the same code.

PR #1559, opened 2019-05-28
add alias go for goroutine and gos for goroutines

two:
goroutine is too long for use


derekparker:
I'm fine with adding an alias, but I think `gr` for goroutine and `grs` for goroutines is better. `go` is a verb and is already associated with the action of creating a goroutine, not the actual name for one.

two:
> I'm fine with adding an alias, but I think `gr` for goroutine and `grs` for goroutines is better. `go` is a verb and is already associated with the action of creating a goroutine, not the actual name for one.

What you said makes sense. I will now change `go` to `gr` , `gos` to `grs`, and it will be easier to use.

PR #1571, opened 2019-06-01
service: return an error when a client calls an unknown method

aarzilli:
```
service: return an error when a client calls an unknown method

Without this a client calling an method on a version of Delve that
doesn't have that method (for example because it's old) will never get
a response back.

```


rogpeppe:
Thanks for this! How about a test too?

derekparker:
I agree I think we can unit test this.

aarzilli:
Done

PR #1573, opened 2019-06-05
proc: add LocationCover method to BinaryInfo

aarzilli:
```
proc: add LocationCover method to BinaryInfo

Also fixes findCompileUnitForOffset which was broken in some edge cases
(when looking up an offset inside the last child of the compilation
unit) which don't happen in normal executables (we only look up types, and those
are always direct childs of compile units).

```

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -61,9 +61,11 @@ type BinaryInfo struct {
 	packageMap map[string]string
 
 	frameEntries frame.FrameDescriptionEntries
-	compileUnits []*compileUnit
-	types        map[string]dwarfRef
-	packageVars  []packageVar // packageVars is a list of all global/package variables in debug_info, sorted by address
+
+	compileUnits []*compileUnit // compileUnits is sorted by increasing DWARF offset
+
+	types       map[string]dwarfRef
+	packageVars []packageVar // packageVars is a list of all global/package variables in debug_info, sorted by address
 
 	gStructOffset uint64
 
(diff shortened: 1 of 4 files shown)


pkg/proc/bininfo.go
```diff
@@ -61,9 +61,12 @@ type BinaryInfo struct {
 	packageMap map[string]string
 
 	frameEntries frame.FrameDescriptionEntries
-	compileUnits []*compileUnit
-	types        map[string]dwarfRef
-	packageVars  []packageVar // packageVars is a list of all global/package variables in debug_info, sorted by address
+
+	compileUnits         []*compileUnit
+	compileUnitsByOffset []*compileUnit
```
derekparker: How often will this actually be used in practice? I'm considering the cost of increased default memory usage storing each pointer twice vs just sorting on demand.

aarzilli: AFAICT it's only used for a workaround that should never happen with never version of Go, so it should be used almost never. It occupies 8*num_packages so to get to ~1MB the project would have to use 100000 packages, which is a lot (delve has 160).

derekparker: Right, it seems like this is a convenience method, it's not used anywhere else except the test in this CL. I think for that reason it's better to pay the cost of a slower search vs (even minimal) increased memory usage.

aarzilli: Done.

derekparker: Is there any reason not to always sort compile units by offset anyways?

aarzilli: Apparently, no. We can't use the sort-by-pc we're doing, probably because they are not contiguous.

derekparker: What are your thoughts on that then, just sort that list by offset always, and then reimplement the sort.Search you had originally which was a bit cleaner.

aarzilli: Done

PR #1585, opened 2019-06-18
cmd/dlv: add --continue to continue process on launch/attach

### LINKED ISSUE #245: Allow attaching to a process without pausing the execution
Sorry, the issue was created too fast :D 
Proper description:
Given the following app:

``` go
package main

import (
    "fmt"
    "net/http"
)

func handler(w http.ResponseWriter, r *http.Request) {
    fmt.Fprintln(w, "hello")
}

func main() {

    handlers := http.NewServeMux()
    handlers.HandleFunc("/", handler)
    server := &http.Server{Addr: ":8889", Handler: handlers}

    server.ListenAndServe()
}
```

When I'm doing the following steps:

``` bash
go build -gcflags "-N -l" -o demo demo.go
./demo &
dlv attach `pgrep demo`
```

I can observe a pause in execution until I send the `continue` command to the debugger.

While I do understand that it might be useful to have that functionality, say for example I want to hook up in a staging app that has traffic on it (sample traffic), attaching to its process and having it paused is not ideal (for example I want to set tracepoints rather that breakpoints).

Do you think it would be useful to have an `attach --no-pause` flag?

briandealwis:
Add `--continue` option for `attach`, `debug`, `exec`, and `trace`, to issue a _continue_ on start.

Fixes #245


service/debugger/debugger.go
```diff
@@ -57,6 +57,9 @@ type Config struct {
 	// attach.
 	AttachPid int
 
+	// ContinueOnStart determiend whether the new process should be paused on start.
```
derekparker: nit: spelling error: s/determiend/determines/

derekparker:
cc @aarzilli for your thoughts as well.

briandealwis:
Sorry for wasting your time @derekparker — I hadn't tested this in an end-to-end situation.

service/debugger/debugger.go
```diff
@@ -99,6 +102,11 @@ func New(config *Config, processArgs []string) (*Debugger, error) {
 			return nil, attachErrorMessage(d.config.AttachPid, err)
 		}
 		d.target = p
+		if d.config.ContinueOnStart {
+			if _, err := d.target.ContinueOnce(); err != nil {
```
aarzilli: ContinueOnce shouldn't be called directly, it should be `proc.Continue`.

cmd/dlv/cmds/commands.go
```diff
@@ -240,6 +246,7 @@ to know what functions your process is executing.`,
 	traceCommand.Flags().BoolVarP(&traceTestBinary, "test", "t", false, "Trace a test binary.")
 	traceCommand.Flags().IntVarP(&traceStackDepth, "stack", "s", 0, "Show stack trace with given depth.")
 	traceCommand.Flags().String("output", "debug", "Output path for the binary.")
+	traceCommand.Flags().BoolVar(&ContinueOnStart, "continue", false, "Continue the traced process on start.")
```
aarzilli: The trace command always continues automatically, there's no need for this.

briandealwis:
Thanks — I'll fix this up and do more comprehensive testing.

derekparker:
> Sorry for wasting your time @derekparker — I hadn't tested this in an end-to-end situation. 

Not a waste of time, just doing my due diligence! I appreciate this PR as it's been a much requested one!

derekparker:
@briandealwis ping on this.

briandealwis:
Sorry, I was pulled away with holidays and other work.  My attempt to use a client like the traceCmd hasn't worked out: dlv just exits.  I'll be digging into this tomorrow.

briandealwis:
So it seemed to make sense to limiting `--continue` to `--headless` since otherwise `--init` can be used. I got it to work with by forcing `AcceptMulti` and then using a separate `Client` instance to connect to the server and then issue a `Disconnect(true)`. But the use of `AcceptMulti` means the delve server process stays around even after the actual process has exited.

I'm currently looking at whether I can do this at a lower level

briandealwis:
Fatfingered the close button.

Would this be better done at lower levels the stack?

Unfortunately I can't seem to debug delve with delve on macOS to get a better understanding of what's going on: I get some ioctl error from the second lldb debugserver.

derekparker:
> My attempt to use a client like the traceCmd hasn't worked out: dlv just exits. 

The trace command could be used as a starting point, I think. But I'm not sure what you mean by "dlv just exits". What does the process look like that you're testing against? Is it a server or something that is intended to be long running? Is there any output when delve exits? Is Delve exiting before the process does?

aarzilli:
> So it seemed to make sense to limiting `--continue` to `--headless` since otherwise `--init` can be used. I got it to work with by forcing `AcceptMulti` and then using a separate `Client` instance to connect to the server and then issue a `Disconnect(true)`. But the use of `AcceptMulti` means the delve server process stays around even after the actual process has exited.
> 
> I'm currently looking at whether I can do this at a lower level

No, I think this level is the correct one. AcceptMulti is necessary because otherwise the whole thing would be useless (you'd have a process running under delve but no way to connect to delve itself). 

> But the use of `AcceptMulti` means the delve server process stays around even after the actual process has exited.

This is always the case, if you run a program under delve and the program terminates you go back to delve's prompt.

briandealwis:
Sorry for lack of response: I was away at a lake and forbidden from touching a laptop :-)

I realize my comments above were from an odd behaviour I hit where `dlv exec --headless --accept-multiclient` of an app that exits normally during the session (like the example in [`_fixtures/buildtest/main.go`](https://github.com/go-delve/delve/blob/master/_fixtures/buildtest/main.go)) renders the headless `dlv` unkillable.  I filed #1617.

> This is always the case, if you run a program under delve and the program terminates you go back to delve's prompt.

I should have provided some context.  My driving use-case is to leverage Delve to allow debugging containerized Go programs in a Kubernete cluster as part of [Skaffold](https://skaffold.dev)(tracked as GoogleContainerTools/skaffold#2306).  `skaffold debug` rewrites the various Kubernetes resource objects to configure application containers for debugging as required for their language runtimes, and exposes the debug port so that IDEs can connect and debug programs in the cluster.  Since the users may not want to connect to each and every container (e.g., in a multi-service setup), we want the containers to be debuggable (able to be connected to) but not paused waiting for a debugger.  So for Go-based containers, that would launch the app with `dlv exec --continue --headless --listen :xxxx --accept-multiclient .../app` (for some port `xxxx`).

In the Kubernetes case, once the user has finished their debugging session and stops the application, then we want the container to exit so that Kubernetes can do its magic and relaunch the container.  As I noticed with #1617, if the application has exited normally then the headless instance becomes unkillable (at least with the dlv terminal client), and the container appears to hang.  (I suppose Kubernetes will eventually deem the container as being unhealthy and restart it?)

(I didn't test whether the RPC _detach_ will properly cause the headless dlv process to exit.)

derekparker:
> Sorry for lack of response: I was away at a lake and forbidden from touching a laptop :-)

No worries, it's good to unplug for a bit!

Saw you pushed some updates but the tests are failing in CI. Does everything pass for you locally?

briandealwis:
The tests are timing out on the `TestContinue` that I introduced earlier which tries to run
```
dlv debug --continue --headless --continue --accept-multiclient --listen :0
```
on the test in `_fixtures/buildtest/main.go.

But I just realized that test doesn't make sense in `--headless`.  Oops.

service/config.go
```diff
@@ -13,6 +13,8 @@ type Config struct {
 	Listener net.Listener
 	// ProcessArgs are the arguments to launch a new process.
 	ProcessArgs []string
+	// ContinueOnStart determines whether the new process should be continued on start.
+	ContinueOnStart bool
```
aarzilli: Am I missing something or this isn't used?

service/debugger/debugger.go
```diff
@@ -57,6 +57,9 @@ type Config struct {
 	// attach.
 	AttachPid int
 
+	// ContinueOnStart determines whether the new process should be paused on start.
+	ContinueOnStart bool
```
aarzilli: Same for this.

derekparker:
Thanks for the patch!

PR #1588, opened 2019-06-22
proc,service: remove support for locspec '<fnname>:0'

aarzilli:
```
proc,service: remove support for locspec '<fnname>:0'

The location specified '<fnname>:0' could be used to set a breakpoint
on the entry point of the function (as opposed to locspec '<fnname>'
which sets it after the prologue).
Setting a breakpoint on an entry point is almost never useful, the way
this feature was implemented could cause it to be used accidentally and
there are other ways to accomplish the same task (by setting a
breakpoint on the PC address directly).

```

### DIFF
--- a/Documentation/api/ClientHowto.md
+++ b/Documentation/api/ClientHowto.md
@@ -177,8 +177,7 @@ as valid.
 
 If you want to let your users specify a breakpoint on a function selected
 from a list of all functions you should specify the name of the function in
-the FunctionName field of Breakpoint and set Line to -1. *Do not omit Line,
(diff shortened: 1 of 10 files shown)


aarzilli:
That's true, this breaks backwards compatibility. However none of the major clients (vim-go, Goland, VSCode-go, the emacs plugins and LiteIDE) make use of this, I suspect if anybody that uses this or will use this in the future is/will be making a mistake. There's no reason I can think of to place a breakpoint at entry instead of after the prologue. IMHO the tradeoff is very favorable.

PR #1596, opened 2019-07-01
Add support for rev prefix and step-instruction

dpapastamos:


Documentation/cli/README.md
```diff
@@ -32,6 +32,7 @@ Command | Description
 [print](#print) | Evaluate an expression.
 [regs](#regs) | Print contents of CPU registers.
 [restart](#restart) | Restart process from a checkpoint or event.
+[reverse-stepi](#reverse-stepi) | Reverse single step a single cpu instruction.
```
aarzilli: We don't have a `stepi` command, it's called `step-instruction` so it's weird that this is `reverse-stepi`.

dpapastamos: Ah I see.  I can rename it to reverse-step-instruction.

aarzilli: Maybe we should make the prefix be `rev-` instead of reverse and have `rev-step-instruction`, `rev-next` and `rev-step`. Alternatively make `reverse`/`rev` be a prefix, like `on` and `deferred` and have `rev step-instruction`, `rev next` and `rev step`. cc @derekparker

dpapastamos: I like the approach to have a rev prefix and act as a modifier.  Once it is agreed on, I will rework my patch and update the PR.

derekparker: @aarzilli +1, I like the prefix approach a lot, `rev [command]` works for me.

Documentation/cli/README.md
```diff
@@ -32,6 +32,7 @@ Command | Description
 [print](#print) | Evaluate an expression.
 [regs](#regs) | Print contents of CPU registers.
 [restart](#restart) | Restart process from a checkpoint or event.
+[rev](#rev) | Performs the reverse operation of the command specified.
```
derekparker: nit: I would be more clear about this. Something more along the lines of `Reverses the execution of the target program for the command specified.`

derekparker:
My only concern here is as it's documented this new prefix does not outline the limitations. Namely, that the `rev` prefix currently only works for `step-instruction`. Either we need to call that out in the documentation, or implement the rest of the commands to work with the prefix before merging.

PR #1600, opened 2019-07-03
vendor: update golang.org/x/sys dependency to latest version

aarzilli:
```
vendor: update golang.org/x/sys dependency to latest version

```


aarzilli:
To be used for #1480

PR #1622, opened 2019-07-17
proc/native/linux: call postExit when process is killed by SIGINT

aarzilli:
```
proc/native/linux: call postExit when process is killed by SIGINT

If we return ErrProcessExited we should also set exited to true.

```


aarzilli:
Found while reviewing #1621

Issue #1625, opened 2019-07-19
Add print depth options for dereferense pointers and etc


msaf1980:
Can you add option for print depth of pointers, nested structs and etc ?
May be use existing MaxVariableRecurse or other options.

derekparker:
Not sure what you mean here, you simply want an option to print the number which represents the depth of nesting or pointer derefs?

msaf1980:
This is a option for evaluation depth of nested struct members, array items, dereference pointers, etc
when traversing into them.
By default its a 1 - so no traverse into. If I need a print value of nested element. I'll need print a casted memory addres. It's not a simple. With this parameter it's more easy.
I think, this will be useful for cli debugging.

aarzilli:
Implemented by 89d2167c31e57c83b716a0b772b1465df6b0b4c4

PR #1626, opened 2019-07-19
add max-variable-recurse parameter

msaf1980:
Add max-variable-recurse parameter for print depth of pointers, nested structs and etc 
See feature request from issue https://github.com/go-delve/delve/issues/1625

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -42,7 +42,7 @@ type Config struct {
 	// MaxArrayValues is the maximum number of array items that the commands
 	// print, locals, args and vars should read (in verbose mode).
 	MaxArrayValues *int `yaml:"max-array-values,omitempty"`
-	// MaxVariableRecurse is output evaluation depth of of nested struct members, array and
+	// MaxVariableRecurse is output evaluation depth of nested struct members, array and
 	// slice items and dereference pointers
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -42,6 +42,9 @@ type Config struct {
 	// MaxArrayValues is the maximum number of array items that the commands
 	// print, locals, args and vars should read (in verbose mode).
 	MaxArrayValues *int `yaml:"max-array-values,omitempty"`
+	// MaxVariableRecurse is output evaluation depth of of nested struct members, array and
```
derekparker: nit: two "of"'s.

pkg/config/config.go
```diff
@@ -216,6 +219,9 @@ substitute-path:
 # Maximum loaded string length.
 # max-string-len: 64
 
+# Output evaluation.
+# max-variable-recurse 1
```
derekparker: must have colon after setting, e.g.: `max-variable-recurse: 1`.

pkg/config/config.go
msaf1980: Not understood, where double s ?

derekparker: No I mean, it says "of of" instead of just "of".

PR #1629, opened 2019-07-22
terminal/command: add support for next [count]

jeremyfaller:
Add support for an optional [count] to the next command -- similar to gdb.

This fixes no outstanding issue, it just makes life a little easier for people familiar with gdb.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -138,7 +138,12 @@ the arguments.  With -noargs, the process starts with an empty commandline.
 		{aliases: []string{"continue", "c"}, cmdFn: c.cont, helpMsg: "Run until breakpoint or program termination."},
 		{aliases: []string{"step", "s"}, cmdFn: c.step, helpMsg: "Single step through program."},
 		{aliases: []string{"step-instruction", "si"}, allowedPrefixes: revPrefix, cmdFn: c.stepInstruction, helpMsg: "Single step a single cpu instruction."},
-		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: "Step over to next source line."},
+		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: `Step over to next source line.
+
+	 next [count]
+
+Optional [count] argument alloes you to skip multiple lines.
+`},
 		{aliases: []string{"stepout"}, cmdFn: c.stepout, helpMsg: "Step out of the current function."},
 		{aliases: []string{"call"}, cmdFn: c.call, helpMsg: `Resumes process, injecting a function call (EXPERIMENTAL!!!)
 	
@@ -864,6 +869,17 @@ func parseArgs(args string) ([]string, error) {
 	return v[0], nil
 }
 
+// parseOptionalCount parses an optional count argument.
+func parseOptionalCount(arg string, unspecifiedValue int64) (val int64, err error) {
+	if len(arg) == 0 {
+		return unspecifiedValue, nil
+	}
+	if val, err = strconv.ParseInt(arg, 0, 64); err != nil {
+		return unspecifiedValue, err
+	}
+	return val, err
+}
+
 func restart(t *Term, ctx callContext, args string) error {
 	v, err := parseArgs(args)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/terminal/command.go
```diff
@@ -864,6 +869,17 @@ func parseArgs(args string) ([]string, error) {
 	return v[0], nil
 }
 
+// parseOptionalCount parses an optional count argument.
+func parseOptionalCount(arg string, unspecifiedValue int64) (val int64, err error) {
```
aarzilli: unspecifiedValue is always 1, there's no reason to abstract it.

pkg/terminal/command.go
```diff
@@ -864,6 +869,17 @@ func parseArgs(args string) ([]string, error) {
...
+	if len(arg) == 0 {
+		return unspecifiedValue, nil
+	}
+	if val, err = strconv.ParseInt(arg, 0, 64); err != nil {
```
aarzilli: You can just `return strconv.ParseInt`, the value return parameter is ignored anyway if err is set.

pkg/terminal/command.go
```diff
@@ -1032,13 +1048,27 @@ func (c *Commands) next(t *Term, ctx callContext, args string) error {
...
-	if err != nil {
-		printcontextNoState(t)
+	var count int64
+	var err error
+	if count, err = parseOptionalCount(args, 1); err != nil {
 		return err
 	}
-	printcontext(t, state)
-	return continueUntilCompleteNext(t, state, "next")
+	if count != 0 {
```
aarzilli: I think it would be better to just return early `if count <= 0`

pkg/terminal/command.go
```diff
@@ -1032,13 +1048,27 @@ func (c *Commands) next(t *Term, ctx callContext, args string) error {
...
+		var state *api.DebuggerState
+		for ; count != 0; count-- {
+			state, err = exitedToError(t.client.Next())
+			if err != nil {
+				break
```
aarzilli: just call `printcontextNoState` and return here instead of breaking.

aarzilli:
I think that continueUntilCompleteNext needs to be called every iteration and it needs to be changed so that, if the user chooses to cancel the operation, the code in `next` will exit the `count` loop early.

pkg/terminal/command.go
jeremyfaller: OK!

pkg/terminal/command.go
jeremyfaller: OK!

pkg/terminal/command.go
jeremyfaller: you got it!

pkg/terminal/command.go
jeremyfaller: ok!

jeremyfaller:
LMK what you think.

jeremyfaller:
Is this failing test a flaky? It doesn't make sense to me why it's failing wrt to this change.

aarzilli:
>Is this failing test a flaky?

tbh this one looks more like a problem with debugserver...

pkg/terminal/command.go
```diff
@@ -138,7 +138,12 @@ the arguments.  With -noargs, the process starts with an empty commandline.
 		{aliases: []string{"continue", "c"}, cmdFn: c.cont, helpMsg: "Run until breakpoint or program termination."},
 		{aliases: []string{"step", "s"}, cmdFn: c.step, helpMsg: "Single step through program."},
 		{aliases: []string{"step-instruction", "si"}, allowedPrefixes: revPrefix, cmdFn: c.stepInstruction, helpMsg: "Single step a single cpu instruction."},
-		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: "Step over to next source line."},
+		{aliases: []string{"next", "n"}, cmdFn: c.next, helpMsg: `Step over to next source line.
+
+	 next [count]
+
+Optional [count] argument alloes you to skip multiple lines.
```
derekparker: s/alloes/allows/

pkg/terminal/command.go
```diff
@@ -864,6 +869,18 @@ func parseArgs(args string) ([]string, error) {
 	return v[0], nil
 }
 
+// parseOptionalCount parses an optional count argument.
+// If there are not arguments, a value of 1 is returned as the default.
+func parseOptionalCount(arg string) (val int64, err error) {
+	if len(arg) == 0 {
+		return 1, nil
+	}
+	if val, err = strconv.ParseInt(arg, 0, 64); err != nil {
```
derekparker: This was marked as resolved but I think @aarzilli comment still makes sense. We ignore the result of this if `err != nil` so just return the result directly.

pkg/terminal/command.go
```diff
@@ -1032,13 +1051,28 @@ func (c *Commands) next(t *Term, ctx callContext, args string) error {
...
-	printcontext(t, state)
-	return continueUntilCompleteNext(t, state, "next")
+	for ; count > 0; count-- {
+		state, err := exitedToError(t.client.Next())
+		if err != nil {
+			printcontextNoState(t)
+			return err
+		}
+		// If we're about the exit the loop, print the context.
+		if count == 1 {
```
derekparker: nit: let's just store the result of this calculation since we use it twice, something like `finishedNext := count == 1`.

pkg/terminal/command.go
jeremyfaller: I stink at splelling. ty.

pkg/terminal/command.go
jeremyfaller: sorry, got lost. I resolved this comment (and the one about the default value), but only actually made the change on one. Sorry, thanks!

pkg/terminal/command.go
jeremyfaller: Sure -- your code base, your style!

derekparker:
Thanks for the patch!

PR #1648, opened 2019-07-31
travis-ci: fix dwz test problems

aarzilli:
```
travis-ci: fix dwz test problems [WIP]

```


aarzilli:
Not sure why it started failing but it's dwz failing when running on the executable file output by go so it's either a bug in dwz or in go, definitely not delve.

PR #1654, opened 2019-08-09
pkg/proc: Reorganize and untangle

derekparker:
This is pt.2 in an ongoing effort to cleanup and untangle `pkg/proc`. I think there are some potential new package boundaries here and I'm trying to shake them out.

I think bininfo, eval, variables, and types still need to be cleaned up with some boundaries enforces. There is a lot of tight coupling between those concepts. Naturally there is some overlap within all these things, but I think we can separate these concerns.

For example, I don't think a `Function` should know what compile unit it came from. Instead we should be asking a compile unit what functions it contains. And technically, nothing aside from bininfo or even DWARF should know *anything* about the concept of a compile unit, or a DWARF entry, or how many object files have been loaded into the process memory. The interfaces between all of these concepts should be clean, and pass along strictly what needs to be known.

Again, this is all in an effort to make `pkg/proc` purely about process manipulation, and move all business logic higher up in `service/debugger`.

The changes here aren't final, just another step towards some cleanup.


derekparker:
Based on #1652

aarzilli:
I disagree with removing the cu field from Function, aside from the inefficiency of calling findCompileUnit every time, I don't think it makes the code more readable or more modular (and it makes it longer).

derekparker:
> I disagree with removing the cu field from Function, aside from the inefficiency of calling findCompileUnit every time, I don't think it makes the code more readable or more modular (and it makes it longer).

ACK, I agree this makes things temporarily less efficient, but I think having these circular references in the code is a good thing. A compile unit knows about the functions within it, and the function knows about the compile unit that encapsulates it. The reason we hold onto the compile unit is to look up the DWARF reader. That just seems backwards, we should have things organized properly so that finding the DWARF section that includes the particular function is efficient. My argument here is essentially that a function should have zero knowledge about what a DWARF reader even is, because we then begin to abuse the `Function` struct by passing it around everything it doesn't need to be in order to access the dwarf reader which is associated with the compile unit. Essentially my argument here is that we're breaking the Law Of Demeter (I know this isn't OO, but bear with me).

aarzilli:
Function is essentially an abstraction over a DWARF entry: it's a node of a tree structure where the first level is bininfo, the second level is the images, the third level is the compile units and the fourth one is functions. The cu pointer in Function and the image pointer in compileUnit are pointers to the parent node. 

I've heard people argue against pointers to parents; I don't think it's usually a good thing, in general, for various reasons. In this particular case you can't migrate a Function between two compileUnits so removing the parent pointer doesn't reduce the coupling between compileUnits and Functions it just makes it hidden and slower.

aarzilli:
We have another notable place with pointers to parents in this codebase and that's the Process pointer inside each thread (both native threads and gdbserial threads have it). You could get rid of that by having each function that takes a thread also take the process (because you can't derive the process of a thread from just the thread itself).

When I started writing gdbserial I did it that way, then gave up on it halfway through. In this case I'm against it because it produces function signatures that give the illusion that you can pass an arbitrary process and thread but it's actually a lie the process you pass must be the parent process of the thread.

derekparker:
@aarzilli thanks for your responses. I really appreciate this discussion a lot and enjoy making sure we are maintaining the cleanest code base we possibly can.

> Function is essentially an abstraction over a DWARF entry: it's a node of a tree structure where the first level is bininfo, the second level is the images, the third level is the compile units and the fourth one is functions. The cu pointer in Function and the image pointer in compileUnit are pointers to the parent node.

I agree with you that this is the overall structure of the data, but I ask you to consider if we maintain these parent pointers *only* for optimization, or because they actually make sense in the code. I think parent pointers can be good in a proper tree or graph data structure where we actually need to traverse from node to node. In this case however, we don't necessarily need to traverse the data structure. We have a distinct idea (a function), and we only hold pointers for convenience, we aren't actually traversing a structure and analyzing it.

Consider two other cases. 

1. The stdlib dwarf package, all leaf references (`FuncType`, `ArrayType`, etc) do not have parent references. Typically if you find the leaf you are looking for, you really don't care about anything above it.

2. Look at this diff from this patch. Here is an example of where I think we are abusing this relationship. We are trying to return a new scope for the given `*Image` (which I think is a strained relationship anyways). However, in order to do this we are forced to create 2 additional levels of indirection. We have to create a fake, empty `*Function` which then contains a fake, empty `*compileUnit`.

```diff
- // globalFor returns a global scope for 'image' with the register values of 'scope'.
- func (scope *EvalScope) globalFor(image *Image) *EvalScope {
- 	r := *scope
- 	r.Regs.StaticBase = image.StaticBase
- 	r.Fn = &Function{cu: &compileUnit{image: image}}
- 	return &r
- }
```

I argue that creating those additional levels of indirection makes the codebase harder to understand. Now, when you look at `EvalScope` (which I also am trying to refactor and slim down), you have to load into your head knowledge of function types (not so bad), but also compileUnits and images (object files). Now, I haven't fully thought through how EvalScope should be refactored, but I think there are things to be done there.

> We have another notable place with pointers to parents in this codebase and that's the Process pointer inside each thread (both native threads and gdbserial threads have it). You could get rid of that by having each function that takes a thread also take the process (because you can't derive the process of a thread from just the thread itself).

I agree with this, and I think we should also look at this relationship critically and see if there are ways to simplify, but I think we leave it for now.

aarzilli:
> I agree with you that this is the overall structure of the data, but I ask you to consider if we maintain these parent pointers only for optimization, or because they actually make sense in the code.

AFAIC it's both, it's a permanent relationship between to entities which we represent with a pointer, it's clearer, it's shorter and it's also faster.

> The stdlib dwarf package, all leaf references (FuncType, ArrayType, etc) do not have parent references. Typically if you find the leaf you are looking for, you really don't care about anything above it.

This is unfortunate, if they did we wouldn't need findCompileUnitForOffset. 

It doesn't matter much in practice because the linker doesn't know which compile unit each type belongs to. If it did maybe it would be more useful.

For example, findMethod, works entirely through string manipulation. If the compiler ever evolved into being able to place methods inside their receiver type (like C++ compilers do) we'd have trouble taking advantage of that.

> Look at this diff from this patch.

Yes, globalFor sucks. I think it's a problem with globalFor though, not with Function. If you look at how globalFor is used it's always used to call either (*EvalScope).newVariable or (*EvalScope).extractVarInfoFromEntry.

The underlying mistake is probably that those two functions shouldn't be methods of EvalScope (there's already a version of newVariable that isn't and it isn't called as a form of conveninence).

derekparker:
@aarzilli you've swayed me on the Function/cu issue for now. I think cleaning up the `globalFor` method will be good enough. I'll revert that patch and take a shot at cleaning up `globalFor`.

derekparker:
@aarzilli pushed update which reverts the Function/cu change. I also removed a couple methods from EvalScope (for function calls). It might make sense to do that with more of the EvalScope methods.

Would you mind taking a shot at cleaning up `globalFor`? Seems like you might have some ideas around that.

PR #1658, opened 2019-08-10
proc: remove (*EvalScope).globalFor

aarzilli:
```
proc: remove (*EvalScope).globalFor

```

### DIFF
--- a/pkg/proc/dwarf_expr_test.go
+++ b/pkg/proc/dwarf_expr_test.go
@@ -212,12 +212,14 @@ func TestDwarfExprLoclist(t *testing.T) {
 	mainfn := bi.LookupFunc["main.main"]
 
 	mem := newFakeMemory(defaultCFA, uint16(before), uint16(after))
-	regs := linutil.AMD64Registers{Regs: &linutil.AMD64PtraceRegs{}}
+	const PC = 0x40100
+	regs := linutil.AMD64Registers{Regs: &linutil.AMD64PtraceRegs{Rip: PC}}
 
-	scope := &proc.EvalScope{Location: proc.Location{PC: 0x40100, Fn: mainfn}, Regs: dwarfRegisters(bi, &regs), Mem: mem, BinInfo: bi}
+	scope := &proc.EvalScope{Location: proc.Location{PC: PC, Fn: mainfn}, Regs: dwarfRegisters(bi, &regs), Mem: mem, BinInfo: bi}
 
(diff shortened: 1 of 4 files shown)


pkg/proc/dwarf_expr_test.go
```diff
@@ -212,12 +212,14 @@ func TestDwarfExprLoclist(t *testing.T) {
 	mainfn := bi.LookupFunc["main.main"]
 
 	mem := newFakeMemory(defaultCFA, uint16(before), uint16(after))
-	regs := linutil.AMD64Registers{Regs: &linutil.AMD64PtraceRegs{}}
+	const PC = 0x40100
+	regs := linutil.AMD64Registers{Regs: &linutil.AMD64PtraceRegs{Rip: 0x40100}}
```
derekparker: You can use the new const here.

pkg/proc/eval.go
```diff
@@ -251,7 +251,7 @@ func (scope *EvalScope) setValue(dstv, srcv *Variable, srcExpr string) error {
 
 // Extracts the name and type of a variable from a dwarf entry
 // then executes the instructions given in the  DW_AT_location attribute to grab the variable's address
-func (scope *EvalScope) extractVarInfoFromEntry(varEntry *dwarf.Entry) (*Variable, error) {
+func extractVarInfoFromEntry(bi *BinaryInfo, image *Image, regs op.DwarfRegisters, mem MemoryReadWriter, varEntry *dwarf.Entry) (*Variable, error) {
```
derekparker: This can be moved out of `eval.go` now, most likely, as it doesn't have anything to do with `EvalScope` anymore.

pkg/proc/eval.go
```diff
@@ -288,6 +287,10 @@ func (scope *EvalScope) extractVarInfoFromEntry(varEntry *dwarf.Entry) (*Variabl
 	return v, nil
 }
 
+func (scope *EvalScope) extractVarInfoFromEntry(varEntry *dwarf.Entry) (*Variable, error) {
```
derekparker: Do we still need this at all?

pkg/proc/eval.go
aarzilli: moved to variables.go

pkg/proc/eval.go
aarzilli: I thought it was used in multiple places but it's actually only called in one place so I removed it now.

PR #1665, opened 2019-08-13
proc/gdbserial: reset thread updater in step loop

### LINKED ISSUE #1659: 'Add after Finish' panic during step command

```
Delve Debugger
Version: 1.2.0
Build: c9c455cc381c500f69fcfae1ebeff9bb1c5898b4
``` 

```
go version go1.12.6 darwin/amd64
```

Mac OS, amd64

4. What did you do?

Debug the following program with a breakpoint the first line of main in a terminal with tty enabled:

```
package main

type T struct {
	s string
}

func main() {
	t := T{} // break
	t.s = "1"
	println(t.s)
	t.s = "2"
	println(t.s)
}
```

Step once debugger reaches the breakpoint

Debugger stops on the next line.

6. What did you see instead?

Debugger panics with the "Add after Finish" error. Log can be found here https://gist.github.com/nd/a6587640ed34cc67f7701a3e5a3aaf98

aarzilli:
```
proc/gdbserial: reset thread updater in step loop

threadUpdater needs to be reset before every possible use.

Fixes #1659

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver_conn.go
+++ b/pkg/proc/gdbserial/gdbserver_conn.go
@@ -599,6 +599,9 @@ func (conn *gdbConn) step(threadID string, tu *threadUpdater, ignoreFaultSignal 
 		if err := conn.send(conn.outbuf.Bytes()); err != nil {
 			return err
 		}
+		if tu != nil {
+			tu.Reset()
+		}
 		var err error
 		_, sig, err = conn.waitForvContStop("singlestep", threadID, tu)
 		if err != nil {


aarzilli:
I don't think it makes sense to write a test for this. The bug report has some specific circumstances but we already have multiple test that do essentially the same thing and they didn't catch it, it must be triggered by something in the OS/environment which doesn't happen in CI.

aarzilli:
cc @nd how does this look?

nd:
Works without panic, thanks!

PR #1677, opened 2019-08-22
pkg/terminal: add support for gdb-style "[un]display" command

jeremyfaller:
This is kind of a clunky way to enable gdb-style [un]display, but incredibly helpful. LMK what you think.


jeremyfaller:
Totally right about next/step. Sorry 'bout that. Will withdraw PR until I get unit tests fixed, and get some more time to think about things.

PR #1679, opened 2019-08-27
pkg/dwarf/frame/parser: Fix `parse augmentation`

chainhelen:
According to the description of "CIE: length, CIE_id, version, augmentation"
in Page 122 of  http://dwarfstd.org/doc/Dwarf3.pdf ,
`augmentation` should exclude `version`


chainhelen:
@derekparker Has added one test for `parseCIE` and rebase commit, please check again.Thx.

PR #1681, opened 2019-08-30
dwarf/line: fix state machine behavior with multi-sequence units

aarzilli:
```
dwarf/line: fix state machine behavior with multi-sequence units

A compile unit can produce a debug_line program consisting of multiple
sequences according to the DWARF standard. The standard guarantees that
addresses monotonically increment within a single sequence but
different sequences may not follow this rule.

This commit changes dwarf/line (in particular PCToLine and
AllPCsBetween) to support debug_line sections containing units with
multiple sequences.

TestPCToLine needs to be changed so that it picks valid addresses (i.e.
addresses covered by a sequence) as values for basePC, instead of just
rounding.

```


aarzilli:
Reviewable.

PR #1682, opened 2019-08-30
proc: round TLS segment size to its alignment

heschik:
The fix for #1428 was buggy, partly because I communicated poorly. Sorry
about that.

The size of the TLS segment should be rounded to its alignment,
not to the machine word size. (For the record, the rounding logic is
working in bits, but PtrSize is in bytes, so it wasn't working as
originally intended either.)


heschik:
Updated the commit to reflect new information from the lld people, please see its description for details.

PR #1690, opened 2019-09-10
delve : adding support for AARCH64 platform

ossdev07:
go debugger (delve) was developed for AMD architecture only and
we added the support for AARCH64 architeture along with AMD

-off-by: ossdev <ossdev@puresoftware.com>


pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,247 @@
...
+
+			inst.Args[i] = arm64asm.ArchArm64(uint64(pc) + uint64(rel) + uint64(inst.Len))
+
+		}
+	}
+}
+
+// Text will return the assembly instructions in human readable format according to
+// the flavour specified.
+var readerAt io.ReaderAt
```
aarzilli: Remove this, pass nil to arm64asm.GoSyntax instead. If I'm not mistaken this variable is always nil nothing reads or writes it and only exists because arm64asm.GoSyntax has it as an argument (which it doesn't use). It makes Text look non-reentrant.

pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,247 @@
...
+	var text string
+
+	switch flavour {
+	case GNUFlavour:
+		text = arm64asm.GNUSyntax(arm64asm.Inst(*inst.Inst))
+	case GoFlavour:
+		text = arm64asm.GoSyntax(arm64asm.Inst(*inst.Inst), inst.Loc.PC, bi.symLookup, readerAt)
+	case IntelFlavour:
+		fallthrough
+	default:
```
aarzilli: Move GNUFlavour clause here, and make it `fallthrough` to `default` (removing the duplicate code)

pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,247 @@
...
+func (inst *AsmInstruction) IsRet() bool {
+	if inst.Inst == nil {
+		return false
+	}
+	return inst.Inst.Op == arm64asm.RET || inst.Inst.Op == arm64asm.ERET
+}
+
+func resolveCallArg(inst *archInst, currentGoroutine bool, regs Registers, mem MemoryReadWriter, bininfo *BinaryInfo) *Location {
+	if inst.Op != arm64asm.BL /*&& inst.Op != arm64asm.B*/ {
+	//if inst.Op != arm64asm.ADRP && inst.Op != arm64asm.LDUR {
```
aarzilli: Remove commented out code.

pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,247 @@
...
+		if !currentGoroutine || regs == nil {
+			return nil
+		}
+		pc, err = regs.Get(int(arg))
+		if err != nil {
+			return nil
+		}
+
+	case arm64asm.RegSP:
+		fmt.Println("1")
```
aarzilli: This function shouldn't print.

pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,247 @@
...
+		return &Location{PC: pc}
+	}
+	return &Location{PC: pc, File: file, Line: line, Fn: fn}
+
+}
+
+type instrseq []arm64asm.Op
+
+// Possible stacksplit prologues are inserted by stacksplit in
+// $GOROOT/src/cmd/internal/obj/x86/obj6.go.
```
aarzilli: This comment is wrong, clearly you didn't get this prologues from the x86 file. Also I did check the correct linker file and the stack split code looks different from what you checking here. If you don't feel like figuring out what the prologues actually look like remove all this code and reject programs built before 1.12 on arm64.

pkg/proc/fncall.go
```diff
@@ -12,13 +12,14 @@ import (
 	"sort"
```
aarzilli: This file shouldn't be changed at all, the Go runtime has no support for injecting calls in non-amd64 architectures. Instead in proc_linux.go the functions Attach and Launch should be changed to pass false to NewCommonProcess when the architecture is arm64.

pkg/proc/linutil/regs.go
```diff
@@ -397,3 +575,38 @@ func AMD64XstateRead(xstateargs []byte, readLegacy bool, regset *AMD64Xstate) er
 
 	return nil
 }
+// // LinuxX86XstateRead reads a byte array containing an XSAVE area into regset.
```
aarzilli: This comment is wrong, it should describe ARM64XstateRead.

pkg/proc/linutil/regs.go
```diff
@@ -397,3 +575,38 @@ func AMD64XstateRead(xstateargs []byte, readLegacy bool, regset *AMD64Xstate) er
...
+		return nil
+	}
+
+	if xstate_bv&(1<<2) == 0 {
+		// AVX state not present
+		return nil
+	}
+
+	avxstate := xstateargs[_XSAVE_EXTENDED_REGION_START:]
+	regset.AvxState = true
```
aarzilli: AVX is a subset of the Intel 86x64 ISA how can this be correct for ARM64?

pkg/proc/native/proc_linux.go
```diff
@@ -178,7 +178,7 @@ func (dbp *Process) addThread(tid int, attach bool) (*Thread, error) {
 	}
 
 	var err error
-	if attach {
+	if attach {
```
aarzilli: Run go fmt.

pkg/proc/native/ptrace_linux.go
```diff
@@ -82,3 +87,15 @@ func PtraceGetRegset(tid int) (regset linutil.AMD64Xstate, err error) {
 	err = linutil.AMD64XstateRead(regset.Xsave, false, &regset)
 	return
 }
+// PtraceGetRegset returns floating point registers of the specified thread
+// using PTRACE.
+// See arm64_linux_fetch_inferior_registers in gdb/arm64-linux-nat.c.html
+// and arm64_supply_xsave in gdb/arm64-tdep.c.html
+func PtraceGetRegset_arm64(tid int) (regset linutil.ARM64PtraceFpRegs, err error) {
```
aarzilli: Create two files ptrace_linux_amd64.go and ptrace_linux_arm64.go. Copy the existing PtraceGetRegset into ptrace_linux_amd64.go and copy this code to ptrace_linux_arm64.go, remove the _arm64 suffix from PtraceGetRegset_arm64.

pkg/proc/native/threads_linux_arm64.go
```diff
@@ -0,0 +1,131 @@
...
+	pc := regs.PC()
+	fn := t.BinInfo().PCToFunc(pc)
+	if fn != nil && ((fn.Name == "runtime.futex") || (fn.Name == "runtime.usleep") || (fn.Name == "runtime.clone")) {
+		return true
+	}
+	return false
+}
+
+func (t *Thread) restoreRegisters(savedRegs proc.Registers) error {		
+		sr := savedRegs.(*linutil.ARM64Registers)
```
aarzilli: Run go fmt (this goes for every other file too).

pkg/proc/threads.go
```diff
@@ -1,585 +1,585 @@
...
-}
-
-func (w *onNextGoroutineWalker) Visit(n ast.Node) ast.Visitor {
-	if binx, isbin := n.(*ast.BinaryExpr); isbin && binx.Op == token.EQL && exprToString(binx.X) == "runtime.curg.goid" {
-		w.ret, w.err = evalBreakpointCondition(w.thread, n.(ast.Expr))
-		return nil
-	}
-	return w
-}
+package proc
```
aarzilli: You changed the line endings on this file from `\n` (unix) to `\r\n` (windows) this is a big problem because:

1. it makes the diff unreadable going forward
2. I can't review the changes because github marks the entire file as changed.

This needs to be fixed.

aarzilli:
PS. Thank you for doing this.

vielmetti:
If you are looking for a CI path for this, Drone Cloud has a native arm64 build and test environment.

aarzilli:
@vielmetti thank you!

derekparker:
@vielmetti thanks for the recommendation!

@aarzilli thanks for jumping in with the quick review on this!

@ossdev07 thanks for this patch!

vendor/golang.org/x/arch/arm64/arm64asm/arg.go
```diff
@@ -0,0 +1,494 @@
+// Generated by ARM internal tool
```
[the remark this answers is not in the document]
ossdev07: @aarzilli can you please provide some document how "go mod vendor" works , In our case it is removing a lot of code from vendor directory which is required for AARCH64 arch

ossdev07:
@aarzilli Please share your suggestion on following issue
https://github.com/go-delve/delve/issues/1715

aarzilli:
Unfortunately I don't own a 64bit ARM system so I have no clue what could be wrong.

erifan:
> @aarzilli Please share your suggestion on following issue
> #1715

Do you want to get the current goroutine's tls ? For Go, tls is just used to store g register of program with cgo, so in fact you want to get g register ? If so, For pure go program, g is r28, for cgo, get tls with instruction “MRS TPIDR_EL0, R0”. Maybe there's better ways to get tls, I'm not sure. For general purpose register, there is something wrong with the current unit.PtraceRegsArm64 API, we'll fix it.

ossdev07:
> > @aarzilli Please share your suggestion on following issue
> > #1715
> 
> Do you want to get the current goroutine's tls ? For Go, tls is just used to store g register of program with cgo, so in fact you want to get g register ? If so, For pure go program, g is r28, for cgo, get tls with instruction “MRS TPIDR_EL0, R0”. Maybe there's better ways to get tls, I'm not sure. For general purpose register, there is something wrong with the current unit.PtraceRegsArm64 API, we'll fix it.

Thank you so much for this wonderful information , now i am able to get g register

hengwu0:
We have finished the disasm of arm64, which include get g register, resolveCallArg. You can review my patch at https://github.com/go-delve/delve/pull/1733/commits/36912a4a17198a27c63a1545998b6ff0b8157b6d, and share your apinion, thanks a lot.

derekparker:
Work on this issue has moved to https://github.com/go-delve/delve/pull/1733.

PR #1691, opened 2019-09-12
pkg/proc: fix can not call method of an embedded filed directly

### LINKED ISSUE #1688: Can not call method of an embedded field directly

```
Delve Debugger
Version: 1.3.0
Build: 69e6b613d81ae115521571c7550b0926be031d52
```

```
go version go1.12.9 linux/amd64
```

linux, amd64

4. What did you do?

Debug the following program with breakpoint on line 13

```
package main

type Base struct{}
type Derived struct {
	Base
}

func (b *Base) Method() int { return 4 }

var d Derived

func main() {
	println(d.Method()) // break
}
```

Execute the `call d.Method()` command once debugger stops at the breakpoint.

Result of `d.Method()`

6. What did you see instead?

Error: `Command failed: d has no member Method`. The `call d.Base.Method()` command works fine.

chainhelen:
`func (v *Variable) findMethod` should support for searching methods of an
embedded filed.

Fixes #1688

### DIFF
--- a/_fixtures/fncall.go
+++ b/_fixtures/fncall.go
@@ -121,6 +121,19 @@ func noreturncall(n int) {
 	return
 }
 
+type Base struct{
+	y int
+}
+
+type Derived struct {
+	x int
+	Base
+}
+
+func (b *Base) Method() int {
+	return b.y
+}
+
 func main() {
 	one, two := 1, 2
 	intslice := []int{1, 2, 3}
@@ -142,9 +155,13 @@ func main() {
 	fn2ptrmeth := pa.PRcvr
 	var fn2nil func()
 
+	d := &Derived{3, Base{4}}
+
 	runtime.Breakpoint()
 	call1(one, two)
 	fn2clos(2)
 	strings.LastIndexByte(stringslice[1], 'w')
-	fmt.Println(one, two, zero, callpanic, callstacktrace, stringsJoin, intslice, stringslice, comma, a.VRcvr, a.PRcvr, pa, vable_a, vable_pa, pable_pa, fn2clos, fn2glob, fn2valmeth, fn2ptrmeth, fn2nil, ga, escapeArg, a2, square, intcallpanic, onetwothree, curriedAdd, getAStruct, getAStructPtr, getVRcvrableFromAStruct, getPRcvrableFromAStructPtr, getVRcvrableFromAStructPtr, pa2, noreturncall, str)
+	d.Method()
+	d.Base.Method()
+	fmt.Println(one, two, zero, callpanic, callstacktrace, stringsJoin, intslice, stringslice, comma, a.VRcvr, a.PRcvr, pa, vable_a, vable_pa, pable_pa, fn2clos, fn2glob, fn2valmeth, fn2ptrmeth, fn2nil, ga, escapeArg, a2, square, intcallpanic, onetwothree, curriedAdd, getAStruct, getAStructPtr, getVRcvrableFromAStruct, getPRcvrableFromAStructPtr, getVRcvrableFromAStructPtr, pa2, noreturncall, str, d)
 }
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -1847,7 +1847,32 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
 		}
 		return r, nil
 	}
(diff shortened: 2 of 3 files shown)


_fixtures/fncall.go
```diff
@@ -121,6 +121,14 @@ func noreturncall(n int) {
 	return
 }
 
+type Base struct{}
```
aarzilli: Change this to be:

```
type Base struct {
    y int
}
```

_fixtures/fncall.go
```diff
@@ -121,6 +121,14 @@ func noreturncall(n int) {
...
+
+type Derived struct {
```
aarzilli: Change this to be:

```
type Derived struct {
    x int
    Base
}
```

_fixtures/fncall.go
```diff
@@ -121,6 +121,14 @@ func noreturncall(n int) {
...
+	Base
+}
+
+func (b *Base) Method() int { return 4 }
```
aarzilli: Change this to be 

```
func (b *Base) Method() int { return b.y }
```

_fixtures/fncall.go
```diff
@@ -142,8 +150,12 @@ func main() {
 	fn2ptrmeth := pa.PRcvr
 	var fn2nil func()
 
+	d := &Derived{}
```
aarzilli: Change this to 

```
d := &Derived{ 3, Base{ 4 } }
```

I want to check that we're getting the right address for the receiver.

pkg/proc/eval.go
```diff
@@ -1828,7 +1828,39 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
...
+func (v *Variable) tryFindMethodInEmbeddedFileds(mname string) (*Variable, error) {
+	structVar := v.maybeDereference()
+	structVar.Name = v.Name
+	if structVar.Unreadable != nil {
+		return structVar, nil
+	}
+	switch t := structVar.RealType.(type) {
+	case *godwarf.StructType:
+		for _, field := range t.Field {
+			isEmbeddedStructMember :=
```
aarzilli: This code was necessary to support older versions of Go, just use `field.Embedded` directly, i.e. remove `isEmbeddedStructMember` and use field.Embedded in its place.

service/test/integration2_test.go
```diff
@@ -1581,6 +1581,55 @@ func TestClientServerFunctionCall(t *testing.T) {
 	})
 }
 
+func TestClientServerFunctionCallEmbeddedFiled(t *testing.T) {
```
aarzilli: Instead of writing this you should add a test case to TestCallFunction in service/test/variables_test.go

It should look like this:

```
{`d.Base.Method()`, []string{ `:int:4` }, nil },
{`d.Method()`, []string{ `:int:4` }, nil },
```

pkg/proc/eval.go
```diff
@@ -1790,7 +1790,7 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
 	}
 
 	if _, istypedef := typ.(*godwarf.TypedefType); !istypedef {
-		return nil, nil
+		return v.tryFindMethodInEmbeddedFileds(mname)
```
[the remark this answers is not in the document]
aarzilli: This is because 1.10 doesn't support function calls and will accept everything and versions of Go prior to 1.12 have a linker bug that's too complicated to explain here (if you are curious look at commit a0f5d5f8830e578892c47f7704e6a2616273aac1 in the go repository). I think you should keep `return nil, nil` here and only test this feature on 1.12 and later (which, if you do the change I suggest below simply means adding the test cases to the testcases112 variable).

chainhelen:
@aarzilli Thanks for your guidances, it's interesting for me. I have done and rebase, please view again.

pkg/proc/eval.go
```diff
@@ -1820,7 +1820,32 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
 		}
 		return r, nil
 	}
+	return v.tryFindMethodInEmbeddedFileds(mname)
+}
 
+func (v *Variable) tryFindMethodInEmbeddedFileds(mname string) (*Variable, error) {
```
derekparker: s/Fileds/Fields/

chainhelen:
@derekparker  Thx. I have done and rebase again.

PR #1699, opened 2019-09-24
Fix starlark map iteration for maps > 64 entries

alxn:
I discovered that `dlv` wasn't correctly iterating over maps larger than 64 elements.

Ref: https://github.com/go-delve/delve/blob/master/Documentation/cli/expr.md#elements-limit
> For arrays, slices, strings and maps delve will only return a maximum of 64 elements at a time:

```go
package main

import (
	"runtime"
	"strconv"
)

func main() {
	m := make(map[string]int)
	for i := 0; i < 256; i++ {
		m[strconv.Itoa(i)] = i
	}
	runtime.Breakpoint()
}
```
```python
i = 0
m = eval(None, "m")
print(m.Variable.Len)
for item in m.Variable.Value:
	i = i + 1
print(i)
```

```
(dlv) source iterate.star
256
128
```

This was because in `mapVariableAsStarlarkValueIterator.Next()`, we are iterating in 2s, but we are not resuming `/2` for the next offset.


alxn:
`    variables_test.go:896: Wrong number of children: 64
`

PR #1702, opened 2019-10-01
*: add option to re-record recorded targets

aarzilli:
```
*: add option to re-record recorded targets

Adds a '-r' option to the 'restart' command (and to the Restart API)
that re-records the target when using rr.

Also moves the code to delete the trace directory inside the gdbserial
package.

```


service/debugger/debugger.go
```diff
@@ -39,6 +39,8 @@ type Debugger struct {
 	target       proc.Process
 	log          *logrus.Entry
 
+	canRestart bool // target was started with Launch and can be restarted
```
derekparker: Can't we infer this from `config` by asserting that we weren't given a PID to attach to?

PR #1717, opened 2019-10-13
proc,debugger: implement logical breakpoints

aarzilli:
```
proc,debugger: implement logical breakpoints

Modifies FindFileLocation, FindFunctionLocation and LineToPC as well as
service/debugger to support inlining and introduces the concept of
logical breakpoints.

For inlined functions FindFileLocation, FindFunctionLocation and
LineToPC will now return one PC address for each inlining and one PC
for the concrete implementation of the function (if present).

A proc.Breakpoint will continue to represent a physical breakpoint, at
a single memory location.

Breakpoints returned by service/debugger, however, will represent
logical breakpoints and may be associated with multiple memory
locations and, therefore, multiple proc.Breakpoints.

The necessary logic is introduced in service/debugger so that a change
to a logical breakpoint will be mirrored to all its physical
breakpoints and physical breakpoints are aggregated into a single
logical breakpoint when returned.

```

### DIFF
--- a/pkg/dwarf/line/state_machine.go
+++ b/pkg/dwarf/line/state_machine.go
@@ -112,33 +112,6 @@ func newStateMachine(dbl *DebugLineInfo, instructions []byte) *StateMachine {
 	return sm
 }
 
-// Returns all PCs for a given file/line. Useful for loops where the 'for' line
-// could be split amongst 2 PCs.
-func (lineInfo *DebugLineInfo) AllPCsForFileLine(f string, l int) (pcs []uint64) {
-	if lineInfo == nil {
-		return nil
-	}
-
-	var (
-		lastAddr uint64
-		sm       = newStateMachine(lineInfo, lineInfo.Instructions)
-	)
-
-	for {
-		if err := sm.next(); err != nil {
-			if lineInfo.Logf != nil {
-				lineInfo.Logf("AllPCsForFileLine error: %v", err)
-			}
-			break
-		}
-		if sm.line == l && sm.file == f && sm.address != lastAddr && sm.isStmt && sm.valid {
-			pcs = append(pcs, sm.address)
-			lastAddr = sm.address
-		}
-	}
-	return
-}
-
 // AllPCsForFileLines Adds all PCs for a given file and set (domain of map) of lines
 // to the map value corresponding to each line.
 func (lineInfo *DebugLineInfo) AllPCsForFileLines(f string, m map[int][]uint64) {
@@ -237,6 +210,13 @@ func (lineInfo *DebugLineInfo) PCToLine(basePC, pc uint64) (string, int) {
 		panic(fmt.Errorf("basePC after pc %#x %#x", basePC, pc))
 	}
 
+	sm := lineInfo.stateMachineFor(basePC, pc)
+
+	file, line, _ := sm.PCToLine(pc)
+	return file, line
+}
+
+func (lineInfo *DebugLineInfo) stateMachineFor(basePC, pc uint64) *StateMachine {
 	var sm *StateMachine
 	if basePC == 0 {
 		sm = newStateMachine(lineInfo, lineInfo.Instructions)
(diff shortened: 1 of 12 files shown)


derekparker:
I'll review tomorrow, please fix conflicts in the meantime.

aarzilli:
Done.

pkg/proc/bininfo.go
```diff
@@ -390,38 +398,61 @@ func (err *ErrCouldNotFindLine) Error() string {
 }
 
 // LineToPC converts a file:line into a memory address.
```
derekparker: Comment needs to be updated as this now returns a slice of addresses.

pkg/dwarf/line/state_machine.go
```diff
@@ -320,6 +298,40 @@ func (lineInfo *DebugLineInfo) LineToPC(filename string, lineno int) uint64 {
 	return fallbackPC
 }
 
+// LineToPCIn returns the first PC for filename:lineno in the interval [startPC, endPC).
+// basePC will be used for caching, it's normally the entry point for the
+// function containing pc.
```
derekparker: Let's add just a little bit more info here, specifically mention something about how we can't rely simply on the file:line and instead must provide an address range because of inlined functions.

pkg/proc/bininfo.go
```diff
@@ -390,38 +398,61 @@ func (err *ErrCouldNotFindLine) Error() string {
...
-	return r
+	pcs = make([]uint64, 0, len(fn.InlinedCalls)+1)
+	pcs = bi.appendLineToPCIn(pcs, filename, lineno, fn.cu, fn, fn.Entry, fn.End)
+	for _, call := range fn.InlinedCalls {
+		pcs = bi.appendLineToPCIn(pcs, filename, lineno, call.cu, bi.PCToFunc(call.LowPC), call.LowPC, call.HighPC)
+	}
+	return pcs, nil
+}
+
+func (bi *BinaryInfo) appendLineToPCIn(pcs []uint64, filename string, lineno int, cu *compileUnit, containingFn *Function, lowPC, highPC uint64) []uint64 {
```
derekparker: This can be a standalone function, it does not use `*BinaryInfo`.

service/debugger/debugger.go
```diff
@@ -442,45 +443,80 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
 	}
-	if err := copyBreakpointInfo(bp, requestedBp); err != nil {
-		if _, err1 := d.target.ClearBreakpoint(bp.Addr); err1 != nil {
-			err = fmt.Errorf("error while creating breakpoint: %v, additionally the breakpoint could not be properly rolled back: %v", err, err1)
+	d.log.Infof("created breakpoint: %#v", createdBp)
+	return createdBp, nil
+}
+
+// createLogicalBreakpoint creates one physical breakpoint for each address
+// ins addrs and associates all of them with the same logical breakpoint.
```
derekparker: s/ins/in/

service/debugger/debugger.go
```diff
@@ -442,45 +443,80 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
...
+func createLogicalBreakpoint(p proc.Process, addrs []uint64, requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
+	bps := make([]*proc.Breakpoint, len(addrs))
+	var err error
+	for i := range addrs {
+		var err2 error
```
derekparker: Why have err and err2 here? We assign `err = err2` on error and break and check `err` anyways.

aarzilli:
Done.

derekparker:
We've marked `api.Breakpoint.Addr` as deprecated but our terminal frontend still uses it. Might be better to update the code which prints out breakpoint information to show all addresses and maybe explain that some belong to inlined functions.

aarzilli:
I've collected all changes to command.go, along with other changes to service/debugger, in a follow up commit.

PR #1718, opened 2019-10-14
service: add min/max supported ver and target ver to GetVersion resp

### LINKED ISSUE #1713: Feature: to include a go compiler version into GetVersion result
It would be great to get the go compiler version so that the client can implement an analog of the `--check-go-version` logic.

aarzilli:
```
service: add min/max supported ver and target ver to GetVersion resp

Fixes #1713

```

### DIFF
--- a/pkg/goversion/compat.go
+++ b/pkg/goversion/compat.go
@@ -5,10 +5,12 @@ import (
 )
 
 var (
-	minSupportedVersionOfGoMinor = 11
-	maxSupportedVersionOfGoMinor = 13
-	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version 1.%d, suppress this error with --check-go-version=false)", minSupportedVersionOfGoMinor)
-	dlvTooOldErr                 = fmt.Errorf("Version of Delve is too old for this version of Go (maximum supported version 1.%d, suppress this error with --check-go-version=false)", maxSupportedVersionOfGoMinor)
+	MinSupportedVersionOfGoMajor = 1
+	MinSupportedVersionOfGoMinor = 11
+	MaxSupportedVersionOfGoMajor = 1
+	MaxSupportedVersionOfGoMinor = 13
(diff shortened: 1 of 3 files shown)


service/debugger/debugger.go
```diff
@@ -1261,6 +1261,12 @@ func (d *Debugger) GetVersion(out *api.GetVersionOut) error {
 			out.Backend = d.config.Backend
 		}
 	}
+
+	out.TargetGoVersion = d.target.BinInfo().Producer()
+
+	out.MinSupportedVersionOfGo = fmt.Sprintf("1.%d.0", goversion.MinSupportedVersionOfGoMinor)
```
derekparker: We should make the major version a const as well just for consistency and so we don't have to change it later.

pkg/goversion/compat.go
```diff
@@ -5,10 +5,12 @@ import (
...
 var (
-	minSupportedVersionOfGoMinor = 11
-	maxSupportedVersionOfGoMinor = 13
-	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version 1.%d, suppress this error with --check-go-version=false)", minSupportedVersionOfGoMinor)
-	dlvTooOldErr                 = fmt.Errorf("Version of Delve is too old for this version of Go (maximum supported version 1.%d, suppress this error with --check-go-version=false)", maxSupportedVersionOfGoMinor)
+	MinSupportedVersionOfGoMajor = 1
+	MinSupportedVersionOfGoMinor = 11
+	MaxSupportedVersionOfGoMajor = 1
+	MaxSupportedVersionOfGoMinor = 13
+	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version 1.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMinor)
```
derekparker: nit: please use new major version const in string interpolation.

aarzilli: Done.

Issue #118, opened 2015-05-03
ARM support


derekparker:
Add support for ARM processors.

tylerarnold:
yes please for ARM processors. I'd love to help but my bandwidth is limited at the moment.

ci-iotsyst:
If this helps, when compiling, I get:

    src/github.com/derekparker/delve/proc/disasm.go:9: undefined: ArchInst

dt-rush:
Can anyone familiar with the technical hurdles give a high level breakdown of what they are?

willwp:
@dt-rush The overall technical hurdle is adding in the arm64 disassembler golang files in /pkg/proc/disasm_arm64.go.  You can find my work on the forked delve repo, [arm64_fix branch](https://github.com/WinstonPrivacyInc/delve/tree/arm64_fix). 
I've taken the arch/arm64 files from the [golang source tree](https://github.com/golang/go/tree/master/src/cmd/vendor/golang.org/x/arch/arm64/arm64asm) and put them in the [vendor/golang.org/x/arch directory](https://github.com/WinstonPrivacyInc/delve/tree/arm64_fix/vendor/golang.org/x/arch/arm64/arm64asm) in the arm64_fix branch of delve.  
The technical hurdles are as follows:

-  There are still calls to the Intel instruction set that need to be converted to arm64/RISC operations in [/pkg/proc/diasm_arm64.go](https://github.com/WinstonPrivacyInc/delve/blob/arm64_fix/pkg/proc/disasm_arm64.go)  Right now there are calls in disasm_arm64.go to CALL, LCALL, and Rel, Segment, Base, Index, and Imm that need to be ported from x86 to arm64 procedures.
- Verification and testing that arm64 conversion works.

alanjds:
I will probable be asking too much, but ARM32 would be nice too.

gidoBOSSftw5731:
whats the status here, is there any way people can contribute to help? I want to code on a chromebook but this makes it hard

christoofar:
:-(

```pi@temp:~/go/src/github.com/derekparker/delve $ make install
go install "-ldflags=-X main.Build=34e802a42b9dc3ba3c4a513b50e9eec6afedcdf2" github.com/derekparker/delve/cmd/dlv
# github.com/derekparker/delve/pkg/proc
pkg/proc/disasm.go:12:14: undefined: archInst```

christoofar:
[...]

PR #1724, opened 2019-10-21
delve: Separate amd64-arch code from the project.

hengwu0:
In order to implement other arch( arm64 ), we need to separate amd64-arch code firstly. This patch do not change any code, except _ArchInfo_ function in _arch_amd64.go_. But i also correct the _dwarf_expr_test.go_, to make sure the testsuits pass.

#118


hengwu0:
> In order to implement other arch( arm64 ), we need to separate amd64-arch code firstly. This patch do not change any code, except _ArchInfo_ function in _arch_amd64.go_. But i also correct the _dwarf_expr_test.go_, to make sure the testsuits pass.

#118

aarzilli:
No, I don't think this is what we want. I can't see any reason why the AMD64 implementation of Arch should be compiled in conditionally. I think an ARM64 build of delve should be able to read AMD64 core files and vice versa. Same thing goes from every other change. There's a reason linux_amd64_core was named linux_amd64_core and not core_linux_amd64, the reason is that it should be compiled in, no matter what architecture or operating system is doing the build.

hengwu0:
> No, I don't think this is what we want. I can't see any reason why the AMD64 implementation of Arch should be compiled in conditionally. I think an ARM64 build of delve should be able to read AMD64 core files and vice versa. Same thing goes from every other change. There's a reason linux_amd64_core was named linux_amd64_core and not core_linux_amd64, the reason is that it should be compiled in, no matter what architecture or operating system is doing the build.

Okey, I agree this idea can let delve be easy-to-use, but hard to code of course. Because that means we should use different funcs according to the arch. I think the best code is like `go objdump` tool, it calls the methods in the class, and different arch use different class. Or just add an parameter in each func which we need. Whatever, we may need to change almost 20 files showed in this patch, to adapt multiple architectures. What's your idea, use class definition? I hope I can help, to let it support arm64.

aarzilli:
Yes, it should be similar to what `go objdump` does. 

> Whatever, we may need to change almost 20 files showed in this patch, to adapt multiple architectures

I'm not sure how many files you would need to change because I'm not sure what the scope of this PR would be. I suggest you do not try to simultaneously port proc/core, proc/native and proc/gdbserial all at once to arm64 and instead just pick one backend and only do that one.

hengwu0:
> I'm not sure what the scope of this PR would be.

This PR will be closed, because there is no need to separate amd64 code with that compatibility goals.

>  I suggest you do not try to simultaneously port proc/core, proc/native and proc/gdbserial all at once to arm64 and instead just pick one backend and only do that one.

I think I will pick _proc/native_, because the local debug is a foundamental function i think. The arch, regs, disasm already had architecture attribute, beside _/proc/stack.go_. As arm64 need different method to parse stack-frame,  i think it should be transformed first.

aarzilli:
stack.go executes instructions read from debug_frame, I expect few changes to that file.

hengwu0:
> stack.go executes instructions read from debug_frame, I expect few changes to that file.

yes, but the instructions are different between arm64 and amd64. So, there may have two things: 
1、pack stack func with interface, or add arch parameter.
2、separate amd64-arch code from proc/native.
the first one seems code frame related, hope you can help. Thanks.

PR #1728, opened 2019-10-22
terminal: let 'list' work on file:line exprs that don't map to code

aarzilli:
```
terminal: let 'list' work on file:line exprs that don't map to code

Make the 'list' command succeed for file:line expressions that don't
map to any instruction.
Adds an argument to the FindLocations API call that makes FindLocations
return if the expression can be parsed, even if it doesn't end up
matching any instruction in debug_line.

```

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -1145,7 +1145,7 @@ func (d *Debugger) convertDefers(defers []*proc.Defer) []api.Defer {
 }
 
 // FindLocation will find the location specified by 'locStr'.
-func (d *Debugger) FindLocation(scope api.EvalScope, locStr string) ([]api.Location, error) {
+func (d *Debugger) FindLocation(scope api.EvalScope, locStr string, includeNonExecutableLines bool) ([]api.Location, error) {
 	d.processMutex.Lock()
 	defer d.processMutex.Unlock()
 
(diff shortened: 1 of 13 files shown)


derekparker:
What's the use case for this?

aarzilli:
`edit` and `list` need to convert a location expression into a (file, line) pair. When the expression is given as <file>:<line> already all they need to do is figure out the full path of file. It's weird that they can only do that when there happens to be an instruction associated with the line in question.

If I did `list blah.go:1` I would almost certainly get an error because the first line of blah.go is typically the package statement and it doesn't have an associated instruction. But we don't need that to just print some text from a file, it doesn't make much sense that we do unless you are privy to the internals of delve.

service/debugger/debugger.go
```diff
@@ -1145,7 +1145,7 @@ func (d *Debugger) convertDefers(defers []*proc.Defer) []api.Defer {
 }
 
 // FindLocation will find the location specified by 'locStr'.
-func (d *Debugger) FindLocation(scope api.EvalScope, locStr string) ([]api.Location, error) {
+func (d *Debugger) FindLocation(scope api.EvalScope, locStr string, returnLines bool) ([]api.Location, error) {
```
derekparker: We should be consistent here and name `returnLines` to `findInstruction`, it also seems more clear to me.

aarzilli: To keep backwards compatibility the default value of the API field (false) must produce the old behavior, hence the switch between findInstruction and returnLines. We could use findInstruction here but the member field that the server API takes needs to stay returnLines (or another name with the same semantics)

derekparker: Ack, good call. How about `includeNonExecutableLines` or something to that effect? I just want to make it a little clearer what is going on and have the name be a bit closer to what we're using elsewhere.

Don't mean to get too dragged down on bikeshedding.

PR #1733, opened 2019-10-24
proc/native: imlement arm64

hengwu0:
As `proc/native` is arch related, it should move some functions to arch-relate file. And this patch can help us to separate the architecture code, make code tidy. So that the merge of arm64 code later will not cause chaos.

#118


aarzilli:
I'd prefer to see this as part of an actual arm64 support patch.

hengwu0:
> I'd prefer to see this as part of an actual arm64 support patch.

Do you mean I should push a full patch here? I reviewed https://github.com/go-delve/delve/pull/1690, and you saied "_Several changes have been made to this file, after the rename, including a global reformatting. This PR should only rename this file and make no further changes._" at _pkg/proc/native/threads_linux_amd64.go_.

aarzilli:
That PR removed threads_linux.go and replaced it with threads_linux_amd64.go and threads_linux_arm64.go, in that circumstance I would expect threads_linux_amd64.go to be equal to threads_linux.go prior to the PR. This commit is fine but I would like to see the rest of the port too, either as other commits in the PR or as another PR.

hengwu0:
delve: support linux-arm64 native debug
* delve now can be built to arm64-arch and running on linux-arm64 OS.  
* arm64 general-purpose registers have completed. 
* arm64 disasm has completed.


Now you can debug an simple program in linux arm64 OS by delve native debug, like this:
```test.go
package main

import "fmt"

func main() {
        fmt.Println("hello arm64 delve!")
}
```

Since this patch just completed general-purpose registers and disasm, you should only use related commands( like disass, regs) and delve's own command( like b, bp, c).

Since this patch is huge and there is no linux-arm64 CI, I will push linux-arm64 testcases next time. This patch completed a fundamental native debug, let's improve it step by step.

pkg/proc/disasm_amd64_arch.go
```diff
@@ -195,3 +197,47 @@ func checkPrologue(s []AsmInstruction, prologuePattern instrseq) bool {
 	}
 	return true
 }
+
+func disassemble(memrw MemoryReadWriter, regs Registers, breakpoints *BreakpointMap, bi *BinaryInfo, startAddr, endAddr uint64, singleInstr bool) ([]AsmInstruction, error) {
```
aarzilli: As far as I can tell the only difference between this and the disassemble in disasm_amd64 is the number of bytes advanced for an unknown instruction (1 here, 4 for arm64). I think this function should be left inside disasm.go and take an `arch Arch` argument. Then:

1. Add a `MinInstructionLength` member to `type Arch interface`
2. Define `func (a *AMD64) MinInstructionLength() int { return 1 }` and `func (a *ARM64) MinInstructionLength() int { return 4 }`
3. Call MinInstructionLength in this function.

pkg/proc/arm64_arch.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	return 8 // general registers
+}
+
+// The mapping between hardware registers and DWARF registers is specified
+// in the DWARF for the ARM® Architecture page 7,
+// Table 1
+// http://infocenter.arm.com/help/topic/com.arm.doc.ihi0040b/IHI0040B_aadwarf.pdf
+var asm64DwarfToHardware_arm64 = map[int]arm64asm.Reg{
```
aarzilli: Rename this to arm64DwarfToHardware

pkg/proc/amd64_arch.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	return 8
+}
+
+// The mapping between hardware registers and DWARF registers is specified
+// in the System V ABI AMD64 Architecture Processor Supplement page 57,
+// figure 3.36
+// https://www.uclibc.org/docs/psABI-x86_64.pdf
+
+var asm64DwarfToHardware = map[int]x86asm.Reg{
```
aarzilli: Rename this to amd64DwarfToHardware, this was always a typo.

pkg/proc/disasm_amd64_arch.go
```diff
@@ -1,3 +1,5 @@
+// +build amd64
```
aarzilli: I think this file should be left with its name `disasm_amd64.go`. Instead add this note:

```
// TODO: disassembler support should be compiled in unconditionally, instead of being decided by the build-target architecture, and be part of the Arch object instead.
```

pkg/proc/native/registers_linux_arm64.go
```diff
@@ -0,0 +1,113 @@
...
+		return err
+	}
+	r := ir.(*linutil.ARM64Registers)
+	r.Regs.Sp = sp
+	thread.dbp.execPtraceFunc(func() { err = ptraceSetGRegs(thread.ID, r.Regs) })
+	return
+}
+
+func (thread *Thread) SetDX(dx uint64) (err error) {
+	/*var ir proc.Registers
```
aarzilli: Remove commented out code.

pkg/proc/native/registers_linux_arm64.go
```diff
@@ -0,0 +1,113 @@
...
+		if err == syscall.ENODEV {
+			err = nil
+		}
+		return
+	} else {
+		err = nil
+	}
+
+	fpregset = arm64_fpregs[:iov.Len-8]
+	//err = proc.LinuxX86XstateRead(regset.Xsave, false, &regset)
```
aarzilli: Remove commented out code.

pkg/proc/registers_arm64_arch.go
```diff
@@ -0,0 +1,52 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"os"
+)
+
+// AppendFPReg appends a 128 bit FP register to regs.
+func AppendFPReg(regs []Register, name string, reg_value []byte) []Register {
+	buf := bytes.NewReader(reg_value)
+	/*type Reader struct {
```
aarzilli: Remove commented out code.

pkg/proc/registers_amd64_arch.go
```diff
@@ -7,7 +7,7 @@ import (
 	"golang.org/x/arch/x86/x86asm"
 )
 
-var dwarfToAsm = map[int]x86asm.Reg{
+var dwarfToAsm_amd64 = map[int]x86asm.Reg{
```
aarzilli: It looks like we aren't using this file anymore. It was superceded by the RegistersToDwarfRegisters stuff in arch.go. Delete it.

aarzilli:
BTW travis-ci now supports arm64 so you can just go ahead and add this to .travis.yml:

```
arch:
  - amd64
  - arm64
```

pkg/proc/disasm_arm64_arch.go
```diff
@@ -0,0 +1,212 @@
...
+	case GNUFlavour:
+		text = arm64asm.GNUSyntax(inst.Inst.Inst)
+	default:
+		text = arm64asm.GoSyntax(inst.Inst.Inst, inst.Loc.PC, bi.symLookup, nil)
+	}
+
+	return text
+}
+
+// IsCall returns true if the instruction is a CALL or LCALL instruction.
```
derekparker: nit: Comment mentioned incorrect instructions.

pkg/proc/disasm_arm64_arch.go
```diff
@@ -0,0 +1,212 @@
...
+
+// IsRet returns true if the instruction is a RET or ERET instruction.
+func (inst *AsmInstruction) IsRet() bool {
+	if inst.Inst == nil {
+		return false
+	}
+	return inst.Inst.Inst.Op == arm64asm.RET || inst.Inst.Inst.Op == arm64asm.ERET
+}
+
+func resolveCallArg(inst *archInst, inst_addr uint64, currentGoroutine bool, regs Registers, mem MemoryReadWriter, bininfo *BinaryInfo) *Location {
```
derekparker: s/inst_addr/instAddr/ per typical Go style guide.

pkg/proc/disasm_arm64_arch.go
```diff
@@ -0,0 +1,212 @@
...
+// instruction is added by load_g_cx in the same file and is either 1 or 2
+// MOVs.
+var prologues []instrseq
+
+func init() {
+	var tinyStacksplit = instrseq{arm64asm.MOV, arm64asm.CMP, arm64asm.B}
+	var smallStacksplit = instrseq{arm64asm.SUB, arm64asm.CMP, arm64asm.B}
+	var bigStacksplit = instrseq{arm64asm.CMP, arm64asm.B, arm64asm.ADD, arm64asm.SUB, arm64asm.MOV, arm64asm.CMP, arm64asm.B}
+	var unixGetG = instrseq{arm64asm.MOV}
+	//var windowsGetG = instrseq{arm64asm.MOV, arm64asm.MOV}
```
derekparker: Why comment this out?

pkg/proc/disasm_arm64_arch.go
```diff
@@ -0,0 +1,212 @@
...
+
+import (
+	"golang.org/x/arch/arm64/arm64asm"
+)
+
+var maxInstructionLength uint64 = 15
+
+type archInst struct {
+	Inst arm64asm.Inst
+	Len  int
```
derekparker: Do we need this? `asmDecode` always puts length at 4, which as I understand it is the instruction length in ARM state, as opposed to 2 in thumb state. Can we always assume ARM state (and thus remove this field and just return 4 in `(inst *archInst) Size()`? If not, is there a way to infer from `arch.Instr` the size or at least which mode we're in?

pkg/proc/disasm_arm64_arch.go
```diff
@@ -0,0 +1,212 @@
...
+				}
+			}
+			return r.Loc.PC, nil
+		}
+	}
+
+	return fn.Entry, nil
+}
+
+func checkPrologue(s []AsmInstruction, prologuePattern instrseq) bool {
```
derekparker: Per my above comment about instruction size, we might be able to consolidate this function as opposed to duplicating is just because we need to access `Inst.Inst`.

pkg/proc/linutil/regs_arm64_arch.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"fmt"
+	"golang.org/x/arch/arm64/arm64asm"
+
+	"github.com/go-delve/delve/pkg/proc"
+)
+
+// Regs is a wrapper for sys.PtraceRegs.
+type ARM64Registers struct {
+	Regs     *ARM64PtraceRegs //通用寄存器
```
derekparker: Please convert comments to English.

pkg/proc/amd64_arch.go
hengwu0: modified

pkg/proc/registers_arm64_arch.go
hengwu0: modified.

pkg/proc/registers_amd64_arch.go
hengwu0: modified.

pkg/proc/disasm_arm64_arch.go
hengwu0: modified.

pkg/proc/disasm_arm64_arch.go
hengwu0: modified.

pkg/proc/disasm_arm64_arch.go
hengwu0: Because go arm64 do not support windows for now.(go1.12)

pkg/proc/disasm_arm64_arch.go
hengwu0: I confirmed that there is no thumb code in arm64, it only exists in arm arch. So, I deleted len arg. Please review.

aarzilli:
> Because it will not test the arm64 code of involving function calls(this func don't have arm64 code in fact). All tests run in amd64-arch should pass of course, because i have not change any amd64 code. BTW, the CI running on arm64 may be failed for now. let's improve it step by step.

Yes, improving it later is fine. The FreeBSD patch was also merged incomplete. However I'd at least like to know how many of the tests pass/fail. Could you add a skip to the tests that don't work on arm64?

```
    t.Skip("broken on arm64")
}
```

hengwu0:
I don't know why the test timeout, since there is no go-test-timeout-log. In my arm64 device, it test passed for `pkg/proc`. There already have an timeout in golang testing, could you cancel the timeout of CI? @aarzilli

derekparker:
@hengwu0 I've restarted the jobs that timed out.

aarzilli:
You can comment out line scripts/make.go:259 to get verbose output from the tests. It may provide a clue about which test might be stalling.

pkg/proc/proc_test.go
```diff
@@ -1339,6 +1375,9 @@ func TestRecursiveStructure(t *testing.T) {
 }
 
 func TestIssue316(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Eval Variable for now.")
```
aarzilli: I'll be honest. I don't see much of a point in merging an arm64 support patch that can't read from memory.

hengwu0:
Looked into the failed tests, the `Eval Variable` function should be well, the terminal of native debug output showed as follow:

```sh
(dlv) c
> main.main() ./testvariables2.go:310 (hits goroutine(1):1 total:1) (PC: 0xa6220)
   305:                 },
   306:         }
   307:
   308:         ll := &List{0, &List{1, &List{2, &List{3, &List{4, nil}}}}}
   309:
=> 310:         var amb1 = 1
   311:         runtime.Breakpoint()
   312:         for amb1 := 0; amb1 < 10; amb1++ {
   313:                 fmt.Println(amb1)
   314:         }
   315:
(dlv) p iface5
interface {}(*main.dstruct) *{
        x: interface {}(*main.dstruct) *{
                x: interface {}(*main.dstruct) ...,},}
(dlv) p aas
[]main.a len: 1, cap: 1, [
        {
                aas: []main.a len: 1, cap: 1, [
                        (*main.a)(0x400008e0a0),
                ],},
]
(dlv) 
```

I found that the reason is arm64-delve do not support `runtime.Breakpoint()` for now. It will hang on after that go-func executed. So, if the test have the runtime.Breakpoint() func, it will be timeouted certainly.
It seems lots tests have `runtime.Breakpoint()`, including the new test TestIssue1736(which lead to timeout of CI now), I think the testsuits modify should not be merged with _support arm64 delve_ together.

derekparker:
> Looked into the failed tests, the `Eval Variable` function should be well, the terminal of native debug output showed as follow:
> 
> ```shell
> (dlv) c
> > main.main() ./testvariables2.go:310 (hits goroutine(1):1 total:1) (PC: 0xa6220)
>    305:                 },
>    306:         }
>    307:
>    308:         ll := &List{0, &List{1, &List{2, &List{3, &List{4, nil}}}}}
>    309:
> => 310:         var amb1 = 1
>    311:         runtime.Breakpoint()
>    312:         for amb1 := 0; amb1 < 10; amb1++ {
>    313:                 fmt.Println(amb1)
>    314:         }
>    315:
> (dlv) p iface5
> interface {}(*main.dstruct) *{
>         x: interface {}(*main.dstruct) *{
>                 x: interface {}(*main.dstruct) ...,},}
> (dlv) p aas
> []main.a len: 1, cap: 1, [
>         {
>                 aas: []main.a len: 1, cap: 1, [
>                         (*main.a)(0x400008e0a0),
>                 ],},
> ]
> (dlv) 
> ```
> 
> I found that the reason is arm64-delve do not support `runtime.Breakpoint()` for now. It will hang on after that go-func executed. So, if the test have the runtime.Breakpoint() func, it will be timeouted certainly.
> It seems lots tests have `runtime.Breakpoint()`, including the new test TestIssue1736(which lead to timeout of CI now), I think the testsuits modify should not be merged with _support arm64 delve_ together.

That is because Delve is programmed to handle SIGTRAP generated from a breakpoint instruction being hit, however it seems that `runtime.Breakpoint()` is actually executing an undefined instruction, which happens to be known in GDB to be a breakpoint: https://github.com/golang/go/blob/master/src/runtime/asm_arm.s#L185. We need to teach Delve the same thing.

aarzilli:
It is likely that runtime.Breakpoint on arm64 generates a SIGILL signal, to handle that correctly it would have to be caught in trapWaitInternal (proc_linux.go) and read the memory at PC to find out the instruction value. I think this is worth doing here to see how many of the skipped tests are just this, as opposed to other problems.

hengwu0:
> > Looked into the failed tests, the `Eval Variable` function should be well, the terminal of native debug output showed as follow:
> > ```shell
> > (dlv) c
> > > main.main() ./testvariables2.go:310 (hits goroutine(1):1 total:1) (PC: 0xa6220)
> >    305:                 },
> >    306:         }
> >    307:
> >    308:         ll := &List{0, &List{1, &List{2, &List{3, &List{4, nil}}}}}
> >    309:
> > => 310:         var amb1 = 1
> >    311:         runtime.Breakpoint()
> >    312:         for amb1 := 0; amb1 < 10; amb1++ {
> >    313:                 fmt.Println(amb1)
> >    314:         }
> >    315:
> > (dlv) p iface5
> > interface {}(*main.dstruct) *{
> >         x: interface {}(*main.dstruct) *{
> >                 x: interface {}(*main.dstruct) ...,},}
> > (dlv) p aas
> > []main.a len: 1, cap: 1, [
> >         {
> >                 aas: []main.a len: 1, cap: 1, [
> >                         (*main.a)(0x400008e0a0),
> >                 ],},
> > ]
> > (dlv) 
> > ```
> > 
> > 
> > I found that the reason is arm64-delve do not support `runtime.Breakpoint()` for now. It will hang on after that go-func executed. So, if the test have the runtime.Breakpoint() func, it will be timeouted certainly.
> > It seems lots tests have `runtime.Breakpoint()`, including the new test TestIssue1736(which lead to timeout of CI now), I think the testsuits modify should not be merged with _support arm64 delve_ together.
> 
> That is because Delve is programmed to handle SIGTRAP generated from a breakpoint instruction being hit, however it seems that `runtime.Breakpoint()` is actually executing an undefined instruction, which happens to be known in GDB to be a breakpoint: https://github.com/golang/go/blob/master/src/runtime/asm_arm.s#L185. We need to teach Delve the same thing.

Thanks, but i think it should be this file: https://github.com/golang/go/blob/master/src/runtime/asm_arm64.s#L93

derekparker:
> > That is because Delve is programmed to handle SIGTRAP generated from a breakpoint instruction being hit, however it seems that `runtime.Breakpoint()` is actually executing an undefined instruction, which happens to be known in GDB to be a breakpoint: https://github.com/golang/go/blob/master/src/runtime/asm_arm.s#L185. We need to teach Delve the same thing.
> 
> Thanks, but i think it should be this file: https://github.com/golang/go/blob/master/src/runtime/asm_arm64.s#L93

Ack, still requires some investigation then as to why Delve is hanging when that is hit. Have you verified control is being handed back to Delve? Are there issues with the handling of `runtime.breakpoint`? We do special things there to step out of that and the enclosing functions.

pkg/proc/disasm_arm64_arch.go
derekparker: Let's remove it then instead of just commenting it out until it's supported.

hengwu0: modified.

hengwu0:
> Ack, still requires some investigation then as to why Delve is hanging when that is hit. Have you verified control is being handed back to Delve? Are there issues with the handling of `runtime.breakpoint`? We do special things there to step out of that and the enclosing functions.

Yes, i investigated and now fixed. The reason is this: PtraceSingleStep seems cannot step over BRK instruction(linux-arm64 feature or kernel bug maybe). And gdb has the same question too, it will hang on forever with `c` command or execute that instruction indefinitely with `s`,`si` command.

hengwu0:
Now, all tests should passed, but it seems there is some error with your CI config?

hengwu0:
travis check fixed by https://github.com/go-delve/delve/pull/1752/commits/aa85dd947759a285a9d546b1d6ede5e45905ee7b. So, now all tests passed.

hengwu0:
travis check fixed by https://github.com/go-delve/delve/pull/1752/commits/aa85dd947759a285a9d546b1d6ede5e45905ee7b. So, now all tests passed. @aarzilli

pkg/proc/proc_test.go
```diff
@@ -2082,6 +2124,9 @@ func TestIssue573(t *testing.T) {
 }
 
 func TestTestvariables2Prologue(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
aarzilli: How is this test failing exactly? We're just checking that we can detect the prologue correctly. It should work and if it isn't then that's probably the reason why several other tests aren't working.

service/test/integration1_test.go
```diff
@@ -191,6 +191,9 @@ func Test1ClientServer_exit(t *testing.T) {
 }
 
 func Test1ClientServer_step(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
aarzilli: I find it hard to believe that this test doesn't work but TestClientServer_step without 1 does since they are the exact same tests but using different API.

service/test/integration2_test.go
```diff
@@ -1708,6 +1747,9 @@ func (c *brokenRPCClient) call(method string, args, reply interface{}) error {
 }
 
 func TestUnknownMethodCall(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
aarzilli: This test simply checks that we don't stall on unknown API calls. It's very strange that it wouldn't work on arm64 given that it's completely architecture independent.

hengwu0: I thought it was a test for func call in debug program execute before, so i skiped it. But I still think it should be skiped for arm64-delve check until remote debug released on arm64.

service/test/integration1_test.go
hengwu0: The remote debug code is not checked, and i tested it may failed contingently before(it may ok after https://github.com/go-delve/delve/pull/1733/commits/b1369a96231ebe9028e576fe61e60b593b655aa2). I think it should be skiped for arm64-delve check until remote debug released on arm64.

pkg/proc/proc_test.go
hengwu0: Yes, it can't skip func prologue correctly on arm64, it still get the start of func include `runtime.morestack` head. I will fix it next time, since it do not affect the debugging on native debug a lot. Let's skip it just for now.

service/test/integration2_test.go
aarzilli: There's no difference between local and remote debugging with delve.

pkg/proc/proc_test.go
aarzilli: Prologue skipping does in fact affect debugging a lot, I'd rather not skip it.

aarzilli:
The number of disabled tests still needs to go down substantially. At the moment enough tests are disabled that even the gdbserial backend passes the tests on arm64 despite being unchanged. That means that, for example, the code that reads G isn't being exercised at all.

service/test/integration2_test.go
hengwu0: Okey, i will check later.

pkg/proc/proc_test.go
hengwu0: Okey

service/test/integration2_test.go
hengwu0: Okey, i will check later.

hengwu0:
> The number of disabled tests still needs to go down substantially. At the moment enough tests are disabled that even the gdbserial backend passes the tests on arm64 despite being unchanged. That means that, for example, the code that reads G isn't being exercised at all.

This patch only completed to build arm64-arch, disass, general-purpose registers. At least `breakpoint`, `c`, `s` and `p` this basis commands can work now. Can https://github.com/go-delve/delve/pull/1733/commits/07ef21c87d4b2e94482bcaaf858dc9f22748e072 be merged firstly, so that we can improve it together, or how about let arm64-arch tests Allow Failed? Since we do not need to provide arm64-delve in this release. And there is still a lot of work todo to pass tests.

pkg/proc/proc_test.go
hengwu0: Fixed. But why CI failed...

hengwu0:
> The number of disabled tests still needs to go down substantially. At the moment enough tests are disabled that even the gdbserial backend passes the tests on arm64 despite being unchanged. That means that, for example, the code that reads G isn't being exercised at all.

I labeled reason why the tests failed, and I think the code that reads G is exercised now.

pkg/proc/proc_test.go
hengwu0: Fixed.

service/test/integration2_test.go
hengwu0: the skip-code removed.

hengwu0:
What need i do for merge this PR now? The `stacktrace` need to separate amd64-arch code again, and i think it should be pulled in next PR. What's your advice? @aarzilli

pkg/proc/native/threads.go
```diff
@@ -119,9 +119,14 @@ func (t *Thread) SetCurrentBreakpoint(adjustPC bool) error {
 	if err != nil {
 		return err
 	}
+	if _, ok := t.Arch().(*proc.ARM64); ok {
```
aarzilli: This change here is very much not obvious and needs some kind of explanatory comment. Also I don't get it. I thought the way break instructions worked on ARM64 the value of PC would stay on the break instruction, that would make the SetPC we do later a no-op, why is this needed?

.travis.yml
```diff
@@ -5,6 +5,10 @@ os:
   - linux
   - osx
 
+arch:
+  - amd64
+  - arm64
```
aarzilli: This file has a line that says `matrix:` in it, under it add this:

```
  exclude:
    - os: osx
      arch: arm64
```

note that the first line starts with two spaces.

pkg/proc/native/threads.go
hengwu0: In amd64, the PC would stay on the next instrcution of breakpoint when hitting the breakpoint. So, we need to adjust the PC back to the place of breakpoint by `SetPC()`, and then replace the breakpoint instrcution by user's origin instruction. The `FindBreakpoint()` can help us to know wheather there is hitting the breakpoint, and it will search the previous instrcution when `adjustPC` is true.
In arm64, the PC would stay on the break instruction when hitting the breakpoint. If we execute `si` command then, after single instruction executed, delve will go into this function, and `FindBreakpoint()` will return the breakpoint location(previous instrcution) and the `SetPC()` will let PC back to the breakpoint location. This will cause `si` to not step one instruction.
In arm64, we donot need to search previous instrcution, so the `adjustPC` should be always false. To the same, there is no need to execute `SetPC()` if `adjustPC` is false.

hengwu0: > This change here is very much not obvious and needs some kind of explanatory comment. 

I write the comment in the head of patch https://github.com/go-delve/delve/pull/1733/commits/6f0cebdb5832eec8660b7e7b5f26913daaed1e96. I think it should be moved in sourcecode, and explain it fully。

aarzilli: Ok. I get it now.

Let's add a new method to Arch:

```
BreakInstrMovesPC() bool
```

and have it return true for AMD64 and false for ARM64

then on this line do:

```
// If the breakpoint instruction does not change the value of PC after being executed we should look for breakpoints with bp.Addr == PC and there is no need to call SetPC after finding one. 
adjustPC = adjustPC && t.Arch().BreakMovesPC()
```

.travis.yml
hengwu0: modified.

pkg/proc/native/threads.go
hengwu0: modified.

pkg/proc/proc_test.go
```diff
@@ -566,6 +572,9 @@ func TestNextConcurrentVariant2(t *testing.T) {
 	if runtime.GOOS == "freebsd" {
 		t.Skip("test is not valid on FreeBSD")
 	}
+	if runtime.GOARCH == "arm64" {
```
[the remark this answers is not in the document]
hengwu0: This test should be ok, we can remove the skip-code directly.

hengwu0:
Can merge now?

pkg/proc/linutil/regs_arm64_arch.go
```diff
@@ -0,0 +1,139 @@
...
+func (r *ARM64Registers) SP() uint64 {
+	return r.Regs.Sp
+}
+
+func (r *ARM64Registers) BP() uint64 {
+	return r.Regs.Regs[29]
+}
+
+// CX returns the value of RCX register.
+func (r *ARM64Registers) CX() uint64 {
```
shawn-xdji: Guess it's not necessary for ARM64.

hengwu0: Firstly, we need those to build delve! Secondly, it's the simplest to keep it until all arm64 functions are completed.
The `Registers interface` defined as followed:

```go
proc/registers.go: 
type Registers interface {
	PC() uint64
	SP() uint64
	BP() uint64
	CX() uint64
	TLS() uint64
	// GAddr returns the address of the G variable if it is known, 0 and false otherwise
	GAddr() (uint64, bool)
	Get(int) (uint64, error)
	Slice(floatingPoint bool) []Register
	// Copy returns a copy of the registers that is guaranteed not to change
	// when the registers of the associated thread change.
	Copy() Registers
}
```

pkg/proc/proc_test.go
```diff
@@ -154,6 +154,9 @@ func TestExit(t *testing.T) {
 }
 
 func TestExitAfterContinue(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
derekparker: Why is this test skipped? It's only checking that we get an exit notification from the process, this shouldn't be dependent on CPU arch.

pkg/proc/disasm_arm64.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+}
+
+// firstPCAfterPrologueDisassembly returns the address of the first
+// instruction after the prologue for function fn by disassembling fn and
+// matching the instructions against known split-stack prologue patterns.
+// If sameline is set firstPCAfterPrologueDisassembly will always return an
+// address associated with the same line as fn.Entry
+func firstPCAfterPrologueDisassembly(p Process, fn *Function, sameline bool) (uint64, error) {
```
derekparker: I think much of this routine (and `checkPrologue`) is arch independent aside from the actual assembly that's being verified. I think if possible let's share this code instead of duplicating it.

pkg/proc/native/threads_linux_arm64.go
```diff
@@ -0,0 +1,43 @@
...
+	t.dbp.execPtraceFunc(func() {
+		restoreRegistersErr = ptraceSetGRegs(t.ID, sr.Regs)
+		if restoreRegistersErr != syscall.Errno(0) {
+			return
+		}
+		if sr.Fpregset != nil {
+			iov := sys.Iovec{Base: &sr.Fpregset[0], Len: uint64(len(sr.Fpregset))}
+			_, _, restoreRegistersErr = syscall.Syscall6(syscall.SYS_PTRACE, sys.PTRACE_SETREGSET, uintptr(t.ID), uintptr(elf.NT_FPREGSET), uintptr(unsafe.Pointer(&iov)), 0, 0)
+		}
+		//return
```
derekparker: Remove commented code.

pkg/proc/proc_test.go
```diff
@@ -808,6 +811,9 @@ func (l1 *loc) match(l2 proc.Stackframe) bool {
 }
 
 func TestStacktrace(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -843,6 +849,9 @@ func TestStacktrace(t *testing.T) {
 }
 
 func TestStacktrace2(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -891,6 +900,9 @@ func stackMatch(stack []loc, locations []proc.Stackframe, skipRuntime bool) bool
 }
 
 func TestStacktraceGoroutine(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -960,6 +972,9 @@ func TestStacktraceGoroutine(t *testing.T) {
 }
 
 func TestKill(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
derekparker: Also curious why this is being skipped on arm. Killing a process should be arch independent.

pkg/proc/proc_test.go
```diff
@@ -1210,6 +1225,9 @@ func TestVariableEvaluation(t *testing.T) {
 }
 
 func TestFrameEvaluation(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -1690,6 +1708,9 @@ func TestIssue384(t *testing.T) {
 }
 
 func TestIssue332_Part1(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -1711,6 +1732,9 @@ func TestIssue332_Part1(t *testing.T) {
 }
 
 func TestIssue332_Part2(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -1868,6 +1892,9 @@ func TestCmdLineArgs(t *testing.T) {
 }
 
 func TestIssue462(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -1897,6 +1924,9 @@ func TestNextParked(t *testing.T) {
 	if runtime.GOOS == "freebsd" {
 		t.Skip("test is not valid on FreeBSD")
 	}
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -1947,6 +1977,9 @@ func TestNextParked(t *testing.T) {
 }
 
 func TestStepParked(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -2270,6 +2303,9 @@ func TestStepConcurrentDirect(t *testing.T) {
 	if runtime.GOOS == "freebsd" {
 		t.Skip("test is not valid on FreeBSD")
 	}
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -2645,6 +2687,9 @@ func getg(goid int, gs []*proc.G) *proc.G {
 }
 
 func TestStacktraceWithBarriers(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/.

pkg/proc/proc_test.go
```diff
@@ -3337,6 +3388,9 @@ func TestCgoSources(t *testing.T) {
 }
 
 func TestSystemstackStacktrace(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("arm64 do not support Stacktrace for now")
```
derekparker: s/do not/does not/ (and for all the rest).

pkg/proc/proc_test.go
```diff
@@ -3521,6 +3584,9 @@ func TestIssue1145(t *testing.T) {
 }
 
 func TestDisassembleGlobalVars(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
derekparker: If I'm not mistaken the disasm has been implemented. Is there any other reason to skip this test?

pkg/proc/proc_test.go
hengwu0: This test passed on my device, but i still skiped it. Because there is an bug in origin `Detach()` function.  It need to investigate later for full information. mentioned in #1770.

pkg/proc/disasm_arm64.go
hengwu0: modified.

pkg/proc/native/threads_linux_arm64.go
hengwu0: modified

pkg/proc/proc_test.go
hengwu0: Mainly may be testsuit's bug, It seems can't find main.v in range of mainfn.Entry and mainfn.End on arm64. Needs to investigate in #1770.

hengwu0:
Fixed. @derekparker

pkg/proc/proc_test.go
derekparker: This is the only thing that concerns me, the rest I understand waiting for follow up patches, but a bug in Detach makes me nervous to merge this right away.

Do you have more details on what the bug is? Aside from this issue I'm alright with merging this and fixing the rest in follow up PRs. @hengwu0 will you be around and have time to continue this work in follow up patches?

hengwu0: > Do you have more details on what the bug is?

It's amd64's bug, and it had not appear in arm64(or unhited when i test on arm64), traced in #1771

> will you be around and have time to continue this work in follow up patches?

Yes, i will. The stacktrace is finished, which will be the next BP(big PR).

hengwu0:
> there is an bug in origin `Detach()` function.

I investigated, and found it's `TestKill`'s bug. here is the detail:
what `Detach(true)` function do:
    1. detach all threads; 2. send SIGINT to tracee.
what `TestKill` do:
    1. Detach(true); 2. Open /proc/%d/
It will fail in `2. Open /proc/tracee_pid/` sporadicly since there is no any sync between send signal and open  /proc/%d/. Especially in some weak arm64 cpu.

hengwu0:
Can merge? @derekparker

hengwu0:
> > there is an bug in origin `Detach()` function.
> 
> I investigated, and found it's `TestKill`'s bug. here is the detail:
> what `Detach(true)` function do:
> 1. detach all threads; 2. send SIGINT to tracee.
> what `TestKill` do:
> 1. Detach(true); 2. Open /proc/%d/
> It will fail in `2. Open /proc/tracee_pid/` sporadicly since there is no any sync between signal sended(tracee handled) and open /proc/%d/. Especially in some weak arm64 cpu.

I think https://github.com/go-delve/delve/pull/1733/commits/b951551e429bc7c8c12ec8e696b571d7f4d3e471 can wait until there is an perfect solution.

aarzilli:
Thank you for this.

hengwu0:
> Thank you for this.

for everyone using arm64-go.

PR #1737, opened 2019-10-26
proc: always resolve array types even if they don't appear in the

### LINKED ISSUE #1736: "function calls not allowed without using 'call'" error on attempt to print a field using direct access syntax
```
Delve Debugger
Version: 1.3.2
Build: d064d1fe05b993e44a844a06df8c5c55c6c5ed5b
```
```
go version go1.12.9 linux/amd64
```
```
linux/amd64
```
4. What did you do?

Debugged the program with breakpoint on line 8:
```
package main

func main() {
	x := make(chan int, 10)
	x <- 1
	x <- 3
	x <- 5
	println(x) // break
}
```
and evaluated channels `buf` field using the direct access syntax `p *(*"[10]int")(824634261520)`.

Contents of the buf field

6. What did you see instead?
Error `function calls not allowed without using 'call'`. Full delve log: https://gist.github.com/nd/6c1fe924252311cb859beddd6f8d3f0d

aarzilli:
```
proc: always resolve array types even if they don't appear in the
program

When evaluating type casts always resolve array types.

Instead of resolving them by looking up the string in debug_info
construct a fake array type so that a type cast to an array type always
works as long as the element type exists.

We already did this for byte arrays, this commit extends this to any
array type. The reason is that we return a fake array type (that
doesn't exist in the target program) for the array of a channel type.

Fixes #1736

```

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -125,7 +125,7 @@ func main() {
 	var fn2 functype = nil
 	var nilslice []int = nil
 	var nilptr *int = nil
-	ch1 := make(chan int, 10)
+	ch1 := make(chan int, 11)
 	ch1 <- 1
 	ch1 <- 4
 	ch1 <- 3
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -15,7 +15,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"reflect"
 	"sort"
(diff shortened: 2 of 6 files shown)


derekparker:
Alright to merge after conflicts resolved.

aarzilli:
> Alright to merge after conflicts resolved.

It's very disappointing that git can never resolve conflicts to proc_test.go given that it's basically an append only file.

PR #1742, opened 2019-11-04
service,terminal: support logical breakpoints

aarzilli:
```
service,terminal: support logical breakpoints

Changes CreateBreakpoint to create a logical breakpoint when multiple
addresses are specified, FindLocation and the api.Location type to
return logical locations and the cli to support logical breakpoints.

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -4,6 +4,7 @@ package terminal
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
 	"go/parser"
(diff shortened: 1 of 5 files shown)


pkg/terminal/command.go
```diff
@@ -2222,9 +2227,24 @@ func formatBreakpointName(bp *api.Breakpoint, upcase bool) string {
 }
 
 func formatBreakpointLocation(bp *api.Breakpoint) string {
+	var out bytes.Buffer
```
derekparker: This might be a little cleaner as a `strings.Builder`.

PR #1746, opened 2019-11-07
Don't stop headless server on SIGINT on windows (go-delve#1745)

nd:
This allows to execute and debug signal handlers in inferior.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -675,9 +675,25 @@ func execute(attachPid int, processArgs []string, conf *config.Config, coreFile 
 		}
 		ch := make(chan os.Signal, 1)
 		signal.Notify(ch, syscall.SIGINT)
-		select {
-		case <-ch:
-		case <-disconnectChan:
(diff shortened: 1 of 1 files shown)


cmd/dlv/cmds/commands.go
```diff
@@ -675,9 +675,25 @@ func execute(attachPid int, processArgs []string, conf *config.Config, coreFile
...
+			// as SIGINT to delve. Ignore it instead of stopping the server
+			// in order to be able to debug signal handlers.
+			go func() {
+				for {
+					select {
+					case <-ch:
+					}
+				}
+			}()
+			select {
```
ProximaB, line 689: @aarzilli  Could it be just one statement  `<-disconnectChan` instead of wrapping it in select?

aarzilli: Sounds right.

ProximaB: I'll make the pull request, alright?

aarzilli: Sure.

PR #1749, opened 2019-11-08
gdbserial: propagate unhandled signals back to a specific thread

### LINKED ISSUE #1744: Cannot debug signal handler on mac

```
Delve Debugger
Version: 1.3.2
Build: 5a947bceffa8aac96574f3f7f50cdede74fc2fa7
```

```
go version go1.13 darwin/amd64
```

MacOS 10.14.6

4. What did you do?

Debug the following program:

```
package main

import (
	"fmt"
	"os"
	"os/signal"
)

func main() {
	sigs := make(chan os.Signal, 1)
	signal.Notify(sigs, os.Interrupt)
	sig :- <- sigs
	fmt.Printf("Got a signal '%v', start shutting down...\n", sig) //break
}

```

and send it a SIGINT signal via `kill -2 <pid>`

Program gets a signal and stops at breakpoint.

6. What did you see instead?

Program doesn't get a signal although it works fine when program is executed without debugger. Debugger log: https://gist.github.com/nd/c5146b71ddfbb836e8c19b9e0ff25e07.

aarzilli:
```
gdbserial: propagate unhandled signals back to a specific thread

Instead of just sending unhandled signals back to the process send them
to the specific thread that received them.
This is important because:

1. debugserver does not appear to support the vCont;CXX packet without
specifying a target thread
2. the non-cooperative preemption change in an upcoming version of Go
(1.15?) will require sending signals to a specific thread.

Fixes #1744

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -355,6 +355,9 @@ func LLDBLaunch(cmd []string, wd string, foreground bool, debugInfoDirs []string
 		if foreground {
 			args = append(args, "--stdio-path", "/dev/tty")
 		}
+		if logflags.LLDBServerOutput() {
+			args = append(args, "-g", "-l", "stdout")
+		}
 		args = append(args, "-F", "-R", fmt.Sprintf("127.0.0.1:%d", listener.Addr().(*net.TCPAddr).Port), "--")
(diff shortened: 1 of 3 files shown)


aarzilli:
cc @nd

nd:
Checked: it does fix #1744, thanks!

PR #1751, opened 2019-11-11
Makefile: discard stderr of "go list"

aarzilli:
```
Makefile: discard stderr of "go list"

In module mode "go" will print messages about downloading modules to
stderr, we shouldn't confuse them for the real command output.

```


aarzilli:
This fixes the Appveyor build now that they have switched to Go 1.13.

derekparker:
Closing, this commit was actually included in https://github.com/go-delve/delve/pull/1755 which was merged.

PR #1752, opened 2019-11-11
proc/linux: do not route signals to threads while stopping

aarzilli:
```
proc/linux: do not route signals to threads while stopping

While we are trying to stop the process we should not route signals
sent to threads because that will result in threads being resumed.
Also keep better track of which threads are stopped.

This fixes an incompatibility with Go 1.14, which sends a lot of
signals to its threads to implement non-cooperative preemption,
resulting in Delve hanging waiting for an already-stopped thread to
stop.

In principle however this bug has nothing to do with Go 1.14 and could
manifest in any instance of high signal pressure.

```


aarzilli:
This fixes the problem with Go 1.14 I was talking about on friday (I don't know why the travis check is stuck on pending, if you click on it, it completed).

PR #1756, opened 2019-11-12
pkg/proc,pkg/dwarf: Introduce loclist package

derekparker:
Move the loclist code out of proc and into its own package in
`pkg/dwarf`.

### DIFF
--- a/pkg/dwarf/loclist/loclist.go
+++ b/pkg/dwarf/loclist/loclist.go
@@ -1,0 +1,81 @@
+package loclist
+
+import (
+	"encoding/binary"
+)
+
+// Reader parses and presents DWARF loclist information.
+type Reader struct {
+	data  []byte
+	cur   int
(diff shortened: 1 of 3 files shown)


pkg/dwarf/loclist/loclist.go
```diff
@@ -0,0 +1,96 @@
...
+		addr := uint64(binary.LittleEndian.Uint64(rdr.read(rdr.ptrSz)))
+		return addr
+	default:
+		panic("bad address size")
+	}
+}
+
+// Entry represents a single entry in the loclist section.
+type Entry struct {
+	lowpc, highpc uint64
```
aarzilli: I think I prefer these and instr to be public and not have reader methods. Like debug/dwarf.LineEntry in the std library.

derekparker:
Done.

PR #1757, opened 2019-11-14
proc: support DW_AT_go_package_name

aarzilli:
```
proc: support DW_AT_go_package_name

Use the name specified by compile unit attribute DW_AT_go_package_name,
introduced in Go 1.13, to map package names to package paths, instead of
trying to deduce it from names of types.
Also use this mapping for resolving global variables and function
expressions.

```


aarzilli:
I'll have to look into windows failure.

aarzilli:
Fixed.

PR #1764, opened 2019-11-21
verify local connections originate from the same UID

stapelberg:
This prevents other users on the same machine (e.g. a production server) from
using dlv under the credentials of another user, which poses a security issue.

### DIFF
--- a/service/rpccommon/sameuser_linux.go
+++ b/service/rpccommon/sameuser_linux.go
@@ -1,0 +1,99 @@
+//+build linux
+
+package rpccommon
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"strings"
+)
+
+// for testing
+var (
+	uid      = os.Getuid()
+	readFile = ioutil.ReadFile
+)
+
+func sameUserForHexLocalAddr(filename, hexaddr string) (bool, error) {
+	b, err := readFile(filename)
+	if err != nil {
+		return false, err
+	}
+	for _, line := range strings.Split(strings.TrimSpace(string(b)), "\n") {
+		// The format contains whitespace padding (%4d, %5u), so we use
+		// fmt.Sscanf instead of splitting on whitespace.
+		var (
+			sl                    int
+			localAddr, remoteAddr string
+			state                 int
+			queue, timer          string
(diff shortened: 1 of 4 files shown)


stapelberg:
Not 100% sure what to make of the travis failure. Given that 2 out of 3 configurations pass, perhaps the remaining one is a flake? Can you restart the run please?

service/rpccommon/sameuser_linux.go
```diff
@@ -0,0 +1,99 @@
...
+	addr, ok := remoteAddr.(*net.TCPAddr)
+	if !ok {
+		panic(fmt.Sprintf("BUG: conn.RemoteAddr is %T, want *net.TCPAddr", remoteAddr))
+	}
+	same, err := sameUserForRemoteAddr(addr)
+	if err != nil {
+		log.Printf("cannot check remote address: %v", err)
+	}
+	if !same {
+		log.Printf("closing connection from different user (%v): connections to localhost are only accepted from the same UNIX user for security reasons (see http://b/138232956)", addr)
```
aarzilli: What's b/138232956?

stapelberg: Sorry about that, an internal reference leaked into this PR here :). That bug serves as a place for more discussion, but that obviously isn’t applicable here.

Removed.

briandealwis:
This is a fairly significant behavioural change.  It broke `skaffold debug`'s support for debugging Go apps in Kubernetes.  I clearly need to develop better integration tests.

There is a work around, to listen on `0.0.0.0`.

aarzilli:
@briandealwis you can also pass `--only-same-user=false`.

briandealwis:
I have a chicken-and-egg problem: I have a deployed base of Skaffold versions that invoke dlv with `--listen=localhost:26486` and do not know of this new option. They install dlv via a container image that is fetched at runtime.

So I can't update to dlv 1.4 until the Skaffold installation base has migrated to a version that uses `--listen=0.0.0.0:26486` — so they won't be bitten by this new behaviour. And then I can stay transitioning back to listening on localhost and disable `--only-same-user`.

stapelberg:
You can also put a `dlv` wrapper into `$PATH`.

briandealwis:
Oh! Good thinking — I'll patch dlv.

Issue #1771, opened 2019-11-27
native/proc: coredump after detach


hengwu0:
1. What version of Delve are you using (`dlv version`)? 
> top version
2. What version of Go are you using? (`go version`)?
> go version go1.11.5 linux/amd64
3. What operating system and processor architecture are you using?
> linux/amd64

```sh
$ cat test.go
package main

import (
        "fmt"
        "os"
        "time"
)

func main() {
        fmt.Printf("pid= %d\n", os.Getpid())
        for i := 0; ; {
                time.Sleep(time.Second)
                fmt.Printf("%d\n", i)
                i++
        }
}

$  ./A
pid= 47418
0
1
trace/breakpoint trap (core dumped)

=====================on another terminal=========================
$ ./dlv_amd64 attach 47418
Type 'help' for list of commands.
(dlv) b test.go:14
Breakpoint 1 set at 0x49e2c6 for main.main() ./test.go:14
(dlv) c
> main.main() ./test.go:14 (hits goroutine(1):1 total:1) (PC: 0x49e2c6)
     9: func main() {
    10:         fmt.Printf("pid= %d\n", os.Getpid())
    11:         for i := 0; ; {
    12:                 time.Sleep(time.Second)
    13:                 fmt.Printf("%d\n", i)
=>  14:                 i++
    15:         }
    16: }
(dlv) q
Would you like to kill the process? [Y/n] y
```

This bug is not be sure to hit, about less than 5% possibility...

PR #1772, opened 2019-11-27
native/proc: coredumped after detach (#1771)

hengwu0:
the tracee may coredumped in detaching: trace/breakpoint trap (core dumped).


hengwu0:
The possibility i can think of is:
  Thread y hits the break point but you have attached to thread x, while detaching.
  This case is a Trace/breakpoint trap.

PR #1774, opened 2019-11-28
proc/core: implementing coredump functionality for ARM64 

ossdev07:
Implementing coredump functionality for ARM64 , now it can be checked by using "dlv core " command.
Floating point registers are not handled yet, working on it.

### DIFF
--- a/pkg/proc/core/linux_core.go
+++ b/pkg/proc/core/linux_core.go
@@ -29,15 +29,65 @@ const NT_X86_XSTATE elf.NType = 0x202 // Note type for notes containing X86 XSAV
 // NT_AUXV is the note type for notes containing a copy of the Auxv array
 const NT_AUXV elf.NType = 0x6
 
+// NT_FPREGSET is the note type for floating point registers.
+const NT_FPREGSET elf.NType = 0x2
+
+// Fetch architecture using exeELF.Machine from core file
+// Refer http://man7.org/linux/man-pages/man5/elf.5.html
+const (
+	EM_AARCH64           = 183
+	EM_X86_64            = 62
+	_ARM_FP_HEADER_START = 512
+)
+
 const elfErrorBadMagicNumber = "bad magic number"
 
-// readLinuxAMD64Core reads a core file from corePath corresponding to the executable at
+func linuxThreadsFromNotes(p *Process, notes []*Note, machineType elf.Machine) {
+	var lastThreadAMD *linuxAMD64Thread
+	var lastThreadARM *linuxARM64Thread
+	for _, note := range notes {
+		switch note.Type {
+		case elf.NT_PRSTATUS:
+			if machineType == EM_X86_64 {
+				t := note.Desc.(*LinuxPrStatusAMD64)
+				lastThreadAMD = &linuxAMD64Thread{linutil.AMD64Registers{Regs: &t.Reg}, t}
+				p.Threads[int(t.Pid)] = &Thread{lastThreadAMD, p, proc.CommonThread{}}
+				if p.currentThread == nil {
+					p.currentThread = p.Threads[int(t.Pid)]
+				}
+			} else if machineType == EM_AARCH64 {
+				t := note.Desc.(*LinuxPrStatusARM64)
+				lastThreadARM = &linuxARM64Thread{linutil.ARM64Registers{Regs: &t.Reg}, t}
+				p.Threads[int(t.Pid)] = &Thread{lastThreadARM, p, proc.CommonThread{}}
+				if p.currentThread == nil {
+					p.currentThread = p.Threads[int(t.Pid)]
+				}
+			}
+		case NT_FPREGSET:
+			if machineType == EM_AARCH64 {
+				if lastThreadARM != nil {
+					lastThreadARM.regs.Fpregs = note.Desc.(*linutil.ARM64PtraceFpRegs).Decode()
+				}
+			}
+		case NT_X86_XSTATE:
+			if machineType == EM_X86_64 {
+				if lastThreadAMD != nil {
+					lastThreadAMD.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
+				}
+			}
+		case elf.NT_PRPSINFO:
+			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
+		}
+	}
+}
+
+// readLinuxCore reads a core file from corePath corresponding to the executable at
 // exePath. For details on the Linux ELF core format, see:
 // http://www.gabriel.urdhr.fr/2015/05/29/core-file/,
 // http://uhlo.blogspot.fr/2012/05/brief-look-into-core-dumps.html,
 // elf_core_dump in http://lxr.free-electrons.com/source/fs/binfmt_elf.c,
 // and, if absolutely desperate, readelf.c from the binutils source.
-func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
+func readLinuxCore(corePath, exePath string) (*Process, error) {
 	coreFile, err := elf.Open(corePath)
 	if err != nil {
 		if _, isfmterr := err.(*elf.FormatError); isfmterr && (strings.Contains(err.Error(), elfErrorBadMagicNumber) || strings.Contains(err.Error(), " at offset 0x0: too short")) {
@@ -62,45 +112,41 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
 		return nil, fmt.Errorf("%v is not an exe file", exeELF)
 	}
 
-	notes, err := readNotes(coreFile)
+	machineType := exeELF.Machine
+	notes, err := readNotes(coreFile, machineType)
 	if err != nil {
 		return nil, err
 	}
 	memory := buildMemory(coreFile, exeELF, exe, notes)
 	entryPoint := findEntryPoint(notes)
-
 	p := &Process{
 		mem:         memory,
 		Threads:     map[int]*Thread{},
 		entryPoint:  entryPoint,
-		bi:          proc.NewBinaryInfo("linux", "amd64"),
 		breakpoints: proc.NewBreakpointMap(),
 	}
 
-	var lastThread *linuxAMD64Thread
-	for _, note := range notes {
-		switch note.Type {
-		case elf.NT_PRSTATUS:
-			t := note.Desc.(*LinuxPrStatus)
-			lastThread = &linuxAMD64Thread{linutil.AMD64Registers{Regs: &t.Reg}, t}
-			p.Threads[int(t.Pid)] = &Thread{lastThread, p, proc.CommonThread{}}
-			if p.currentThread == nil {
-				p.currentThread = p.Threads[int(t.Pid)]
-			}
-		case NT_X86_XSTATE:
-			if lastThread != nil {
-				lastThread.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
-			}
-		case elf.NT_PRPSINFO:
-			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
-		}
+	switch machineType {
+	case EM_X86_64:
+		p.bi = proc.NewBinaryInfo("linux", "amd64")
+		linuxThreadsFromNotes(p, notes, machineType)
+	case EM_AARCH64:
+		p.bi = proc.NewBinaryInfo("linux", "arm64")
+		linuxThreadsFromNotes(p, notes, machineType)
+	default:
+		return nil, fmt.Errorf("unsupported machine type")
 	}
 	return p, nil
 }
 
 type linuxAMD64Thread struct {
 	regs linutil.AMD64Registers
-	t    *LinuxPrStatus
+	t    *LinuxPrStatusAMD64
+}
+
+type linuxARM64Thread struct {
+	regs linutil.ARM64Registers
+	t    *LinuxPrStatusARM64
 }
 
 func (t *linuxAMD64Thread) registers(floatingPoint bool) (proc.Registers, error) {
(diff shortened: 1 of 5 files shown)


pkg/proc/core/linux_core_amd64.go
```diff
@@ -8,7 +8,7 @@ import (
 	"io"
 	"os"
 	"strings"
-
+
```
aarzilli: Run go fmt on changed files.

pkg/proc/core/linux_core_arm64.go
```diff
@@ -0,0 +1,350 @@
+package core
```
aarzilli: I think that it should be possible to read arm64 core dumps on non-arm64 hosts, the name of this file should be linux_arm64_core.go so that it is compiled in unconditionally.

Also linux_amd64_core.go should not be renamed to linux_core_amd64.go.

pkg/proc/core/linux_core.go
```diff
@@ -29,108 +22,217 @@ const NT_X86_XSTATE elf.NType = 0x202 // Note type for notes containing X86 XSAV
...
-	}
-	if exeELF.Type != elf.ET_EXEC && exeELF.Type != elf.ET_DYN {
-		return nil, fmt.Errorf("%v is not an exe file", exeELF)
+const (
+	_ARM_FP_HEADER_START = 512
+)
+
+// Copied from golang.org/x/sys/unix.Timeval since it's not available on all
+// systems.
+type LinuxCoreTimeval struct {
```
aarzilli: Please limit the intermixing of cosmetic changes (such as moving the declaration of structs around) with functional changes. Reviewing this PR is very hard.

ossdev07: @aarzilli will remove cosmetic changes to make it simple for review

aarzilli: Thank you.

ossdev07: > Thank you.

changes are Done now

pkg/proc/core/linux_core.go
```diff
@@ -61,46 +73,90 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
...
-	var lastThread *linuxAMD64Thread
-	for _, note := range notes {
-		switch note.Type {
-		case elf.NT_PRSTATUS:
-			t := note.Desc.(*LinuxPrStatus)
-			lastThread = &linuxAMD64Thread{linutil.AMD64Registers{Regs: &t.Reg}, t}
-			p.Threads[int(t.Pid)] = &Thread{lastThread, p, proc.CommonThread{}}
-			if p.currentThread == nil {
-				p.currentThread = p.Threads[int(t.Pid)]
+		var lastThread *linuxAMD64Thread
```
aarzilli: split this code to a different function called `linuxThreadsFromNotesAMD64(*Process, []*Note)`.

pkg/proc/core/linux_core.go
```diff
@@ -61,46 +73,90 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
 	if exeELF.Type != elf.ET_EXEC && exeELF.Type != elf.ET_DYN {
 		return nil, fmt.Errorf("%v is not an exe file", exeELF)
 	}
+	switch getMachineType := exeELF.Machine; getMachineType {
```
aarzilli: rename `getMachineType` to `machineType`.

pkg/proc/core/linux_core.go
```diff
@@ -129,8 +198,8 @@ type Note struct {
 	Desc interface{} // Decoded Desc from the
 }
 
-// readNotes reads all the notes from the notes prog in core.
-func readNotes(core *elf.File) ([]*Note, error) {
+// readNotesAMD64 reads all the notes from the notes prog in core.
+func readNotesAMD64(core *elf.File) ([]*Note, error) {
```
aarzilli: This function and `readNotesARM64` are basically the same, remove `readNotesARM64` call this one `readNotes` and add `machineType` as a parameter.

pkg/proc/core/linux_core.go
```diff
@@ -155,8 +224,8 @@ func readNotes(core *elf.File) ([]*Note, error) {
 	return notes, nil
 }
 
-// readNote reads a single note from r, decoding the descriptor if possible.
-func readNote(r io.ReadSeeker) (*Note, error) {
+// readNoteAMD64 reads a single note from r, decoding the descriptor if possible.
+func readNoteAMD64(r io.ReadSeeker) (*Note, error) {
```
aarzilli: This function and `readNoteARM64` are basically the same. Rename this to `readNote`, remove `readNoteARM64`. Add a `machineType` parameter to this. Use the `machineType` parameter to decide what to do inside the switch cases `elf.NT_PRSTATUS`, `elf.NT_X86_STATE` (return an error if machineType is not amd64) and `elf.NT_FPREGSET`.

hengwu0:
> this will have to wait for stacktraces on arm64 before being merged.

arm64 stacktraces is almost done and there are some bugfixes work left.

pkg/proc/core/linux_core.go
```diff
@@ -29,15 +29,69 @@ const NT_X86_XSTATE elf.NType = 0x202 // Note type for notes containing X86 XSAV
...
+const EM_X86_64 = 62
+
+const (
+	_ARM_FP_HEADER_START = 512
+)
+
 const elfErrorBadMagicNumber = "bad magic number"
 
-// readLinuxAMD64Core reads a core file from corePath corresponding to the executable at
+func (lastThread *linuxAMD64Thread) linuxThreadsFromNotesAMD64(p *Process, notes []*Note) (*Process, error) {
```
aarzilli: No reason to have lastThread as a receiver here, it can just be a local variable. Also, no reason to return a *Process since p is already a pointer.

pkg/proc/core/linux_core.go
```diff
@@ -29,15 +29,69 @@ const NT_X86_XSTATE elf.NType = 0x202 // Note type for notes containing X86 XSAV
...
+				lastThread.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
+			}
+		case elf.NT_PRPSINFO:
+			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
+		}
+	}
+	return p, nil
+}
+
+func (lastThread *linuxARM64Thread) linuxThreadsFromNotesARM64(p *Process, notes []*Note) (*Process, error) {
```
aarzilli: Same as linuxThreadsFromNotesAMD64

pkg/proc/core/linux_core.go
```diff
@@ -62,45 +116,53 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
...
-			}
-		case NT_X86_XSTATE:
-			if lastThread != nil {
-				lastThread.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
-			}
-		case elf.NT_PRPSINFO:
-			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
+	switch machineType {
+	case EM_X86_64:
+		p := &Process{
```
aarzilli: This code is almost the asme as the code in the switch case below. Take it out (with the exception of the assignment to the member variable bi) then in the switch case do: `p.bi = proc.NewBinaryInfo(...)`.

pkg/proc/core/linux_core.go
```diff
@@ -183,9 +258,16 @@ func readNote(r io.ReadSeeker) (*Note, error) {
 	descReader := bytes.NewReader(desc)
 	switch note.Type {
 	case elf.NT_PRSTATUS:
-		note.Desc = &LinuxPrStatus{}
-		if err := binary.Read(descReader, binary.LittleEndian, note.Desc); err != nil {
-			return nil, fmt.Errorf("reading NT_PRSTATUS: %v", err)
+		if machineType == EM_X86_64 {
+			note.Desc = &LinuxPrStatusAMD64{}
+			if err := binary.Read(descReader, binary.LittleEndian, note.Desc); err != nil {
```
aarzilli: This if is the same in both branches, move it out. Leave only the assignment to `note.Desc` inside the two branches.

pkg/proc/core/linux_core.go
```diff
@@ -210,13 +292,26 @@ func readNote(r io.ReadSeeker) (*Note, error) {
...
 	case NT_X86_XSTATE:
-		var fpregs linutil.AMD64Xstate
-		if err := linutil.AMD64XstateRead(desc, true, &fpregs); err != nil {
+		if machineType == EM_X86_64 {
+			var fpregs linutil.AMD64Xstate
+			if err := linutil.AMD64XstateRead(desc, true, &fpregs); err != nil {
+				return nil, err
+			}
+			note.Desc = &fpregs
+		} else {
```
aarzilli: The else branch needs to be removed.

ossdev07:
@aarzilli ,kindly confirm, is stacktraces implemented?
So that, this PR can be merged.

aarzilli:
> kindly confirm, is stacktraces implemented?

No.

pkg/proc/linutil/regs_arm64_arch.go
```diff
@@ -130,7 +130,23 @@ func (r *ARM64Registers) Copy() proc.Registers {
...
+		regs = proc.AppendFPReg(regs, fmt.Sprintf("V%d", i/16), fpregs.vregs[i:i+16])
+	}
+	return
+}
+
+func (fpregs *ARM64PtraceFpRegs) Byte() []byte {
+	return fpregs.vregs[:]
+}
+
 func Decode(fpregs []byte) (regs []proc.Register) {
```
hengwu0: Why left two `Decode()`?

ossdev07: We cannot use/modify already defined Decode API 

> func Decode(fpregs []byte) (regs []proc.Register) 

as we need to use `ARM64PtraceFpRegs` here so I created new

> func (fpregs *ARM64PtraceFpRegs) Decode() (regs []proc.Register)

hengwu0:
I mean `linutil.Decode()` should be removed since there is new `ARM64PtraceFpRegs.Decode()`. They are duplicated codes.

hengwu0:
LGTM

pkg/proc/core/linux_core.go
```diff
@@ -62,45 +113,39 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
...
-			if lastThread != nil {
-				lastThread.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
-			}
-		case elf.NT_PRPSINFO:
-			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
-		}
+	switch machineType {
+	case EM_X86_64:
+		p.bi = proc.NewBinaryInfo("linux", "amd64")
+		linuxThreadsFromNotes(p, notes, machineType)
```
derekparker: This and the same call below can be moved out of the switch statement.

derekparker: This and the same call below can be moved out of the switch statement.

pkg/proc/core/linux_core.go
```diff
@@ -29,15 +29,66 @@ const NT_X86_XSTATE elf.NType = 0x202 // Note type for notes containing X86 XSAV
...
 
+// NT_FPREGSET is the note type for floating point registers.
+const NT_FPREGSET elf.NType = 0x2
+
+// Fetch architecture using exeELF.Machine from core file
+// Refer http://man7.org/linux/man-pages/man5/elf.5.html
+const EM_AARCH64 = 183
+const EM_X86_64 = 62
+
+const (
```
derekparker: Let's fold all consts into this `const (...)` declaration.

pkg/proc/core/linux_core.go
```diff
@@ -62,45 +113,39 @@ func readLinuxAMD64Core(corePath, exePath string) (*Process, error) {
...
-				p.currentThread = p.Threads[int(t.Pid)]
-			}
-		case NT_X86_XSTATE:
-			if lastThread != nil {
-				lastThread.regs.Fpregs = note.Desc.(*linutil.AMD64Xstate).Decode()
-			}
-		case elf.NT_PRPSINFO:
-			p.pid = int(note.Desc.(*LinuxPrPsInfo).Pid)
-		}
+	switch machineType {
```
derekparker: Since we're checking machine type here, we should return an error if we get an unsupported machine type.

pkg/proc/core/linux_core.go
```diff
@@ -302,8 +375,8 @@ type LinuxPrPsInfo struct {
 	Args                 [80]uint8
 }
 
-// LinuxPrStatus is a copy of the prstatus kernel struct.
-type LinuxPrStatus struct {
+// LinuxPrStatusAMD64 is a copy of the prstatus kernel struct.
+type LinuxPrStatusAMD64 struct {
```
derekparker: It's a shame we have to dup these structs just because of the Reg value, tried to think of a way to prevent that using a custom binary unmarshaler, but I don't think it'll work. Oh well.

ossdev07: @derekparker is this fine or you want some changes ? it's not clear

hengwu0: > It's a shame we have to dup these structs just because of the Reg value, tried to think of a way to prevent that using a custom binary unmarshaler, but I don't think it'll work. Oh well.

Yes, how about move `proc/registers.go` into `proc/linutil/`, and use interface instead in packages. (not in this patch, code refactor next time)

derekparker: > @derekparker is this fine or you want some changes ? it's not clear

Sorry, no changes requested right now. I was just thinking out loud, apologies for not being clearer.

ossdev07:
@derekparker it's long pending PR so can we do all refactoring in a new PR and let this PR go in master branch

derekparker:
@ossdev07 there are merge conflicts anyways. I'd say resolve the conflicts and the feedback at the same time and push your changes. After that I will merge quickly. Sorry for the delay.

ossdev07:
> @ossdev07 there are merge conflicts anyways. I'd say resolve the conflicts and the feedback at the same time and push your changes. After that I will merge quickly. Sorry for the delay.

@derekparker Review comments are incorporated. please have a look

PR #1777, opened 2019-11-28
proc: build disassemblers unconditionally

aarzilli:
```
proc: build disassemblers unconditionally

Remove build tags from disassembler code, move architecture specific
functionality inside proc.Arch.
This is necessary because Delve should be able to debug corefiles
cross-platform.

vendor: Fix inconsistent vendoring

The ARM64 left the vendor directory in an inconsistent state, go1.13
doesn't care so we didn't catch it but go1.14 does.

```


hengwu0:
The filename of this project is like __goversion__, __logflags__, __bininfo.go__. How about __disassarm64.go__?

aarzilli:
Just keeping consistency with amd64_arch and the stuff in pkg/proc/core.

hengwu0:
I think you should change `maxInstructionLength` with `Arch.MaxInstructionLength()`here:  https://github.com/go-delve/delve/blob/23d080abe31b843d0b2820d477ce6d29e8cb3d15/pkg/proc/proc_test.go#L2483

aarzilli:
Done.

hengwu0:
> Done.

LGTM

pkg/proc/amd64_arch.go
```diff
@@ -47,10 +47,9 @@ func (a *AMD64) PtrSize() int {
 	return 8
 }
 
-// MinInstructionLength returns the min lenth
-// of the instruction
-func (a *AMD64) MinInstructionLength() int {
-	return 1
+// MaxInstructionLength returns the maximum lenght of an instruction.
```
derekparker: nit: s/lenght/length/

Issue #1778, opened 2019-11-28
disass on arm64 can't show GlobalVars


hengwu0:
> 1. What version of Delve are you using (`dlv version`)?

top version

> 2. What version of Go are you using? (`go version`)?

go version go1.12.12 linux/arm64

> 3. What operating system and processor architecture are you using?

arm64/linux

```sh
$ ./dlv exec _fixtures/teststepconcurrent
(dlv) b main.main
Breakpoint 1 set at 0x95950 for main.main() ./_fixtures/teststepconcurrent.go:32
(dlv) c
> main.main() ./_fixtures/teststepconcurrent.go:32 (hits goroutine(1):1 total:1) (PC: 0x95950)
    27:                         wg.Done()
    28:                 }()
    29:         }
    30: }
    31:
=>  32: func main() {
    33:         x := v
    34:         y := x * x
    35:         var z int
    36:         Threads(Foo)
    37:         for i := 0; i < 100; i++ {
(dlv) disass
TEXT main.main(SB) /home/wuheng/github_git/src/github.com/go-delve/delve/_fixtures/teststepconcurrent.go
        teststepconcurrent.go:32        0x95940         810b40f9        MOVD 16(R28), R1
        teststepconcurrent.go:32        0x95944         e2c300d1        SUB $48, RSP, R2
        teststepconcurrent.go:32        0x95948         5f0001eb        CMP R1, R2
        teststepconcurrent.go:32        0x9594c         c9080054        BLS 70(PC)
=>      teststepconcurrent.go:32        0x95950*        fe0f15f8        MOVD.W R30, -176(RSP)
        teststepconcurrent.go:32        0x95954         fd831ff8        MOVD R29, -8(RSP)
        teststepconcurrent.go:32        0x95958         fd2300d1        SUB $8, RSP, R29
        teststepconcurrent.go:33        0x9595c         5b0600f0        ADRP 831488(PC), R27
        teststepconcurrent.go:33        0x95960         7b430391        ADD $208, R27, R27
        teststepconcurrent.go:33        0x95964         600340f9        MOVD (R27), R0
        teststepconcurrent.go:33        0x95968         e03300f9        MOVD R0, 96(RSP)
        teststepconcurrent.go:34        0x9596c         007c009b        MUL R0, R0, R0
        teststepconcurrent.go:34        0x95970         e02f00f9        MOVD R0, 88(RSP)
[...]

PR #1779, opened 2019-11-28
test: add reason for skipping TestDisassembleGlobalVars on arm64 (1778)

hengwu0:
It seems that there is no GlobalVars be showed in disass on arm64. It's arm64/linux objdump feature.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3595,7 +3595,7 @@ func TestIssue1145(t *testing.T) {
 
 func TestDisassembleGlobalVars(t *testing.T) {
 	if runtime.GOARCH == "arm64" {
-		t.Skip("test is not valid on ARM64")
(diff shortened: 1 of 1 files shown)


pkg/proc/proc_test.go
```diff
@@ -3595,7 +3595,7 @@ func TestIssue1145(t *testing.T) {
 
 func TestDisassembleGlobalVars(t *testing.T) {
 	if runtime.GOARCH == "arm64" {
-		t.Skip("test is not valid on ARM64")
+		t.Skip("there is always no GlobalVars in disass on ARM64")
```
aarzilli: say: "On ARM64 symLookup can't look up variables due to how they are loaded, see issue #1778"

hengwu0: Fixed.

Issue #1770, opened 2019-11-26
Next steps of the remaining work of debugging support on ARM64


hengwu0:
1、There is bug in orig-Detach function, it needs to be investigated to fix `TestExit` and `TestExitAfterContinue`.
2、`TestDisassembleGlobalVars` needs to be investigated. The disasm has been implemented and all other disasm tests passed, except this one. It seems can't find main.v in range of mainfn.Entry and mainfn.End on arm64.
3、To implemente stacktrace with separate amd64-arch code again.
4、To implemente Fp registers and Fp instructions

#1733, #118

hengwu0:
> 1、There is bug in orig-Detach function, it needs to be investigated to fix `TestExit` and `TestExitAfterContinue`.

fixed by https://github.com/go-delve/delve/commit/eb6d01a63ef371edc8039965b48b84e74a35576d https://github.com/go-delve/delve/pull/1733/commits/b951551e429bc7c8c12ec8e696b571d7f4d3e471.

hengwu0:
> 2、`TestDisassembleGlobalVars` needs to be investigated. The disasm has been implemented and all other disasm tests passed, except this one. It seems can't find main.v in range of mainfn.Entry and mainfn.End on arm64.

traced in https://github.com/go-delve/delve/issues/1778

hengwu0:
> 3、To implemente stacktrace with separate amd64-arch code again.

traced in #1780

hengwu0:
> 4、To implemente Fp registers and Fp instructions

The Fp registers and Fp instructions can be got correctly in delve already, but it seems that the floating-point registers are displayed in wrong format and hardly read when debugging. which traced in #2198

hengwu0:
I closed this issue, since all lists are done.

PR #1780, opened 2019-11-29
proc: implemente stacktrace of arm64

hengwu0:
steps:
1、separate amd64 code about stacktrace, so we can add arm64 stacktrace code.
2、add arm64 stacktrace code.
3、testcases and bugfix.

#1770


hengwu0:
PTAL

aarzilli:
There are vast amounts of code duplicated between arm64_stack and amd64_stack, that shouldn't happen. The idea here is that stack.go should be mostly architecture independent, it's an interpreter for programs written for a Virtual Machine defined by DWARFv4. Anything that needs to be architecture dependent should become a function call into some method of Arch.

Also you seem to have based your arm64_stack.go from an old version of stack.go so the duplicated code has a couple of old bugs that were fixed in stack.go.

For now I'd worry about making the normal stacktraces work, I wouldn't worry about cgo or the switchStack function. Also I'm pretty sure you have to change the code that reads goroutines to read the 'lr' field of g.sched, I don't think it's going to work otherwise.

hengwu0:
> There are vast amounts of code duplicated between arm64_stack and amd64_stack, that shouldn't happen. The idea here is that stack.go should be mostly architecture independent, it's an interpreter for programs written for a Virtual Machine defined by DWARFv4. Anything that needs to be architecture dependent should become a function call into some method of Arch.

I know they both use DWARFv4, but the regs and disasm are different, especially vast differences of _asm_arm64.s_ and _asm_amd64.s_ file. That's the reason it need different stackIterator.

> Also you seem to have based your arm64_stack.go from an old version of stack.go so the duplicated code has a couple of old bugs that were fixed in stack.go.

Yes, it should be updated.

>I'm pretty sure you have to change the code that reads goroutines to read the 'lr' field of g.sched. 

Before child function, The __lr__ in first frame is `X30 register`. It will be transferred into `StackFrame` after then. Is there need to read g.sched's lr?

aarzilli:
> I know they both use DWARFv4, but the regs and disasm are different, 

Most (all?) of the differences should be captured by the contents of debug_frame and FixFrameUnwindContext.

> especially vast differences of asm_arm64.s and asm_amd64.s file.

Very true, for that a method of Arch should be added and called.

aarzilli:
> Before child function, The lr in first frame is X30 register. It will be transferred into StackFrame after then. Is there need to read g.sched's lr?

I think so because that code is also supposed to work on parked goroutines which won't have registers besides the one read from the g.sched struct.

hengwu0:
> > I know they both use DWARFv4, but the regs and disasm are different,
> 
> Most (all?) of the differences should be captured by the contents of debug_frame and FixFrameUnwindContext.
> 
> > especially vast differences of asm_arm64.s and asm_amd64.s file.
> 
> Very true, for that a method of Arch should be added and called.

ok, BUT i don't want to mass amd64 with arm64 stacktrace together for now, since the amd64 is excellent while arm64 is steadying. Can this work be to do later?

hengwu0:
> > Before child function, The lr in first frame is X30 register. It will be transferred into StackFrame after then. Is there need to read g.sched's lr?
> 
> I think so because that code is also supposed to work on parked goroutines which won't have registers besides the one read from the g.sched struct.

Agreed! thanks.

aarzilli:
> > > I know they both use DWARFv4, but the regs and disasm are different,
> > 
> > 
> > Most (all?) of the differences should be captured by the contents of debug_frame and FixFrameUnwindContext.
> > > especially vast differences of asm_arm64.s and asm_amd64.s file.
> > 
> > 
> > Very true, for that a method of Arch should be added and called.
> 
> ok, BUT i don't want to mass amd64 with arm64 stacktrace together for now, since the amd64 is excellent while arm64 is steadying.

Yes, this is why we have a test set and do reviews.

> Can this work be to do later?

As far as I am concerned, no. I expect the changes to be small besides switchStack and FixFrameUnwindContext.

hengwu0:
> > > > I know they both use DWARFv4, but the regs and disasm are different,
> > > 
> > > 
> > > Most (all?) of the differences should be captured by the contents of debug_frame and FixFrameUnwindContext.
> > > > especially vast differences of asm_arm64.s and asm_amd64.s file.
> > > 
> > > 
> > > Very true, for that a method of Arch should be added and called.
> > 
> > 
> > ok, BUT i don't want to mass amd64 with arm64 stacktrace together for now, since the amd64 is excellent while arm64 is steadying.
> 
> Yes, this is why we have a test set and do reviews.
> 
> > Can this work be to do later?
> 
> As far as I am concerned, no. I expect the changes to be small besides switchStack and FixFrameUnwindContext.

Alright, can i push the patch append? Or must be repushed?

aarzilli:
> Alright, can i push the patch append? Or must be repushed?

Whatever you prefer.

pkg/proc/proc_test.go
```diff
@@ -4103,7 +4049,7 @@ func TestNextUnknownInstr(t *testing.T) {
 
 func TestReadDeferArgs(t *testing.T) {
 	if runtime.GOARCH == "arm64" {
-		t.Skip("arm64 does not support Stacktrace for now")
+		t.Skip("arm64 does not support ReadDeferArgs for now")
```
aarzilli: Can you do an experiment for me?
In stack.go there's a function called EvalScope with receiver *Defer, inside this function there's a line that says:

```
scope.Regs.Regs[scope.Regs.SPRegNum].Uint64Val = uint64(scope.Regs.CFA - int64(bi.Arch.PtrSize()))
```

can you change it to:

```
switch bi.Arch.(type) {
case *AMD64:
     scope.Regs.Regs[scope.Regs.SPRegNum].Uint64Val = uint64(scope.Regs.CFA - int64(bi.Arch.PtrSize()))
case *ARM64:
     scope.Regs.Regs[scope.Regs.SPRegNum].Uint64Val = uint64(scope.Regs.CFA)
}
```

And then see if the test passes.
I'm not sure that's the problem but it seems like it would be the most likely cause.

pkg/proc/proc_test.go
```diff
@@ -3262,9 +3229,6 @@ func frameInFile(frame proc.Stackframe, file string) bool {
 }
 
 func TestCgoStacktrace(t *testing.T) {
-	if runtime.GOARCH != "amd64" {
```
aarzilli: This test is failing, if you aren't going to fix it (it's probably going to be fairly complicated) then you should keep it disabled.

hengwu0: To my honest that my code is on the base of go1.10, which works well done on CgoStacktrace... And i confirmed than golang changed some asm-code in `asmcgocall` and it's call-chain... That's why it needs some some more time on cgo, and it may hard to compatible low go version(which we using now).

pkg/proc/stack.go
```diff
@@ -589,8 +590,7 @@ func (it *stackIterator) executeFrameRegRule(regnum uint64, rule frame.DWRule, c
 	case frame.RuleUndefined:
 		return nil, nil
 	case frame.RuleSameVal:
-		reg := *it.regs.Reg(regnum)
-		return &reg, nil
+		return it.regs.Reg(regnum), nil
```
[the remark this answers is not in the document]
hengwu0: the `it.regs.Reg(regnum)` may `nil` sometimes on arm64.

pkg/proc/proc_test.go
hengwu0: emm, no any change(even the err value):
```sh
=== RUN   TestReadDeferArgs
--- FAIL: TestReadDeferArgs (9.68s)
    proc_test.go:4095: value of argument 'a' at frame 1, deferred call 1: 0 (expected 42)
    proc_test.go:4099: value of argument 'b' at frame 1, deferred call 1: 0 (expected 61)
    proc_test.go:4095: value of argument 'a' at frame 2, deferred call 2: 72 (expected 1)
    proc_test.go:4099: value of argument 'b' at frame 2, deferred call 2: 64 (expected -1)
FAIL
FAIL    github.com/go-delve/delve/pkg/proc      10.516s
FAIL
```

hengwu0:
> I think that the switch cases for `"runtime.asmcgocall"` and for `"runtime.cgocallback_gofunc"` should just return false for arm64. Which means they should be moved to some method of Arch. This will probably cause some other tests to fail, but that's fine because they were passing by accident.

I reseted the patch since it can't parse cgo of go1.13, and i think the combine work should do after all stacktrace work done. It may need another couple of days.(P.S. It should switch sp before switchStack on arm64, which opposited on amd64. So it may need some artifice to combine these two arch-code…)

pkg/proc/stack.go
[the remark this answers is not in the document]
aarzilli: Then do:

```
if it.regs.Reg(regnum) == nil {
    return nil
}
<what was there before>
```

pkg/proc/proc_test.go
aarzilli: Ok, then I don't know, but the problem should be somewhere in that function.

pkg/proc/proc_test.go
aarzilli: Ok.

hengwu0:
Sorry for waiting, it seems there has bugs in both amd64-code about cgo stacktrace(https://github.com/go-delve/delve/issues/1796). The work becomes harder…… @aarzilli

hengwu0:
Stacktrace on arm64 seems finished, but how to make it architecture independent? do you have some advice? @aarzilli

aarzilli:
I'm not sure what happened here, the last time I reviewed this PR, on december 10th, it was significantly better than it is now. What I see now has essentially all the problems it had when I reviewed it on december 2nd, namely: there are vast amounts of duplicated code between arm64_stack.go and amd64_stack, and arm64_stack.go is based on an old version of stack.go with known problems.

The version I saw on december 10th was closer to being correct, the differences between ARM64 and AMD64 should be concentrated in the switchStack function and in the FixFrameUnwindContext method of Arch.

hengwu0:
> The version I saw on december 10th was closer to being correct, the differences between ARM64 and AMD64 should be concentrated in the switchStack function and in the FixFrameUnwindContext method of Arch.

That version is no cgo supported, now cgo is completed. And There have others differents include advanceRegs function. And it seems hardly to move into *_arch.go file.

hengwu0:
Done.

pkg/proc/proc.go
```diff
@@ -226,6 +226,9 @@ func Continue(dbp Process) error {
 				return conditionErrors(threads)
 			case g == nil || dbp.Common().fncallForG[g.ID] == nil:
 				// a hardcoded breakpoint somewhere else in the code (probably cgo)
+				if !curthread.Arch().BreakInstrMovesPC() {
+					curthread.SetPC(loc.PC + uint64(curthread.Arch().BreakpointSize()))
```
aarzilli: IIRC this case also happens when the target process is stopped manually, in that case the instruction at PC won't be a hardcoded breakpoint but some random instruction so we can't blindly advance by BreakpointSize. This code should read the memory at loc.PC with dbp.CurrentThread().ReadMemory and compare it to dbp.BinInfo().Arch.BreakpointInstruction()

pkg/proc/stack.go
```diff
@@ -252,8 +252,15 @@ func (it *stackIterator) Next() bool {
 	}
 
 	if it.opts&StacktraceSimple == 0 {
-		if it.switchStack() {
-			return true
+		switch it.bi.Arch.(type) {
```
aarzilli: Add a `SwitchStack` method to Arch and make amd64switchStack and arm64switchStack methods of ARM64 and AMD64.

pkg/proc/stack.go
```diff
@@ -387,12 +397,120 @@ func (it *stackIterator) switchStack() bool {
...
+	}
+}
+
 func (it *stackIterator) switchToGoroutineStack() {
 	it.systemstack = false
 	it.top = false
 	it.pc = it.g.PC
 	it.regs.Reg(it.regs.SPRegNum).Uint64Val = it.g.SP
 	it.regs.Reg(it.regs.BPRegNum).Uint64Val = it.g.BP
+	if _, ok := it.bi.Arch.(*ARM64); ok {
```
aarzilli: Do `if it.regs.Reg(it.regs.LRRegNum) != nil` instead

pkg/proc/stack.go
```diff
@@ -387,12 +397,120 @@ func (it *stackIterator) switchStack() bool {
 	}
 }
 
+func (it *stackIterator) switchSp() bool {
```
aarzilli: The return value isn't used so it should be removed. Also the name should have a hint that this is arm64 specific (arm64SwitchSp, for example).

pkg/proc/stack.go
```diff
@@ -543,6 +661,10 @@ func (it *stackIterator) advanceRegs() (callFrameRegs op.DwarfRegisters, ret uin
 		framectx = it.bi.Arch.FixFrameUnwindContext(fde.EstablishFrame(it.pc), it.pc, it.bi)
 	}
 
+	if _, ok := it.bi.Arch.(*ARM64); ok {
```
aarzilli: I think this if and the call to switchSp can be moved outside of `advanceRegs` and directly into `Next`.

scripts/make.go
```diff
@@ -255,7 +255,7 @@ func testFlags() []string {
 	if err != nil {
 		log.Fatal(err)
 	}
-	testFlags := []string{"-count", "1", "-p", "1"}
+	testFlags := []string{"-count", "1", "-p", "1", "-timeout", "0"}
```
aarzilli: Having a timeout set for the tests is very useful and 10 minutes is a reasonable default, if you need infinite timeouts add a new flag to make.go to enable those manually.

pkg/proc/proc.go
hengwu0: AYS? The comment here said "a hardcoded breakpoint somewhere else".

pkg/proc/stack.go
hengwu0: The `switchSp` should be a part of `advanceRegs`, because it must get right SP value before get other regs. And the `advanceRegs()` called in two places of proc.go, that means `switchSp()` will appear twice too, that's duplicated i think.

pkg/proc/proc.go
hengwu0: modified.

aarzilli: > AYS? The comment here said "a hardcoded breakpoint somewhere else".

Yes, the comment is wrong (or, rather, incomplete).

pkg/proc/stack.go
aarzilli: The other place that calls advanceRegs is amd64's switchStack for which it isn't relevant.

pkg/proc/stack.go
hengwu0: The `it.regs.LRRegNum`'s value is 0 on amd64 for default.

scripts/make.go
hengwu0: modified.

pkg/proc/stack.go
hengwu0: modified.

pkg/proc/stack.go
hengwu0: The return value is left for only change SP and ignore other regs. Need to remove?

pkg/proc/stack.go
hengwu0: I reorganize `switchSp` to `Arch interface`, is it ok?

pkg/proc/arm64_arch.go
```diff
@@ -258,6 +376,7 @@ func (a *ARM64) RegistersToDwarfRegisters(staticBase uint64, regs Registers) op.
 	dregs[arm64DwarfIPRegNum] = op.DwarfRegisterFromUint64(regs.PC())
 	dregs[arm64DwarfSPRegNum] = op.DwarfRegisterFromUint64(regs.SP())
 	dregs[arm64DwarfBPRegNum] = op.DwarfRegisterFromUint64(regs.BP())
+	dregs[arm64DwarfLRRegNum] = op.DwarfRegisterFromUint64(regs.LR())
```
derekparker: Instead of adding `LR()` to the overall Registers interface, can we just type cast once within this function and access LR directly? I'd like to avoid adding all the methods that simply return 0 if possible, and unless I missed something that method is only used here where we would always expect the ARM register struct.

pkg/proc/stack.go
```diff
@@ -542,6 +425,7 @@ func (it *stackIterator) advanceRegs() (callFrameRegs op.DwarfRegisters, ret uin
 	} else {
 		framectx = it.bi.Arch.FixFrameUnwindContext(fde.EstablishFrame(it.pc), it.pc, it.bi)
 	}
+	it.bi.Arch.SwitchSp(it)
```
derekparker: Why do we only want to switch stack pointer here? Is there a reason to special case this and not just use `SwitchStack`?

pkg/proc/proc.go
```diff
@@ -225,7 +226,15 @@ func Continue(dbp Process) error {
 				}
 				return conditionErrors(threads)
 			case g == nil || dbp.Common().fncallForG[g.ID] == nil:
-				// a hardcoded breakpoint somewhere else in the code (probably cgo)
+				// a hardcoded breakpoint somewhere else in the code (probably cgo), or manual stop in cgo
+				if !curthread.Arch().BreakInstrMovesPC() {
```
derekparker: Let's be consistent with how we access Arch as it's likely going to change in the future.

For now just store it as `arch := dbp.BinaryInfo().Arch` and use that instead of always calling `curthread.Arch()` or `dbp.BinInfo().Arch`.

pkg/proc/stack.go
hengwu0: The `SwitchSp` is a part of `advanceRegs`, and it must get right SP value before get other regs. But the `SwitchStack` is called after all regs got in `Next`. That's the reason we need `SwitchSp` there.

pkg/proc/amd64_arch.go
```diff
@@ -160,6 +161,129 @@ func (a *AMD64) FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *
 	return fctxt
 }
 
+// cgocallSPOffsetSaveSlot is the offset from systemstack.SP where
+// (goroutine.SP - StackHi) is saved in runtime.asmcgocall after the stack
+// switch happens.
+const amd64cgocallSPOffsetSaveSlot = 0x28
+
+// switchStack will use the current frame to determine if it's time to
```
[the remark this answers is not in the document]
hengwu0: Fixed.

pkg/proc/amd64_arch.go
```diff
@@ -160,6 +161,129 @@ func (a *AMD64) FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *
...
+	}
+	switch it.frame.Current.Fn.Name {
+	case "runtime.asmcgocall":
+		if it.top || !it.systemstack {
+			return false
+		}
+
+		// This function is called by a goroutine to execute a C function and
+		// switches from the goroutine stack to the system stack.
+		// Since we are unwinding the stack from callee to caller we have  switch
```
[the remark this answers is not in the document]
hengwu0: Fixed.

pkg/proc/arm64_arch.go
hengwu0: Removed.

pkg/proc/stack.go
aarzilli: Yes, that's good, but it should still be called in Next instead of advanceRegs.

aarzilli:
I think that SwitchStack and SwitchSp are basically the same thing and should be combined in a single function, see https://github.com/go-delve/delve/pull/1832/commits/d37a357d97064b50cfb6e02d1c2434469f3ae52e

hengwu0:
> LGTM (with squash&rebase before merging)

What need i do?

aarzilli:
> What need i do?

Nothing. Just a reminder for us to squash the commits before merging them.

hengwu0:
> I think that SwitchStack and SwitchSp are basically the same thing and should be combined in a single function, see [d37a357](https://github.com/go-delve/delve/commit/d37a357d97064b50cfb6e02d1c2434469f3ae52e)

Thanks for your assistance very much.

hengwu0:
Can merge? @derekparker

hengwu0:
Thanks @derekparker @aarzilli . And Happy Chinese New Year to All!

PR #1784, opened 2019-12-03
proc,service: construct import path to directory path map

aarzilli:
```
proc,service: construct import path to directory path map

Adds an API call that returns a map from import paths to directory
paths that can be used by IDEs to map source file paths between
different development environments.

```

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -351,7 +351,7 @@ func (bi *BinaryInfo) LineToPC(filename string, lineno int) (pcs []uint64, err e
 	fileFound := false
 	var pc uint64
 	for _, cu := range bi.compileUnits {
-		if cu.lineInfo.Lookup[filename] == nil {
+		if cu.lineInfo != nil && cu.lineInfo.Lookup[filename] == nil {
 			continue
 		}
 		fileFound = true
@@ -411,7 +411,7 @@ func (bi *BinaryInfo) AllPCsForFileLines(filename string, linenos []int) map[int
 		r[line] = make([]uint64, 0, 1)
 	}
 	for _, cu := range bi.compileUnits {
-		if cu.lineInfo.Lookup[filename] != nil {
+		if cu.lineInfo != nil && cu.lineInfo.Lookup[filename] != nil {
 			cu.lineInfo.AllPCsForFileLines(filename, r)
 		}
 	}
@@ -1317,8 +1317,8 @@ func (bi *BinaryInfo) loadDebugInfoMaps(image *Image, debugLineBytes []byte, wg 
 			if len(cu.ranges) >= 1 {
 				cu.lowPC = cu.ranges[0][0]
 			}
-			lineInfoOffset, _ := entry.Val(dwarf.AttrStmtList).(int64)
-			if lineInfoOffset >= 0 && lineInfoOffset < int64(len(debugLineBytes)) {
+			lineInfoOffset, hasLineInfo := entry.Val(dwarf.AttrStmtList).(int64)
+			if hasLineInfo && lineInfoOffset >= 0 && lineInfoOffset < int64(len(debugLineBytes)) {
 				var logfn func(string, ...interface{})
 				if logflags.DebugLineErrors() {
 					logger := logrus.New().WithFields(logrus.Fields{"layer": "dwarf-line"})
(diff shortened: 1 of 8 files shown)


aarzilli:
cc @nd

pkg/proc/bininfo.go
```diff
@@ -1756,3 +1756,36 @@ func (bi *BinaryInfo) symLookup(addr uint64) (string, uint64) {
...
+		ip := strings.Replace(cu.name, "\\", "/", -1)
+		if _, ok := m[ip]; ok {
+			continue
+		}
+		path := cu.lineInfo.FirstFile()
+		if ext := filepath.Ext(path); ext != ".go" && ext != ".s" {
+			continue
+		}
+		dp := filepath.Dir(path)
+		m[ip] = dp
```
nd: Use of map here implies one to one correspondence between import path and directory path. If I understand correctly this might not be the case for bazel builds: https://github.com/bazelbuild/intellij/blob/master/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java#L78.

nd:
Thank you, the API looks convenient to use in IDE.

pkg/proc/bininfo.go
aarzilli: That's annoying, I don't think there's a way to do this for bazel builds. We could also return a list of all files that partecipated on the build of each package, then the IDE could use the directory guess in modules mode and the file list in bazel mode.

nd: I think any of the following will be fine for IDE: a list of files per import path, or a list of directories per import path, or the current solution but changed to allow several pairs with the same import path but different directories.

aarzilli:
Second attempt

nd:
I think that will work for IDE. I guess we will ignore `PackageBuildInfo.DirectoryPath` because in bazel it seems like the order of files in BUILD.bazel affects the order in dwarf. I've created a small go bazel project: https://github.com/nd/gobazel. Dwarfdump shows: 

```
COMPILE_UNIT<header overall offset = 0x00041f43>:
< 0><0x0000000b>  DW_TAG_compile_unit
                    DW_AT_name                  github.com/my/pkg
                    DW_AT_language              DW_LANG_Go
                    DW_AT_stmt_list             0x00018606
                    DW_AT_low_pc                0x004525b0
                    DW_AT_ranges                0x00000060
                ranges: 2 at .zdebug_ranges offset 96 (0x00000060) (32 bytes)
                        [ 0] range entry    0x00000000 0x000000b6
                        [ 1] range end      0x00000000 0x00000000
                    DW_AT_comp_dir              .
                    DW_AT_producer              Go cmd/compile go1.13.5
                    <Unknown AT value 0x2905>   pkg

LOCAL_SYMBOLS:
< 1><0x0000004d>    DW_TAG_subprogram
                      DW_AT_name                  github.com/my/pkg.GGGG
                      DW_AT_low_pc                0x004525b0
                      DW_AT_high_pc               0x00452601
                      DW_AT_frame_base            DW_OP_call_frame_cfa
                      DW_AT_decl_file             0x00000001 ./subdir/bar.go
                      DW_AT_external              yes(1)
< 1><0x0000007d>    DW_TAG_subprogram
                      DW_AT_name                  github.com/my/pkg.FFFF
                      DW_AT_low_pc                0x00452610
                      DW_AT_high_pc               0x00452666
                      DW_AT_frame_base            DW_OP_call_frame_cfa
                      DW_AT_decl_file             0x00000002 ./foo.go
                      DW_AT_external              yes(1)
```

so if I understand correctly delve will return "./subdir" as a `DirectoryPath` for "github.com/my/pkg". This is fine for us - we can ignore it and use only files, but might be confusing for others.

aarzilli:
> I guess we will ignore PackageBuildInfo.DirectoryPath because in bazel it seems like the order of files in BUILD.bazel affects the order in dwarf

Yes, you can't use DirectoryPath for bazel builds because there is no correct guess for bazel.
Note that if inlining is enabled each file can appear in multiple packages so just looking at the file list won't work either (if you care about debugging optimized programs).

Of course there is no way to solve this problem for programs built with bazel and optimizations turned on, after all C debuggers don't have this feature for a reason.

aarzilli:
cc @derekparker PTAL

derekparker:
@aarzilli just needs rebase, then can merge.

PR #1786, opened 2019-12-04
travis: exclude builds of Go other than most recent from ARM64 builds

aarzilli:
```
travis: exclude builds of Go other than most recent from ARM64 builds

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,12 @@ matrix:
   exclude:
     - os: osx
       arch: arm64
+    - os: linux
+      go: 1.11.x
+      arch: arm64
+    - os: linux
+      go: 1.12.x
+      arch: arm64
 
 before_install:
   - export GOFLAGS=-mod=vendor


aarzilli:
Travis-CI take forever now that we are building all four supported versions of Go on ARM64 I'd like to get some of the speed back and I don't think we actually need to test every possible combination since different version of Go are more likely to affect the symbolic layer and most of the differences between ARM64 and AMD64 are in the backend and are likely to be caught by the test on AMD64.

hengwu0:
emm... it is not good for arm64-delve...

aarzilli:
@hengwu0 why? You think just testing linux/arm/go1.13 is not enough?

hengwu0:
> @hengwu0 why? You think just testing linux/arm/go1.13 is not enough?

some tests failed on specific go version like: https://travis-ci.org/go-delve/delve/builds/620146852?utm_source=github_status&utm_medium=notification

PR #1804, opened 2019-12-30
arm64 core-file reader (this is needed by core backtrace)

MichaelPenk:
The "core.go" file was modified to allow arm64 core-files to be opened.
This was needed to get a simple backtrace ("bt" or "bt -full") to work.
If delve is executing on either an amd64 or on an arm64, the appropriate
reader will be called, which will setup the appropriate calls to enable
parsing of the core-file.

When running on an arm64, core.go will call "linux_arm64_core.go" to
setup reading the arm64 core file.  This file was derived from the
repository "https://github.com/ossdev07/delve.gt", fetched on Oct 25.

Several lines were commented out (identified by "//yyy"), as the functions
they referred to were not in main delve repository.  These functions
primarily involved accessing the floating point registers.  At some point,
these functions (or their equivalents) should be implemented.

### DIFF
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"go/ast"
 	"io"
+	"runtime"
 
 	"github.com/go-delve/delve/pkg/proc"
 )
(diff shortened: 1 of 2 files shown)


derekparker:
I'm going to close this as there hasn't been a response, and we are planning to merge #1774 instead.

PR #1807, opened 2019-12-30
proc: fix inlined stack reading for midstack inlined calls

### LINKED ISSUE #1795: Next does not skip mid-stack inlined function call
dlv version:
      Delve Debugger
      Version: 1.3.2
      Build: $Id: 569ccbd514fc47c8b4c521b142556867ec5e6917 $
GO version:
      go version go1.13.4 darwin/amd64
OS:
      MAC PRO 15 2018
      Mojave 10.14.6
Operate:
    When i use dlv's command next to debug, it step into the anotherfunction.
(dlv) next
    56:	func SetServiceRootConfig(rc string) {
=>  57:		if !regexpForServiceRootConfig.MatchString(rc) {
    58:			log.Panicf("[E] registry: invalid ServiceRootConfig string:%s", rc)
    59:			return
    60:		}
(dlv) 
   513:	func (re *Regexp) doMatch(r io.RuneReader, b []byte, s string) bool {
=> 514:		return re.doExecute(r, b, s, 0, 0, nil) != nil
   515:	}

the doMatch function is in called by MatchString function

I think this is wrong

aarzilli:
```
proc: fix inlined stack reading for midstack inlined calls

Due to a bug in the Go compiler midstack inlined calls do not report
their ranges correctly. We can't check if an address is in the range of
a DIE by simply looking at that DIE's range, we should also recursively
check the DIE's children's ranges.

Also fixes the way stacktraces of midstack inlined calls are reported
(they used to be inverted, with the deepest inlined stack frame
reported last).

Fixes #1795

```

### DIFF
--- a/_fixtures/issue1795.go
+++ b/_fixtures/issue1795.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"runtime"
+)
+
+func main() {
+	r := regexp.MustCompile("ab")
+	runtime.Breakpoint()
+	out := r.MatchString("blah")
+	fmt.Printf("%v\n", out)
+}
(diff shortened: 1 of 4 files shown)


derekparker:
@aarzilli just needs rebase.

PR #1810, opened 2020-01-04
service/debugger: support full path matching in location expressions.

### LINKED ISSUE #1808: symbol link issue

Version: 1.3.2
go version go1.12.4 darwin/amd64
macOS 10.14.6
4. What did you do?
see below
see below
6. What did you see instead?
see below

## Reproduce
```
$ cd tmp
$ ln -s ../a.go
$ realpath a.go
/Users/eph/test/a.go
$ pwd
/Users/eph/test/tmp
$ dlv debug a.go
(dlv) b /Users/eph/test/a.go:14
Command failed: Location "/Users/eph/test/a.go:14" not found
(dlv) b /Users/eph/test/tmp/a.go:14
Breakpoint 1 set at 0x10f47db for main.main() ./a.go:14
```

pdb3 (Python debugger) can handle this situation

chainhelen:
Support both relative path and symbolic link in location expressions.

Fixes #1808

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -57,6 +57,10 @@ type Config struct {
 	// DebugFileDirectories is the list of directories Delve will use
 	// in order to resolve external debug info files.
 	DebugInfoDirectories []string `yaml:"debug-info-directories"`
+
+	// If FullPathMatch is true Delve will try to resolve symbolic links when searching source files.
+	// Note: this option may cause performance degradation.
+	FullPathMatch bool `yaml:"full-path-match"`
 }
 
 // LoadConfig attempts to populate a Config object from the config.yml file.
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -125,6 +125,17 @@ type Thread struct {
 	ReturnValues []Variable
 }
 
+// FindLocCfg describes how to find location from target.
+type FindLocCfg struct {
+	// IncludeNonExecutableLines makes FindLocations return if the expression can be parsed,
+	// even if it doesn't end up matching any instruction in debug_line.
+	IncludeNonExecutableLines bool
+
+	// If FullPathMatch is true Delve will try to resolve symbolic links when searching source files.
+	// Note: this option may cause performance degradation.
+	FullPathMatch bool
+}
+
 // Location holds program location information.
 // In most cases a Location object will represent a physical location, with
 // a single PC address held in the PC field.
--- a/service/debugger/locations.go
+++ b/service/debugger/locations.go
@@ -3,6 +3,7 @@ package debugger
 import (
 	"fmt"
 	"go/constant"
+	"os"
 	"path"
 	"path/filepath"
 	"reflect"
@@ -17,7 +18,7 @@ import (
 const maxFindLocationCandidates = 5
 
 type LocationSpec interface {
-	Find(d *Debugger, scope *proc.EvalScope, locStr string, includeNonExecutableLines bool) ([]api.Location, error)
+	Find(d *Debugger, scope *proc.EvalScope, locStr string, cfg api.FindLocCfg) ([]api.Location, error)
 }
 
 type NormalLocationSpec struct {
@@ -251,7 +252,7 @@ func packageMatch(specPkg, symPkg string, packageMap map[string][]string) bool {
 	return partialPackageMatch(specPkg, symPkg)
 }
 
-func (loc *RegexLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string, includeNonExecutableLines bool) ([]api.Location, error) {
+func (loc *RegexLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string, cfg api.FindLocCfg) ([]api.Location, error) {
 	funcs := d.target.BinInfo().Functions
 	matches, err := regexFilterFuncs(loc.FuncRegex, funcs)
 	if err != nil {
@@ -267,7 +268,7 @@ func (loc *RegexLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr st
 	return r, nil
 }
 
-func (loc *AddrLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string, includeNonExecutableLines bool) ([]api.Location, error) {
+func (loc *AddrLocationSpec) Find(d *Debugger, scope *proc.EvalScope, locStr string, cfg api.FindLocCfg) ([]api.Location, error) {
 	if scope == nil {
 		addr, err := strconv.ParseInt(loc.AddrExpr, 0, 64)
 		if err != nil {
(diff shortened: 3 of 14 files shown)


service/debugger/locations.go
```diff
@@ -303,8 +304,33 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
 	return partialPathMatch(loc.Base, path)
 }
 
-func tryMatchRelativePathByProc(expr, debugname, file string) bool {
-	return len(expr) > 0 && expr[0] == '.' && file == path.Join(path.Dir(debugname), expr)
+func (loc *NormalLocationSpec) tryMatchFullPath(d *Debugger, file string) bool {
```
derekparker: It seems like the only thing this method uses from `*Debugger` is `d.processArgs`. Let's just pass that in by itself instead of passing in the entire `Debugger` struct.

service/test/common_test.go
```diff
@@ -131,3 +132,29 @@ func getCurinstr(d3 api.AsmInstructions) *api.AsmInstruction {
 	}
 	return nil
 }
+
+func createSymbolLink(t *testing.T, originFile string, generatedPath string) (string, string) {
```
derekparker: nit: rename to `createSymLink` as it's really short for `symbolic link`.

service/test/common_test.go
```diff
@@ -131,3 +132,29 @@ func getCurinstr(d3 api.AsmInstructions) *api.AsmInstruction {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	symbolLinkAbsDir, err := filepath.Abs(filepath.Join(fixturesDir, generatedPath))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// ensure clean old dir
+	_ = os.RemoveAll(symbolLinkAbsDir)
```
derekparker: Just ignore the return value here, no need to assign to `_`.

service/test/integration2_test.go
```diff
@@ -734,6 +734,16 @@ func TestClientServer_FindLocations(t *testing.T) {
 	}
 }
 
+func TestClientServer_FindLocationsSymbolLink(t *testing.T) {
+	// Issue #1808
+	originAbsPathFile, symbolLinkAbsDir := createSymbolLink(t, "locationsprog3.go", "symbollink")
```
derekparker: nit: s/symbollink/symlink/

service/debugger/locations.go
```diff
@@ -303,8 +304,33 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
...
+	if !filepath.IsAbs(locBaseAbsPath) {
+		return false
+	}
+
+	curFile := file
+	for {
+		if curFile == locBaseAbsPath {
+			return true
+		}
+		desSymbolLinkFile, err := os.Readlink(curFile)
```
derekparker: nit: s/desSymbolLinkFile/symLinkFile/

service/debugger/locations.go
derekparker: nit: s/tryMatchFullPath/fullPathMatch/

This is basically a predicate function, we can just basically ask and return a bool.

service/debugger/locations.go
```diff
@@ -303,8 +304,33 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
...
+	locBaseAbsPath := loc.Base
+	if !filepath.IsAbs(locBaseAbsPath) && len(d.processArgs) >= 1 {
+		locBaseAbsPath = path.Join(path.Dir(d.processArgs[0]), loc.Base)
+	}
+	if !filepath.IsAbs(locBaseAbsPath) {
+		return false
+	}
+
+	curFile := file
+	for {
```
derekparker: Couldn't we replace this loop with https://godoc.org/path/filepath#EvalSymlinks ?

chainhelen: Ex: ...-> a.go -> b.go -> c.go -> d.go, and we build b.go into b.  
The target of b contian information of b.go.  
The key point is I think set breakpoint on b.go/c.go/d.go for b is ok, but a.go not (Maybe I'm wrong, what's your opinion?).  
So if we use EvalSymlinks(a.go/b.go/c.go) == d.go, we can't distinguish between a.go and c.go.

chainhelen:
@derekparker 
Thx for guidance, except for `EvalSymlinks`(I can't make a decision, need your opinion) everything has done and rebase.

service/debugger/locations.go
```diff
@@ -303,8 +304,33 @@ func (loc *NormalLocationSpec) FileMatch(path string) bool {
...
+		if curFile == locBaseAbsPath {
+			return true
+		}
+		symLinkFile, err := os.Readlink(curFile)
```
aarzilli: I'm concerned with the performance of this. How fast is this going to be for a project that has 100k source files? I'm personally leaning towards not supporting this use case at all.

aarzilli: Like, what happens if the file system happens to be a network mountpoint and we stat 100000 files?

chainhelen: I test the performance of Readlink. My pc is 4 core i7 cpu / 8G. 
Create 100K files, and then ReadLink all files. [test_source_code](https://gist.github.com/chainhelen/5732fcd52513f835d9f21f857ca4392f).   
The average cost of Readlink for 100k files is approximate 1.5s 
```
1: 1.49343594s
2: 1.502350766s
3: 1.502523769s
4: 1.562780299s
5: 1.599868709s
6: 1.526759598s
7: 1.586828969s
8: 1.527909997s
9: 1.556420267s
10: 1.505442886s
```  

en,  this pr maybe closed,If we consider the performance problem of 100K.

chainhelen: I have no good idea for 100k, or just only support realpath and save the realpath when debug is starting.

derekparker: I agree with @aarzilli we could see a huge performance hit with this. I'm also leaning towards not implementing this unless we can find a more performant way.

chainhelen: @aarzilli @derekparker   Yes, I agree, thx. 

But I have one more idea, we can add a config to control(default off) and let the user decide(I will finish it tonight).

chainhelen:
Done and has rebased, please check again. Thx.

service/rpc2/server.go
```diff
@@ -571,9 +571,9 @@ func (c *RPCServer) AttachedToExistingProcess(arg AttachedToExistingProcessIn, o
 }
 
 type FindLocationIn struct {
-	Scope                     api.EvalScope
-	Loc                       string
-	IncludeNonExecutableLines bool
+	Scope api.EvalScope
+	Loc   string
+	Cfg   *api.FindLocCfg
```
aarzilli: API changes need to be backwards compatible: you can't remove the IncludeNonExecutableLines parameter from here. Just add the FullPathMatch parameter.

chainhelen: Yes, I do so because I think [ccf57b94](https://github.com/go-delve/delve/commit/ccf57b94#diff-ef4dee432332f81f2fbd37326df20715L5700) use api.FindLocCfg maybe better for compatible.   

I gree with you, I will just add FullPathMatch parameter for backwards compatible.  

Thx.

chainhelen:
Has done and rebase , please check again, thx.

pkg/config/config.go
```diff
@@ -57,6 +57,11 @@ type Config struct {
 	// DebugFileDirectories is the list of directories Delve will use
 	// in order to resolve external debug info files.
 	DebugInfoDirectories []string `yaml:"debug-info-directories"`
+
+	// FullPathMatch will try to eval all file paths to realpath if location contains filename.
```
aarzilli: "When FullPathMatch is true Delve will try to resolve symbolic links when setting breakpoints. Note: this option may cause performance degradation."

service/api/types.go
```diff
@@ -125,6 +125,18 @@ type Thread struct {
 	ReturnValues []Variable
 }
 
+// FindLocCfg describes how to find location from target.
+type FindLocCfg struct {
+	// IncludeNonExecutableLines makes FindLocations return if the expression can be parsed,
+	// even if it doesn't end up matching any instruction in debug_line.
+	IncludeNonExecutableLines bool
+
+	// FullPathMatch will try to eval all file paths to realpath if location contains filename.
```
aarzilli: Same comment description as above.

pkg/config/config.go
chainhelen: @aarzilli en, maybe just `setting breakpoints` is accurate, ex : `list` cmd. 
Need other description, how about changing `setting breakpoints` into `searching source files` above?

chainhelen: Have changed the description of `FullPathMatch`, please check again, thx very much.

aarzilli: This is fine.

derekparker:
See my comment on the original issue. I'm hesitant to merge this as I don't think the original issue it's trying to solve is really a compelling issue that cannot be solved with the existing `substitute-path` config.

derekparker:
@chainhelen I appreciate your work on this, but I am going to close this patch without merging for the issues listed in the original issue. I don't think this is a use case we need to support.

chainhelen:
@derekparker I get it, thx for your gudience. I am happy that there is a conclusion about original issue.

Issue #1800, opened 2019-12-19
print can't output digits with HEX format.


hengwu0:
It's really hard to debug dlv with dlv, if we can't print the values with hex format.

Issue #1038, opened 2017-12-05
Would it be possible to add a feature to dump structures in hexadecimal?


mimoo:
Hello!

At the moment I can print some slices, but what I would like to do would be to do something like this:

```go
print hex.Dump(something)
```

Is this something that is currently possible?

aarzilli:
what do you mean exactly? What would the output look like?

mimoo:
the output would be the hexadecimal string representation of a byte/int slice

derekparker:
This could be solved when we do better conversions for interface values:

```
(dlv) call fmt.Printf("%#v", 4)
> main.main() ./server.go:19 (hits goroutine(1):3 total:3) (PC: 0x73f918)
Command failed: can not convert 4 constant to []interface {}
```

PR #1814, opened 2020-01-07
pkg,service: add cmd `examinemem`(`x`) for examining memory.

### LINKED ISSUE #1584: Could delve inspect goroutine stack frame layout?
I want to list golang stack frame memory layout for diving into golang runtime。and noticed that gdb can inspect memory layout.

I list all delve commands with `help`, and test them one by one, but failed to find which I want.

can delve do it? or any plan about this? 

thank you

chainhelen:
According to #1800 #1584 #1038, `dlv` should enable the user to dive into
memory. User can print binary data in specific memory address range.
But not support for sepecific variable name or structures temporarily.(Because
I have no idea that modify `print` command.)  

Close #1584.

Next some questions: 
1. support the format for `print`cmd, include `address` , `variable name` and `structures`(no idea).  
2. support the form about `t(binary), f(float), a(address), i(instruction), c(char), s(string)` .  
3. support the negative position from address. (This is why also use `x /<FMT> [address]` like gdb, but not `x -<FMT> [address]`).  

----
I can close this pr if you against.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -7,6 +7,10 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/cosiner/argv"
+	"github.com/go-delve/delve/service"
+	"github.com/go-delve/delve/service/api"
+	"github.com/go-delve/delve/service/debugger"
 	"go/parser"
 	"go/scanner"
 	"io"
@@ -20,11 +24,6 @@ import (
 	"strconv"
 	"strings"
 	"text/tabwriter"
-
-	"github.com/cosiner/argv"
-	"github.com/go-delve/delve/service"
-	"github.com/go-delve/delve/service/api"
-	"github.com/go-delve/delve/service/debugger"
 )
 
 const optimizedFunctionWarning = "Warning: debugging optimized function"
@@ -368,6 +367,15 @@ Defines <alias> as an alias to <command> or removes an alias.`},
 	
 If locspec is omitted edit will open the current source file in the editor, otherwise it will open the specified location.`},
 		{aliases: []string{"libraries"}, cmdFn: libraries, helpMsg: `List loaded dynamic libraries`},
+
+		{aliases: []string{"examinemem", "x"}, cmdFn: examineMemoryCmd, helpMsg: `Examine memory:
+
+    examinemem [-fmt <format>] [-len <length>] <address>
+
+Format represents the data format and the value is one of this list (default hex): oct(octal), hex(hexadecimal), dec(decimal), bin(binary).
+Length is the number of bytes (default 1) and must be less than or equal to 1000.
+Address is the memory location of the target to examine.
+For example:  x -fmt hex -len 20 0xc00008af38`},
 	}
 
 	if client == nil || client.Recorded() {
@@ -1339,6 +1347,81 @@ func edit(t *Term, ctx callContext, args string) error {
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
+}
+
+func examineMemoryCmd(t *Term, ctx callContext, args string) error {
+	v := strings.FieldsFunc(args, func(c rune) bool {
+		return c == ' '
+	})
+
+	var (
+		address int64
+		err     error
+		ok      bool
+	)
+
+	// Default value
+	priFmt := byte('x')
+	length := 1
+
+	for i := 0; i < len(v); i++ {
+		switch v[i] {
+		case "-fmt":
+			i++
+			if i >= len(v) {
+				return fmt.Errorf("expected argument after -fmt")
+			}
+			fmtMapToPriFmt := map[string]byte{
+				"oct":         'o',
+				"octal":       'o',
+				"hex":         'x',
+				"hexadecimal": 'x',
+				"dec":         'd',
+				"decimal":     'd',
+				"bin":         'b',
+				"binary":      'b',
+			}
+			priFmt, ok = fmtMapToPriFmt[v[i]]
+			if !ok {
+				return fmt.Errorf("%q is not a valid format", v[i])
+			}
+		case "-len":
+			i++
+			if i >= len(v) {
+				return fmt.Errorf("expected argument after -len")
+			}
+			var err error
+			length, err = strconv.Atoi(v[i])
+			if err != nil || length <= 0 {
+				return fmt.Errorf("len must be an positive integer")
+			}
+			// TODO, maybe configured by user.
+			if length > 1000 {
+				return fmt.Errorf("len must be less than or equal to 1000")
+			}
+		default:
+			if i != len(v)-1 {
+				return fmt.Errorf("unknown option %q", v[i])
+			}
+			// TODO, maybe we can support expression.
+			address, err = strconv.ParseInt(v[len(v)-1], 0, 64)
+			if err != nil {
+				return fmt.Errorf("convert address into uintptr type failed, %s", err)
+			}
+		}
+	}
+
+	if address == 0 {
+		return fmt.Errorf("no address specified")
+	}
+
+	memArea, err := t.client.ExamineMemory(uintptr(address), length)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(api.PrettyExamineMemory(uintptr(address), memArea, priFmt))
+	return nil
 }
 
 func printVar(t *Term, ctx callContext, args string) error {
--- a/service/api/prettyprint.go
+++ b/service/api/prettyprint.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"strconv"
 	"strings"
 )
 
@@ -353,4 +354,54 @@ func (v *Variable) writeSliceOrArrayTo(buf io.Writer, newlines bool, indent stri
 	}
 
 	fmt.Fprint(buf, "]")
+}
+
+func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
+	cols := 8
+	// Avoid emitting rows that are too long when using binary format
+	if format == 'b' {
+		cols = 4
+	}
+
+	l := len(memArea)
+	rows := l / cols
+	if l%cols != 0 {
+		rows++
+	}
+
+	var colFormat string
+	// Leading zero and occupy 8 for binary
+	if format == 'b' {
+		colFormat = "    %#08b"
+	} else {
+		var maxColCharNum int
+		for i := 0; i < rows; i++ {
+			for j := 0; j < cols && i*cols+j < l; j++ {
+				curColCharNum := len(fmt.Sprintf("%#"+string(format), memArea[i*cols+j]))
+				if curColCharNum > maxColCharNum {
+					maxColCharNum = curColCharNum
+				}
+			}
+		}
+		colFormat = "    %#-" + strconv.Itoa(maxColCharNum) + string(format)
+	}
+
+	lines := ""
+	for i := 0; i < rows; i++ {
+		lines += fmt.Sprintf("%#x:", address)
+		for j := 0; j < cols && i*cols+j < l; j++ {
+			curOutput := fmt.Sprintf(colFormat, memArea[i*cols+j])
+
+			// Diffrent versions of golang output differently if binary.
+			// See https://ci.appveyor.com/project/derekparker/delve-facy3/builds/30179356.
+			// Remove prefix `0b` if binary in some versions of golang because it is not graceful.
+			if format == 'b' && strings.Contains(curOutput, "0b") {
+				curOutput = "    " + curOutput[6:]
+			}
+			lines += curOutput
+		}
+		lines += "\n"
+		address += uintptr(cols)
+	}
+	return lines
 }
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -1310,6 +1310,25 @@ func (d *Debugger) ListDynamicLibraries() []api.Image {
 	return r
 }
 
+// ExamineMemory returns the raw memory stored at the given address.
+// The amount of data to be read is specified by length.
+// This function will return an error if it reads less than `length` bytes.
+func (d *Debugger) ExamineMemory(address uintptr, length int) ([]byte, error) {
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
+	thread := d.target.CurrentThread()
+	data := make([]byte, length)
+	n, err := thread.ReadMemory(data, address)
+	if err != nil {
+		return nil, err
+	}
+	if length != n {
+		return nil, errors.New("the specific range has exceeded readable area")
+	}
+	return data, nil
+}
+
 func (d *Debugger) GetVersion(out *api.GetVersionOut) error {
 	if d.config.CoreFile != "" {
 		if d.config.Backend == "rr" {
--- a/service/rpc2/server.go
+++ b/service/rpc2/server.go
@@ -751,3 +751,27 @@ func (s *RPCServer) ListPackagesBuildInfo(in ListPackagesBuildInfoIn, out *ListP
 	out.List = s.debugger.ListPackagesBuildInfo(in.IncludeFiles)
 	return nil
 }
+
+// ExamineMemoryIn holds the arguments of ExamineMemory
+type ExamineMemoryIn struct {
+	Address uintptr
+	Length  int
+}
+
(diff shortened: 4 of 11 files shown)


aarzilli:
See test failures.

chainhelen:
Sorry for test failures of docs, I have done.
(Something faield with others now, not involved with this pr.)

service/debugger/debugger.go
```diff
@@ -1310,6 +1310,16 @@ func (d *Debugger) ListDynamicLibraries() []api.Image {
...
 }
 
+// ExamineMemory returns a list of specific address memory that len is num.
+func (d *Debugger) ExamineMemory(address uintptr, num int) ([]byte, error) {
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
+	thread := d.target.CurrentThread()
+
+	return proc.ExamineMemory(thread, address, num)
```
aarzilli: `proc.ExamineMemory` is only used here, it should be inlined here instead of being a separate function.

service/rpc2/server.go
```diff
@@ -731,3 +731,26 @@ func (s *RPCServer) ListDynamicLibraries(in ListDynamicLibrariesIn, out *ListDyn
 	out.List = s.debugger.ListDynamicLibraries()
 	return nil
 }
+
+// ExamineMemoryIn holds the arguments of ExamineMemory
+type ExamineMemoryIn struct {
+	Num     int
```
aarzilli: I'd prefer Count or Size instead of Num

service/rpc2/server.go
```diff
@@ -731,3 +731,26 @@ func (s *RPCServer) ListDynamicLibraries(in ListDynamicLibrariesIn, out *ListDyn
...
+	Address uintptr
+}
+
+// ExaminedMemoryOut holds the return values of ExamineMemory
+type ExaminedMemoryOut struct {
+	Address uintptr
+	MemData []byte
```
aarzilli: Either Mem or Data instead of MemData.

service/rpc2/server.go
```diff
@@ -731,3 +731,26 @@ func (s *RPCServer) ListDynamicLibraries(in ListDynamicLibrariesIn, out *ListDyn
...
+
+// ExamineMemoryIn holds the arguments of ExamineMemory
+type ExamineMemoryIn struct {
+	Num     int
+	Address uintptr
+}
+
+// ExaminedMemoryOut holds the return values of ExamineMemory
+type ExaminedMemoryOut struct {
+	Address uintptr
```
aarzilli: This has to be the same as the input address so it doesn't need to be sent back.

pkg/terminal/command.go
```diff
@@ -368,6 +369,19 @@ Defines <alias> as an alias to <command> or removes an alias.`},
 	
 If locspec is omitted edit will open the current source file in the editor, otherwise it will open the specified location.`},
 		{aliases: []string{"libraries"}, cmdFn: libraries, helpMsg: `List loaded dynamic libraries`},
+
+		{aliases: []string{"examinemem", "x"}, cmdFn: examineMemoryCmd, helpMsg: `Examine memory: x /[<FMT>] ADDRESS.
```
aarzilli: 1. This description makes it look like the '/' should always be included, i.e. that to use the default for FMT you should type `x / ADDRESS`. 
2. It should use the full name of the command (`examinemem`) instead of `x` for the syntax.
3. To be consistent with the way we describe syntax elsewhere both FMT and ADDRESS should be lowercase and inside angular parenthesis.
4. We don't use a `/` for options anywhere else, it should be a `-` (I get it that it's following gdb syntax but I would prefer internal consistency to external consistency).

service/debugger/debugger.go
chainhelen: Done

service/rpc2/server.go
chainhelen: Done

service/rpc2/server.go
chainhelen: Done

service/rpc2/server.go
chainhelen: Done

pkg/terminal/command.go
chainhelen: Done

service/debugger/debugger.go
```diff
@@ -1310,6 +1310,23 @@ func (d *Debugger) ListDynamicLibraries() []api.Image {
 	return r
 }
 
+// ExamineMemory returns a list of specific address memory that len is num.
```
derekparker: This comment should be rewritten to something like: `ExamineMemory returns the raw memory stored at the given address. The amount of data to be read is specified by `length`. This function will return an error if it reads less than ``length`` bytes.`

service/debugger/debugger.go
```diff
@@ -1310,6 +1310,23 @@ func (d *Debugger) ListDynamicLibraries() []api.Image {
...
+func (d *Debugger) ExamineMemory(address uintptr, count int) ([]byte, error) {
```
derekparker: s/count/length/ .. I think we use `size` internally in `proc/mem.go` as well, but I think either works a bit btter than `count`.

service/debugger/debugger.go
```diff
@@ -1310,6 +1310,23 @@ func (d *Debugger) ListDynamicLibraries() []api.Image {
...
+	d.processMutex.Lock()
+	defer d.processMutex.Unlock()
+
+	thread := d.target.CurrentThread()
+	data := make([]byte, count)
+	rn, err := thread.ReadMemory(data, address)
```
derekparker: s/rn/read/ for a bit more descriptive name.

pkg/terminal/command.go
```diff
@@ -1341,6 +1355,148 @@ func edit(t *Term, ctx callContext, args string) error {
...
+func parseFmtStr(fmtStr string) (count int, format byte, size int, err error) {
+	if len(fmtStr) <= 2 {
+		return 0, ' ', 0, fmt.Errorf("invalid argument for fmt")
+	}
+	if fmtStr[0] != '-' {
+		return 0, ' ', 0, fmt.Errorf("the first letter of fmt should be \"-\"")
+	}
+	i := 1
+	for ; i < len(fmtStr); i++ {
+		if fmtStr[i] < '0' || fmtStr[i] > '9' {
```
derekparker: Could use `unicode.IsDigit` here.

pkg/terminal/command.go
```diff
@@ -368,6 +369,19 @@ Defines <alias> as an alias to <command> or removes an alias.`},
 	
 If locspec is omitted edit will open the current source file in the editor, otherwise it will open the specified location.`},
 		{aliases: []string{"libraries"}, cmdFn: libraries, helpMsg: `List loaded dynamic libraries`},
+
+		{aliases: []string{"examinemem", "x"}, cmdFn: examineMemoryCmd, helpMsg: `Examine memory: examinemem [-<fmt>] <address>.
+	Fmt is a count followed by a format letter and a size letter.
```
derekparker: I wonder if we could just simplify this. Do we really need to specify byte vs halfword vs word etc...?

I'm leaning towards not giving control over chunk size and simply having flags for `-len` and `-fmt`. Length is always in bytes and maybe defaults to the arch pointer size and fmt defaults to hex.

chainhelen: Oh, thx for your guidance. I get something, and need some time to change it.

chainhelen:
Examine memory:

    examinemem [-fmt <format>] [-len <length>] <address>

Format represents the data format and the value is one of this list (default hex): oct(octal), hex(hexadecimal), dec(decimal), bin(binary).
Length is the number of bytes (default 1), please ensure len is not more than 1k.
Address is the memory location of the target to examine.
For example:  x -fmt hex -len 20 0xc00008af38

-------- 

Has done and rebase , please check it again. Thx.

chainhelen:
Sorry for ci failed, it is because I add binary format. It looks diffrent versions of golang output this  differently. I will fix it.

pkg/terminal/command.go
```diff
@@ -1341,6 +1351,175 @@ func edit(t *Term, ctx callContext, args string) error {
 	return cmd.Run()
 }
 
+func parseFmtStr(fmtStr string) (count int, format byte, size int, err error) {
```
aarzilli: This doesn't get called anymore.

pkg/terminal/command.go
```diff
@@ -1341,6 +1351,175 @@ func edit(t *Term, ctx callContext, args string) error {
...
+	})
+	if len(v) != 1 && len(v) != 3 && len(v) != 5 {
+		return fmt.Errorf("wrong arguments for input")
+	}
+	// Default value
+	format := "hex"
+	length := 1
+	addressStr := v[len(v)-1]
+
+	StringArrIndex := func(strArr []string, str string) int {
```
aarzilli: In principle this is fine, however it's also going to make adding new arguments unnecessarily complicated. Something like this is just as simple and more flexible:

```
for i := 0; i < len(v); i++ {
    switch v[i] {
    case "-fmt":
         i++
         // parse argument...
    case "-len":
         i++
         // parse argument...
    default:
         if i != len(v)-1 {
             return fmt.Errorf(...)
         }
         // parse argument...
    }
}
```

similar to parseStackArgs.

pkg/terminal/command.go
```diff
@@ -1341,6 +1351,175 @@ func edit(t *Term, ctx callContext, args string) error {
...
+			return fmt.Errorf("len must be an positive integer")
+		}
+		if length <= 0 {
+			return fmt.Errorf("len must be an positive integer")
+		}
+	}
+
+	// TODO, maybe configured by user.
+	if length > 1000 {
+		return fmt.Errorf("len must be not more than 1k")
```
aarzilli: "len must be less than 1000"

service/api/prettyprint.go
```diff
@@ -354,3 +355,42 @@ func (v *Variable) writeSliceOrArrayTo(buf io.Writer, newlines bool, indent stri
 
 	fmt.Fprint(buf, "]")
 }
+
+func PrettyExamineMemory(address uintptr, res []uint64, format byte) string {
+	cols := int(8)
```
aarzilli: cols is a constant if I'm not mistaken, so:

```
const cols = 8
```

service/api/prettyprint.go
```diff
@@ -354,3 +355,42 @@ func (v *Variable) writeSliceOrArrayTo(buf io.Writer, newlines bool, indent stri
...
+	}
+	colFormat := "    %#-" + strconv.Itoa(maxColCharNum) + string(format)
+
+	lines := ""
+	for i := 0; i < rows; i++ {
+		lines += fmt.Sprintf("%#x:", address)
+		for j := 0; j < cols && i*cols+j < l; j++ {
+			curOutput := fmt.Sprintf(colFormat, res[i*cols+j])
+
+			// Diffrent versions of golang output differently if binanry.
```
aarzilli: s/binanry/binary/

pkg/terminal/command.go
```diff
@@ -1341,6 +1351,175 @@ func edit(t *Term, ctx callContext, args string) error {
...
+	if err != nil {
+		return fmt.Errorf("convert address into uintptr type failed, %s", err)
+	}
+
+	memArea, err := t.client.ExamineMemory(uintptr(address), length)
+	if err != nil {
+		return err
+	}
+
+	res := make([]uint64, 0, length)
```
aarzilli: I think you can skip this copy and just pass memArea to PrettyExamineMemory.

pkg/terminal/command.go
chainhelen: I use `not more than` because `length` belong to [1, 1000] instead of [1, 1000). 
So maybe "len must be less than or equal to 1000“ better, what's your options?

pkg/terminal/command.go
chainhelen: Oh, it is brilliant, thx

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
 	return cmd.Run()
 }
 
+func examineMemoryCmd(t *Term, ctx callContext, args string) error {
+	v := strings.FieldsFunc(args, func(c rune) bool {
+		return c == ' '
+	})
+	if len(v) != 1 && len(v) != 3 && len(v) != 5 {
```
aarzilli: This is not necessary anymore.

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
...
+				"hex":         'x',
+				"hexadecimal": 'x',
+				"dec":         'd',
+				"decimal":     'd',
+				"bin":         'b',
+				"binary":      'b',
+			}
+			priFmt, ok = fmtMapToPriFmt[v[i]]
+			if !ok {
+				return fmt.Errorf("a wrong format argument followed with -fmt")
```
aarzilli: `fmt.Errorf("%q is not a valid format", v[i])`

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
...
+	// Default value
+	priFmt := byte('x')
+	length := 1
+
+	for i := 0; i < len(v); i++ {
+		switch v[i] {
+		case "-fmt":
+			i++
+			if i >= len(v) {
+				return fmt.Errorf("a format must be followed with -fmt")
```
aarzilli: `"expected argument after -fmt"`

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
...
+				"binary":      'b',
+			}
+			priFmt, ok = fmtMapToPriFmt[v[i]]
+			if !ok {
+				return fmt.Errorf("a wrong format argument followed with -fmt")
+			}
+		case "-len":
+			i++
+			if i >= len(v) {
+				return fmt.Errorf("a length must be followed with -len")
```
aarzilli: `"expected argument after -len"`

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
...
+			if err != nil || length <= 0 {
+				return fmt.Errorf("len must be an positive integer")
+			}
+			// TODO, maybe configured by user.
+			if length > 1000 {
+				return fmt.Errorf("len must be less than or equal to 1000")
+			}
+		default:
+			if i != len(v)-1 {
+				return fmt.Errorf("wrong arguments for input")
```
aarzilli: `fmt.Errorf("unknown option %q", v[i])`

pkg/terminal/command.go
```diff
@@ -1341,6 +1349,84 @@ func edit(t *Term, ctx callContext, args string) error {
...
+			// TODO, maybe we can support expression.
+			address, err = strconv.ParseInt(v[len(v)-1], 0, 64)
+			if err != nil {
+				return fmt.Errorf("convert address into uintptr type failed, %s", err)
+			}
+		}
+	}
+
+	if address == 0 {
+		return fmt.Errorf("wrong argument for input, please specify a address")
```
aarzilli: `fmt.Errorf("no address specified")`

service/api/prettyprint.go
```diff
@@ -354,3 +355,53 @@ func (v *Variable) writeSliceOrArrayTo(buf io.Writer, newlines bool, indent stri
 
 	fmt.Fprint(buf, "]")
 }
+
+func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
+	cols := 8
+	// Avoid one row that are too long for binary
```
aarzilli: `// Avoid emitting rows that are too long when using binary format`

hengwu0:
> 1. support the format for `print`cmd, include `address` , `variable name` and `structures`(no idea).

If you are investigativing, please fix this error:
```sh
jim@localhost:~/test$ dlv exec ./A
(dlv) p 123//9
123
(dlv)
```
The Exp:`123//9` should be en error actually.

chainhelen:
> > 1. support the format for `print`cmd, include `address` , `variable name` and `structures`(no idea).
> 
> If you are investigativing, please fix this error:
> 
> ```shell
> jim@localhost:~/test$ dlv exec ./A
> Type 'help' for list of commands.
> (dlv) p 123//9
> 123
> (dlv)
> ```
> 
> The Exp:`123//9` should be en error actually.

I will do my best.

PR #1818, opened 2020-01-09
service/test,pkg: Ensure LineToPC can return pc when pc is not zero.

### LINKED ISSUE #1817: input/output error when setting a breakpoint in some cases

master(059c51e)  

go version go1.12.5 linux/amd64

Ubuntu 18.04.2 LTS  
Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz  

4. What did you do?  
set breakpoint 
```
// t6.go
        package main
        
        import (
        	"fmt"
        	"unsafe"
        	"runtime"
        )
        
        func main() {
        	l := int(51)
        	bs := make([]byte, l)
        	for i := 0; i < l; i++ {
        		bs[i] = byte(i + int(10))
        	}
        
16 ==>  	p := uintptr(unsafe.Pointer(&bs))
        
        	fmt.Println(p)
        
        	bs[0] = 254
        
        	runtime.KeepAlive(bs)
        }
```
`dlv debug t6.go`  and type `b t6.go:16`  

`Breakpoint 1 set at...` successed  Or `Command failed: could not find statement` failed

6. What did you see instead?  
`Command failed: input/output error`   

-----
This is becase [bininfo.go#L358](https://github.com/go-delve/delve/blob/059c51ea0caf6cba882f248284c52245423a14c6/pkg/proc/bininfo.go#L358) get one nonzero `pc`, but [bininfo.go#L390](https://github.com/go-delve/delve/blob/059c51ea0caf6cba882f248284c52245423a14c6/pkg/proc/bininfo.go#L390)`appendLineToPCIn` doesn't append it. Actually the essential difference is the judgemet about `isStmt` of `StateMachine` between  [LineToPC](https://github.com/go-delve/delve/blob/059c51ea0caf6cba882f248284c52245423a14c6/pkg/dwarf/line/state_machine.go#L272) and [LineToPCIn](https://github.com/go-delve/delve/blob/059c51ea0caf6cba882f248284c52245423a14c6/pkg/dwarf/line/state_machine.go#L306).  (It is involved in ##1717.) 

I not sure whether the compiler with `-gcflag="all=-N -l"` doesn't write any `lineentry` with `isstmt` being true for `16th line` is correct or not ? (I need more time to check it because my current weak ability). 

But I think I can add one pr 
```
	if len(pcs) == 0 {
		return []uint64{pc}, nil
	}
```
and one testcase to ensure backward compatibility.

chainhelen:
`LineToPc` return zero that cause `list filename:lineno` cmd display `(PC: 0x0)`
instead of source code in some case. This commit ensure `LineToPC` can return pc at lease
when pc is not zero.

Fix #1817

### DIFF
--- a/_fixtures/locationsprog.go
+++ b/_fixtures/locationsprog.go
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"runtime"
+	"unsafe"
 )
 
 type SomeType struct {
@@ -35,6 +37,12 @@ func main() {
 	a.SomeFunction()
 	anotherFunction()
 	ioutil.ReadFile("nonexistent.file.txt")
+
+	//Issue #1817
+	bs := make([]byte, 100)
+	p := uintptr(unsafe.Pointer(&bs))
+	fmt.Println(p)
+	runtime.KeepAlive(bs)
 }
 
 var amap map[string]func()
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -389,6 +389,11 @@ func (bi *BinaryInfo) LineToPC(filename string, lineno int) (pcs []uint64, err e
 	for _, call := range fn.InlinedCalls {
 		pcs = appendLineToPCIn(pcs, filename, lineno, call.cu, bi.PCToFunc(call.LowPC), call.LowPC, call.HighPC)
 	}
+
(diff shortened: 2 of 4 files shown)


aarzilli:
The right way to fix this is to fix `LineToPCIn`.

chainhelen:
Done in #1821, close this pr. Thx for your guidence.

aarzilli:
Apologies for the trouble.

Issue #1750, opened 2019-11-08
refactor: Simplify pkg/proc


derekparker:
I think that `pkg/proc` has gotten way too big in scope, making it difficult to understand especially for those new to the project. I believe this clutter also makes it more complicated to add new arches and OSes. I propose we simplify `pkg/proc` by limiting its scope exclusively to low-level process manipulation, and move everything else into independent packages, and move any higher-level logic into `service/debugger`. As it stands right now, `pkg/proc` is aware of: low-level process execution / memory manipulation, DWARF logic (compilation units, line info, etc), various binary sections including moduledata, disassembly, variables and variable types, parsing the Go AST, the function call protocol, AND implementing other even lower-level backends.

I propose that `pkg/proc` implement a simple interface which abstracts what backend we are using, and is only responsible for manipulation of the process execution and memory. All other functionality should be split up into independent packages, and used by a higher-level concept which would be the `debugger` package.

The logic that should be moved out into independent packages, as I see it, is as follows:

- Variable eval / AST parsing code

I think this logic can be implemented in its own package. Functions can take a `proc.MemReadWriter` in order to actually get the values from memory and turn them into something useful, but this shouldn't be a part of the `proc` package itself.

- BinaryInfo

A lot of this code is higher level than `proc` as it deals with parsing the various sections in the binary we need to gather info from in order to unwind the stack, read variables, etc. I think it can be moved into its own package and used by debugger instead of `proc`.

- Stack unwinding

I believe this can be implemented in `service/debugger` by taking advantage of the new packages above and low-level functionality provided by the `proc` package.

- Next'ing, function call protocol and other high-level logic

This can be implemented by the `debugger` package as well. At a process level there is no concept of source code lines. This logic should be implemented at a higher level, leaving `proc` to implement _only_ the low-level functionality that the logic builds on top of.

- Disassembly

Again, this is high level, this could be it's own standalone package.

- Breakpoints

The concept of breakpoints is only of use to the debugger, an actual process knows nothing about them. This concept should be moved higher up.

I am creating this issue for awareness and discussion. At the time of writing there are also arch ports in play which this refactor would disrupt, so it should be completed after.

aarzilli:
I agree that the disassembler could and should be split, and also that 'pkg/proc' is a bad name for what's going on inside 'pkg/proc'. I'd also suggest that the loclist code in bininfo.go is a good candidate for a new 'pkg/dwarf' subpackage.

[...]

PR #1823, opened 2020-01-10
*: Introduce debug package and Target concept

derekparker:
Introduce a new package `debug` with a new main exported type `Target`. 

Package `proc` was getting out of hand. It simply knew "too much". It knew how to interface with the operating system to control a process, how to parse some DWARF information, how to perform complex high level tasks such as Next / Step / StepInto, it knew about the Go AST, variables, etc. It essentially held all functionality in a single bloated package. With this there was also a lot of duplication in the backend implementations.

This PR is the first towards refactoring the internals of Delve a bit. Simplify the code, remove duplication, and ensure that functionality is split into separate package where appropriate with easy to use public interfaces. This not only helps us internally with maintaining the code and potentially optimizing it, but also opens up new possibilities for building tools that take advantage of Delve internals.

To that end here we introduce a `debug.Target` struct which acts as the main type for interacting with the process we are debugging. It utilized the low-level concepts within `pkg/proc` and builds on top of them to create higher level functionality. In doing so we are able to get rid of code duplication especially in startup and detaching, but also code used throughout the lifecycle of the debug session.

This is only the first PR of this refactor, so not everything is perfect nor split up into separate packages quite yet.

Updates #1750


derekparker:
@aarzilli things are working with with this, with the exception of Windows. Windows seems to keep failing on `StepConcurrent*` tests. There is an unexpected fault error which causes a panic, which causes Delve to switch thread and goroutine which causes the test to fail. I haven't yet tracked down what is causing this failure, but I hoped that by posting this now perhaps more eyes could help reveal the issue.

derekparker:
Going to close this out and try and different approach to end up at (somewhat) of the same place, but in a slower more controlled way to allow for review and not disrupting other work that's happening.

I think the discussion that's happened so far has been nevertheless productive and this example got a great conversation started.

PR #1824, opened 2020-01-11
*: Introduce debug target pt1

derekparker:
#1823 is way too large of a PR to be meaningfully reviewed and merged, but I think having the end result visible somewhere is still important.

Here is Pt.1 of that refactor, which is mostly mechanical changes and code consolidation. Here we introduce the `debug` package and the `debug.Target` type and move some of the code and files around into the `debug` package.

@aarzilli this should be easier to review, if you'd like it broken down even further let me know. We can still reference the other PR for what the end result should eventually look like.


pkg/debug/target.go
```diff
@@ -0,0 +1,289 @@
...
+// It is responsible for implementing the high level logic
+// that is used to manipulate and inspect a running process.
+type Target struct {
+	proc.Process
+
+	bi *proc.BinaryInfo
+}
+
+// New returns an initialized Target.
+func New(p proc.Process, os, arch string, debugInfoDirs []string) (*Target, error) {
```
aarzilli: I don't like this, Launch, Attach and OpenCoreOrRecording. I think the way we have it now, with the backends creating and returing a proc.Process object is better. I do like the debug.Target struct but I think we should keep the setup where initialization is done in the backends and they instead return a debug.Target object.

Doing it this way (with debug.Launch, debug.Attach...) leads to the awkward and confusing SetTarget method. I know this is just a step towards a refactoring where SetTarget shouldn't be necessary but (a) I don't think it's possible to rewrite the backends in such a way that they don't need to access the Target object, (b) even if it's possible I don't believe it to be a worthwhile goal, in fact I think it will make delve's codebase less flexible.

Doing it this way also makes the initialization procedure more complicated: note for example how the core backend needs to return the os/arch pair to pkg/debug which will then call an Initialization method inside the core package, vs. how it worked before.

pkg/debug/target.go
```diff
@@ -0,0 +1,289 @@
...
+	t.SetSelectedGoroutine(g)
+
+	createUnrecoveredPanicBreakpoint(t)
+	createFatalThrowBreakpoint(t)
+	return nil
+}
+
+// Detach will force the target to stop tracing the process.
+// If kill is true the process will be killed during the detach.
+func (t *Target) Detach(kill bool) error {
```
aarzilli: This method, Pid, SelectedGoroutine, Recorded, Restart, Direction, When, Checkpoint, etc. do nothing since embedding will automatically take care of them.

pkg/proc/bininfo.go
```diff
@@ -255,33 +255,25 @@ type ElfDynamicSection struct {
...
-		bi.lastModified = fi.ModTime()
-	}
-
-	bi.debugInfoDirectories = debugInfoDirs
-
-	return bi.AddImage(path, entryPoint)
-}
-
-func loadBinaryInfo(bi *BinaryInfo, image *Image, path string, entryPoint uint64) error {
+func parseBinarySections(bi *BinaryInfo, image *Image, path string, entryPoint uint64) error {
```
aarzilli: I'd rather keep the old name for now.

pkg/proc/bininfo.go
```diff
@@ -296,6 +288,12 @@ func loadBinaryInfo(bi *BinaryInfo, image *Image, path string, entryPoint uint64
 	return errors.New("unsupported operating system")
 }
 
+// SetLogger sets the internal logger to use for
+// the BinaryInfo object. Mostly useful for tests.
+func (bi *BinaryInfo) SetLogger(l *logrus.Entry) {
```
aarzilli: It looks like this is only used by a single test, isn't there any way around this?

pkg/proc/bininfo.go
```diff
@@ -503,6 +501,12 @@ func (image *Image) registerRuntimeTypeToDIE(entry *dwarf.Entry, ardr *reader.Re
 // the relocation offset) for all other images.
 // The first image added must be the executable file.
 func (bi *BinaryInfo) AddImage(path string, addr uint64) error {
+	if bi.lastModified.IsZero() {
```
aarzilli: This is altering the semantics of bi.lastModified. It used to be the last modified time of the executable or zero if the executable can not be stat'ed. Now it is the last modified time of the first image we can stat. The old semantics is better.

pkg/proc/core/core.go
```diff
@@ -243,9 +249,9 @@ func (p *Process) EntryPoint() (uint64, error) {
 	return p.entryPoint, nil
 }
 
-// writeBreakpoint is a noop function since you
+// WriteBreakpoint is a noop function since you
 // cannot write breakpoints into core files.
-func (p *Process) writeBreakpoint(addr uint64) (file string, line int, fn *proc.Function, originalData []byte, err error) {
+func (p *Process) WriteBreakpoint(addr uint64) (file string, line int, fn *proc.Function, originalData []byte, err error) {
```
aarzilli: writeBreakpoint really shouldn't be an exported function, since it can't be called safely by anyone except the breakpoint map.

pkg/proc/core/core.go
```diff
@@ -391,9 +397,9 @@ func (p *Process) ClearInternalBreakpoints() error {
 	return nil
 }
 
-// ContinueOnce will always return an error because you
+// Resume will always return an error because you
 // cannot control execution of a core file.
-func (p *Process) ContinueOnce() (proc.Thread, error) {
+func (p *Process) Resume() (proc.Thread, error) {
```
aarzilli: I prefer to keep the old name for now.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1329,7 +1337,7 @@ func (t *Thread) Blocked() bool {
 // inferior's thread.
 func (p *Process) loadGInstr() []byte {
 	var op []byte
-	switch p.bi.GOOS {
+	switch runtime.GOOS {
```
aarzilli: Given what we are doing now this is fine however in principle one could use the gdbserial package to connect to a stub running on a different OS/architecture than the one where delve is running. We'll probably never do this but there's no reason to make this harder, either.

pkg/debug/target.go
derekparker: > I don't like this, Launch, Attach and OpenCoreOrRecording. I think the way we have it now, with the backends creating and returing a proc.Process object is better. I do like the debug.Target struct but I think we should keep the setup where initialization is done in the backends and they instead return a debug.Target object.
> 

I disagree; one large goal of this refactor is to reduce duplication and make the backends as simple as possible. They should, as much as possible, _only_ know how to do low-level process related tasks (continue a process, read memory, write memory, stop a process, etc) with a higher level package building on top of the simple interface. If we don't centralize the initialization logic then we have every single backend knowing how to initialize the BinaryInfo object, so we repeat that code with every backend. 

> Doing it this way (with debug.Launch, debug.Attach...) leads to the awkward and confusing SetTarget method. I know this is just a step towards a refactoring where SetTarget shouldn't be necessary but (a) I don't think it's possible to rewrite the backends in such a way that they don't need to access the Target object, (b) even if it's possible I don't believe it to be a worthwhile goal, in fact I think it will make delve's codebase less flexible.
> 

This is why I also wanted to post to full change list. This was just a refactor step which let me start breaking things up while still being able to run the tests locally and ensure the refactor wasn't breaking anything. It's temporary and is not present in the final commits of the larger PR.

> Doing it this way also makes the initialization procedure more complicated: note for example how the core backend needs to return the os/arch pair to pkg/debug which will then call an Initialization method inside the core package, vs. how it worked before.

One backend having to return a few more parameters which only the `debug` package sees seems like a win over _every_ backend having to repeat startup and initialization code.

pkg/debug/target.go
derekparker: Agreed, they are redundant and can be removed for the sake of the diff. I created them originally because I didn't simply embed `proc.Process` at first, that change came a bit later.

pkg/proc/bininfo.go
derekparker: Fine by me.

pkg/proc/bininfo.go
derekparker: Hopefully, I can look into it, I'd rather remove it as well.

pkg/proc/bininfo.go
derekparker: Ack, you're right. I'll try and find a better way.

pkg/proc/core/core.go
derekparker: In my opinion I don't think anything in `pkg/proc` should even know about breakpoints after this refactor.

pkg/proc/core/core.go
derekparker: Fine by me.

pkg/proc/gdbserial/gdbserver.go
derekparker: Ack, agreed.

pkg/debug/dwarf_expr_test.go
```diff
@@ -27,7 +28,8 @@ func fakeBinaryInfo(t *testing.T, dwb *dwarfbuilder.Builder) (*proc.BinaryInfo,
 	dwdata, err := dwarf.New(abbrev, aranges, frame, info, line, pubnames, ranges, str)
 	assertNoError(err, t, "creating dwarf")
 
-	bi := proc.NewBinaryInfo("linux", "amd64")
+	bi := &proc.BinaryInfo{GOOS: "linux", Arch: proc.AMD64Arch("linux")}
```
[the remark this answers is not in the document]
derekparker: I don't remember why I changed this... doesn't make sense, I'll change it back.

pkg/debug/target.go
derekparker: I've replaced `SetTarget` with `SetBinaryInfo` to simply re-inject it into the `proc.Process`. This is still a win overall because the actual creation of the BinaryInfo is still done at a higher level. Again, I think eventually this should not be necessary.

aarzilli: >I disagree; one large goal of this refactor is to reduce duplication and make the backends as simple as possible

I think we have hit an irreconciliable difference of opinion. I don't think what you want to do can or should be done and I don't see myself ever being swayed to the contrary.

Let's approach this from a different angle. Let's call point A the state of the code base as it is now. And let's call B the state of the code base as it would be after this PR. Let's call C the state where debug.Target gets introduced but the initialization procedure of the backend does not change and there is no debug.Launch/debug.Attach/etc. This introduces debug.Target which we both like and gets us closer to removing proc.CommonProcess which we both don't like, right? Can we also agree that point C is intermediate between A and B? If so can we just get to C for now, finish removing CommonProcess and then revisit the initialization procedure later?

derekparker: I'm 100% fine with compromising on option C if only for the sake of moving forward, but I kind of want to understand your hesitation to the consolidation of the startup code. From my perspective, of all the proposed changes this seems the most benign.

Here's my thoughts and justification on this change:

1. Less code will always decrease maintenance load. If packageA, packageB and packageC (native, gdbserial, core) all must perform the same action, and as a result of that end up producing a `debug.Target` struct, why not move all the code that all 3 packages are copying from each other into one central location?
2. There will only ever be one, single `debug.Target` implementation. There are already 3 backend implementations and there's the possibility for more. The more logic we consolidate into `debug.Target` means less logic _every_ _single_ backend will have to implement and duplicate, which I would argue leads to fewer bugs in the long run. Right now both native and gdbserial know about the goroutine cache. What if we find a bug where the cache isn't being cleared properly? Now we have to make sure it's fixed 2 places instead of just one. How is that a win?
3. Having the backend packages know about `*debug.Target` seems backwards to me and provides a leaky abstraction. In this case, now everything in the project has to know about the various backend packages AND the `debug` package, so instead of simplifying things and removing we're now instead _adding_ a new concept _on_ _top_ of the existing concepts. This refactor is meant to be reductive not additive.

Other than the fact that it's going to be tricky and must be done with care, I'm still not convinced that simplifying the backends in any way we possibly can would hurt us in the long run. The simpler and dumber they are, the easier it will be to add new OS support. If all we have to do is satisfy a tiny interface that is basically "start process, stop process, read memory, write memory, set breakpoint (in the case of the gdbserial backend we can leave this for now)" then adding a new backend becomes trivial. It doesn't have to worry if it's clearing the G cache, or creating the binary info struct, or setting the "current thread" or "current goroutine". Everything that is common and _must_ be done by all backends should be lifted out and implemented in one single place. Then we only ever have to write that code once, and we can add a million backends after that, and that code will never be duplicated.

aarzilli: 1. I don't think this actually removes much duplicate code, anything that can be done with this can be done with PostInitializationSetup.

2. It makes the control flow more complicated. For example now a backend's Launch in the backend calls NewBreakpointMap, NewBinaryInfo and then PostInitializationSetup. With this PR debug.Launch calls backend.Launch which does some work, returns to debug.Launch which calls debug.(*Target).Initialize which calls back the backend's Initialize function. I don't like the control bounching back and forth between debug and the backend.

3. The final version of this PR has a couple of downcasts in debug.(*Target).Initialize to execute different code for special backends, that suggests this is the wrong approach.

4. I think nobody will ever write another implementation of proc.Process, ever. The last time it happened it was april 2017, roughly 60% of the way through the entire life of this project.

derekparker: I agree with the points you made, especially regarding having to type cast and having control flow bouncing back and forth. I'll rework this a bit.

derekparker:
Going to close this out and try and different approach to end up at (somewhat) of the same place, but in a slower more controlled way to allow for review and not disrupting other work that's happening.

I think the discussion that's happened so far has been nevertheless productive and this example got a great conversation started.

PR #1827, opened 2020-01-13
service/test: Fix a wrong judgment about err.

chainhelen:
Should judge `state.Err` instead of `err`.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -364,8 +364,8 @@ func TestClientServer_breakpointInMainThread(t *testing.T) {
 		}
 
 		state := <-c.Continue()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


service/test/integration2_test.go
```diff
@@ -364,7 +364,7 @@ func TestClientServer_breakpointInMainThread(t *testing.T) {
 		}
 
 		state := <-c.Continue()
-		if err != nil {
+		if state.Err != nil {
 			t.Fatalf("Unexpected error: %v, state: %#v", err, state)
```
derekparker: This should also be `state.Err`.

chainhelen:
Oh, sorry, has done.

PR #1829, opened 2020-01-14
cmd,pkg,service: support mtls and tls with token.

### LINKED ISSUE #861: Support token and TLS for remote debugging
This is actually a suggestion / feature request.

I've just realized that it would be preferable to support TLS and token authentication for the debugger when running in the remote configuration in order to protect against possible issues.

What do you think?

chainhelen:
Ensure the safety of data transmission during remote debugging by mtls or tls
with token.

Fix #861 #1332  

---- 
I implement two ways about tls  (Tls should be L4, and token seem like L7).    

**1. mtls**  
Generate 3 key pairs(6 files) for ca/client/server. (no need to have token. Maybe not friendly for goland? and very complicated to me), can use the following steps to test. 
 
　　1.1.  Generate (Note: the ip for server default 127.0.0.1) 
       ```use openssl by ./scripts/openssl_generate_mtls_cert.sh (or go run ./scripts/openssl_generate_mtls_cert.go --host 127.0.0.1 --output ./mtls)```  

　　1.2. Server side 
       ```dlv debug main.go --mtls-ca-crt ./mtls/ca.crt --mtls-crt ./mtls/server.crt --mtls-key ./mtls/server.key  --headless --listen 127.0.0.1:3333```    

　　1.3. Client side
       ```dlv connect 127.0.0.1:3333 --mtls-ca-crt ./mtls/ca.crt --mtls-crt ./mtls/client.crt --mtls-key ./mtls/client.key```   

**2. tls(One-Way SSL) and token**
    2.1. Just only need the machine of server side to generate one key pairs(2 files, crt is signed with itself without ca)    
  ```   
# use openssl
# some parts are borrowed by https://stackoverflow.com/questions/21488845/how-can-i-generate-a-self-signed-certificate-with-subjectaltname-using-openssl   

# tlstoken.sh
mkdir tlstoken 

# specity the ip that the server will listen or client will connect
echo -e "[req]\nx509_extensions = v3_ca\ndistinguished_name = req_distinguished_name\n[req_distinguished_name]\n\n[v3_ca]\n\nsubjectAltName = @alt_names\nbasicConstraints = CA:FALSE\nkeyUsage = nonRepudiation, digitalSignature, keyEncipherment\n\n\n[alt_names]\nIP = 127.0.0.1" > ./tlstoken/config.cnf  

openssl genrsa -out ./tlstoken/server.key 2048  

openssl req -new -x509 -key ./tlstoken/server.key -days 3650 -subj "/CN=go-delve"  -config ./tlstoken/config.cnf -out ./tlstoken/server.crt
```
or 
```
use "go run $GOROOT/src/crypto/tls/generate_cert.go --host 127.0.0.1"  to generate (Be careful to the name and path of files maybe different with below) 
```

　　2.2 Server side 
　　　```dlv debug main.go --tls-crt ./tlstoken/server.crt --tls-key ./tlstoken/server.key --tls-token godelve  --headless --listen 127.0.0.1:3333```    

　　2.3 Client side (need to copy server.crt from server side)  
　　　```dlv connect 127.0.0.1:3333 --tls-crt ./tlstoken/server.crt --tls-token godelve```    

----   

I think it's possible that mtls can be removed because it is unfriendly to operate, just save tls with token.  
Need discussion,  @dlsniper @justinclift (#861) @antong (#1332) , thx.

### DIFF
--- a/Documentation/cli/starlark.md
+++ b/Documentation/cli/starlark.md
@@ -20,6 +20,7 @@ Function | API Call
 amend_breakpoint(Breakpoint) | Equivalent to API call [AmendBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AmendBreakpoint)
 ancestors(GoroutineID, NumAncestors, Depth) | Equivalent to API call [Ancestors](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Ancestors)
 attached_to_existing_process() | Equivalent to API call [AttachedToExistingProcess](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AttachedToExistingProcess)
+auth(Token) | Equivalent to API call [Auth](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Auth)
 cancel_next() | Equivalent to API call [CancelNext](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.CancelNext)
 checkpoint(Where) | Equivalent to API call [Checkpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Checkpoint)
 clear_breakpoint(Id, Name) | Equivalent to API call [ClearBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.ClearBreakpoint)
--- a/Documentation/usage/dlv.md
+++ b/Documentation/usage/dlv.md
@@ -30,6 +30,12 @@ Pass flags to the program you are debugging using `--`, for example:
       --log                  Enable debugging server logging.
       --log-dest string      Writes logs to the specified file or file descriptor (see 'dlv help log').
       --log-output string    Comma separated list of components that should produce debug output (see 'dlv help log')
+      --mtls-ca-crt string   Specify ca crt file path of mtls to establish tcp connection.
+      --mtls-crt string      Specify crt file path of mtls to establish tcp connection.
+      --mtls-key string      Specify key file path of mtls to establish tcp connection.
+      --tls-crt string       Specify crt file path of tls, need to be generated manually by server and set by server/client.
+      --tls-key string       Specify key file path of tls, need to be generated manually by server and only set by server.
+      --tls-token string     Specify token string for authentication, set same value by server/client.
       --wd string            Working directory for running the program. (default ".")
 ```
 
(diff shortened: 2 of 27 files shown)


justinclift:
> Need discussion, @dlsniper @justinclift (#861) @antong (#1332) , thx.

I'm probably not the right person to discuss this at the moment.  I'm deeply into some completely unrelated stuff, and haven't even thought about Go nor TLS in months. :wink:

aarzilli:
I'm not competent to review the TLS stuff, the auth token seems to work the opposite of how you would expect it: it allows the client to authenticate that it's connecting to a specific server, which is not that useful (as opposed to allowing the server to check that the client was authorized).

chainhelen:
@aarzilli 
1. mtls is double mutual authentication, The authority control lies in CA issuing certificate to who. 
We can konw this is no problem, but maybe complex.

2. Then the tls with token. 
In fact, token auth has no direct relationship with tls.(Just I bind them)
When the server verifies the token from client, it will trust this tcp.   
Tls just ensure the packets of L4 will not be eavesdropped (like token) or modified message. Token ensure L7 auth.  
And the final is also two-way, so I think there is no opposite. (Just a little winding)      

Logically, security is guaranteed by tls. I do it by suggession from #861. I find that `tls` implement is not graceful enough when I finish, or `tls` is not suitable for this scenario.

Maybe using ssh (do it in programm, not ssh cli) like [scenarios](https://nodejs.org/de/docs/guides/debugging-getting-started/#enabling-remote-debugging-scenarios) is better? I will try it. 
Any other discussion can continue, I think.   

(If any other problem, I can close this pr, thx)

dlsniper:
Thanks for the mention. I'll cc my colleague who's working on the debugger interaction, @nd.

I haven't had time to fully look at the PR, nor am I qualified to do a review from a security perspective.

That being said, I think that securing the communication with TLS is a good start on this.

aarzilli:
> When the server verifies the token from client, it will trust this tcp.

But the server doesn't do anything with this verification, the verification can fail and the client will still be able to send more commands. In fact the client can skip requesting the token verification entirely. The only thing this accomplishes is that the client verifies that the server knows the secret too, and in the process it reveals the secret to the server, so it's only useful the first time anyway.

In general I'm still skeptical that we should be doing this in delve. There are several ways to add encryption to remote delve connections: stunnel, an ssh tunnel or some kind of VPN. For local connections we are now verifying that these have the same user as origin, at least on linux, so it shouldn't be a problem there anymore. It remains a problem on Windows and macOS, but those also tend to be single-user systems.

And as far as I can tell neither gdbserver nor lldb-server implement anything like this.

derekparker:
Sorry for taking so long to get to this.

I agree with @aarzilli in that I think this should not be the responsibility of Delve. I'm not for generating our own keys, etc. I think implementing security via an SSH tunnel could be interesting, but I think that's about it.

Thanks for sending in this patch, however, and taking the time to work on this.

chainhelen:
As my understanding of the project deepens, I also agree with @aarzilli . 
Just some docs to explain how to deal with some security scenarios  are better.
And delve focus on debugger itself.

PR #1830, opened 2020-01-14
proc: cache mapping between goroutine ID and runtime.allgs index

aarzilli:
```
proc: cache mapping between goroutine ID and runtime.allgs index

Caches the mapping between a goroutine ID and its position inside the
runtime.allgs array, this allows speeding up FindGoroutine and makes
commands like 'goroutines -t' not be accidentally quadratic in the
number of goroutines.

```

### DIFF
--- a/pkg/proc/interface.go
+++ b/pkg/proc/interface.go
@@ -114,10 +114,18 @@ type BreakpointManipulation interface {
 // CommonProcess contains fields used by this package, common to all
 // implementations of the Process interface.
 type CommonProcess struct {
-	allGCache     []*G
+	goroutineCache
+
 	fncallEnabled bool
 
 	fncallForG map[int]*callInjection
+}
+
+type goroutineCache struct {
+	partialGCache map[int]*G
+	allGCache     []*G
+
+	allgentryAddr, allglenAddr uint64
 }
 
 type callInjection struct {
@@ -132,9 +140,4 @@ type callInjection struct {
 // all process implementations.
 func NewCommonProcess(fncallEnabled bool) CommonProcess {
 	return CommonProcess{fncallEnabled: fncallEnabled, fncallForG: make(map[int]*callInjection)}
-}
-
-// ClearAllGCache clears the cached contents of the cache for runtime.allgs.
-func (p *CommonProcess) ClearAllGCache() {
-	p.allGCache = nil
 }
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -18,6 +18,8 @@ var ErrNotExecutable = errors.New("not an executable file")
 // only possible on recorded (traced) programs.
 var ErrNotRecorded = errors.New("not a recording")
 
+var ErrNoRuntimeAllG = errors.New("could not find goroutine array")
+
 const (
 	// UnrecoveredPanic is the name given to the unrecovered panic breakpoint.
 	UnrecoveredPanic = "unrecovered-panic"
@@ -71,6 +73,8 @@ func PostInitializationSetup(p Process, path string, debugInfoDirs []string, wri
 
 	createUnrecoveredPanicBreakpoint(p, writeBreakpoint)
 	createFatalThrowBreakpoint(p, writeBreakpoint)
+
+	p.Common().goroutineCache.init(p.BinInfo())
 
 	return nil
 }
@@ -507,12 +511,9 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
 		}
 	}
 
-	exeimage := dbp.BinInfo().Images[0] // Image corresponding to the executable file
-
 	var (
 		threadg = map[int]*G{}
 		allg    []*G
-		rdr     = exeimage.DwarfReader()
 	)
 
 	threads := dbp.ThreadList()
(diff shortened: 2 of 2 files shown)


derekparker:
Any reason not to maintain the list of Goroutines as a map regardless? `map[<id>]*G`.

aarzilli:
We would have to nuke that every time we resume the target while the map to indexes can survive that.

pkg/proc/proc.go
```diff
@@ -588,6 +567,53 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
 	return allg, -1, nil
 }
 
+func getAllGs(dbp Process) (uint64, uint64, error) {
```
derekparker: This function technically only needs a `*BinaryInfo` struct and a `MemoryReadWriter`. Instead of passing in the whole `Process` let's just pass in `dbp.BinInfo()` and `dbp.CurrentThread()`.

pkg/proc/proc.go
```diff
@@ -588,6 +567,53 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
...
+	allgptr := binary.LittleEndian.Uint64(faddr)
+
+	return allgptr, allglen, nil
+}
+
+// gorountineAtIndex returns the goroutine at index i in runtime.allgs
+func goroutineAtIndex(p Process, i uint64) (*G, error) {
+	allgptr, allglen, err := getAllGs(p)
+	if i >= allglen {
+		return nil, fmt.Errorf("Index out of bounds")
```
derekparker: nit: s/Index/index/.

pkg/proc/proc.go
derekparker: Also, shouldn't this be the level in which we maintain the allgcache? This gets called from both `GoroutinesInfo` and `FindGoroutine`. We could cache the address of runtime.allg as well. This would offer significant improvement as we won't need to issue a bunch of calls to read process memory, dig through debug info, etc.

pkg/proc/interface.go
```diff
@@ -114,6 +114,7 @@ type BreakpointManipulation interface {
 // CommonProcess contains fields used by this package, common to all
 // implementations of the Process interface.
 type CommonProcess struct {
+	goIDToIdx     map[int]uint64
```
derekparker: Maybe we have a new type here.

Something like (and this is just a sketch):

```
type GoroutineCache struct {
  allgAddr uint64
  allg []*G
  idx  map[int]uint64
}

func (gc *GoroutineCache) AllGAddr() uint64 {
  return gc.allgAddr
}

func (gc *GoroutineCache) AddGoroutine(...) {
  ...
}

func (gc *GoroutineCache) FindGoroutine(...) *G {
 ...
}
```

That way we can abstract it a bit more and callers don't need to know the internals of where the idx map is allocated.

pkg/proc/proc.go
```diff
@@ -588,6 +567,53 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
...
+	}
+	gvar, err := newGVariable(p.CurrentThread(), uintptr(allgptr+(uint64(i)*uint64(p.BinInfo().Arch.PtrSize()))), true)
```
derekparker: Can we improve the caching by using a *G we already parsed? I understand that the GoID -> index survives longer because it doesn't depend on our internal cache, but is there a possibility to link them somehow? Like if we already have a parsed the goroutine at this index before and it's just sitting in our allgcache for example, can we just figure out a way for the index cache to return that instead of reading the memory all over again? Since we're directly calling `CurrentThread().ReadMemory()` we're not even getting the benefit of memory caching even if we've already read it before.

aarzilli:
I've added a cache for *G objects and removed the GoID -> index cache. I think that if we cache *G objects there is no code path where the other cache gets used.

pkg/proc/proc.go
```diff
@@ -588,6 +568,56 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
...
+	_, err = mem.ReadMemory(faddr, uintptr(gcache.allgentryAddr))
+	if err != nil {
+		return 0, 0, err
+	}
+	allgptr := binary.LittleEndian.Uint64(faddr)
+
+	return allgptr, allglen, nil
+}
+
+func (gcache *goroutineCache) addGoroutine(i uint64, g *G) {
```
derekparker: `i` isn't actually used.

pkg/proc/proc.go
```diff
@@ -588,6 +568,56 @@ func GoroutinesInfo(dbp Process, start, count int) ([]*G, int, error) {
...
+
+	rdr.Seek(0)
+	gcache.allgentryAddr, err = rdr.AddrFor("runtime.allgs", exeimage.StaticBase)
+	if err != nil {
+		// try old name (pre Go 1.6)
+		gcache.allgentryAddr, _ = rdr.AddrFor("runtime.allg", exeimage.StaticBase)
+	}
+}
+
+func (gcache *goroutineCache) getAllGs(bi *BinaryInfo, mem MemoryReadWriter) (uint64, uint64, error) {
```
derekparker: nit: The name of this method is a bit misleading. It isn't actually getting Gs it's just returning the pointer to the allg struct and the length.

aarzilli: Changed to getRuntimeAllg, open to alternatives.

PR #1833, opened 2020-01-16
pkg/proc: Remove disasm based prologue detection

derekparker:
Go started using the prologue_end DWARF attr in v1.11. Since that is now
the minimum supported Go version (Travis only tests 1.11 and above)
let's remove the legacy code which tries to disassemble the function to
find the end of the prologue.


derekparker:
Wishful thinking... I forgot this was only implemented for x86.

PR #1834, opened 2020-01-17
pkg/proc: Introduce Target and remove CommonProcess

derekparker:
I've distilled a lot of the conversation that came about the large refactor I submitted and have found a path to perform a more measured refactor. This patch set serves as part 1 of that refactor.

Introduce a new type `proc.Target` which will build on top of the various backends to remove complexity and duplication. The first step is removing `proc.CommonProcess` and pushing all that functionality further up the stack to `proc.Target`.

### DIFF
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -132,7 +132,7 @@ func (callCtx *callContext) doReturn(ret *Variable, err error) {
 // EvalExpression, EvalExpressionWithCalls is not a method of EvalScope.
 func EvalExpressionWithCalls(p Process, g *G, expr string, retLoadCfg LoadConfig, checkEscape bool) error {
 	bi := p.BinInfo()
-	if !p.Common().fncallEnabled {
+	if ok, _ := p.Recorded(); ok {
 		return errFuncCallUnsupportedBackend
 	}
 
@@ -234,7 +234,7 @@ func evalFunctionCall(scope *EvalScope, node *ast.CallExpr) (*Variable, error) {
 
 	p := scope.callCtx.p
 	bi := scope.BinInfo
-	if !p.Common().fncallEnabled {
+	if ok, _ := p.Recorded(); ok {
 		return nil, errFuncCallUnsupportedBackend
 	}
 
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -90,7 +90,7 @@ type functionCallState struct {
 }
 
 type callContext struct {
-	p Process
+	p *Target
 
 	// checkEscape is true if the escape check should be performed.
 	// See service/api.DebuggerCommand.UnsafeCall in service/api/types.go.
@@ -140,7 +140,7 @@ func (callCtx *callContext) doReturn(ret *Variable, err error) {
 // EvalExpression, EvalExpressionWithCalls is not a method of EvalScope.
 func EvalExpressionWithCalls(t *Target, g *G, expr string, retLoadCfg LoadConfig, checkEscape bool) error {
 	bi := t.BinInfo()
-	if ok, _ := t.Recorded(); ok {
+	if !t.SupportsFunctionCalls() {
 		return errFuncCallUnsupportedBackend
 	}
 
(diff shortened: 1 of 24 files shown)


pkg/proc/fncall.go
```diff
@@ -130,9 +138,9 @@ func (callCtx *callContext) doReturn(ret *Variable, err error) {
 // EvalExpressionWithCalls is like EvalExpression but allows function calls in 'expr'.
 // Because this can only be done in the current goroutine, unlike
 // EvalExpression, EvalExpressionWithCalls is not a method of EvalScope.
-func EvalExpressionWithCalls(p Process, g *G, expr string, retLoadCfg LoadConfig, checkEscape bool) error {
-	bi := p.BinInfo()
-	if !p.Common().fncallEnabled {
+func EvalExpressionWithCalls(t *Target, g *G, expr string, retLoadCfg LoadConfig, checkEscape bool) error {
+	bi := t.BinInfo()
+	if ok, _ := t.Recorded(); ok {
```
aarzilli: I think we should keep fncallEnabled, even though we are currently badly failing at setting the correct value for it (it should be false for rr and for arm64). For two reasons:

1. Function calls can be disabled for backends other than recordings (core files, the arm64 backend)
2. It should be actually possible to make function calls on the 'rr' backend through a facility called 'excursion', it isn't properly documented and I couldn't get it to work, but it exists.

I think it can be a parameter to NewTarget.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -885,7 +841,6 @@ func (p *Process) Restart(pos string) error {
 
 	p.exited = false
 
-	p.common.ClearAllGCache()
```
aarzilli: I think this is needed, in some form.

pkg/proc/threads.go
```diff
@@ -20,7 +20,7 @@ type Thread interface {
 	Location() (*Location, error)
 	// Breakpoint will return the breakpoint that this thread is stopped at or
 	// nil if the thread is not stopped at any breakpoint.
-	Breakpoint() BreakpointState
+	Breakpoint() *BreakpointState
```
aarzilli: We can change it but I liked that it wasn't a pointer because it implied that it can't return nil.

pkg/proc/fncall.go
```diff
@@ -234,7 +242,7 @@ func evalFunctionCall(scope *EvalScope, node *ast.CallExpr) (*Variable, error) {
 
 	p := scope.callCtx.p
 	bi := scope.BinInfo
-	if !p.Common().fncallEnabled {
+	if ok, _ := p.Recorded(); ok {
```
aarzilli: Same as other comment about fncallEnabled.

pkg/proc/gdbserial/gdbserver.go
derekparker: Added a `Restart` method to `(*Target)` which clears the cache and calls `Process.Restart`.

pkg/proc/threads.go
derekparker: Had to change because all methods worked on a pointer receiver and the compiler complained about taking the address of it once things were moved around.

pkg/proc/fncall.go
derekparker: Changed.

pkg/proc/gdbserial/gdbserver.go
aarzilli: What do you think about separating the interface that the backend provides to pkg/proc from the interface that can be called outside of proc, like this: https://github.com/aarzilli/delve/commit/2a2a54467899ca116fef2158679eda29ceaa28b8

ContinueOnce, SetSelectedGoroutine and methods for writing and removing breakpoints would also go inside proc.ProcessInternal.

derekparker: That's an interesting thought, but I want to see how a few more iterations play out first, because I think the embedding of `Process` into `Target` might end up being temporary.

How do you feel about merging as is and revisiting this suggestion?

aarzilli: Sure, go ahead.

PR #1836, opened 2020-01-18
proc,service: expose goroutine pprof labels in api

### LINKED ISSUE #1763: Feature: to expose goroutine labels in API
It is possible to [set pprof labels for a goroutine](https://tip.golang.org/pkg/runtime/pprof/#Do). It would be great to have access to these labels via API. They can make it easier to find a goroutine without looking into its stack.

nd:
Labels can help in identifying a particular goroutine during debugging.

Fixes #1763

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -210,6 +210,8 @@ type G struct {
 	variable *Variable
 
 	Unreadable error // could not read the G struct
+
+	Labels map[string]string // G's pprof labels
 }
 
 // Defer returns the top-most defer of the goroutine.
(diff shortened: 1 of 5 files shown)


pkg/proc/variables.go
```diff
@@ -558,6 +560,27 @@ func (v *Variable) parseG() (*G, error) {
 
 	status, _ := constant.Int64Val(v.fieldVariable("atomicstatus").Value)
 	f, l, fn := v.bi.PCToLine(uint64(pc))
+
```
aarzilli: GetG, and consequently parseG, is called quite a lot. I would prefer if parseG only saved the value of labelsVar and then a method of *G did the rest of the work and cached the result.

nd: Does it make sense to reuse the variable field and call `loadFieldNamed("labels")` in the method of *G?

aarzilli: Yes, that's fine.

aarzilli:
LGTM

PR #1841, opened 2020-01-20
cmd: support `LookPath` when debugging on windows.

chainhelen:
Add suffix `.exe` when building executable on windows.

Fix go-delve#505

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -211,7 +211,7 @@ Alternatively you can specify a package name, and Delve will debug the tests in
 that package instead.`,
 		Run: testCmd,
 	}
-	testCommand.Flags().String("output", "debug.test", "Output path for the binary.")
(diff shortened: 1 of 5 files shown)


cmd/dlv/cmds/commands.go
```diff
@@ -347,6 +347,9 @@ func debugCmd(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(os.Stderr, "%v\n", err)
 			return 1
 		}
+		if runtime.GOOS == "windows" {
+			debugname += ".exe"
```
aarzilli: I fear this would break backwards compatiblity. I don't know if anyone is using the output option but if they are we'll break them if we start changing the name of the file they are passing. If the problem is with the default value of output why not just change that?

chainhelen: Yes, sorry for this pr. 
Some infomation about the processor of `go build` in my mind is wrong, I have tested some cases on windows.

Have no idea about `LookPath`, but this pr can be closed

PR #1846, opened 2020-01-25
Reduce latency of conditional breakpoints/stop-resume cycles {WIP]

aarzilli:
```
proc: optimize parseG

runtime.g is a large and growing struct, we only need a few fields.
Instead of using loadValue to load the full contents of g cache its
memory and then only load the fields we care about.

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        10013510647 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	9330025748 ns/op

Conditional breakpoint evaluation: 1.0ms -> 0.93ms

Updates #1549

proc: optimize RegistersToDwarfRegisters

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        11570508729 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	10013510647 ns/op

Conditional breakpoint evaluation 1.2ms -> 1.0ms

Updates #1549

proc: cache result of GetG

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        15929810602 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	11570508729 ns/op

Conditional breakpoint evaluation 1.6ms -> 1.2ms

Updates #1549

proc,proc/*: move SelectedGoroutine to proc.Target

moves SelectedGoroutine, SwitchThread and SwitchGoroutine to
proc.Target, merges PostInitializationSetup with NewTarget.

proc/native: optimize native.status through buffering

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        17294564246 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	15929810602 ns/op

Conditional breakpoint evaluation 1.7ms -> 1.6ms

Updates #1549

proc: do not load g0 until it's needed when stacktracing

The stacktrace code occasionally needs the value of g.m.g0.sched.sp to
switch stacks. Since this is only needed rarely and calling parseG is
relatively expensive we should delay doing it until we know it will be
needed.

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        18397062447 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	17294564246 ns/op

i.e. this commit improves the time needed to evaluate a conditional
breakpoint from 1.8ms to 1.7ms.

Updates #1549

proc: only format registers value when it's necessary

Reduces the minimum amount of time needed to stop and resume a process
that hits a breakpoint.
A significant amount of time is spent generating the strings for the
proc.Registers object of each thread, since this field is rarely used
(only when the Registers API is called) it should be generated on
demand.

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	22218288218 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	18397062447 ns/op

i.e. this commit reduces the time spent stopping and resuming a process
from 2.2ms to 1.8ms.

Updates #1549

dwarf/reader: precalcStack does not need to read past the first entry

It was reading all the way to the end of the debug_info section,
slowing down stacktraces substantially.

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	80344642562 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	22218288218 ns/op

i.e. a reduction of the cost of a breakpoint hit from 8ms to 2.2ms

Updates #1549

tests: add benchmark for conditional breakpoints

proc: remove CX method from proc.Registers

It is not used anymore besides internally by the proc/gdbserial
backend.

```

### DIFF
--- a/_fixtures/issue1549.go
+++ b/_fixtures/issue1549.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+func main() {
+	sum := int64(0)
+	start := time.Now()
+	for value := int64(0); value < 10000; value++ {
+		sum += value
+	}
+	elapsed := time.Since(start)
+	fmt.Printf("Sum: %d\nTook %s\n", sum, elapsed)
+}
--- a/pkg/dwarf/reader/reader.go
+++ b/pkg/dwarf/reader/reader.go
@@ -426,6 +426,9 @@ childLoop:
 		default:
 			irdr.reader.SkipChildren()
 		}
+		if rentry == nil {
+			break
+		}
 	}
 
 	if rentry != nil && rentry.Tag == dwarf.TagInlinedSubroutine {
--- a/pkg/proc/amd64_arch.go
+++ b/pkg/proc/amd64_arch.go
@@ -1,7 +1,11 @@
 package proc
 
 import (
+	"bytes"
 	"encoding/binary"
+	"fmt"
+	"math"
+	"os"
 	"strings"
 
 	"github.com/go-delve/delve/pkg/dwarf/frame"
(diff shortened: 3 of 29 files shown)


aarzilli:
Closing this, I've been submitting it piecemeal anyway.

PR #1848, opened 2020-01-25
pkg: fix typo in the comment of `PtraceGetFpRegset`

chainhelen:


hengwu0:
Yes, Fp registers and Fp instructions of arm64 is still on working, and the code here is useless. Which traced in #1770.
If it confused you, it's ok to fix here.
LGTM

chainhelen:
Oh, I see. 
Thx for your explantation, I compare the difference of implementation about arm64 and amd64 indeed.

PR #1851, opened 2020-01-27
Fix serious performance degradation when building stack traces

aarzilli:
```
dwarf/reader: precalcStack does not need to read past the first entry

It was reading all the way to the end of the debug_info section,
slowing down stacktraces substantially.

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	80344642562 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	22218288218 ns/op

i.e. a reduction of the cost of a breakpoint hit from 8ms to 2.2ms

Updates #1549

tests: add benchmark for conditional breakpoints

proc: remove CX method from proc.Registers

It is not used anymore besides internally by the proc/gdbserial
backend.

```

### DIFF
--- a/_fixtures/issue1549.go
+++ b/_fixtures/issue1549.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+func main() {
+	sum := int64(0)
+	start := time.Now()
+	for value := int64(0); value < 10000; value++ {
+		sum += value
+	}
+	elapsed := time.Since(start)
+	fmt.Printf("Sum: %d\nTook %s\n", sum, elapsed)
+}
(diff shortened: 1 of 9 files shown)


aarzilli:
This PR adds a benchmark for conditional breakpoints and adds a serious performance degradation to stacktraces introduced by commit adb1746c6019cd2ac40e37bb00d71647245268a2.

This part of a [larger series of optimizations](https://github.com/go-delve/delve/pull/1846) that reduces the latency of a conditional breakpoint evaluation from 8ms to 0.9ms. The majority of the latency is caused by this particular bug (from 8ms down to 2.2ms).

PR #1854, opened 2020-01-30
[WIP] pkg/proc: Move breakpoint logic up to Target

derekparker:
This patch moves all breakpoint handling from the various backends
up to Target, removing a bunch of duplication.

When doing this I ran into an issue with Windows. For some reason I
don't fully understand, unlike the other backends, if you blindly
iterate over each thread and call SetCurrentBreakpoint it breaks Windows
in unexpected ways which manifests itself in the `StepConcurrent` tests.
This is resolved by returning not only the initial thread that we first
received the trap from, but all other threads that may have hit a
breakpoint as well. And then we only call SetCurrentBreakpoints on those
threads.

### DIFF
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -390,6 +390,11 @@ func (p *Process) ContinueOnce() (proc.Thread, []proc.Thread, error) {
 	return nil, nil, ErrContinueCore
 }
 
+// CurrentDirection always returns forward for non-recorded processes.
+func (p *Process) CurrentDirection() proc.Direction {
+	return proc.Forward
+}
+
 // StepInstruction will always return an error
 // as you cannot control execution of a core file.
 func (p *Process) StepInstruction() error {
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -220,7 +220,7 @@ func OpenCore(corePath, exePath string, debugInfoDirs []string) (*proc.Target, e
 // initialize for core files doesn't do much
 // aside from call the post initialization setup.
 func (p *Process) initialize(path string, debugInfoDirs []string) error {
-	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.WriteBreakpointFn)
+	return proc.PostInitializationSetup(p, path, debugInfoDirs)
 }
 
 // AdjustsPCAfterBreakpoint always returns false as the core backend does
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -392,8 +392,8 @@ func (p *Process) ClearInternalBreakpoints() error {
 
 // ContinueOnce will always return an error because you
 // cannot control execution of a core file.
-func (p *Process) ContinueOnce() (proc.Thread, error) {
-	return nil, ErrContinueCore
+func (p *Process) ContinueOnce() (proc.Thread, []proc.Thread, error) {
+	return nil, nil, ErrContinueCore
 }
 
 // StepInstruction will always return an error
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -223,9 +223,9 @@ func (p *Process) initialize(path string, debugInfoDirs []string) error {
 	return proc.PostInitializationSetup(p, path, debugInfoDirs)
 }
 
-// AdjustsPCAfterBreakpoint always returns false as the core backend does
+// AdjustPCAfterBreakpoint always returns false as the core backend does
 // not execute instructions or hit breakpoints.
-func (dbp *Process) AdjustsPCAfterBreakpoint() bool { return false }
+func (dbp *Process) AdjustPCAfterBreakpoint() bool { return false }
 
 // BinInfo will return the binary info.
 func (p *Process) BinInfo() *proc.BinaryInfo {
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -222,7 +222,7 @@ func OpenCore(corePath, exePath string, debugInfoDirs []string) (*proc.Target, e
 // initialize for core files doesn't do much
 // aside from call the post initialization setup.
 func (p *Process) initialize(path string, debugInfoDirs []string) error {
-	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.writeBreakpoint)
+	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.WriteBreakpointFn)
 }
 
 // BinInfo will return the binary info.
@@ -240,12 +240,6 @@ func (p *Process) SetSelectedGoroutine(g *proc.G) {
 // EntryPoint will return the entry point address for this core file.
 func (p *Process) EntryPoint() (uint64, error) {
 	return p.entryPoint, nil
-}
-
-// writeBreakpoint is a noop function since you
-// cannot write breakpoints into core files.
-func (p *Process) writeBreakpoint(addr uint64) (file string, line int, fn *proc.Function, originalData []byte, err error) {
-	return "", 0, nil, nil, errors.New("cannot write a breakpoint to a core file")
 }
 
 // Recorded returns whether this is a live or recorded process. Always returns true for core files.
@@ -452,6 +446,14 @@ func (p *Process) SelectedGoroutine() *proc.G {
 // SetBreakpoint will always return an error for core files as you cannot write memory or control execution.
 func (p *Process) SetBreakpoint(addr uint64, kind proc.BreakpointKind, cond ast.Expr) (*proc.Breakpoint, error) {
 	return nil, ErrWriteCore
+}
+
+func (p *Process) WriteBreakpointFn(addr uint64) (string, int, *proc.Function, []byte, error) {
+	return "", 0, nil, nil, ErrWriteCore
+}
+
+func (p *Process) ClearBreakpointFn(uint64, []byte) error {
+	return ErrWriteCore
 }
 
 // SwitchGoroutine will change the selected and active goroutine.
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -3,7 +3,6 @@ package core
 import (
 	"errors"
 	"fmt"
-	"go/ast"
 	"io"
 
 	"github.com/go-delve/delve/pkg/proc"
@@ -156,7 +155,6 @@ type Process struct {
 	entryPoint uint64
 
 	bi                *proc.BinaryInfo
-	breakpoints       proc.BreakpointMap
 	currentThread     *Thread
 	selectedGoroutine *proc.G
 }
@@ -228,10 +226,6 @@ func (p *Process) initialize(path string, debugInfoDirs []string) error {
 // AdjustsPCAfterBreakpoint always returns false as the core backend does
 // not execute instructions or hit breakpoints.
 func (dbp *Process) AdjustsPCAfterBreakpoint() bool { return false }
-
-func (dbp *Process) FindBreakpoint(pc uint64) (*proc.Breakpoint, bool) {
-	return nil, false
-}
 
 // BinInfo will return the binary info.
 func (p *Process) BinInfo() *proc.BinaryInfo {
@@ -375,23 +369,6 @@ func (t *Thread) SetDX(uint64) error {
 	return ErrChangeRegisterCore
 }
 
-// Breakpoints will return all breakpoints for the process.
-func (p *Process) Breakpoints() *proc.BreakpointMap {
-	return &p.breakpoints
-}
-
-// ClearBreakpoint will always return an error as you cannot set or clear
-// breakpoints on core files.
-func (p *Process) ClearBreakpoint(addr uint64) (*proc.Breakpoint, error) {
-	return nil, proc.NoBreakpointError{Addr: addr}
-}
-
-// ClearInternalBreakpointsInternal will always return nil and have no
-// effect since you cannot set breakpoints on core files.
-func (p *Process) ClearInternalBreakpointsInternal() error {
-	return nil
-}
-
 // ContinueOnce will always return an error because you
 // cannot control execution of a core file.
 func (p *Process) ContinueOnce() (proc.Thread, []proc.Thread, error) {
@@ -454,11 +431,6 @@ func (p *Process) ResumeNotify(chan<- struct{}) {
 // goroutine.
 func (p *Process) SelectedGoroutine() *proc.G {
 	return p.selectedGoroutine
-}
-
-// SetBreakpoint will always return an error for core files as you cannot write memory or control execution.
-func (p *Process) SetBreakpoint(addr uint64, kind proc.BreakpointKind, cond ast.Expr) (*proc.Breakpoint, error) {
-	return nil, ErrWriteCore
 }
 
 func (p *Process) WriteBreakpointFn(addr uint64) (string, int, *proc.Function, []byte, error) {
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -225,6 +225,14 @@ func (p *Process) initialize(path string, debugInfoDirs []string) error {
 	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.WriteBreakpointFn)
 }
 
+// AdjustsPCAfterBreakpoint always returns false as the core backend does
+// not execute instructions or hit breakpoints.
+func (dbp *Process) AdjustsPCAfterBreakpoint() bool { return false }
+
+func (dbp *Process) FindBreakpoint(pc uint64) (*proc.Breakpoint, bool) {
+	return nil, false
+}
+
 // BinInfo will return the binary info.
 func (p *Process) BinInfo() *proc.BinaryInfo {
 	return p.bi
@@ -378,9 +386,9 @@ func (p *Process) ClearBreakpoint(addr uint64) (*proc.Breakpoint, error) {
 	return nil, proc.NoBreakpointError{Addr: addr}
 }
 
-// ClearInternalBreakpoints will always return nil and have no
+// ClearInternalBreakpointsInternal will always return nil and have no
 // effect since you cannot set breakpoints on core files.
-func (p *Process) ClearInternalBreakpoints() error {
+func (p *Process) ClearInternalBreakpointsInternal() error {
 	return nil
 }
 
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -108,8 +108,6 @@ type Process struct {
 
 	manualStopRequested bool
 
-	breakpoints proc.BreakpointMap
-
 	gcmdok         bool   // true if the stub supports g and G commands
 	threadStopInfo bool   // true if the stub supports qThreadStopInfo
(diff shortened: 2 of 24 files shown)


derekparker:
@aarzilli this is ready for review.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -699,41 +689,42 @@ continueLoop:
...
+			err = errors.New("arithmetic exception")
+		case 0x94:
+			err = errors.New("emulation exception")
+		case 0x95:
+			err = errors.New("software exception")
+		case 0x96:
+			err = errors.New("breakpoint exception")
+		}
+		r := make([]proc.Thread, 0, len(p.threads))
+		for _, t := range p.threads {
```
aarzilli: I don't like that we are not using Thread.setbp here and instead just setting a breakpoint on every thread. I think the new logic should follow the old logic. Also the change to the loop looking for the thread doesn't have anything to do with this PR and I'd like to leave it out simply to minimize the patch size.

pkg/proc/core/core.go
```diff
@@ -222,9 +220,13 @@ func OpenCore(corePath, exePath string, debugInfoDirs []string) (*proc.Target, e
 // initialize for core files doesn't do much
 // aside from call the post initialization setup.
 func (p *Process) initialize(path string, debugInfoDirs []string) error {
-	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.writeBreakpoint)
+	return proc.PostInitializationSetup(p, path, debugInfoDirs, p.WriteBreakpointFn)
```
aarzilli: Since WriteBreakpointFn is now part of the Process interface there is no reason to pass it here.

pkg/proc/interface.go
```diff
@@ -102,8 +102,6 @@ type ProcessManipulation interface {
 
 // BreakpointManipulation is an interface for managing breakpoints.
 type BreakpointManipulation interface {
-	Breakpoints() *BreakpointMap
-	SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr) (*Breakpoint, error)
-	ClearBreakpoint(addr uint64) (*Breakpoint, error)
-	ClearInternalBreakpoints() error
+	WriteBreakpointFn(addr uint64) (string, int, *Function, []byte, error)
```
aarzilli: I'm going to insist that these internal functions are moved to an interface other than proc.Process that we don't embed inside Target. We had two "internal methods users of proc aren't supposed to call" before and now we have four.

pkg/proc/interface.go
```diff
@@ -70,6 +66,10 @@ type Info interface {
 	BinInfo() *BinaryInfo
 	EntryPoint() (uint64, error)
 
+	AdjustsPCAfterBreakpoint() bool
```
aarzilli: IMHO this is evidence that the code that this PR is unifying isn't actually common among different backends and shouldn't be unified.

I also don't think that having a variable called adjustPC and a method called AdjustsPCAfterBreakpoint have opposite meanings (AdjustsPCAfterBreakpoint returning true means we set adjustPC to false) is a good idea.

pkg/proc/target.go
```diff
@@ -1,28 +1,119 @@
...
 	fncallForG map[int]*callInjection
 
 	// gcache is a cache for Goroutines that we
 	// have read and parsed from the targets memory.
 	// This must be cleared whenever the target is resumed.
 	gcache goroutineCache
+
+	// threadToBreakpoint maps threads to the breakpoint that they
+	// have were trapped on.
+	threadToBreakpoint map[int]*BreakpointState
```
aarzilli: I don't like that a chunk of the thread state now goes into a map instead of being in its logical place (inside the thread object).

pkg/proc/proc.go
```diff
@@ -523,21 +534,34 @@ func StepInstruction(dbp *Target) (err error) {
...
 
+func (dbp *Target) threadStepInstruction(thread Thread, setbp bool) error {
+	bp := dbp.ThreadToBreakpoint(thread)
+	if bp.Breakpoint != nil {
+		if err := dbp.Process.ClearBreakpointFn(bp.Addr, bp.OriginalData); err != nil {
+			return err
+		}
+		defer dbp.Process.WriteBreakpointFn(bp.Addr)
+	}
+	if err := thread.StepInstruction(); err != nil {
```
aarzilli: The way this works out is that it becomes far more inefficient in the gdbserial backend, the original code called the internal gdbserial.(*Thread).stepInstruction, which did not reload the thread registers (because we don't need to know in this situation), while this calls StepInstruction which will reload the threads.

aarzilli:
As an alternative, most of the (actually) duplicated code between backends is the FindBreakpoint and SetCurrentBreakpoint methods. If we added a single method:

```
func (bpmap *BreakpointMap) EvalBreakpointForThread(th Thread, adjustPC bool) BreakpointState
```

then both FindBreakpoint and SetCurrentBreakpoint could be removed and calls to SetCurrentBreakpoint could be replaced with:

```
th.CurrentBreakpoint = p.Breakpoints().EvalBreakpointForThread(th, adjustPC)
```

This would remove the duplicate code, cause less disruption and leave the breakpoint state where it belongs.

pkg/proc/core/core.go
derekparker: Done.

pkg/proc/gdbserial/gdbserver.go
derekparker: I've added back in `Thread.setbp` in the logic of which threads to return. The change from looping and using the map directly was because the exit condition changed and I thought it clearer to check the map for the thread and then iterate the list of threads instead of having nested loops.

pkg/proc/interface.go
derekparker: I disagree, I don't think it's a problem to "ask" the backend something, this is a simple predicate method. I don't think it's worth duplication of code just to prevent asking a single question to determine whether we should do some arithmetic on a pointer.

pkg/proc/interface.go
derekparker: I'm fine with that as long as we agree it's temporary. I think in the end all backends should be in `pkg/proc/internal/...` so they are completely unexposed to anything outside. Everything should use `Target` once this refactor is done, and the backends should not leak out as they simply aren't safe to use by just anything.

pkg/proc/proc.go
derekparker: Perhaps that's a sign the method is doing too much. Why should `StepInstruction` have a side effect of reloading registers? I think a more proper place for that would be in the first call to `Thread.Registers` and then the result is memoized. Essentially calling and caching on demand instead of as a result of calling `StepInstruction`.

pkg/proc/target.go
derekparker: Honestly I could go either way on this.

My thoughts are removing it from the Thread object(s) has a few benefits:

1. Remove redundancy (not a huge win overall, I admit)
2. We keep better track of the state this way. Instead of always iterating every thread and asking it if it's stopped at a breakpoint, we have a definitive mapping right here.
3. Easier to cleanup and manage; this is ephemeral state, not necessarily an actual piece of information on an operating system thread. We can easily cleanup this state by wiping it out and creating a new map.
4. The backend(s) no longer care about breakpoint state for a thread, so we'd be storing this information in the backend solely for the use of the consumer, even though the consumer itself (`Target`) is using and managing the information.

derekparker:
> In general I do not like this change. I think that its expanding too much the set of methods of Process that we expose outside proc despite the fact that they can not be called safely by anyone outside of proc, it touches a lot of sensitive code in every backend and moves a bunch of state that belongs inside the thread object outside of it.

Just wanted to respond to this directly, as I didn't when I pushed my changes the other day. The exposing of a subset of the Process methods is only temporary. Ideally nothing else aside from `Target` should even know about `Process`, with all backends being "hidden" via `pkg/proc/internal/...` directory hierarchy.

The touching of sensitive code in the backends also speaks towards the need for this refactor. We shouldn't have a bunch of sensitive code that we don't have confidence in changing spread across multiple backends and implementations. Any "tricky" code should exist only one single place so it's done right once and never duplicated.

As for the moving of thread state to another location, again I don't feel particularly strongly about that, but I did voice some of the reasoning above.

pkg/proc/proc.go
aarzilli: > Perhaps that's a sign the method is doing too much

You are not wrong,  but the gdbserial backend was designed to always keep the regs field of Thread always updated. I did that because it was simpler but doing it when it's needed and memoizing is definitely the way forward.

However it's a rather complicated change so it should be done in a separate PR, rather than as part of an unrelated refactor. And either that goes first or this PR goes in without moving the thread resume logic out of the backends.

pkg/proc/target.go
aarzilli: Responding point to point:

1. Agree, not much of a win given that it's just a field and a getter
2. There are two places in this patch where Target.threadToBreakpoint is iterated, by contrast there are four places where we have a Thread and we want to get its breakpoint
3. Of the two places where we iterate threadToBreakpoint one is the thread resume logic where we do not want to just wipe the map because each thread needs to be single stepped, so in 50% of the cases the easy cleanup is not a win
4. In principle, I'd rather the backend knew. Even if it doesn't need to know now it might in the future and there's no reason to erect artificial barriers to it

derekparker: I agree with your points, there is one more things to consider though. Once all backends are in `pkg/internal`, we will have to initialize them via `Target`, which means `pkg/proc` will be importing the backends instead of the other way around.

If both `pkg/proc` and the various backends know about breakpoints we get into a cyclic dependency issue.

pkg/proc/proc.go
derekparker: Agreed.

pkg/proc/interface.go
derekparker: I'm voting for leaving it as a method for now, unless you really object. I think it's a cleaner solution instead of adding another config param to NewTarget.

I'm open to other names for this method as well. My thought process on the current name is that we are asking the backend if it adjusts the pc after hitting a breakpoint, so that's why I phrased it as I did. Setting `adjustPC` to true when the method call is false is saying "we must adjust the PC if the backend does not".

pkg/proc/target.go
derekparker: @aarzilli I think this is the only outstanding item of contention.

aarzilli: Either a third package will initialize them or the initialization will keep happening as it does now and their path won't have 'internal' in it. What does putting them into an internal package get us?

At the moment proc can call anything in a backend and the backend can call anything in proc. I'd like that to continue to be true.

pkg/proc/interface.go
aarzilli: Let's call it AdjustPC and have it match the direction of adjustPC.

pkg/proc/interface.go
```diff
@@ -14,10 +10,21 @@ import (
...
+// ProcessInternal holds a set of methods that are
+// not meant to be called by anyone except for an instance of
+// `proc.Target`. These methods are not safe to use by themselves
+// and should never be called directly outside of the `proc` package.
+// This is temporary and in support of an ongoing refactor.
+type ProcessInternal interface {
+	WriteBreakpointFn(addr uint64) (string, int, *Function, []byte, error)
+	ClearBreakpointFn(uint64, []byte) error
+	AdjustsPCAfterBreakpoint() bool
+	CurrentDirection() Direction
```
aarzilli: Restart, SetSelectedGoroutine, ContinueOnce and Detach should also be moved here.

pkg/proc/target.go
derekparker: The backends aren't safe to use by themselves, so I'd prefer if nothing outside of proc is able to import them.

derekparker: In any event, we can go back and forth on this point at a later time in the refactor. I'm working on reverting this so "current breakpoint" state stays on threads for now.

aarzilli: > The backends aren't safe to use by themselves, so I'd prefer if nothing outside of proc is able to import them

If you look at what the backends actually export right now, almost everything you can call can be safely called by anyone, because the only things you can call are the launch and attach functions, which return Target objects. The exception here are the Ptrace* functions in the native package that are exported for no reason at all.

aarzilli: They also all export the Process and Thread concrete types but there's no way to create them and we could very easily un-export them.

derekparker: @aarzilli I agree that we should likely un-export the Process and Thread types in the backends, and anything else that doesn't need to be exported. I guess my position on it is, we don't expect or want anything outside of proc to import the backends directly, so why not enforce that using an internal import path? Right now the only reason why the backends are being imported is, as you mentioned, the Launch, Attach, etc functions. To me though that feels backwards as well, that was something I relented on during initial refactor discussions but still feel can be done better. I don't think the backends returning `proc.Target` is the right thing. I still prefer the more generic Launch / Attach methods that were in the original refactor sketch, something like `proc.Launch` instead of the debugger package making the determination on which backend to use. 

Ideally I see it as:

1. Debugger (or test code) wants to start a process
2. Debugger (or test code) asks proc package to Launch or Attach
3. proc package handles figuring out which backend to use and initialize it
4 proc package returns its own `proc.Target` object, which holds an instance of the selected backend.

This ends up looking like:

```
debugger --> proc.target --> { native, gdbserial, core }
```

Instead of:

```
debugger --> { native, gdbserial, core } --> proc.target
```

Either way, the only relevance that change has to this current patch is handling the cyclic dependency problem. If we wait on that we can discuss and solve this at another time.

I regret that I didn't first draft a design document to discuss the scope of this refactor and the desired end result. I think that would have helped guide the discussion a lot better. It might still be helpful, now that I think about it.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1503,23 +1407,6 @@ func (t *Thread) reloadGAtPC() error {
 	cx := t.regs.CX()
 	pc := t.regs.PC()
 
-	// We are partially replicating the code of GdbserverThread.stepInstruction
```
aarzilli: This code shouldn't be removed, removing it doesn't break debugserver on macOS because we can allocate memory on the target process but it will break lldb-server on linux and the spec is not clear on what would happen in this circumstance anyway.

pkg/proc/target.go
aarzilli: > I guess my position on it is, we don't expect or want anything outside of proc to import the backends directly, so why not enforce that using an internal import path? Right now the only reason why the backends are being imported is, as you mentioned, the Launch, Attach, etc functions. To me though that feels backwards as well

I don't agree with this. Yes, we don't expect anything outside of *debugger* to import the backends, we also do not expect anything outside of debugger to import proc itself. And we don't expect anything outside of service to import debugger. 

And I don't think having the constructors in the backend is backwards: you want to use the native backend, you import the native backend. If having the switch statement duplicated in debugger and proc_test we can use a mechanism like database/sql with the init function of the backends calling proc.Register.

aarzilli: > calling proc.Register

It does occour to me now that proc.Register is already take, but you get the point.

Also, I'm not in principle opposed to proc.Launch/proc.Attach it's just that I don't see it as a big win and having the backends be unable to see inside pkg/proc is too great of a downside.

derekparker: > > I guess my position on it is, we don't expect or want anything outside of proc to import the backends directly, so why not enforce that using an internal import path? Right now the only reason why the backends are being imported is, as you mentioned, the Launch, Attach, etc functions. To me though that feels backwards as well
> 
> I don't agree with this. Yes, we don't expect anything outside of _debugger_ to import the backends, we also do not expect anything outside of debugger to import proc itself. And we don't expect anything outside of service to import debugger.

I should have been more clear, I was phrasing that based on future work, if we were to invert the initialization (using only proc, not the various backends). As of now, you're correct, nothing except `debugger` (and tests) should be importing the backends.

pkg/proc/gdbserial/gdbserver.go
derekparker: Is this due to lldb-server thinking that there is still a breakpoint (or breakpoints) present in this address range and changing the behavior of `step`?

derekparker: Looking into this more, is there really no way to get the value of `fs_base` out of LLDB/GDB? I know neither return the value in the registers command (CLI or wire protocol) but in GDB you can `print $fs_base`, so it has the information somewhere.

I'd like to get rid of having to special case this backend in terms of how we read into thread local storage to get the `*G`. I've done some digging in the documentation but wasn't able to find anything, however I didn't go too deep. I work with some GDB / LLDB maintainers so I think I might ping them on if there is a better way to get that value.

aarzilli: > Looking into this more, is there really no way to get the value of fs_base out of LLDB/GDB?

GDB, yes it is actually in returned by the 'g' command (but this code doesn't work with gdb for other reasons). LLDB, no, as far as I know.

aarzilli: PS. this would be a behaviour change so a refactoring wouldn't be the place to do it anyway.

derekparker: @aarzilli of course, I was planning on investigating this and then submitting as a separate patch.

derekparker: Coming back to this, I’m not sure I actually fully understand why this is an issue. Essentially we are clearing the breakpoints present in `[pc, pc+len(movinstr)]`, however we overwrite that memory anyways, meaning there will be no breakpoint instruction there. What’s the bug, exactly, that this prevents? Will lldb-server on linux complain and prevent us from writing to this memory if it knows there should be a breakpoint there?

Also, it seems lldb-server support on Linux is broken currently in master anyways (just ran `go test -v -backend=lldb` on my Fedora machine with failures) and we do not test it in Travis, so who knows for how long it's been broken. I will try and bisect to figure out what broke it, but regardless it is currently broken.

(On this note, I wish we could also test RR via CI, but Travis runs on GCE which does not support the counters RR needs).

derekparker: I actually just tried to bisect the lldb-server failures on linux, but I'm getting failures as far back as `v1.2.0`. 

Go v1.13
lldb-server v9.0.1

aarzilli: I only have v6.0.0 installed, that one works. Regardless, this code has been here since the beginning, and the other branch depends on having memory allocation available, we don't know if that's available on every possible combination of lldb-server, debugserver and version of macOS so I wouldn't remove it.

In general, I'm not a fan of removing visibility into the list of breakpoints from the backends. It breaks this but this is just an example of things that we can't do if the backends can't se the breakpoint map. I like moving the breakpoint map into the target object but I'd also like to save a pointer to the target object inside gdbserial.Process and native.Process, so that we can access the state of Target from the backends when we need to.

derekparker:
Marking this as WIP for now. I'm most likely going to just harvest this PR for code and submit a few other PRs individually so that this doesn't continue to rot. I also want to unblock other PRs that are on hold due to changes to breakpoint handling.

derekparker:
Closing this due to age and rot. Will try another approach.

PR #1857, opened 2020-02-11
proc/gdbserial: Reload thread registers on demand

derekparker:
Instead of reloading the registers for every thread every time the
process executes, reload the registers on demand for individual threads
and memoize the result.


derekparker:
@aarzilli done.

Issue #1515, opened 2019-03-18
Debug Adapter Protocol


dsvensson:
Not sure if it would be better to have something like this external to Delve, but a sibling project to the Language Server Protocol is starting to take form, in Debug Adapter Protocol. This is something that is rapidly gaining support in Emacs, but most servers are currently ripped out from Visual Studio Code plugins. There's currently a wrapper around delve for making it conform to DAP in Visual Studio Code. It's written in TypeScript, so it adds an extra universe of dependencies. It would be nice to cut a few layers for a more lean integration.

Emacs dap-mode:
https://github.com/emacs-lsp/dap-mode

VSCode Delve wrapper, also used by Emacs dap-mode:
https://github.com/Microsoft/vscode-go/tree/master/src/debugAdapter

Debug Adapter Protocol:
https://microsoft.github.io/debug-adapter-protocol/

aarzilli:
If you want to implement this you can take a look at https://github.com/go-delve/delve/blob/master/service/rpc2/server.go. It would just be a matter of implementing the protocol and hooking it up to a command line command in https://github.com/go-delve/delve/blob/master/cmd/dlv/cmds/commands.go

dsvensson:
Yep, started looking in that direction. Would you consider this a separate project or an addition to delve, like `cmd/dap` or something?

aarzilli:
I think it would be fine to have a subcommand, like `dlv debugger-adapter-server`.

eliben:
@aarzilli a potential issue with a new command like `dlv debugger-adapter-server` is that currently delve can accept several commands in "server mode", like `dlv exec` and `dlv debug` with `--headless` (and also derivative commands like `attach`, and probably more). It'd be good to be able to do all these in DAP mode too.

It almost seems like DAP could be a new "API version", like: `dlv exec --headless --api-version dap`, but API version is an int flag. So a few options:

1. We could change the flag type to be string and accept "dap" in addition to "1" and "2"
2. We could use some numeric API version for DAP, e.g. version 3 is DAP, or version 30, etc.
3. We could just use a different flag, like `--dap`. This flag would be mutually exclusive with `--api-version` and would only work in `--headless` mode.

What do you think?

aarzilli:
[...]

PR #1858, opened 2020-02-11
service/dap: Initial implementation for 'dlv dap'

polinasok:
This change adds an initial synchronous implementation for running delve over DAP that supports a handful of requests (initialize, launch, set breakpoint, continue, disconnect). This min viable version is lacking support for many requests and other features (see comments in the code), but this should be enough to agree on the overall approach before diving into more details. (I do plan to add more tests to service_test and dlv_test in future PRs.)

Updates #1515

### DIFF
--- a/_fixtures/helloworld.go
+++ b/_fixtures/helloworld.go
@@ -1,0 +1,9 @@
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
+	fmt.Println("HELLO WORLD!")
+}
--- a/_fixtures/helloworld.go
+++ b/_fixtures/helloworld.go
@@ -1,9 +1,0 @@
-package main
-
-import (
-	"fmt"
-)
-
-func main() {
-	fmt.Println("HELLO WORLD!")
-}
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -335,6 +335,7 @@ names selected from this list:
 	lldbout		Copy output from debugserver/lldb to standard output
 	debuglineerr	Log recoverable errors reading .debug_line
 	rpc		Log all RPC messages
+	dap		Log all DAP messages
 	fncall		Log function call protocol
 	minidump	Log minidump loading
 
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -19,6 +19,7 @@ import (
 	"github.com/go-delve/delve/pkg/version"
 	"github.com/go-delve/delve/service"
 	"github.com/go-delve/delve/service/api"
+	"github.com/go-delve/delve/service/dap"
 	"github.com/go-delve/delve/service/rpc2"
 	"github.com/go-delve/delve/service/rpccommon"
 	"github.com/spf13/cobra"
@@ -151,6 +152,22 @@ option to let the process continue or kill it.
 		Run: connectCmd,
 	}
 	RootCommand.AddCommand(connectCommand)
+
+	// 'dap' subcommand.
+	dapCommand := &cobra.Command{
+		Use:   "dap",
+		Short: "[EXPERIMENTAL] Starts a TCP server communicating via Debug Adaptor Protocol (DAP).",
+		Long: `[EXPERIMENTAL] Starts a TCP server communicating via Debug Adaptor Protocol (DAP).
+
+The server supports debugging of a precompiled binary akin to 'dlv exec' via a launch request.
+It does not yet support support specification of program arguments.
+It does not yet support launch requests with 'debug' and 'test' modes that require compilation.
+It does not yet support attach requests to debug a running process like with 'dlv attach'.
+It does not yet support asynchronous request-response communication.
+The server does not accept multiple client connections.`,
+		Run: dapCmd,
+	}
+	RootCommand.AddCommand(dapCommand)
 
 	// 'debug' subcommand.
 	debugCommand := &cobra.Command{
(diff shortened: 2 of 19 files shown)


polinasok:
Hmm, it appears that the tests are behaving differently here than they did for me locally. Investigating.

cmd/dlv/cmds/commands.go
```diff
@@ -344,6 +361,63 @@ func remove(path string) {
...
+		if len(dlvArgs) > 0 {
+			fmt.Fprintf(os.Stderr, "Warning: debug arguments ignored with dap; specify via launch/attach request instead\n")
+		}
+		if len(targetArgs) > 0 {
+			fmt.Fprintf(os.Stderr, "Warning: program flags ignored with dap; specify via launch/attach request instead\n")
+		}
+
+		listener, err := net.Listen("tcp", Addr)
+		if err != nil {
+			fmt.Printf("couldn't start listener: %s\n", err)
```
aarzilli: os.Stderr?

pkg/logflags/logflags.go
```diff
@@ -151,6 +171,8 @@ func Setup(logFlag bool, logstr string, logDest string) error {
 			debugLineErrors = true
 		case "rpc":
 			rpc = true
+		case "dap":
```
aarzilli: The values that we recognize here are documented inside cmd/dlv/cmds/commands.go (grep for "Help about logging flags"). Please add it there too.

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+	case *dap.BreakpointLocationsRequest:
+		s.sendUnsupportedErrorResponse(request.Request)
+	default:
+		// This is a DAP message that go-dap has a struct for, so
+		// decoding succeeded, but this function does not know how
+		// to handle. We should be sending an ErrorResponse, but
+		// we cannot get to Seq and other fields from dap.Message.
+		// TODO(polina): figure out how to handle this better.
+		// Consider adding GetSeq() method to dap.Message interface.
+		panic(fmt.Sprintf("Unable to process %#v", request))
```
aarzilli: I'd rather log an error than panic here.

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+				fmt.Println("DAP error:", err)
+			}
+			return
+		}
+		s.handleRequest(request)
+	}
+}
+
+func (s *Server) handleRequest(request dap.Message) {
+	jsonmsg, _ := json.Marshal(request)
```
aarzilli: I think that either this function or its caller should have some kind of panic guard, like service/rpccommon/server.go does (see newInternalError and its callers). The idea is that if we have a bug and we accidentally do an out of bounds slice access it's safer to recover from that panic and return it to the client (as an internal error) than to let the panic bring down Delve, because that will also kill the user's process, which is highly disruptive.

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+		CoreFile:             "",
+		Backend:              s.config.Backend,
+		Foreground:           s.config.Foreground,
+		DebugInfoDirectories: s.config.DebugInfoDirectories,
+		CheckGoVersion:       s.config.CheckGoVersion,
+	}
+	var err error
+	if s.debugger, err = debugger.New(config, s.config.ProcessArgs); err != nil {
+		s.sendErrorResponse(request.Request,
+			3000, "Failed to launch", err.Error())
```
aarzilli: 3000 should be a named constant.

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+	s.send(&dap.InitializedEvent{Event: *newEvent("initialized")})
+	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+}
+
+// onDisconnectRequest handles the DisconnectRequest. Per the DAP spec,
+// it disconnects the debuggee and signals that the debug adaptor
+// (in our case this TCP server) can be terminated.
+func (s *Server) onDisconnectRequest(request *dap.DisconnectRequest) {
+	s.send(&dap.DisconnectResponse{Response: *newResponse(request.Request)})
+	// TODO(polina): only halt if the program is running
```
aarzilli: I don't think it matters (but I haven't checked).

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+		err = s.debugger.Detach(kill)
+		if err != nil {
+			s.log.Error(err)
+		}
+	}
+	// TODO(polina): make thread-safe when handlers become asynchronous.
+	s.signalDisconnect()
+}
+
+func (s *Server) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
```
aarzilli: One thing to check is if VSCode sends this request while the program is running, if it does know that in that case Debugger.CreateBreakpoint will block until the program stops.

service/dap/server.go
```diff
@@ -0,0 +1,464 @@
...
+		default:
+		}
+		return
+	}
+	if state.Exited {
+		e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
+		s.send(e)
+	} else {
+		e := &dap.StoppedEvent{Event: *newEvent("stopped")}
+		e.Body.Reason = "breakpoint"
```
aarzilli: This could also be `pause` if a halt command was sent. I don't know if it makes a difference and it will probably be hard to make this distinction (so if it doesn't matter in practice it probably isn't worth bothering with this).

_fixtures/helloworld.go
```diff
@@ -0,0 +1,9 @@
+package main
```
aarzilli: We already have other fixtures that don't do a whole lot, in the interest of avoiding fixture proliferation I suggest using increment.go or testargs.go

service/dap/server.go
polinasok: There is such thing as a PauseRequest, but I have not implemented that in this PR, so no checking for this case here yet. I don't think I can tell from the debugger if this was a pause or breakpoint, can I? I guess I could keep track if this was called in a context of a breakpoint or a halt request.

service/dap/server.go
polinasok: I am going to add a TODO for now if that's OK. I have not yet implemented attach, so taking down the program as delve goes down comes with the limited experimental territory right now. I will need to figure out what returning the error to the client would mean here as the responses are always sent with the seq id of the request. And I can use the extra time to learn about this language feature (I am a go newbie).

cmd/dlv/cmds/commands.go
aarzilli: No, we should probably change the other one. Don't worry about it.

service/dap/server.go
aarzilli: Ok.

service/dap/server.go
aarzilli: > Would there be any issues with the debugger object if I were to call Debugger.CreateBreakpoint in one goroutine and while it was blocking another call to the debugger was made from another goroutine?

Yes, it would block until the other call to the debugger was finished. Debugger serializes all its calls with a mutex, with the exception of the Halt command and `State` with `nowait == true`.

service/dap/server.go
aarzilli: > I don't think I can tell from the debugger if this was a pause or breakpoint, can I? I guess I could keep track if this was called in a context of a breakpoint or a halt request.

Yes, you would have to keep track (and there would be a race condition if the user requested a pause just as the program hit a breakpoint).

service/dap/server.go
aarzilli: I think you should use the logger with loglevel Error (see https://github.com/go-delve/delve/pull/1864)

service/dap/server.go
```diff
@@ -0,0 +1,466 @@
...
+	// debugger is the underlying debugger service.
+	debugger *debugger.Debugger
+	// log is used for structured logging.
+	log *logrus.Entry
+	// stopOnEntry is set to automatically stop the debugee after start.
+	stopOnEntry bool
+}
+
+// NewServer creates a new DAP Server. It takes an opened Listener
+// via config and assumes its ownerhsip. Optinally takes DisconnectChan
```
derekparker: s/ownerhsip/ownership/

derekparker: s/Optinally/Optionally/

polinasok:
Thank you both for speedy code reviews and useful comments!

aarzilli:
You're welcome!

Issue #1838, opened 2020-01-19
'regs' does not show selected frame registers.


klauspost:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?

Latest master.
```
λ dlv version
Delve Debugger
Version: 1.3.2
Build: $Id: 569ccbd514fc47c8b4c521b142556867ec5e6917 $
```

2. What version of Go are you using? (`go version`)?

`go version go1.13.6 windows/amd64`

3. What operating system and processor architecture are you using?

windows/amd64

4. What did you do?

Reproduction sample: [repro.zip](https://github.com/go-delve/delve/files/4082975/repro.zip)

Run the sample. Crash occurs. We switch to the frame causing the crash (3). Print registers.
```
λ dlv test                                                                                                               
Type 'help' for list of commands.                                                                                        
(dlv) continue                                                                                                           
unexpected fault address 0xc000270398                                                                                    
fatal error: fault                                                                                                       
> [runtime-fatal-throw] runtime.fatalthrow() c:/go/src/runtime/panic.go:820 (hits goroutine(6):1 total:1) (PC: 0x4356f0) 
Warning: debugging optimized function                                                                                    
   815: // fatalthrow implements an unrecoverable runtime throw. It freezes the                                          
   816: // system, prints stack traces starting from its caller, and terminates the                                      
   817: // process.                                                                                                      
   818: //                                                                                                               
   819: //go:nosplit                                                                                                     
=> 820: func fatalthrow() {                                                                                              
   821:         pc := getcallerpc()                                                                                      
   822:         sp := getcallersp()                                                                                      
   823:         gp := getg()                                                                                             
   824:         // Switch to the system stack to avoid any stack growth, which                                           
   825:         // may make things worse if the runtime is in a bad state.                                               
(dlv) stack                                                                                                              
0  0x00000000004356f0 in runtime.fatalthrow                                                                              
   at c:/go/src/runtime/panic.go:820                                                                                     
1  0x0000000000435579 in runtime.throw                                                                                   
   at c:/go/src/runtime/panic.go:774                                                                                     
2  0x0000000000448463 in runtime.sigpanic                                                                                
   at c:/go/src/runtime/signal_windows.go:236                                                                            
3  0x0000000000560598 in _/e_/temp/weird.repro                                                                           
   at e:/temp/weird/repro.s:6                                                                                            
[...]

PR #1860, opened 2020-02-12
proc: only format registers value when it's necessary

aarzilli:
```
proc: only format registers value when it's necessary

A significant amount of time is spent generating the string
representation for the proc.Registers object of each thread, since this
field is rarely used (only when the Registers API is called) it should
be generated on demand.

Also by changing the internal representation of proc.Register to be
closer to that of op.DwarfRegister it will help us implement #1838
(when Delve will need to be able to display the registers of an
internal frame, which we currently represent using op.DwarfRegister
objects).

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	22292554301 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	17326345671 ns/op

Reduces conditional breakpoint latency from 2.2ms to 1.7ms.

Updates #1549, #1838

```

### DIFF
--- a/pkg/proc/amd64_arch.go
+++ b/pkg/proc/amd64_arch.go
@@ -1,12 +1,15 @@
 package proc
 
 import (
+	"bytes"
 	"encoding/binary"
+	"fmt"
+	"math"
+	"os"
 	"strings"
 
 	"github.com/go-delve/delve/pkg/dwarf/frame"
 	"github.com/go-delve/delve/pkg/dwarf/op"
-	"golang.org/x/arch/x86/x86asm"
 )
 
 // AMD64 represents the AMD64 CPU architecture.
@@ -300,24 +303,24 @@ func (a *AMD64) RegSize(regnum uint64) int {
 // figure 3.36
 // https://www.uclibc.org/docs/psABI-x86_64.pdf
 
-var amd64DwarfToHardware = map[int]x86asm.Reg{
-	0:  x86asm.RAX,
-	1:  x86asm.RDX,
-	2:  x86asm.RCX,
-	3:  x86asm.RBX,
-	4:  x86asm.RSI,
-	5:  x86asm.RDI,
-	8:  x86asm.R8,
-	9:  x86asm.R9,
-	10: x86asm.R10,
-	11: x86asm.R11,
-	12: x86asm.R12,
-	13: x86asm.R13,
-	14: x86asm.R14,
-	15: x86asm.R15,
-}
-
 var amd64DwarfToName = map[int]string{
+	0:  "Rax",
+	1:  "Rdx",
+	2:  "Rcx",
+	3:  "Rbx",
+	4:  "Rsi",
+	5:  "Rdi",
+	6:  "Rbp",
+	7:  "Rsp",
+	8:  "R8",
+	9:  "R9",
+	10: "R10",
+	11: "R11",
+	12: "R12",
+	13: "R13",
+	14: "R14",
+	15: "R15",
+	16: "Rip",
 	17: "XMM0",
 	18: "XMM1",
 	19: "XMM2",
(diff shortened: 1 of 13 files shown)


aarzilli:
I realize that the amd64DwarfToName map isn't used anymore, except to build it's inverse map (amd64NameToDwarf) and so we could remove. However I'd like to keep it around since it will eventually be used to implement #1838 (we'll need a way to put a name on registers for which we only have a dwarf register number).

Also, I explored the possibility of a much more radical change where the backend specifies the DWARF register number directly (instead of the thing we have now where we find the register number by using the register name). Technically it's possible, however for the gdbserial backend we only have the register names so it would simply push the ugly (register name → register number) table down, so it isn't really worth it.

Finally the extra register names that get added manually in the constructor of amd64NameToDwarf are alternate names for some registers that mozilla rr uses.

pkg/proc/arm64_arch.go
```diff
@@ -406,3 +409,45 @@ func (a *ARM64) AddrAndStackRegsToDwarfRegisters(staticBase, pc, sp, bp, lr uint
...
+		var v4 [4]float32
+		for i := range v4 {
+			binary.Read(buf, binary.LittleEndian, &v4[i])
+		}
+		fmt.Fprintf(&out, "\tv4_float={ %g %g %g %g }", v4[0], v4[1], v4[2], v4[3])
+
+		return out.String()
+	} else if reg.Bytes == nil || (reg.Bytes != nil && len(reg.Bytes) < 16) {
+		return fmt.Sprintf("%#016x", reg.Uint64Val)
+	} else {
```
derekparker: nit: drop this else, just unconditionally return at the end.

pkg/proc/amd64_arch.go
```diff
@@ -356,13 +359,25 @@ var amd64DwarfToName = map[int]string{
 	66: "SW",
 }
 
+var amd64NameToDwarf = func() map[string]int {
```
derekparker: Clever. I went back and forth during the review on whether this should happen in an init function, but I kind of like the initialization happening where the var declaration is.

pkg/proc/arm64_arch.go
```diff
@@ -406,3 +409,45 @@ func (a *ARM64) AddrAndStackRegsToDwarfRegisters(staticBase, pc, sp, bp, lr uint
 		LRRegNum:   arm64DwarfLRRegNum,
 	}
 }
+
+func (a *ARM64) DwarfRegisterToString(name string, reg *op.DwarfRegister) string {
```
derekparker: Unless I'm missing something `a` (method receiver) isn't actually being used here. What if this (and the other arch implementation) is a `String` method on the register type?

pkg/proc/amd64_arch.go
aarzilli: I didn't come up with this, I saw it somewhere.

pkg/proc/arm64_arch.go
aarzilli: It doesn't use a but it's different depending on the architecture. One thing we could do is stuff a function pointer inside op.DwarfRegister and then its string method would be:

```
func (reg *DwarfRegister) String() string {
    return reg.fmtFunc(reg)
}
```

However at some point we'll also want a way to get the register name given a register number and DwarfRegisterToString could be extended to do that, otherwise we'll need something else (probably another method of Arch).

aarzilli: Correction: it would be proc.Register not op.DwarfRegister to get the String method, of course.

derekparker: I was thinking more each concrete implementation would have this method, but after thinking it through a bit more I think your approach is cleaner.

aarzilli: I think it would be a shame to make it into an interface just for this thing, given that everything else is the same. Also we'll have to go back from register numbers to this interface objects somehow, and I don't know how good that would look.

PR #1871, opened 2020-02-18
pkg/proc: Allow function calls on non-struct types

derekparker:
Removes the restriction that the DWARF type for the receiver of a method
must be a TypeDef. This seems reasonable in theory, but in practice it turns out
Go DWARF does not consider

```
type X int
```

to be a typedef. This patch also allows for calling a method where the
receiver is not used or passed in, such as:

```
func (_ X) Method() { println("why") }
```

### DIFF
--- a/_fixtures/fncall.go
+++ b/_fixtures/fncall.go
@@ -121,7 +121,7 @@ func noreturncall(n int) {
 	return
 }
 
-type Base struct{
+type Base struct {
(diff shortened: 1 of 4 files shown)


derekparker:
Fixes #1871 1870

ear7h:
That was quick! This patch fixes the issues I was having, thank you!

PR #1873, opened 2020-02-19
Fixed DirIdx (index starts at one)

klemens-morgenstern:
I am using the elf to load C++ based elf and there the index starts at one and not zero, hence the minor fix.


klemens-morgenstern:
The current test doesn't do much since GoLang doesn't seem to use the includeDirs.

aarzilli:
This PR is making tests fail, I think the reason is that our IncludeDirs slice has, as the first element the compilation directory so the indexes are usually right (i.e. DirIdx 1 is the first element of the directory table).

I think the actual bug is that we only add the compilation directory to IncludeDirs if it isn't the empty string, so when there is no compilation directory the IncludeDirs slice is wrong.

The reason this bug doesn't show up with GoLang is that it will always include a non-empty compilation directory in the compile unit DIE, because otherwise dsymutil on macOS won't work.

pkg/dwarf/line/line_parser_test.go
```diff
@@ -289,3 +298,48 @@ func BenchmarkPCToLine(b *testing.B) {
 		runTestPCToLine(b, lineInfos, entries, basePCs, false, 0x10000)
 	}
 }
+
+func TestDebugLineC(t * testing.T) {
+
```
aarzilli: Remove newline

pkg/dwarf/line/line_parser_test.go
```diff
@@ -289,3 +298,48 @@ func BenchmarkPCToLine(b *testing.B) {
...
+	if len(parsed) == 0 {
+		t.Fatal("Parser result is empty")
+	}
+
+	file := []string{"main.c", "/mnt/c/develop/delve/_fixtures/main.c" ,"/usr/lib/gcc/x86_64-linux-gnu/7/include/stddef.h",
+			 "/usr/include/x86_64-linux-gnu/bits/types.h" ,"/usr/include/x86_64-linux-gnu/bits/libio.h", "/usr/include/stdio.h",
+			 "/usr/include/x86_64-linux-gnu/bits/sys_errlist.h"}
+
+	for _, ln := range parsed {
+
```
aarzilli: remove newline

pkg/dwarf/line/line_parser_test.go
```diff
@@ -289,3 +298,48 @@ func BenchmarkPCToLine(b *testing.B) {
...
+		if len(ln.FileNames) == 0 {
+			print("Parser could not parse Filenames")
```
aarzilli: t.Fatal instead of print

klemens-morgenstern:
Ah that makes more sense, I usually have an empty string there and the debug tools list it differently. I will see to fix it.

pkg/dwarf/line/line_parser_test.go
```diff
@@ -289,3 +298,44 @@ func BenchmarkPCToLine(b *testing.B) {
...
+			for _, cmp := range file {
+				if filepath.ToSlash(fn.Path) == cmp {
+					found = true
+					break
+				}
+			}
+			if !found {
+				t.Fatalf("Found %s does not appear in the filelist\n", fn.Path)
+			}
+
```
derekparker: nit: Remove newline.

pkg/dwarf/line/line_parser_test.go
```diff
@@ -289,3 +298,44 @@ func BenchmarkPCToLine(b *testing.B) {
...
+	if err != nil {
+		t.Fatal("Could not find test data", p, err)
+	}
+
+	data, err := ioutil.ReadFile(p)
+	if err != nil {
+		t.Fatal("Could not read test data", err)
+	}
+
+	parsed  :=  ParseAll(data, nil, 0, true)
```
derekparker: nit: cleanup whitespace (double space before and after `:=`).

PR #1874, opened 2020-02-20
Better native.Process.stop performance for linux backend

aarzilli:
```
proc/native/linux: do not call (*Thread).Stopped inside (*Process).stop

(*Thread).Stopped is slow because it needs to open, read and parse a
file in /proc, we don't actually need to do that, we can just rely on
the value of Thread.os.running.

Benchmark before:

BenchmarkConditionalBreakpoints-4              1        12476166303 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	10403533675 ns/op

Conditional breakpoint evaluation: 1.24ms -> 1ms

Updates #1549

proc/native/linux: only set breakpoints on threads that receive SIGTRAP

```

### DIFF
--- a/pkg/proc/native/proc_linux.go
+++ b/pkg/proc/native/proc_linux.go
@@ -250,16 +250,31 @@ func findExecutable(path string, pid int) string {
 }
 
 func (dbp *Process) trapWait(pid int) (*Thread, error) {
-	return dbp.trapWaitInternal(pid, false)
+	return dbp.trapWaitInternal(pid, 0)
 }
 
-func (dbp *Process) trapWaitInternal(pid int, halt bool) (*Thread, error) {
+type trapWaitOptions uint8
+
+const (
+	trapWaitHalt trapWaitOptions = 1 << iota
+	trapWaitNohang
+)
+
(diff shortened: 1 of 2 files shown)


pkg/proc/native/proc_linux.go
```diff
@@ -442,14 +460,26 @@ func (dbp *Process) stop(trapthread *Thread) (err error) {
...
+
+	for _, th := range dbp.threads {
+		th.os.setbp = false
+	}
+	trapthread.os.setbp = true
+
+	// check if any other thread simultaneously received a SIGTRAP
+	for {
+		th, _ := dbp.trapWaitInternal(-1, trapWaitNohang)
+		if th == nil {
```
derekparker: Wouldn't we want to set `th.os.setbp = true` in this case as well (when `th != nil`)?

aarzilli: trapWaitInternal sets setbp appropriately when the signal is SIGTRAP, doing it here would set it for threads that received the SIGSTOP we sent.

PR #1877, opened 2020-02-20
proc,proc/*: add StopReason field to Target

aarzilli:
```
proc,proc/*: add StopReason field to Target

Adds a StopReason field to the Target object describing why the target
process is currently stopped. This will be useful for the DAP server
(which needs to report this reason in one of its requests) as well as
making pull request #1785 (reverse step) conformant to the new
architecture.

proc,proc/*: move SelectedGoroutine to proc.Target, remove PostInitializationSetup

moves SelectedGoroutine, SwitchThread and SwitchGoroutine to
proc.Target, merges PostInitializationSetup with NewTarget.

```

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -1,8 +1,18 @@
 package proc
 
+import (
+	"fmt"
+)
+
 // Target represents the process being debugged.
 type Target struct {
 	Process
+
+	proc ProcessInternal
+
+	// Goroutine that will be used by default to set breakpoint, eval variables, etc...
+	// Normally selectedGoroutine is currentThread.GetG, it will not be only if SwitchGoroutine is called with a goroutine that isn't attached to a thread
+	selectedGoroutine *G
 
 	// fncallForG stores a mapping of current active function calls.
 	fncallForG map[int]*callInjection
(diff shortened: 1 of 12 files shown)


aarzilli:
cc @polinasok

pkg/proc/target.go
```diff
@@ -26,8 +28,25 @@ type Target struct {
...
+// A process could be stopped for multiple simultaneous reason, in which
+// case only one will be reported.
+type StopReason uint8
+
+const (
+	StopUnknown             StopReason = iota
+	StopLaunched                       // The process was just launched
+	StopAttached                       // The debugger stopped the process after attaching
+	StopExited                         // The target proces terminated
+	StopBreakpoint                     // The target process hit one or more software breakpoints
```
polinasok: Is it not worth differentiating what kind of breakpoint this was (function, data condition, etc)?

aarzilli: We don't have data breakpoints, when we do we'll do that. Differentiating between function breakpoints and breakpoints on things other than the first line of a function would be complicated and IMHO not very useful.

pkg/proc/target.go
```diff
@@ -16,19 +28,60 @@ type Target struct {
...
+	StopHardcodedBreakpoint            // The target process hit a hardcoded breakpoint (for example runtime.Breakpoint())
+	StopManual                         // A manual stop was requested
+	StopNextFinished                   // The next/step/stepout command terminated
+	StopCallReturned                   // An injected call commpleted
+)
+
 // NewTarget returns an initialized Target object.
-func NewTarget(p Process, disableAsyncPreempt bool) *Target {
+func NewTarget(p Process, path string, debugInfoDirs []string, writeBreakpoint WriteBreakpointFn, disableAsyncPreempt bool, stopReason StopReason) (*Target, error) {
```
derekparker: This function signature is getting a little out of hand. Maybe we have a config object?

aarzilli: Done.

pkg/proc/target.go
```diff
@@ -1,9 +1,21 @@
...
+	"fmt"
+)
+
 // Target represents the process being debugged.
 type Target struct {
 	Process
 
+	proc ProcessInternal
+
+	StopReason StopReason
```
derekparker: nit: Add a doc comment to this.

aarzilli: Done.

PR #1879, opened 2020-02-20
terminal: add ability to print goroutine labels

aarzilli:
```
terminal: add ability to print goroutine labels

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -189,7 +189,8 @@ Print out info for every goroutine. The flag controls what information is shown 
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
 	-s	displays location of the start function
-	-t	displays stack trace of goroutine
+	-t	displays goroutine's stacktrace
+	-l	displays goroutine's labels
 
 If no flag is specified the default is -u.`},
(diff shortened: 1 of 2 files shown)


pkg/terminal/command.go
```diff
@@ -650,7 +660,9 @@ func goroutines(t *Term, ctx callContext, argstr string) error {
 			case "-s":
 				fgl = fglStart
 			case "-t":
-				bPrintStack = true
+				flags |= printGoroutinesStack
+			case "-l":
```
derekparker: This new flag needs to be documented: https://github.com/go-delve/delve/blob/2359e6b2ff949a536e636a1a5c8f7d30f630fe0a/pkg/terminal/command.go#L192.

pkg/terminal/command.go
```diff
@@ -877,6 +889,36 @@ func writeGoroutineLong(w io.Writer, g *api.Goroutine, prefix string) {
...
+		more = true
+		keys = keys[:maxNumberOfGoroutineLabels]
+	}
+	fmt.Fprintf(w, "%sLabels: ", prefix)
+	for i, k := range keys {
+		fmt.Fprintf(w, "%q:%q", k, g.Labels[k])
+		if i != len(keys)-1 {
+			fmt.Fprintf(w, ", ")
+		} else if more {
+			fmt.Fprintf(w, "...")
```
derekparker: nit: Maybe specify the number that we left out? (e.g. foo,bar,baz ... + n more)

PR #1881, opened 2020-02-20
service/dap: avoid hardcoded []byte slices in DAP tests

eliben:
DAP tests will check server responses by analyzing interesting fields on a field-by-field basis rather than comparing serialized messages with `[]byte` golden strings.

Updates #1515


eliben:
@polinasok

service/dap/daptest/client.go
```diff
@@ -49,30 +49,96 @@ func (c *Client) send(request dap.Message) {
...
-	switch response.(type) {
-	case *dap.ErrorResponse:
-		return response, nil
-	default:
-		return nil, errors.New(fmt.Sprintf("not an ErrorResponse: %#v", response))
+	return m.(*dap.ErrorResponse)
+}
+
+func (c *Client) ExpectContinueResponse(t *testing.T) *dap.ContinueResponse {
+	m, err := dap.ReadProtocolMessage(c.reader)
```
aarzilli: I think you would be better served by defining a helper function:

```
func mustReadProtocolMessage(t *testing.T, c *Client) dap.Message {
   m, err := dap.ReadProtocolMessage(c.reader)
   if err != nil {
       t.Error(err)
   }
   return m
}
```

then you wouldn't even need these, you could just `mustReadProtocolMessage(t, c).(*dap.ConcreteResponseType)`

eliben: Thanks for the suggestion. Just to make sure I understand precisely, so you mean I should use (1) this helper function to simplify the `Expect*` methods, or (2) replace all calls to `Expect*` with this?

E.g., replace:

```
contResp := client.ExpectContinueResponse(t)
```

With:

```
contResp := mustReadProtocolMessage(t, client).(*dap.ContinueResponse)
```

The tradeoff, IMHO is:

* The original has much cleaner call sites in tests, so the tests themselves are easier to read.
* We have some custom logic in some `Expect*` method (e.g. `ExpectInitializeResponse`) and may add more such logic in the future. Removing the helpers means this logic needs to be duplicated at call sites, or have specific helpers just for relevant methods.

On the other hand, your proposal helps remove a bunch of repetitive methods, which is an advantage.

A middle way is (1), replace:

```
func (c *Client) ExpectContinueResponse(t *testing.T) *dap.ContinueResponse {
	m, err := dap.ReadProtocolMessage(c.reader)
	if err != nil {
		t.Error(err)
	}
	return m.(*dap.ContinueResponse)
}
```

By:

```
func (c *Client) ExpectContinueResponse(t *testing.T) *dap.ContinueResponse {
       return c.mustReadProtocolMessage().(*dap.ContinueResponse)
}
```

This keeps all the `Expect*` methods but makes them one-liners, except those where extra logic is needed. The call sites in tests remain short and readable, but the amount of repetitive code is reduced.

WDYT?

aarzilli: I was thinking just:

```
contResp := mustReadProtocolMessage(t, client).(*dap.ContinueResponse)
```

but if you think you'll need to add extra logic then keeping the Expect methods makes sense.

polinasok: I like the idea of keeping the Expect methods, but factoring out all the common parts into a mustRead healper.

polinasok: Except for consistency it should be called ExpectProtocolMessage or the other one should be called MustReadFooRequest.

service/dap/daptest/client.go
```diff
@@ -49,30 +49,96 @@ func (c *Client) send(request dap.Message) {
 	dap.WriteProtocolMessage(c.conn, request)
 }
 
-// ReadBaseMessage reads and returns a json-encoded DAP message.
-func (c *Client) ReadBaseMessage() ([]byte, error) {
-	message, err := dap.ReadBaseMessage(c.reader)
+func (c *Client) ExpectDisconnectResponse(t *testing.T) *dap.DisconnectResponse {
```
polinasok: Should these be calling t.Helper()?
(Just learned about it from PR #1882)

service/dap/daptest/client.go
eliben: Done

service/dap/daptest/client.go
eliben: Done

eliben:
The 1.12 macOS builder errors out with:

```
Building without native backend

go build "-ldflags=-s -X main.Build=1f13e2e24d039036bccdd2d9c4a679aca0b75dd1" github.com/go-delve/delve/cmd/dlv

Testing

ok  	github.com/go-delve/delve/cmd/dlv	28.310s

?   	github.com/go-delve/delve/cmd/dlv/cmds	[no test files]

ok  	github.com/go-delve/delve/pkg/config	0.049s

?   	github.com/go-delve/delve/pkg/dwarf/dwarfbuilder	[no test files]

ok  	github.com/go-delve/delve/pkg/dwarf/frame	0.048s

?   	github.com/go-delve/delve/pkg/dwarf/godwarf	[no test files]

ok  	github.com/go-delve/delve/pkg/dwarf/line	1.756s

?   	github.com/go-delve/delve/pkg/dwarf/loclist	[no test files]

ok  	github.com/go-delve/delve/pkg/dwarf/op	0.049s

?   	github.com/go-delve/delve/pkg/dwarf/reader	[no test files]

ok  	github.com/go-delve/delve/pkg/dwarf/util	0.049s

ok  	github.com/go-delve/delve/pkg/goversion	0.075s

?   	github.com/go-delve/delve/pkg/logflags	[no test files]

ok  	github.com/go-delve/delve/pkg/proc	118.970s

ok  	github.com/go-delve/delve/pkg/proc/core	0.054s

?   	github.com/go-delve/delve/pkg/proc/core/minidump	[no test files]

ok  	github.com/go-delve/delve/pkg/proc/fbsdutil	0.052s

ok  	github.com/go-delve/delve/pkg/proc/gdbserial	1.676s

ok  	github.com/go-delve/delve/pkg/proc/linutil	0.056s

?   	github.com/go-delve/delve/pkg/proc/native	[no test files]

?   	github.com/go-delve/delve/pkg/proc/test	[no test files]

?   	github.com/go-delve/delve/pkg/proc/winutil	[no test files]

ok  	github.com/go-delve/delve/pkg/terminal	18.903s

?   	github.com/go-delve/delve/pkg/terminal/starbind	[no test files]

?   	github.com/go-delve/delve/pkg/version	[no test files]

?   	github.com/go-delve/delve/service	[no test files]

?   	github.com/go-delve/delve/service/api	[no test files]

ok  	github.com/go-delve/delve/service/dap	1.533s

?   	github.com/go-delve/delve/service/dap/daptest	[no test files]

ok  	github.com/go-delve/delve/service/debugger	0.053s

?   	github.com/go-delve/delve/service/rpc1	[no test files]

?   	github.com/go-delve/delve/service/rpc2	[no test files]

?   	github.com/go-delve/delve/service/rpccommon	[no test files]

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

Is this known flakiness? (timeout?)

Is there a way to kick a build without pushing a new commit?

aarzilli:
PS. I think the timeout is a problem with travis, actually.

Issue #20, opened 2014-11-14
Support for 32-bit systems


muniz95:
Correct me if I am wrong, but your tool only works in 64-bit systems. I tried to install **dlv** via **go get** and got this error:

<code>
go build github.com/derekparker/delve/proctl: no buildable Go source files in /home/user/go/src/github.com/derekparker/delve/proctl
</code>

AFAIK those who had 64-bit O.S. got **dlv** working properly.

P.S: I have libreadline installed.

sethwklein:
I'm out of time now, but I want to throw some stuff in here to save time for the next person who looks at this. These instructions trade the precision of a branch for markers of intent that hopefully will prevent them from bit rotting so fast. The 32bit branch no longer applies cleanly to master.

To start running tests in `proctl`...
* move `proctl_linux_amd64.go` -> `proctl_linux.go`
* move `breakpoints_linux_amd64.go` -> `breakpoints_linux.go`
* move `setHardwareBreakpoint` from `breakpoints_linux.go` to `breakpoints_linux_amd64.go` and add C imports from `breakpoints_linux.go` but change the definition of `offset` to a declaration
* copy `breakpoints_linux_amd64.go` -> `breakpoints_linux_386.go` and replace `C.DR_LEN_8` with whatever is correct. I don't know, but I used `C.DR_LEN_4`
* copy `threads_linux_amd64.go` -> `threads_linux_386.go` and replace `Rsp` with whatever is correct. I don't know, but I used uint64(Esp)

At this point, proctl tests should run, but fire hose your terminal with "Hello, World!". You may have to `killall testprog` from another terminal to make it stop. After you fix that, you should see the tests panicking in `"delve/dwarf/frame".parseFDE`. If you make a test like the following, you'll find that this doesn't happen on an actual binary, but the proctl code calls `frame.Parse` on `/proc/{{fd}}/exe`, not directly on a binary.

```go
package frame

import (
	"debug/elf"
	"os"
	"os/exec"
	"testing"
)

func DirectTest(t *testing.T) {
	const progName = "./testprog"
	cmd := exec.Command("go", "build", "-gcflags=-N -l", "-o", progName, "../../_fixtures/"+progName+".go")
	err := cmd.Run()
	if err != nil {
		t.Fatalf("Couldn't compile test program: %v\n", err)
	}
	defer os.Remove(progName)

	exe, err := elf.Open(progName)
	if err != nil {
		t.Fatalf("ELF missing: %v\n", err)
	}

[...]

PR #1884, opened 2020-02-21
pkg,service: support 386 on linux

chainhelen:
Implement debugging function for 386 on linux with reference to AMD64.
There are a few remaining problems that need to be solved in another time.

1. The stacktrace of cgo are not exactly as expected.
2. Not implement `core` for now.
3. Not implement `call` for now. Can't not find `runtime·debugCallV1` or
   similar function in $GOROOT/src/runtime/asm_386.s.

Update #20

------------------

Testcase pass on go1.13.6 linux/386 for me. (Skip some cases referred problems above ).

But when I want to add `arch 386` into .travis.yml to run ci, I find [Add 32-bit environments](https://github.com/travis-ci/travis-ci/issues/986) and [32 bit versions of linux images](https://github.com/travis-ci/travis-ci/issues/5770).  It looks travis doesn't support 32bit directly.

I can close this pr if dlv will not support 32 bit.


chainhelen:
I check the below travis CI on 64bit .  
goversion 1.13 sucess and 1.12, 1.11 failed. 
It looks go 1.12 , go 1.11 on travis pull some dlv's codes from remote `go-delve/delve`, but not my this pr. This [commit](https://github.com/go-delve/delve/commit/17f2fa7908308c2b467ff4605388b63e3d1b76fb#diff-f07ec4d9a9e44ace9b300b2d6fc87952R302) of dlv master affect this pr's testcase for go1.12 and go 1.11.So I think it doesn't matter.

aarzilli:
The reason the test is failing is that travis is testing the merge of your changes with master. A recent change was made to master that added a test in line_parser_test.go. You have to rebase this test on master to make it work.

pkg/proc/threads.go
```diff
@@ -209,6 +209,24 @@ func next(dbp Process, stepInto, inlinedStepOut bool) error {
 				continue
 			}
 
+			// On 386 linux when buildmode is pie, compiler insert more instructions (ex: call __x86.get_pc_thunk.).
```
aarzilli: I think this is too hacky, it would be preferrable to figure out the address of x86.get_pc_thunk by reading the symbol table and add an entry for that in bi.Functions.

service/test/integration2_test.go
```diff
@@ -1132,7 +1140,9 @@ func TestSkipPrologue2(t *testing.T) {
 		callme3 := findLocationHelper(t, c, "main.callme3", false, 1, 0)[0]
 		callme3Z := uint64(clientEvalVariable(t, c, "main.callme3").Addr)
 		ver, _ := goversion.Parse(runtime.Version())
-		if ver.Major < 0 || ver.AfterOrEqual(goversion.GoVer18Beta) {
+
+		is64Bit := runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64"
+		if (ver.Major < 0 || ver.AfterOrEqual(goversion.GoVer18Beta)) && is64Bit {
```
aarzilli: I've compiled callme.go with GOARCH=386 and it does have a prologue, there's something wrong with prologue detection on i386 if this change is necessary.

pkg/proc/i386_arch.go
```diff
@@ -0,0 +1,422 @@
...
+// SwitchStack will use the current frame to determine if it's time to
+// switch between the system stack and the goroutine stack or vice versa.
+// Sets it.atend when the top of the stack is reached.
+func (i *I386) SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
+	if it.frame.Current.Fn == nil {
+		return false
+	}
+	switch it.frame.Current.Fn.Name {
+	case "runtime.asmcgocall":
+		if it.top || !it.systemstack {
```
aarzilli: From what I can tell from the tests, cgo stacktraces don't work. Therefore I would prefer that SwitchStack didn't try to do things it can't do and simply remove the cases for runtime.asmcgocall and runtime.cgocallback_gofunc for now.

chainhelen:
Ok, I will rebase. Thx for review.

service/test/integration2_test.go
chainhelen: I find [unixGetG](https://github.com/go-delve/delve/pull/1884/commits/d7dc84ac9ea5289b7aee4632f9f983ab61b0dab0#diff-1aec94b2d00c8b6e87bc18b22ed320cfR32) should be {`mov`, `mov`}, I will remove [windowsGetG](https://github.com/go-delve/delve/pull/1884/commits/d7dc84ac9ea5289b7aee4632f9f983ab61b0dab0#diff-1aec94b2d00c8b6e87bc18b22ed320cfR33) temporarily.   

---  

I feel puzzled at  `something wrong with prologue detection`.  
```
19   func callme3() {
20 	 callme2()
21   }

// I disassme on 386
        .:0             0x80c566f       cc              int3
        callme.go:19    0x80c5670       658b0d00000000  mov ecx, dword ptr gs:[]
        callme.go:19    0x80c5677       8b89fcffffff    mov ecx, dword ptr [ecx+0xfffffffc]
        callme.go:19    0x80c567d       3b6108          cmp esp, dword ptr [ecx+0x8]
        callme.go:19    0x80c5680       7606            jbe 0x80c5688
        callme.go:20    0x80c5682       e819ffffff      call $main.callme2
        callme.go:21    0x80c5687       c3              ret
        callme.go:19    0x80c5688       e883c8fcff      call $runtime.morestack_noctxt
        callme.go:19    0x80c568d       ebe1            jmp $main.callme3
```
It is my understanding that `0x80c5670 ~ 0x80c5680`  for  line 19 are `stacksplit`'s codes. 
When  cmd `list main.callme3`, in fact it become line 20.  
So cmd `list callme.go:19` can only select `0x80c5670`.  

If I understand it wrong, please point it out. Thx.

pkg/dwarf/util/ptr_size.go
```diff
@@ -0,0 +1,19 @@
+package util
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func PtrSizeByRuntimeArch() int {
```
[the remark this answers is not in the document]
chainhelen: > This is only ever used for tests so I would prefer that it be inside the `_test.go` files that need it. Also the body can be replaced by:
> 
> ```
> return unsafe.Sizeof(uintptr(0))
> ```

Oh,  the code is so cool, I will do it.

service/test/integration2_test.go
chainhelen: So I think this testcase may need special treatment for 386.

pkg/proc/threads.go
chainhelen: En, I have try to get symbol from symbol table by the addr calleld, it is ok. 

----

But  the comments of `Functions` is
`Functions is a list of all DW_TAG_subprogram entries in debug_info, sorted by entry point`. 

I am not sure additional funcs from symbol table, whether destroy the original logic of `bi.Functions`(addtional funcs has no fn.entry, fn.end, dwarfoffset and compileUnit, just has name and addr if my understanding is right).  

I think other way is  add one filed named `SymNames`( type is `map[addr]elf.Symbol`). This map is just used by  `bi.symLookup`

service/test/integration2_test.go
aarzilli: Ok, you're right about this. However I think the test should be `runtime.GOARCH != "386"` rather than `runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64"`.

pkg/proc/threads.go
aarzilli: True, it would probably be a problem to have Function objects without a compile unit. 

Instead of keeping around the whole gosym.Table map I'd just use `gosym.(*Table).PCToFunc("__x86.get_pc_thunk")` to get its entry point and then the check is `instr.IsCall() && instr.DestLoc.PC == entryPointOfx86getPCThunk`.

chainhelen: 1.
__x86.get_pc_thunk is a array, even same name but in diffrent place.

```
addr       name
127520 __x86.get_pc_thunk.ax
127530 __x86.get_pc_thunk.cx
127540 __x86.get_pc_thunk.dx
127550 __x86.get_pc_thunk.bx
127560 __x86.get_pc_thunk.bp
127570 __x86.get_pc_thunk.si
127580 __x86.get_pc_thunk.di
12750c __x86.get_pc_thunk.dx
127510 __x86.get_pc_thunk.bx
``` 
Can't not use entryPointOfx86getPCThunk.   

-----

2. 
And one benefit for addtional field SymNames in `bi` to help search name.

diassemble
```
(dlv) disassemble
TEXT main.Foo(SB) /root/workspace/go-delve/delve/_fixtures/teststepconcurrent.go
        teststepconcurrent.go:13        0x1c0eb0        e81b6ef7ff              call $C._IO_2_1_stdout_+163024
        teststepconcurrent.go:13        0x1c0eb5        c7c1fcffffff            mov ecx, 0xfffffffc
        teststepconcurrent.go:13        0x1c0ebb        658b09                  mov ecx, dword ptr gs:[ecx]
        teststepconcurrent.go:13        0x1c0ebe        3b6108                  cmp esp, dword ptr 
```

will become

```
(dlv) disassemble
TEXT main.Foo(SB) /root/workspace/go-delve/delve/_fixtures/teststepconcurrent.go
        teststepconcurrent.go:13        0x1c0eb0        e81b6ef7ff              call $__x86.get_pc_thunk.cx
        teststepconcurrent.go:13        0x1c0eb5        c7c1fcffffff            mov ecx, 0xfffffffc
        teststepconcurrent.go:13        0x1c0ebb        658b09                  mov ecx, dword ptr gs:[ecx]
        teststepconcurrent.go:13        0x1c0ebe        3b6108                  cmp esp, dword ptr [ecx+0x8]

```  
---- 

3. 
I find that here should be clean for `theard.go`. 
Place function `bi.IsX86GetPCThunk(pc uint64)`  into bi. 
And just keep clean here.
```
if instr.IsCall() && instr.DestLoc != nil && dbp.BinInfo().IsX86GetPCThunk(instr.DestLoc.PC) {
     continue
}
```

aarzilli: I don't want to depend on the assembly formatter behaving in a certain way to implement an unrelated feature. And the `IsX86GetPCThunk` should be called `InhibitStepInto` (or something like that) and be part of the Arch object, probably.

chainhelen: I agree with you and get it.

pkg/dwarf/util/util.go
```diff
@@ -124,3 +126,31 @@ func ParseString(data *bytes.Buffer) (string, uint32) {
 
 	return str[:len(str)-1], uint32(len(str))
 }
+
+func ReadUintRaw(reader io.Reader, order binary.ByteOrder, ptrSize int) (uint64, error) {
```
aarzilli: Comment:

```
// ReadUintRaw reads an integer of ptrSize bytes, with the specified byte order, from reader.
```

pkg/dwarf/util/util.go
```diff
@@ -124,3 +126,31 @@ func ParseString(data *bytes.Buffer) (string, uint32) {
...
+		var n uint64
+		if err := binary.Read(reader, order, &n); err != nil {
+			return 0, err
+		}
+		return n, nil
+	}
+	return 0, fmt.Errorf("not supprted ptr size %d", ptrSize)
+}
+
+func WriteUint(writer io.Writer, order binary.ByteOrder, ptrSize int, data uint64) error {
```
aarzilli:
```
// WriteUint writes an integer of ptrSize bytes to writer, in the specified byte order.
```

pkg/proc/bininfo.go
```diff
@@ -974,19 +1007,15 @@ func (bi *BinaryInfo) setGStructOffsetElf(image *Image, exe *elf.File, wg *sync.
...
 	for _, prog := range exe.Progs {
 		if prog.Type == elf.PT_TLS {
 			tls = prog
 			break
 		}
 	}
-	if tls == nil {
-		bi.gStructOffset = ^uint64(8) + 1 // -8
+	if tlsg == nil || tls == nil {
+		bi.gStructOffset = ^uint64(bi.Arch.PtrSize()) + 1 //-ptrSize + 1
```
aarzilli: The comment is wrong, it's `// -ptrSize` not `-ptrSize + 1` (it's taking the 2s complement of 8).

pkg/proc/i386_arch.go
```diff
@@ -0,0 +1,369 @@
...
+			return name, false, fmt.Sprintf("%#x", reg.Bytes)
+		}
+	}
+}
+
+// InhibitStepInto returns whether StepBreakpoint can be set at pc.
+// When cgo or pie on 386 linux, compiler will insert more instructions (ex: call __x86.get_pc_thunk.).
+// StepBreakpoint shouldn't be set on __x86.get_pc_thunk and skip it.
+// See comments on stacksplit in $GOROOT/src/cmd/internal/obj/x86/obj6.go for generated instructions details.
+func (i *I386) InhibitStepInto(bi *BinaryInfo, pc uint64) bool {
```
aarzilli: cc @derekparker what if instead of doing this and having to read the symbol table we just didn't step into functions for which we don't have a debug_info entry? We could make a note about this, patch the linker to emit DIEs for __x86.get_pc_thunk and then use that in a couple of years.

pkg/proc/linutil/dynamic.go
```diff
@@ -21,6 +22,24 @@ const (
 	_DT_DEBUG = 21 // DT_DEBUG as defined by SysV ABI specification
 )
 
+func readUintRaw(reader io.Reader, order binary.ByteOrder, ptrSize int) (uint64, error) {
```
aarzilli: What's the reason for not using dwarf/util.ReadUintRaw?

service/test/integration1_test.go
```diff
@@ -945,6 +945,14 @@ func Test1Disasm(t *testing.T) {
 			if strings.HasPrefix(curinstr.Text, "call") || strings.HasPrefix(curinstr.Text, "CALL") {
 				t.Logf("call: %v", curinstr)
 				if curinstr.DestLoc == nil || curinstr.DestLoc.Function == nil {
+					// On 386 linux when pie, maybe meet  "call __x86.get_pc_thunk." first but not find any symbol name or file name.
```
aarzilli: I think that instead of doing this we should skip the test for 386. Before the for loop:

```
if runtime.GOOS == "386" {
    // Skip the rest of the test because on intel 386 the compiler will insert calls to __x86.get_pc_thunk which do not have DIE and we can't resolve.
    return
}
```

service/test/integration2_test.go
```diff
@@ -1052,6 +1052,15 @@ func TestDisasm(t *testing.T) {
 			if strings.HasPrefix(curinstr.Text, "call") || strings.HasPrefix(curinstr.Text, "CALL") {
 				t.Logf("call: %v", curinstr)
 				if curinstr.DestLoc == nil || curinstr.DestLoc.Function == nil {
+					// On 386 linux when pie, maybe meet  "call $__x86.get_pc_thunk." first
```
aarzilli: Same as Test1Disasm.

pkg/dwarf/line/state_machine_test.go
```diff
@@ -44,6 +44,10 @@ func TestGrafana(t *testing.T) {
 		t.Skip("filepath.Join ruins this test on windows")
 	}
 
+	if ptrSizeByRuntimeArch() != 8 {
```
aarzilli: Remove this `if` and then in this test instead of calling `ptrSizeByRuntimeArch` hardcode 8

pkg/proc/bininfo.go
chainhelen: Oh, sorry for this mistake

pkg/proc/linutil/dynamic.go
chainhelen: Because I think `dwarf/util.ReadUintRaw` maybe just used for `dwarf` or something about `dwarf`.  
In fact, I want to put `ReadUintRaw` input one common tool package, but I can't find such package.

service/test/integration1_test.go
chainhelen: 1. when nopie for 386 linux, it is ok

2. For user, `step` can't enter `__x86.get_pc_thunk`, but `step instruction` can.  We maybe need to keep this testcase to verify this logic.

pkg/dwarf/line/state_machine_test.go
chainhelen: Oh, I'm sorry for my understanding. I get it.

pkg/proc/bininfo.go
chainhelen: Done

pkg/dwarf/util/util.go
chainhelen: Done

pkg/dwarf/util/util.go
chainhelen: Done

pkg/proc/i386_arch.go
chainhelen: Need a decision

pkg/proc/linutil/dynamic.go
chainhelen: Need to discuss

service/test/integration1_test.go
chainhelen: Need to discuss

service/test/integration2_test.go
chainhelen: Need to discuss

pkg/dwarf/line/state_machine_test.go
chainhelen: Done

service/test/integration1_test.go
aarzilli: Ok then `if runtime.GOOS == "386" && buildMode == "pie"`

chainhelen:
Rebase by master. 
It looks `ProcessVmRead` a little bit repetitive when supporting 386, but I have no a better solution.

service/test/integration1_test.go
chainhelen: Done

pkg/proc/threads.go
aarzilli: As I was saying before the test calling `bi.Arch.InhibitStepInto(bi, instr.DestLoc.PC)` should go inside setStepIntoBreakpoint, remove the check for `instr.DestLoc.Fn != nil` in the if at line 216 and inside setStepIntoBreakpoint do:

```
if dbp.BinInfo().Arch.InhibitStepInto(dbp.BinInfo(), instr.DestLoc.PC) {
    return nil
}
```

around the same place where we skip private runtime functions.

pkg/proc/native/ptrace_linux_amd64.go
```diff
@@ -39,3 +39,31 @@ func PtraceGetRegset(tid int) (regset linutil.AMD64Xstate, err error) {
 	err = linutil.AMD64XstateRead(regset.Xsave, false, &regset)
 	return
 }
+
+// ProcessVmRead calls process_vm_readv
+func ProcessVmRead(tid int, addr uintptr, data []byte) (int, error) {
```
aarzilli: Make a file called ptrace_linux_64bit.go and add this at the top:

```
// +build linux,amd64 linux,arm64
```

and put ProcessVmRead, ProcessVmWrite inside it.

service/test/integration2_test.go
chainhelen: Done

pkg/proc/native/ptrace_linux_amd64.go
chainhelen: Ok, I will do it.

pkg/proc/threads.go
chainhelen: Ok, done and rebase.

pkg/proc/threads.go
```diff
@@ -399,6 +398,11 @@ func setStepIntoBreakpoint(dbp Process, text []AsmInstruction, cond ast.Expr) er
 
 	pc := instr.DestLoc.PC
 
+	// Skip InhibitStepInto functions for different arch.
+	if instr.IsCall() && dbp.BinInfo().Arch.InhibitStepInto(dbp.BinInfo(), pc) {
```
aarzilli: No need to have `instr.IsCall()` here, it's always a call instruction.

service/test/integration1_test.go
aarzilli: What I meant is to add:

```
if runtime.GOOS == "386" && buildMode == "pie" {
    // Skip the rest of the test because on intel 386 with PIE build mode the compiler will insert calls to __x86.get_pc_thunk which do not have DIEs and we can't resolve.
    return
}
```

before the for loop.

chainhelen: Sorry, has done.

pkg/proc/threads.go
chainhelen: Ok, done.

pkg/proc/i386_arch.go
derekparker: Sorry for the delay, I didn't see the ping on this. I think I'm alright with this solution as is.

chainhelen:
There is a issue about ci as what I said above. 

1. 
Travis not support 32bit. According to [travis-ci#5770](https://github.com/travis-ci/travis-ci/issues/5770), need to use docker. 
If we change the .travis.yml to use docker for 32bit, maybe cause unfriendly ci report or complex configuration? 


2. 
I find [scipy](https://github.com/scipy/scipy) solve it by azure, but azure also not support 32bit linux (32bit window is ok, and [introduce is here](https://docs.microsoft.com/en-us/azure/site-recovery/vmware-physical-azure-support-matrix#ubuntu-kernel-versions)), it use docker too.  
And I investigate [scipy#azure-pipelines.yml](https://github.com/scipy/scipy/blob/master/azure-pipelines.yml), then write a azure-pipelines.yml for 386 on linux.  
```
trigger:
  branches:
    include:
    - '*'  

jobs:
- job: linux_32bit
  pool:
    vmImage: 'ubuntu-18.04'
  
  strategy:
    matrix:
      go12:
        go_version: "1.12"
      go13:
        go_version: "1.13"
      go14:
        go_version: "1.14"
    maxParallel: 3
  
  variables:
    go_project: "delve"
  
  steps:
  - script: |
      docker pull i386/ubuntu:bionic
      docker run -v $(pwd):/$(go_project) i386/ubuntu:bionic /bin/bash -c "cd $(go_project) && \
      apt-get -y update && \
      apt-get -y install software-properties-common && \
      apt-get -y install git && \
      add-apt-repository ppa:longsleep/golang-backports && \
      apt-get -y install golang-$(go_version)-go && \
      export PATH=$PATH:/usr/lib/go-$(go_version)/bin && \
      go version && \
      uname -a && \
      make test"
    displayName: 'linux 32bit'
```

ps: I have try it on other project. Just change `go_project` name, others are same as mine.   

---

Need to discuss.

chainhelen:
> Maybe we run a 32bit docker image of golang on Cirrus CI or something?  

Ok , I will Investigate this and fix tow places above today.

pkg/proc/i386_arch.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+}
+
+// PtrSize returns the size of a pointer
+// on this architecture.
+func (i *I386) PtrSize() int {
+	return 4
+}
+
+// MaxInstructionLength returns the maximum lenght of an instruction.
```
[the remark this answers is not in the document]
chainhelen: Done

pkg/proc/bininfo.go
```diff
@@ -98,7 +102,7 @@ type BinaryInfo struct {
 }
 
 // ErrUnsupportedLinuxArch is returned when attempting to debug a binary compiled for an unsupported architecture.
-var ErrUnsupportedLinuxArch = errors.New("unsupported architecture - only linux/amd64 and linux/arm64 are supported")
+var ErrUnsupportedLinuxArch = errors.New("unsupported architecture - only linux/amd64, linux/arm64 and linux/i386 are supported")
```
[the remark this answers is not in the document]
chainhelen: I remove `ErrUnsupportedLinuxArch`, `ErrUnsupportedWindowsArch` `ErrUnsupportedDarwinArch` . 

Create new error `ErrUnsupportedArch` for unsupported circumstances.  
(via the `supportedLinuxArch`, `supportedWindowsArch`, `supportedDarwinArch` map)

derekparker:
@chainhelen thanks! I'm going to re-run the tests. I don't know of a way to collapse the test results as with Travis. We might be able to set up a matrix or something on CirrusCI, but that doesn't have to happen here.

derekparker:
@chainhelen right! Added it in, can't believe I overlooked that one.

mewmew:
Incredible work @chainhelen! The original 32-bit issue was opened 6 years ago. It's great to see progress being made to add 32-bit support to Delve!

So, I tried it out for a sample binary (at rev c63ae072bcfcf418a6af772e906738b78ff7435b) and first I made the mistake of running `dlv` compiled for my native machine architecture (AMD64) when trying to debug a 32-bit ELF sample. This resulted in the wrong pointer size being used (as expected).

```
[u@x220 rodnad]$ dlv exec ./sample
goarch: amd64
elfFile.Machine: EM_386
bi.Arch.PtrSize(): 8
panic: runtime error: slice bounds out of range [16:12]

goroutine 7 [running]:
github.com/go-delve/delve/pkg/dwarf/frame.parseFDE(0xc0001da000, 0xbd1f68)
	/home/u/goget/src/github.com/go-delve/delve/pkg/dwarf/frame/parser.go:88 +0x2d4
```

As the 32-bit support was implemented for the `native` package I tried to rebuild `dlv` with `GOARCH=386`, but then ran into the following build error:

```
[u@x220 ~]$ GOOS=linux GOARCH=386 go install -v github.com/go-delve/delve/cmd/dlv
github.com/go-delve/delve/pkg/proc/native
# github.com/go-delve/delve/pkg/proc/native
goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:66:13: undefined: ptraceGetTls
goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:83:51: undefined: ptraceGetRegset
goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:97:16: undefined: processVmWrite
goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:112:9: undefined: processVmRead
```

Is this to be expected?

Once more, thanks for working on this! It's tremendously useful to have a debugger for Go written in Go with support for the various architectures that Go compiles down to.

Cheerful regards,
Robin

chainhelen:
> Incredible work @chainhelen! The original 32-bit issue was opened 6 years ago. It's great to see progress being made to add 32-bit support to Delve!
> 
> So, I tried it out for a sample binary (at rev [c63ae07](https://github.com/go-delve/delve/commit/c63ae072bcfcf418a6af772e906738b78ff7435b)) and first I made the mistake of running `dlv` compiled for my native machine architecture (AMD64) when trying to debug a 32-bit ELF sample. This resulted in the wrong pointer size being used (as expected).
> 
> ```
> $ dlv exec ./sample
> goarch: amd64
> elfFile.Machine: EM_386
> bi.Arch.PtrSize(): 8
> panic: runtime error: slice bounds out of range [16:12]
> 
> goroutine 7 [running]:
> github.com/go-delve/delve/pkg/dwarf/frame.parseFDE(0xc0001da000, 0xbd1f68)
> 	/home/u/goget/src/github.com/go-delve/delve/pkg/dwarf/frame/parser.go:88 +0x2d4
> ```
> 
> For reference, the out-of-bounds access happens on this line (and it is expected to happen as I had dlv compiled for 64-bit, but debugged a 32-bit program):
> 
> From `pkg/dwarf/frame/parser.go:88`:
> 
> ```go
> 	ctx.frame.Instructions = r[2*ctx.ptrSize:]
> ```
> 
> As the 32-bit support was implemented for the `native` package I tried to rebuild `dlv` with `GOARCH=386`, but then ran into the following build error:
> 
> ```
> $ GOOS=linux GOARCH=386 go install -v github.com/go-delve/delve/cmd/dlv
> github.com/go-delve/delve/pkg/proc/native
> # github.com/go-delve/delve/pkg/proc/native
> goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:66:13: undefined: ptraceGetTls
> goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:83:51: undefined: ptraceGetRegset
> goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:97:16: undefined: processVmWrite
> goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:112:9: undefined: processVmRead
> ```
> 
> Is this to be expected?
> 
> Once more, thanks for working on this! It's tremendously useful to have a debugger for Go written in Go with support for the various architectures that Go compiles down to.
> 
> Cheerful regards,
> Robin

Please notice cgo when cross compile.

chainhelen:
> Incredible work @chainhelen! The original 32-bit issue was opened 6 years ago. It's great to see progress being made to add 32-bit support to Delve!
> 
> So, I tried it out for a sample binary (at rev [c63ae07](https://github.com/go-delve/delve/commit/c63ae072bcfcf418a6af772e906738b78ff7435b)) and first I made the mistake of running `dlv` compiled for my native machine architecture (AMD64) when trying to debug a 32-bit ELF sample. This resulted in the wrong pointer size being used (as expected).
> 
> ```
> $ dlv exec ./sample
> goarch: amd64
> elfFile.Machine: EM_386
> bi.Arch.PtrSize(): 8
> panic: runtime error: slice bounds out of range [16:12]
> 
> goroutine 7 [running]:
> github.com/go-delve/delve/pkg/dwarf/frame.parseFDE(0xc0001da000, 0xbd1f68)
> 	/home/u/goget/src/github.com/go-delve/delve/pkg/dwarf/frame/parser.go:88 +0x2d4
> ```
> 
> For reference, the out-of-bounds access happens on this line (and it is expected to happen as I had dlv compiled for 64-bit, but debugged a 32-bit program):
> 
> From `pkg/dwarf/frame/parser.go:88`:
> 
> ```go
> 	ctx.frame.Instructions = r[2*ctx.ptrSize:]
> ```
> 
> As the 32-bit support was implemented for the `native` package I tried to rebuild `dlv` with `GOARCH=386`, but then ran into the following build error:
> 
> ```
> $ GOOS=linux GOARCH=386 go install -v github.com/go-delve/delve/cmd/dlv
> github.com/go-delve/delve/pkg/proc/native
> # github.com/go-delve/delve/pkg/proc/native
> goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:66:13: undefined: ptraceGetTls
> goget/src/github.com/go-delve/delve/pkg/proc/native/register_linux_386.go:83:51: undefined: ptraceGetRegset
> goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:97:16: undefined: processVmWrite
> goget/src/github.com/go-delve/delve/pkg/proc/native/threads_linux.go:112:9: undefined: processVmRead
> ```
> 
> Is this to be expected?
> 
> Once more, thanks for working on this! It's tremendously useful to have a debugger for Go written in Go with support for the various architectures that Go compiles down to.
> 
> Cheerful regards,
> Robin

Please notice cgo when cross compile.

mewmew:
> Please notice cgo when cross compile.

Thanks! You were right, I had forgotten to set `CGO_ENABLED`. Now the build completes successfully.

```
$ CGO_ENABLED=1 GOOS=linux GOARCH=386 go install -v github.com/go-delve/delve/cmd/dlv
```

I tried using `dlv` on the 32-bit sample and it worked perfectly.

Thanks!

PR #1885, opened 2020-02-21
logflags: Add pid to server-listening-at message

polinasok:
This change makes it easier to get to the process id of the delve server when killing it manually, especially since SIGINT from Ctrl-C gets sent to the debugged process.


aarzilli:
The "API server listening at" message gets parsed by frontends and therefore can not be changed.

polinasok:
I didn't realize this. Thank you for the info. Any objections to make this its line then?

aarzilli:
If you want to add it to the "DAP server listening" line, that's fine. Otherwise I think it should go through the logger, probably enabled by --log-output=debugger. To be honest I've always just used psgrep for this.

polinasok:
I have used ps/pgrep myself, but found it quite annoying when I had more than one dlv running. I encountered this both with my dap testing and with --headless, so figured it would be best to make the change for both. It might be annoying to have to remember to use the log flags to get to this though. And if we do, it should be under log-output=rpc/dap, not debugger since we would be printing these on server creation, not debugger creation, wouldn't you agree?

aarzilli:
> It might be annoying to have to remember to use the log flags to get to this though.

We don't have any other mandatory logging except errors and the "API server listening" message which exists only so that front ends can connect.

> And if we do, it should be under log-output=rpc/dap, not debugger since we would be printing these on server creation, not debugger creation, wouldn't you agree?

Yes, you have a point.

derekparker:
@polinasok just wanted to ping you on this, there's a few CI runs that have failed. Are you still working on this?

PR #1889, opened 2020-02-25
service/dap: clean shutdown sequence in server

eliben:
Using a stop channel to signal clean shutdown to the server's goroutines. Also updates the tests to trigger `disconnectChan`.

This helps avoid spurious error logs when the client disconnects, when Ctrl+C is pressed, etc.

Updates #1515


aarzilli:
There's a merge conflict, otherwise I'm ok with this.

eliben:
> There's a merge conflict, otherwise I'm ok with this.

Resolved.

PR #1890, opened 2020-02-25
proc/gdbserial,debugger: allow clients to stop a recording

aarzilli:
```
proc/gdbserial,debugger: allow clients to stop a recording

Allows Delve clients to stop a recording midway by sending a Command('halt')
request.

This is implemented by changing debugger.New to start recording the process
on a separate goroutine while holding the processMutex locked. By locking
the processMutex we ensure that almost all RPC requests will block until the
recording is done, since we can not respond correctly to any of them.
API calls that do not require manipulating or examining the target process,
such as "IsMulticlient", "SetApiVersion" and "GetState(nowait=true)" will
work while we are recording the process.

Two other internal changes are made to the API: both GetState and Restart
become asynchronous requests, like Command. Restart because this way it can
be interrupted by a Command(halt) request if the rerecord option is passed.
GetState because clients need a call that will block until the recording is
compelted and can also be interrupted with a Command(halt)

Clients that are uninterested in allowing the user to stop a recording can
ignore this change, since eventually they will make a request to Delve that
will block until the recording is completed.

Clients that wish to support this feature must:

1. call GetState(nowait=false) after connecting to Delve, before any call
that would need to manipulate the target process
2. allow the user to send a Command(halt) request during the initial
GetState call
3. allow the user to send a Command(halt) request during any subsequent
Restart(rerecord=true) request (if supported).

Implements #1747

```

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -122,6 +122,15 @@ func (t *Term) Close() {
 func (t *Term) sigintGuard(ch <-chan os.Signal, multiClient bool) {
 	for range ch {
 		t.starlarkEnv.Cancel()
+		state, err := t.client.GetStateNonBlocking()
+		if err == nil && state.Recording {
+			fmt.Printf("received SIGINT, stopping recording (will not forward signal)\n")
+			err := t.client.StopRecording()
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "%v\n", err)
+			}
+			continue
+		}
 		if multiClient {
 			answer, err := t.line.Prompt("Would you like to [s]top the target or [q]uit this client, leaving the target running [s/q]? ")
 			if err != nil {
@@ -215,6 +224,10 @@ func (t *Term) Run() (int, error) {
 	}
 
 	var lastCmd string
+
+	// Ensure that the target process is neither running nor recording by
+	// making a blocking call.
+	_, _ = t.client.GetState()
 
 	for {
 		cmdstr, err := t.promptForInput()
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -19,6 +19,11 @@ var ErrNotExecutable = proc.ErrNotExecutable
 type DebuggerState struct {
 	// Running is true if the process is running and no other information can be collected.
 	Running bool
+	// Recording is true if the process is currently being recorded and no other
+	// information can be collected. While the debugger is in this state
+	// sending a StopRecording request will halt the recording, every other
+	// request will block until the process has been recorded.
+	Recording bool
 	// CurrentThread is the currently selected debugger thread.
 	CurrentThread *Thread `json:"currentThread,omitempty"`
 	// SelectedGoroutine is the currently selected goroutine
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"go/parser"
+	"os"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -43,6 +44,9 @@ type Debugger struct {
 
 	running      bool
 	runningMutex sync.Mutex
+
+	stopRecording func() error
+	recordMutex   sync.Mutex
 }
 
 // Config provides the configuration to start a Debugger.
@@ -133,7 +137,10 @@ func New(config *Config, processArgs []string) (*Debugger, error) {
 			}
 			return nil, err
 		}
-		d.target = p
+		if p != nil {
+			// if p == nil and err == nil then we are doing a recording, don't touch d.target
+			d.target = p
+		}
 		if err := d.checkGoVersion(); err != nil {
 			d.target.Detach(true)
 			return nil, err
@@ -158,6 +165,10 @@ func (d *Debugger) checkGoVersion() error {
 	if !d.config.CheckGoVersion {
 		return nil
 	}
+	if d.isRecording() {
+		// do not do anything if we are still recording
+		return nil
+	}
 	producer := d.target.BinInfo().Producer()
 	if producer == "" {
 		return nil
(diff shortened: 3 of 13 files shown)


aarzilli:
The biggest API problem with this is the GetVersion call. We accidentally allowed GetVersion to access `d.target` without acquiring the processMutex lock to populate the TargetGoVersion field of the response. There are two possible solutions:

1. check if we are recording or running and only access it if we aren't (as implemented in this patch)
2. acquire the processMutex lock

both solutions have problems. (1) is backwards incompatible, because a client could always assume that the TargetGoVersion field is populated, (2) restricts the ability of clients to call GetVersion during the handshake.

aarzilli:
cc @nd

nd:
In addition to GetVersion() GoLand is also affected by changes in RPCServer.Recorded(). At the moment we are relying on RPCServer.Recorded() being blocking and this is changed in the recording scenario. Will try to change that and will post update here.

I wonder if it makes sense to introduce a dedicated 'StopRecording' command to avoid potential confusion? E.g. a client sending 'StopRecording' in non-recording session is definitely wrong. Also sending 'halt' due to some bug in client won't stop recording accidentally. This is minor, we can of course use halt.

aarzilli:
> In addition to GetVersion() GoLand is also affected by changes in RPCServer.Recorded().

Thinking about it, it should probably be left blocking.

> I wonder if it makes sense to introduce a dedicated 'StopRecording' command to avoid potential confusion?

I don't have a preference in this regard.

derekparker:
I like having a more explicit `StopRecording` command as well.

nd:
Checked: GoLand can adopt to this change. We will stop relying on blocking RPCServer.Recorded() behavior and instead will send a blocking State request if recording is in progress. Also the Version() will be sent only when the process is not running/recording.

aarzilli:
Made Command(halt) blocking during a recording, added StopRecording command to stop the recording. Just to be clear, will making the Recording call blocking create problems for GoLand? Should we do anything different about GetApiVersion?

nd:
I think both blocking and and non-blocking Recording call is fine for GoLand: we will call it only in the state when a process is not running and not recording. Current implementation of GetApiVersion is also fine, we will also send it only when it can return information about target version.

aarzilli:
Reverted Recorded to be blocking.

service/api/types.go
```diff
@@ -19,6 +19,11 @@ var ErrNotExecutable = proc.ErrNotExecutable
 type DebuggerState struct {
 	// Running is true if the process is running and no other information can be collected.
 	Running bool
+	// Recoring is true if the process is currently being recorded and no other
+	// information can be collected. While the debugger is in this state
+	// sending a Command("halt") request will halt the recording, every other
```
nd: halt should probably be changed to StopRecording?

aarzilli: Done, thanks.

service/api/types.go
```diff
@@ -19,6 +19,11 @@ var ErrNotExecutable = proc.ErrNotExecutable
 type DebuggerState struct {
 	// Running is true if the process is running and no other information can be collected.
 	Running bool
+	// Recoring is true if the process is currently being recorded and no other
```
derekparker: s/Recoring/Recording/

pkg/terminal/terminal.go
```diff
@@ -213,6 +222,10 @@ func (t *Term) Run() (int, error) {
 		}
 	}
 
+	// Insure that the target process is neither running nor recording by
```
derekparker: s/Insure/Ensure/

service/debugger/debugger.go
```diff
@@ -1476,6 +1582,18 @@ func (d *Debugger) ListPackagesBuildInfo(includeFiles bool) []api.PackageBuildIn
 	return r
 }
 
+// StopRecording stops a recording (if one is in progress)
+func (d *Debugger) StopRecording() error {
+	d.recordMutex.Lock()
+	if d.stopRecording == nil {
+		d.recordMutex.Unlock()
```
derekparker: `defer d.recordMutex.Unlock()`

aarzilli:
@sourya: I think deepsource should run its tests on the merge of the PR with master rather than on the PR commit otherwise it will show errors for things that were fixed on master already but don't appear on the PR branch:

https://github.com/go-delve/delve/pull/1890/checks?check_run_id=521373326

this is what happened before I rebased this PR. This would be especially a problem for projects using a merge-commit strategy (which delve does not).

service/debugger/debugger.go
aarzilli: Done.

sourya:
> @sourya: I think deepsource should run its tests on the merge of the PR with master rather than on the PR commit.

Hey @aarzilli, we prefer to denote if the code is good by labeling the checks as passed or failed. While we run the analysis on the merge of a PR to master as well, we think it is essential to detect issues before they are merged into master.

> it will show errors for things that were fixed on master already but don't appear on the PR branch

Thank you for the feedback. I can only think of rebasing as a fix right now. Let me discuss internally with my teammates and see what can be done about it.

> This would be especially a problem for projects using a merge-commit strategy.

We will add support for this soon. Thank you for bringing this to attention.

service/debugger/debugger.go
```diff
@@ -291,6 +369,7 @@ func (d *Debugger) detach(kill bool) error {
 }
 
 var ErrCanNotRestart = errors.New("can not restart this target")
+var ErrNotRecording = errors.New("debugger is not recording")
```
derekparker: nit: Add doc comment to this (and above since we're here).

aarzilli: Done.

aarzilli:
There seems to be some problem with CI, deepsource is flagging preexisting problems as introduced by this PR and Travis isn't testing at all...

derekparker:
@aarzilli yeah, don't know what's going on with deepsource with this.

aarzilli:
> @aarzilli yeah, don't know what's going on with deepsource with this.

It's not just this, it's doing it with everything now.

sourya:
Hey @aarzilli, we are looking into the issue. We pushed a new release today, and maybe it has something to do with it.

sourya:
We've fixed the issue now. There are some duplicate issues that had popped up due to the bug. We'll clear them in a few hours.

PR #1894, opened 2020-02-25
pkg,service: Use PCToFunc instead of PCToLine if it only needs func

chainhelen:


chainhelen:
Appveyor show pending but has succeeded in fact. [smile]

PR #1896, opened 2020-02-27
test: fix probabilistic error when teststepconcurrent

chainhelen:
Ensure other goroutines has finished when we test `step loop` after
`concurrent`. Or it enter other goroutine probably that cause CI failed
sometime. 

---
fo example about ci test : 
[291401517](https://travis-ci.com/go-delve/delve/jobs/291401517)
[291203525](https://travis-ci.com/go-delve/delve/jobs/291203525)
and so on.

if my understanding is wrong, please point out and i will close.

### DIFF
--- a/_fixtures/teststepconcurrent.go
+++ b/_fixtures/teststepconcurrent.go
@@ -2,8 +2,8 @@ package main
 
 import (
 	"fmt"
-	"time"
 	"sync"
+	"time"
 )
 
 var v int = 99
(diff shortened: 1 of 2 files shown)


aarzilli:
The error with TestStepConcurrent* is actually exposing a real bug with our linux backend. Doing this would fix the error by neutering the test (the loop is meant to be run in parallel with the goroutines to produce concurrenty).

I've researched this bug and my understanding is that we've always had this bug and optimizing the debugger to be faster made it more frequent, particularly c272212baa3b3ff81408b1c6f323c8ad8b14e93a.

As far as I can tell the actual fix for this is part of #1874, specifically the commit about only setting breakpoints on threads that received SIGTRAP. I've never been able to reproduce the TestStepConcurrent bug on my laptop but the other commit of #1874 (avoiding calls to Thread.Status) actually made the bug much more frequent (about 0.5% of the runs on my laptop), enough that I could debug it.

chainhelen:
Oh, I get it. The stopped thread should be same as previous if I just step.

PR #1897, opened 2020-02-27
pkg/logflags: Display warning when unknown component for --log-output

chainhelen:


pkg/logflags/logflags.go
```diff
@@ -177,8 +180,11 @@ func Setup(logFlag bool, logstr string, logDest string) error {
 			fnCall = true
 		case "minidump":
 			minidump = true
-			// If adding another value, do make sure to
-			// update "Help about logging flags" in commands.go.
+		default:
+			if !unKnownHasPrint { // Just print first unknown component.
```
derekparker: I don't think we need to have this restriction, I don't mind printing a warning for each unknown log item.

pkg/logflags/logflags.go
```diff
@@ -177,8 +180,11 @@ func Setup(logFlag bool, logstr string, logDest string) error {
...
+				fmt.Fprintf(os.Stderr, "Warning: unknown %q for '--log-output', run 'dlv help log' for usage.\n", logcmd)
```
derekparker: nit: `Warning: unknown log output value %s for ...`

pkg/logflags/logflags.go
chainhelen: I have try, it looks ok. I will change and rebase.

pkg/logflags/logflags.go
chainhelen: I remove `for '--log-output'` that the same meaning as `log output value` , avoid too long one line.

chainhelen:
I will change and rebase.

PR #1899, opened 2020-02-27
dlv_test: add a basic test for dap command

polinasok:
Updates #1515


polinasok:
It doesn't look like the failure is related. Is there a way for me to rerun Travis without a commit?

chainhelen:
@polinasok Don't worry it.  You can see #1896 for detail from aarzilli  about the failed case. Just wait #1874 to be merged.

PR #1901, opened 2020-02-28
service/dap: Add support for debug and test modes

polinasok:
This change also:
-- refactors building helper functions from commands.go into its own gobuild package.
-- makes processing untyped launch args (that are implementation-specific and not part of the DAP spec) more robust, verified by additional tests

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -21,8 +21,8 @@ func Remove(path string) {
 	}
 }
 
-// OptFlags generates default build flags to turn off optimization and inlining.
-func OptFlags(args []string) []string {
+// optflags generates default build flags to turn off optimization and inlining.
+func optflags(args []string) []string {
 	// after go1.9 building with -gcflags='-N -l' and -a simultaneously works.
 	// after go1.10 specifying -a is unnecessary because of the new caching strategy,
 	// but we should pass -gcflags=all=-N -l to have it applied to all packages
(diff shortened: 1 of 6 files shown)


pkg/gobuild/gobuild.go
```diff
@@ -0,0 +1,73 @@
...
+// This can be used to remove the temporary binary generated for the session.
+func Remove(path string) {
+	err := os.Remove(path)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "could not remove %v: %v\n", path, err)
+	}
+}
+
+// OptFlags generates default build flags to turn off optimization and inlining.
+func OptFlags(args []string) []string {
```
aarzilli: As far as I can tell this is only called inside gobuild, it doesn't need to be exported.

polinasok: Good point. Fixed.

service/dap/server.go
```diff
@@ -286,34 +292,71 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
...
+			s.sendInternalErrorResponse(request.Seq, err.Error())
+			return
+		}
+
+		buildFlags, ok := request.Arguments["buildFlags"].(string)
+		if !ok {
+			buildFlags = ""
+		}
+
+		if mode == "debug" {
```
derekparker: nit: Switch here instead?

PR #1904, opened 2020-02-28
proc/native/linux: try to use process_vm_readv/writev

cuviper:
This change adds `ProcessVmRead` and `ProcessVmWrite` wrappers around
the syscalls `process_vm_readv` and `process_vm_writev`, available since
Linux 3.2. These follow the same permission model as `ptrace`, but they
don't actually require being attached, which means they can be called
directly from any thread in the debugger. They also use `iovec` to write
entire blocks at once, rather than having to peek/poke each `uintptr`.

These wrappers are used in `Thread.ReadMemory` and `WriteMemory`, still
falling back to `ptrace` if that fails for any reason.  Notably,
`process_vm_writev` respects memory protection, so it can't modify
read-only memory like `ptrace`. This frequently occurs when writing
breakpoints in read-only `.text`, so to avoid a lot of wasted `EFAULT`
calls, we only try `process_vm_writev` for larger writes.

### DIFF
--- a/pkg/proc/native/ptrace_linux.go
+++ b/pkg/proc/native/ptrace_linux.go
@@ -49,3 +49,31 @@ func PtracePeekUser(tid int, off uintptr) (uintptr, error) {
 	}
 	return val, nil
 }
+
+// ProcessVmRead calls process_vm_readv
+func ProcessVmRead(tid int, addr uintptr, data []byte) (int, error) {
+	len_iov := uint64(len(data))
+	local_iov := sys.Iovec{Base: &data[0], Len: len_iov}
+	remote_iov := sys.Iovec{Base: (*byte)(unsafe.Pointer(addr)), Len: len_iov}
+	p_local := uintptr(unsafe.Pointer(&local_iov))
+	p_remote := uintptr(unsafe.Pointer(&remote_iov))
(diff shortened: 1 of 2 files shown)


cuviper:
I'm not sure how to add any test case for this, other than making sure nothing regressed. I did try attaching to delve with `strace -e ptrace,process_vm_readv,process_vm_writev -c ...`. Here are results while debugging a very simple test program, just `b main.main; c; c; q`.

On master:
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100.00    0.003562           8       421         4 ptrace
```

On this PR:
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 88.34    0.001273          16        77         6 ptrace
 11.66    0.000168           7        24         2 process_vm_readv
```

cuviper:
@derekparker suggested that this might particularly benefit conditional breakpoints -- seems so!

On master:
```
BenchmarkConditionalBreakpoints-4              1        6057932939 ns/op
```

On this PR:
```
BenchmarkConditionalBreakpoints-4              1        4051294019 ns/op
```

PR #1906, opened 2020-03-02
proc/gdbserial: disable async preemption on macOS

### LINKED ISSUE #1905: Turn off asyncpreempt on MacOS
Running a binary built with Go 1.14 and the async preemption causes significant slowdown on MacOS (see https://github.com/golang/go/issues/37528). We already disable this by default on Windows, so I don't think it would hurt to also disable on MacOS.

aarzilli:
```
proc/gdbserial: disable async preemption on macOS

Disables async preemption on macOS

Fixes #1905

```

### DIFF
--- a/pkg/proc/proc.go
+++ b/pkg/proc/proc.go
@@ -7,8 +7,10 @@ import (
 	"go/ast"
 	"go/constant"
 	"go/token"
+	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/go-delve/delve/pkg/goversion"
 )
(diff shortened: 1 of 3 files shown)


pkg/proc/proc.go
```diff
@@ -845,3 +847,17 @@ func setAsyncPreemptOff(p *Target, v int64) {
 	err = scope.setValue(asyncpreemptoffv, newConstant(constant.MakeInt64(v), scope.Mem), "")
 	logger.Warnf("could not set asyncpreemptoff %v", err)
 }
+
+// DisableAsyncPreemptEnv returns a process environment (like os.Environ)
+// where asyncpreemptoff is set to 1.
+func DisableAsyncPreemptEnv() []string {
```
derekparker: We already disable async preemption by writing the value we want in the process memory at debug.asyncpreemptoff, why add this method in as well?

aarzilli: Because that happens before any runtime code runs, if asyncpreemptoff in GODEBUG is set it will overwrite any change we make (on the other hand if we are attaching the value of GODEBUG will do nothing -- unless we attached very early in the process life).

derekparker: Ack, sorry that's my fault, I misread this patch thinking this was net new, but we were already disabling this in the env, which makes complete sense. Apologies for the noise.

PR #1907, opened 2020-03-03
Got basic backtracing working on ARM64 running Linux.

MichaelPenk:
Got an improved version of backtracing going for the ARM64 version of GO
running under Linux.  It now handles exception records for GO versions
1.12, 1.13, and 1.14 (and probably also works for earlier version),
and is less dependent upon the GO runtime code.

Also fixed a bug that would cause bactraces to fail around 12% of the time.

Fixed a bug that caused most arguments to GO routines to be printed out
incorrectly.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -118,6 +118,13 @@ func (scope *EvalScope) Locals() ([]*Variable, error) {
 			addr := afterLastArgAddr(vars)
 			if addr == 0 {
 				addr = uintptr(scope.Regs.CFA)
+				// The address needed on the ARM64 by the Dwarf lookup routines
+				// is offset from scope.Regs.CFA by the length of a 64-bit
+				// pointer.  This is corrected here.  It probably would be better
+				// to store the corrected address in scope.Regs.CFA.
+				if _, isARM64 := scope.BinInfo.Arch.(*ARM64); isARM64 {
+					addr = addr + uintptr(scope.PtrSize())
+				}
 			}
 			addr = uintptr(alignAddr(int64(addr), val.DwarfType.Align()))
 			val = newVariable(val.Name, addr, val.DwarfType, scope.BinInfo, scope.Mem)
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -5,11 +5,20 @@ import (
 	"errors"
 	"fmt"
 	"go/constant"
+	"sort"
 
 	"github.com/go-delve/delve/pkg/dwarf/frame"
 	"github.com/go-delve/delve/pkg/dwarf/op"
 	"github.com/go-delve/delve/pkg/dwarf/reader"
 )
+
+// Define a list of key indicies.  This is used to call the sort package.
+
+type keyList []uint64
+
+func (a keyList) Len() int           { return len(a) }
+func (a keyList) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a keyList) Less(i, j int) bool { return a[i] < a[j] }
 
 // This code is partly adapted from runtime.gentraceback in
 // $GOROOT/src/runtime/traceback.go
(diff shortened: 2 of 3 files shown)


pkg/proc/stack.go
```diff
@@ -438,13 +447,24 @@ func (it *stackIterator) advanceRegs() (callFrameRegs op.DwarfRegisters, ret uin
 	// implicit.
 	// See also the comment in dwarf2_frame_default_init in
 	// $GDB_SOURCE/dwarf2-frame.c
-	if _, ok := it.bi.Arch.(*ARM64); ok {
+	_, isARM64 := it.bi.Arch.(*ARM64)
+	if isARM64 {
```
aarzilli: Do `callFrameRegs.AddReg(callFrameRegs.SPRegNum, cfareg)` here instead.

pkg/proc/eval.go
```diff
@@ -118,6 +118,13 @@ func (scope *EvalScope) Locals() ([]*Variable, error) {
 			addr := afterLastArgAddr(vars)
 			if addr == 0 {
 				addr = uintptr(scope.Regs.CFA)
+				// The address needed on the ARM64 by the Dwarf lookup routines
```
aarzilli: If CFA was wrong then a lot of things wouldn't work so the suggestion that it should be changed is inappropriate.

pkg/proc/stack.go
```diff
@@ -455,12 +475,15 @@ func (it *stackIterator) advanceRegs() (callFrameRegs op.DwarfRegisters, ret uin
 				it.err = err
 			} else {
 				ret = reg.Uint64Val
+				if isARM64 && it.bi.GOOS == "linux" {
```
aarzilli: This shouldn't be done here but in (*ARM64).SwitchStack (ret is saved in it.frame.Ret).

pkg/proc/stack.go
MichaelPenk: The new code variable "isARM64" was only added on the assumption that it is probably cheaper to refer to this variable later rather than re-execute "it.bi.Arch.(*ARM64)" later.

pkg/proc/eval.go
MichaelPenk: I was intending the comment only as a possible reminder to look at all usages of CFA at some time in the future.  It seems odd to use CFA on the AMD64 architecture but to use CFA + 8 on the AMD64.  I'm fine with the comment being removed.

pkg/proc/stack.go
MichaelPenk: If the code is put into (*ARM64).SwitchStack, the backtrace no longer works.  In particular, the "retaddr = uint64(it.regs.CFA + regRule.Offset)" assignment in "advanceRegs" computes the wrong value, and it seems tricky to correct this value later.

pkg/proc/stack.go
chainhelen: `callFrameRegs.AddReg(callFrameRegs.SPRegNum, cfareg)` is better here, because it has same abstract logic in diffrent arch.  
You can see this [pr](https://github.com/go-delve/delve/pull/1884/files#diff-2f765e651083ae7ea60fe0b2437bd7b4L441-R442) about supporting 386 linux.

aarzilli: The isARM64 variable is fine, it's the `if` I have an issue with (yes, it's not your fault it exists but while we are here we might as well fix it).

pkg/proc/eval.go
aarzilli: This piece of code is used to work around a bug in the compiler where sometimes the arguments of optimized functions do not get an address recorded in debug_info. It almost never runs. Conversely the value of CFA is used to calculate the value of every other variable as well as implementing step over. If it was wrong almost nothing would work.

pkg/proc/stack.go
aarzilli: If I'm not mistaken, retaddr gets stored in addrret and only used to do a check related to stack barriers, a feature that isn't present in recent versions of Go and wouldn't matter in the case of exception frames anyway.

pkg/proc/stack.go
MichaelPenk: OK, will fix it.  I'm busy for the next few days, but should be able to issue a corrected pull request soon.

pkg/proc/stack.go
MichaelPenk: I ran a few more checks, and still can't get the code to work when it is placed in SwitchStack.  I've got some other work to do for the next few days, but should be able to get back to this early next week.  I'll keep you posted.

derekparker:
Closing due to lack of updates.

PR #1911, opened 2020-03-04
go.mod: Commit go.mod changes

derekparker:
When running the test suite off of master I get go.mod changes. Ran go
mody tidy again as well.


aarzilli:
This is very strange. I ran the tests with 1.14 and 1.13 and neither added ginkgo, and the others, back to the dependencies. It doesn't make sense that it should be there because I can pass -mod=vendor and the tests all compile and pass, but ginkgo is not inside vendor! And doing a `grep -R ginkgo` returns zero hits (vendor included).

aarzilli:
Also the version of golang.org/x/sys is changed but there are no updates in vendor.

derekparker:
Hmm... I was running 1.14rc1, I upgraded to 1.14 official release and I'm not seeing the go.mod file change anymore. Must have been something locally for me.

PR #1917, opened 2020-03-09
terminal: add display command

aarzilli:
```
terminal: add display command

Implements #1256

terminal: divide commands into categories

There are too many commands, for clarity they should be divided into
categories when printing and generating documentation.

```

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -374,7 +374,7 @@ If locspec is omitted edit will open the current source file in the editor, othe
 
 		{aliases: []string{"examinemem", "x"}, group: dataCmds, cmdFn: examineMemoryCmd, helpMsg: `Examine memory:
 
-    examinemem [-fmt <format>] [-len <length>] <address>
+	examinemem [-fmt <format>] [-len <length>] <address>
 
 Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal),.
 Length is the number of bytes (default 1) and must be less than or equal to 1000.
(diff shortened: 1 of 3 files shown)


pkg/terminal/command.go
```diff
@@ -1075,6 +1087,7 @@ func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string, sho
 		if shouldPrintFile {
 			printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
 		}
+		t.onStop()
```
derekparker: For these cases with multiple returns we should just `defer t.onStop()`.

aarzilli: Good idea, done.

chainhelen:
Ex:
```
package main

import (
	"fmt"
)

func main() {
	s := 100   
	p := 101         // set breakpoint and continue here.
	fmt.Printf("%d %d\n", s, p)
}
```

```
(dlv) display -a s
0: s = 100
(dlv) display -a p
1: p = 4236625  

(dlv) c
100 101
0: s = error Process 32653 has exited with status 0
1: p = error Process 32653 has exited with status 0
Process 32653 has exited with status 0

(dlv) n
Process 32653 has exited with status 0

(dlv) s
Process 32653 has exited with status 0

(dlv) si
0: s = error Process 32653 has exited with status 0
1: p = error Process 32653 has exited with status 0
Process 32653 has exited with status 0

(dlv) c
0: s = error Process 32653 has exited with status 0
1: p = error Process 32653 has exited with status 0
Process 32653 has exited with status 0
```

It looks some diffrents about `c`/`s`/`si`/`n` after process existed. 
And I think `dlv` could not display any info of  cmd `display` when process existed.

aarzilli:
Fixed, thanks.

PR #1919, opened 2020-03-10
*: Remove appveyor, use travis windows support

derekparker:
Travis has windows CI support, so drop Appveyor and just use Travis instead. Takes our CI system load from 3 to 2.


aarzilli:
No sure why it's failing on windows on travis-ci

aarzilli:
Should we limit the tests with windows to only 1.14 like it was before? I don't think testing the full matrix is all that useful (most bugs are going to be either os/arch specific or go-version specific) and with linux/arm64 and linux/i386 it's starting to get out of hand (I'd cut those to the last version too).

derekparker:
@aarzilli PTAL. I reduced the matrix down to Linux amd64 runs all 3 supported versions + tip, every other combo runs only latest supported + tip. The yaml file looks a little scary now, but we cut down out CI usage.

derekparker:
Once this merges we can just delete the appveyor webhook.

PR #1922, opened 2020-03-11
terminal: add `-l` prompt on goroutines help

chainhelen:
Update: #1879

-----  

ps: 
In fact, I think the prompt of `goroutines` cmd maybe too long. 

How about this? 
```
goroutines [-u (default)| -r | -g | -s ](optional location) [-t (stack trace)] [-l (labels)]
``` 
the whole help info :
```
List program goroutines.

  goroutines [-u (default)| -r | -g | -s ](optional location) [-t (stack trace)] [-l (labels)]

  -u  displays location of topmost stackframe in user code
  -r  displays location of topmost stackframe (including frames inside private runtime functions)
  -g  displays location of go instruction that created the goroutine
  -s  displays location of the start function
  -t  displays goroutine's stacktrace
  -l  displays goroutine's labels

If no flag is specified the default is -u.
```

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -301,7 +301,7 @@ Aliases: gr
 ## goroutines
 List program goroutines.
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [ -t (stack trace)]
+	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [-t (stack trace)] [-l (labels)]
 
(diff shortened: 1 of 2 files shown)


chainhelen:
I notice that `Cirrus CI` not run, but I don't konw why. 
I test with two accounts on other project, it is ok.  

[create pr here](https://github.com/chainhelen/godbg/pull/5)  and the same [.cirrus.yml](https://github.com/chainhelen/godbg/blob/master/.cirrus.yml) as `delve`.

PR #1931, opened 2020-03-14
proc,dwarf: cache debug.Entry objects

aarzilli:
```
proc,dwarf: cache debug.Entry objects

Instead of rescanning debug_info every time we want to read a function
(either to find inlined calls or its variables) cache the tree of
dwarf.Entry that we would generate and use that.

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	5164689165 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	4817425836 ns/op

Updates #1549

```

### DIFF
--- a/pkg/dwarf/reader/reader.go
+++ b/pkg/dwarf/reader/reader.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/go-delve/delve/pkg/dwarf/godwarf"
 	"github.com/go-delve/delve/pkg/dwarf/op"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/dwarf/reader/reader.go
```diff
@@ -320,52 +290,6 @@ func (reader *Reader) NextCompileUnit() (*dwarf.Entry, error) {
...
-			break
-		}
-	}
-
-	return compositeEntry(r), entry.Offset
-}
-
 // InlineStackReader provides a way to read the stack of inlined calls at a
 // specified PC address.
 type InlineStackReader struct {
```
derekparker: We should be able to get rid of this type completely now, correct?

aarzilli: Done.

PR #1935, opened 2020-03-17
*: Fix go vet complaints

derekparker:
Run `go vet` on project, fix complaints.


aarzilli:
Is the [WIP] still in effect or can we merge this?

derekparker:
@aarzilli no I just forgot to remove it.

PR #1941, opened 2020-03-19
Exclude `_fixtures` for DeepSource issues

sourya:


derekparker:
I assume this is preferable to adding it to `test_patterns` correct? Should I close the other PR in favor of this change? On my end this makes more sense as we don't want to analyze that code at all.

sourya:
> I assume this is preferable to adding it to `test_patterns` correct? Should I close the other PR in favor of this change? On my end this makes more sense as we don't want to analyze that code at all.

Yes, in that case, this would be the preferred PR. You can close the other one.

PR #1947, opened 2020-03-19
travis.yml: Add go_import_path

derekparker:
This makes it so that CI will still work on forks, if necessary.


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: go
 sudo: required
+go_import_path: github.com/go-delve/delve
```
aarzilli: This should also be added to scripts/gen-travis.go

PR #1948, opened 2020-03-20
proc: fix switch to goroutine stack

aarzilli:
```
proc: fix switch to goroutine stack

When switching to the goroutine stack the stack iterator registers
might not have any entry for the BP register, make sure to add it
instead of just trying to change it.

```


aarzilli:
Saw this problem as a CI failure on i386.

PR #1949, opened 2020-03-21
pkg/config: add `disassemble-flavor` option for config 

### LINKED ISSUE #415: Allow user to specify syntax flavor in disassemble command
Should have a flag to specify output syntax flavor of assembly in the `disassemble` command.

chainhelen:
Allow user to specify output syntax flavor of assembly in the
disassemble command.

Close #415

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1914,6 +1914,18 @@ func disassCommand(t *Term, ctx callContext, args string) error {
 		rest = argv[1]
 	}
 
+	flavor := api.IntelFlavour
+	if t.conf != nil && t.conf.DisassembleFlavor != nil {
+		switch *t.conf.DisassembleFlavor {
+		case "go":
+			flavor = api.GoFlavour
+		case "gnu":
+			flavor = api.GNUFlavour
+		default:
+			flavor = api.IntelFlavour
+		}
+	}
+
 	var disasm api.AsmInstructions
 	var disasmErr error
 
@@ -1923,7 +1935,7 @@ func disassCommand(t *Term, ctx callContext, args string) error {
 		if err != nil {
 			return err
 		}
-		disasm, disasmErr = t.client.DisassemblePC(ctx.Scope, locs[0].PC, api.IntelFlavour)
+		disasm, disasmErr = t.client.DisassemblePC(ctx.Scope, locs[0].PC, flavor)
 	case "-a":
 		v := split2PartsBySpace(rest)
 		if len(v) != 2 {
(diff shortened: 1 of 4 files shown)


pkg/terminal/command.go
```diff
@@ -1898,55 +1899,74 @@ func (c *Commands) sourceCommand(t *Term, ctx callContext, args string) error {
 	return c.executeFile(t, args)
 }
 
-var disasmUsageError = errors.New("wrong number of arguments: disassemble [-a <start> <end>] [-l <locspec>]")
+var disasmUsageError = errors.New("wrong number of arguments: disassemble [-a <start> <end>] [-l <locspec>] [-f <flavor>]")
 
 func disassCommand(t *Term, ctx callContext, args string) error {
-	var cmd, rest string
+	v := strings.FieldsFunc(args, func(c rune) bool {
```
aarzilli: The argument to the `-l` is a location expression which could be a Go expression, you can't split the entire argument list, you have to do it one by one until you encounter `-l` and then stop.

pkg/terminal/command.go
```diff
@@ -1898,55 +1899,106 @@ func (c *Commands) sourceCommand(t *Term, ctx callContext, args string) error {
...
+				return i
+			}
+		}
+	}
+	return -1
+}
 
-func disassCommand(t *Term, ctx callContext, args string) error {
-	var cmd, rest string
+func nextDisasmArgs(args string, splitStrs []string) (string, string, error) {
```
aarzilli: I think we should use `split2PartsBySpace` instead of doing this.

chainhelen: > I think we should use `split2PartsBySpace` instead of doing this.

Do this because I notice go allow this expression `main      .       main` for `-l`. 
so 
```
disasm   -l    main         .   main     -f    intel 
``` 
is ok.  

Even 
```
// main.go
package main

import (
	"fmt"
)

func main() {
	fmt   .       Printf("hello world\n")
}
```
`go run main.go` run ok.

aarzilli: Yes but consider this:

```
(dlv) disasm -l m["a          string    -f      "]
```

aarzilli:
Thinking more about this, wouldn't it be better to have it as an option in the configuration file instead of a flag? Presumably nobody wants to switch between different syntax flavors.

chainhelen:
En, will this pr be merged or closed?

derekparker:
@chainhelen please update the default config output text in `writeDefaultConfig` in `pkg/config/config.go`.

chainhelen:
Thx, done and reabse.

PR #1950, opened 2020-03-23
go.mod: update dependencies and go version

moricho:
This PR update dependencies and version up Go 1.11 => 1.13.


aarzilli:
Also the vendor directory should be updated, too. Appveyor used to fail if it wasn't, we should put in something that does.

derekparker:
Closing as updates have happened already via other patches.

PR #1956, opened 2020-03-24
debugger: ClearBreakpoint should clear a logical breakpoint

### LINKED ISSUE #1955: `clear` only delete one breakpoint of a logical breakpoint once a time.
Delve Debugger
Version: 1.4.0
Build: 697310fc296883780f1a57cf162a8c68ebb187d3

go version go1.13 linux/amd64 

linux/amd64

4. What did you do?
set breakpoint at an inline function, then clear it.

the breakpoint should be deleted.

6. What did you see instead?
have to `clear` many time as the inline function appear.

aarzilli:
```
debugger: ClearBreakpoint should clear a logical breakpoint

Clear all physical breakpoints associated with a logical breakpoint in
ClearBreakpoint.

Fixes #1955

```

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -1,6 +1,7 @@
 package debugger
 
 import (
+	"bytes"
 	"debug/dwarf"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


service/debugger/debugger.go
```diff
@@ -557,14 +557,59 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
 	d.processMutex.Lock()
 	defer d.processMutex.Unlock()
 
-	var clearedBp *api.Breakpoint
-	bp, err := d.target.ClearBreakpoint(requestedBp.Addr)
+	var bps []*proc.Breakpoint
+	var remaddrs []string
+	var errs []error
+	addrnum := 0
```
derekparker: Do we need this? Can't we infer from `len(bps) == 0` that we weren't able to clear any of the breakpoints?

service/debugger/debugger.go
```diff
@@ -557,14 +557,59 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
...
+	case 1:
+		err = errs[0]
+	default:
+		err = fmt.Errorf("multiple errors deleting breakpoint, first one: %v", errs[0])
+	}
+
 	if err != nil {
-		return nil, fmt.Errorf("Can't clear breakpoint @%x: %s", requestedBp.Addr, err)
+		if len(errs) == addrnum {
+			return nil, fmt.Errorf("Can't clear breakpoint @%x: %s", requestedBp.Addr, err)
```
derekparker: s/Can't/unable to/

service/debugger/debugger.go
```diff
@@ -557,14 +557,59 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
...
+		} else {
+			return nil, fmt.Errorf("Can't clear some of the physical breakpoints (remaining %s): %v", strings.Join(remaddrs, ", "), err)
```
derekparker: s/Can't/unable to/

service/debugger/debugger.go
```diff
@@ -557,14 +557,59 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
...
+	var bps []*proc.Breakpoint
+	var remaddrs []string
+	var errs []error
+	addrnum := 0
+
+	clear := func(addr uint64) {
+		addrnum++
+		bp, err := d.target.ClearBreakpoint(addr)
+		if err != nil {
+			remaddrs = append(remaddrs, fmt.Sprintf("%d", addr))
```
derekparker: Instead of printing the remaining addresses, could we remove this variable altogether and make better use of the error messages returned to inform the user about what went wrong?

derekparker: The error output should have enough context to inform the user what's gone wrong. So instead of collecting all failed addresses and presenting them to the user, can we just present them err.Error() instead for each error that failed? So instead of just showing the first error, maybe print something like (psuedocode): `"the following error(s) occurred when clearing the breakpoint at <addr>"; for (error in errors) { println(error.Error()) }`

aarzilli: How about this?

service/debugger/debugger.go
```diff
@@ -659,14 +660,52 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
...
+		for i, err := range errs {
+			fmt.Fprintf(buf, "%s", err)
+			if i != len(errs)-1 {
+				fmt.Fprintf(buf, ", ")
+			}
+		}
+
+		if len(bps) == 0 {
+			return nil, fmt.Errorf("unable to clear breakpoint %d: %v", requestedBp.ID, buf.String())
+		} else {
```
derekparker, line 698: We can drop this `else` clause.

service/test/integration2_test.go
```diff
@@ -1904,3 +1903,26 @@ func TestStopRecording(t *testing.T) {
 		assertNoError(err, t, "GetState()")
 	})
 }
+
+func TestIssue1955(t *testing.T) {
```
derekparker, line 1907: I'd like to break the pattern of naming tests after Github issues. Let's just call it `TestClearLogicalBreakpoint`.

PR #1957, opened 2020-03-24
pkg/proc: Move proc exec funcs to Target methods

derekparker:
Also renames proc.go to target_exec.go.

This should preserve most of the git blame ease. I also didn't update the receiver from `dbp` to `t`... for the sake of the spirit of this change and minimizing overall churn I'll leave it alone and accept the inconsistency for now.


derekparker:
I'm going to ignore the deepsource failures for this PR, as it's a refactor it wouldn't be a good idea to fix all the resurfaced static analysis issues with this particular PR.

PR #1959, opened 2020-03-25
proc/tests: disable TestIssue414 on linux/386/pie

aarzilli:
```
scripts/make: do not test buildmode=pie on linux/386

The tests have a tendency to hang, see for example:

https://travis-ci.com/github/go-delve/delve/jobs/301890354

```

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -1803,12 +1803,18 @@ func TestIssue396(t *testing.T) {
 }
 
 func TestIssue414(t *testing.T) {
+	if runtime.GOOS == "linux" && runtime.GOARCH == "386" && buildMode == "pie" {
+		t.Skip("test occasionally hangs on linux/386/pie")
+	}
 	// Stepping until the program exits
 	protest.AllowRecording(t)
 	withTestProcess("math", t, func(p *proc.Target, fixture protest.Fixture) {
 		setFileBreakpoint(p, t, fixture.Source, 9)
 		assertNoError(p.Continue(), t, "Continue()")
 		for {
+			pc := currentPC(p, t)
+			f, ln := currentLineNumber(p, t)
+			t.Logf("at %s:%d %#x\n", f, ln, pc)
 			err := p.Step()
 			if err != nil {
 				if _, exited := err.(proc.ErrProcessExited); exited {


derekparker:
I think it's worth investigating why instead of just outright disabling the test.

Any thoughts @chainhelen ?

chainhelen:
```
Tow scripts 

for ((i=1; i<=100; i++)); do go test -v -test.run TestNextConcurrent$ ; done

for ((i=1; i<=100; i++)); do go test -test-buildmode=pie -v -test.run TestNextConcurrent$ ; done
```

(Not only TestNextConcurrent, other cases are same)
In fact, I have tow 386 circumstances. 


1. I386/centos (2Core/4G) remote vps hang (exceed 10m) with high probability on both pie and nopie. I can see the generated exe didn't exit and output nothing when hanging.

2. I386/centos local docker on amd64 is always ok.    

I can't explain this , myabe need to collect more information to find mini reproduction-conditions.

aarzilli:
If your local docker instance only has access to one cpu core that could be the reason you can't reproduce it with it.

chainhelen:
**1.** 
Not one cpu, in my local docker 
```
// top
%Cpu0  :  3.9 us,  0.0 sy,  0.0 ni, 95.1 id,  0.0 wa,  0.0 hi,  1.0 si,  0.0 st
%Cpu1  :  2.9 us,  1.9 sy,  0.0 ni, 95.2 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
%Cpu2  :  2.9 us,  1.0 sy,  0.0 ni, 96.1 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
%Cpu3  :  1.9 us,  1.0 sy,  0.0 ni, 96.1 id,  0.0 wa,  0.0 hi,  1.0 si,  0.0 st
```
```
// main.go
package main

import (
  "fmt"
  "runtime"
)

func main() {
  fmt.Println(runtime.NumCPU())
}

go run main.go

4
```

**2.** 
And I bought the vps so that I could finish 386 linux debugger code, nothing else on that machine. 
I will send ip/user/passward to your email, you can also  check what happend.   


**3.** 
I think I need to build one full 386 linux vm on local pc and  try to find more infomation.

chainhelen:
[ubuntu]
I build i386 ubuntu (2 core cpu) by virtulbox, script never hang. 

[centos]
1.The installation of  centos i386  (http://mirror.aktkn.sg/centos-altarch/7.7.1908/isos/i386/CentOS-7-i386-Everything-1908.iso) with multicore cpu always failed by virtulbox. I don't know why.
2. centos i386 with single core  never hang.

And I see this ci case of  [amd64](https://travis-ci.com/github/go-delve/delve/jobs/314197696) has also hung.

aarzilli:
> And I see this ci case of [amd64](https://travis-ci.com/github/go-delve/delve/jobs/314197696) has also hung.

There seems to be some problems, sometimes, with Travis CI itself. A while back the macOS builder started doing it. It's possible that this is also what's happening with i386 but it seemed more frequent and more deterministic (i.e. it stops always in the same spot, whereas when it's travis it's usually in a random place).

chainhelen:
@derekparker I think it is more likely to be an problem with travis ci itself. (Based on my tests  above) .
  
https://travis-ci.com/github/go-delve/delve/jobs/318730386 again. I approved of this pr because i also have no a better idea.

aarzilli:
I've narrowed it down to just TestIssue414, it seems that that's what's getting stuck. I don't think this warrants further investigation and I'd rather just disable it.

PR #1960, opened 2020-03-25
tests: relax TestStacktraceGoroutine  with Go 1.14

aarzilli:
```
tests: relax TestStacktraceGoroutine  with Go 1.14

The test was always flaky because we can't fully control the state of
all goroutines in the target program, Go 1.14's asynchronous preemption
exacerbates the problem.

See for example:
https://travis-ci.com/github/go-delve/delve/jobs/302407282

This commit relaxes the checks made by the test to avoid irrelevante
flakiness.

```


pkg/proc/proc_test.go
```diff
@@ -946,6 +946,19 @@ func TestStacktraceGoroutine(t *testing.T) {
 		{{10, "main.agoroutine"}},
 	}
 
+	tgtAgouroutineCount := 10
```
derekparker: s/tgtAgouroutineCount/tgtAgoroutineCount

aarzilli: Done.

PR #1963, opened 2020-03-26
*: un-export unnecessarily public symbols

aarzilli:
```
*: un-export unnecessarily public symbols

vendor: check vendor directory

Remove gitignore rule preventing vendor/modules.txt file from being
commited, add check that the vendor directory is synchronized.

```

### DIFF
--- a/pkg/dwarf/frame/entries.go
+++ b/pkg/dwarf/frame/entries.go
@@ -53,7 +53,7 @@ func (fde *FrameDescriptionEntry) EstablishFrame(pc uint64) *FrameContext {
 
 type FrameDescriptionEntries []*FrameDescriptionEntry
 
-func NewFrameIndex() FrameDescriptionEntries {
+func newFrameIndex() FrameDescriptionEntries {
 	return make(FrameDescriptionEntries, 0, 1000)
 }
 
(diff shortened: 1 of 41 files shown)


aarzilli:
I imagine we should probably discuss the _Process/_Thread thing. I didn't want to use `thread` because we use that as a variable name a lot and it would be confusing to read. _Process is for simmetry with _Thread, there would be no problem with using `process`.

derekparker:
> I imagine we should probably discuss the _Process/_Thread thing. I didn't want to use `thread` because we use that as a variable name a lot and it would be confusing to read. _Process is for simmetry with _Thread, there would be no problem with using `process`.

How about `gdbthread`, `nativethread`, etc?

aarzilli:
And nativeprocess/gdbprocess or just process? What about natthread instead of nativethread?

derekparker:
> And nativeprocess/gdbprocess or just process? What about natthread instead of nativethread?

Yeah, for consistency let's name `*process` the same way. And I prefer `native*` instead of `nat*`.

aarzilli:
nativeProcess/nativeThread/gdbProcess/gdbThread?

derekparker:
> nativeProcess/nativeThread/gdbProcess/gdbThread?

That sgtm.

derekparker:
Also, while we're here... we should be able to un-export all of the flag variables in `cmd/dlv/cmds/commands.go`.

PR #1971, opened 2020-03-28
terminal/command: Add 'reload' command

### LINKED ISSUE #1551: Feature: Add reload command to refresh code being debugged
1. What version of Delve are you using (`dlv version`)? 1.2.0
2. What version of Go are you using? (`go version`)? 1.12.1 darwin/amd64
3. What operating system and processor architecture are you using? OS x/intel
4. What did you do? I ran delve to debug an issue and set some breakpoints, found the issue and updated my source code.  Afterwards, in the same delve instance, I issued a restart command
5. What did you expect to see? N/A
6. What did you see instead? new source files and old code running.

The current behavior is similar to how gdb works.  After you edit the code, gdb will show the new source files but run old code.  This can be handy if you want to avoid recompiling because your disk is excessively slow.

However, in gdb you can issue a ```file "executable.exe"``` command.  At this point it reloads the executable from disk and you now have updated machine code.  You can issue a ```restart``` and verify the changes you made are working as expected, with your existing breakpoints and traces still in place.

The ```file``` command wouldn't make sense for delve because of delves ability to run the code without an executable file on disk.  How about adding a ```reload``` command that triggers delve to read sources from disk and recompile?  Is this sort of thing possible? hard? easy? foolhardy?

alexsaezm:
These changes add the 'reload' command, which allows us to rebuild the project
and start the debugging session again. Currently, if the project's code is
updated while debugging it, Delve shows the new source code, but it's still
running the old one. With 'reload', the whole binary is rebuilt, and the
process starts again.

Fixes #1551

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -754,7 +754,7 @@ func execute(attachPid int, processArgs []string, conf *config.Config, coreFile 
 				Foreground:           headless && tty == "",
 				Packages:             dlvArgs,
 				BuildFlags:           buildFlags,
-				Kind:                 kind,
+				ExecuteKind:          kind,
 				DebugInfoDirectories: conf.DebugInfoDirectories,
 				CheckGoVersion:       checkGoVersion,
 				TTY:                  tty,
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -391,11 +391,11 @@ func dapCmd(cmd *cobra.Command, args []string) {
 		if workingDir != "" {
 			fmt.Fprintf(os.Stderr, "Warning: working directory ignored with dap; launch requests must specify full program path\n")
 		}
-		DlvArgs, TargetArgs := splitArgs(cmd, args)
-		if len(DlvArgs) > 0 {
+		dlvArgs, targetArgs := splitArgs(cmd, args)
+		if len(dlvArgs) > 0 {
 			fmt.Fprintf(os.Stderr, "Warning: debug arguments ignored with dap; specify via launch/attach request instead\n")
 		}
-		if len(TargetArgs) > 0 {
+		if len(targetArgs) > 0 {
 			fmt.Fprintf(os.Stderr, "Warning: program flags ignored with dap; specify via launch/attach request instead\n")
 		}
 
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -210,7 +210,7 @@ or later, -gcflags="-N -l" on earlier versions of Go.`,
 			return nil
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			os.Exit(execute(0, args, conf, "", executingExistingFile))
+			os.Exit(execute(0, args, conf, "", debugger.ExecutingExistingFile, args, buildFlags))
 		},
 	}
 	execCommand.Flags().StringVar(&tty, "tty", "", "TTY to use for the target program")
@@ -312,7 +312,7 @@ https://github.com/mozilla/rr
 			},
 			Run: func(cmd *cobra.Command, args []string) {
 				backend = "rr"
-				os.Exit(execute(0, []string{}, conf, args[0], executingOther))
+				os.Exit(execute(0, []string{}, conf, args[0], debugger.ExecutingOther, args, buildFlags))
 			},
 		}
 		rootCommand.AddCommand(replayCommand)
@@ -440,8 +440,8 @@ func debugCmd(cmd *cobra.Command, args []string) {
 			return 1
 		}
 		defer gobuild.Remove(debugname)
-		processArgs := append([]string{debugname}, TargetArgs...)
-		return execute(0, processArgs, conf, "", executingGeneratedFile)
+		processArgs := append([]string{debugname}, targetArgs...)
+		return execute(0, processArgs, conf, "", debugger.ExecutingGeneratedFile, dlvArgs, buildFlags)
 	}()
 	os.Exit(status)
 }
@@ -465,21 +465,21 @@ func traceCmd(cmd *cobra.Command, args []string) {
 		var regexp string
 		var processArgs []string
 
-		DlvArgs, TargetArgs := splitArgs(cmd, args)
+		dlvArgs, TargetArgs := splitArgs(cmd, args)
 
 		if traceAttachPid == 0 {
-			var dlvArgsLen = len(DlvArgs)
+			var dlvArgsLen = len(dlvArgs)
 
 			if dlvArgsLen == 1 {
 				regexp = args[0]
-				DlvArgs = DlvArgs[0:0]
+				dlvArgs = dlvArgs[0:0]
 			} else if dlvArgsLen >= 2 {
 				if traceExecFile != "" {
 					fmt.Fprintln(os.Stderr, "Cannot specify package when using exec.")
 					return 1
 				}
-				regexp = DlvArgs[dlvArgsLen-1]
-				DlvArgs = DlvArgs[:dlvArgsLen-1]
+				regexp = dlvArgs[dlvArgsLen-1]
+				dlvArgs = dlvArgs[:dlvArgsLen-1]
 			}
 
 			debugname := traceExecFile
@@ -590,9 +590,9 @@ func testCmd(cmd *cobra.Command, args []string) {
 			return 1
 		}
 		defer gobuild.Remove(debugname)
-		processArgs := append([]string{debugname}, TargetArgs...)
+		processArgs := append([]string{debugname}, targetArgs...)
 
-		return execute(0, processArgs, conf, "", executingGeneratedTest)
+		return execute(0, processArgs, conf, "", debugger.ExecutingGeneratedTest, dlvArgs, buildFlags)
 	}()
 	os.Exit(status)
 }
@@ -603,11 +603,11 @@ func attachCmd(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(os.Stderr, "Invalid pid: %s\n", args[0])
 		os.Exit(1)
 	}
-	os.Exit(execute(pid, args[1:], conf, "", executingOther))
+	os.Exit(execute(pid, args[1:], conf, "", debugger.ExecutingOther, args, buildFlags))
 }
 
 func coreCmd(cmd *cobra.Command, args []string) {
-	os.Exit(execute(0, []string{args[0]}, conf, args[1], executingOther))
+	os.Exit(execute(0, []string{args[0]}, conf, args[1], debugger.ExecutingOther, args, buildFlags))
 }
 
 func connectCmd(cmd *cobra.Command, args []string) {
@@ -616,7 +616,7 @@ func connectCmd(cmd *cobra.Command, args []string) {
 		fmt.Fprint(os.Stderr, "An empty address was provided. You must provide an address as the first argument.\n")
 		os.Exit(1)
 	}
-	os.Exit(connect(addr, nil, conf, executingOther))
+	os.Exit(connect(addr, nil, conf, debugger.ExecutingOther))
 }
 
 // waitForDisconnectSignal is a blocking function that waits for either
@@ -657,7 +657,7 @@ func splitArgs(cmd *cobra.Command, args []string) ([]string, []string) {
 	return args, []string{}
 }
 
-func connect(addr string, clientConn net.Conn, conf *config.Config, kind executeKind) int {
+func connect(addr string, clientConn net.Conn, conf *config.Config, kind debugger.ExecuteKind) int {
 	// Create and start a terminal - attach to running instance
 	var client *rpc2.RPCClient
 	if clientConn != nil {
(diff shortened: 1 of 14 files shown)


service/debugger/debugger.go
```diff
@@ -420,6 +428,12 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 		p, err = d.recordingRun(run)
 		d.recordingDone()
 	} else {
+		if rebuild {
```
aarzilli: This should go before the `if recorded` line, otherwise it won't have an effect on recorded targets.

Also, breakpoints that can't be set by (file, line) should be discarded in the `for _, oldBp := ...` loop below.

service/debugger/debugger.go
```diff
@@ -420,6 +428,12 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
...
+			err := gobuild.GoBuild(d.processArgs[0], d.config.Packages, d.config.BuildFlags)
```
aarzilli: Some error message should be returned if the debugger was started with `attach` or `exec`, also this doesn't work if the debugger was started with `test`.

pkg/terminal/command.go
```diff
@@ -142,6 +142,7 @@ For live targets the command takes the following forms:
 If newargv is omitted the process is restarted (or re-recorded) with the same argument vector.
 If -noargs is specified instead, the argument vector is cleared.
 `},
+		{aliases: []string{"reload"}, group: runCmds, cmdFn: c.reload, allowedPrefixes: revPrefix, helpMsg: "Refresh code."},
```
aarzilli: I think `rebuild` would be a better name.

cmd/dlv/cmds/commands.go
```diff
@@ -47,6 +47,10 @@ var (
 	InitFile string
 	// BuildFlags is the flags passed during compiler invocation.
 	BuildFlags string
+	// DlvArgs DlvArgs contains debug arguments.
+	DlvArgs []string
```
derekparker: I don't think we need to add these here and export them.

cmd/dlv/cmds/commands.go
```diff
@@ -477,20 +481,20 @@ func traceCmd(cmd *cobra.Command, args []string) {
...
+					if err := gobuild.GoBuild(debugname, DlvArgs, BuildFlags); err != nil {
 						fmt.Fprintf(os.Stderr, "%v\n", err)
 						return 1
 					}
 				}
 				defer gobuild.Remove(debugname)
 			}
 
-			processArgs = append([]string{debugname}, targetArgs...)
+			processArgs = append([]string{debugname}, TargetArgs...)
```
derekparker: (I can't add a review comment on the appropriate line cause it wasn't changed so I'll add it here)

You must fill in the new struct members you added to `service.Config` below (`Packages` and `BuildFlags`). Currently they are not being passed through and are always empty, it appears.

alexsaezm:
I added a test for the feature, and also move the `ExecuteKind` from `cmds` to `debugger`. Also, I renamed the feature because yeah, makes total sense to call it rebuild.

service/debugger/debugger.go
```diff
@@ -434,6 +434,15 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 	}
 	var p *proc.Target
 	var err error
+
+	// TODO executeKind from command.go
+	if rebuild && d.canRestart() {
```
aarzilli, line 443: The code above has a `if !d.canRestart() { return ... }` why do you need to check this here too?

Also the loop that sets breakpoints below should discard the ones that can't be set by file/line.

go.mod
```diff
@@ -20,6 +20,6 @@ require (
 	go.starlark.net v0.0.0-20190702223751-32f345186213
 	golang.org/x/arch v0.0.0-20190927153633-4e8777c89be4
 	golang.org/x/sys v0.0.0-20190626221950-04f50cda93cb
-	golang.org/x/tools v0.0.0-20191127201027-ecd32218bd7f
+	golang.org/x/tools v0.0.0-20200415034506-5d8e1897c761
```
aarzilli, line 23: I don't think this needs an updated version of golang.org/x/tools so this should be reverted. If I'm wrong and it does the vendor directory should also be updated.

alexsaezm: Yeah, for sure it is not necessary. My bad, I think I update this while doing the rebase of the branch for some reason.

service/debugger/debugger.go
alexsaezm: You are totally right. I removed the d.canRestart.

Can you please clarify a little more the loop part? I don't really understand what you mean.

aarzilli: In this function there's a loop, `for _, oldBp := ...` that restores breakpoints. It either does it by file:line (`if len(oldBp).File > 0`) or by address (`else`). We shouldn't try to set them by address because addresses have changed after rebuilding.

service/debugger/debugger.go
```diff
@@ -420,6 +439,20 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 	}
 	var p *proc.Target
 	var err error
+
+	if rebuild {
+		// We cannot restart processes that we don't know how to build. If
+		// d.config.Kind is different from ExecutingGeneratedFile that means we
+		// didn't build the binary and we are just running it
+		if d.config.Kind == ExecutingExistingFile {
+			return nil, fmt.Errorf("cannot restart process Delve did not create")
```
aarzilli, line 448: s/cannot restart/cannot rebuild/

cmd/dlv/cmds/commands.go
```diff
@@ -387,11 +387,11 @@ func dapCmd(cmd *cobra.Command, args []string) {
 		if workingDir != "" {
 			fmt.Fprintf(os.Stderr, "Warning: working directory ignored with dap; launch requests must specify full program path\n")
 		}
-		dlvArgs, targetArgs := splitArgs(cmd, args)
-		if len(dlvArgs) > 0 {
+		DlvArgs, TargetArgs := splitArgs(cmd, args)
```
aarzilli, line 390: why the capitalization?

service/debugger/debugger.go
```diff
@@ -420,6 +439,20 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
...
+		}
+		err := gobuild.GoBuild(d.processArgs[0], d.config.Packages, d.config.BuildFlags)
```
aarzilli, line 450: This should call GoTestBuild if `Kind == ExecutingGeneratedTest`.

service/test/integration2_test.go
```diff
@@ -74,6 +75,9 @@ func startServer(name string, buildFlags protest.BuildFlags, t *testing.T) (clie
 		Debugger: debugger.Config{
 			Backend:        testBackend,
 			CheckGoVersion: true,
+			Packages:       []string{fixture.Source},
+			BuildFlags:     "", // build flags can be an empty string here because the only test that uses it, does not set special flags.
+			Kind:           debugger.ExecutingGeneratedTest,
```
aarzilli, line 80: This should say ExecutingGeneratedFile, it doesn't really matter but for consistency that's what they are.

service/test/integration2_test.go
```diff
@@ -110,6 +114,7 @@ func TestRunWithInvalidPath(t *testing.T) {
 		APIVersion:  2,
 		Debugger: debugger.Config{
 			Backend: testBackend,
+			Kind:    debugger.ExecutingGeneratedTest,
```
aarzilli, line 117: Also ExecutingGeneratedFile.

service/debugger/debugger.go
alexsaezm: Thanks for the explanation! I updated the loop to avoid the case

service/debugger/debugger.go
alexsaezm: Fixed

cmd/dlv/cmds/commands.go
alexsaezm: Because I totally mixed that up :disappointed: . Changed

service/test/integration2_test.go
alexsaezm: Makes sense. I changed it.

service/rpccommon/server.go
```diff
@@ -116,6 +116,8 @@ func (s *ServerImpl) Run() error {
 		Foreground:           s.config.Foreground,
 		DebugInfoDirectories: s.config.DebugInfoDirectories,
 		CheckGoVersion:       s.config.CheckGoVersion,
+		Packages:             s.config.Packages,
```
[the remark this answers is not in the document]
alexsaezm: It should work now :)

service/debugger/debugger.go
```diff
@@ -420,6 +439,21 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 	}
 	var p *proc.Target
 	var err error
+
+	if rebuild {
+		// We cannot build processes that we don't know how to build. If
+		// d.config.Kind is different from ExecutingGeneratedFile that means we
+		// didn't build the binary and we are just running it
+		if d.config.Kind == ExecutingExistingFile {
+			err = gobuild.GoTestBuild(d.processArgs[0], d.config.Packages, d.config.BuildFlags)
```
aarzilli, line 448: This is still wrong:

ExecutingExistingFile, ExecutingOther → it can't be rebuilt at all (either an separately built executable file or a core dump/recording)
ExecutingGeneratedFile → it can be rebuilt by GoBuild
ExecutingGeneratedTest → it can be rebuilt by GoTestBuild

it doesn't make sense to call GoTestBuild on ExecutingExistingFile and it doesn't make sense to call GoBuild on anything other than ExecutingGeneratedFile.

alexsaezm: Updated. Thanks for the explanation :)

aarzilli:
I'm sorry I forgot to tell you this yesterday but there is a conflict between your PR and something that has already been merged into master that needs to be solved. You need to do a git rebase.

alexsaezm:
I rebased the fork to keep it up to date with master.

pkg/terminal/command.go
```diff
@@ -143,6 +143,7 @@ For live targets the command takes the following forms:
 If newargv is omitted the process is restarted (or re-recorded) with the same argument vector.
 If -noargs is specified instead, the argument vector is cleared.
 `},
+		{aliases: []string{"rebuild"}, group: runCmds, cmdFn: c.rebuild, allowedPrefixes: revPrefix, helpMsg: "Rebuild the code."},
```
derekparker, line 146: We need more documentation here. Add in more information about how it will recompile code, etc, etc...

pkg/terminal/command.go
```diff
@@ -1073,6 +1074,18 @@ func printcontextNoState(t *Term) {
...
 }
 
+func (c *Commands) rebuild(t *Term, ctx callContext, args string) error {
+	if ctx.Prefix == revPrefix {
+		return c.rewind(t, ctx, args)
+	}
+	defer t.onStop()
+	discarded, err := t.client.Restart(true)
+	if len(discarded) > 0 {
+		fmt.Printf("Some of the breakpoints could not be recovered")
```
derekparker, line 1084: nit: no need for capitalization. Also, maybe rephrase to `not all breakpoints could be restored`.

service/config.go
```diff
@@ -36,4 +36,17 @@ type Config struct {
 
 	// DisconnectChan will be closed by the server when the client disconnects
 	DisconnectChan chan<- struct{}
+
+	// TTY is passed along to the target process on creation. Used to specify a
+	// TTY for that process.
+	TTY string
```
derekparker, line 42: This flag can be removed.

service/debugger/debugger.go
```diff
@@ -421,6 +440,26 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
...
+				return nil, fmt.Errorf("could not rebuild process: %s", err)
+			}
+		case ExecutingGeneratedTest:
+			err = gobuild.GoTestBuild(d.processArgs[0], d.config.Packages, d.config.BuildFlags)
+			if err != nil {
+				return nil, fmt.Errorf("could not rebuild process: %s", err)
+			}
+		default:
+			// We cannot build a process that we didn't start, because we don't know
+			// how it was build.
```
derekparker, line 458: s/build/built/

service/debugger/debugger.go
```diff
@@ -450,6 +489,10 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 			}
 			createLogicalBreakpoint(p, addrs, oldBp)
 		} else {
+			// Avoid setting a breakpoint based on address when rebulding
```
derekparker, line 492: s/rebulding/rebuilding/

aarzilli:
> @aarzilli did you have a chance to take another look as well?

I have now. I don't have anything to add. For the rebuild command documentation I think something like this would work:

```
Rebuild the target executable and restarts it. Does not work if the executable was not built by delve.
```

cmd/dlv/cmds/commands.go
```diff
@@ -465,7 +465,7 @@ func traceCmd(cmd *cobra.Command, args []string) {
 		var regexp string
 		var processArgs []string
 
-		dlvArgs, targetArgs := splitArgs(cmd, args)
+		dlvArgs, TargetArgs := splitArgs(cmd, args)
```
derekparker, line 468: This does not need to be exported.

service/config.go
```diff
@@ -36,4 +36,13 @@ type Config struct {
 
 	// DisconnectChan will be closed by the server when the client disconnects
 	DisconnectChan chan<- struct{}
+
+	// Packages contains the packages that we are debugging.
```
derekparker, line 40: These three fields can be removed because we've added them to the `debugger.Config` as well, which we include within this config already anyway.

service/debugger/debugger.go
```diff
@@ -98,6 +108,15 @@ type Config struct {
...
 	TTY string
+
+	// Packages contains the packages that we are debugging.
+	Packages []string
+
+	// BuildFlags contains the flags passed to the compiler.
+	BuildFlags string
+
+	// ExecuteKind contains the kind of the executed program.
+	Kind ExecuteKind
```
derekparker, line 119: nit: `Kind` is a little vague, maybe we stutter a bit and name this field `ExecuteKind` as well.

alexsaezm: Done. `ExecuteKind` sounds perfect. I updated a few more places where it appears.

PR #1972, opened 2020-03-29
proc: convert Arch into a struct

aarzilli:
```
proc: convert Arch into a struct

Replace the interface type Arch with a struct with the same
functionality.

```

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -5,47 +5,87 @@ import (
 	"github.com/go-delve/delve/pkg/dwarf/op"
 )
 
-// Arch defines an interface for representing a
-// CPU architecture.
-type Arch interface {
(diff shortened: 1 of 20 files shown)


pkg/proc/arch.go
```diff
@@ -5,47 +5,86 @@ import (
...
 
-// Arch defines an interface for representing a
-// CPU architecture.
-type Arch interface {
-	// PtrSize returns the size of a pointer for the architecture.
-	PtrSize() int
-	// MaxInstructionLength is the maximum size in bytes of an instruction.
-	MaxInstructionLength() int
-	// AsmDecode decodes the assembly instruction starting at mem[0:] into asmInst.
+type Arch struct {
```
derekparker: Doc comment for exported type.

aarzilli: Done.

Issue #1973, opened 2020-03-29
nothing put out


PeterlitsZo:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
Delve Debugger
Version: 1.4.0
Build: $Id: 67422e6f7148fa1efa0eac1423ab5594b223d93b $

2. What version of Go are you using? (`go version`)?
go version go1.14.1 linux/amd64

3. What operating system and processor architecture are you using?
Debian Linux(wsl)

4. What did you do?
follow your doc:
> 1. `go get -u github.com/go-delve/delve/cmd/dlv`
> 2. try to edit a test.go:
> ```
> $ cat test.go
> package main
> 
> import "fmt"
> 
> func main () {
>     fmt.Println("hello")
> }
> ```
> 3. try to debug, `dlv debug ./temp.go`

5. What did you expect to see?'
something, such as a error message or "Type 'help' for list of commands...." like your doc say

6. What did you see instead?
nothing happend, it just wait. I do not know what step is worry....

aarzilli:
>  Debian Linux(wsl)

is this windows subsystem for linux 1 or 2? 1 isn't supported and doesn't work.

PeterlitsZo:
I enter `wsl -l -v` and then it show me the usage -- It looks be under wsl 1...
[this](https://scotch.io/bar-talk/trying-the-new-wsl-2-its-fast-windows-subsystem-for-linux) show how to update my wsl. I think I will try.
Thanks.

derekparker:
Closing this due to noted incompatibility above. We may want to document this in some way, however perhaps differently than the above linked PR.

PR #1975, opened 2020-03-30
add message of wsl

PeterlitsZo:
link: issues #1973


PeterlitsZo:
But I think we need to do something for wsl, isn't it?
To be honest, linux do not support my net card, the only way to use linux is wsl. (so... Windows is the best version of linux XD? just kidding)

PR #1976, opened 2020-03-31
service/dap: Change event reason to "entry" when stopping on entry

polinasok:
Existing code matched [vscode-go](https://github.com/microsoft/vscode-go/blob/master/src/debugAdapter/goDebug.ts#L824), setting reason to 'breakpoint', but that's misleading and according to the [spec](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Stopped), there is a special option for entry. Use that option. This impacts what is displayed in the UI ('paused on breakpoint' vs 'paused on entry').
pause-on-entry


polinasok:
cc @eliben

polinasok:
Matching PR to vscode-go was accepted as well: https://github.com/microsoft/vscode-go/pull/3150.

PR #1977, opened 2020-03-31
pkg/proc: Use consistent naming for Target receiver

derekparker:


derekparker:
I realize this will create `git blame` churn, but I really value the consistency when reading the code base. We also have other PRs in flight which will add some blame churn, so overall I feel that keeping a high standard of consistency and readability in the code base should be of greater importance overall.

aarzilli:
I don't want this much diff churn in those functions, I don't think it's worth it just for consistency. I'd be fine with changing the name of the receiver if the first line of the function was `dbp := t` and the rest of the body was intact.

derekparker:
> I don't want this much diff churn in those functions, I don't think it's worth it just for consistency. I'd be fine with changing the name of the receiver if the first line of the function was `dbp := t` and the rest of the body was intact.

I really don't understand this argument, and I don't think we benefit from doing git blame driven development. I value clean, readable code over ease of git blame. I take pride in the quality, consistency, and readability of the code, not how easy it is to run `git blame` on a file.

Also in the end, we're not losing any history... it's all still there. Worst case scenario you add a couple flags to `git blame` or checkout another revision and `git blame` from there.

aarzilli:
> I really don't understand this argument, and I don't think we benefit from doing git blame driven development. I value clean, readable code over ease of git blame. I take pride in the quality, consistency, and readability of the code, not how easy it is to run git blame on a file.

I like very much to have a clean diff history because it's helpful when trying to determine why a certain code decision was made. I weight this against other concerns, for example: recent mass-unexporting produced a lot of diff garbage but it was mostly limited to function headers, and in return we have cleaner autogenerated docs. Another example: moving next from threads.go was worth it because it doesn't have anything to do with threads.

Names of local variables don't matter much so I'm very biased against changing them.

derekparker:
> I like very much to have a clean diff history because it's helpful when trying to determine why a certain code decision was made. I weight this against other concerns, for example: recent mass-unexporting produced a lot of diff garbage but it was mostly limited to function headers, and in return we have cleaner autogenerated docs. Another example: moving next from threads.go was worth it because it doesn't have anything to do with threads.

I completely understand this, I think being able to do forensic analysis on a code base is important. However this change doesn't outright prevent it, just adds one more step for certain lines.

> Names of local variables don't matter much so I'm very biased against changing them.

I'd half agree normally (I do think proper variable naming is important), but this particular local variable is a method receiver. I know it's all just syntactic sugar over this being the first argument to a function, but I think it looks sloppy to have multiple receiver names. And again, I think more people are going to be looking at the code than they are going to be `git blame`ing it, so I'd really like to optimize for viewing and overall cleanliness.

I'm fine with drawing the line here, it's really the fact that it's a receiver name that has me so determined to make it consistent. I'm not advocating changing every single instance where `Target` is passed into a function to have a consistent variable name, but I think all methods should use the name receiver name.

aarzilli:
> I'd half agree normally (I do think proper variable naming is important), but this particular local variable is a method receiver.

Which is why I was suggesting to change the receiver name and then assign it to `dbp` and continue using the old name inside the function.

derekparker:
Closing this for now.

PR #1979, opened 2020-04-01
*: Better error launching invalid binary format

### LINKED ISSUE #1310: Clarify error when delve attempts to execute a binary built for another os.

```bash
steel@mac:~$dlv version
Delve Debugger
Version: 1.0.0
Build: 9335c54014f987055f327014e3dc8b51465cf414
```
```bash
steel@mac:~$go version
go version go1.10.3 darwin/amd64
```
`macOS High Sierra 10.13.6`

As dumb as it is, I unknowingly ran into this issue today in vscode while attempting to solve another problem. The error message `could not launch process: stub exited while waiting for connection: exit status 0` is a tad generic, but googling it sends you off in a direction that doesn't end up pointing to my stupidity.

Is there some way delve could be update to output a better message in the instance where the binary is not built for the OS?

- Example command(OSX): `env GOOS=linux dlv debug main.go`
- Example command(Linux): `env GOOS=osx dlv debug main.go`

derekparker:
This improves the overall error message when trying to run an executable of an invalid binary format for the host system. Also consolidates some basic sanity checks (is this file an executable?).

Before:

(MacOS):

```
could not launch process: stub exited while waiting for connection: exit status 0
```

(Linux):

```
could not launch process: fork/exec /home/deparker/code/rr-test/__debug_bin: exec format error
```

After:

```
could not launch process: could not launch executable /Users/derekparker/code/delve-test-progs/interface-test/__debug_bin: invalid executable binary format
```

Fixes #1310

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -25,6 +25,17 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	// ErrCanNotRestart is returned when the target cannot be restarted.
+	// This is returned for targets that have been attached to, or when
+	// debugging core files.
+	ErrCanNotRestart = errors.New("can not restart this target")
+
+	// ErrNotRecording is returned when StopRecording is called while the
+	// debugger is not recording the target.
+	ErrNotRecording = errors.New("debugger is not recording")
+)
+
 // Debugger service.
 //
 // Debugger provides a higher level of
@@ -178,6 +189,9 @@ func (d *Debugger) checkGoVersion() error {
 
 // Launch will start a process with the given args and working directory.
 func (d *Debugger) Launch(processArgs []string, wd string) (*proc.Target, error) {
+	if err := verifyBinaryFormat(processArgs[0]); err != nil {
+		return nil, err
+	}
 	switch d.config.Backend {
 	case "native":
 		return native.Launch(processArgs, wd, d.config.Foreground, d.config.DebugInfoDirectories)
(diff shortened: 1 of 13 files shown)


service/debugger/debugger.go
```diff
@@ -176,8 +191,50 @@ func (d *Debugger) checkGoVersion() error {
...
+		return err
+	}
+	defer f.Close()
+
+	fi, err := f.Stat()
+	if err != nil {
+		return err
+	}
+	// check that the argument is an executable file
+	if (fi.Mode()&0111) == 0 {
```
aarzilli, line 206: windows executables don't have the executable bit set, this is what's causing the test to fail.

derekparker: Fixed, I was looking at the sources of exec.LookPath which checked the unix executable bit on godoc but it went straight to the _unix.go file and I wasn't paying attention. I was planning to keep working on this last night so it wasn't still failing in the morning, but it was getting late.

derekparker:
> Also it looks like there is a legit test failure for linux for the new test, but I don't know why.

It's failing on ARM64, it's trying to cross compile from Linux arm64 to Windows arm64, which won't work. I'll fix it.

derekparker:
The error message is different from what I originally posted above, but it's still way better and more consistent than what we had before.

service/debugger/debugger_windows.go
```diff
@@ -14,3 +20,23 @@ func stopProcess(pid int) error {
...
+
+	// Make sure the binary exists and is an executable file
+	if filepath.Base(exePath) == exePath {
+		if _, err := exec.LookPath(exePath); err != nil {
+			return err
+		}
+	}
+
+	if _, err = pe.NewFile(f); err != nil {
+		return proc.ErrNotExecutable
```
aarzilli, line 39: Is this error ever returned anywhere else now? If not it should probably be moved here.

derekparker: No it's not... I think we can get rid of it and just return the `api.ErrNotExecutable` error instead.

PR #1981, opened 2020-04-02
Only load floating point registers when needed

aarzilli:
```
proc/*: only load floating point registers when needed

Changes implementations of proc.Registers interface and the
op.DwarfRegisters struct so that floating point registers can be loaded
only when they are needed.
Removes the floatingPoint parameter from proc.Thread.Registers.
This accomplishes three things:

1. it simplifies the proc.Thread.Registers interface
2. it makes it impossible to accidentally create a broken set of saved
registers or of op.DwarfRegisters by accidentally calling
Registers(false)
3. it improves general performance of Delve by avoiding to load
floating point registers as much as possible

Floating point registers are loaded under two circumstances:

1. When the Slice method is called with floatingPoint == true
2. When the Copy method is called

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	4327350142 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	3852642917 ns/op

Updates #1549

proc: do not convert floating point registers until needed

```

### DIFF
--- a/pkg/dwarf/op/regs.go
+++ b/pkg/dwarf/op/regs.go
@@ -12,18 +12,46 @@ type DwarfRegisters struct {
 	CFA       int64
 	FrameBase int64
 	ObjBase   int64
-	Regs      []*DwarfRegister
+	regs      []*DwarfRegister
 
 	ByteOrder binary.ByteOrder
 	PCRegNum  uint64
 	SPRegNum  uint64
 	BPRegNum  uint64
 	LRRegNum  uint64
+
+	FloatLoadError   error // error produced when loading floating point registers
+	loadMoreCallback func()
 }
 
 type DwarfRegister struct {
 	Uint64Val uint64
 	Bytes     []byte
+}
+
+// NewDwarfRegisters returns a new DwarfRegisters object.
+func NewDwarfRegisters(staticBase uint64, regs []*DwarfRegister, byteOrder binary.ByteOrder, pcRegNum, spRegNum, bpRegNum, lrRegNum uint64) *DwarfRegisters {
+	return &DwarfRegisters{
+		StaticBase: staticBase,
+		regs:       regs,
+		ByteOrder:  byteOrder,
+		PCRegNum:   pcRegNum,
+		SPRegNum:   spRegNum,
+		BPRegNum:   bpRegNum,
+		LRRegNum:   lrRegNum,
+	}
+}
+
+// SetLoadMoreCallback sets a callback function that will be called the
+// first time the user of regs tries to access an undefined register.
+func (regs *DwarfRegisters) SetLoadMoreCallback(fn func()) {
+	regs.loadMoreCallback = fn
+}
+
+// CurrentSize returns the current number of known registers. This number might be
+// wrong if loadMoreCallback has been set.
+func (regs *DwarfRegisters) CurrentSize() int {
+	return len(regs.regs)
 }
 
 // Uint64Val returns the uint64 value of register idx.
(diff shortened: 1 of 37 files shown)


aarzilli:
Looks like I broke arm64, I'll have to look into it.

chainhelen:
En, about
```
Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	4327350142 ns/op

Benchmark after:

BenchmarkConditionalBreakpoints-4   	       1	3852642917 ns/op
```
When I do benchmark (previous similar optimizations), the fluctuation is very big every time and the value is inconspicuous. Of course it must be optimized in theory. 

I just want to konw how to generate reliable data. Maybe remove the maximum, remove the minimum, take the average?

aarzilli:
> Looks like I broke arm64, I'll have to look into it.

Fixed it.

> When I do benchmark (previous similar optimizations), the fluctuation is very big every time and the value is inconspicuous. Of course it must be optimized in theory.
>
> I just want to konw how to generate reliable data. Maybe remove the maximum, remove the minimum, take the average?

I don't get a lot of variance. If you are using 1.14 disable async preemption, that adds a lot of noise, otherwise change the number of iterations in issue1549.go.

pkg/dwarf/op/regs.go
```diff
@@ -12,27 +12,55 @@ type DwarfRegisters struct {
...
+	loadMoreCallback func()
 }
 
 type DwarfRegister struct {
 	Uint64Val uint64
 	Bytes     []byte
 }
 
+// NewDwarfRegisters returns a new DwarfRegisters object.
+func NewDwarfRegisters(staticBase uint64, regs []*DwarfRegister, byteOrder binary.ByteOrder, pcRegNum, spRegNum, bpRegNum, lrRegNum uint64) *DwarfRegisters {
```
derekparker, line 33: This is just a thin wrapper around initializing the struct. All function args become struct members with no processing done by the function. Maybe we just let the callers initialize the struct themselves?

pkg/dwarf/op/regs.go
```diff
@@ -52,10 +80,22 @@ func (regs *DwarfRegisters) Bytes(idx uint64) []byte {
 
 // Reg returns register idx or nil if the register is not defined.
 func (regs *DwarfRegisters) Reg(idx uint64) *DwarfRegister {
-	if idx >= uint64(len(regs.Regs)) {
-		return nil
+	if idx >= uint64(len(regs.regs)) {
+		if regs.loadMoreCallback != nil {
```
derekparker, line 84: nit: I think this would read better as:

```
if regs.loadMoreCallback == nil {
        return nil
}
regs.loadMoreCallback()
regs.loadMoreCallback = nil
if idx >= uint64(len(regs.regs)) {
	return nil
}
```

pkg/dwarf/op/regs.go
```diff
@@ -52,10 +80,22 @@ func (regs *DwarfRegisters) Bytes(idx uint64) []byte {
...
+			regs.loadMoreCallback()
```
derekparker, line 85: I think we should put this into a `regs.loadMore` method which can wrap the logic of checking if the callback is not nil and then setting it to nil once it's been called.

pkg/dwarf/op/regs.go
```diff
@@ -72,12 +112,25 @@ func (regs *DwarfRegisters) BP() uint64 {
...
-		copy(newRegs, regs.Regs)
-		regs.Regs = newRegs
+		copy(newRegs, regs.regs)
+		regs.regs = newRegs
+	}
+	regs.regs[idx] = reg
+}
+
+// ClearRegisters clears all registers except for PC, SP and BP
+func (regs *DwarfRegisters) ClearRegisters() {
```
derekparker, line 124: I think the name should reflect that this is selective in which registers it clears (e.g. `ClearGeneralPurposeRegisters` or something). Also, why be selective?

pkg/dwarf/op/regs.go
derekparker: Ack, makes perfect sense.

PR #1986, opened 2020-04-04
*: Add --tty flag for debug / exec

derekparker:
This flag allows users on UNIX systems to set the tty for the program
being debugged by Delve. This is useful for debugging command line
applications which need access to their own TTY, and also for
controlling the output of the debugged programs so that IDEs may open a
dedicated terminal to show the output for the process.


Documentation/faq.md
```diff
@@ -35,3 +35,13 @@ dlv exec --headless --continue --listen :4040 --accept-multiclient /path/to/exec
 ```
 
 Note that the connection to Delve is unauthenticated and will allow arbitrary remote code execution: *do not do this in production*.
+
+#### How can I use Delve to debug a CLI application?
+
+There are two good ways to go about this. The first is to run your CLI application in a separate
```
aarzilli, line 41: Here we should mention that there's also the option of running a headless instance of delve in a terminal and then connecting to it from a different terminal.

derekparker: Agreed. Also updated the part below, instead of pointing to `reptyr` the docs now point to a little utility I wrote called `ptyme` to create a pty and proxy it.

Documentation/faq.md
```diff
@@ -35,3 +35,16 @@ dlv exec --headless --continue --listen :4040 --accept-multiclient /path/to/exec
 ```
 
 Note that the connection to Delve is unauthenticated and will allow arbitrary remote code execution: *do not do this in production*.
+
+#### How can I use Delve to debug a CLI application?
+
+There are three good ways to go about this. The first is to run your CLI application in a separate
+terminal and then attach to it via `dlv attach`. The other way is to assign the process its own
```
aarzilli, line 42: s/The other/Another/ or The second.

Documentation/faq.md
```diff
@@ -35,3 +35,16 @@ dlv exec --headless --continue --listen :4040 --accept-multiclient /path/to/exec
...
+TTY. This can be done on UNIX systems via the `--tty` flag for the `dlv debug` and `dlv exec` commands.
+
+The second is to run Delve in headless mode via `dlv debug --headless` and then connect to it from
```
aarzilli, line 45: s/The second/The third/

Documentation/faq.md
```diff
@@ -35,3 +35,16 @@ dlv exec --headless --continue --listen :4040 --accept-multiclient /path/to/exec
...
+another terminal. This will place the process being debugged in the foreground and allow it to access
+the terminal TTY.
+
+For the best experience, you should create your own PTY and assign it as the TTY. This can be done
```
aarzilli, line 49: This should probably go next to the paragraph describing `--tty`.

PR #1987, opened 2020-04-05
cmd/dlv: Add support for piping data to cmd stdin

simon3z:
This change adds the support for piping data to the command stdin, as for example:

    $ cat <file> | dlv debug --headless ...

In any case it shouldn't be a problem to pass the existing stdin to the process that is being debugged.

### DIFF
--- a/_fixtures/stdintest/main.go
+++ b/_fixtures/stdintest/main.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+)
+
+func main() {
+	line, _, err := bufio.NewReader(os.Stdin).ReadLine()
+
+	if err != nil {
+		os.Exit(1)
+	}
+
+	fmt.Println("echo:", string(line))
+}
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -253,6 +253,44 @@ func TestContinue(t *testing.T) {
 	cmd.Wait()
 }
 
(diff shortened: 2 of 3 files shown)


simon3z:
Hi @aarzilli I am not sure if this is the most correct fix, I am sure you know this part much better than me, what do you think?

simon3z:
> Not passing stdin to the target process when it isn't foregrounded (i.e. when it isn't running inside a headless instance) is done in purpose, to avoid it interfering with delve's command line.

@aarzilli if it's headless there's no command line AFAIK.

> Incidentally the test isn't testing anything because the code does the same thing before and after this change if --headless is specified.

The point is to pass the stdin to the debugged process not to delve (which indeed you shouldn't have had --headless).

Part of the problem (and here is my doubt) is that the `foreground` bool is done using an irrelevant check (IMO) `isatty.IsTerminal(os.Stdin.Fd())` ...which means that if you pipe a file (as mentioned in the description) it would be `false` even though it's indeed foreground (it just has file pipe attached).

@aarzilli comments / thoughts?

aarzilli:
Yes, I forgot the isatty check for foreground, still this isn't the way to do it. I think if `foreground == true` but `!isatty` we should set `foreground = false` but still pass Stdin to the target process.

Also, if you only intend to implement this for linux you should say so in the commit and gate the test on `runtime.GOOS`.

derekparker:
Going to close this for now as there have not been any updates since Apr 6.

PR #1994, opened 2020-04-06
gdbserial/gdbserver: Dynamically resolve debugserver binary

### LINKED ISSUE #986: lldb-server needs to be installed in $PATH
```
$ dlv version
Delve Debugger
Version: 1.0.0-rc.1
Build: $Id: b6077a6a35c48a31716aad2ad3fdf77bbdf70abd $
```
```
$ go version
go version go1.9.1 darwin/amd64
```
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.3
BuildVersion:	16D32
```
4. What did you do?
```
$ go get -u github.com/derekparker/delve/cmd/dlv
$ dlv exec k6 run samples/http_2.js
```
The Go debugger

6. What did you see instead?
```
could not launch process: exec: "lldb-server": executable file not found in $PATH
```

xyzst:
Instead of hardcoding the absolute path to the Command Line
Tools (CLT) binary, will attempt to resolve the path at the
$PATH, or at the Xcode bundle. If none are available, will
fallback to the default CLT location.

Fixes #986


xyzst:
I understand that unit test are required before it's merged but I want to make sure I am on the right track before I am committed.

If I am on the right track, what's the best approach for testing this type of change in go? My plan is to mock the fs by creating temporary files and directories at test runtime with the `os` package functions.

aarzilli:
I think it would be better if this was done in LLDBLaunch and LLDBAttach rather than in a initializer. Also you should add a log line so that we can find out which program ends up being launched.

xyzst:
@aarzilli thanks for reviewing my PR.

Is this more of a style concern, or is there a behavioral difference when calling the helper function eagerly vs lazily? I am new to go, so I would appreciate an explanation or a pointer to related documentation. 

As for the logging, I will look into it after I finish up my daily work 😃

aarzilli:
> Is this more of a style concern, or is there a behavioral difference when calling the helper function eagerly vs lazily?

It's just going to add to the load time for delve, even if it doesn't get used at all.

The test doesn't work because it doesn't have permission to go write in system folders, I don't think we need a test for this.

xyzst:
@aarzilli thanks for the explanation. Entirely reasonable, will keep that in mind going forward. 

Will update the PR after work hours and remove the tests.

PR #1996, opened 2020-04-07
*: Update freebsd image

derekparker:


derekparker:
@aarzilli yeah if you go to the `checks` tab and click on the CirrusCI section, at the bottom of the page there's a link that says `View more details on Cirrus CI` which gives all the information.

And you're correct, they did update to 1.14.

PR #1997, opened 2020-04-08
proc: better support for C pointers

aarzilli:
```
proc: better support for C pointers

- treat C pointers as arrays
- print 'char *' variables as strings

proc: fix findCompileUnitForOffset when plugins are used

Splits the compileUnits slice between images so that we can search for
an offset inside the debug info of a specific image file.

```

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -75,6 +75,8 @@ const (
 	// the variable is the return value of a function call and allocated on a
 	// frame that no longer exists)
 	VariableFakeAddress
+	// VariableCPrt means the variable is a C pointer
+	VariableCPtr
 )
 
 // Variable represents a variable. It contains the address, name,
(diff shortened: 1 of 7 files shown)


aarzilli:
No longer WIP.

pkg/proc/variables.go
```diff
@@ -1341,9 +1374,27 @@ func readStringValue(mem MemoryReadWriter, addr uintptr, strlen int64, cfg LoadC
 		return "", fmt.Errorf("could not read string at %#v due to %s", addr, err)
 	}
 
-	retstr := *(*string)(unsafe.Pointer(&val))
+	//retstr := *(*string)(unsafe.Pointer(&val))
```
derekparker, line 1377: Dangling comment?

pkg/proc/variables.go
```diff
@@ -1341,9 +1374,27 @@ func readStringValue(mem MemoryReadWriter, addr uintptr, strlen int64, cfg LoadC
...
 
-	return retstr, nil
+	return string(val), nil
+}
+
+func readCStringValue(mem MemoryReadWriter, addr uintptr, cfg LoadConfig) (string, bool, error) {
+	done := false
+	buf := make([]byte, 1)
```
derekparker, line 1384: Reading a single byte at a time until we hit a nul terminator is going to cause a lot of syscall context switching. Might be a premature optimization, but would be preferable to read into a larger buffer and search for the index of the nul terminator instead of going byte by byte.

pkg/proc/variables.go
aarzilli: Yes.

pkg/proc/variables.go
aarzilli: Done. It's a bit complicated because of pagefaults.

PR #1998, opened 2020-04-08
proc: fix findCompileUnitForOffset when plugins are used

aarzilli:
```
proc: fix findCompileUnitForOffset when plugins are used

Splits the compileUnits slice between images so that we can search for
an offset inside the debug info of a specific image file.

```


aarzilli:
Rebased.

PR #2000, opened 2020-04-08
dwarf/op: return register values when they are the only piece

aarzilli:
```
dwarf/op: return register values when they are the only piece

When there is a single piece and it's a register value just return it.
This is important for clang compiled programs which will use DW_OP_regN
to specify the value of the frame base.

```


aarzilli:
Rebased.

PR #2005, opened 2020-04-12
gdbserial/gdbserver: optimize the search of debugserver binary

chainhelen:
Unexport `GetDebugServerAbsolutePath` and avoid unnecessary repeated calls.
Remove `os.Stat` because `Exec.LookPath` has already used `os.Stat`.And Fix
some comments.

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -94,11 +94,17 @@ const (
 
 const heartbeatInterval = 10 * time.Second
 
+var debugserverExecutablePaths = []string{
+	"debugserver",
+	"/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver",
+	"/Applications/Xcode.app/Contents/SharedFrameworks/LLDB.framework/Versions/A/Resources/debugserver",
+}
+
 // ErrDirChange is returned when trying to change execution direction
 // while there are still internal breakpoints set.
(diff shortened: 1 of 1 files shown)


chainhelen:
To be on the safe side, can you help me re-run ci for travis? thx.

aarzilli:
Done.

chainhelen:
Ha ha, so fortunate again.

derekparker:
Hm, yeah in theory this should work. The only difference is before we always returned a path (the last fallback path was never looked up via ExecPath).

chainhelen:
Ok, I want to try to test failed CI for this pr. So don't merge temporarily.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -363,8 +361,8 @@ func LLDBLaunch(cmd []string, wd string, foreground bool, debugInfoDirs []string
 	var listener net.Listener
 	var port string
 	var process *exec.Cmd
-	if _, err := os.Stat(GetDebugServerAbsolutePath()); err == nil {
-		listener, err = net.Listen("tcp", "127.0.0.1:0")
+	if debugserverExecutable := getDebugServerAbsolutePath(); debugserverExecutable != "" {
+		listener, err := net.Listen("tcp", "127.0.0.1:0")
```
derekparker, line 365: This is the cause of the breakage. This variable shadows the `listener` variable above, causing it to never be set, thus we end up with a zero-value listener even though we were able to find the debug server executable.

chainhelen: Thx, done

PR #2007, opened 2020-04-13
*: Consolidate service/debugger config

derekparker:
Embed the debugger config object in the service config object to avoid needless duplication of fields.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -89,7 +89,7 @@ func (s *Server) Stop() {
 		s.conn.Close()
 	}
 	if s.debugger != nil {
-		kill := s.config.AttachPid == 0
+		kill := s.config.Debugger.AttachPid == 0
 		if err := s.debugger.Detach(kill); err != nil {
 			s.log.Error(err)
 		}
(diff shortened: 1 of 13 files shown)


service/dap/server.go
```diff
@@ -354,24 +354,15 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-		DebugInfoDirectories: s.config.DebugInfoDirectories,
-		CheckGoVersion:       s.config.CheckGoVersion,
-		TTY:                  s.config.TTY,
-	}
+	s.config.Debugger.WorkingDir = filepath.Dir(program)
+
+	config := s.config.Debugger
 	var err error
-	if s.debugger, err = debugger.New(config, s.config.ProcessArgs); err != nil {
+	if s.debugger, err = debugger.New(&config, s.config.ProcessArgs); err != nil {
```
chainhelen, line 365: En, maybe we can remove temporary variable `config` and 
```
debugger.New(&s.config.Debugger, s.config.ProcessArgs)
```
is enough.

derekparker: True, that would work. I did this mostly so that the debugger object got a pointer to a copy of the config and not the exact config embedded in `s.config`. Not a big deal either way, I suppose.

PR #2009, opened 2020-04-15
proc: use cached packageVars in proc.(*EvalScope).PackageVariables

aarzilli:
```
proc: use cached packageVars map in PackageVariables

```


aarzilli:
This is in part to avoid rescanning the debug_info section, in part to avoid an having an extra code path that scans debug_info in a different way and also there's a bug in the current version of PackageVariables because we don't reset the reader after reading the undefined type, which means that some global variables get skipped if types and variables are intermixed inside debug_info (which wasn't the case for go when this code was written but is true now).

PR #2011, opened 2020-04-16
Go 1.15 support branch

### LINKED ISSUE #1134: Printing an uninitialized variable prints seemingly random information
With this simple program:

```
package main

import (
        "fmt"
)

func main() {
        branch := "master"
        fmt.Println(branch)
}
```

Breaking on the `branch := master` line gives the following output:

```
$ dlv debug .

(dlv) break a.go:8
Breakpoint 1 set at 0x49c1af for main.main() ./a.go:8

(dlv) c
> main.main() ./a.go:8 (hits goroutine(1):1 total:1) (PC: 0x49c1af)
     3: import (
     4:         "fmt"
     5: )
     6:
     7: func main() {
=>   8:         branch := "master"
     9:         fmt.Println(branch)
    10: }

(dlv) locals
branch = "H�D$8�FH�D$(H���\x00\x00\x00H�\f$�G_\x03\x00H�D$\bH�D$8H�L$(Hǁ�\x00\x00\x00\x00\x00\x00\x00H�\f$H�D$\b\x03...+842350542940 more"

(dlv) step
> main.main() ./a.go:9 (PC: 0x49c1c4)
     4:         "fmt"
     5: )
     6:
     7: func main() {
     8:         branch := "master"
=>   9:         fmt.Println(branch)
    10: }

(dlv) locals
branch = "master"
```

Similar output occurs with other variable types; for example `branch := 42` is `branch = 842350543032` on `locals`.

I understand that this happens because the `branch` variable isn't assigned a value yet, but printing seemingly random information doesn't strike me as a good user experience, and something that might confuse people a lot. I think it would be better to print something like `<uninitialized>` or the type's zero value.

```
$ dlv version
Delve Debugger
Version: 1.0.0

$ go version
go version go1.10 linux/amd64

$ uname -a
Linux arch.arp242.net 4.15.2-2-ARCH #1 SMP PREEMPT Thu Feb 8 18:54:52 UTC 2018 x86_64 GNU/Linux
```

aarzilli:
```
goversion: add Go 1.15 to supported versions

core: support core files for PIEs on windows

Makefile: enable PIE tests on windows/Go 1.15

tests: properly check if cgo is enabled for cgo related tests

proc: silence go vet error

proc: start variable visibility one line after their decl line

In most cases variables shouldn't be visible on their declaration line
because they won't be initialized there.
Function arguments are treated as an exception.

This fix is only applied to programs compiled with Go 1.15 or later as
previous versions of Go did not report the correct declaration line for
variables captured by closures.

Fixes #1134

proc: use cached packageVars in proc.(*EvalScope).PackageVariables

```

### DIFF
--- a/_fixtures/decllinetest.go
+++ b/_fixtures/decllinetest.go
@@ -2,14 +2,15 @@ package main
 
 import (
 	"fmt"
-	"runtime"
 )
 
 func main() {
 	a := 0
-	runtime.Breakpoint()
 	a++
 	b := 0
-	runtime.Breakpoint()
-	fmt.Println(a, b)
+	f1(a, b)
+}
+
+func f1(a, b int) {
+	fmt.Printf("%d %d\n", a, b)
 }
--- a/_fixtures/testvariables.go
+++ b/_fixtures/testvariables.go
@@ -20,8 +20,8 @@ type Nest struct {
 }
 
 func barfoo() {
-	runtime.Breakpoint()
 	a1 := "bur"
+	runtime.Breakpoint()
 	fmt.Println(a1)
 }
 
--- a/_scripts/make.go
+++ b/_scripts/make.go
@@ -70,15 +70,7 @@ func NewMakeCommands() *cobra.Command {
 		Short: "Tests delve",
 		Long: `Tests delve.
 
-Use the flags -s, -r and -b to specify which tests to run. Specifying nothing is equivalent to:
-
-	go run _scripts/make.go test -s all -b default
-	go run _scripts/make.go test -s basic -b lldb    # if lldb-server is installed and Go < 1.14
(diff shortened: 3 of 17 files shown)


nd:
In GoLand we made an exception for loop variables wrt declline and shown them on the line where they are declared. Would it be possible to do the same in Delve or to allow clients to filter vars?

aarzilli:
@derekparker Go 1.15beta1 has been out for a while and rc1 should be close so I think it's time to review this. Mostly it's changes to the call injection code to mach the changes done to the Go runtime: the injected call now gets executed on a different goroutine, we have to figure out the association between call injection goroutine and call execution goroutine by noticing that the thread is the same.

Aside from that I've enabled the PIE tests on windows (that build mode is now available there) and implemented PIE support for windows minidumps (it's just five lines of code).

PR #2016, opened 2020-04-17
proc/gdbserial: replace stdio-path option with stdin/stdout/stderr-path

### LINKED ISSUE #1964: Cannot read from stdin with buffered reader
Go 1.14.1
Delve Debugger
Version: 1.4.0
Build: 8bb93e9ae19589b7ee665225c2206ef118928ae9

Run this code in debug mode and then just continue execution. On MacOS I cannot type any symbol after that.

```
package main

import (
	"bufio"
	"fmt"
	"log"
	"os"
)

func main() {
	fmt.Println("hello")
	rdr := bufio.NewReader(os.Stdin)
	str, err := rdr.ReadString('\n')
	if err != nil {
		log.Fatalf("User input is not a number: %v\n", str)
	}
}

```

related issue: https://youtrack.jetbrains.com/issue/GO-8997

aarzilli:
```
proc/gdbserial: replace stdio-path option with stdin/stdout/stderr-path

Makes target instances running inside debugserver work when stdin is
not a teletyle.

Fixes #1964

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -81,7 +81,6 @@ import (
 	"github.com/go-delve/delve/pkg/logflags"
 	"github.com/go-delve/delve/pkg/proc"
 	"github.com/go-delve/delve/pkg/proc/linutil"
-	isatty "github.com/mattn/go-isatty"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


aarzilli:
I'll have to figure out how to do this better.

PR #2018, opened 2020-04-20
[WIP] pkg/proc: avoid target process leaks.

### LINKED ISSUE #2017: `dlv exec` leave target process running that was not successfully launch caused by missing debug info
Delve Debugger
Version: 1.4.0
Build: $Id: 67422e6f7148fa1efa0eac1423ab5594b223d93b $

go version go1.13.5 linux/amd64

linux/amd64

4. What did you do?
`dlv exec` a stripped binary. 

dlv exit

6. What did you see instead?
dlv exit with error message `could not launch process: could not open debug info` but leaving the target process running.

chainhelen:
Target process should exit when dlv launch failed.

Fix #2017.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -41,7 +41,7 @@ func TestMain(m *testing.M) {
 			}
 		}
 	}
-	os.Exit(m.Run())
+	os.Exit(protest.RunTestsWithFixtures(m))
 }
 
 func assertNoError(err error, t testing.TB, s string) {
@@ -251,6 +251,50 @@ func TestContinue(t *testing.T) {
 		t.Fatalf("error detaching from headless instance: %v", err)
 	}
 	cmd.Wait()
+}
+
+// TestChildProcessExitWhenNoDebugInfo verifies that the child process exits when dlv launch the binary without debug info
+func TestChildProcessExitWhenNoDebugInfo(t *testing.T) {
+	if runtime.GOOS == "darwin" {
+		t.Skip("test skipped on darwin, see https://github.com/go-delve/delve/pull/2018 for details")
+	}
+
+	if _, err := exec.LookPath("ps"); err != nil {
+		t.Skip("test skipped, `ps` not found")
+	}
+
+	dlvbin, tmpdir := getDlvBin(t)
+	defer os.RemoveAll(tmpdir)
+
+	fix := protest.BuildFixture("http_server", protest.LinkStrip)
+
+	// dlv exec the binary file and expect error.
+	if _, err := exec.Command(dlvbin, "exec", fix.Path).CombinedOutput(); err == nil {
+		t.Fatalf("Expected err when launching the binary without debug info, but got nil")
+	}
+
+	// search the running process named fix.Name
+	cmd := exec.Command("ps", "-aux")
+	stdout, err := cmd.StdoutPipe()
+	assertNoError(err, t, "stderr pipe")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("`ps -aux` failed: %v", err)
+	}
+
+	var foundFlag bool
+	scan := bufio.NewScanner(stdout)
+	for scan.Scan() {
+		t.Log(scan.Text())
+		if strings.Contains(scan.Text(), fix.Name) {
+			foundFlag = true
+			break
+		}
+	}
+	cmd.Wait()
+
+	if foundFlag {
+		t.Fatalf("Expected child process exited, but found it running")
+	}
 }
 
 func checkAutogenDoc(t *testing.T, filename, gencommand string, generated []byte) {
--- a/pkg/proc/native/proc_freebsd.go
+++ b/pkg/proc/native/proc_freebsd.go
@@ -58,7 +58,7 @@ func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tt
 	dbp := newProcess(0)
 	defer func() {
 		if err != nil && dbp.pid != 0 {
-			_ = dbp.detach(true)
+			_ = dbp.Detach(true)
 		}
 	}()
 	dbp.execPtraceFunc(func() {
--- a/pkg/proc/native/proc_freebsd.go
+++ b/pkg/proc/native/proc_freebsd.go
@@ -56,6 +56,11 @@ func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tt
 	}
 
 	dbp := newProcess(0)
+	defer func() {
+		if err != nil && dbp.pid != 0 {
+			_ = dbp.detach(true)
+		}
+	}()
 	dbp.execPtraceFunc(func() {
 		process = exec.Command(cmd[0])
 		process.Args = cmd
@@ -85,12 +90,10 @@ func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tt
 	dbp.childProcess = true
 	_, _, err = dbp.wait(process.Process.Pid, 0)
 	if err != nil {
-		_ = dbp.Detach(true)
 		return nil, fmt.Errorf("waiting for target execve failed: %s", err)
 	}
 	tgt, err := dbp.initialize(cmd[0], debugInfoDirs)
 	if err != nil {
-		_ = dbp.Detach(true)
 		return nil, err
 	}
 	return tgt, nil
--- a/pkg/proc/native/proc_freebsd.go
+++ b/pkg/proc/native/proc_freebsd.go
@@ -85,10 +85,12 @@ func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tt
 	dbp.childProcess = true
 	_, _, err = dbp.wait(process.Process.Pid, 0)
 	if err != nil {
+		_ = dbp.Detach(true)
 		return nil, fmt.Errorf("waiting for target execve failed: %s", err)
 	}
 	tgt, err := dbp.initialize(cmd[0], debugInfoDirs)
 	if err != nil {
+		_ = dbp.Detach(true)
 		return nil, err
 	}
 	return tgt, nil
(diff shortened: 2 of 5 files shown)


dlsniper:
Sorry to step in but I have something that's a bit unclear for me, so please correct me if I'm wrong.

From reading #2017, it looks like the user tries to `attach` to a long running process to debug it, however that fails.

If we apply this patch, whenever we fail to attach to a process, that would kill the process.
Debugging a production service with stripped binary information would then kill the long running process, and that's a side-effect that we shouldn't create imho.
Just because I want to attach to a process, it doesn't mean I also want to kill it when I finish debugging.

As an anecdote, sometimes I fail to select the proper Go process from a list of running processes, and often try to connect to Docker. After applying this patch, I would nuke all containers by mistakes, just because I selected the wrong target process to attach to.

aarzilli:
This patch only applies to processes launched by delve, the related issue talks about 'exec' not attach.

chainhelen:
@dlsniper Yes, as aarzilli say , just `exec` not `attach`.

chainhelen:
@derekparker @aarzilli  I am sure this logic take effect, but I don't know how to add this case.


Because there is a issue about here.
```
func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tty string) (*proc.Target, error) {
	defer func() {
		if err != nil && dbp != nil && dbp.pid != 0 {
			_ = killProcess(dbp.pid)
		}
	}()
}
```
Child process doesn't exit directly  after `killProcess` until parent exit.  It looks child process doesn't immediately receive `kill` signal  from os. I don't konw why.

aarzilli:
> It looks child process doesn't immediately receive kill signal from os. I don't konw why.

It could be because it's trace stopped. Doing it by launching another instance of dlv, like the tests in cmd/dlv/dlv_test.go do, might work.

chainhelen:
> > It looks child process doesn't immediately receive kill signal from os. I don't konw why.
> 
> It could be because it's trace stopped. Doing it by launching another instance of dlv, like the tests in cmd/dlv/dlv_test.go do, might work.

Oh, yes. Thx for guidance, I will try.

aarzilli:
Looking more at the Launch code, I think it might be more appropriate to call dbp.Detach(true) in the last `if err != nil` for linux and freebsd. The native/windows code is already doing that so the bug shouldn't affect it.

chainhelen:
Ok, good idea, I will try.

cmd/dlv/dlv_test.go
```diff
@@ -253,6 +255,57 @@ func TestContinue(t *testing.T) {
...
+// TestChildProcessExitWhenNoDebugInfo verifies that the child process exits when dlv launch the binary without debug info
+func TestChildProcessExitWhenNoDebugInfo(t *testing.T) {
+	if _, err := exec.LookPath("ps"); err != nil {
+		t.Skip("test skipped, `ps` not found")
+	}
+
+	dlvbin, tmpdir := getDlvBin(t)
+	defer os.RemoveAll(tmpdir)
+
+	// Make a (good enough) random temporary file name
```
aarzilli, line 267: Instead of copying this here call BuildFixture passing the LinkStrip build flag. You will have to change TestMain to say `os.Exit(support.RunTestsWithFixtures(m))`

chainhelen: OK, I will finish it tomorrow.

chainhelen:
Why does this way of `BuildFixture` on mac say `error: The specified item could not be found in the keychain.`? Here is [ci report](https://travis-ci.com/github/go-delve/delve/jobs/327229448).

aarzilli:
I think that's because we are trying to sign the executable and there is no certificate installed. It probably doesn't have anything to do with your failure.

pkg/proc/native/proc_linux.go
```diff
@@ -162,7 +164,7 @@ func initialize(dbp *nativeProcess) error {
 		}
 		comm = match[1]
 	}
-	dbp.os.comm = strings.Replace(string(comm), "%", "%%", -1)
+	dbp.os.comm = strings.ReplaceAll(string(comm), "%", "%%")
```
chainhelen, line 167: For deepsource.

pkg/proc/native/proc_linux.go
```diff
@@ -362,14 +364,12 @@ func (dbp *nativeProcess) trapWaitInternal(pid int, options trapWaitOptions) (*n
...
 			th.os.running = false
 			return th, nil
-		} else {
-			if err := th.resumeWithSig(int(status.StopSignal())); err != nil {
-				if err == sys.ESRCH {
-					dbp.postExit()
-					return nil, proc.ErrProcessExited{Pid: dbp.pid}
-				}
-				return nil, err
+		} else if err := th.resumeWithSig(int(status.StopSignal())); err != nil {
```
chainhelen, line 367: For deepsource.

aarzilli: As far as I am concerned, you shouldn't worry about deepsource errors if they aren't in code you wrote.

cmd/dlv/dlv_test.go
```diff
@@ -253,6 +253,50 @@ func TestContinue(t *testing.T) {
...
+
+	fix := protest.BuildFixture("http_server", protest.LinkStrip)
+
+	// dlv exec the binary file and expect error.
+	if _, err := exec.Command(dlvbin, "exec", fix.Path).CombinedOutput(); err == nil {
+		t.Fatalf("Expected err when launching the binary without debug info, but got nil")
+	}
+
+	// search the running process named fix.Name
+	cmd := exec.Command("ps", "-aux")
```
qingyunha, line 277: `ps --ppid=dlv_pid` may be more neat.

pkg/proc/native/proc_freebsd.go
```diff
@@ -85,10 +85,12 @@ func Launch(cmd []string, wd string, foreground bool, debugInfoDirs []string, tt
 	dbp.childProcess = true
 	_, _, err = dbp.wait(process.Process.Pid, 0)
 	if err != nil {
+		_ = dbp.Detach(true)
```
derekparker, line 88: This (and all other such occurrences) seem like they would be perfectly suited for a `defer` call. I'd rather something like this:

```
defer func() { 
        if err != nil {
                dbp.Detach(true)
        }
}()
```

chainhelen: Done

aarzilli:
Your new TestChildProcessExitWhenNoDebugInfo seems to fail systematically on freebsd: https://cirrus-ci.com/task/5833222820986880?command=test

Personally I'm ok with just disabling it on that GOOS.

chainhelen:
> Your new TestChildProcessExitWhenNoDebugInfo seems to fail systematically on freebsd: https://cirrus-ci.com/task/5833222820986880?command=test
> 
> Personally I'm ok with just disabling it on that GOOS.

Done

derekparker:
I don't think we should ignore the failure on freebsd. This patch touches the native backend(s), one of which is freebsd, so if that backend is failing for the test this patch introduces, seems like it might be a legit failure.

chainhelen:
It's a very small probability but happened about TestChildProcessExitWhenNoDebugInfo on freebsd.  

I will try some times for ci to judge whether to skip on freebsd.

polinasok:
I came here because of this skip reference:
https://github.com/go-delve/delve/blob/a19931c9d3a22d1c3015281a29910ecb9e6b904a/cmd/dlv/dlv_test.go#L282-L286
Why was this disabled? It's not clear to me from reading the comments.

polinasok:
I checked both tests on linux. `proc_test` conveniently prints the expected error:
```
$ go test -run ^TestAttachStripped$ -v
=== RUN   TestAttachStripped
    proc_test.go:3173: error is could not open debug info
--- PASS: TestAttachStripped (0.79s)
```
but `dlv_test` if modified to print output and error, has the following:
```
    dlv_test.go:276: [output] time="2022-03-16T07:03:59Z" level=warning msg="CGO_CFLAGS already set, Cgo code could be optimized." layer=dlv
        Stdin is not a terminal, use '-r' to specify redirects for the target process or --allow-non-terminal-interactive=true if you really want to specify a redirect for Delve
    dlv_test.go:277: [error] exit status 1
```
That doesn't look right, does it?

aarzilli:
I think the reason it is skipped on darwin is https://github.com/go-delve/delve/pull/2018#issuecomment-623422482

> but `dlv_test` if modified to print output and error, has the following:
> 
> ```
>     dlv_test.go:276: [output] time="2022-03-16T07:03:59Z" level=warning msg="CGO_CFLAGS already set, Cgo code could be optimized." layer=dlv
>         Stdin is not a terminal, use '-r' to specify redirects for the target process or --allow-non-terminal-interactive=true if you really want to specify a redirect for Delve
>     dlv_test.go:277: [error] exit status 1
> ```
> 
> That doesn't look right, does it?

Modified how? Depending how delve is getting started it could be the right error message, some tests in dlv_test do need the `--allow-non-terminal-interactive` option to work.

aarzilli:
The process always gets launched, the test checks that the process gets closed after delve returns an error.

polinasok:
Summarizing conclusions for completeness:

>> dlv_test.go:276: [output] time="2022-03-16T07:03:59Z" level=warning msg="CGO_CFLAGS already set, Cgo code could be optimized." layer=dlv
        Stdin is not a terminal, use '-r' to specify redirects for the target process or --allow-non-terminal-interactive=true if you really want to specify a redirect for Delve
> > That doesn't look right, does it? 

> Modified how? Depending how delve is getting started it could be the right error message, some tests in dlv_test do need the `--allow-non-terminal-interactive` option to work.

When debugging this test, I did not change how the delve command was started, but only logged its combined output, which revealed that it failed with an error other than the one that the test was intended for and before any process launching was attempted. The delve command inside of such tests should be run either with `--headless` or `--allow-non-terminal-interactive` to get through command set-up to actually launching the debugger and the process. #2949 fixes the issue.

PR #2020, opened 2020-04-22
proc: keep debugger state consistent if callInjectionProtocol fails

aarzilli:
```
proc: keep debugger state consistent if callInjectionProtocol fails

pickCurrentThread should always run.

```

### DIFF
--- a/pkg/proc/target_exec.go
+++ b/pkg/proc/target_exec.go
@@ -78,13 +78,17 @@ func (dbp *Target) Continue() error {
 
 		threads := dbp.ThreadList()
 
-		callInjectionDone, err := callInjectionProtocol(dbp, threads)
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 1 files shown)


pkg/proc/target_exec.go
```diff
@@ -78,15 +78,19 @@ func (dbp *Target) Continue() error {
 
 		threads := dbp.ThreadList()
 
-		callInjectionDone, err := callInjectionProtocol(dbp, threads)
-		if err != nil {
-			return err
-		}
+		callInjectionDone, callErr := callInjectionProtocol(dbp, threads)
+		// callErr check delayed until after pickCurrentThread, which must always
+		// happen, otherwise the debugger will could be left in an inconsisten
```
derekparker, line 83: s/will could/could/
s/inconsisten/inconsistent/

Issue #2024, opened 2020-04-23
dlv attach can't find the external debug file properly


coderplay:
1. What version of Delve are you using (`dlv version`)?

```
$ dlv version
Delve Debugger
Version: 1.4.0
Build: $Id: 67422e6f7148fa1efa0eac1423ab5594b223d93b
```

2. What version of Go are you using? (`go version`)?

```
$ go version
go version go1.14.2 linux/amd64
```

3. What operating system and processor architecture are you using?

```
$ uname -a
Linux 4.15.0-1057 #59-Ubuntu SMP Wed Dec 4 10:02:00 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

4. What did you do?

I was attaching dlv to a daemon process of of kubernetes.  
executable path:  `/usr/bin/kubelet`
external debug file : `/usr/lib/debug/kubelet.debug`
debug-info-directories: `["/usr/lib/debug/.build-id", "/usr/lib/debug"]`

```bash
$ sudo dlv attach `pgrep kubelet`
```

5. What did you expect to see?

dlv should able to attach the target running process.

6. What did you see instead?

```
could not attach to pid 8020: could not open debug info
```

7. Root cause

The commit 
https://github.com/go-delve/delve/commit/f3b149bda706005283f547e0cf090457d084349c#diff-102e824b326c8dd9e8721394368821a8L625  switch the executable path to `image.Path` 

``` go
potentialDebugFilePath = fmt.Sprintf("%s/%s.debug", dir, filepath.Base(image.Path))
```

The above `image.Path` was originally from `proc_linux.go` passed by its `Attach` method
```
func Attach(pid int, debugInfoDirs []string) (*proc.Target, error) {
...
	tgt, err := dbp.initialize(findExecutable("", dbp.pid), debugInfoDirs)
...
}

func findExecutable(path string, pid int) string {
	if path == "" {
		path = fmt.Sprintf("/proc/%d/exe", pid)
	}
	return path
}
```

So for my case the `image.Path` was `/proc/8020/exe`.  Delve tried to open `/usr/lib/debug/exe.debug` instead of `/usr/lib/debug/kubelet.debug`.

aarzilli:
Fixed by #2025

PR #2025, opened 2020-04-23
proc: Find executable should follow symbol links.

coderplay:
https://github.com/go-delve/delve/issues/2024

On linux platform, we simply treated `/proc/$pid/exe` as the
executable of targeting process when doing `dlv attach`. The
`/proc/$pid/exe` is a symbol link of the real executable file.
Delve couldn't find the corrsponding external debug file based on the
symbol link:

```
could not attach to pid $pid: could not open debug info
```

The fix is to evaluate the symbol links to the actual executable path.

### DIFF
--- a/pkg/proc/native/proc_linux.go
+++ b/pkg/proc/native/proc_linux.go
@@ -119,7 +119,12 @@ func Attach(pid int, debugInfoDirs []string) (*proc.Target, error) {
 		return nil, err
 	}
 
-	tgt, err := dbp.initialize(findExecutable("", dbp.pid), debugInfoDirs)
+	execPath, err := findExecutable(pid)
+	if err != nil {
+		return nil, err
+	}
+
+	tgt, err := dbp.initialize(execPath, debugInfoDirs)
 	if err != nil {
 		dbp.Detach(false)
 		return nil, err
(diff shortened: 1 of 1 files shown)


coderplay:
> I think this would be more appropriate in `openSeparateDebugInfo` in pkg/proc/bininfo.go, change:
> 
> ```
> potentialDebugFilePath = fmt.Sprintf("%s/%s.debug", dir, filepath.Base(image.Path))
> ```
> 
> into:
> 
> ```
> potentialDebugFilePath = fmt.Sprintf("%s/%s.debug", dir, filepath.Base(filepath.EvalSymlinks(image.Path)))
> ```

Thanks for taking a look, Alessandro!   I agree that patching on `openSeparateDebugInfo` could be better in terms of side effects.   That actually what I thought  at the very beginning.   I turned to modify `findExecutable` in `proc_linux.go` because of below reasons:

1.  I looked at the `findExecutable` implementations for other platforms. They are returning actual executable paths.  I thought the return result for linux might cause other potential problems giving that it's not consistent with the other platform's.

https://github.com/go-delve/delve/blob/master/pkg/proc/native/proc_windows.go#L110

https://github.com/go-delve/delve/blob/master/pkg/proc/native/proc_darwin.c#L93

https://github.com/go-delve/delve/blob/master/pkg/proc/native/proc_freebsd.c#L31

2.  From the literal meaning of `findExecutable` and the further usage of it in `NewTargetConfig.Path` or `Image.Path`,  the actual executable path makes more sense than `/proc/$pid/exe`

3.  From the documentation of [dlv attach](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_attach.md). User can optionally pass an executable path to the command, which I think in most of cases is the actual path rather than `/proc/$pid/exe`  

```
dlv attach pid [executable]
```

Please correct me if I am wrong, and please leave your comments after seeing above. Thanks!

PR #2031, opened 2020-04-26
cmd/dlv: make core command limitations more prominent

### LINKED ISSUE #2026: document the limitation of Mac OS coredump analysis
Currently `dlv` cannot process core dumps from Mac for various reasons.
Document the limitations. 

@heschik

aarzilli:
```
cmd/dlv: make core command limitations more prominent

Fixes #2026

```


aarzilli:
Our test for usage being up to date was broken, we didn't check whether the script actually ran and it didn't because the vendor directory didn't contain the packages to run it.

cmd/dlv/tools.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+// List packages used by _scirpts
```
derekparker, line 3: s/_scirpts/scripts/

And is this because it's ignored now that it's been moved to `_scripts` and the underscore is blocking the Go tool from finding the deps?

aarzilli: I don't know. Maybe it's just been broken for a while and we didn't notice.

PR #2034, opened 2020-05-01
Add some simple examples for the list command

colinnewell:
These are just some simple examples of how the list command can be used.  I'm not sure how you'd prefer to document these things, so I figured it would be worth throwing something together to at least start a discussion about what you'd prefer.

I was looking at the documentation struggling to remember how to use the list command and some concrete examples would be useful for me.  I essentially grabbed these from your tests.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -348,6 +348,13 @@ Show source code.
 
 Show source around current point or provided linespec.
 
+For example:
+
+	frame 1 list 69
+	list testvariables.go:10000
+	list main.main:30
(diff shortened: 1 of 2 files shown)


derekparker:
I think it's a good idea to add some more examples. However this new documentation should be added in `pkg/terminal/command.go` in the description for the `list` command. Documentation will then need to be regenerated via `go run _scripts/gen-cli-docs.go`.

colinnewell:
Thanks for the pointer.  I have updated the commit accordingly.

PR #2039, opened 2020-05-05
Highlight printed go code syntax on CLI

### LINKED ISSUE #1273: Syntax highlighting

```
Delve Debugger
Version: 1.0.0
```
`go version go1.10.2 linux/amd64`
Ubuntu 18.04 LTS x86_64
4. What did you do?
`l`
5. What did you ~expect~ want to see?
Syntax highlighted code
6. What did you see instead?
Non syntax highlighted code

I now this is difficult but it would be quite useful.

bulentkazanci:
To increase the readability of printed Go code on CLI, a syntax highlighter was added. The example can be seen below.

Closes #1273

**Showcase:**
delve-syntax-highlight-ex


aarzilli:
Delve uses vendored dependencies, if you add dependencies you need to run `go mod vendor`.

This PR introduces some dubious dependencies that I'd much rather not take:

* github.com/alecthomas/kong: we already have a command line parser, we don't need another one and I don't see why one should be needed to print syntax highlighting at all
* github.com/dlclark/regexp2: we already have a regexp library, it's in go's standard library. Also no regexp library at all is needed to parse go.
* github.com/mitchellh/mapstructure: not sure what this has to do with syntax highlighting at all
* github.com/danwakefield/fnmatch: this doesn't even document what it does
* github.com/sergi/go-diff: what do diffs have to do with syntax highlighting?
* github.com/stretchr/testify: we already have a testing library in the standard library.
* why does this want to update golang.org/x/sys? why would a syntax highlighting library need to make system calls?

bulentkazanci:
I'll change the new highlighting package I used with a lightweight one to get rid of those dependencies and update here.

bulentkazanci:
@aarzilli Hi I changed the package for highlighting. Could you review it again?

pkg/config/config.go
```diff
@@ -53,9 +53,9 @@ type Config struct {
 	// expression for its argument.
 	ShowLocationExpr bool `yaml:"show-location-expr"`
 
-	// Source list line-number color (3/4 bit color codes as defined
-	// here: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors)
-	SourceListLineColor int `yaml:"source-list-line-color"`
+	// Style to be used to highlight source code as defined
+	// here: https://godoc.org/github.com/alecthomas/chroma#Style
```
aarzilli, line 57: I presume the url needs to be changed.

pkg/config/config.go
```diff
@@ -53,9 +53,9 @@ type Config struct {
...
+	SourceCodeStyle string `yaml:"source-code-template"`
```
aarzilli, line 58: Having the ability to specify the base color for the source code was a feature added by someone. I don't use it but I think they probably liked it and I would rather not remove it. I'm not sure how the themes work here, but it seems moot given that there is only one.

pkg/terminal/terminal.go
```diff
@@ -268,12 +247,13 @@ func (t *Term) Run() (int, error) {
...
-		terminalColorEscapeCode := fmt.Sprintf(terminalHighlightEscapeCode, t.conf.SourceListLineColor)
-		prefix = fmt.Sprintf("%s%s%s", terminalColorEscapeCode, prefix, terminalResetEscapeCode)
-	}
-	fmt.Fprintf(t.stdout, "%s%s\n", prefix, str)
+func (t *Term) Println(str string) {
+	fmt.Fprintf(t.stdout, "%s\n", str)
+}
+
+func (t *Term) PrintHighlightedCode(code string) error {
+	err := beni.Highlight(strings.NewReader(code), t.stdout, "Go", t.conf.SourceCodeStyle, "Terminal256")
```
aarzilli, line 255: I don't think you can call this with a chunk of a file and expect it to work, the first line that we print could be inside a comment, for example.

pkg/config/config.go
bulentkazanci: So far the So far [packaga]() has only one theme but I think in time more can be added. If you don't like this restriction, do you have any suggestions to replace it?

pkg/terminal/terminal.go
bulentkazanci: I tested it with a chunk starting with a commented line before. What could be the possible reason for failure?

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/google/go-dap v0.2.0
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/koron/beni v0.0.0-20150126044346-f85aa8e52ecf
```
komuw, line 12: `github.com/koron/beni` does not have a license on their repo.

pkg/config/config.go
derekparker: There hasn't been any updated here in a while. @aarzilli do you have a response to this?

derekparker:
Sorry for such a delay on this. I was waiting for the other comments to be resolved but I think we need to start moving forward on this if we want to get this feature merged. I agree having source code coloring would be a good feature, so I'm
 happy to help push this forward.

pkg/config/config.go
aarzilli: I still think we shouldn't remove the option for specifying the default color.

aarzilli:
I think there are two main problems with this PR as it stands: (1) the dependency it's introducing doesn't have a licence file, people are going to take issue with that, (2) a tokenizer or parser based highlighter isn't going to work properly on an arbitrary chunk of a file, so when it's used it will need to see the entire file and then the part we display will need to be cut from the highlighted output.

go.mod
komuw: created https://github.com/koron/beni/issues/8

derekparker:
Closing as we will likely go with https://github.com/go-delve/delve/pull/2294.

PR #2050, opened 2020-05-12
pkg/proc: Fix crash when LocationExpr is nil

### LINKED ISSUE #2049: Internal debugger error: runtime error: invalid memory address or nil pointer dereference.
1.4.0

1.14

Docker

4. What did you do?
Ran `ENTRYPOINT ["dlv", "exec", "/app/frontend", "--continue", "--accept-multiclient", "--api-version=2", "--headless", "--listen=:3000", "--log"]` and then reloaded the application to hit the breakpoint in VSCode.

I expected the debugger to stop on the breakpoint.

6. What did you see instead?
The instance shut down and got the following error message:
```
Unable to produce stack trace: "Internal debugger error: runtime error: invalid memory address or nil pointer dereference
runtime.gopanic (0x434455)
	/usr/local/go/src/runtime/panic.go:969
runtime.panicmem (0x44abe9)
	/usr/local/go/src/runtime/panic.go:212
runtime.sigpanic (0x44aa38)
	/usr/local/go/src/runtime/signal_unix.go:695
github.com/go-delve/delve/pkg/proc.(*EvalScope).Locals (0x6eecc9)
	/go/src/github.com/go-delve/delve/pkg/proc/eval.go:269
github.com/go-delve/delve/pkg/proc.(*EvalScope).LocalVariables (0x6f00ce)
	/go/src/github.com/go-delve/delve/pkg/proc/eval.go:411
github.com/go-delve/delve/service/debugger.(*Debugger).convertStacktrace (0x78eecb)
	/go/src/github.com/go-delve/delve/service/debugger/debugger.go:1389
github.com/go-delve/delve/service/debugger.(*Debugger).Stacktrace (0x78e165)
	/go/src/github.com/go-delve/delve/service/debugger/debugger.go:1325
github.com/go-delve/delve/service/rpc2.(*RPCServer).Stacktrace (0x9b4346)
	/go/src/github.com/go-delve/delve/service/rpc2/server.go:187
reflect.Value.call (0x4c30da)
	/usr/local/go/src/reflect/value.go:460
reflect.Value.Call (0x4c25e3)
	/usr/local/go/src/reflect/value.go:321
github.com/go-delve/delve/service/rpccommon.(*ServerImpl).serveJSONCodec.func2 (0xa09186)
	/go/src/github.com/go-delve/delve/service/rpccommon/server.go:311
github.com/go-delve/delve/service/rpccommon.(*ServerImpl).serveJSONCodec (0xa07539)
	/go/src/github.com/go-delve/delve/service/rpccommon/server.go:313
runtime.goexit (0x464a90)
	/usr/local/go/src/runtime/asm_amd64.s:1373"
```

chainhelen:
Add nil check for v.LocationExpr

Fix #2049

-------------------   

Is it necessary to add a testcase?But how to construct such a testcase?   

----------

Has rebase.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -261,13 +261,16 @@ func (scope *EvalScope) Locals() ([]*Variable, error) {
 			locationExpr := v.LocationExpr
 			declLine := v.DeclLine
 			v = v.maybeDereference()
-			if v.Addr == 0 {
+			if v.Addr == 0 && v.Unreadable == nil {
 				v.Unreadable = fmt.Errorf("no address for escaped variable")
 			}
 			v.Name = name[1:]
 			v.Flags |= VariableEscaped
-			locationExpr.isEscaped = true
-			v.LocationExpr = locationExpr
+			// See https://github.com/go-delve/delve/issues/2049 for details
+			if locationExpr != nil {
+				locationExpr.isEscaped = true
+				v.LocationExpr = locationExpr
+			}
 			v.DeclLine = declLine
 			vars[i] = v
 		}


derekparker:
I agree we should be checking if `v.LocationExpr` is nil, but I'm curious why the variable doesn't have a `LocationExpr`. It seems that only happens when `v.Unreadable != nil`, so I wonder if this panic is hiding an error where we couldn't parse a var.

chainhelen:
Agree, wait for more information about a small reproducer.

aarzilli:
I'd say that we've let this one slip. A reproducer would be cool, if nothing else as a regression test. However one failed to appear. My proposal is to apply this with two changes: (1) the one I've already requested about saving LocationExpr before maybeDereference and (2) change `if v.Addr == 0 { v.Unreadable = ... }` into `if v.Addr == 0 && v.Unreadable == nil { v.Unreadable = ... }`. 

This way the nil pointer dereference will stop masking the real problem and if somebody re-encounters it the unreadable error at its root will be reported to us.

chainhelen:
Not any process on reproducer and avoid to cover up the real problem. Rebase and done.

derekparker:
> My proposal is to apply this with two changes: (1) the one I've already requested about saving LocationExpr before maybeDereference and (2) change `if v.Addr == 0 { v.Unreadable = ... }` into `if v.Addr == 0 && v.Unreadable == nil { v.Unreadable = ... }`.

Where was the first suggestion? I don't see it in this thread or in the issue. Does the current patch include the changes you wanted to see? If so I agree that I think we should merge this fix without a reproducer.

> This way the nil pointer dereference will stop masking the real problem and if somebody re-encounters it the unreadable error at its root will be reported to us.

Agreed, this sounds good.

aarzilli:
> Does the current patch include the changes you wanted to see?

Yes. But now it doesn't have the check for `locationExpr != nil` before setting `locationExpr.isEscaped = true`, which was the original point of the PR. I think @chainhelen dropped one commit here.

chainhelen:
Done and rebase.

PR #2054, opened 2020-05-15
gdbserial: ensure registers are loaded in SetDX/SetSP/SetPC

aarzilli:
```
gdbserial: ensure registers are loaded in SetDX/SetSP/SetPC

These methods only work if registers have been loaded once after the
last resume, there's probably no code path that calls SetXX before
Thread.Registers but lets make sure it can't happen anyway.

```

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -1840,6 +1840,7 @@ func (r *gdbRegisters) FloatLoadError() error {
 
 // SetPC will set the value of the PC register to the given value.
 func (t *gdbThread) SetPC(pc uint64) error {
+	_, _ = t.Registers() // Registes must be loaded first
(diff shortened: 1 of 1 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1871,6 +1871,7 @@ func (r *gdbRegisters) FloatLoadError() error {
 
 // SetPC will set the value of the PC register to the given value.
 func (t *gdbThread) SetPC(pc uint64) error {
+	_, _ = t.Registers()
```
derekparker, line 1874: nit: we should add a comment here explaining this as it may be non-obvious for a reader that we're calling it for the side effect of ensuring `t.regs` is set properly.

aarzilli: Done.

PR #2056, opened 2020-05-17
service/dap: Add buildFlags error checking and tests in launch requests

polinasok:
Updates #1515

### DIFF
--- a/pkg/proc/test/support.go
+++ b/pkg/proc/test/support.go
@@ -81,6 +81,12 @@ const (
 
 // BuildFixture will compile the fixture 'name' using the provided build flags.
 func BuildFixture(name string, flags BuildFlags) Fixture {
+	return BuildFixtureWithFlags(name, flags, "")
+}
+
+// BuildFixtureWithFlags will compile the fixture 'name' using the provided predefined
+// and custom build flags.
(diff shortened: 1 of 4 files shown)


polinasok:
Do you want me to fix the anti-pattern that DeepSource detected in the original code?

pkg/proc/test/support.go
```diff
@@ -81,6 +81,13 @@ const (
 
 // BuildFixture will compile the fixture 'name' using the provided build flags.
 func BuildFixture(name string, flags BuildFlags) Fixture {
+	return BuildFixtureWithFlags(name, flags, "")
+}
+
+// BuildFixtureWithFlags will compile the fixture 'name' using the provided
+// a la carte build flags as well as additional freeform build flags specified as
+// a string to be passed as-is to the 'go' command.
+func BuildFixtureWithFlags(name string, flags BuildFlags, moreflags string) Fixture {
```
aarzilli, line 90: Do you plan to use this in the future?

polinasok: I am reverting these changes. I tried to do something with build fixtures for testing, but then came up with a better way, so this is no longer needed.

PR #2059, opened 2020-05-21
*: v1.4.1

derekparker:
Bump version to v1.4.1.

There's a good amount of fixes and optimizations in this release and a some new features as well.

I'm marking the date as tomorrow because that's likely when this will get merged. Also removing some out of date text from the top of the CHANGELOG.


CHANGELOG.md
```diff
@@ -3,7 +3,37 @@
 All notable changes to this project will be documented in this file.
 This project adheres to Semantic Versioning.
 
-All changes mention the author, unless contributed by me (@derekparker).
+## [1.4.1] 2020-05-22
+
+### Added
```
chainhelen, line 8: Can we add `Support linux/386` (#1884)? As I understand, this is the first time we support 32-bit os.

derekparker:
@chainhelen right! Added it in.

PR #2061, opened 2020-05-25
proc: step into unexported runtime funcs when already inside runtime

aarzilli:
```
proc: step into unexported runtime funcs when already inside runtime

Normally we don't step into unexported runtime functions because the
compiler is free to insert them into the code and they are not relevant
to the user, however if we are already stepping through a runtime
function we should let step into work normally and step into other
runtime functions.

```

### DIFF
--- a/pkg/proc/target_exec.go
+++ b/pkg/proc/target_exec.go
@@ -7,6 +7,7 @@ import (
 	"go/ast"
 	"go/token"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


aarzilli:
I would lean against writing a test for this because it would depend on implementation details of the runtime.

derekparker:
Will merge after rebase.

aarzilli:
Rebased.

PR #2062, opened 2020-05-25
proc,proc/native: avoid context switches by executing Continue in ptrace thread

### LINKED ISSUE #1549: Conditional breakpoints are slow

Delve Debugger 
Version: 1.2.0 
Build: ac3b1c7a786d681a5aefcdded9888090d69b3832

go version go1.12.2 windows/amd64

Windows 10 64bit
AMD Ryzen 1800X

4. What did you do?

Put a conditional breakpoint inside an inner loop running millions of times. The condition is on the integer loop variable. 

I expected the loop run up to a few times slower due to the conditional breakpoint.

6. What did you see instead?

The loop runs about 1000x slower, which makes the conditional breakpoint useless if the loop count is higher than a few thousand.

A speedy workaround is to put the same condition as an if statement and an unconditional breakpoint inside it, but it is inconvenient, since needs recompilation and re-running the binary. 

I am using delve via GoLand as the GUI. JetBrains tested my code (see below) via delve (without GoLand) and it had the same performance issue. So this is not due to the GUI integration.

Binary inside GoLand: `%AppData%\Local\JetBrains\Toolbox\apps\Goland\ch-0\191.6183.86\plugins\go\lib\dlv\windows>dlv`

How to reproduce:

```go
package main

import ( 
    "fmt" 
    "time" 
)

func main() { 
    sum := int64(0) 
    start := time.Now() 
    for value := int64(0); value < 10000000; value++ { 
        sum += value 
    } 
    elapsed := time.Since(start) 
    fmt.Printf("Sum: %d\nTook %s", sum, elapsed) 
}
```

Run without debugging:

Sum: 49999995000000 
Took 6.0053ms

Run with debugger, without conditional breakpoint:

Sum: 49999995000000 
Took 24.0209ms

Put breakpoint with condition `value == -1` on code line `sum += value`.

Debug run with conditional breakpoint:

Paused it after a few minutes and checked the current value. It got only 36k iterations far.

So I slashed the loop count by 1000 from 10M to 10000 and run it again:

Sum: 49995000 
Took 32.4708425s

Based on the above result 10M iterations would have taken 32471 seconds (9 hours). More than 1000 times slowdown due to the presence of the conditional breakpoint.

This is why conditional breakpoints cannot be used conveniently inside inner loops right now, because they are plainfully slow.

Workaround is to add the condition to the code as an if statement and put a breakpoint inside it:

```go
if value == -1 { 
    fmt.Print("Put breakpoint here") 
} 
```

It is barely slower than running the original code in debug mode: 

Sum: 49999995000000 
Took 28.0243ms

There must be some crazy high overhead of calling out to check those breakpoint conditions. I guess it calls out to Java, extract the variable values from the stack / heap, then somehow interprets the condition there.

Would it be possible to apply the above workaround automatically at least if such a conditional breakpoint is set before starting the debugged application? It would help a lot with catching corner cases in inner loops and analyzing them in their context.

Maybe there is a way to add a call between each pair of instructions, allowing the debugger to inject conditional checks there while the code is already running. Or some "NOP" instructions which can be replaced later at runtime with a debug trap. There must be a solution, in case of C/C++ it had good performance.

I understand if this is not possible, then I just stick with my workaround.

aarzilli:
```
proc,proc/native: avoid context switches by executing Continue in ptrace thread

Avoids a lot of context switches while executing Continue by running
inside the ptrace thread.

Benchmark before:

BenchmarkConditionalBreakpoints-4   	       1	3554266510 ns/op

After:

BenchmarkConditionalBreakpoints-4   	       1	1807164143 ns/op

Fixes #1549

```

### DIFF
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -38,6 +38,8 @@ type nativeProcess struct {
 	// this process.
 	ctty *os.File
 
+	insidePtraceThread bool // currently executing inside the ptrace thread
+
 	exited, detached bool
 }
 
@@ -296,20 +298,48 @@ func (dbp *nativeProcess) handlePtraceFuncs() {
 	runtime.LockOSThread()
 
 	for fn := range dbp.ptraceChan {
+		dbp.insidePtraceThread = true
 		fn()
+		dbp.insidePtraceThread = false
 		dbp.ptraceDoneChan <- nil
 	}
+	close(dbp.ptraceDoneChan)
 }
 
 func (dbp *nativeProcess) execPtraceFunc(fn func()) {
+	if dbp.insidePtraceThread {
+		fn()
+		return
+	}
 	dbp.ptraceChan <- fn
 	<-dbp.ptraceDoneChan
+}
+
+// ExecOnMagicThread executes fn on the thread that can make ptrace
+// calls (or the equivalent of ptrace on windows). If fn does a lot of calls
+// to other methods of Process that need to do ptrace calls this will speed
+// it up considerably by avoiding a large number of context switches.
+// It is very important that fn does not panic, you must defer a recover
+// call here.
+//
+// This function works by calling fn in the ptrace goroutine and then
+// disabling the code that sends ptrace calls to that goroutine, executing
+// them directly instead.
+// It is very important that only the goroutine that calls this function
+// will attempt to execute ptrace calls, otherwise the ptrace calls will be
+// executed by a goroutine that can't execute them and return the ESRCH / no
+// such process (or access denied on windows).
+// For Delve this property is ensured by the debugger package which takes
+// care to serialize every call to pkg/proc.
+// The only risky thing is function call injection which happens inside a
+// different goroutine.
+func (dbp *nativeProcess) ExecOnMagicThread(fn func()) {
+	dbp.execPtraceFunc(fn)
 }
 
 func (dbp *nativeProcess) postExit() {
 	dbp.exited = true
 	close(dbp.ptraceChan)
-	close(dbp.ptraceDoneChan)
 	dbp.bi.Close()
 	if dbp.ctty != nil {
 		dbp.ctty.Close()
--- a/pkg/proc/core/core.go
+++ b/pkg/proc/core/core.go
@@ -447,3 +447,8 @@ func (p *process) FindThread(threadID int) (proc.Thread, bool) {
 func (p *process) SetCurrentThread(th proc.Thread) {
 	p.currentThread = th.(*thread)
 }
+
+// ExecOnMagicThread calls fn.
+func (p *process) ExecOnMagicThread(fn func()) {
+	fn()
+}
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -635,6 +635,11 @@ func (p *gdbProcess) SetCurrentThread(th proc.Thread) {
 	p.currentThread = th.(*gdbThread)
 }
 
+// ExecOnMagicThread calls fn.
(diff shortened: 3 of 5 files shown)


aarzilli:
The performance improvement is significant (over twice as fast) but it's also awful.

pkg/proc/native/proc.go
```diff
@@ -333,20 +335,48 @@ func (dbp *nativeProcess) handlePtraceFuncs() {
...
+// calls (or the equivalent of ptrace on windows). If fn does a lot of calls
+// to other methods of Process that need to do ptrace calls this will speed
+// it up considerably by avoiding a large number of context switches.
+// It is very important that fn does not panic, you must defer a recover
+// call here.
+//
+// This function works by calling fn in the ptrace goroutine and then
+// disabling the code that sends ptrace calls to that goroutine, executing
+// them directly instead.
+// It is very important that only the goroutine that calls this function
```
derekparker, line 365: Since we no longer have the guarantee that the calls will be serialized we need to enforce it and ensure we produce correct behavior in some other way than relying on the channel to serialize.

I'd be curious to see the performance implications of ensuring we're executing on the correct thread. E.g. using `syscall.Gettid()` and checking against the thread we know to be the "ptrace" or "magic" thread.

pkg/proc/native/proc.go
```diff
@@ -333,20 +335,48 @@ func (dbp *nativeProcess) handlePtraceFuncs() {
...
+		dbp.insidePtraceThread = true
 		fn()
+		dbp.insidePtraceThread = false
 		dbp.ptraceDoneChan <- nil
 	}
+	close(dbp.ptraceDoneChan)
 }
 
 func (dbp *nativeProcess) execPtraceFunc(fn func()) {
+	if dbp.insidePtraceThread {
```
derekparker, line 347: Maybe instead of checking this state we store the thread id of the "magic" thread and then compare here. Something like `if dbp.magicThread == syscall.Gettid()`, I wonder if we lose out on the performance improvements by injecting a syscall in this path.

Otherwise I fear we're opening ourselves to bugs where this global state causes other goroutines to attempt to run ptrace operations due to programmer error.

aarzilli: I just tried it and it's actually fast. However I don't think this will help.

The problem only arises when an Interloper-goroutine runs in parallel with the Continue-goroutine and tries to do a ptrace call. It stands to reason that the Continue-goroutine will always have to synchronize with the Interloper-goroutine, if it didn't the Interloper-goroutine could do ptrace calls *after* the Continue-goroutine has resumed the target process, which would fail (here I'm assuming that the Interloper-goroutine will limit itself to reading/writing memory, i.e. that we will never have two goroutines trying to simultaneously manage the run state of the target process, I think this is reasonable).

If we stored and checked the thread id of the magic thread (which would be the thread executing the Continue-goroutine) then the Interloper-goroutine would try to send on ptraceChan, but ptraceChan wouldn't be read until the Continue-goroutine terminates, and since the Continue-goroutine needs to wait for the Interloper-goroutine before resuming the thread we would always get a deadlock.

We could check if the thread ID is the same as the ExecOnMagicThread thread ID and return an error if it isn't, I think a runtime error is marginally preferable to a deadlock, but the OS is already doing this for us (it's ESRCH on linux).

And there's also the fact that Gettid doesn't exist anywhere else but linux. In fact from a quick look to the syscall package I can't figure out how to do this on freebsd and windows.

aarzilli:
You know, the more I think about this the less I feel good about it. Maybe we should just keep the performance penalty.

derekparker:
Closing for now. I think this is a great thing to do, but might involve some other refactorings or changes to be done with a bit more confidence.

PR #2064, opened 2020-05-26
proc/*: move Set/Clear Breakpoint methods to Target

aarzilli:
```
proc/*: move Set/Clear Breakpoint methods to Target

```


pkg/proc/breakpoints.go
```diff
@@ -304,18 +296,22 @@ func (bpmap *BreakpointMap) Set(addr uint64, kind BreakpointKind, cond ast.Expr,
 }
 
 // SetWithID creates a breakpoint at addr, with the specified logical ID.
-func (bpmap *BreakpointMap) SetWithID(id int, addr uint64, writeBreakpoint WriteBreakpointFn) (*Breakpoint, error) {
-	bp, err := bpmap.Set(addr, UserBreakpoint, nil, writeBreakpoint)
+func (t *Target) setWithID(id int, addr uint64) (*Breakpoint, error) {
```
derekparker, line 299: This method looses some context when attached to `Target` instead of `BreakpointMap`. This might be better as just a function.

aarzilli: How about setBreakpointWithID?

PR #2065, opened 2020-05-28
cmd/delve: Add command line flag to retain output

jquirke:
We add an option to keep the temporary binary file build by Delve.

Currently, the delve cli removes the temporary files created by the
build system when it exits.

Unfortunately, this means go build must re-link the output on each
session, even when the source files have not changed, and can result
in significant time waiting.

e.g.

```
$ echo -n 'q' | time dlv debug
        22.48 real        20.97 user         6.82 sys
$ echo -n 'q' | time dlv debug --keep-output
        21.70 real        19.45 user         6.60 sys
$ echo -n 'q' | time dlv debug
        2.74 real         3.64 user         2.90 sys
$ echo '// force build' >> main.go
$ echo -n 'q' | time dlv debug --keep-output
        22.29 real        21.01 user         6.36 sys
```

### DIFF
--- a/Documentation/usage/dlv.md
+++ b/Documentation/usage/dlv.md
@@ -26,6 +26,7 @@ Pass flags to the program you are debugging using `--`, for example:
       --check-go-version     Checks that the version of Go in use is compatible with Delve. (default true)
       --headless             Run debug server only, in headless mode.
       --init string          Init file, executed by the terminal client.
+      --keep-output          Keep output binary.
   -l, --listen string        Debugging server listen address. (default "127.0.0.1:0")
       --log                  Enable debugging server logging.
       --log-dest string      Writes logs to the specified file or file descriptor (see 'dlv help log').
--- a/Documentation/usage/dlv_attach.md
+++ b/Documentation/usage/dlv_attach.md
@@ -32,6 +32,7 @@ dlv attach pid [executable]
       --check-go-version     Checks that the version of Go in use is compatible with Delve. (default true)
       --headless             Run debug server only, in headless mode.
       --init string          Init file, executed by the terminal client.
+      --keep-output          Keep output binary.
   -l, --listen string        Debugging server listen address. (default "127.0.0.1:0")
       --log                  Enable debugging server logging.
       --log-dest string      Writes logs to the specified file or file descriptor (see 'dlv help log').
--- a/Documentation/usage/dlv_backend.md
+++ b/Documentation/usage/dlv_backend.md
@@ -25,6 +25,7 @@ are:
       --check-go-version     Checks that the version of Go in use is compatible with Delve. (default true)
       --headless             Run debug server only, in headless mode.
       --init string          Init file, executed by the terminal client.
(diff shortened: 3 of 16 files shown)


derekparker:
I prefer not to add this because we already have a mechanism for doing this. What you really want is `dlv exec` with a prebuilt binary, I think.

jquirke:
Fair point, but on large repos the link phase can dominate, even on 'test' command.

e.g. here is a moderate size repo at Uber. If I jump out of the debugger for any reason, say to change a file, etc, I have to wait nearly a minute for the link.

```
qjeremy$ echo -n q | time dlv test  ./services/XXX-rides
       61.55 real        48.72 user        17.14 sys
qjeremy$ echo -n q | time dlv test  ./services/XXX-rides
       48.88 real        45.39 user        16.44 sys
```

The work around is then to do:

```
go test -c -o test.bin && ./services/XXX-rides dlv exec test.bin
```

which works but is meh. None of these solutions are great (including my pull request).

dgladkov:
Considering that building application or tests manually and running `dlv exec` significantly complicates CLI usage, is there a specific reason why we prefer to remove the binaries in the first place?

derekparker:
@dgladkov @jquirke sorry for the late reply on this, it wasn't my intention to close this and then ignore it. This isn't a hard no, I'm listening to the rationale, I just don't see a compelling reason to add this as of right now, I want to understand where the pain points are.

Is the problem with the existing workflow in building a debug binary that you can keep around in between invocations of the `dlv exec` command? I understand if it's easier to have Delve build a debug binary because it will always know the correct flags to build the best binary, whereas a user may not, so I can definitely understand that argument.

Aside from the `dlv exec` option, what about just leaving the session open and issuing a `restart` command? The goal isn't to rebuild the application, just essentially continuing to debug a binary that Delve has built, correct? In that case, is it possible to leave the debug session open and just `restart` when you want to start the debug session over again?

jquirke:
> @dgladkov @jquirke sorry for the late reply on this, it wasn't my intention to close this and then ignore it. This isn't a hard no, I'm listening to the rationale, I just don't see a compelling reason to add this as of right now, I want to understand where the pain points are.
> 
> Is the problem with the existing workflow in building a debug binary that you can keep around in between invocations of the `dlv exec` command? I understand if it's easier to have Delve build a debug binary because it will always know the correct flags to build the best binary, whereas a user may not, so I can definitely understand that argument.
> 
> Aside from the `dlv exec` option, what about just leaving the session open and issuing a `restart` command? The goal isn't to rebuild the application, just essentially continuing to debug a binary that Delve has built, correct? In that case, is it possible to leave the debug session open and just `restart` when you want to start the debug session over again?

Yes all of this is possible, and this is a minor nuisance, due to the various workarounds, but our "magical" idea of dlv test ideally should be the same as go test; itshould be **fast**. 

Currently it is not for projects of any size, due in part to the way the go build system works, and in part due to the way we remove binaries that would not require a rebuild.

PR #2068, opened 2020-06-01
Documentation: clarify API version status

### LINKED ISSUE #2066: Confusing --api-version default of 1 and related documentation
RPC APIv1 is deprecated. Documentation [says](https://github.com/go-delve/delve/blob/master/service/rpc1/readme.txtr) RPC APIv1 is only kept around for backward compatibility and [notes](https://github.com/go-delve/delve/blob/master/Documentation/api/json-rpc/README.md#api-versions) new clients must use v2.

Yet the [--api-version flag](https://github.com/go-delve/delve/blob/f96663a243ac3d844a3e43731d7eaf5b1bf3dfab/cmd/dlv/cmds/commands.go#L118) that sets the API for a headless server defaults to 1. This became a point of confusion in a recent meeting with my workgroup, when developers tried to use the flag default as a way to look up which API to focus the client code on.

It appears that instead of relying on this flag on server start-up (which is not even always possible as the debugger might be started independently), the clients are supposed to force the API to be v2 using [RPCServer.SetApiVersion](https://github.com/go-delve/delve/blob/c06a1a0252ee164a4a775c587396945ab771c498/service/rpccommon/server.go#L387). Yet I do not see this captured in the [documentation](https://github.com/go-delve/delve/blob/master/Documentation/api/json-rpc/README.md#api-versions). 

Using SetApiVersion is indeed what dlv frontend does itself as part of [rpc2 client set up](https://github.com/go-delve/delve/blob/8bb93e9ae19589b7ee665225c2206ef118928ae9/service/rpc2/client.go#L36). And this results in the following a bit unexpected log:
```
$ dlv exec --log --log-output=rpc helloworld
2020-05-27T10:03:11-07:00 info layer=rpc Using API v1
2020-05-27T10:03:11-07:00 debug layer=rpc <- RPCServer.SetApiVersion(
api.SetAPIVersionIn{"APIVersion":2})
2020-05-27T10:03:11-07:00 debug layer=rpc -> *api.SetAPIVersionOut{} error:""
2020-05-27T10:03:11-07:00 debug layer=rpc <-RPCServer.IsMulticlient(rpc2.IsMulticlientIn{})
... more rpc2 calls ...
```

I therefore propose the following improvements (and happy to make all or some these changes myself):
1.  Update --api-version flag default to 2.
2. Update --api-version flag description to stress that rpc2 is the encouraged version for all new clients while rpc1 is only around for backward compatibility. 
3. Update --api-version flag description to mention SetAPIVersion rpc for cases when changing the flag value is outside of user control.
4. Update  [API documentation](https://github.com/go-delve/delve/blob/master/Documentation/api/json-rpc/README.md#api-versions) to encourage using SetAPIVersion to switch to rpc2.

aarzilli:
```
Documentation: clarify API version status

Clarify that the currently supported version of the API is 2 and that
the preferred method for selecting it is sending a SetApiVersion
command after connection.

Fixes #2066

```

### DIFF
--- a/Documentation/api/ClientHowto.md
+++ b/Documentation/api/ClientHowto.md
@@ -65,6 +65,17 @@ Delve will respond by sending a response packet that will look like this:
 {"id":27, "result": {"Breakpoint": {"id":3, "name":"", "addr":4538829, "file":"/User/you/some/file.go", "line":16, "functionName":"main.main", "Cond":"", "continue":false, "goroutine":false, "stacktrace":0, "LoadArgs":null, "LoadLocals":null, "hitCount":{}, "totalHitCount":0}}, "error":null}
 ```
 
+## Selecting the API version
+
+Delve currently supports two version of its API, APIv1 and APIv2. By default
+a headless instance of `dlv` will serve APIv1 for backward-compatibility
+with older clients, however new clients should use APIv2 as new features
+will only be made available through version 2. The preferred method of
+switching to APIv2 is to send the `RPCServer.SetApiVersion` command right
+after connecting to the backend.
+Alternatively the `--api-version=2` command line option can be used when
+spawning the backend.
+
 ## Diagnostics
 
 Just like any other program, both Delve and your client have bugs. To help
--- a/Documentation/api/json-rpc/README.md
+++ b/Documentation/api/json-rpc/README.md
@@ -8,7 +8,8 @@ Here is an (incomplete) [list of language implementations](http://json-rpc.org/w
 
 # API versions
 
-Delve currently supports two versions of its API. By default a headless instance of `dlv` will serve APIv1, however new clients should use APIv2 as new features will only be made available through version 2. To select APIv2 use `--api-version=2` command line argument.
+Delve currently supports two versions of its API. By default a headless instance of `dlv` will serve APIv1 for backward compatibility with old clients, however new clients should use APIv2 as new features will only be made available through version 2. To select APIv2 use `--api-version=2` command line argument. 
+Clients can also select APIv2 by sending a [SetApiVersion](https://godoc.org/github.com/go-delve/delve/service/rpccommon#RPCServer.SetApiVersion) request specifying `APIVersion = 2` after connecting to the headless instance.
 
 # API version 2 documentation
 
--- a/Documentation/usage/dlv.md
+++ b/Documentation/usage/dlv.md
@@ -20,7 +20,7 @@ Pass flags to the program you are debugging using `--`, for example:
 
 ```
       --accept-multiclient   Allows a headless server to accept multiple client connections.
-      --api-version int      Selects API version when headless. (default 1)
+      --api-version int      Selects API version when headless. New clients should use v2. Can be reset via RPCServer.SetApiVersion. See Documentation/api/json-rpc/README.md. (default 1)
       --backend string       Backend selection (see 'dlv help backend'). (default "default")
       --build-flags string   Build flags, to be passed to the compiler.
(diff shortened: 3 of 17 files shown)


aarzilli:
cc @polinasok

Documentation/usage/dlv.md
```diff
@@ -20,7 +20,7 @@ Pass flags to the program you are debugging using `--`, for example:
 
 ```
       --accept-multiclient   Allows a headless server to accept multiple client connections.
-      --api-version int      Selects API version when headless. (default 1)
+      --api-version int      Selects API version when headless (see Documentation/api/json-rpc/README.md). (default 1)
```
polinasok, line 23: I think people generally prefer not to open their browser and follow links when doing --help on the command line, so it might be helpful to include the info that is needed here.
What about:

Selects API version when headless. New clients should use v2. Can be reset via RPCServer.SetApiVersion. See Documentation/api/json-rpc/README.md.

Documentation/api/ClientHowto.md
```diff
@@ -65,6 +65,16 @@ Delve will respond by sending a response packet that will look like this:
 {"id":27, "result": {"Breakpoint": {"id":3, "name":"", "addr":4538829, "file":"/User/you/some/file.go", "line":16, "functionName":"main.main", "Cond":"", "continue":false, "goroutine":false, "stacktrace":0, "LoadArgs":null, "LoadLocals":null, "hitCount":{}, "totalHitCount":0}}, "error":null}
 ```
 
+## Selecting the API version
+
+Delve currently supports two version of its API, APIv1 and APIv2. By default
+a headless instance of `dlv` will serve APIv1, however new clients should
```
polinasok, line 71: It is a bit odd that the default is not the recommended version. I think it is worth being explicit about it in the documentation by adding something like "will serve APIv1 for backward-compatibility with older clients".

polinasok:
There is something off with the formatting of the top comment, which makes "Fixes #2066" not a link. I don't have edit permissions to fix it myself. Could you please?

Documentation/api/ClientHowto.md
aarzilli: Done.

Documentation/usage/dlv.md
aarzilli: Done.

aarzilli:
> There is something off with the formatting of the top comment, which makes "Fixes #2066" not a link. I don't have edit permissions to fix it myself. Could you please?

Do you mean the opening message of the PR? Ages ago I wrote a program to automatically create PRs from branches, similar to `git codereview mail`, because clicking on the web interface and manually copying the commit message was boring. I put the commit message in backticks because otherwise markdown would mess up the formatting, that's why it doesn't have a link.

Documentation/api/json-rpc/README.md
```diff
@@ -8,7 +8,8 @@ Here is an (incomplete) [list of language implementations](http://json-rpc.org/w
 
 # API versions
 
-Delve currently supports two versions of its API. By default a headless instance of `dlv` will serve APIv1, however new clients should use APIv2 as new features will only be made available through version 2. To select APIv2 use `--api-version=2` command line argument.
+Delve currently supports two versions of its API. By default a headless instance of `dlv` will serve APIv1, however new clients should use APIv2 as new features will only be made available through version 2. To select APIv2 use `--api-version=2` command line argument.
```
polinasok, line 11: This one is missing "for backward-compatibility
with older clients"

aarzilli: Done.

PR #2088, opened 2020-06-19
Update cosiner/argv to v0.1.0

gadelkareem:
Fixes breaking changes in github.com/cosiner/argv v0.1.0

Fixes https://github.com/beego/bee/pull/665 but there are probably a lot of other libs cannot use `go get -u` because of this

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/go-delve/delve
 go 1.11
 
 require (
-	github.com/cosiner/argv v0.0.0-20170225145430-13bacc38a0a5
+	github.com/cosiner/argv v0.1.0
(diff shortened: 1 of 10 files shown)


aarzilli:
I'm not interested on walking the upgrade threadmill, dependencies are pinned to a specific version for a reason.

gadelkareem:
The problem is that bee is using delve and it must be included in the go.mod file. I tried to replace  `github.com/cosiner/argv` with the version you pinned but that only works for `go mod` but you try to `go get -u github.com/beego/bee` it will fetch it from github.

derekparker:
I don't see the harm in updating the dependency. This patch provides a fix to a breaking change and updates the dep, seems reasonable.

Issue #2079, opened 2020-06-09
Weird stepOut behavior on last method in interface (alphabetically) in go1.13


nd:
1. What version of Delve are you using (`dlv version`)?

```
Delve Debugger
Version: 1.4.1
Build: 708eadd5538edd222d8553b08b47c295474f38fa
```

2. What version of Go are you using? (`go version`)?

```
go version go1.13.12 windows/amd64
```

3. What operating system and processor architecture are you using?

windows/amd64

4. What did you do?

Debug the program:

```
package main

type I interface {
	foo()
	bar()
}

type T struct {}

func (t T) foo() {
	println("ok")
}

func (t T) bar() {
	println("ok")
}

func main() {
	t := T{}
	var i I = t
	i.bar() //break
 	i.foo()
}
```

Once breakpoint is reached, step into method bar, and then stepOut. Current location is reported to be main.main as expected. Step into method foo() and stepOut.

5. What did you expect to see?

Current location is main.main.

6. What did you see instead?

Current location is main.(T*).foo in main.go:1.

Interestingly if I add a method which is alphabetically after foo (e.g. foo1 or fop) to the interface and implement it for T, then stepOut from foo starts to work properly. Adding a method which is alphabetically before foo() doesn't change foo's stepOut behavior. Also stepOut seems to work fine in go1.14.

Full log: https://gist.github.com/nd/04a2e4685c496b995c6c44e43dbda853

chainhelen:
I check it on linux,  same issue when go1.13.12 linux/amd64 && dlv 1.4.1

aarzilli:
I think that our interpretation of debug_line matches the specification. Seeing that:

1. this has been fixed in 1.14 because the code that emits debug_line changed, and
2. the effect is just cosmetic (the autogen filter not working doesn't impact functionality)

I'd say that we shouldn't do anything about this. If you have objections please reopen.

PR #2089, opened 2020-06-20
proc: use file:line at entry point in skipAutogeneratedWrappersOut

### LINKED ISSUE #2086: Stepout skips one-line methods and functions

```
Delve Debugger
Version: 1.4.1
Build: 4af62f57b5c9ee31ccddb1a091cb0201d89803f8
```

```
go version go1.14.4 windows/amd64
```

but it seems like all go1.14.* versions are affected

Windows10/amd64, but it also happens on linux and mac.

4. What did you do?

Debugged the program:

```
package main

var i int

type T struct {}

func (t T) m() { stop() }

func stop() {
     println() //break
     i++
}

func main() { T{}.m() }
```

Once the breakpoint is reached stepped out.

Program stops at main.go:7 (`func (t T) m() { stop() }`)

6. What did you see instead?

Program stops at go1.14.4/src/runtime/proc.go:212 (T.m and main are skipped).

Full log: https://gist.github.com/nd/0f41486ebf5a2d9a24dbad611131df00

aarzilli:
The file:line information for the entrypoint is more acccurate than the
file:line information at a return point, which could be affected by a
compiler bug.

Fixes #2086


aarzilli:
@nd I think this also fixes the similar bug you reported a couple of weeks ago that I said I wasn't going to fix.

However you will also find out that there is another bug now, namely when stepping out you'll end up in <autogenerated>:1 instead of a sensible line. This is a bug with the compiler and should be reported to golang/go. I think @thanm already knows what this bug is, it was discussed in https://go-review.googlesource.com/c/go/+/235739

thanm:
@aarzilli thanks for the reminder. Filed a bug to track this:

https://github.com/golang/go/issues/39757

It is late in the release cycle, so this fix won't appear until 1.16 I think.

nd:
@aarzilli did you mean it can also fix #2079? As far as I can see it doesn't change behavior at least for go1.13.12 on windows. Step out from the alphabetically last method on interface still reports position on the first line of a file (position is not marked as <autogenerated>). Currently we workaround it by sending one more stepout if there are no function declarations on line 1.

aarzilli:
> did you mean it can also fix #2079?

yes.

> As far as I can see it doesn't change behavior at least for go1.13.12 on windows.

Strange.

> Step out from the alphabetically last method on interface still reports position on the first line of a file (position is not marked as autogenerated). 

Are you checking what stepout does or what `stack` reports? Because after this the behavior of skipAutogeneratedWrappersOut will not be based on the line reported by `stack` anymore.

nd:
> Are you checking what stepout does or what stack reports? 

As far as I can see they both report position at main.go:1: https://gist.github.com/nd/19d014ad9cce2ba8e9e297ee871c38cc.

PR #2093, opened 2020-06-26
service/dap: Support stackTrace request

polinasok:


service/dap/handles.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func (hs *handlesMap) create(value interface{}) int {
+	next := hs.nextHandle
+	hs.nextHandle++
+	hs.handleToVal[next] = value
+	return next
+}
+
+func (hs *handlesMap) get(handle int) (interface{}, bool) {
```
polinasok, line 31: The next PR I have in the pipeline uses this function. I guess I could take it out now and add it there.

service/dap/server.go
```diff
@@ -623,10 +625,57 @@ func (s *Server) onPauseRequest(request *dap.PauseRequest) { // TODO V0
...
+
+	goroutineID := request.Arguments.ThreadId
+	// Borrow the defaults for the arguments from the original vscode-go adapter.
+	// TODO(polina): is 'startFrame'+'levels' a better alternative for depth?
+	depth := 50 // length of the returned list of Stackframes
+	// TODO(polina): this is hardcoded throughout the codebase - make this a constant?
+	cfg := &api.LoadConfig{FollowPointers: true, MaxVariableRecurse: 1, MaxStringLen: 64, MaxArrayValues: 64, MaxStructFields: -1}
+	opts := api.StacktraceOptions(0)
+
+	locs, err := s.debugger.Stacktrace(goroutineID, depth, opts, api.LoadConfigToProc(cfg))
```
aarzilli, line 648: You should pass nil as the last argument if you aren't using frame.Locals/frame.Arguments.

service/dap/handles.go
aarzilli: You can leave it.

service/dap/server.go
```diff
@@ -623,10 +625,57 @@ func (s *Server) onPauseRequest(request *dap.PauseRequest) { // TODO V0
...
+		s.sendErrorResponse(request.Request, UnableToProduceStackTrace, "Unable to produce stack trace", err.Error())
+		return
+	}
+
+	stackFrames := make([]dap.StackFrame, len(locs))
+	for i, loc := range locs {
+		uniqueStackFrameID := s.stackFrameHandles.create(stackFrame{goroutineID, i})
+		stackFrames[i] = dap.StackFrame{Id: uniqueStackFrameID, Line: loc.Line}
+		if loc.Function != nil {
+			stackFrames[i].Name = loc.Function.Name()
```
aarzilli, line 659: Unless there's a specific reason why unknown functions should be called `<unknown>` you should call `loc.Function.Name()` directly, without the if.

derekparker:
> @derekparker: it's bothering me somewhat that dap is calling debugger.go which converts the result into a service/api type that then dap converts into a dap type. I think that debugger.go should return proc.StackFrame and then service/rpc* and service/dap should do the conversion starting from that.
> 
> Same thing for debugger.(*Debugger).ListGoroutines.
> 
> I'm not saying we should do it with this PR but what do you think?

I agree it's definitely not ideal. Let's not do anything in this PR but it's certainly worth looking into how to simplify the various conversions. It may be that the API layer directly converts to the type it needs to return from the raw proc types. The downside to this is it kind of breaks the contract we set up where the API layer doesn't really know about the lower-level proc types.

service/dap/server.go

PR #2098, opened 2020-07-07
service: move some type conversions from service/debugger to rpc pkgs

aarzilli:
Move the conversion of some 'proc' types from service/debugger into
service/rpc1 and service/rpc2. The methods of
service/debugger.(*Debugger) are also used by service/dap which
requires these types to be converted differently and converting them
twice is inefficent and doesn't make much sense.


service/debugger/debugger.go
```diff
@@ -1401,15 +1362,15 @@ func (d *Debugger) Ancestors(goroutineID, numAncestors, depth int) ([]api.Ancest
...
+		r[i].Stack, err = d.ConvertStacktrace(frames, nil)
 		if err != nil {
 			r[i].Unreadable = fmt.Sprintf("could not read ancestor stacktrace: %v", err)
 		}
 	}
 	return r, nil
 }
 
-func (d *Debugger) convertStacktrace(rawlocs []proc.Stackframe, cfg *proc.LoadConfig) ([]api.Stackframe, error) {
+func (d *Debugger) ConvertStacktrace(rawlocs []proc.Stackframe, cfg *proc.LoadConfig) ([]api.Stackframe, error) {
```
derekparker, line 1373: I think we should move this off of `Debugger` and into `api` as it's own function. Same goes for any other `Convert*` methods on `Debugger`.

Can happen in another commit though.

aarzilli:
> I feel that if debugger is going to start returning proc types and leave the conversion to the caller it should do so consistently

I don't disagree, but there are some problems with doing that. Looking at the api types we use in debugger:

1. `api.DiscardedBreakpoint` and `api.Breakpoint` don't have equivalent types in `proc` since `proc.Breakpoint` is a physical breakpoint, `proc` doesn't care about logical breakpoints.
2. `api.DebuggerState` doesn't have anything equivalent in `proc` and fairly complicated to put together outside of `debugger`.
3. Everything that takes `api.EvalScope` as input would have to be changed to take a `(goroutine ID, frame index, deferred call index)` triplet. I think this is very inconvenient
4. `api.Stackframe` is different from `proc.Stackframe` because it has locals and arguments, we can't get those without the `proc.Target` object but that's private to `debugger.Debugger` (as it should be). Doing it by calling `debugger.(*Debugger).LocalVariables` on every frame would make the conversion quadratic, so we'd need another method in debugger with the signature `debugger.(*Debugger).LocalVariablesForFrame(cfg *proc.LoadConfig, frames []proc.Stackframe)` which would return the variables for `frames[0]` (but wants the whole list of frames for optimization reasons). Also where would the ConvertStacktrace live? It can't be `service/api` because it would create an import loop between it and `service/debugger`.
5. `api.Thread`, `api.Registers`, `api.Location`, `api.AsmInstructions`, `api.Checkpoints`, `api.Image`, `api.PackageBuildInfo` and `api.Defer` are easy

We could do all types listed in (5) with this PR, it would take care of all the low hanging fruits. Doing `api.DebuggerState` would be worthwhile: `service/dap` calls `Command` but doesn't need all the stuff that `Command` spends time collecting, but it's complicated enough that it would merit its own PR.

derekparker:
> > I feel that if debugger is going to start returning proc types and leave the conversion to the caller it should do so consistently
> 
> I don't disagree, but there are some problems with doing that. Looking at the api types we use in debugger:
> 
>     1. `api.DiscardedBreakpoint` and `api.Breakpoint` don't have equivalent types in `proc` since `proc.Breakpoint` is a physical breakpoint, `proc` doesn't care about logical breakpoints.
> 
>     2. `api.DebuggerState` doesn't have anything equivalent in `proc` and fairly complicated to put together outside of `debugger`.
> 
>     3. Everything that takes `api.EvalScope` as input would have to be changed to take a `(goroutine ID, frame index, deferred call index)` triplet. I think this is very inconvenient
> 
>     4. `api.Stackframe` is different from `proc.Stackframe` because it has locals and arguments, we can't get those without the `proc.Target` object but that's private to `debugger.Debugger` (as it should be). Doing it by calling `debugger.(*Debugger).LocalVariables` on every frame would make the conversion quadratic, so we'd need another method in debugger with the signature `debugger.(*Debugger).LocalVariablesForFrame(cfg *proc.LoadConfig, frames []proc.Stackframe)` which would return the variables for `frames[0]` (but wants the whole list of frames for optimization reasons). Also where would the ConvertStacktrace live? It can't be `service/api` because it would create an import loop between it and `service/debugger`.
> 
>     5. `api.Thread`, `api.Registers`, `api.Location`, `api.AsmInstructions`, `api.Checkpoints`, `api.Image`, `api.PackageBuildInfo` and `api.Defer` are easy
> 
> 
> We could do all types listed in (5) with this PR, it would take care of all the low hanging fruits. Doing `api.DebuggerState` would be worthwhile: `service/dap` calls `Command` but doesn't need all the stuff that `Command` spends time collecting, but it's complicated enough that it would merit its own PR.

Sorry for the delay in response on this.

I think let's move forward with the rest of the low-hanging fruit via this PR. Then, before or upon merging let's distill the other comments you had (which I agree with) and put them into an issue which we can use to track and discuss the rest of the work left to do.

aarzilli:
Done, also see #2161. I was wrong about api.Location, it's actually complicated to remove for the same reason as api.Breakpoint is complicated.

derekparker (approved):
LGTM

Just needs rebase.

PR #2103, opened 2020-07-09
service/dap: resolve minor error code TODO

polinasok:
A while back I was able to clarify with vscode maintainers any questions we had about the error code values. @weinand wrote "The error numbers are private to some of our implementations. You should pick your own. There is no benefit of trying to reuse these numbers".


polinasok:
cc @eliben @hyangah

PR #2111, opened 2020-07-29
service/dap: Initial support for scopes and variables requests

polinasok:
Updates #1515


service/dap/server_test.go
```diff
@@ -375,6 +388,70 @@ func expectStackFrames(t *testing.T, got *dap.StackTraceResponse,
...
+	}
+}
+
+// expectVar is a helper for verifying the values within a VariablesResponse.
+//     i - index of the scope within VariablesRespose.Body.Variables array
+//     name - name of the variable
+//     value - the value of the variable
+//     useExactMatch - true if value is to be compared to exactly, false if to be used as regex
+//     ref - reference to retrieve children of this variable
+func expectVar(t *testing.T, got *dap.VariablesResponse, i int, name, value string, useExactMatch bool, ref int) {
```
aarzilli, line 422: I think you should support `i == -1`, where you scan `got` for the first variable with that name, and then use this everywhere the order of variables doesn't matter. I think as it is this set of tests is too fragile. For example if the definition of the chan struct changes a little bit (because a field gets inserted) the tests will fail. Same thing for the order of variables returned by `LocalVariables`.

Same thing for `ref`, allow `ref < 0` to match anything and return the `VariableReference` field of the variable that's been tested, so that tests don't have to depend on variable references being produced in a certain order.

service/dap/server_test.go
```diff
@@ -500,9 +1082,14 @@ func runDebugSessionWithBPs(t *testing.T, client *daptest.Client, launchRequest
 
 	// Program automatically continues to breakpoint or completion
 
-	for _, onBP := range onBreakpoints {
+	for _, onBP := range onBPs {
```
aarzilli, line 1085: IMHO this would be clearer if it was more like withTestProcessArgs in proc_test: a single function pointer gets called here and then if it wants to continue it calls client.ContinueRequest/client.ExpectContinueResponse/client.ExpectStoppedEvent (possibly using a helper function).

However this is a personal opinion, so you can ignore this comment.

service/dap/server.go
```diff
@@ -686,16 +689,222 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
...
+		if v.Base == 0 {
+			value = "nil <" + v.Type + ">"
+		} else {
+			value = fmt.Sprintf("<%s> (length: %d)", v.Type, v.Len)
+			if len(v.Children) > 0 {
+				variablesReference = s.variableHandles.create(v)
+			}
+		}
+	case reflect.String:
+		if v.Unreadable != "" {
```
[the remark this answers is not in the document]
polinasok: done

service/dap/server.go
```diff
@@ -686,16 +689,222 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
...
+					Name:               key,
+					Value:              val,
+					VariablesReference: valref,
+				}
+				if keyref != 0 {
+					kvvar.VariablesReference = keyref
+				} else if valref != 0 {
+					kvvar.VariablesReference = valref
+				}
+				children = append(children, kvvar)
```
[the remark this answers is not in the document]
polinasok: I was worried about this, but decided that since we only expand the key while the value is obviously a non-expandable scalar, then it will not be confusing. The problem with the original TypeScript implementation is that it does not support key expansion in any scenario. It only supports non-scalar values. So instead of feature parity, I decided to actually fix things here. I am not super happy with how things look, but can't think of a better way. I am open to suggestions.

Before:
image
After:
image

service/dap/server_test.go
polinasok: I do see your point. Thank you for the suggestion. I added a TODO and will take this into consideration next time I have to reshuffle the tests.

service/dap/server_test.go
polinasok: Yes, that's a very good idea. And it would definitely make it easier to update the tests. Will give this a try next and push in a separate commit.

service/dap/server_test.go
```diff
@@ -375,6 +388,70 @@ func expectStackFrames(t *testing.T, got *dap.StackTraceResponse,
...
+	if len(got.Body.Scopes) <= i {
+		t.Errorf("\ngot  %d\nwant len(Scopes)>%d", len(got.Body.Scopes), i)
+	}
+	goti := got.Body.Scopes[i]
+	if goti.Name != name || goti.VariablesReference != varRef || goti.Expensive {
+		t.Errorf("\ngot  %#v\nwant Name=%q VariablesReference=%d Expensive=false", goti, name, varRef)
+	}
+}
+
+// expectVars is a helper for verifying the number of variables withing a VariablesResponse.
```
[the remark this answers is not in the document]

_fixtures/testvariables3.go
```diff
@@ -0,0 +1,20 @@
+package main
```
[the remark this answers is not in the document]
aarzilli: > I am glad you made the fix to hide this and make such test case unnecessary.

Note that it could still happen, either because of corrupted memory or because of non-live in-scope variables (although that would be pretty unlikely).

service/dap/server.go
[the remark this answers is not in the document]
aarzilli: It looks somewhat strange but better than I thought, I guess it's fine.

service/dap/server_test.go
aarzilli: Do you mean a separate commit in this PR or a separate PR?

_fixtures/testvariables3.go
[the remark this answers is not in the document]
polinasok: Yes of course. The code still has extra checks.

service/dap/server.go
```diff
@@ -686,16 +689,222 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
...
+func (s *Server) onScopesRequest(request *dap.ScopesRequest) {
+	sf, ok := s.stackFrameHandles.get(request.Arguments.FrameId)
+	if !ok {
+		s.sendErrorResponse(request.Request, UnableToListLocals, "Unable to list locals", fmt.Sprintf("unknown frame id %d", request.Arguments.FrameId))
+		return
+	}
+
+	scope := api.EvalScope{GoroutineID: sf.(stackFrame).goroutineID, Frame: sf.(stackFrame).frameIndex}
+	// TODO(polina): Support setting config via launch/attach args
+	cfg := &api.LoadConfig{FollowPointers: true, MaxVariableRecurse: 1, MaxStringLen: 64, MaxArrayValues: 64, MaxStructFields: -1}
```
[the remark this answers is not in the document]
polinasok: I changed the code to skip the conversation. I think I saw this way of doing of setting up and using a default config somewhere else in the code and did not question it much. But now that I looked closely, I can just use proc.LoadConfig directly, can't I?

service/dap/server_test.go
polinasok: Already pushed a commit in this PR last night.

polinasok:
@aarzilli, did you have a chance to check out the odd test failures in the jobs that are allowed to fail?

polinasok:
Thanks @aarzilli for offline explanation. Those jobs are failing "because of the change on how we scope variables amb1 will be in scope for go < 1.15 but out of scope for go 1.15".

polinasok:
The tip tests are now also passing.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: I found a bug and just pushed a fix for this case. Even though I was returning the right number of elements, vscode only displayed a single one because all compound keys had the same value of their type. Now this looks like so:
image

service/dap/server.go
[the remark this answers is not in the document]
aarzilli: Yes.

service/dap/server_test.go
```diff
@@ -442,7 +536,472 @@ func TestStackTraceRequest(t *testing.T) {
...
+
+					client.VariablesRequest(1000)
+					args := client.ExpectVariablesResponse(t)
+					expectChildren(t, args, "Arguments", 0)
+
+					// Locals
+
+					client.VariablesRequest(1001)
+					locals := client.ExpectVariablesResponse(t)
+					expectChildren(t, locals, "Locals", 93)
```
aarzilli, line 835: Given the problem with Go 1.15 I'd say we shouldn't test this. Besides you are already checking the variables individually, this shouldn't break if we add a new one.

polinasok: I already reworked the test, so 1.15 changes don't cause different behavior for this check.

aarzilli: I see, still I don't think we should do this test, it doesn't add much and it'll break if we add something to testvariables2.go.

PR #2122, opened 2020-08-06
Add arm-linux support to delve

puppywang:
This is the intial support of arm arch for linux os, to address the problem in this issue https://github.com/go-delve/delve/issues/20,  The arm-linux's ptrace is lacking support of PTRACE_SINGLESTEP, so we need to emulate it by setup breakpoint at any possible next instructions, and then continue execution. I try to fix all the bug that I found, but there may still have plenty of them. Any code review comments is welcome, thanks.


chainhelen:
So great.

pkg/proc/native/threads_linux_arm.go
```diff
@@ -0,0 +1,247 @@
...
+							}
+							pc += regVal
+						}
+					}
+					nextPcs = append(nextPcs, pc)
+				}
+			}
+			return nextPcs, nil, nil
+		}
+		nextPcs, nops, err := resolvePCs()
```
aarzilli, line 179: Shouldn't all this only happen once, outside of the for loop?

pkg/proc/native/threads_linux_arm.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (t *nativeThread) singleStep() (err error) {
+	// Arm don't have ptrace singleStep implemented, so we use breakpoint to emulate it.
+	for {
+		regs, err := t.Registers()
+		if err != nil {
+			return err
+		}
+		resolvePCs := func() ([]uint64, []uint64, error) {
```
aarzilli, line 57: I think this function would be easier to understand if it was defined outside of (*nativeThread).singleStep

pkg/proc/native/threads_single_step_linux.go
```diff
@@ -0,0 +1,32 @@
+// +build !arm
```
aarzilli, line 1: This file shoud be called threads_hardware_singlestep_linux.go

pkg/proc/target_exec.go
```diff
@@ -125,7 +125,7 @@ func (dbp *Target) Continue() error {
 			case loc.Fn.Name == "runtime.breakpoint":
 				// In linux-arm64, PtraceSingleStep seems cannot step over BRK instruction
 				// (linux-arm64 feature or kernel bug maybe).
-				if !arch.BreakInstrMovesPC() {
```
aarzilli, line 128: You should make BreakInstrMovesPC return true if you don't want this for "arm"

pkg/proc/native/threads_single_step_linux.go
puppywang: Got it.

pkg/proc/native/threads_linux_arm.go
puppywang: Yes, I will fix it.

pkg/proc/target_exec.go
puppywang: Golang use UND instruction `0xf0, 0x01, 0xf0, 0xe7` to act as break instruction, this instruction will cause ptrace to stop at the PC of current UND instruction, so BreakInstrMovesPC should be false I think, and as we do implementation the singleStep, we are able to move PC no matter what instruction it is, so I change there to compare with `BreakInstrMovesPC` to compare with `arm64` arch, as the comment here said `PtraceSingleStep seems cannot step over BRK instruction // (linux-arm64 feature or kernel bug maybe).`

pkg/proc/native/threads_linux_arm.go
puppywang: I will refactor it as member function of nativeThread.

pkg/proc/native/threads_linux_arm.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func (t *nativeThread) singleStep() (err error) {
+	// Arm don't have ptrace singleStep implemented, so we use breakpoint to emulate it.
+	for {
+		regs, err := t.Registers()
+		if err != nil {
+			return err
+		}
+		nextPcs, nops, err := t.resolvePC(regs)
```
aarzilli, line 182: Shouldn't this, the execPtraceFunc call and the defer all be outside of the for loop? I don't see how it could work like this.

pkg/proc/target_exec.go
aarzilli: In that case I think you should revert this change to check for BreakInstrMovesPC and then remove the special code in resolvePCs and singleStep to NOP breakpoint instructions.

puppywang: Got, I change it back to check against BreakInstrMovesPC , thanks

pkg/proc/native/threads_linux_arm.go
puppywang: Got, I've move those logic outside the for loop, and they only need to be run once.

pkg/proc/native/threads_linux_arm.go
```diff
@@ -0,0 +1,237 @@
...
+	return restoreRegistersErr
+}
+
+// resolvePCForArm is used to resolve all next PC for current instruction.
+func (t *nativeThread) resolvePC(regs proc.Registers) ([]uint64, error) {
+	// Use ptrace to get better performance.
+	nextInstrLen := t.BinInfo().Arch.MaxInstructionLength()
+	nextInstrBytes := make([]byte, nextInstrLen)
+	var err error
+	t.dbp.execPtraceFunc(func() {
```
aarzilli, line 55: Why not just use t.ReadMemory (here and in the other places in this function where PtracePeekData is called)?

aarzilli (commented):
In a previous comment I was asking "How do we test this?". This is a question we need to find an answer to. We like to have continuous integration for the architectures we support.

pkg/proc/native/threads_linux_arm.go
puppywang: As this is inside low level emulation of ptrace singlestep, I want to make sure it first enter the context of ptrace channel, then do the peek and poke work, so I skip the upper Read/WriteMemory function, and use ptrace function directly.

puppywang:
> In a previous comment I was asking "How do we test this?". This is a question we need to find an answer to. We like to have continuous integration for the architectures we support.

Yes, currently I do only a little test on raspberry pi4, and I am looking for help to make a more coverage test, do we have any guideline to make such test on different platform? Thanks.

aarzilli:
> Yes, currently I do only a little test on raspberry pi4, and I am looking for help to make a more coverage test, do we have any guideline to make such test on different platform? Thanks.

So far we've gotten along using free services like travis, see `.travis.yml`. It doesn't have 32bit arm but I don't know if it's possible to do the same thing we do with linux/386.

aarzilli:
Regarding testing this, I tried just adding an `export GOARCH=arm` to the build script. This works on my raspberry pi but it doesn't on Travis-CI, presumably because they didn't enable the appropriate kernel option. 
Result of the experiment here: https://github.com/go-delve/delve/pull/2218

Additionally the tests hang randomly on my raspberrypi inside `native.(*nativeThread).singleStep`.

puppywang:
> (*

Got, let me check with my arm device.

alexis-evelyn:
Can confirm d75633f5b1e444a876ede6ef99b77d845c98f443 works on my Raspberry Pi 3 with an armhf type processor.

daresan:
So no success, to make delve work on a 32bit raspberry Pi OS yet? Just asking, because I just spent more than 2 hours trying to find a 32bit version for raspbian. It should be mentioned in the documentation.

xcaptain:
Also got a raspberry pi and can't run dlv
image
would take some time to try this branch and see if it works on raspberry pi 4b+

puppywang:
> Also got a raspberry pi and can't run dlv
> image
> would take some time to try this branch and see if it works on raspberry pi 4b+

Before this merge is accept, you could try my working branch, try following cmd.

```
git clone https://github.com/puppywang/delve
go install
``
And try debug with delve.

krasi-georgiev:
magic I just tested it and it works!

krasi-georgiev:
spoke too soon, stepping in almost never works.

derekparker:
I've been holding off on a deeper review of this since we've never been able to answer the question of how do we have CI for this? I agree 32-bit ARM support would be great, but we need CI to verify this (I don't even have a Pi or other 32-bit ARM system to test this on, unfortunately).

aarzilli:
@derekparker we could probably do the same thing we do on 386 and run it inside docker. If the kernel is compiled with compatibility with arm32 binaries the tests can be run directly on arm64. See my comment https://github.com/go-delve/delve/pull/2122#issuecomment-721048134. The problem however is that it randomly hangs inside singleStep.

derekparker:
> @derekparker we could probably do the same thing we do on 386 and run it inside docker. If the kernel is compiled with compatibility with arm32 binaries the tests can be run directly on arm64. See my comment [#2122 (comment)](https://github.com/go-delve/delve/pull/2122#issuecomment-721048134). The problem however is that it randomly hangs inside singleStep.

Ack, alright that makes sense.

@puppywang will you be able to investigate this outstanding issue?

chrisdutz:
Another option I discovered recently would be to use a GitHub Action and to define a "local runner" ... there a RPi could possibly be setup as local runner.

krasi-georgiev:
or use the new Balena open fleet :)

https://www.balena.io/blog/introducing-open-fleets-and-self-submitted-apps-and-blocks-on-balenahub/?mc_cid=092773ae21#open-fleets

chrisdutz:
So yesterday I successfully used what we use for building our application for arm32 devices and was able to compile the arm version on my normal ubuntu laptop (So it should be buildable on a normal CI agent or in a Docker container). I was able to run the Delve binary on my edge device, but yes ... I also did encounter these lockups. 

Is there anything I can help with?

jochen42:
i am using a patched dlv-binary (the code of the mr) and go 1.17 on linux-arm7 and remote-debugging. work's like charm until now.
do you plan to integrate this into the stable binanry? Can i support?

aarzilli:
> i am using a patched dlv-binary (the code of the mr) and go 1.17 on linux-arm7 and remote-debugging. work's like charm until now.
> do you plan to integrate this into the stable binanry? Can i support?

Last time I looked at this it didn't pass most of the tests reliably, so there are no plans at the moment to merge this. If you want to rebase it and fix the problems yourself see: https://github.com/go-delve/delve/blob/master/Documentation/internal/portnotes.md

chrisdutz:
We decided to simply switch the platform ... with Arm64 we're quite happy with how things are working. Anyway ... thanks for your hard work on this.

derekparker:
Closing stale PR.

PR #2123, opened 2020-08-08
proc: prevent internal breakpoint conditions from failing

### LINKED ISSUE #2113: Stepout doesn't preserve current goroutine

Version: 1.4.1
Build: f9c8f7f55b798a472013dd6a20bf688387eb0c53

go1.15rc1

linux/amd64

4. What did you do?

Debugging slightly modified version of https://github.com/aarzilli/delve_client_testing/blob/master/08gswitch/main.go:

```
package main

import (
	"fmt"
	"sync"
	"runtime"
)

func coroutine(i int, start, finish *sync.WaitGroup) {
	defer finish.Done()

	j := i * 2

	if i == 99 {
		runtime.Breakpoint()
		start.Done()
	} else {
		start.Wait()
  }

	fmt.Println("hello ", i, j)
	fmt.Println("goodbye", i, j)
}

func main() {
	i := 0
	var start, finish sync.WaitGroup
	start.Add(1)
	for ; i < 100; i++ {
		finish.Add(1)
		go coroutine(i, &start, &finish)
	}
	finish.Wait()
	println(i)
}
```

Once a breakpoint is reached, switched to another goroutine with main.coroutine in its stack. Stepped out.

Stepout preserves the goroutine I've switched to and returns a state where current thread runs the selected goroutine.

6. What did you see instead?

Step out returns the current thread with a different goroutine (0 instead 6). 

This doesn't always happen, sometimes stepout returns the expected result.

Full log: https://gist.github.com/nd/038386b9fbdfe306753cdf17127fdb43.

aarzilli:
An internal breakpoint condition shouldn't ever error:
* use a ThreadContext to evaluate conditions if a goroutine isn't
available
* evaluate runtime.curg to a fake g variable containing only
`goid == 0` when there is no current goroutine

Fixes #2113


aarzilli:
@nd please take a look
		
@derekparker What do we do with freebsd? The new test for this issue doesn't work there, because somehow it triggers a case where the native backend doesn't work correctly. There's multiple tests that have been failing for years and it doesn't work with async breakpoints. I don't want to delete it but I also don't want to spend my time making it work. Should we write something in the docs about it?

nd:
I cannot reproduce the original issue anymore.

Very seldom (1 in several thousands runs) I get a strange behavior when the program exits unexpectedly.

We reach a breakpoint and switch to another goroutine without resuming a program. After that we try to step out, but program said to be terminated. I saw it happened when the goroutine we switched to was inside runtime.deferproc (https://gist.github.com/nd/cabc370e70a840d4a2172b28d612b08c) and inside sync.(*WaitGroup).Wait (https://gist.github.com/nd/e566cff5ac7424d9204d9e196618fb6d).

aarzilli:
Both of those cases are goroutines executing on the system stack so the return frame happens to be systemstack_switch. They could be doing some stack manipulation we can't follow or we could have some problem specifically with systemstack_switch, either way it isn't interesting enough to fix. We should really be using a different stacktrace mode for stepout on the system stack, but it isn't a high priority problem since basically nobody uses delve for this.

PR #2124, opened 2020-08-12
Documentation: update supported architectures

aarzilli:


derekparker:
We also support i386 which isn't mentioned.

PR #2128, opened 2020-08-13
cmd/dlv: subcommand 'dlv test' should switch to package directory

### LINKED ISSUE #2125: Should dlv test chdir to the package directory like go test does?
```
Delve Debugger
Version: 1.4.1
Build: $Id: bda606147ff48b58bde39e20b9e11378eaa4db46 $
```

2. What version of Go are you using? (`go version`)? `go version go1.13.6 linux/amd64`
3. What operating system and processor architecture are you using? Linux x86/64
4. What did you do?
```
# On a test that attempts to read a file from ./testdata/foo.txt
dlv test ./mypackage
```

Similar to what `go test` does, `dlv test` chdirs to `./mypackage` before running the test, so the test can open the file

6. What did you see instead?
Test fails to open the file, because dlv stays in the current directory

aarzilli:
Match 'go test' behaviour and switch to package directory, unless one
is specified with the '--wd' option.

Fixes #2125


derekparker:
This does change behavior which some might be surprised with in a patch release. I don't feel that holding this off to preserve semver semantics is worth it however.

PR #2140, opened 2020-08-19
proc/native/windows: do not call _DebugBreakProcess on a stopped process

### LINKED ISSUE #2138: 'Access is denied' error after sequence of halt commands on windows
```
Delve Debugger
Version: 1.5.0
Build: 328cf87808822693dc611591519689dcd42696a3
```

```
go version go1.15 windows/amd64
```

windows/amd64

4. What did you do?

Debugged the program:
```
package main

import (
	"time"
)

func main() {
	time.Sleep(time.Hour)
	println("ok")
}
```

Send 3 halt requests in a row and then continue.

Program resumed

6. What did you see instead?

Got the 'Access is denied.' error. It happens inside native._SuspendThread(). SysInternals process explorer also cannot open the thread which Delve fails to suspend. The thread is always the DbgUiRemoteBreakin thread created by windows. After several halts there are number of such threads in the process, Delve tries to suspend threads including DbgUiRemoteBreakin and fails. It looks like continue doesn't work after that, the returned state always has running=false.

Rpc log: https://gist.github.com/nd/d42c18b597602fbb0e61b9958318b1c2.

Is it responsibility of a client to avoid sending halt when process is already halted?

aarzilli:
Fixes #2138

### DIFF
--- a/_fixtures/issue2138.go
+++ b/_fixtures/issue2138.go
@@ -1,0 +1,10 @@
+package main
+
+import (
+	"time"
+)
+
+func main() {
+	time.Sleep(time.Hour)
+	println("ok")
+}
--- a/pkg/proc/native/proc_windows.go
+++ b/pkg/proc/native/proc_windows.go
@@ -17,6 +17,7 @@ type osProcessDetails struct {
 	hProcess    syscall.Handle
 	breakThread int
 	entryPoint  uint64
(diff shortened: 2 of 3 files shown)


aarzilli:
cc @nd

nd:
Thank you, tested - it works.

PR #2146, opened 2020-08-22
Target redirects

aarzilli:
cmd,proc,terminal,debugger: Support default file descriptor redirects

Adds features to support default file descriptor redirects for the
target process:

1. A new command line flag '--redirect' and '-r' are added to specify file redirects for the target process
2. New syntax is added to the 'restart' command to specify file redirects.
3. Interactive instances will check if stdin/stdout and stderr are terminals and print a helpful error message if they aren't.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -412,16 +412,22 @@ Restart process.
 
 For recorded targets the command takes the following forms:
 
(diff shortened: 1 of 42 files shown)


derekparker:
Just resolve conflicts.

PR #2147, opened 2020-08-22
feature(examinemem): x -fmt <?> -count <?> -size <?> <address>, works like gdb `x/FMT`

hitzhangjie:
Make `examinemem` works like gdb x/FMT, in which FMT contains
- a number, represents a repeat times, similar to -count in examinemem
- a format letter, similar to -fmt in examinemem
- a size letter(b,d,w,h..), similar to -size in examinemem

Besides, I add a new format `address`, which has fixed size of sizeof(uintptr).

We can use it like:

1.address `x -fmt addr [-count ?]  <address>`
2.hex `x -fmt hex -count 16 -size 8  <address>`

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"text/tabwriter"
+	"unsafe"
 
 	"github.com/cosiner/argv"
 	"github.com/go-delve/delve/pkg/locspec"
@@ -390,15 +391,15 @@ If locspec is omitted edit will open the current source file in the editor, othe
 
 		{aliases: []string{"examinemem", "x"}, group: dataCmds, cmdFn: examineMemoryCmd, helpMsg: `Examine memory:
 
-	examinemem [-fmt <format>] [-len <length>] <address>
+	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <address>
 
-Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal),.
+Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal), addr(address).
 Length is the number of bytes (default 1) and must be less than or equal to 1000.
-Address is the memory location of the target to examine.
+Address is the memory location of the target to examine. Please note '-len' is deprecated by '-count and -size'.
 
 For example:
 
-    x -fmt hex -len 20 0xc00008af38`},
+    x -fmt hex -count 20 -size 1 0xc00008af38`},
 
 		{aliases: []string{"display"}, group: dataCmds, cmdFn: display, helpMsg: `Print value of an expression every time the program stops.
 
@@ -1562,7 +1563,9 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 
 	// Default value
 	priFmt := byte('x')
-	length := 1
+	count := 1
+	size := 1
+	ptrSize := int(unsafe.Sizeof(uintptr(0)))
 
 	for i := 0; i < len(v); i++ {
 		switch v[i] {
@@ -1580,24 +1583,32 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 				"decimal":     'd',
 				"bin":         'b',
 				"binary":      'b',
+				"addr":        'a',
+				"address":     'a',
 			}
 			priFmt, ok = fmtMapToPriFmt[v[i]]
 			if !ok {
 				return fmt.Errorf("%q is not a valid format", v[i])
 			}
-		case "-len":
+		case "-count", "-len":
 			i++
 			if i >= len(v) {
-				return fmt.Errorf("expected argument after -len")
+				return fmt.Errorf("expected argument after -count")
 			}
 			var err error
-			length, err = strconv.Atoi(v[i])
-			if err != nil || length <= 0 {
-				return fmt.Errorf("len must be an positive integer")
+			count, err = strconv.Atoi(v[i])
+			if err != nil || count <= 0 {
+				return fmt.Errorf("count must be an positive integer")
 			}
-			// TODO, maybe configured by user.
-			if length > 1000 {
-				return fmt.Errorf("len must be less than or equal to 1000")
+		case "-size":
+			i++
+			if i >= len(v) {
+				return fmt.Errorf("expected argument after -size")
+			}
+			var err error
+			size, err = strconv.Atoi(v[i])
+			if err != nil || size <= 0 {
+				return fmt.Errorf("size must be an positive integer")
 			}
 		default:
 			if i != len(v)-1 {
@@ -1611,16 +1622,28 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 		}
 	}
 
+	if priFmt == 'a' {
+		if size != 1 && size != ptrSize {
+			return fmt.Errorf("ptrsize on you mach is %d, not %d", ptrSize, size)
+		}
+		size = ptrSize
+	}
+
+	// TODO, maybe configured by user.
+	if count*size > 1000 {
+		return fmt.Errorf("read memory range (count*size) must be less than or equal to 1000 bytes")
+	}
+
 	if address == 0 {
 		return fmt.Errorf("no address specified")
 	}
 
-	memArea, err := t.client.ExamineMemory(address, length)
+	memArea, err := t.client.ExamineMemory(uintptr(address), count*size)
 	if err != nil {
 		return err
 	}
 
-	fmt.Print(api.PrettyExamineMemory(uintptr(address), memArea, priFmt))
+	fmt.Print(api.PrettyExamineMemory(uintptr(address), memArea, priFmt, size))
 	return nil
 }
 
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1627,12 +1627,11 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 		return fmt.Errorf("no address specified")
 	}
 
-	memArea, err := t.client.ExamineMemory(uintptr(address), count*size)
+	memArea, isLittleEndian, err := t.client.ExamineMemory(uintptr(address), count*size)
 	if err != nil {
 		return err
 	}
-
-	fmt.Print(api.PrettyExamineMemory(uintptr(address), memArea, priFmt, size))
+	fmt.Print(api.PrettyExamineMemory(uintptr(address), memArea, isLittleEndian, priFmt, size))
 	return nil
 }
 
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 	"strings"
 	"text/tabwriter"
-	"unsafe"
 
 	"github.com/cosiner/argv"
 	"github.com/go-delve/delve/pkg/locspec"
@@ -1565,7 +1564,6 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 	priFmt := byte('x')
 	count := 1
 	size := 1
-	ptrSize := int(unsafe.Sizeof(uintptr(0)))
 
 	for i := 0; i < len(v); i++ {
 		switch v[i] {
@@ -1583,8 +1581,6 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 				"decimal":     'd',
 				"bin":         'b',
 				"binary":      'b',
-				"addr":        'a',
-				"address":     'a',
 			}
 			priFmt, ok = fmtMapToPriFmt[v[i]]
 			if !ok {
@@ -1593,12 +1589,12 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 		case "-count", "-len":
 			i++
 			if i >= len(v) {
-				return fmt.Errorf("expected argument after -count")
+				return fmt.Errorf("expected argument after -count/-len")
 			}
 			var err error
 			count, err = strconv.Atoi(v[i])
 			if err != nil || count <= 0 {
-				return fmt.Errorf("count must be an positive integer")
+				return fmt.Errorf("count/len must be a positive integer")
 			}
 		case "-size":
 			i++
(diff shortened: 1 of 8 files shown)


chainhelen:
> @chainhelen Could you help me about the Travis CI build error? It seems like some testcases on linux/arm64 platform failed. How to fix this kind of problems? I cannot reproduce it in my development environment.

Don't worry. This error should have nothing to do with this submission. 
Some failures about linux/arm64 occur after go1.15.  
```
```

@hengwu0

pkg/terminal/command.go
```diff
@@ -1537,24 +1540,32 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
...
-				return fmt.Errorf("expected argument after -len")
+				return fmt.Errorf("expected argument after -count")
 			}
 			var err error
-			length, err = strconv.Atoi(v[i])
-			if err != nil || length <= 0 {
-				return fmt.Errorf("len must be an positive integer")
+			count, err = strconv.Atoi(v[i])
+			if err != nil || count <= 0 {
+				return fmt.Errorf("count must be an positive integer")
```
aarzilli, line 1558: s/an positive/a positive/

service/api/prettyprint.go
```diff
@@ -377,21 +386,25 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
 		colFormat = "%08b"
 	case 'o':
 		cols = 8
-		colFormat = "%04o" // Always keep one leading zero for octal.
+		colFormat = "0%03o" // Always keep one leading zero for octal.
```
aarzilli, line 389: The number in the format string should be `colBytes*number_of_character_in_a_byte_for_the_current _format`

service/api/prettyprint.go
```diff
@@ -377,21 +386,25 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
 	case 'd':
 		cols = 8
 		colFormat = "%03d"
 	case 'x':
 		cols = 8
 		colFormat = "0x%02x" // Always keep one leading '0x' for hex.
+	case 'a':
+		cols = 4             // Print 4 address in one row.
```
aarzilli, line 397: Why?

service/api/prettyprint.go
```diff
@@ -403,14 +416,40 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
+			}
 		}
 		fmt.Fprintln(w, "")
 		address += uintptr(cols)
 	}
 	w.Flush()
 	return b.String()
 }
+
+func byteArrayToUInt(arr []byte) interface{} {
```
aarzilli, line 437: 1. the idiomatic way of doing this is returning `(uint, error)`
2. but really the return type should be uint64 or maybe math/big.Int
3. if the former you should clamp the maximum value of size to 8

service/api/prettyprint.go
```diff
@@ -403,14 +416,40 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
+	reverse(arr)
```
aarzilli, line 438: The right way of doing this is with shift and addition, or encoding/binary.ReadUintX

aarzilli (commented):
I don't like the new 'address' format for two reasons:

1. It's not orthogonal to either other formats or sizes: as a format it is the same thing as hexadecimal, but unlike other formats it forces the size to be some specific number.
2. The size of a pointer shouldn't be determined by the frontend, consider the case of an instance of delve debugging a i386 core dump on a amd64 host, the correct pointer size is 4 but this code will determine it to be 8.

I think it would be better if a special value for size (for example -size 0) meant "size of the pointer" and the backend determined its value.

pkg/terminal/command.go
hitzhangjie: Thanks @aarzilli  for pointing out the problems.

The format of address is indeed hexidecimal. We can use `hexdecimal` to check that by `-fmt hex -size 8` or `-fmt hex -size 4`, to avoid the cases `a i386 core dump on a amd64 host`.

hitzhangjie:
> I don't like the new 'address' format for two reasons:
> 
> 1. It's not orthogonal to either other formats or sizes: as a format it is the same thing as hexadecimal, but unlike other formats it forces the size to be some specific number.
> 2. The size of a pointer shouldn't be determined by the frontend, consider the case of an instance of delve debugging a i386 core dump on a amd64 host, the correct pointer size is 4 but this code will determine it to be 8.
> 
> I think it would be better if a special value for size (for example -size 0) meant "size of the pointer" and the backend determined its value.

Thanks @aarzilli for pointing out the problems.

The format of address is indeed hexidecimal. We can use hexdecimal to check that by -fmt hex -size 8 or -fmt hex -size 4, to avoid the cases a i386 core dump on a amd64 host.

pkg/terminal/command.go
hitzhangjie: done

service/api/prettyprint.go
hitzhangjie: done

service/api/prettyprint.go
hitzhangjie: drop `address` format ... done

service/api/prettyprint.go
hitzhangjie: Here use uint64 and limit the maximum of `-size` to 8

service/api/prettyprint.go
hitzhangjie: aarzilli, I use shift and addition to reimplement this.

But I have another problem: how to determine the byte size? Here, I use little endian by default.
I just don't know how to determine the byte order in a core file.

hitzhangjie: done

service/api/prettyprint.go
aarzilli: Where?

hitzhangjie: Oh, I revert this. 
---------------
I suggest use 1byte zero padding to format the number, it will be shorter
and looks a little better, for example, a 8-byte hexidecimal, 0x12345678
may be better than 0x0000000012345678.

service/api/prettyprint.go
```diff
@@ -403,14 +411,53 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
 	return b.String()
 }
+
+func byteArrayToUInt64(buf []byte) (uint64, error) {
+	if len(buf) == 0 {
+		return 0, errors.New("empty buf")
+	}
+	var n uint64
+
+	if isLittleEndian() {
```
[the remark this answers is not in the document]
hitzhangjie: I understand your meaning now. Maybe the the machine where dlv runs has different architecture from the backend debugger server.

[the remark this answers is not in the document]
hitzhangjie: Hi, @aarzilli I changed the RPC method ExamineMemory letting it to return the `IsLittleEndian`.

And, PrettyExamineMemory still keeps the logic to calculate the number and print. I didn't move it to the RPC server side.
I think this way the ExamineMemory method may be reused in the future.

service/api/prettyprint.go
```diff
@@ -403,14 +411,53 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
 		fmt.Fprintln(w, "")
 		address += uintptr(cols)
 	}
 	w.Flush()
 	return b.String()
 }
+
+func byteArrayToUInt64(buf []byte) (uint64, error) {
+	if len(buf) == 0 {
+		return 0, errors.New("empty buf")
```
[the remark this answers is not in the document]
hitzhangjie: changed to `func byteArrayToUint64(buf []byte) uint64`

hengwu0:
> > @chainhelen Could you help me about the Travis CI build error? It seems like some testcases on linux/arm64 platform failed. How to fix this kind of problems? I cannot reproduce it in my development environment.
> 
> Don't worry. This error should have nothing to do with this submission.
> Some failures about linux/arm64 occur after go1.15.
> 
> @hengwu0

It seems independence with go1.15. In https://travis-ci.com/github/go-delve/delve/jobs/376630572, here is one of the `stacktrace`: 
```sh
    proc_test.go:3209: 	0x63d1c -0x480 -0x488 runtime.memclrNoHeapPointers at memclr_arm64.s:174
    proc_test.go:3209: 	0x1e344 -0x3f0 -0x4000187000 runtime.heapBits.initSpan at mbitmap.go:822
    proc_test.go:3209: 	0x205b8 -0x3a0 -0x4000187000 runtime.(*mcentral).grow at mcentral.go:515
    proc_test.go:3209: 	0x2026c -0x310 -0x4000187000 runtime.(*mcentral).cacheSpan at mcentral.go:177
    proc_test.go:3209: 	0x1fbcc -0x2e0 -0x4000187000 runtime.(*mcache).refill at mcache.go:142
    proc_test.go:3209: 	0x1968c -0x290 -0x4000187000 runtime.(*mcache).nextFree at malloc.go:880
    proc_test.go:3209: 	0x19ed4 -0x1d0 -0x4000187000 runtime.mallocgc at malloc.go:1061
    proc_test.go:3209: 	0x1a2e8 -0x1a0 -0x4000187000 runtime.newobject at malloc.go:1195
    proc_test.go:3209: 	0x3beac -0x120 -0x4000187000 runtime.acquireSudog at proc.go:345
    proc_test.go:3209: 	0x13ad8 -0xa0 -0x4000187000 runtime.chansend at chan.go:238
    proc_test.go:3209: 	0x139d0 -0x60 -0x4000187000 runtime.chansend1 at chan.go:143
    proc_test.go:1279: Goroutine 38: could not find correct frame
```
and here is the source code:
```go
		found := make([]bool, 10)
		for _, g := range gs {
			frame := -1
			frames, err := g.Stacktrace(10, 0)
                        ... ...
			for i := range frames {
				if frames[i].Call.Fn != nil && frames[i].Call.Fn.Name == "main.agoroutine" {
					frame = i
					break
				}
			}

			if frame < 0 {
				t.Logf("Goroutine %d: could not find correct frame", g.ID)
				continue
			}
```
This means `g.Stacktrace(10, 0)` is too less to print the stacktrace to find `main.agoroutine call`

hengwu0:
> ```
> https://travis-ci.com/github/go-delve/delve/jobs/373024405
> https://travis-ci.com/github/go-delve/delve/jobs/373024272
> https://travis-ci.com/github/go-delve/delve/jobs/373024766
> ```
> 
> @hengwu0

 I have checked fleetingly, and i think we can just ignore them here by rerun the test. They are occured rarely.
Thank you for collect these errors, and i think we should trace them in a new issue.

chainhelen:
> > ```
> > https://travis-ci.com/github/go-delve/delve/jobs/373024405
> > https://travis-ci.com/github/go-delve/delve/jobs/373024272
> > https://travis-ci.com/github/go-delve/delve/jobs/373024766
> > ```
> > 
> > 
> > @hengwu0
> 
> I have checked fleetingly, and i think we can just ignore them here by rerun the test. They are occured rarely.
> Thank you for collect these errors, and i think we should trace them in a new issue.

New issue #2151

service/api/prettyprint.go
aarzilli: But then columns aren't going to be aligned.

service/rpc2/server.go
```diff
@@ -783,6 +785,7 @@ func (s *RPCServer) ExamineMemory(arg ExamineMemoryIn, out *ExaminedMemoryOut) e
 	}
 
 	out.Mem = Mem
+	out.IsLittleEndian = isLittleEndian()
```
aarzilli, line 788: Change it to `out.IsLittleEndian = true` and add this comment:

```
//TODO: get byte order from debugger.target.BinInfo().Arch
```

service/api/prettyprint.go
hitzhangjie: Yes ... done

service/rpc2/server.go
hitzhangjie: done
-----------------------------------------------------------
Please let me know if I misunderstood something.

service/api/prettyprint.go
```diff
@@ -375,23 +381,27 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
+		colFormat = "0%03o"                            // Always keep one leading zero for octal.
+		colFormat = fmt.Sprintf("0%%0%do", colBytes*3) // Always keep one leading zero for octal.
 	case 'd':
 		cols = 8
 		colFormat = "%03d"
+		colFormat = fmt.Sprintf("%%0%dd", colBytes*3)
 	case 'x':
 		cols = 8
-		colFormat = "0x%02x" // Always keep one leading '0x' for hex.
+		colFormat = "0x%02x"                            // Always keep one leading '0x' for hex.
```
aarzilli, line 395: This line can be deleted, I believe it was accidentally left here.

hitzhangjie: done

aarzilli (approved):
LGTM

service/api/prettyprint.go
```diff
@@ -403,14 +409,34 @@ func PrettyExamineMemory(address uintptr, memArea []byte, format byte) string {
...
+			}
 		}
 		fmt.Fprintln(w, "")
 		address += uintptr(cols)
 	}
 	w.Flush()
 	return b.String()
 }
+
+func byteArrayToUInt64(buf []byte, isLittleEndian bool) uint64 {
```
derekparker, line 430: This can be replaced with `binary.<Endian>.Uint64`, no?

service/api/prettyprint_test.go
```diff
@@ -32,3 +33,28 @@ func TestPrettyExamineMemory(t *testing.T) {
...
 	}
 }
+
+func Test_byteArrayToUInt64(t *testing.T) {
+	tests := []struct {
+		name string
+		args []byte
+		want uint64
+	}{
+		// TODO: Add test cases.
```
derekparker, line 43: This can be removed now right?

service/rpc2/server.go
```diff
@@ -770,7 +770,8 @@ type ExamineMemoryIn struct {
 
 // ExaminedMemoryOut holds the return values of ExamineMemory
 type ExaminedMemoryOut struct {
-	Mem []byte
+	Mem            []byte
+	IsLittleEndian bool
```
derekparker, line 774: Maybe we can just store `binary.BigEndian` or `binary.LittleEndian` and just use the methods off of them when we need to convert, etc.

aarzilli: I don't think those can be serialized to json.

service/api/prettyprint.go
hitzhangjie: `binary.<Endian>.Uint64` only works if `len(buf)==8`.

service/api/prettyprint_test.go
hitzhangjie: done

service/rpc2/server.go
derekparker: Of course not... obvious oversight on my part.

derekparker (approved):
LGTM

Just needs rebase.

hitzhangjie:
> LGTM
> 
> Just needs rebase.

ok, done

Issue #2151, opened 2020-08-25
"could not find correct frame" error ci tests on arm64/linux occasionally


chainhelen:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
1.5.0 

2. What version of Go are you using? (`go version`)?
1.15

3. What operating system and processor architecture are you using?
CI on arm64/linux

4. What did you do? 

5. What did you expect to see?
No ci error

6. What did you see instead?

https://travis-ci.com/github/go-delve/delve/jobs/376630572
https://travis-ci.com/github/go-delve/delve/jobs/373024405
https://travis-ci.com/github/go-delve/delve/jobs/373024272
https://travis-ci.com/github/go-delve/delve/jobs/373024766

relate #2147

chainhelen:
hengwu0: 

It seems independence with go1.15. In https://travis-ci.com/github/go-delve/delve/jobs/376630572, here is one of the stacktrace:

    proc_test.go:3209: 	0x63d1c -0x480 -0x488 runtime.memclrNoHeapPointers at memclr_arm64.s:174
    proc_test.go:3209: 	0x1e344 -0x3f0 -0x4000187000 runtime.heapBits.initSpan at mbitmap.go:822
    proc_test.go:3209: 	0x205b8 -0x3a0 -0x4000187000 runtime.(*mcentral).grow at mcentral.go:515
    proc_test.go:3209: 	0x2026c -0x310 -0x4000187000 runtime.(*mcentral).cacheSpan at mcentral.go:177
    proc_test.go:3209: 	0x1fbcc -0x2e0 -0x4000187000 runtime.(*mcache).refill at mcache.go:142
    proc_test.go:3209: 	0x1968c -0x290 -0x4000187000 runtime.(*mcache).nextFree at malloc.go:880
    proc_test.go:3209: 	0x19ed4 -0x1d0 -0x4000187000 runtime.mallocgc at malloc.go:1061
    proc_test.go:3209: 	0x1a2e8 -0x1a0 -0x4000187000 runtime.newobject at malloc.go:1195
    proc_test.go:3209: 	0x3beac -0x120 -0x4000187000 runtime.acquireSudog at proc.go:345
    proc_test.go:3209: 	0x13ad8 -0xa0 -0x4000187000 runtime.chansend at chan.go:238
[...]

PR #2152, opened 2020-08-25
[verifying] Not found frame arm64 (don't merged directly)

chainhelen:
relate #2151

1. Enlarging depth into 30 from 10.
2. Add 100 loops for `TestFrameEvaluation`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ matrix:
 
 before_install:
   - export GOFLAGS=-mod=vendor
+  - export GODEBUG=asyncpreemptoff=1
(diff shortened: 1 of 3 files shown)


chainhelen:
Enlarging into 100.

hengwu0:
https://travis-ci.com/github/go-delve/delve/jobs/377316078
amd64 seems have the same problom?

chainhelen:
> https://travis-ci.com/github/go-delve/delve/jobs/377316078
> amd64 seems have the same problom?
> 
> ```shell
>     proc_test.go:1270: Goroutine 25 <nil>
>     proc_test.go:3211: 	0x426a40 -0xc0000947e0 -0xc000094800 runtime.gopark at proc.go:305
>     proc_test.go:3211: 	0x0 0x0 0x0 ? at .:0
>     proc_test.go:1280: Goroutine 25: could not find correct frame
> ```

Yes, but the frequency of arm64 is higher.

PR #2156, opened 2020-08-27
proc: fix dlv can't find the g in runtime.gopark

### LINKED ISSUE #2151: "could not find correct frame" error ci tests on arm64/linux occasionally

hengwu0:
fix #2151. I find the g can't be geted because of `th.Blocked()`

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -310,9 +310,6 @@ func GoroutinesInfo(dbp *Target, start, count int) ([]*G, int, error) {
 
 	threads := dbp.ThreadList()
 	for _, th := range threads {
-		if th.Blocked() {
-			continue
-		}
 		g, _ := GetG(th)
 		if g != nil {
 			threadg[g.ID] = g


pkg/proc/variables.go
```diff
@@ -311,7 +311,7 @@ func GoroutinesInfo(dbp *Target, start, count int) ([]*G, int, error) {
 	threads := dbp.ThreadList()
 	for _, th := range threads {
 		if th.Blocked() {
-			continue
+			//continue
```
aarzilli, line 314: This is on purpose. We either aren't sure that we can read memory from thread if Blocked is true or we don't want to because it could be wrong (runtime.clone). We can't do this.

hengwu0: Okey, I  should fix the `th.Blocked` actually.

hengwu0: > This is on purpose. We either aren't sure that we can read memory from thread if Blocked is true or we don't want to because it could be wrong (runtime.clone). We can't do this.

1. There is an state that the `th.Blocked()` returned true, and i checked than the fn.Name is "runtime.futex”. **But** the `g, _ := GetG(th)` still can get the right `g` and the `proc.G{Status: 0x02}` in `g` means it is running.
2. I checked `GetG()` and it already has the code to check whether it's in ` runtime.clone`: https://github.com/go-delve/delve/blob/788c41be761b9c4b84979f9e8b0d9416dd7bcfb3/pkg/proc/variables.go#L242

So, why there need to check the threads is Blocked? We can use `GetG` to check whether there is a `g` in thread directly.

aarzilli: I'm not sure why any of this matters.

Jim-wiselike: When the thread going to `runtime·futex`, there is no _p_ maybe, but it still have _g_ and _g_'s status is still running. I think it is save to get the _g_ maybe, otherwise we may lost some _g_ when listing gorutines at that instant.

hengwu0:
closed this in #2206

aarzilli:
Even though we didn't merge this PR, thank you for looking into this.

hengwu0:
> Even though we didn't merge this PR, thank you for looking into this.

I checked that the `th.Blocked()` is removed in #2206, so there is nothing to do in this PR. It should be closed.

PR #2158, opened 2020-09-03
service/dap: Use 'nil' as string value for interfaces with 0 addr

polinasok:
This is based on the pretty print logic for CLI. Testing appears too tricky to implement reliably.


aarzilli (approved):
lgtm

PR #2160, opened 2020-09-04
service/dap: Add support for package globals to scopes/variables requests

polinasok:
Updates #1515


service/dap/server.go
```diff
@@ -752,22 +752,22 @@ func (s *Server) onScopesRequest(request *dap.ScopesRequest) {
...
-		var globals []api.Variable
 		if state.CurrentThread != nil && state.CurrentThread.File != "" {
+			currPkg := "??"
 			currDir := path.Dir(filepath.ToSlash(state.CurrentThread.File)) // Normalized across OS
 			// Retrieve package-to-compiled-directory mapping, so we can match
 			// the directory of the current file with the breakpoint to the import
 			// path of its package. The import path can then be used to filter
 			// all global package variables returned by the debugger by default
 			// down to only the ones in the current package.
 			pkgInfo := s.debugger.ListPackagesBuildInfo( /*includeFiles*/ false)
```
polinasok, line 767: As per offline discussion with @aarzilli, I am looking into replacing this file-to-package logic with parsing the package name from the current function name. 

I have access to state.CurrentThread.Function, which is of type *api.Function. I believe it is converted from *proc.Function in api.ConvertFunction(). And proc.Function has PackageName. I can expand api.Function to include package name separately from the fully qualified Name (as well as cu.isgo) and copy those over in api.ConvertFunction().

aarzilli: I was thinking of just adding a CurrentPackage method to debugger.(*Debugger) actually.

service/dap/server.go
```diff
@@ -464,6 +469,9 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		s.args.stackTraceDepth = int(depth)
 	}
 
+	globals, ok := request.Arguments["showGlobalVariables"]
+	s.args.showGlobalVariables = ok && globals == true
```
aarzilli, line 473: you can replace this with just `s.args.showGlobalVariables, _ = request.Arguments["showGlobalVariables"].(bool)`

polinasok: Agree, that's better, more succinct. However, now that I am thinking about this, I should rewrite this to be conditional. I recently made a change to have the defaults in a struct that I set in the constructor, so I should no longer rely on these expressions producing the right defaults.

service/dap/server.go
polinasok: How about now?

aarzilli (approved):
LGTM

PR #2163, opened 2020-09-06
proc/*,service: replace uses of uintptr with uint64

aarzilli:
Since proc is supposed to work independently from the target
architecture it shouldn't use architecture-dependent types, like
uintptr. For example when reading a 64bit core file on a 32bit
architecture, uintptr will be 32bit but the addresses proc needs to
represent will be 64bit.


derekparker (approved):
LGTM

Issue #2162, opened 2020-09-06
dlv panic 


rmGFW:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
1.5.0
2. What version of Go are you using? (`go version`)?
go1.14 linux/amd64
3. What operating system and processor architecture are you using?
Tencent tlinux release 2.2 (Final)  in x86_64
4. What did you do?
use dlv debug command to debug a go program 
5. What did you expect to see?
i hope dlv can work
6. What did you see instead?
![image](https://user-images.githubusercontent.com/66858140/92329396-cdb3da00-f099-11ea-98a4-ce15c3bc9552.png)

chainhelen:
This is a bug.

rmGFW:
> This is a bug.

but dlv work on my mac well .

i install dlv by brew 

but on my linux pc . i install by go get

chainhelen:
Not your fault, don't worry, I will fix it.

aarzilli:
I think this is already fixed by f90134eb4db1c423e24fddfbc6eff41b288e6297, however stepping like that won't do anything useful: you should set a breakpoint on main and continue.

chainhelen:
> I think this is already fixed by [f90134e](https://github.com/go-delve/delve/commit/f90134eb4db1c423e24fddfbc6eff41b288e6297), however stepping like that won't do anything useful: you should set a breakpoint on main and continue.



@aarzilli I have try master branch (include f90134e), also panic. So I create one new pr to solve this issue.

aarzilli:
Are you sure about this? Because I can't reproduce it on master.

chainhelen:
@aarzilli Yes.  so strange?

OS:
```
// also happened go1.15.1 linux/amd64
go version go1.14.8 linux/amd64
```

dlv version:
```
// install by master
commit 7555d1c06339e51a4f1e2d7a62dc9c44745e10b7 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
cmd,proc,terminal,debugger: Support default file descriptor redirects

make install 
go install "-ldflags=-X main.Build=7555d1c06339e51a4f1e2d7a62dc9c44745e10b7" github.com/go-delve/delve/cmd/dlv
```


```
// cat main.go
package main

import "fmt"

func main() {
	fmt.Println("test dlv")
}


dlv debug main.go   
[...]

PR #2164, opened 2020-09-07
pkg/proc: fix dlv panic when sameGCond is nil.

chainhelen:
`sameFrameCond` should not be constructed as one `And Express` when
sameGCond which is the first child of `BinaryExpr` is nil.

Fixes: #2162


aarzilli (approved):
LGTM

PR #2170, opened 2020-09-10
proc: resolve symlinks when searching for split debug_info if path is /proc/pid/exe

### LINKED ISSUE #2168: Delve should not attempt to follow procfs exe symlink
1. What version of Delve are you using (`dlv version`)? 1.5.0 (build `7555d1c06339e51a4f1e2d7a62dc9c44745e10b7`)
2. What version of Go are you using? (`go version`)? go1.14.5
3. What operating system and processor architecture are you using? Linux on 64-bit x86.
4. What did you do? Attempt to attach to an existing process whose binary had been replaced via `sudo ./dlv attach 20447 ./gitaly-ssh`.
5. What did you expect to see? A successful attachment.
6. What did you see instead?

```
$ sudo ./dlv attach 20447 ./gitaly-ssh

could not attach to pid 20447: lstat /opt/gitlab/embedded/bin/gitaly-ssh (deleted): no such file or directory
```

Here's the deal: procfs is weird.

When you `stat /proc/<pid>/exe`, it looks like a symlink. Except when the inode of the binary changed since the process started, it appends ` (deleted)` which then is no longer a symlink you can actually follow.

Delve is trying to be nice in following the symlink.

Except `/proc/<pid>/exe` in fact _doesn't_ behave like a symlink when you `open` it. In that case it will give you a fd to the memory mapped (older) copy of the program.

I believe delve should not try to resolve the symlink and instead directly attempt to open the procfs exe here.

aarzilli:
### proc: resolve symlinks when searching for split debug_info if path is /proc/pid/exe

Fixes #2168

### Revert "proc: Find executable should follow symbol links."

This reverts commit 3e04ad0fada0c3ab57caf58bc024e4c0f9a3e01a.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -1037,6 +1037,11 @@ func (bi *BinaryInfo) openSeparateDebugInfo(image *Image, exe *elf.File, debugIn
 				continue
 			}
 			potentialDebugFilePath = fmt.Sprintf("%s/%s/%s.debug", dir, desc1, desc2)
+		} else if strings.HasPrefix(image.Path, "/proc") {
+			path, err := filepath.EvalSymlinks(image.Path)
+			if err == nil {
+				potentialDebugFilePath = fmt.Sprintf("%s/%s.debug", dir, filepath.Base(path))
+			}
 		} else {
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

PR #2172, opened 2020-09-12
proc/gdbserial: fix two protocol bugs

aarzilli:
During the testing of the core dump generation feature two bugs were
discovered in gdbserial:

1. we don't check that both bytes of the checksum are read, if the
buffer only has one byte we can end up reading only one byte instead
of two and the second byte will mess up the parsing of the next
packet
2. binary encoded packets can start with an 'E' and not be errors, when
using binary responses add an extra check for the lenght of the
response before deciding that the response is an error.

Unfortunately this encoding is inherently ambiguous (we can't
distinguish a 3 byte response starting with 'E' from an error) so
binary requests that lead to short responses should be avoided.

Testing this is complicated, they will be tested implicitly by the
upcoming core dump test.


derekparker (approved):
LGTM

PR #2173, opened 2020-09-12
terminal,service,proc/*: adds dump command (gcore equivalent)

aarzilli:
### terminal,service,proc/*: adds dump command (gcore equivalent)

Adds the `dump` command that creates a core file from the target process.

Backends will need to implement a new, optional, method `MemoryMap` that
returns a list of mapped memory regions.

Additionally the method `DumpProcessNotes` can be implemented to write out
to the core file notes describing the target process and its threads. If
DumpProcessNotes is not implemented `proc.Dump` will write a description of
the process and its threads in a OS/arch-independent format (that only Delve
understands).

Currently only linux/amd64 implements `DumpProcessNotes`.

Core files are only written in ELF, there is no minidump or macho-o writers.


derekparker:
@aarzilli please rebase and I will review. This week might be tougher with Gophercon going on but I'll try and get to it before too long.

pkg/proc/core/delve_core.go
```diff
@@ -0,0 +1,158 @@
...
+		return v[0], v[1], nil
+	}
+	panic("internal error")
+}
+
+func threadsFromDelveNotes(p *process, notes []*note) (proc.Thread, error) {
+	var currentThread proc.Thread
+	for _, note := range notes {
+		if note.Type == elfwriter.DelveHeaderNoteType {
+			const ()
```
derekparker, line 35: nit: remove unused const declaration.

pkg/proc/core/delve_core.go
```diff
@@ -0,0 +1,158 @@
...
+			for _, line := range strings.Split(string(note.Desc.([]byte)), "\n") {
```
derekparker, line 36: nit: This can be a bit cleaner / more efficient with a `bytes.Buffer` and then using the `buf.ReadString('\n')` method.

pkg/proc/proc_test.go
```diff
@@ -4979,3 +4982,166 @@ func TestStepOutPreservesGoroutine(t *testing.T) {
 		}
 	})
 }
+
+func TestDump(t *testing.T) {
+	if runtime.GOOS == "freebsd" || (runtime.GOOS == "darwin" && testBackend == "native") {
+		t.Skip("not supported")
+	}
+	if os.Getenv("CI") == "true" && buildMode == "pie" {
```
derekparker, line 4990: Does this need to be updated for TeamCity CI?

service/api/types.go
```diff
@@ -24,6 +24,8 @@ type DebuggerState struct {
 	// sending a StopRecording request will halt the recording, every other
 	// request will block until the process has been recorded.
 	Recording bool
+	// Dumping currently in progress.
+	Dumping bool
```
derekparker, line 28: nit: `CoreDumping` instead?

service/client.go
```diff
@@ -165,6 +165,13 @@ type Client interface {
 	// StopRecording stops a recording if one is in progress.
 	StopRecording() error
 
+	// DumpStart starts creating a core dump to the specified file
+	DumpStart(dest string) (api.DumpState, error)
```
derekparker, line 169: nit: Prefix with `Core`.

service/debugger/debugger.go
```diff
@@ -37,6 +37,12 @@ var (
 	// ErrNotRecording is returned when StopRecording is called while the
 	// debugger is not recording the target.
 	ErrNotRecording = errors.New("debugger is not recording")
+
+	// ErrDumpInProgress is returned when a core dump is already in progress.
+	ErrDumpInProgress = errors.New("core dump in progress")
```
derekparker, line 42: nit: For consistency prefer `CoreDump` over just `Dump`.

derekparker (changes requested):
A few minor things, looking forward to this feature!

pkg/proc/proc_test.go
aarzilli: I've updated the skip message but it's still occasionally failing. I'm leaning towards this being caused by a legitimate bug, but I can't reproduce it locally. Maybe it has something to do with docker.

aarzilli:
Everything except the test skip should be done.

derekparker (approved):
LGTM

aarzilli:
I can't make it fail anymore. Weird.

derekparker:
@aarzilli just needs rebase.

derekparker:
@aarzilli `TestDump` seems to be failing on CI.

aarzilli:
Yes, I'm looking into it, I can reproduce it now. I think I have figured it out.

PR #2179, opened 2020-09-22
proc/native: fix flakyness of TestStepConcurrentDirect on linux/386

aarzilli:
TestStepConcurrentDirect will occasionally fail (7% of the time on my
setup) by either causing the target processs to execute an invalid
instruction or (more infrequently) by switching to the wrong thread.

Both of those are caused by receiving SIGTRAPs for threads hitting a
breakpoint after it has been removed (the thread hits the breakpoint,
we stop everything and remove the breakpoint and only after we receive
the signal).

Change native.(*nativeProcess).stop to handle SIGTRAPs that can't be
attributed to a breakpoint, a hardcoded breakpoint in the program's
text, or manual stops (and therefore are likely caused by phantom
breakpoint hits).

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -9,13 +9,14 @@ import (
 type Arch struct {
 	Name string // architecture name
 
-	ptrSize               int
-	maxInstructionLength  int
-	prologues             []opcodeSeq
-	breakpointInstruction []byte
-	breakInstrMovesPC     bool
-	derefTLS              bool
-	usesLR                bool // architecture uses a link register, also called RA on some architectures
+	ptrSize                  int
+	maxInstructionLength     int
+	prologues                []opcodeSeq
+	breakpointInstruction    []byte
+	altBreakpointInstruction []byte
+	breakInstrMovesPC        bool
+	derefTLS                 bool
+	usesLR                   bool // architecture uses a link register, also called RA on some architectures
 
 	// asmDecode decodes the assembly instruction starting at mem[0:] into asmInst.
 	// It assumes that the Loc and AtPC fields of asmInst have already been filled.
@@ -64,6 +65,11 @@ func (a *Arch) MaxInstructionLength() int {
 // the given architecture.
 func (a *Arch) BreakpointInstruction() []byte {
 	return a.breakpointInstruction
+}
+
+// AltBreakpointInstruction returns an alternate encoding for the breakpoint instruction.
+func (a *Arch) AltBreakpointInstruction() []byte {
+	return a.altBreakpointInstruction
 }
 
 // BreakInstrMovesPC is true if hitting the breakpoint instruction advances the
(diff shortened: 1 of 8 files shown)


aarzilli:
Deepsource problem is for a pre-existing thing, Travis-CI getting stuck is a Travis-CI bug.

pkg/proc/arch.go
```diff
@@ -65,6 +66,11 @@ func (a *Arch) BreakpointInstruction() []byte {
 	return a.breakpointInstruction
 }
 
+// AltBreakpointInstruction returns an alternate encoding for the breakpoint instruction.
+func (a *Arch) AltBreakpointInstruction() []byte {
```
derekparker, line 70: Do you think it'd make sense to just change the original method to be `BreakpointInstructions` (plural) instead and just always return a slice?

derekparker (changes requested):

pkg/proc/native/proc_linux.go
```diff
@@ -524,14 +525,69 @@ func (dbp *nativeProcess) stop(trapthread *nativeThread) (err error) {
...
+				}
+			}
+		}
 	}
+
+	if switchTrapthread {
+		*ptrapthread = nil
+		for _, th := range dbp.threads {
+			if th.os.setbp && th.ThreadID() != trapthread.ThreadID() {
+				fmt.Printf("Trapthread switch successful\n")
```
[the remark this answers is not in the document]
aarzilli: Yes. Removed.

pkg/proc/arch.go
aarzilli: No strong opinions here. My reasoning was that BreakpointInstruction was the one we use and the other would be breakpoint instructions we might encounter.

aarzilli: But changing to a slice is fine too.

derekparker: Ack, that makes sense to me.

derekparker (approved):

pkg/proc/native/proc_darwin.go
```diff
@@ -422,7 +422,8 @@ func (dbp *nativeProcess) resume() error {
 }
 
 // stop stops all running threads and sets breakpoints
-func (dbp *nativeProcess) stop(trapthread *nativeThread) (err error) {
+func (dbp *nativeProcess) stop(ptrapthread **nativeThread) (err error) {
```
derekparker, line 425: I'm not all that excited about the pointer-to-pointer change.

I think from a clarity of intent perspective it's cleaner to keep it as it, but instead return `(*nativeThread, err)` and let the caller do something like `trapthread = dbp.stop(trapthread)`.

Thoughts?

derekparker (changes requested):

pkg/proc/native/proc_darwin.go
aarzilli: Done.

derekparker (approved):
LGTM

PR #2180, opened 2020-09-22
dwarf/line: ignore end_of_sequence entry if AllPCsBetween

aarzilli:
Go 1.15 (but possibly prior versions of Go too) has a tendency to use
an address in the middle of an instruction for this entry, but if it
was correct it would be after the last instruction of the function
anyway.

This problem manifests especially frequently as a target crash in
TestStepConcurrentPtr on linux/arm64 (~6% of the runs).


derekparker (approved):
LGTM

PR #2182, opened 2020-09-23
service/*: remove threadID argument of (*Debugger).PackageVariables

aarzilli:
Which thread is used doesn't make any difference to the list of package
variables that is returned and this option was only ever used by an old
v1 API call.


derekparker (approved):
LGTM

PR #2185, opened 2020-09-28
service/dap: support evaluate requests with expressions and calls

polinasok:
updates #1515


service/dap/server.go
```diff
@@ -987,16 +1004,81 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
...
+		frame = sf.(stackFrame).frameIndex
+	}
+	// TODO(polina): Support config settings via launch/attach args vs auto-loading?
+	apiCfg := &api.LoadConfig{FollowPointers: true, MaxVariableRecurse: 1, MaxStringLen: 64, MaxArrayValues: 64, MaxStructFields: -1}
+	prcCfg := proc.LoadConfig{FollowPointers: true, MaxVariableRecurse: 1, MaxStringLen: 64, MaxArrayValues: 64, MaxStructFields: -1}
+
+	response := &dap.EvaluateResponse{Response: *newResponse(request.Request)}
+	isCall, err := regexp.MatchString(`^\s*call\s+\S+`, request.Arguments.Expression)
+	if err == nil && isCall { // call {expression}
+		state, err := s.debugger.Command(&api.DebuggerCommand{
```
aarzilli, line 1041: I think we should complain if `frame != 0`.

service/dap/server.go
```diff
@@ -987,16 +1004,81 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
...
+		if _, isexited := err.(proc.ErrProcessExited); isexited || err == nil && state.Exited {
+			e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
+			s.send(e)
+			return
+		}
+		if err != nil {
+			s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
+			return
+		}
+		retVars := s.debugger.ReturnValues(&prcCfg)
```
aarzilli, line 1057: The problem with this is that there's no guarantee that the return values for the call will be available the very next time the debugger stops. It could be that the debugger stopped because it encountered a breakpoint, either inside the function that was called or some other breakpoint in a different goroutine.

I'm not sure what the right way to handle this would be, I think it depends on what the UI of VSCode can do. For example what happens when using VSCode with gdb, doing a function call while a breakpoint is set in the function being called?

service/dap/server.go
```diff
@@ -987,16 +1004,81 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
...
+		if retVars != nil && len(retVars) > 0 {
```
aarzilli, line 1058: See DeepSource error.

service/dap/server_test.go
```diff
@@ -1132,6 +1150,239 @@ func TestLaunchRequestWithStackTraceDepth(t *testing.T) {
...
+						t.Errorf("\ngot %#v\nwant Format=\"Unable to evaluate expression: could not find symbol value for a1\"", erres)
+					}
+				},
+				disconnect: false,
+			}})
+	})
+}
+
+func TestEvaluateCallRequest(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
```
aarzilli, line 1303: you should call proc/test.MustSupportFunctionCalls here.

service/dap/server.go
polinasok: I added a draft implementation. See TODO please - I am not fully sure what the cleanest way is to detect if the call completed or not.

Also, I added a unittest for a breakpoint inside of the call. I still need to add a test where the call restarts other goroutines that stop on a breakpoint. In the meantime, I tested this manually with https://github.com/aarzilli/delve_client_testing/blob/master/05nextbp/main.go. It works when one-off calls are issued from the debug console. But if we put a call into a watch panel, then everything gets totally screwed. The call restarts everything, so before it finishes, another goroutine stops. That triggers watch re-eval, but relative to another goroutine, so it is allowed. This again restarts everything and just keeps going like continue, never really stopping at any breakpoints for longer than a split second. Should we be disallowing calls as watch expressions?   That would be  unfortunate because it could be such a neat way to pretty print most interesting variables.

service/dap/server_test.go
polinasok: Done

service/dap/server.go
aarzilli: > See TODO please - I am not fully sure what the cleanest way is to detect if the call completed or not.

You should iterate over the thread list and check if one has the goroutine ID you want and `ReturnValues != nil`. I don't think the way you are doing this works. First off there's no guarantee that the injected goroutine will make progress before we are interrupted by a simultaneous breakpoint, secondly consider what happens if the goroutine stopped on a breakpoint inside `main.f` and the user injects a call to `main.g` which calls `main.f` again.

> Should we be disallowing calls as watch expressions to avoid such circular behavior? That would be unfortunate because it could be such a neat way to pretty print most interesting variables in the watch panel and have them automatically update.

IMHO there's no way to do it cleanly. I'd like to know what VSCode's solution is for C-like languages, maybe there's something I haven't thought of.

service/dap/server_test.go
aarzilli: I meant instead of checking `runtime.GOARCH == "arm64"`.

aarzilli:
PS. see the failure of TestScopesAndVariables on windows/go1.15. I think you need to relax that test, it keeps going back and forth.

aarzilli:
I think travis is running the tests because of the merge conflict. If you don't know how to get the full log of cirrus-ci, it's here: https://cirrus-ci.com/task/6100324710612992.

polinasok:
It took me a moment to figure out, but I did eventually get to the log by clicking some obscure link on the bottom. Let's see what things look like after the merge.

polinasok:
I think the proc test failures are unrelated.

service/dap/server.go
```diff
@@ -1032,16 +1051,131 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
...
+				continue
+			}
+			if t.Line == stateBeforeCall.SelectedGoroutine.CurrentLoc.Line && t.ReturnValues != nil {
+				// The call completed. Get the return values.
+				proct, _ := s.debugger.FindThread(t.ID)
+				if err != nil {
+					s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
+					return
+				}
+				retVars = proct.Common().ReturnValues(prcCfg)
```
aarzilli, line 1135: There should be a s.debugger.LockTarget() / s.debugger.UnlockTarget() around this. Also at this point we can exit the loop early with `break`.

service/debugger/debugger.go
```diff
@@ -1518,6 +1518,16 @@ func (d *Debugger) Recorded() (recorded bool, tracedir string) {
 	return d.target.Recorded()
 }
 
+// ReturnValues returns the return values of the function we just stepped out of.
```
aarzilli, line 1521: No longer used?

service/dap/server.go
```diff
@@ -1032,16 +1051,131 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
...
+		// After the call is done, the goroutine where we injected the call should
+		// return to the original stopped line with return values. However,
+		// it is not guaranteed to be selected due to the possibility of the
+		// of simultaenous breakpoints. Therefore, we check all threads.
+		var retVars []*proc.Variable
+		for _, t := range state.Threads {
+			if t.GoroutineID != stateBeforeCall.SelectedGoroutine.ID {
+				continue
+			}
+			if t.Line == stateBeforeCall.SelectedGoroutine.CurrentLoc.Line && t.ReturnValues != nil {
```
aarzilli, line 1128: This if and the one above can be fused into a single one 
```
if t.GoroutineID == stateBeforeCall.SelectedGoroutine.ID && t.Line == stateBeforeCall.SelectedGoroutine.CurrentLoc.Line && t.ReturnValues != nil { ...```

service/dap/server.go

service/debugger/debugger.go
polinasok: Reused for ReturnValues of a given thread.

aarzilli (approved):
LGTM

PR #2186, opened 2020-09-28
service/dap: add "panic" and "fatal error" as stopped reasons

suzmue:
The unrecovered panic and fatal throw breakpoints are not set by the
user. We now check for these special breakpoints and send appropriate
stopped reasons to the client.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1173,13 +1173,13 @@ func (s *Server) doCommand(command string) {
 			stopped.Body.Reason = "step"
 		default:
 			stopped.Body.Reason = "breakpoint"
-			if state.CurrentThread.Breakpoint != nil {
-				switch state.CurrentThread.Breakpoint.Name {
-				case proc.FatalThrow:
-					stopped.Body.Reason = "fatal error"
-				case proc.UnrecoveredPanic:
-					stopped.Body.Reason = "panic"
-				}
+		}
+		if state.CurrentThread.Breakpoint != nil {
+			switch state.CurrentThread.Breakpoint.Name {
+			case proc.FatalThrow:
+				stopped.Body.Reason = "fatal error"
+			case proc.UnrecoveredPanic:
+				stopped.Body.Reason = "panic"
 			}
 		}
 		s.send(stopped)
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1150,12 +1150,35 @@ func (s *Server) doCommand(command string) {
 	stopped.Body.AllThreadsStopped = true
 
 	if err == nil {
-		stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		if state.SelectedGoroutine != nil {
+			stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		} else {
+			// If there is no selectedGoroutine, get the list of goroutines and select the first one.
+			// TODO(polina): validate the assumption in this code that the first goroutine
+			// is the current one. So far it appears to me that this is always the main goroutine
+			// with id 1.
+			gs, _, err := s.debugger.Goroutines(0, 1)
+			if err != nil {
+				s.log.Error(err)
+			}
+			if len(gs) > 0 {
+				stopped.Body.ThreadId = gs[0].ID
+			}
+		}
+
 		switch command {
 		case api.Next, api.Step, api.StepOut:
 			stopped.Body.Reason = "step"
 		default:
 			stopped.Body.Reason = "breakpoint"
+			if state.CurrentThread.Breakpoint != nil {
+				switch state.CurrentThread.Breakpoint.Name {
+				case proc.FatalThrow:
+					stopped.Body.Reason = "fatal error"
+				case proc.UnrecoveredPanic:
+					stopped.Body.Reason = "panic"
+				}
+			}
 		}
 		s.send(stopped)
 	} else {
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -962,7 +962,21 @@ func (s *Server) convertVariable(v *proc.Variable) (value string, variablesRefer
 		} else if len(v.Children) == 0 || v.Children[0].Kind == reflect.Invalid && v.Children[0].Addr == 0 {
 			value = "nil <" + typeName + ">"
 		} else {
-			value = "<" + typeName + ">"
+			value = "<" + typeName + "(" + v.Children[0].TypeString() + ")" + ">"
+			// TODO(polina): should we remove one level of indirection and skip "data"?
+			// Then we will have:
+			// Before:
+			//   i: <interface{}(int)>
+			//      data: 123
+			// After:
+			//   i: <interface{}(int)> 123
+			// Before:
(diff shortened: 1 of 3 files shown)


suzmue:
@polinasok

service/dap/server.go
```diff
@@ -1150,12 +1150,35 @@ func (s *Server) doCommand(command string) {
 	stopped.Body.AllThreadsStopped = true
 
 	if err == nil {
-		stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		if state.SelectedGoroutine != nil {
+			stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		} else {
+			// If there is no selectedGoroutine, get the list of goroutines and select the first one.
```
aarzilli, line 1156: This isn't the correct behavior, if there is no selectedGoroutine that means that the current thread is not executing a goroutine. It would be better to set ThreadId = 0 in that case.

service/dap/server.go
```diff
@@ -1150,12 +1150,35 @@ func (s *Server) doCommand(command string) {
...
+			gs, _, err := s.debugger.Goroutines(0, 1)
+			if err != nil {
+				s.log.Error(err)
+			}
+			if len(gs) > 0 {
+				stopped.Body.ThreadId = gs[0].ID
+			}
+		}
+
 		switch command {
```
aarzilli, line 1169: After this block of code was written we added a StopReason field to the Target object, this could be a good time to use that instead of looking at the command field.

service/dap/server.go
suzmue: Sounds good. The ThreadId for the stopped event is optional in the DAP spec, so it probably makes the most sense to not set the ThreadId in this case.

service/dap/server.go
suzmue: Is there a way that StopReason is exposed? Debugger.target is not exported and the info is not available through the DebugState

service/dap/server.go
```diff
@@ -1150,12 +1150,35 @@ func (s *Server) doCommand(command string) {
 	stopped.Body.AllThreadsStopped = true
 
 	if err == nil {
-		stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		if state.SelectedGoroutine != nil {
```
polinasok, line 1153: Good catch, Suzie. We should definitely be checking for no selected goroutine!

service/dap/server.go
```diff
@@ -1150,12 +1150,35 @@ func (s *Server) doCommand(command string) {
...
+				stopped.Body.ThreadId = gs[0].ID
+			}
+		}
+
 		switch command {
 		case api.Next, api.Step, api.StepOut:
 			stopped.Body.Reason = "step"
 		default:
 			stopped.Body.Reason = "breakpoint"
+			if state.CurrentThread.Breakpoint != nil {
```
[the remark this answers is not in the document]
suzmue: Good point! Using StopReason seems like it'll help make the logic a bit cleaner. I'll definitely add a test for that.

service/dap/server.go
suzmue: Yeah this seems like a UI issue for the editors that are using the results of this call.

service/dap/server.go
suzmue: I added it to this PR, if I should separate this out I am happy to do so!

service/dap/server_test.go
```diff
@@ -1254,6 +1254,32 @@ func TestPanicBreakpoint(t *testing.T) {
...
+			fixture.Source, []int{5},
+			[]onBreakpoint{{
+				execute: func() {
+					handleStop(t, client, 1, 5)
+
+					client.NextRequest(1)
+					client.ExpectNextResponse(t)
+
+					se := client.ExpectStoppedEvent(t)
+					if se.Body.ThreadId != 1 || se.Body.Reason != "panic" {
```
polinasok, line 1276: Why is thread id not 1?

service/dap/server.go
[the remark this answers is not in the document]
polinasok: I thought the comment for StopReason said it could be any one reason when there are multiple ones (so could be breakpoint or could be next here, no?) Should the panic/fatal check be outside of the switch to make sure it supplies panic/fatal reason regardless of StopReason?

service/dap/server.go
polinasok: It is good as-is. For some reason, I imagined a more complicated refactor, channeling more reasons through through, but we could always do that later. Makes sense to keep the status quo in this change.

service/dap/server.go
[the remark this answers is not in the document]
aarzilli: @polinasok yes, that's probably better. Also I suspect that this is the cause of the test failure on Go 1.13 that Travis-CI is reporting.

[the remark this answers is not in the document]
suzmue: I tried moving the check outside of the switch, but when running with Go 1.13.15, none of the goroutines in the debug state are stopped on a breakpoint when it hits the panic during next.

This is because the breakpoint that is hit is an internal NextDeferBreakpoint, which is cleared before we get to that part of the code. This breakpoint gets set in 1.13 because the "_defer" var is actually pointing to an address in the runtime package (defer at src/runtime/proc.go:157) as opposed to in 1.14 and 1.15 where _defer points to 0. The solution may be to run a different test here (or a separate test for 1.13?), since the StopReasons do appear to be correct.

I can definitely move the checks for these breakpoints outside of the switch to be safe, but it would not fix this breaking test.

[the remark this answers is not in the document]
aarzilli: Makes sense. I'd move the check outside and skip the test on 1.13.

aarzilli (approved):
LGTM

service/dap/server_test.go
```diff
@@ -1269,12 +1269,31 @@ func TestPanicBreakpointOnNext(t *testing.T) {
...
-					client.NextRequest(1)
-					client.ExpectNextResponse(t)
-
-					se := client.ExpectStoppedEvent(t)
-					if se.Body.ThreadId != 1 || se.Body.Reason != "panic" {
-						t.Errorf("\ngot  %#v\nwant ThreadId=1 Reason=\"panic\"", se)
+					// Send a next request until the panic breakpoint is hit.
+					// We expect to hit the unrecovered panic breakpoint and
+					// return a "panic" stopped event.
+					// In Go 1.13, 'next' will step into the defer in the runtime
```
polinasok, line 1275: I would suggest being explicit in the test and having a separate block for Go 1.13 vs after. And maybe even skip 1.13 completely in your test. With the next Go version, 1.13 will probably be removed from the build set-up, but the complicated test will stay forever. I think you can use goversion.VersionAfterOrEqual to detect the version.

service/dap/server_test.go
polinasok: Please ignore this comment. I don't know what I was thinking :)

service/dap/server_test.go
suzmue: Done!

polinasok:
Thank you for updating the test.
LGTM

PR #2188, opened 2020-10-01
service/dap: support clearing breakpoints and setting breakpoint conditions

polinasok:
Updates #1515

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -380,6 +380,7 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 	// TODO(polina): Respond with an error if debug session is in progress?
 	response := &dap.InitializeResponse{Response: *newResponse(request.Request)}
 	response.Body.SupportsConfigurationDoneRequest = true
+	response.Body.SupportsConditionalBreakpoints = true
 	// TODO(polina): support this to match vscode-go functionality
 	response.Body.SupportsSetVariable = false
 	// TODO(polina): support these requests in addition to vscode-go feature parity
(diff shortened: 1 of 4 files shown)


service/dap/server.go
```diff
@@ -546,9 +582,9 @@ func (s *Server) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
 	i := 0
 	for _, b := range request.Arguments.Breakpoints {
 		bp, err := s.debugger.CreateBreakpoint(
-			&api.Breakpoint{File: request.Arguments.Source.Path, Line: b.Line})
+			&api.Breakpoint{File: request.Arguments.Source.Path, Line: b.Line, Cond: b.Condition})
 		if err != nil {
-			s.log.Error("ERROR:", err)
+			s.log.Error("ERROR: ", err)
```
aarzilli, line 587: I took a look at the SetBreakpointsRequest specification and it seems to me that when there is an error the adapter can return `verified == false` and set the `message` field to explain why the breakpoint couldn't be verified, wouldn't that be better?

I'm approving this, this is just something to consider.

aarzilli (approved):

service/dap/server.go

service/dap/server.go
```diff
@@ -535,9 +536,44 @@ func (s *Server) onDisconnectRequest(request *dap.DisconnectRequest) {
...
-		s.log.Error("ERROR: Unable to set breakpoint for empty file path")
+		s.sendErrorResponse(request.Request, UnableToSetBreakpoints, "Unable to set or clear breakpoints", "empty file path")
+		return
+	}
+
+	// According to the spec we should "set multiple breakpoints for a single source
+	// and clear all previous breakpoints in that source." The simplest way is
+	// to clear all and then set all.
+	//
+	// TODO(polina): should we optimize this as follows?
```
polinasok, line 550: Do you have an opinion here? (This was a user request).

aarzilli: I'm skeptical that the performance difference would be visible. Maybe when the adapter was javascript talking to delve using our API over laggy network.

PR #2192, opened 2020-10-11
proc: Remove (*Arch).Prologues method

aarzilli:
It returns an unexported type and is only used in a single place.


aarzilli:
DeepSource is always complaining about this method.

derekparker (approved):
LGTM

PR #2193, opened 2020-10-11
proc/test: fix TestStepCallPtr on linux/386

aarzilli:
The test needs to set a breakpoint on main.CallFn after the prologue,
on linux/386 this function does not have any instruction after the
prologue on the function header line  because it doesn't need to
allocate space for local variables. Change the fixture so that this
isn't a problem.

This bug results on the test failing a small percentage of the time.


derekparker (approved):
LGTM

PR #2194, opened 2020-10-11
proc/test: enable TestCgoStacktrace2

aarzilli:
Due to a missing check TestCgoStacktrace2 didn't actually check
anything. Enable it and then skip it on linux/386 where it's broken.


derekparker (approved):
LGTM

PR #2195, opened 2020-10-11
proc/native/linux: get exit status if proc exits after receiving signal

aarzilli:
If the process receives a signal (or sends a singal to itself) and then
dies before we can route the signal back to it we still need to
retrieve its exit status.

Fixes a rare failure of TestIssue1101 in proc_test.go


derekparker (approved):
LGTM

PR #2199, opened 2020-10-13
Add display of floating-point registers under ARM64 

### LINKED ISSUE #2198: Incorrect display of floating-point registers under ARM64

Delve Debugger
Version: 1.5.0

go version go1.14.6 linux/amd64

linux/arm64 

4. What did you do?
Use dlv to debug the program, enter the “regs -a“ command

Display all 128-bit Q registers, 64-bit D registers, 32-bit S registers, 16-bit H registers, and 8-bit B registers in all signed and unsigned hexadecimal and decimal systems
图片1

6. What did you see instead?
Wrong floating-point registers is shown
图片2

dujinze:
fix #2198  I added the display function for floating-point registers under ARM64. 
Part of the current display effect is:

图片1
图片3
图片4
图片5
图片6
图片7

This shows all 128-bit Q registers, 64-bit D registers, 32-bit S registers, 16-bit H registers, and 8-bit B registers in hexadecimal, decimal signed and unsigned modes.

### DIFF
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -34,16 +34,6 @@ type Arch struct {
 	addrAndStackRegsToDwarfRegisters func(uint64, uint64, uint64, uint64, uint64) op.DwarfRegisters
 	// DwarfRegisterToString returns the name and value representation of the given register.
 	DwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
-	//QDwarfRegisterToString returns the Q registers
-	QDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
-	//DDwarfRegisterToString returns the D registers
-	DDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
-	//SDwarfRegisterToString returns the S registers
-	SDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
-	//HDwarfRegisterToString returns the H registers
-	HDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
-	//BDwarfRegisterToString returns the B registers
-	BDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
 	// inhibitStepInto returns whether StepBreakpoint can be set at pc.
 	inhibitStepInto func(bi *BinaryInfo, pc uint64) bool
 
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -34,6 +34,16 @@ type Arch struct {
 	addrAndStackRegsToDwarfRegisters func(uint64, uint64, uint64, uint64, uint64) op.DwarfRegisters
 	// DwarfRegisterToString returns the name and value representation of the given register.
 	DwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
+	//QDwarfRegisterToString returns the Q registers
+	QDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
+	//DDwarfRegisterToString returns the D registers
+	DDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
+	//SDwarfRegisterToString returns the S registers
+	SDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
+	//HDwarfRegisterToString returns the H registers
+	HDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
+	//BDwarfRegisterToString returns the B registers
+	BDwarfRegisterToString func(int, *op.DwarfRegister) (string, bool, string)
 	// inhibitStepInto returns whether StepBreakpoint can be set at pc.
 	inhibitStepInto func(bi *BinaryInfo, pc uint64) bool
 
--- a/pkg/proc/arm64_arch.go
+++ b/pkg/proc/arm64_arch.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"math/big"
 
 	"github.com/go-delve/delve/pkg/dwarf/frame"
 	"github.com/go-delve/delve/pkg/dwarf/op"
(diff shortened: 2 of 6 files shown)


aarzilli (commented):
This change breaks linux/386 support. It is not appropriate to add a bunch of function pointers to proc.Arch that are only used by arm64. In general this seems way more invasive than it should be and basically all these changes should be localized inside arm64DwarfRegisterToString.

aarzilli:
Why is reporting Q0/D0/S0/H0/B0 needed when their values have already been reported as part of displaying V0?

dujinze:
> This change breaks linux/386 support. It is not appropriate to add a bunch of function pointers to proc.Arch that are only used by arm64. In general this seems way more invasive than it should be and basically all these changes should be localized inside arm64DwarfRegisterToString.

Okay, I didn't think about it at the time, it has been modified now

dujinze:
> Why is reporting Q0/D0/S0/H0/B0 needed when their values have already been reported as part of displaying V0?

Firstly, these are several display methods with different focuses. There are also several ways to display floating-point registers in gdb.
Secondly, the D, S, H, B, and Q in the V register do not provide unsigned and signed display modes

aarzilli:
> > Why is reporting Q0/D0/S0/H0/B0 needed when their values have already been reported as part of displaying V0?
> 
> Firstly, these are several display methods with different focuses. There are also several ways to display floating-point registers in gdb.
> Secondly, the D, S, H, B, and Q in the V register do not provide unsigned and signed display modes

This also applies to amd64, we don't do it there we shouldn't do it on arm64. If we started doing it it would be in a way that applied to both with a similar interface, not with special arm64-only functions in proc.Arch.

dujinze:
> > > Why is reporting Q0/D0/S0/H0/B0 needed when their values have already been reported as part of displaying V0?
> > 
> > 
> > Firstly, these are several display methods with different focuses. There are also several ways to display floating-point registers in gdb.
> > Secondly, the D, S, H, B, and Q in the V register do not provide unsigned and signed display modes
> 
> This also applies to amd64, we don't do it there we shouldn't do it on arm64. If we started doing it, it would be in a way that applied to both with a similar interface, not with special arm64-only functions in proc.Arch.

Hey aarzilli, do you mean that this display mode should also be provided under amd64? But the registers display under amd64 is correct and does not need to be changed. Besides, The registers in amd64 are very different from the registers in arm64.

aarzilli:
> Hey aarzilli, do you mean that this display mode should also be provided under amd64?

No, I mean it shouldn't be done on arm64. When we decide that we want to display the same register in multiple ways we should add a way to do it that involves a general, architecture independent, interface in Arch.

dujinze:
> No, I mean it shouldn't be done on arm64. When we decide that we want to display the same register in multiple ways we should add a way to do it that involves a general, architecture independent, interface in Arch.

Hello aarzilli，the floating-point registers under different architectures are different, and the display modes are also different. For specific situations, we need to refer to their chip manuals. Therefore, I think the register display function under ARM64 may not be placed in an architecture independent interface.

aarzilli (changes requested):
> Therefore, I think the register display function under ARM64 may not be placed in an architecture independent interface.

That's not acceptable as far as I am concerned.

dujinze:
> That's not acceptable as far as I am concerned.
Hi aarzilli，the display function of this floating-point register is necessary under arm64, because the user needs to judge the correctness of the floating-point operation based on the corresponding display result. Now that I have implemented this function, in your opinion, where should I put this function in the source code?

aarzilli:
> > That's not acceptable as far as I am concerned.
> 
> Hi aarzilli，the display function of this floating-point register is necessary under arm64, because the user needs to judge the correctness of the floating-point operation based on the corresponding display result. Now that I have implemented this function, in your opinion, where should I put this function in the source code?

I think your changes to arm64DwarfRegisterToString can be worked in.

dujinze:
> I think your changes to arm64DwarfRegisterToString can be worked in.

Most of my changes are made in arm64DwarfRegisterToString.Some changes are made in api.ConvertRegisters, because 6 sets of registers need to be appended to ([ ]Register type) out under arm64,this is not the same as amd64.

dujinze:
> I think your changes to arm64DwarfRegisterToString can be worked in.

Can you tell me which part of my code does not meet the requirements?

aarzilli:
> > I think your changes to arm64DwarfRegisterToString can be worked in.
> 
> Can you tell me which part of my code does not meet the requirements?

You shouldn't return more than one result per register in the register list.

dujinze:
> You shouldn't return more than one result per register in the register list.

Do you mean that I only get the register once in arm64DwarfRegisterToString, and then Fprintf to out in a different way?

aarzilli:
> > You shouldn't return more than one result per register in the register list.
> 
> Do you mean that I only get the register once in arm64DwarfRegisterToString, and then Fprintf to out in a different way?

Yes.

dujinze:
> Yes.

Ok, I see
Is there anything else unreasonable besides this?

aarzilli:
> > Yes.
> 
> Ok, I see
> Is there anything else unreasonable besides this?

Using a global variable as a way to pass arguments to a function.

aarzilli:
Also doing different things based on the value of Arch.Name, we do it in a couple of places but it shouldn't be proliferated.

dujinze:
> Using a global variable as a way to pass arguments to a function.

Got it.

dujinze:
> Also doing different things based on the value of Arch.Name, we do it in a couple of places but it shouldn't be proliferated.

But I can’t think of other ways to distinguish the registers display under different architectures.Do you have any good solutions?

derekparker:
This has been lingering for a while, @aarzilli you did the initial review of this patch. What's the status?

aarzilli:
The changes to arm64DwarfRegistersToString are fine, everything else shouldn't be merged. I've made a reduced version of this PR in #2289. The criticism that it doesn't display every possible interpretation of the registers is valid but it applies in general, on amd64 we aren't displaying the value of the AH part of RAX either, for example. I don't think that the solution is to make the output or regs even more bloated.

derekparker:
Closing this for now. Please reopen if this should be considered for merging.

PR #2202, opened 2020-10-15
proc/*: add launch option to disable ASLR

### LINKED ISSUE #1847: close "ALSR" to make debug convenient
I checked and find delve Do Not Close Address Space Layout Randomization(ALSR) when debugging. It made us hard to locate the problems when randomization of  debuggee's virtual address space is on. So, i think it needs to be enhanced(like GDB already seted ASLR closed by default).

aarzilli:
Fixes #1847


hengwu0:
Thanks for this function!

derekparker (approved):
LGTM

PR #2206, opened 2020-10-20
proc/*: remove proc.Thread.Blocked, refactor memory access

### LINKED ISSUE #2151: "could not find correct frame" error ci tests on arm64/linux occasionally

aarzilli:
### proc/*: remove proc.Thread.Blocked, refactor memory access

On linux we can not read memory if the thread we use to do it is
occupied doing certain system calls. The exact conditions when this
happens have never been clear.

This problem was worked around by using the Blocked method which
recognized the most common circumstances where this would happen.

However this is a hack: Blocked returning true doesn't mean that the
problem will manifest and Blocked returning false doesn't necessarily
mean the problem will not manifest. A side effect of this is issue
#2151 where sometimes we can't read the memory of a thread and find its
associated goroutine.

This commit fixes this problem by always reading memory using a thread
we know to be good for this, specifically the one returned by
ContinueOnce. In particular the changes are as follows:

1. Remove (ProcessInternal).CurrentThread and
(ProcessInternal).SetCurrentThread, the "current thread" becomes a
field of Target, CurrentThread becomes a (*Target) method and
(*Target).SwitchThread basically just sets a field Target.

2. The backends keep track of their own internal idea of what the
current thread is, to use it to read memory, this is the thread they
return from ContinueOnce as trapthread

3. The current thread in the backend and the current thread in Target
only ever get synchronized in two places: when the backend creates a
Target object the currentThread field of Target is initialized with the
backend's current thread and when (*Target).Restart gets called (when a
recording is rewound the currentThread used by Target might not exist
anymore).

4. We remove the MemoryReadWriter interface embedded in Thread and
instead add a Memory method to Process that returns a MemoryReadWriter.

The  backends will return something here that will read memory using
the current thread saved by the backend.

5. The Thread.Blocked method is removed

One possible problem with this change is processes that have threads
with different memory maps. As far as I can determine this could happen
on old versions of linux but this option was removed in linux 2.5.

Fixes #2151


derekparker:
Overall LGTM, this approach is a lot cleaner.

Needs rebase.

derekparker:
@aarzilli this just needs rebase and can be merged.

aarzilli:
It conflicts with #2179, since that looks close to reviewed I was waiting for it to be merged.

derekparker:
> 
> 
> It conflicts with #2179, since that looks close to reviewed I was waiting for it to be merged.

Ack, sorry for holding this up on the other review. That's approved and merged now.

aarzilli:
Rebased

derekparker (approved):
LGTM

PR #2208, opened 2020-10-20
Adds toggle command

### LINKED ISSUE #2008: Feature: support  breakpoint disable/enable

alexsaezm:
These changes add the 'toggle' command, which allows us to enable or disable a breakpoint.

Fixes #2008

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -299,8 +299,8 @@ func (t *Target) SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr) 
 	return newBreakpoint, nil
 }
 
-// setBreakpointWithID creates a breakpoint at addr, with the specified logical ID.
-func (t *Target) setBreakpointWithID(id int, addr uint64) (*Breakpoint, error) {
+// SetBreakpointWithID creates a breakpoint at addr, with the specified logical ID.
+func (t *Target) SetBreakpointWithID(id int, addr uint64) (*Breakpoint, error) {
 	bpmap := t.Breakpoints()
 	bp, err := t.SetBreakpoint(addr, UserBreakpoint, nil)
 	if err == nil {
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -200,6 +200,9 @@ Current limitations:
 	clearall [<linespec>]
 
 If called with the linespec argument it will delete all the breakpoints matching the linespec. If linespec is omitted all breakpoints are deleted.`},
+		{aliases: []string{"toggle"}, group: breakCmds, cmdFn: toggle, helpMsg: `Toggles on or off a breakpoint.
+
+toggle <breakpoint name or id>`},
 		{aliases: []string{"goroutines", "grs"}, group: goroutineCmds, cmdFn: goroutines, helpMsg: `List program goroutines.
 
 	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [-t (stack trace)] [-l (labels)]
@@ -1470,6 +1473,24 @@ func clearAll(t *Term, ctx callContext, args string) error {
 	return nil
 }
 
+func toggle(t *Term, ctx callContext, args string) error {
+	if args == "" {
+		return fmt.Errorf("not enough arguments")
+	}
+	id, err := strconv.Atoi(args)
+	var bp *api.Breakpoint
+	if err == nil {
+		bp, err = t.client.ToggleBreakpoint(id)
+	} else {
+		bp, err = t.client.ToggleBreakpointByName(args)
+	}
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s toggled at %s\n", formatBreakpointName(bp, true), t.formatBreakpointLocation(bp))
+	return nil
+}
+
 // byID sorts breakpoints by ID.
 type byID []*api.Breakpoint
 
@@ -2708,7 +2729,11 @@ func formatBreakpointName(bp *api.Breakpoint, upcase bool) string {
 	if id == "" {
 		id = strconv.Itoa(bp.ID)
 	}
-	return fmt.Sprintf("%s %s", thing, id)
+	state := "(enabled)"
+	if bp.Disabled {
+		state = "(disabled)"
+	}
+	return fmt.Sprintf("%s %s %s", thing, id, state)
 }
 
 func (t *Term) formatBreakpointLocation(bp *api.Breakpoint) string {
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -712,6 +712,7 @@ func isBreakpointExistsErr(err error) bool {
 }
 
 // AmendBreakpoint will update the breakpoint with the matching ID.
+// It also enables or disables the breakpoint.
 func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
@@ -733,7 +734,7 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 		delete(d.disabledBreakpoints, amend.ID)
 	}
 	if amend.Disabled && !disabled { // disable the breakpoint
-		if _, err := d.ClearBreakpoint(amend); err != nil {
+		if _, err := d.clearBreakpoint(amend); err != nil {
 			return err
 		}
 		d.disabledBreakpoints[amend.ID] = amend
@@ -775,7 +776,11 @@ func copyBreakpointInfo(bp *proc.Breakpoint, requested *api.Breakpoint) (err err
 func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
+	return d.clearBreakpoint(requestedBp)
+}
 
+// clearBreakpoint clears a breakpoint, we can consume this function to avoid locking a goroutine
+func (d *Debugger) clearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
 	if bp, ok := d.disabledBreakpoints[requestedBp.ID]; ok {
 		delete(d.disabledBreakpoints, bp.ID)
 		return bp, nil
@@ -826,29 +831,6 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
 	}
 	d.log.Infof("cleared breakpoint: %#v", clearedBp)
 	return clearedBp[0], nil
-}
-
-// ToggleBreakpoint toggles on or off a breakpoint.
-func (d *Debugger) ToggleBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
-	if dbp, ok := d.disabledBreakpoints[requestedBp.ID]; ok {
-		bp, err := d.target.SetBreakpointWithID(dbp.ID, dbp.Addr)
-		if err != nil {
-			return nil, err
-		}
-		requestedBp.Disabled = false
-		copyBreakpointInfo(bp, requestedBp)
-		delete(d.disabledBreakpoints, requestedBp.ID)
-		return api.ConvertBreakpoint(bp), nil
-	}
-
-	_, err := d.ClearBreakpoint(requestedBp)
-	if err != nil {
-		return nil, err
-	}
-	requestedBp.Disabled = true
-	d.disabledBreakpoints[requestedBp.ID] = requestedBp
-	d.log.Infof("toggled breakpoint: %#v", requestedBp)
-	return d.FindBreakpoint(requestedBp.ID), nil
 }
 
 // Breakpoints returns the list of current breakpoints.
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -70,6 +70,10 @@ type Debugger struct {
 	recordMutex   sync.Mutex
 
 	dumpState proc.DumpState
+	// Debugger keeps a map of disabled breakpoints
+	// so lower layers like proc doesn't need to deal
+	// with them
+	disabledBreakpoints map[int]*api.Breakpoint
 }
 
 type ExecuteKind int
@@ -198,6 +202,9 @@ func New(config *Config, processArgs []string) (*Debugger, error) {
 			return nil, err
 		}
 	}
+
+	d.disabledBreakpoints = make(map[int]*api.Breakpoint)
+
 	return d, nil
 }
 
@@ -502,7 +509,9 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 	}
 
 	discarded := []api.DiscardedBreakpoint{}
-	for _, oldBp := range api.ConvertBreakpoints(d.breakpoints()) {
+	breakpoints := api.ConvertBreakpoints(d.breakpoints())
+	d.target = p
+	for _, oldBp := range breakpoints {
 		if oldBp.ID < 0 {
 			continue
 		}
@@ -512,7 +521,7 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 				discarded = append(discarded, api.DiscardedBreakpoint{Breakpoint: oldBp, Reason: err.Error()})
 				continue
 			}
-			createLogicalBreakpoint(p, addrs, oldBp)
+			createLogicalBreakpoint(d, addrs, oldBp)
 		} else {
 			// Avoid setting a breakpoint based on address when rebuilding
 			if rebuild {
@@ -527,7 +536,6 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 			}
 		}
 	}
-	d.target = p
 	return discarded, nil
 }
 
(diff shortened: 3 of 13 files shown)


aarzilli (commented):
There is a major problem with this: it doesn't work with stepping.
Consider this situation:

```
func main() {
	lineOne()	// <- debugger is stopped here
	lineTwo()	// breakpoint 1 is set here
	lineThree()
}
```

The user now disables breakpoint 1 with `toggle 1` and then uses the `next` command.
The expected behavior is moving to the call to `lineTwo()`.
What happens at the moment is that the debugger returns the `next while nexting` error, beacuse the (*Breakpoint).IsInternal function is broken. 
That's not a big deal, however if that was fixed what it would do is step to the call to `lineThree()` skipping the second line.

This will happen because `(*Target).SetBreakpoint` believes that there is already a breakpoint on the call to `lineTwo()` and will add the internal breakpoint needed to make `next` work to it. However because that breakpoint is disabled it will never be hit.

This can be solved but IMO implementing disabled breakpoints like this is going to make breakpoint handling in pkg/proc too tricky. I think it would be better to hold a list of disabled breakpoints inside `debugger.Debugger` and then toggle either:

1. copies the breakpoint into the disabled list and calls `(*Target).ClearBreakpoint`, or
2. calls `(*Target).SetBreakpoint` and removes the breakpoint from the disabled list

The changes to make then are:

1. Change `(*Debugger).CreateBreakpoint` to complain if the address is already in the disabled breakpoint list
2. Change `(*Debugger).Breakpoints`, `(*Debugger).AmendBreakpoint`, `(*Debugger).ClearBreakponit` to also look in the disabled breakpoint list
3. Change `(*Debugger).Restart` to update the disabled breakpoints list

alexsaezm:
I tried implementing your suggestions and I see a behavior that I'm not sure how to handle. When I enable a breakpoint the breakpoint ID is different because `CreateBreakpoint` calls `SetBreakpoint` which increases `breakpointIDCounter`.

```
(dlv) breakpoints
Breakpoint 1 at 0x4b856f for main.lineOne() ./main.go:7 (1)
Breakpoint 2 at 0x4b860f for main.lineTwo() ./main.go:11 (0)

(dlv) toggle 1
Breakpoint 1 toggled at 0x4b856f for main.lineOne() ./main.go:7

(dlv) breakpoints
Breakpoint 2 at 0x4b860f for main.lineTwo() ./main.go:11 (0)

(dlv) toggle 1
Breakpoint 1 toggled at 0x4b856f for main.lineOne() ./main.go:7

(dlv) breakpoints
Breakpoint 2 at 0x4b860f for main.lineTwo() ./main.go:11 (0)
Breakpoint 3 at 0x4b856f for main.lineOne() ./main.go:7 (0)
```

(In the output, the `breakpoints` command still doesn't show the disabled breakpoints)

Any suggestions? I understand that a ID should be reuse.

service/test/integration2_test.go
```diff
@@ -506,6 +506,80 @@ func TestClientServer_clearBreakpoint(t *testing.T) {
...
+			if state.Exited {
+				continue
+			}
+			if state.Err != nil {
+				t.Fatalf("Unexpected error during continue: %v\n", state.Err)
+			}
+
+		}
+
+		if enableCount != 2 {
```
alexsaezm, line 573: Currently this test fails here. Although I think there's should be 2 hits it counts only 1. Maybe I don't understand how the test suite performs the execution of the test

alexsaezm:
Regarding `(*Debugger).Restart`, what should be the expected behavior? Currently it keeps the disabled breakpoints disabled.

service/debugger/debugger.go
```diff
@@ -629,6 +639,12 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoin
 		return nil, err
 	}
 
+	for _, addr := range addrs {
```
aarzilli, line 642: Can you move this check inside createLogicalBreakpoint? In case we ever reuse it somewhere else we'll want this check in there.

service/debugger/debugger.go
```diff
@@ -685,15 +701,21 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 	defer d.targetMutex.Unlock()
 
 	originals := d.findBreakpoint(amend.ID)
-	if originals == nil {
+	disabled := d.findDisabledBreakpoint(amend.ID)
```
aarzilli, line 704: Do it like this

```
if originals == nil {
    disabled := d.findDisabledBreakpoint(amend.ID)
    if disabled != nil {
        originals = []*api.Breakpoint{ disabled }
    }
}
```

then the rest of the function stays the same.

service/debugger/debugger.go
```diff
@@ -729,12 +751,18 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
 	defer d.targetMutex.Unlock()
 
 	var bps []*proc.Breakpoint
+	var dbps []*api.Breakpoint
```
aarzilli, line 754: Because of how this works you can just add this at the start and leave the rest as it is:

```
if bp, ok := d.disabledBreakpoints[requestedBp.Addr]; ok {
    delete(d.disabledBreakpoints, addr)
    return bp, nil
}
```

service/debugger/debugger.go
```diff
@@ -798,6 +867,7 @@ func (d *Debugger) FindBreakpoint(id int) *api.Breakpoint {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
 	bps := api.ConvertBreakpoints(d.findBreakpoint(id))
+	bps = append(bps, d.findDisabledBreakpoint(id))
```
aarzilli, line 870: Should only do the append if the result of `findDisabledBreakpoints` is not nil.

service/debugger/debugger.go
```diff
@@ -814,11 +884,25 @@ func (d *Debugger) findBreakpoint(id int) []*proc.Breakpoint {
...
+	return nil
+}
+
 // FindBreakpointByName returns the breakpoint specified by 'name'
 func (d *Debugger) FindBreakpointByName(name string) *api.Breakpoint {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
-	return d.findBreakpointByName(name)
+
+	bp := d.findBreakpointByName(name)
```
aarzilli, line 901: I think it's easier to just change `findBreakpointByName` to also search the disabledBreakpoints map and remove `findDisabledBreakpointByName`.

service/debugger/debugger.go
```diff
@@ -768,20 +796,61 @@ func (d *Debugger) ClearBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint
...
 	}
 	d.log.Infof("cleared breakpoint: %#v", clearedBp)
 	return clearedBp[0], nil
 }
 
+// ToggleBreakpoint toggles on or off a breakpoint.
+func (d *Debugger) ToggleBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
+	toggle := func(addr uint64) {
+		if dbp, ok := d.disabledBreakpoints[addr]; ok {
+			if bp, err := d.target.SetBreakpointWithID(dbp.ID, dbp.Addr); err == nil {
```
aarzilli, line 811: If this is successful you should call `copyBreakpointInfo(bp, dbp)`, otherwise the new breakpoint won't have the correct attributes.

aarzilli (commented):
The reason your test fails is that continue will stop after the first breakpoint it encounters, if you want it to continue until the end you have to set `Tracepoint: true` when you create bp1, bp2 and bp3 (this still won't work at the moment because you aren't calling copyBreakpointInfo, see comments above).

Regarding Restart, I think it's fine to leave it alone, if the address doesn't exist anymore the user will receive an error when trying to re-enable the breakpoint. That's fine.

I have some exceptions with the API changes however, I think it would be better to add a `Disabled bool` field to api.Breakpoint (service/api/types.go) instead of returning two lists in ListBreakpoints. Instead of the ToggleBreakpoint API call the AmendBreakpoint method will check


```
if !amend.Disabled && d.disabledBreakpoints[amend.Addr] != nil {
	// disable the breakpoint
} else if amend.Disabled && d.disabledBreakpoints[amend.Addr] == nil {
	// enable the breakpoint
}
```

When the breakpoint gets moved to the `disabledBreakpoints` map its Disabled field gets set to true.

service/debugger/debugger.go
```diff
@@ -190,6 +195,11 @@ func New(config *Config, processArgs []string) (*Debugger, error) {
 			return nil, err
 		}
 	}
+
+	if d.disabledBreakpoints == nil {
```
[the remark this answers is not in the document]
alexsaezm: 100% right

aarzilli:
Do you want a review or are you still working on this? (if former it needs to be rebased due to a merge conflict in debugger).

alexsaezm:
> Do you want a review or are you still working on this? (if former it needs to be rebased due to a merge conflict in debugger).

I'm still working on it. I realized I broke some tests and I think the CLI is right now not working as expected. I'll rebase the branch too.

Thanks

alexsaezm:
@aarzilli right now the integration1_test.go fails and I'm kind of confused why. 

```
--- FAIL: Test1Restart_breakpointPreservation (0.11s)
    integration1_test.go:121: Restart
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x8ebc99]

goroutine 30 [running]:
testing.tRunner.func1.1(0x985260, 0xd9ef40)
        /usr/lib/golang/src/testing/testing.go:1072 +0x30d
testing.tRunner.func1(0xc000f60000)
        /usr/lib/golang/src/testing/testing.go:1075 +0x41a
panic(0x985260, 0xd9ef40)
        /usr/lib/golang/src/runtime/panic.go:969 +0x1b9
github.com/go-delve/delve/service/test_test.Test1Restart_breakpointPreservation.func1(0xc000b420c0)
        /home/alex/Developer/src/github.com/go-delve/delve/service/test/integration1_test.go:125 +0x2f9
github.com/go-delve/delve/service/test_test.withTestClient1.func1(0xc000b420c0, 0xa220ad, 0x10, 0xc0000ce140, 0x1e, 0xc0000ba140, 0x50, 0xc00021a200, 0x3c)
        /home/alex/Developer/src/github.com/go-delve/delve/service/test/integration1_test.go:26 +0x2f
github.com/go-delve/delve/service/test_test.withTestClient1Extended(0xa220ad, 0x10, 0xc000f60000, 0xc000d93f20)
        /home/alex/Developer/src/github.com/go-delve/delve/service/test/integration1_test.go:59 +0x3b9
github.com/go-delve/delve/service/test_test.withTestClient1(0xa220ad, 0x10, 0xc000f60000, 0xc000057760)
        /home/alex/Developer/src/github.com/go-delve/delve/service/test/integration1_test.go:25 +0x67
github.com/go-delve/delve/service/test_test.Test1Restart_breakpointPreservation(0xc000f60000)
        /home/alex/Developer/src/github.com/go-delve/delve/service/test/integration1_test.go:107 +0x65
testing.tRunner(0xc000f60000, 0xa47cb0)
        /usr/lib/golang/src/testing/testing.go:1123 +0xef
created by testing.(*T).Run
        /usr/lib/golang/src/testing/testing.go:1168 +0x2b3
FAIL    github.com/go-delve/delve/service/test  0.626s
FAIL
exit status 1
make: *** [Makefile:16: test] Error 1
```

I'm not really sure why `state.CurrentThread` is nil at [line 125](https://github.com/go-delve/delve/blob/master/service/test/integration1_test.go#L125)

Would you mind take a look at it?

aarzilli:
> Would you mind take a look at it?

It happens because the process exited, which happens because the breakpoint called `firstbreakpoint` doesn't get properly restored by `Restart` which happens because you changed `createLogicalBreakpoint` to take a `*Debugger` instead of a `*Target` but at the point where `Restart` calls `createLogicalBreakpoint` the value of `d.target` hasn't been updated yet to be the new `*Target` object (notice where the `d.target = p` line is). I think you can move the update of `d.target` up (it's probably better, actually), but you have to be careful about the `d.breakpoints()` call, which needs to run with the old target.

alexsaezm:
> > Would you mind take a look at it?
> 
> It happens because the process exited, which happens because the breakpoint called `firstbreakpoint` doesn't get properly restored by `Restart` which happens because you changed `createLogicalBreakpoint` to take a `*Debugger` instead of a `*Target` but at the point where `Restart` calls `createLogicalBreakpoint` the value of `d.target` hasn't been updated yet to be the new `*Target` object (notice where the `d.target = p` line is). I think you can move the update of `d.target` up (it's probably better, actually), but you have to be careful about the `d.breakpoints()` call, which needs to run with the old target.

Thanks a lot for the explanation! Now it works flawlessly.

aarzilli:
> Thanks a lot for the explanation! Now it works flawlessly.

Good to know. Ping when you are done with the changes.

alexsaezm:
> > Thanks a lot for the explanation! Now it works flawlessly.
> 
> Good to know. Ping when you are done with the changes.

Does it look good to you @aarzilli ?

pkg/proc/breakpoints.go
```diff
@@ -285,7 +285,7 @@ func (t *Target) SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr)
 	}
 
 	if kind != UserBreakpoint {
-		bpmap.internalBreakpointIDCounter++
+		// TODO: Delete this line -> bpmap.internalBreakpointIDCounter++
```
aarzilli, line 288: If this just needs to be deleted the comment should be removed too. But what's the problem?

pkg/terminal/command.go
```diff
@@ -189,7 +189,10 @@ Current limitations:
 
 	clearall [<linespec>]
 
-If called with the linespec argument it will delete all the breakpoints matching the linespec. If linespec is omitted all breakpoints are deleted.`},
+	If called with the linespec argument it will delete all the breakpoints matching the linespec. If linespec is omitted all breakpoints are deleted.`},
```
aarzilli, line 192: Remove the new tab (the idea is to have only the syntax indented so that it is rendered in monospace in markdown).

service/debugger/debugger.go
```diff
@@ -701,6 +716,13 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
 	defer d.targetMutex.Unlock()
 
 	originals := d.findBreakpoint(amend.ID)
+	if originals == nil {
+		if _, ok := d.disabledBreakpoints[amend.ID]; ok {
+			d.disabledBreakpoints[amend.ID] = amend
```
aarzilli, line 721: Here you should check if `amend.Disabled` is false and if it is re-enable the breakpoint, like you do in the first half of `(*Debugger).ToggleBreakpoint`.

After the ValidBreakpointName check you should check if `amend.Disabled` is true and if it is disable the breakpoint.

This way the functionality of ToggleBreakpoint can be provided through the `AmendBreakpoint` API interface.

service/debugger/debugger.go
```diff
@@ -701,6 +716,13 @@ func (d *Debugger) AmendBreakpoint(amend *api.Breakpoint) error {
...
+		}
+		return nil
```
aarzilli, line 723: This needs to return an error like we did before (the `if originals == nil` below can never be reached).

pkg/proc/breakpoints.go
alexsaezm: While I was doing my tests, that line created some issues. I just tested it again and everything works so I'll remove my change.

aarzilli:
In general this is ok. The new API method ToggleBreakpoint however is now just a special case of AmendBreakpoint and should be removed. Clients will just call FindBreakpoint + AmendBreakpoint.

aarzilli:
PS. Also if you could do a rebase on master, so we can run the tests on mac and arm64.

derekparker:
> In general this is ok. The new API method ToggleBreakpoint however is now just a special case of AmendBreakpoint and should be removed. Clients will just call FindBreakpoint + AmendBreakpoint.

I actually like the `ToggleBreakpoint` client API method even if it may just be a convenience. Also, with the toggling we move the breakpoint to another list internally, etc... so I think it's appropriate to let it continue to have its own method.

aarzilli:
> > In general this is ok. The new API method ToggleBreakpoint however is now just a special case of AmendBreakpoint and should be removed. Clients will just call FindBreakpoint + AmendBreakpoint.
> 
> I actually like the `ToggleBreakpoint` client API method even if it may just be a convenience. Also, with the toggling we move the breakpoint to another list internally, etc... so I think it's appropriate to let it continue to have its own method.

In that case it should just be in rpc2/server.go, there's no need for a ToggleBreakpoint method in Debugger that duplicates the work that AmendBreakpoint does.

derekparker:
> > > In general this is ok. The new API method ToggleBreakpoint however is now just a special case of AmendBreakpoint and should be removed. Clients will just call FindBreakpoint + AmendBreakpoint.
> > 
> > 
> > I actually like the `ToggleBreakpoint` client API method even if it may just be a convenience. Also, with the toggling we move the breakpoint to another list internally, etc... so I think it's appropriate to let it continue to have its own method.
> 
> In that case it should just be in rpc2/server.go, there's no need for a ToggleBreakpoint method in Debugger that duplicates the work that AmendBreakpoint does.

Agreed.

alexsaezm:
> > > In general this is ok. The new API method ToggleBreakpoint however is now just a special case of AmendBreakpoint and should be removed. Clients will just call FindBreakpoint + AmendBreakpoint.
> > 
> > 
> > I actually like the `ToggleBreakpoint` client API method even if it may just be a convenience. Also, with the toggling we move the breakpoint to another list internally, etc... so I think it's appropriate to let it continue to have its own method.
> 
> In that case it should just be in rpc2/server.go, there's no need for a ToggleBreakpoint method in Debugger that duplicates the work that AmendBreakpoint does.

I removed the ToggleBreakpoint method in Debugger and refactored a little the ClearBreakpoint method because I hit a lock.

aarzilli (approved):
LGTM, thanks.

derekparker:
I think the failing tests are unrelated. Running the TeamCity build again.

PR #2210, opened 2020-10-21
proc: support reading deferred calls' arguments on linux/arm64

aarzilli:

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -3,10 +3,9 @@ Tests skipped by each supported backend:
 * 386 skipped = 2.1% (3/145)
 	* 1 broken
 	* 2 broken - cgo stacktraces
-* arm64 skipped = 2.8% (4/145)
+* arm64 skipped = 2.1% (3/145)
 	* 2 broken
 	* 1 broken - global variable symbolication
-	* 1 broken - reading defers
 * darwin/lldb skipped = 0.69% (1/145)
 	* 1 upstream issue
 * freebsd skipped = 7.6% (11/145)
--- a/pkg/proc/arch.go
+++ b/pkg/proc/arch.go
@@ -15,6 +15,7 @@ type Arch struct {
 	breakpointInstruction []byte
 	breakInstrMovesPC     bool
 	derefTLS              bool
+	usesLR                bool // architecture uses a link register, also called RA on some architectures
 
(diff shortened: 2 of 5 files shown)


aarzilli:
cc @hengwu0

hengwu0:
1、Sorry i forgot this feature is undone.
2、The patch in `EvalScope` is Very well done.
3、I’m dissagreed to add `usesLR` variable in **Arch struct** actually. For Two reason:
&nbsp;&nbsp;&nbsp;&nbsp;a. The `LR` Only exists in arm, this bool variable works like `if Arch.Name=="arm64"`
&nbsp;&nbsp;&nbsp;&nbsp;b. If we continue to add `Architecture independent variables` in `Arch struct` on arm32/mips/ppc with different OS, the `Arch struct` will become more and more complex and huge.

aarzilli:
> The LR Only exists in arm, this bool variable works like if Arch.Name=="arm64"

It also exists on arm (non-64bit version), MIPS (both 32 and 64bit), PowerPC, RISC V, and (according to go source code) s390x.

hengwu0:
> > The LR Only exists in arm, this bool variable works like if Arch.Name=="arm64"
> 
> It also exists on arm (non-64bit version), MIPS (both 32 and 64bit), PowerPC, RISC V, and (according to go source code) s390x.

I know it is called `RA(return address) register` in mips(both 32 and 64bit) and riscv. If you mean LR is RA, the name of `usesLR` may not precise. :)

aarzilli:
> I know it is called RA(return address) register in mips(both 32 and 64bit) and riscv. If you mean LR is RA, the name of usesLR may not precise. :)

It's the name that the runtime uses. PC is called IP on amd64/386

hengwu0:
> It's the name that the runtime uses. PC is called IP on amd64/386

PC called IP on amd64/386 is well known for everyone, and RA called LR may not be popular. How about add an comment here to let mips/riscv developers known?

aarzilli:
> > It's the name that the runtime uses. PC is called IP on amd64/386
> 
> PC called IP on amd64/386 is well known for everyone, and RA called LR may not be popular. How about add an comment here to let mips/riscv developers known?

Done

hengwu0:
Thanks for accepting my suggestion, LGTM.

derekparker (approved):
LGTM

PR #2211, opened 2020-10-22
service/rpccommon: log error for conns rejected by --only-same-user

### LINKED ISSUE #2209: Better reporting for failed connection due to --only-same-user
```
Delve Debugger
Version: 1.5.0
```

```
go version go1.15.3 linux/amd64
```

Win10 19041.572 and WSL.

4. What did you do?

Started delve inside WSL with the command

```
dlv debug --headless --listen=127.0.0.1:2345 --api-version=2 --accept-multiclient .
```

And tried to connect to it from inside GoLand in host Windows.

Connection either succeeds or fails with some way to understand why it failed.

6. What did you see instead?

Connection fails. Delve inside WSL doesn't print anything to console. Once started with the additional `--log --log-output=rpc` option, it prints that the reason is the only-same-user policy. 

It would be great to get this reason in the response from delve, or at least printed to the console.

aarzilli:
If no logger is enabled manually write to stderr instead.

Fixes #2209

### DIFF
--- a/pkg/logflags/logflags.go
+++ b/pkg/logflags/logflags.go
@@ -127,7 +127,7 @@ func WriteAPIListeningMessage(addr string) {
 	writeListeningMessage("API", addr)
 }
 
-func writeListeningMessage(server string, addr string) {
+func writeListeningMessage(server, addr string) {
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

Issue #2203, opened 2020-10-16
Apply `config substitute-path` to `break` and `trace`


hyangah:
Currently `config substitute-path` rule applies only to source code retrieval, but does not apply to `break` and `trace`. 
https://github.com/go-delve/delve/blob/71a460fc0fe578cc81dab152e3dd2515a9304da1/pkg/config/config.go#L231-L232

When users opt into use `config substitute-path`, they need to interact with the source code in different location and absolute file paths.  They _often incorrectly assume_ the rule would apply to other commands. Please consider expanding the scope of path substitution.

Also, I noticed that the `list` or source code visualization uses the substitute path only to retrieve the source code. It keeps using the original path in display. If this feature request is accepted, changing that with the substituted path will be desirable for consistency.

```
(dlv) config substitute-path /dir/a /dir/b
(dlv) b /dir/a/main.go:9
Breakpoint 1 set at 0x106cedf for main.main() /dir/a/main.go:9
(dlv) c
1
> main.main() /dir/a/main.go:9 (hits goroutine(1):1 total:1) (PC: 0x106cedf)
     4:	
     5:	var a = 1
     6:	
     7:	func main() {
     8:		println(a)
=>   9:		a++
    10:		
 ...
```

aarzilli:
This isn't a simple change, the argument of `break` (and `trace`) is not a filepath, it's a location expression that's interpreted on the backend-side of delve, whereas substitutions are currently applied by the frontend.

hyangah:
It would be nice if users can choose this to be applied to backends - e.g.  rpc methods like `CreateBreakpoint` that take Breakpoint with the file name as the input, and the dap implementation. cc @polinasok

aarzilli:
Implemented by db93049813b48b48ca60187aea1ab6e15b04826f

PR #2213, opened 2020-10-26
service,terminal: apply substitute path to trace/break argument

aarzilli:
Change FindLocation to apply substitute path rules to location
expressions. Changes terminal to always print paths after applying
substitutions.

Implements #2203

### DIFF
--- a/pkg/locspec/locations.go
+++ b/pkg/locspec/locations.go
@@ -20,7 +20,7 @@ const maxFindLocationCandidates = 5
 // LocationSpec is an interface that represents a parsed location spec string.
 type LocationSpec interface {
 	// Find returns all locations that match the location spec.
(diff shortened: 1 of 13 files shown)


pkg/locspec/locations.go
```diff
@@ -434,6 +438,30 @@ func (loc *NormalLocationSpec) Find(t *proc.Target, processArgs []string, scope
 	return []api.Location{addressesToLocation(addrs)}, nil
 }
 
+func substitutePath(path string, rules [][2]string) string {
```
derekparker, line 441: I'm not a big fan of duplicating this from the terminal package. What are your thoughts on moving this logic here instead and exporting it such that the frontend can use it as well?

aarzilli: Done.

derekparker (approved):
LGTM

PR #2214, opened 2020-10-29
Go 1.16 support branch

aarzilli:

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -1709,7 +1709,7 @@ func TestAcceptMulticlient(t *testing.T) {
 			},
 		})
 		if err := server.Run(); err != nil {
-			t.Fatal(err)
+			panic(err)
 		}
 		<-disconnectChan
 		server.Stop()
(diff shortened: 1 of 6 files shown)


oxisto:
This might actually fix a problem which I have on #2254 on arm64 with the `TestCGONext` test.

service/test/integration2_test.go
```diff
@@ -1709,7 +1709,7 @@ func TestAcceptMulticlient(t *testing.T) {
 			},
 		})
 		if err := server.Run(); err != nil {
-			t.Fatal(err)
+			panic(err)
```
derekparker, line 1712: This is the only change that doesn't make immediate sense to me.

oxisto: I had the same change on my other branch. It seems that DeepSource complaints about this, since the error is related to the server run but not the test itself, thus it seems more appropriate to panic instead of fail the test.

derekparker: Ack, makes sense.

derekparker (approved):
LGTM

PR #2220, opened 2020-11-03
proc/native: increase max size of XSAVE area to support newer CPUs

### LINKED ISSUE #2219: Command failed: could not restore registers: bad address
Delve apparently cannot evaluate expressions, even in the simplest cases. Reproduction:

```
$ cat > hello.go <<'EOF'
package main

import "fmt"

func Add(a, b int) int {
        return a + b
}

func main() {
        fmt.Println("Hello, World")
        fmt.Printf("The result is %d\n", Add(1, 2))
}
EOF
$ dlv debug
(dlv) b hello.go:11
Breakpoint 1 set at 0x4bc67a for main.main() ./hello.go:11
(dlv) c
...
(dlv) call Add(4,5)
> main.main() ./hello.go:11 (hits goroutine(1):2 total:2) (PC: 0x4bc67a)
Command failed: could not restore registers: bad address
```

I expected to see "9" or some version of it. It looks like the only plausible place where this could come from is one of these `SYS_PTRACE` calls:

https://github.com/go-delve/delve/blob/3c8d4d52b825e313f21824da63c0904ba2ebb76e/pkg/proc/native/threads_linux_amd64.go#L34

but I'm afraid my knowledge of ptrace isn't enough to go further. This is similar to https://github.com/go-delve/delve/issues/1802, except completely reproducible.

Delve version:
```
Delve Debugger
Version: 1.5.0
```

Go version: `go version go1.15 linux/amd64`
OS: `5.7.10-1rodete2-amd64 #1 SMP Debian 5.7.10-1rodete2 (2020-08-12 > 2018) x86_64 GNU/Linux` (Xeon Gold 6154 in case that matters)

aarzilli:
Fixes #2219

### DIFF
--- a/pkg/proc/native/registers_freebsd_amd64.go
+++ b/pkg/proc/native/registers_freebsd_amd64.go
@@ -71,7 +71,7 @@ func registers(thread *nativeThread) (proc.Registers, error) {
 }
 
 const (
-	_X86_XSTATE_MAX_SIZE = 2688
+	_X86_XSTATE_MAX_SIZE = 2696
 	_NT_X86_XSTATE       = 0x202
 
 	_XSAVE_HEADER_START          = 512
--- a/pkg/proc/native/registers_linux_amd64.go
+++ b/pkg/proc/native/registers_linux_amd64.go
@@ -66,7 +66,7 @@ func registers(thread *nativeThread) (proc.Registers, error) {
 }
 
 const (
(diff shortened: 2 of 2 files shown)


aarzilli:
GDB had the same problem https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1647053.html

derekparker (approved):
LGTM

PR #2224, opened 2020-11-09
[WIP] fix linux/arm64 on Go 1.15.4

aarzilli:
### dwarf/line: pick last exact match in PCToLine

When there are multiple exact match entries for a single PC value pick
the last one.


aarzilli:
Closing in favor of #2235

PR #2225, opened 2020-11-11
Travis-CI: hold back version of Go 1.15.x to 1.15.3

aarzilli:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ arch:
   - arm64
 
 go:
-  - 1.15.x
+  - 1.15.3 # hold back version of Go, see issue #42484
   - 1.14.x
   - 1.13.x
   - tip


aarzilli:
There is a problem with the debug_lines section generated by Go 1.15.4, see https://github.com/golang/go/issues/42484, I propose that for now we keep our CI on 1.15.3. In practice this seems to only affect linux/arm64 (at least on our tests, I think it can affect amd64 binaries as well) but I couldn't figure out how to just have arm64 on 1.15.3 and I don't think it's a big deal.

If my fix for Go is approved and backported we'll just reenable 1.15.x, if it isn't I have [a workaround for delve](https://github.com/go-delve/delve/pull/2224) that passes tests on 1.15.4 that's not very elegant.

derekparker (approved):
LGTM

PR #2230, opened 2020-11-19
proc: add flag to distinguish ReturnValues

aarzilli:
Adds a flag that distinguishes the return values of an injected
function call from the return values of a function call executed by the
target program.


aarzilli:
cc @polinasok

polinasok (approved):
Oh, I have been planning to file an issue for this, but here is the fix already! This is great! Thank you for making the change!

derekparker (approved):
LGTM

PR #2232, opened 2020-11-23
proc: make nested function calls work when stopped at a sw breakpoint

aarzilli:
evalFunctionCall needs to remove the breakpoint from the current thread
after starting the function call injection, otherwise Continue will
think that the thread is stopped at a breakpoint and return to the user
instead of continuing the call injection.

### DIFF
--- a/_fixtures/fncall.go
+++ b/_fixtures/fncall.go
@@ -193,7 +193,7 @@ func main() {
 
 	d := &Derived{3, Base{4}}
 
-	runtime.Breakpoint()
+	runtime.Breakpoint() // breakpoint here
 	call1(one, two)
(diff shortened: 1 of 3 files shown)


aarzilli:
Funny thing with this. The check in isCallInjectionStop was always wrong: if we catch a thread on the first instruction of runtime.debugCallV1 it will look at the padding the compiler adds between functions (which is all int3 instructions). It didn't cause problems because we didn't clear the breakpoint flag on the thread so on resume ContinueOnce would always do a single step on the threads being injected with a call. Fixing that bug made the isCallInjectionStop bug happen much more frequently.

derekparker (approved):
LGTM

PR #2235, opened 2020-11-25
dwarf/line: do not consider end of sequence entries valid

aarzilli:
This is needed to fix the problem we have with 1.15.4+ after the
backport is applied to it. The DWARF standard isn't clear on what
should happen with the end_of_sequence opcode but this is consistent
with debug/dwarf.LineReader as well as gdb.


derekparker (approved):
LGTM

PR #2254, opened 2020-12-07
Fixing darwin native backend [WIP]

### LINKED ISSUE #1112: could not get thread count - strikes again
00e473157ad1759aba846b8420443c59aece253f
Go 1.9.3
OSX 10.13.3 / amd64
4. What did you do?
Debugged this app:
```
package main

import (
    "time"
    "fmt"
)

func main() {
    o, ok := time.ParseInLocation("2006-01-02 15:04:05", "2017-10-29 00:01:49", time.Local)
    o = o.Local()
    t, ok := time.ParseInLocation("2006-01-02 15:04:05", "2017-10-30 00:01:49", time.Local)
    if ok != nil {
        fmt.Printf("Could not parse")
    } else {
        fmt.Printf("Now in nano : %d ", t.UnixNano())
    }
    t = t.Local()
    n := time.Unix(0, t.UnixNano())
    fmt.Printf("Original : %s, Converted back : %s\n", t.Format(time.RFC3339Nano), n.Format(time.RFC3339Nano))
    if ok != nil {
        fmt.Printf("Failed to parse time, cannot compute: %v", ok)
    }
}
```
The debugger working
6. What did you see instead?
```
could not launch process: could not get thread count

Debugger finished with exit code 1
```

Additional information

```
ls -l /Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver
rwxr-xr-x  1 root  admin  4778272 17 nov 09:48 /Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver
Pingus-2:opushserver charlus$ /Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver
debugserver-@(#)PROGRAM:debugserver  PROJECT:debugserver-900.0.64
 for x86_64.
Usage:
  debugserver host:port [program-name program-arg1 program-arg2 ...]
  debugserver /path/file [program-name program-arg1 program-arg2 ...]
  debugserver host:port –attach=<pid>
  debugserver /path/file –attach=<pid>
  debugserver host:port –attach=<process_name>
  debugserver /path/file –attach=<process_name>
```

Originally reported at https://youtrack.jetbrains.com/issue/GO-5305

### LINKED ISSUE #2246: Apple M1 ARM support
Hi All

Hope you're all safe and well.

You're probably already aware of this but just in case.

When trying to debug on Apple M1, it comes back with "rosetta error: failed to allocate vm space for aot".

Just wondering if there's any recommendations please.

Love your work 😊👍

oxisto:
Added correct entitlements and replaced the fork_exec mechanism to a POSIX spawn. This now enables the launch of the inferior and successfully calls task_for_pid().

The following functionality has been tested:
* Setting breakpoints
* Thread continue
* Thread cpu instruction step
* Reading go routine
* Printing stracktrace
* Reading variables / locals

If finished, this fixes #2246 and it fixes #1112. The reason, I chose to fix #2246 this way is because I am running into more blocking issues using Apple's debugserver and the gdbserver approach. Generally, I would be fine with the gdbserver as default backend on darwin amd64 and the native approach on darwin arm64.


oxisto:
Ah, and it seems that for some reason, `make test` does not work (yet). You have to manually call `make.go` - not sure why. For example:
```bash
➜  delve git:(fix-native-backend) ✗ go run _scripts/make.go test -b native -s github.com/go-delve/delve/pkg/proc -v -r ^TestBreakpoint$
go test -count 1 -p 1 -v -exec=/Users/oxisto/Repositories/delve/_scripts/testsign -tags=macnative "-ldflags=-s -X main.Build=979d6d6836af0b15b8746cea26ec5e78e0f89b0a" github.com/go-delve/delve/pkg/proc -run=^TestBreakpoint$ -backend=native
=== RUN   TestBreakpoint
    support.go:248: enabling recording for github.com/go-delve/delve/pkg/proc_test.TestBreakpoint
--- PASS: TestBreakpoint (0.51s)
PASS
ok  	github.com/go-delve/delve/pkg/proc	1.056s```

File: pkg/proc/native/threads_darwin.go [hunk 1]
```diff
@@ -1,4 +1,4 @@
-//+build darwin,macnative
+//+build !arm64,macnative
```
aarzilli: The file name should be changed to threads_darwin_amd64.go and the build constraint should be `darwin,amd64,macnative`.

pkg/proc/native/threads_darwin_arm64.go
```diff
@@ -0,0 +1,134 @@
+//+build arm64 darwin,macnative
```
aarzilli, line 1: This should be darwin,arm64,macnative

pkg/proc/native/registers_darwin_arm64.go
```diff
@@ -0,0 +1,290 @@
+//+build arm64 darwin,macnative
```
aarzilli, line 1: Should be darwin,arm64,macnative.

pkg/proc/native/registers_darwin_amd64.go
```diff
@@ -1,4 +1,4 @@
-//+build darwin,macnative
+//+build !arm64 macnative
```
aarzilli, line 1: Should be darwin,amd64,macnative.

service/test/integration2_test.go
```diff
@@ -1708,7 +1708,7 @@ func TestAcceptMulticlient(t *testing.T) {
 			},
 		})
 		if err := server.Run(); err != nil {
-			t.Fatal(err)
+			panic(err)
```
aarzilli, line 1711: why?

pkg/proc/native/threads_darwin.go
oxisto: I think you were looking at an old version of the code. I have extracted the common parts of the Darwin threads into threads_darwin.go, the specifics go into the _arm64 and _amd64.go files. Since the only specific thing is the thread_state struct I was actually thinking about a typedef to avoid having those platform specific files alltogether.

service/test/integration2_test.go
oxisto: I can revert that. I think go `vet` was complaining at some point since the Fatal is not tied to the test itself so it should not throw a `t.Fatal`.

oxisto:
How should I proceed with these issues from DeepSource? They all seem not to be related to this PR, not sure why they creep up now.

Is there a way to re-trigger the Travis CI tests? I guess cirrus only checks Linux, right?

aarzilli:
> How should I proceed with these issues from DeepSource?

Ignore them if it's not something you added.

> Is there a way to re-trigger the Travis CI tests?

No, Travis CI recently changed their plan for open source software and now we run out of credits in a couple of days. This is an ongoing problem at the moment.

> I guess cirrus only checks Linux, right?

Freebsd, actually.

pkg/proc/native/proc_darwin.c
```diff
@@ -133,67 +133,138 @@ thread_count(task_t task) {
...
+		return "EXC_BREAKPOINT";
+	} else if (code == EXC_BAD_ACCESS) {
+		return "EXC_BAD_ACCESS";
+	}
+
+	return NULL;
+}
+
+const char* signal_string(int signal) {
+	if (signal == SIGHUP) {
```
rumpelsepp, line 150: Maybe like this?

``` c
switch(signal) {
case SIGHUP:
// …
}
```

pkg/proc/native/exec_darwin.c
```diff
@@ -5,110 +5,76 @@
...
+            exit(1);
+        }
+    }
 
-	sleep(1);
+    short flags = POSIX_SPAWN_START_SUSPENDED | POSIX_SPAWN_SETSIGDEF |
+          POSIX_SPAWN_SETSIGMASK;
+
+#if defined(__arm64__)
+	// Force ASLR disable on arm64. This might not be needed if a non-PIE
```
derekparker, line 53: We have a flag to control disabling ASLR, why must we force it here?

pkg/proc/native/proc_darwin.c
derekparker: Yeah, I agree, we can switch these if/else chains with a switch.

pkg/proc/native/proc_darwin.c
```diff
@@ -133,67 +133,133 @@ thread_count(task_t task) {
...
-	kret = mach_msg(&msg.hdr, opts,
-			0, sizeof(msg.data), port_set, 10, MACH_PORT_NULL);
+	kret = mach_msg((mach_msg_header_t*) &req, opts,
+			0, sizeof(req), port_set, 10, MACH_PORT_NULL);
 	if (kret == MACH_RCV_INTERRUPTED) return kret;
 	if (kret != MACH_MSG_SUCCESS) return 0;
 
+	__Request__mach_exception_raise_t* msg = (__Request__mach_exception_raise_t*) req;
+	// TODO: we actually want to propagate the task back to Go, but this seems to randomly hang
+	// *task = msg->task.name;
```
derekparker, line 184: Are there cases where we wouldn't want to?

derekparker (changes requested):
Initial pass done, few nits and questions.

I do want to say a huge thank you for taking this on, however. Cleaning up and fixing the native backend has been on my list for a long time now, and I'm really happy you've submitted this patch to improve things, so thanks a lot!

_scripts/testsign
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
-codesign -s $CERT $1
+scripts_folder=$(dirname $0)
+CERT=-
+codesign -v -v $1
```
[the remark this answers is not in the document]
oxisto: The first one was just for debugging purposes and can be removed.

pkg/proc/native/exec_darwin.c
```diff
@@ -5,110 +5,76 @@
...
-		size_t n = read(efd[0], &w, 1);
-		close(efd[0]);
-		if (n != 0) {
-			// Child died, reap it.
-			waitpid(pid, NULL, 0);
-			return -1;
-		}
-		return pid;
-	}
+	kern_return_t kret = 0;
```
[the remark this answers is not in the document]
oxisto: Is there a clang-format file or something similar to make it consistent?

pkg/proc/native/exec_darwin.c
oxisto: This is a somewhat weird behaviour that occurs on darwin/arm64. Even with `buildmode=exe`, it seems that always a PIE exe is built, so we need to disable ASLR to counter it. Another, properly better way would be to use the flag and force the flag on darwin/arm64.

pkg/proc/native/proc_darwin.c
oxisto: Well, propagating the task back to Go has one main use case: Getting an information from the inferior, if it uses `execve`. Previously, this was needed because of the exec/fork-flow, which was quite unstable and susceptible to race conditions. Now, the only use case that remains would be, if an inferior itself uses `execve` so the underlying task would change and we want to propagate this back to Go.

I can have another look at this, the whole function is not ideal, since it is recursively done in C. While I doubt that there is a large number of recursions, this is still not ideal.

A "cleaner" way would probably be to return to Go for each exception, have a look at the exception in Go and decide to wait again there. This could however be a little bit slower because of additional context switches.

pkg/proc/native/exec_darwin.c
oxisto: I am running into a problem using the `disableASLR` flag. While it is easy to use it for the regular build, I have no way to set it during the tests. Should I extend the tests to pass it through, similarly to the backend - but this will take a lot of adjustments, or somehow force it on another level?

pkg/proc/native/exec_darwin.c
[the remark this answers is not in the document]
derekparker: No, we don't have anything like that, but I wouldn't be opposed to it while we have to maintain this C code.

pkg/proc/native/exec_darwin.c
derekparker: Hm, yeah enabling this for tests can be tricky. You would have to modify `withTestProcessArgs` to accept launch flags. I don't know if this is worth it for now. If darwin/arm64 simply won't work without it disabled, I'm ok with a heavier-handed approach like this, but I wanted to call some attention to it first.

pkg/proc/native/proc_darwin.c
derekparker: Ack, makes sense.

I agree with tossing things over the fence back to Go for inspection, etc. I'm probably fine with taking the performance penalty there for the sake of added safety and correctness.

oxisto: I addressed this with a goto/label approach and also the task propagation seems to work now after my recent commit

pkg/proc/native/exec_darwin.c
oxisto: I have decided to force it in the `proc_darwin.go` instead of burying deep down in the C code. Probably the lesser of the evils.

pkg/proc/native/exec_darwin.c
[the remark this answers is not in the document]
oxisto: Added a .clang-format which *seems* to be very close to the syntax that was used in those C files.

oxisto:
Unfortunately, I am running into some issues with implementing trace functionality, i.e. attaching to an existing PID under `darwin/arm64`.

As you can see here, Go forces PIE under `darwin/arm64`:
https://github.com/golang/go/blob/cb84d831c956026f477b52c9f8a7c1ed2b2724ad/src/cmd/link/internal/ld/config.go#L39-L40

While launching an executable, we can disable ASLR, but when attaching to existing processes, we can not :(

oxisto:
> Unfortunately, I am running into some issues with implementing trace functionality, i.e. attaching to an existing PID under `darwin/arm64`.
> 
> As you can see here, Go forces PIE under `darwin/arm64`:
> https://github.com/golang/go/blob/cb84d831c956026f477b52c9f8a7c1ed2b2724ad/src/cmd/link/internal/ld/config.go#L39-L40
> 
> While launching an executable, we can disable ASLR, but when attaching to existing processes, we can not :(

Good news, I managed to retrieve the start of the `__TEXT` section using `mach_vm_region`, so debugging with ASLR enabled works now!

derekparker:
Looks like CI is passing this this patch! I've been off for the holidays the last 2 weeks so I'm in catch up mode, but I will review this patch this week.

My current line of thinking is to cut the initial v1.6 release with lldb-server based darwin/arm64 support and then review & merge this, let it simmer for a while, let folks test it, and then include it when we do a v1.6.x release.

oxisto:
> Looks like CI is passing with this patch! I've been off for the holidays the last 2 weeks so I'm in catch up mode, but I will review this patch this week.
> 
> My current line of thinking is to cut the initial v1.6 release with lldb-server based darwin/arm64 support and then review & merge this, let it simmer for a while, let folks test it, and then include it when we do a v1.6.x release.

Sounds like a good idea! In the long run you might want to decide whether to keep the native backend or not. Now at least there is Darwin/amd64 support through the lldb server.

AndrewRPorter:
Any updates on this? I am unable to debug my go application, presumably because of this? How can I test this fix locally?

oxisto:
After a long absence on my part, I have decided the rebase this against the current master. I have also reverted some changes that made this the default since we now have support using LLDB on darwin/arm64 now.

How should we proceed with this @derekparker? I did not have have time yet to check what has changed in the last couple of months. What is your future plan with regards to this?

derekparker:
@oxisto if you are still open to completing this work, feel free to rebase this PR and reopen it and I will make it a priority to review and give feedback.

oxisto:
> @oxisto if you are still open to completing this work, feel free to rebase this PR and reopen it and I will make it a priority to review and give feedback.

Unfortunately I do not see any free time this year to tackle this :( Maybe next year.

PR #2256, opened 2020-12-09
proc: use CPUID to determine maximum size of XSAVE area

aarzilli:
the maximum size of the Xsave area was previously hardcoded but the
CPUID instruction can be used to determine its maximum size.

### DIFF
--- a/pkg/proc/amd64util/xsave.go
+++ b/pkg/proc/amd64util/xsave.go
@@ -1,0 +1,131 @@
+package amd64util
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/go-delve/delve/pkg/proc"
+)
+
+// AMD64Xstate represents amd64 XSAVE area. See Section 13.1 (and
+// following) of Intel® 64 and IA-32 Architectures Software Developer’s
+// Manual, Volume 1: Basic Architecture.
+type AMD64Xstate struct {
+	AMD64PtraceFpRegs
+	Xsave       []byte // raw xsave area
+	AvxState    bool   // contains AVX state
+	YmmSpace    [256]byte
+	Avx512State bool // contains AVX512 state
+	ZmmSpace    [512]byte
+}
+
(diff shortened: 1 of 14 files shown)


pkg/proc/amd64util/xsave.go
```diff
@@ -0,0 +1,130 @@
...
+	Rdp      uint64
+	Mxcsr    uint32
+	MxcrMask uint32
+	StSpace  [32]uint32
+	XmmSpace [256]byte
+	Padding  [24]uint32
+}
+
+// Decode decodes an XSAVE area to a list of name/value pairs of registers.
+func (xsave *AMD64Xstate) Decode() (regs []proc.Register) {
```
derekparker, line 39: nit: why the named return here? We don't really use it.

pkg/proc/amd64util/xsave_amd64.go
```diff
@@ -0,0 +1,30 @@
+package amd64util
+
+import (
+	"sync"
+)
+
+var xstateMaxSize int
+var loadXstateMaxSizeOnce sync.Once
+
+func cpuid(axIn, cxIn uint32) (axOut, bxOut, cxOut, dxOut uint32)
```
derekparker, line 10: nit: we could use something like: https://github.com/intel-go/cpuid.

derekparker (commented):
Small nits otherwise LGTM.

pkg/proc/amd64util/xsave.go
aarzilli: Done.

pkg/proc/amd64util/xsave_amd64.go
aarzilli: In general I think it's easier to read and understand the code as is than to understand the terminology used by that library (the intel manual doesn't use the term "extra features" as far as I can tell). In this particular case, it doesn't seem have one of the things that we need: it doesn't call `CPUID AX=0xd CX=0x0` which returns the size of the XSAVE area.

derekparker: Ack.

derekparker (approved):
LGTM

PR #2260, opened 2020-12-11
service/dap: Support local attach

polinasok:
Updates #1515

### DIFF
--- a/_fixtures/loopprog.go
+++ b/_fixtures/loopprog.go
@@ -6,7 +6,7 @@ func loop() {
 	i := 0
 	for {
 		i++
-		if (i % 100000000) == 0 {
+		if (i % 100000) == 0 {
 			fmt.Println(i)
 		}
 	}
--- a/_fixtures/loopprog.go
+++ b/_fixtures/loopprog.go
@@ -6,7 +6,7 @@ func loop() {
 	i := 0
 	for {
 		i++
-		if (i % 100000) == 0 {
+		if (i % 100000000) == 0 {
 			fmt.Println(i)
 		}
 	}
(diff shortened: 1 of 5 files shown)


polinasok:
I think there is a race condition that I need to debug.

aarzilli:
I think it would be better if the test for this was more like TestAttachDetach in proc_test and used testnextnethttp, which will wait forever, instead of changing loopprog.

polinasok:
> I think it would be better if the test for this was more like TestAttachDetach in proc_test and used testnextnethttp, which will wait forever, instead of changing loopprog.

I changed loopprog to print status messages a bit less often, so I could use it for manual testing. The change is not necessary for the unittest. I can remove it.

Using testnextnethttp has more moving parts. I would need to wait for the server to start listening, then have a client and send a request to trigger the code that has the breakpoint. What's the advantage of using that? I am not quite following. Also, my current test harness doesn't have a good spot to run the client code while guaranteeing that it would trigger the breakpoint after the debugger is fully attached and ready to process the breakpoint.

aarzilli (commented):
> What's the advantage of using that?

You could test that detach without kill works, but on second thought this is fine.
It seems that dap tests are getting stuck on freebsd, I don't know if it's your fault or a preexisting problem in the freebsd backend, if it's the latter you can disable this new test on freebsd.

polinasok:
> It seems that dap tests are getting stuck on freebsd, I don't know if it's your fault or a preexisting problem in the freebsd backend, if it's the latter you can disable this new test on freebsd.

How can you tell? I can't seem to make sense of this new output format :( Also, the output is not verbose and doesn't have RUN + PASS/FAIL messages, so I can't even tell which test functions worked.

aarzilli:
> > It seems that dap tests are getting stuck on freebsd, I don't know if it's your fault or a preexisting problem in the freebsd backend, if it's the latter you can disable this new test on freebsd.
> 
> How can you tell? I can't seem to make sense of this new output format :( Also, the output is not verbose and doesn't have RUN + PASS/FAIL messages, so I can't even tell which test functions 

You can force the verbose output by temporarily changing _scripts/make.go.

aarzilli (approved):
LGTM

PR #2264, opened 2020-12-12
service/api: Expose WaitSince and WaitReason fields for goroutines

dlsniper:
This adds the WaitSince and WaitReason fields for the goroutines to allow the users to easily understand why a goroutine is waiting.


aarzilli (approved):
LGTM

PR #2266, opened 2020-12-14
proc: update a comment

andreimatei:
The comment on the advanceRegs() method was referencing several
nonexistent fields. This patch fixes that, and improves the comment to
reference the peculiar interaction between the method and it.regs.

### DIFF
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -407,9 +407,15 @@ func (it *stackIterator) appendInlineCalls(frames []Stackframe, frame Stackframe
 	return append(frames, frame)
 }
 
-// advanceRegs calculates it.callFrameRegs using it.regs and the frame
-// descriptor entry for the current stack frame.
-// it.regs.CallFrameCFA is updated.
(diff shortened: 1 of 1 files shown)


andreimatei:
Side-note to this patch: the way in which `stackIterator.advanceRegs()` updates the `it.regs.CFA`, but nothing else in `it.regs` feels wrong to me, but I'm not sure. I've noticed that the new value of `it.regs.CFA` is used by `newStackFrame() -> stackIterator.frameBase()`. Perhaps that `FrameBase` computation should move into `advanceRegs()`, and the `advanceRegs` wouldn't need to modify `it.regs` at all?
Generally I'd be grateful for any comments you can give me around the dance involved in computing the registers for the next frame based on the registers of the current frame.

aarzilli (approved):
The rules in the frame descriptor can access the value of CFA, so AFAIK it needs to be updated first.

derekparker:
I could see the possibility of calculating `it.regs.FrameBase` within `it.advanceRegs` at the very end once we've executed all the frame reg rules, but it would take a little bit of refactroring.

PR #2270, opened 2020-12-16
terminal,service: print WaitReason, WaitSince for goroutines

### LINKED ISSUE #637: Add wait reason to output of parked goroutines
For any sleeping goroutine we should be able to get the wait reason as to why they are parked. We should display this information in the `goroutines` output.

aarzilli:
Fixes #637

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -322,9 +322,7 @@ func TestRedirect(t *testing.T) {
 
 	// and detach from and kill the headless instance
 	client := rpc2.NewClient(listenAddr)
-	if err := client.Detach(true); err != nil {
-		t.Fatalf("error detaching from headless instance: %v", err)
-	}
+	_ = client.Detach(true)
 	cmd.Wait()
 }
 
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	"github.com/cosiner/argv"
 	"github.com/go-delve/delve/pkg/locspec"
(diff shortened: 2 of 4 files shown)


cmd/dlv/dlv_test.go
```diff
@@ -322,9 +322,7 @@ func TestRedirect(t *testing.T) {
 
 	// and detach from and kill the headless instance
 	client := rpc2.NewClient(listenAddr)
-	if err := client.Detach(true); err != nil {
```
derekparker, line 325: I'm confused why this is included in this patch.

pkg/terminal/command.go
```diff
@@ -917,11 +918,58 @@ func (t *Term) formatGoroutine(g *api.Goroutine, fgl formatGoroutineLoc) string
 		locname = "Start"
 		loc = g.StartLoc
 	}
-	thread := ""
+
+	buf := new(bytes.Buffer)
```
derekparker, line 922: We should use `strings.Builder` here. It's slightly more efficient in the `String` method, and shows intent a little better.

https://golang.org/pkg/strings/#Builder

pkg/terminal/command.go
```diff
@@ -917,11 +918,58 @@ func (t *Term) formatGoroutine(g *api.Goroutine, fgl formatGoroutineLoc) string
...
+		if g.WaitSince > 0 {
+			fmt.Fprintf(buf, " %s", time.Since(time.Unix(0, g.WaitSince)).String())
+		}
+		fmt.Fprintf(buf, "]")
+	}
+
+	return buf.String()
+}
+
+var waitReasonStrings = [...]string{
```
derekparker, line 945: Is there ever a case we'd want this string outside of just the terminal? For example, returning wait reason as a string via API calls?

derekparker (changes requested):

pkg/terminal/command.go
dlsniper: As a potential consumer of the API (and the one that added support initially for this), I second @aarzilli's message.

derekparker: Ack, makes sense.

derekparker (approved):

PR #2271, opened 2020-12-16
proc/*: implement proc.(*compositeMemory).WriteMemory

aarzilli:
Delve represents registerized variables (fully or partially) using
compositeMemory, implementing proc.(*compositeMemory).WriteMemory is
necessary to make SetVariable and function calls work when Go will
switch to using the register calling convention in 1.17.

This commit also makes some refactoring by moving the code that
converts between register numbers and register names out of pkg/proc
into a different package.

### DIFF
--- a/pkg/proc/arm64_arch.go
+++ b/pkg/proc/arm64_arch.go
@@ -8,14 +8,8 @@ import (
 
 	"github.com/go-delve/delve/pkg/dwarf/frame"
 	"github.com/go-delve/delve/pkg/dwarf/op"
+	"github.com/go-delve/delve/pkg/dwarf/regnum"
 	"golang.org/x/arch/arm64/arm64asm"
-)
-
-const (
-	arm64DwarfIPRegNum uint64 = 32
-	arm64DwarfSPRegNum uint64 = 31
-	arm64DwarfLRRegNum uint64 = 30
-	arm64DwarfBPRegNum uint64 = 29
 )
 
 var arm64BreakInstruction = []byte{0x0, 0x0, 0x20, 0xd4}
@@ -40,6 +34,8 @@ func ARM64Arch(goos string) *Arch {
 		inhibitStepInto:                  func(*BinaryInfo, uint64) bool { return false },
 		asmDecode:                        arm64AsmDecode,
 		usesLR:                           true,
+		PCRegNum:                         regnum.ARM64_PC,
+		SPRegNum:                         regnum.ARM64_SP,
 	}
 }
 
@@ -69,24 +65,24 @@ func arm64FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *Binary
 		// here).
 
 		return &frame.FrameContext{
-			RetAddrReg: arm64DwarfIPRegNum,
+			RetAddrReg: regnum.ARM64_PC,
 			Regs: map[uint64]frame.DWRule{
-				arm64DwarfIPRegNum: frame.DWRule{
+				regnum.ARM64_PC: frame.DWRule{
 					Rule:   frame.RuleOffset,
 					Offset: int64(-a.PtrSize()),
 				},
-				arm64DwarfBPRegNum: frame.DWRule{
+				regnum.ARM64_BP: frame.DWRule{
 					Rule:   frame.RuleOffset,
 					Offset: int64(-2 * a.PtrSize()),
 				},
-				arm64DwarfSPRegNum: frame.DWRule{
+				regnum.ARM64_SP: frame.DWRule{
 					Rule:   frame.RuleValOffset,
 					Offset: 0,
 				},
 			},
 			CFA: frame.DWRule{
 				Rule:   frame.RuleCFA,
-				Reg:    arm64DwarfBPRegNum,
+				Reg:    regnum.ARM64_BP,
 				Offset: int64(2 * a.PtrSize()),
 			},
 		}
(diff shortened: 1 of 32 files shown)


aarzilli:
Since this isn't urgent, I'd sit on this until we can run the full test matrix on it.

derekparker:
I think I'm going to hold off review and merge of this until after we cut `v1.6` unless you have any objections to that @aarzilli.

aarzilli:
> I think I'm going to hold off review and merge of this until after we cut `v1.6` unless you have any objections to that @aarzilli.

No objections.

pkg/proc/arm64_arch.go
```diff
@@ -8,14 +8,15 @@ import (
...
+	"github.com/go-delve/delve/pkg/dwarf/regnum"
 	"golang.org/x/arch/arm64/arm64asm"
 )
 
 const (
-	arm64DwarfIPRegNum uint64 = 32
-	arm64DwarfSPRegNum uint64 = 31
-	arm64DwarfLRRegNum uint64 = 30
-	arm64DwarfBPRegNum uint64 = 29
+	arm64DwarfIPRegNum uint64 = regnum.ARM64_PC
```
derekparker, line 16: I think we can simply drop these constants now.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -2190,30 +2185,29 @@ func (t *gdbThread) SetPC(pc uint64) error {
 	return t.p.conn.writeRegister(t.strID, reg.regnum, reg.value)
 }
 
-// SetSP will set the value of the SP register to the given value.
-func (t *gdbThread) SetSP(sp uint64) error {
-	_, _ = t.Registers() // Registes must be loaded first
-	t.regs.setSP(sp)
-	if t.p.gcmdok {
-		return t.p.conn.writeRegisters(t.strID, t.regs.buf)
+// ChangeRegs will change the value of a list of registers
```
derekparker, line 2188: Comment is incorrect.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -2190,30 +2185,29 @@ func (t *gdbThread) SetPC(pc uint64) error {
...
+func (t *gdbThread) SetReg(regNum uint64, reg *op.DwarfRegister) error {
+	var regName string
+	switch t.p.bi.Arch.Name {
```
derekparker, line 2191: Perhaps `Arch` should have a function pointer for translating, like: `Arch.RegisterNumberToName` which behind the scenes will have been set to the correct function from the `regnum` package.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -2190,30 +2185,29 @@ func (t *gdbThread) SetPC(pc uint64) error {
...
-	return t.p.conn.writeRegister(t.strID, reg.regnum, reg.value)
-}
-
-// SetDX will set the value of the DX register to the given value.
-func (t *gdbThread) SetDX(dx uint64) error {
-	if t.p.bi.Arch.Name == "arm64" {
-		return fmt.Errorf("not supported")
+	_, _ = t.Registers() // Registers must be loaded first
+	gdbreg, ok := t.regs.regs[regName]
+	if !ok && strings.HasPrefix(regName, "xmm") {
```
derekparker, line 2199: Can you add a comment around this change with some context.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -2190,30 +2185,29 @@ func (t *gdbThread) SetPC(pc uint64) error {
...
+		gdbreg, ok = t.regs.regs["y"+regName[1:]]
 	}
-
-	_, _ = t.Registers() // Registes must be loaded first
-	t.regs.setDX(dx)
-	if t.p.gcmdok {
-		return t.p.conn.writeRegisters(t.strID, t.regs.buf)
+	if !ok {
+		return fmt.Errorf("could not set regster %s: not found", regName)
```
derekparker, line 2203: Typo in register.

pkg/proc/i386_arch.go
```diff
@@ -3,15 +3,17 @@ package proc
...
 	"github.com/go-delve/delve/pkg/dwarf/op"
-	"strings"
+	"github.com/go-delve/delve/pkg/dwarf/regnum"
 )
 
 const (
-	i386DwarfIPRegNum uint64 = 8
-	i386DwarfSPRegNum uint64 = 4
-	i386DwarfBPRegNum uint64 = 5
+	i386DwarfIPRegNum uint64 = regnum.I386_Eip
```
derekparker, line 14: Same comment as before but I think we can remove these constants now.

pkg/proc/linutil/regs_amd64_arch.go
```diff
@@ -307,3 +311,71 @@ func (r *AMD64Registers) Copy() (proc.Registers, error) {
...
+		r.loadFpRegs = nil
+	}
+
+	if regNum < regnum.AMD64_XMM0 || regNum > regnum.AMD64_XMM0+15 {
+		return false, fmt.Errorf("can not set %s", regnum.AMD64ToName(regNum))
+	}
+
+	reg.FillBytes()
+
+	err := r.Fpregset.SetXmmRegister(int(regNum-regnum.AMD64_XMM0), reg.Bytes)
```
derekparker, line 376: Would it make more sense to do this arithmetic within `SetXmmRegister` instead of relying on the caller to do it? (In reference to `regNum-regnum.AMD64_XMM0`).

pkg/proc/mem.go
```diff
@@ -129,8 +130,41 @@ func (mem *compositeMemory) ReadMemory(data []byte, addr uint64) (int, error) {
...
 
 func (mem *compositeMemory) WriteMemory(addr uint64, data []byte) (int, error) {
-	//TODO(aarzilli): implement
-	return 0, errors.New("can't write composite memory")
+	addr -= fakeAddress
+	if addr >= uint64(len(mem.data)) || addr+uint64(len(data)) > uint64(len(mem.data)) {
+		return 0, errors.New("write out of bounds")
+	}
+	if mem.regs.ChangeFunc == nil {
+		return 0, errors.New("can not write registers")
```
derekparker, line 138: Will this error ever propagate to the user? If so we should make it a little friendlier and explanatory.

derekparker: Additionally, does the absence of a `ChangeFunc` indicate that register writing is not supported by the current debug mode, e.g. debugging core files? Or should we panic here to indicate a programmer error that we never properly set this value at startup in some area of the code.

derekparker (changes requested):
A few questions / nits. I really like the new `regnum` package, cleans things up a lot.

aarzilli (commented):
For some reason teamcity isn't picking this PR up so I ran a build manually: https://delve.beta.teamcity.com/viewLog.html?buildId=2667&tab=buildResultsDiv&buildTypeId=Delve_AggregatorBuild

pkg/proc/linutil/regs_amd64_arch.go
derekparker: Ack, your rationale makes sense.

pkg/proc/gdbserial/gdbserver.go
derekparker: LGTM

derekparker (approved):
LGTM

PR #2273, opened 2020-12-18
Add example of passing multiple arguments to `--build-flags=`

zamai:
This simple example removes confusion when user is trying to pass multiple arguments to the compiler via `--build-flags=`


aarzilli:
The documentation is autogenerated, it needs to be changed in cmd/dlv/cmds/commands.go. Aside from that I'm weakly inclined to reject this because it doesn't have anything to do with delve, this is a general shell behavior.

zamai:
@aarzilli I've updated the string in the `cmd/dlv/cmds/commands.go`, if its general shell behavior would more examples in documentation hurt it? I was stumbled on passing multiple arguments to compiler thought delve, I'm sure somebody will too in the future. This is a help note for them.

aarzilli:
I should have told you before, you need to `go run _scripts/gen-usage-docs.go` to regenerate the documentation.

zamai:
@aarzilli added generated documentation

derekparker:
I'm alright with adding some clarity to the documentation but might a suggest a slight rewrite. I think something like: 

`Build flags, to be passed to the compiler. For example: --build-flags="-tags=integration -mod=vendor -cover -v"`

It's just a smidge more succinct.

zamai:
@derekparker sure, updated the example

derekparker (approved):
LGTM

PR #2274, opened 2020-12-18
Miscellaneous fixes for problems discovered by Github Actions tests

aarzilli:
In Go 1.15 the call injection will be executed on a different goroutine
from the goroutine where it was started on to avoid confusing the
garbage collector, the test must be aware of this fact and use the
goroutine ID from the stopped response instead of assuming 1 is the
currently selected goroutine.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -1462,17 +1462,12 @@ func TestEvaluateCallRequest(t *testing.T) {
 					// and not the return to the original point of injection after it
 					// completed.
 					client.EvaluateRequest("call makeclos(nil)", 1000, "not watch")
-					client.ExpectStoppedEvent(t)
+					stopped := client.ExpectStoppedEvent(t)
 					erres = client.ExpectVisibleErrorResponse(t)
 					if erres.Body.Error.Format != "Unable to evaluate expression: call stopped" {
 						t.Errorf("\ngot %#v\nwant Format=\"Unable to evaluate expression: call stopped\"", erres)
 					}
-					if (goversion.VersionAfterOrEqual(runtime.Version(), 1, 15) && (runtime.GOOS == "linux" || runtime.GOOS == "windows")) ||
-						runtime.GOOS == "freebsd" {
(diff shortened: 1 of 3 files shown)


aarzilli:
cc @polinasok see the change to TestEvaluateCallRequest.

service/dap/server_test.go
```diff
@@ -1462,17 +1462,12 @@ func TestEvaluateCallRequest(t *testing.T) {
...
 					// completed.
 					client.EvaluateRequest("call makeclos(nil)", 1000, "not watch")
-					client.ExpectStoppedEvent(t)
+					stopped := client.ExpectStoppedEvent(t)
 					erres = client.ExpectVisibleErrorResponse(t)
 					if erres.Body.Error.Format != "Unable to evaluate expression: call stopped" {
 						t.Errorf("\ngot %#v\nwant Format=\"Unable to evaluate expression: call stopped\"", erres)
 					}
-					if (goversion.VersionAfterOrEqual(runtime.Version(), 1, 15) && (runtime.GOOS == "linux" || runtime.GOOS == "windows")) ||
-						runtime.GOOS == "freebsd" {
```
polinasok, line 1471: The goroutine id change makes sense. 
But what about the differences in stack frames that I observed on different systems?

aarzilli: On Go 1.15 the topmost frame on the original goroutine will always be `runtime.debugCallWrap`, including on darwin, which didn't report an error before because on travis function calls weren't tested.

polinasok: Ah, got it. That must be the frame for launching the separate call goroutine. I added that clause because I observed some test failures on Travis (everything was working on my Mac locally), but with your fix and explanation this makes way more sense now!

polinasok (approved):

derekparker (approved):
LGTM

derekparker:
@aarzilli do you mind rebasing your other PRs onto main now that it's merged, to fix all the PRs being marked as having failed CI.

aarzilli:
Done, there are a couple of them that are failing, I'll look into them.

PR #2276, opened 2020-12-19
proc/native: fix linux/386 native backend

aarzilli:
Change 24ec175 (use CPUID to determine maximum size of XSAVE area)
broke the linux/386 backend, this commit fixes it.


derekparker (approved):
LGTM

PR #2278, opened 2020-12-24
terminal: add way to cancel goroutines command with ctrl-C

aarzilli:
The goroutines command can take a long time to complete if there are
many goroutines, add the possibility to terminate it early by pressing
ctrl-C.


derekparker (approved):
LGTM

PR #2279, opened 2020-12-24
pkg/proc/native: fix nil pointer dereference in linux backend

aarzilli:


pkg/proc/native/proc_linux.go
```diff
@@ -591,9 +591,10 @@ func (dbp *nativeProcess) stop(trapthread *nativeThread) (*nativeThread, error)
 	}
 
 	if switchTrapthread {
+		trapthreadID := trapthread.ID
```
derekparker, line 594: I assume that `trapthread` was the nil pointer, so then how does this fix it?

derekparker (approved):
LGTM

PR #2280, opened 2020-12-24
native/proc/linux: wait for the target process to be killed in kill

aarzilli:
Waits for the target process to have received the kill signal in
native.nativeProcess.kill. Fixes an infrequent error in TestKill.


derekparker (approved):
LGTM

PR #2285, opened 2020-12-31
proc/gdbserial: Added support for darwin/arm64 using gdbserver

### LINKED ISSUE #2246: Apple M1 ARM support

oxisto:
This is an alternative fix for #2246 and one that might be easier to merge in than fixing the native darwin backend (see #2254).

Tests almost succeed, and it looks like those that do not, i.e. cgo-related ones, might be related to some Go 1.16 problems. I can provide a complete log of `make test` (it is currently running).

Fixes #2246

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -77,6 +77,7 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/arch/arm64/arm64asm"
 	"golang.org/x/arch/x86/x86asm"
 
 	"github.com/go-delve/delve/pkg/logflags"
@@ -292,6 +293,13 @@ func (p *gdbProcess) Connect(conn net.Conn, path string, pid int, debugInfoDirs 
 					p.gcmdok = false
 					break
 				}
+
+				// Workaround for darwin arm64. Apple's debugserver seems to have a problem
+				// with not selecting the correct thread in the 'g' command and the returned
+				// registers are empty / an E74 error is thrown.
+				if v[len("version:"):] == "1200" && p.bi.Arch.Name == "arm64" {
+					p.gcmdok = false
+				}
 			}
 		}
 	}
@@ -301,17 +309,19 @@ func (p *gdbProcess) Connect(conn net.Conn, path string, pid int, debugInfoDirs 
 		return nil, err
 	}
 
-	// None of the stubs we support returns the value of fs_base or gs_base
-	// along with the registers, therefore we have to resort to executing a MOV
-	// instruction on the inferior to find out where the G struct of a given
-	// thread is located.
-	// Here we try to allocate some memory on the inferior which we will use to
-	// store the MOV instruction.
-	// If the stub doesn't support memory allocation reloadRegisters will
-	// overwrite some existing memory to store the MOV.
-	if addr, err := p.conn.allocMemory(256); err == nil {
-		if _, err := p.conn.writeMemory(addr, p.loadGInstr()); err == nil {
-			p.loadGInstrAddr = addr
+	if runtime.GOARCH != "arm64" {
+		// None of the stubs we support returns the value of fs_base or gs_base
+		// along with the registers, therefore we have to resort to executing a MOV
+		// instruction on the inferior to find out where the G struct of a given
+		// thread is located.
+		// Here we try to allocate some memory on the inferior which we will use to
+		// store the MOV instruction.
+		// If the stub doesn't support memory allocation reloadRegisters will
+		// overwrite some existing memory to store the MOV.
+		if addr, err := p.conn.allocMemory(256); err == nil {
+			if _, err := p.conn.writeMemory(addr, p.loadGInstr()); err == nil {
+				p.loadGInstrAddr = addr
+			}
 		}
 	}
 
@@ -557,12 +567,29 @@ func LLDBAttach(pid int, path string, debugInfoDirs []string) (*proc.Target, err
 // debugging PIEs.
 func (p *gdbProcess) EntryPoint() (uint64, error) {
 	var entryPoint uint64
-	if auxv, err := p.conn.readAuxv(); err == nil {
-		// If we can't read the auxiliary vector it just means it's not supported
-		// by the OS or by the stub. If we are debugging a PIE and the entry point
-		// is needed proc.LoadBinaryInfo will complain about it.
-		entryPoint = linutil.EntryPointFromAuxv(auxv, p.BinInfo().Arch.PtrSize())
+	if runtime.GOOS == "darwin" {
+		// There is no auxv on darwin, however, we can get the location of the mach-o
+		// header from the debugserver by going through the loaded libraries, which includes
+		// the exe itself
+		images, _ := p.conn.getLoadedDynamicLibraries()
+		for _, image := range images {
+			if image.MachHeader.FileType == macho.TypeExec {
+				// This is a bit hacky. This is technically not the entrypoint,
+				// but rather we use the variable to points at the mach-o header,
+				// so we can get the offset in bininfo
+				entryPoint = image.LoadAddress
+				break
+			}
+		}
+	} else {
+		if auxv, err := p.conn.readAuxv(); err == nil {
+			// If we can't read the auxiliary vector it just means it's not supported
+			// by the OS or by the stub. If we are debugging a PIE and the entry point
+			// is needed proc.LoadBinaryInfo will complain about it.
+			entryPoint = linutil.EntryPointFromAuxv(auxv, p.BinInfo().Arch.PtrSize())
+		}
 	}
+
 	return entryPoint, nil
 }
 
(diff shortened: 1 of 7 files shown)


oxisto:
Here is the log of running the tests on my M1 machine: [test.log](https://github.com/go-delve/delve/files/5757813/test.log)

But, somehow it looks like I broke `darwin/amd64`, meh.

pkg/proc/gdbserial/gdbserver_conn_arm64.go
```diff
@@ -0,0 +1,16 @@
+//+build arm64
+
+package gdbserial
+
+const (
+	regnamePC = "pc"
```
oxisto, line 6: It would probably be a good idea to define those constants somewhere else, since multiple backends might need to refer to them.

oxisto:
Existing tests should now all run flawlessly. Ready to review and open for suggestions how to improve it :)

pkg/proc/native/support_sentinel.go
```diff
@@ -1,5 +1,5 @@
 // This file is used to detect build on unsupported GOOS/GOARCH combinations.
 
-//+build !linux,!darwin,!windows,!freebsd linux,!amd64,!arm64,!386 darwin,!amd64 windows,!amd64 freebsd,!amd64
+//+build !linux,!darwin,!windows,!freebsd linux,!amd64,!arm64,!amd64 windows,!amd64 freebsd,!amd64
```
aarzilli, line 3: I think this is wrong. It should be:

```
//+build !linux,!darwin,!windows,!freebsd linux,!amd64,!arm64,!386 darwin,!amd64,!arm64 windows,!amd64 freebsd,!amd64
```

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1926,6 +2179,10 @@ func (t *gdbThread) SetSP(sp uint64) error {
 
 // SetDX will set the value of the DX register to the given value.
 func (t *gdbThread) SetDX(dx uint64) error {
+	if runtime.GOARCH == "arm64" {
```
aarzilli, line 2182: Wherever you use runtime.GOARCH can you use `t.p.bi.Arch.Name`/`p.bi.Arch.Name` instead?

aarzilli (commented):
Overall LGTM, just a couple of minor things, glad to see you made it work!

pkg/proc/gdbserial/gdbserver.go
oxisto: Yeah sure, makes sense to use the target architecture instead of the runtime.

I am wondering whether we should consider some cornercases where one might need to debug a x86 application on an arm64 System - which theoretically should be possible, since lldb also manages to do that. I can keep investigating this once the initial PR is done

aarzilli: Ideally those regnumPC constants would be member fields of gdbProcess (or gdbConn) and setup correctly on initialization, so it wouldn't need conditional compilation and (should) work across architectures. But we don't need to do this here.

oxisto: > Wherever you use runtime.GOARCH can you use `t.p.bi.Arch.Name`/`p.bi.Arch.Name` instead?

I cannot access the process in the `gdbRegisters` functions :( Such as `func (regs *gdbRegisters) TLS() uint64` any ideas?

aarzilli: `gdbRegistes.setDX` is only called by `gdbThread.SetDX` so it can be done there. For TLS and GAddr instead of changing those functions you can change the final part of reloadRegisters to set `t.regs.tls`, `t.regs.hasgaddr` and `t.regs.gaddr` appropriately. That leaves `Get` for which there is no easy solution, you can change gdbRegisters.init to take a `goarch string` parameter and save it as a field of gdbRegisters and use that in Get. BTW do you mind also changing the constants in Get to this:

```
const (
    mask8 = 0xff
    mask16 = 0xffff
    mask32 = 0xffffffff
)
```

Thanks.

oxisto: I have added a *proc.Arch field to the gdbRegisters and the init function - or would you prefer a string field? TLS and GAddr are set in reloadRegisters now as you suggested

aarzilli: > I have added a *proc.Arch field to the gdbRegisters and the init function - or would you prefer a string field?

either way is fine.

oxisto: > Ideally those regnumPC constants would be member fields of gdbProcess (or gdbConn) and setup correctly on initialization, so it wouldn't need conditional compilation and (should) work across architectures. But we don't need to do this here.

I tried that quickly, but then I am running into the same issue as above, with the `gdbRegisters` not able to access them if they are either in the process or in the connection. I think the best way to handle this would actually be in in `proc.Arch` since this seems the obvious choice to know what the "special" registers are on each platform. Maybe this even something that DWARF can solve? I am not an expert on this. But this probably then goes beyond this PR - it is however needed if we can to be able to debug translated rosetta2 binaries.

aarzilli: I'm going to make a few changes to how we manipulate registers to support writing to variables stored in registers (#2271), it's possible that it will help, however some things here are a bit specific to gdb. I agree that it should be postponed.

aarzilli (approved):
LGTM

ebilling:
I hate to ask, and I know this isn't the right place, but @oxisto, I can't get this to work on my setup.  I've been pulling down every version you push to try them out.

Darwin_arm64 laptop,
go version devel +a508840c67 Tue Dec 15 16:31:07 2020 +0000 darwin/amd64
% dlv version
Delve Debugger
Version: 1.5.1
Build: 15456ea98f1ca3fb8379730ff46ae8e08a054ef2

When I try to debug something using Microsoft Code, I get: 

`could not launch process: stub exited while waiting for connection: exit status 0`

This works on my old laptop just fine, so has to be part of the arm move.  Have you run into this?  Are there additional instructions required in the Readme.md?

aarzilli:
Could be that you are using a version of go compiled for amd64.

oxisto:
> Could be that you are using a version of go compiled for amd64.

Yes, exactly. This PR adds support for `darwin/arm64` binaries using a `darwin/arm64` version of Go. All other combinations, such as using a `amd64` go binary or to debug a `amd64` binary using Rosetta 2 is not supported. Once this PR is done, I will have a look at what other options delve can support.

To test it, you can get the Go 1.16 beta that supports `darwin/am64` from the Go downloads page: https://golang.org/dl/go1.16beta1.darwin-arm64.pkg. You should then be able to `make install` delve using this fork/branch. Afterwards, using `dlv debug path-to-your-program.go` should be able to debug your program.

You can even test it in Visual Studio code, but you probably need to download the *very experimental* arm64 builds of VSCode for that (https://code.visualstudio.com/insiders/# / https://code.visualstudio.com/sha/download?build=insider&os=darwin-arm64). I would not recommend this for any productive workflow yet.

ebilling:
Thank you both.  I had built the darwin/arm64 version of go, but vs code was still pointing to a different version, and not paying attention to my PATH.  I appreciate the pointers @oxisto and @aarzilli. Thank you.

gitslav:
@oxisto I think you forgot to bump the version number in compat.go
https://github.com/oxisto/delve/blob/15456ea98f1ca3fb8379730ff46ae8e08a054ef2/pkg/goversion/compat.go#L11

aarzilli:
> @oxisto I think you forgot to bump up the version number in compat.go
> https://github.com/oxisto/delve/blob/15456ea98f1ca3fb8379730ff46ae8e08a054ef2/pkg/goversion/compat.go#L11

That's going to be done by #2214

artspb:
If you want to try it in GoLand, you can specify `-Ddlv.path=/Users/artspb/path/to/arm64/dlv` under _Help | Edit Custom VM Options..._

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1742,157 +1780,369 @@ func (regs *gdbRegisters) byName(name string) uint64 {
 }
 
 func (regs *gdbRegisters) Get(n int) (uint64, error) {
-	reg := x86asm.Reg(n)
 	const (
-		mask8  = 0x000f
-		mask16 = 0x00ff
-		mask32 = 0xffff
+		mask8  = 0xff
```
derekparker, line 1784: This has been fixed elsewhere, but we really should export these somewhere for consistency.

derekparker: To be clear, I don't think this should happen in this patch, just making a note.

derekparker (approved):
LGTM

pkg/proc/gdbserial/gdbserver.go
aarzilli: It makes more sense to do it after the compositeMemory.WriteMemory PR, since that changes some other things about registers, we should just go through op.DwarfRegisters.

zsyu9779:
when I try to  **go get github.com/go-delve/delve/cmd/dlv**
then   ../../service/debugger/debugger.go:25:2: found packages native (nonative_darwin.go) and your_operating_system_and_architecture_combination_is_not_supported_by_delve (support_sentinel.go) in /Users/admin/go/src/delve/pkg/proc/native

shortcircuit3:
I get this error when debugging

```
Failed to continue - runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation]
Unable to propogate EXC_BAD_ACCESS signal to target process and panic (see https://github.com/go-delve/delve/issues/852)
Last known immediate stacktrace (goroutine id 0):
	?:-1
		<unknown>
	:0
		<unknown>
```

xibolun:
> runtime/cgo gcc_arm64.S

check your go IDE is arm version?  or you use another go version to run code?

dragonly:
I still run into trouble debugging TiDB.

https://github.com/go-delve/delve/issues/2246#issuecomment-793464029

PR #2286, opened 2021-01-01
pkg/prog/native/linux: change handling of ESRCH in trapWaitInternal

aarzilli:
If a thread exits while we are looking at it just treat it as if the
status had reported it as exited instead of doing something special.

Fixes flakiness in TestIssue387.


derekparker (approved):
LGTM

Issue #2284, opened 2020-12-30
Could not find abstract origin & use of closed network


exapsy:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

---

1. What version of Delve are you using (`dlv version`)?

**Docker container:**

```
Delve Debugger
Version: 1.5.1
Build: $Id: bca418ea7ae2a4dcda985e623625da727d4525b5
```

**Desktop:**
```
Delve Debugger
Version: 1.5.0
Build: $Id: ca5318932770ca063fc9885b4764c30bfaf8a199
```

---

2. What version of Go are you using? (`go version`)?

**Docker container:**

```
go version go1.15.6 linux/amd64
```

**Desktop:**

```
go version go1.15.5 linux/amd64
```

---

3. What operating system and processor architecture are you using?

**Docker container:**

```
golang:1.15.6-alpine3.12
```

**Desktop:**

```
5.7.19-2-MANJARO
```

---

4. What did you do?

I'm running this on a docker container to run a headless delve server on top of my Go web server

```
dlv exec /app/tmp/service --continue --accept-multiclient --listen=:40000 --headless=true --api-version=2 --log --build-flags='-mod=vendor'
```

And I'm connecting to it via this configuration in VSCode in `launch.json`

```json
    {
      "name": "Golang delve docker",
      "type": "go",
      "request": "attach",
      "mode": "remote",
      "apiVersion": 2,
      "trace": "verbose",
      "port": 40000,
      "host": "localhost"
    },
```

---

5. What did you expect to see?

No error

---

6. What did you see instead?

**When I restart the server**

```
app_1            | 2020-12-30T12:43:42Z warning layer=debugger reading debug_info: could not find abstract origin (0x391128) of inlined call at 0x31a4e7
app_1            | 2020-12-30T12:43:42Z warning layer=debugger reading debug_info: could not find abstract origin (0x391128) of inlined call at 0x31a502
app_1            | 2020-12-30T12:43:42Z warning layer=debugger reading debug_info: could not find abstract origin (0x391128) of inlined call at 0x31a51d
app_1            | 2020-12-30T12:43:42Z warning layer=debugger reading debug_info: could not find abstract origin (0x391128) of inlined call at 0x31a538
app_1            | 2020-12-30T12:43:42Z warning layer=debugger reading debug_info: could not find abstract origin (0x391128) of inlined call at 0x31a96f
[...]

PR #2293, opened 2021-01-09
proc: do not assume abstract origins precede their uses

aarzilli:
The DWARF standard does not say that a DW_ATTR_abstract_origin can only
reference entries that appear before it, this change fixes BinaryInfo
to comply. See #2284 for an example of this happening.


derekparker (approved):
LGTM

PR #2298, opened 2021-01-13
Introduce TeamCity builds

artspb:
This pool request introduces the TeamCity build configuration (https://delve.beta.teamcity.com/). Additionally, it fixes two problems that prevent tests from running. One is relevant for darwin/386 (`restore gdbserial constants for 386`), another is for windows/amd64 (`skip TestAttachRequest on Windows as it never finishes`). I didn't investigate these problems. Instead, I've just implemented workarounds so that tests can run. Feel free to drop these commits if you feel that they bring more harm than good.

If you decide to accept this pool request, you'll need to additionally update the TeamCity configuration, so the server points at the original repo instead of my fork. It should be a rather trivial task. I've sent invites to Alessandro and Derek and will grant admin access as soon as they register.

Currently, tests are run against the following sets (see `settings.kts:36`).

```
"linux/386/1.15.2",
"linux/386/1.16beta1",
"linux/386/latest",
"linux/amd64/1.15.3",
"linux/amd64/1.16beta1",
"linux/amd64/latest",
"windows/amd64/1.15.3",
"windows/amd64/1.16beta1",
"windows/amd64/latest"
```

Tests on Linux are run in Docker (Ubuntu 20.04 LTS) on both 386/amd64. Tests on Windows are only available for amd64. At the moment, you can only add more Go versions to these sets. Additional machines (linux/arm64, darwin/amd64, and darwin/arm64). Are on the way. We'll create agents for them as soon as they arrive.

Builds should run not only against the master branch but also against pool requests. I've enabled this feature but didn't have a chance to test it. You'll need to update the GitHub token for it to work.

And last but not least. You can use both Kotlin DSL and TeamCity UI to configure builds. In either case, the configuration will end up in the repository under `.teamcity`. When using the UI, patches are created that don't look nice and hard to maintain. So, in general, it's recommended to stick to Kotlin or at least merge these patches into the main configuration file from time to time. In any case, we (JetBrains) are always ready to help with any related questions or problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ cmd/dlv/dlv
 .vagrant
 **/*.swp
 localtests
+.idea
+*.iml
+.teamcity/target
--- a/.teamcity/pom.xml
+++ b/.teamcity/pom.xml
@@ -1,0 +1,104 @@
+<?xml version="1.0"?>
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <name>Delve Config DSL Script</name>
+  <groupId>Delve</groupId>
+  <artifactId>Delve_dsl</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <parent>
+    <groupId>org.jetbrains.teamcity</groupId>
+    <artifactId>configs-dsl-kotlin-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+
+  <repositories>
+    <repository>
+      <id>jetbrains-all</id>
+      <url>https://download.jetbrains.com/teamcity-repository</url>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+    <repository>
+      <id>teamcity-server</id>
+      <url>https://delve.beta.teamcity.com/app/dsl-plugins-repository</url>
+      <snapshots>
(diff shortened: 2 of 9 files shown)


karanagai:
It is very much advised to set Versioned settings options to "always use current settings" rather than "use settings from VCS" for this project in TeamCity. Otherwise anyone will be able to submit any DSL code as a pull request and affect the configuration of the resulting build exploiting it in one way or another.

artspb:
@karanagai That's how it's configured now. Thanks!

aarzilli:
> One is relevant for darwin/386

We've never supported darwin on 386, only amd64 and, recently, arm64.

aarzilli (approved):
LGTM, thank you!

artspb:
>>One is relevant for darwin/386

>We've never supported darwin on 386, only amd64 and, recently, arm64.

Sorry, I misspelled it. I meant that it's relevant for any 386. I actually found it when tried to run linux/386 tests.

aarzilli:
> > > One is relevant for darwin/386
> 
> > We've never supported darwin on 386, only amd64 and, recently, arm64.
> 
> Sorry, I misspelled it. I meant that it's relevant for any 386. I actually found it when tried to run linux/386 tests.

Ok, I understand now.

PR #2309, opened 2021-01-19
fix/pmem: prettyprint outputs wrong starting address

hitzhangjie:
The starting address should be incremented by `address += uintptr(cols*colBytes)`.


aarzilli (approved):
LGTM

PR #2310, opened 2021-01-20
investigating test generated docs failure

aarzilli:


nd:
Debugged it locally: it looks like the fix should be

```
saved = bytes.ReplaceAll(saved, []byte{'\r'}, []byte{})
```

`generated` always uses \n's.

PR #2315, opened 2021-01-20
Change TeamCity test matrix and drop appveyor/travis-ci configurations

aarzilli:


aarzilli:
@nd do you know what would be the way to test something like this? I tried making a custom build but the settings change was rejected with this message: "Unsupported change of snapshot dependencies in the build configuration".

nd:
As far as I know there is no easy way to test a change like this in a separate branch. When build uses custom settings from version control TeamCity takes into account only settings for a build, like build script to run, but more global settings like dependencies between builds are ignored. More info [here](https://www.jetbrains.com/help/teamcity/storing-project-settings-in-version-control.html#Storing+Secure+Settings), search for 'will not affect the build'.

If you want to change it safely without breaking current builds, then perhaps the easiest way is to create a parallel build chain (a new Aggregator and tests), ensure that it works, and then replace the old one with the new. The changed build scripts will have to handle old and new builds, so maybe it is too much trouble and it is easier just to commit to master and tweak it until it works.

aarzilli:
@nd how bad would I screw things up if I changed the VCS root to aarzilli/delve temporarily, tested the changes and then set it back to go-delve/delve?

nd:
I think that would work. You can also change only the branch in the current VCS root to point to the refs/pull/2315/head. I'd say it should work and if it doesn't you can always change it back. 

The only thing is that it will affect how TeamCity computes changes in builds (what is shown on the 'Changes' tab). Some builds will have no changes, but they should clone the right revision for the build anyway.

aarzilli:
@nd thanks, changing the VCS root worked well.

@derekparker I've changed the test matrix to make it more like what we had before the travispocalypse, I've changed the scripts to automatically select the latest minor version for go (so we don't have to change them every time a minor release is made) and added a configuration that tests using tip, disabled by default, that we can use for the next-version support branch.

I've removed the configurations for appveyor and travis-ci, appveyor didn't work anyway, travis-ci didn't come back even tough it's been more than a month since we ran out of free credits.

derekparker (approved):
LGTM

PR #2320, opened 2021-01-23
proc: fix embedded field search

### LINKED ISSUE #2316: dlv promotes wrong method when embedding interface in struct
```
Delve Debugger
Version: 1.5.1
Build: $Id: bca418ea7ae2a4dcda985e623625da727d4525b5 $
```
repros on both 1.14.10 and 1.15.5
linux/amd64
4. What did you do?
Given a struct which embeds both another struct and an interface, and a field which exists on both the interface's dynamic type and the struct type, I printed the promoted field in Delve (`print v.F`).  Full details below.
I expected Delve to do the same as go, which is to print the promoted field of the struct type.
6. What did you see instead?
Instead, Delve printed the field of the interface type, as if it were promoted.

In more detail, I have the following code:
```go
package main

import "fmt"

type T struct {
	F string
}

type W1 struct {
	T
}

func (*W1) M() {}

type I interface{ M() }

var _ I = (*W1)(nil)

type W2 struct {
	I
	T
}

var _ I = (*W2)(nil)

func main() {
	v := &W2{I: &W1{T{"F within W1"}}, T: T{"F within W2"}}
	fmt.Printf("%#v\n", v)
	fmt.Println(&v.F, v.F)                     // prints "F within W2"
	fmt.Println(&v.T.F, v.T.F)                 // same as &v.F
	fmt.Println(&v.I.(*W1).T.F, v.I.(*W1).T.F) // different from &v.F
}
```

According to Go, the field `v.F` refers unambiguously to `v.T.F` -- `I` has no field `F` (it's an interface), and even if the field `v.I.(*W1).T.F` were available, it would be hidden by `v.T.F` which is at a shallower depth.  (The details are in the [Go spec](https://golang.org/ref/spec#Selectors).)  Indeed running the program gives:
```
&main.W2{I:(*main.W1)(0xc000010200), T:main.T{F:"F within W2"}}
0xc00000c070 F within W2
0xc00000c070 F within W2
0xc000010200 F within W1
```

Delve helpfully allows accessing the field of `W1` directly as `v.I.T.F`.  The bug is that it also incorrectly promotes this field to `v.F`.  In this case that's doubly wrong: embedding an interface never promotes fields, only methods included in the interface; and additionally `v.T.F` is shallower.  Specifically, with Delve stopped on the last line, we get:
```
(dlv) print &v.F
(*string)(0xc000010200)
(dlv) print v.F
"F within W1"
(dlv) print &v.T.F
(*string)(0xc00000c070)
(dlv) print v.T.F
"F within W2"
(dlv) print &v.I.T.F
(*string)(0xc000010200)
(dlv) print v.I.T.F
"F within W1"
```
The first pair of prints should match the second pair, like Go does, but instead match the third.

aarzilli:
Both structMember and findMethod implemented a depth-first search in
embedded fields but the Go specification requires a breadth-first
search. They also allowed promotion of fields in the concrete type of
embedded interfaces even though this is not allowed by Go.

Furthermore they both lacked protection from infinite recursion
when a type embeds itself and the user requests a non-existent field.

Fixes #2316

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -106,6 +106,38 @@ type List struct {
 	Next *List
 }
 
+type T struct {
+	F string
+}
+
+type W1 struct {
+	T
+}
+
+func (*W1) M() {}
+
+type I interface{ M() }
+
+type W2 struct {
+	W1
+	T
+}
+
+type W3 struct {
+	I
+	T
+}
+
+type W4 struct {
+	I
+}
+
+type W5 struct {
+	*W5
+}
+
+var _ I = (*W2)(nil)
+
 func main() {
 	i1 := 1
 	i2 := 2
@@ -312,6 +344,12 @@ func main() {
 	ll := &List{0, &List{1, &List{2, &List{3, &List{4, nil}}}}}
 	unread := (*int)(unsafe.Pointer(uintptr(12345)))
 
+	w2 := &W2{W1{T{"T-inside-W1"}}, T{"T-inside-W2"}}
+	w3 := &W3{&W1{T{"T-inside-W1"}}, T{"T-inside-W3"}}
+	w4 := &W4{&W1{T{"T-inside-W1"}}}
+	w5 := &W5{nil}
+	w5.W5 = w5
+
 	var amb1 = 1
 	runtime.Breakpoint()
 	for amb1 := 0; amb1 < 10; amb1++ {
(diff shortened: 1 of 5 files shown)


derekparker (approved):
LGTM

derekparker:
@aarzilli the TeamCity build is failing but the GH action builds are fine. I assume this is just an issue with TeamCity that would likely get solved by a rebase or something?

aarzilli:
> I assume this is just an issue with TeamCity that would likely get solved by a rebase or something?

Correct.

aarzilli:
Given that this is a higher risk change (a lot of code calls structMembe and findMethod) and that the bug it's fixing only affects corner cases, I think we should merge this after the release of 1.6. Go1.16 is coming out today anyway.

derekparker:
> Given that this is a higher risk change (a lot of code calls structMember and findMethod) and that the bug it's fixing only affects corner cases, I think we should merge this after the release of 1.6. Go1.16 is coming out today anyway.

Ack, agreed.

PR #2321, opened 2021-01-23
proc: remove dead code

aarzilli:

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -24,9 +24,6 @@ const (
 
 	maxArrayStridePrefetch = 1024 // Maximum size of array stride for which we will prefetch the array contents
 
-	chanRecv = "chan receive"
-	chanSend = "chan send"
-
 	hashTophashEmptyZero = 0 // used by map reading code, indicates an empty cell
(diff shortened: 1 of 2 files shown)


pkg/proc/variables.go
```diff
@@ -1063,59 +1060,61 @@ func (v *Variable) structMember(memberName string) (*Variable, error) {
...
-				continue
-			}
-			// Check for embedded field referenced by type name
-			parts := strings.Split(field.Name, ".")
-			if len(parts) > 1 && parts[1] == memberName {
+	queue := []*Variable{v}
+	seen := map[string]struct{}{} // prevent infinite loops
+	first := true
+
+	for len(queue) > 0 {
```
derekparker, line 1068: This changeset seems more like a refactor than dead code elimination. Would you mind sending this patch in separate?

derekparker: Ah, it seems you have. I'll take a look at the other PR first then, perhaps we can rebase this after.

aarzilli: Yep. I think I messed up a rebase, actually. Fixed it now.

derekparker (approved):
LGTM

PR #2327, opened 2021-01-26
Issue 2319

thanm:
Hi Delve maintainers -- here is a tentative fix for issue 2319. Let me know if you have questions.  I don't have a lot of familiarity with github PR workflow, so let me know if I missed anything from the usual process. Thanks-- Than


pkg/proc/proc_test.go
```diff
@@ -4979,3 +4979,29 @@ func TestStepOutPreservesGoroutine(t *testing.T) {
...
+	protest.MustHaveCgo()
+
+	// The test fixture uses linux/amd64 assembly and a *.syso file
+	// that is linux/amd64, so skip for other architectures.
+	if runtime.GOOS != "linux" || runtime.GOARCH != "amd64" {
+		t.Skipf("skipping since not linux/amd64")
+	}
+	fixture := protest.BuildFixture("issue2319/", protest.BuildModeExternalLinker)
+
+	dest := "/tmp/issue2319.exe"
```
aarzilli, line 4998: Why is this needed? (also the call to MustHaveCgo needs to be `MustHaveCgo(t)`)

thanm: Ack, sorry. Looks like I sent the wrong version. I'll fix that up.

thanm:
Not sure if I understand the "Aggregator (Delve) TeamCity build failed" errors -- looks like

`Number of tests 0 is 1 less than the provided threshold 1`

Is there some sort of book-keeping regarding test counts that I missed? Pls advise.

pkg/proc/stack.go
```diff
@@ -377,7 +377,10 @@ func (it *stackIterator) appendInlineCalls(frames []Stackframe, frame Stackframe
 		if !okname || !okfileidx || !okline {
 			break
 		}
-		if fileidx-1 < 0 || fileidx-1 >= int64(len(frame.Current.Fn.cu.lineInfo.FileNames)) {
+		var bi *BinaryInfo
+		var e *dwarf.Entry
+		filepath, fileOk := bi.cuFilePath(frame.Current.Fn.cu, int(fileidx), e)
```
aarzilli, line 382: It's pretty weird that it calls a method on a guaranteed nil object that is never used by the method. I think it should be a method of compileUnit. Instead of the logging inside cuFilePath it could return an error that the invocation in bininfo.go will print and here will just be ignored (`if err != nil { break }`).

aarzilli (commented):
> Not sure if I understand the "Aggregator (Delve) TeamCity build failed" errors -- looks like

It should go away if you rebase on master, but it isn't important. We've been setting up the TeamCity builds while you were making this and it's using the new configuration of teamcity with the old versions of the build scripts, which doesn't work.

pkg/proc/stack.go
thanm: Agreed, the nil receiver is a bit weird. OK, I'll convert to compileUnit method and move the logging outside. I'll work on updating that.

aarzilli (approved):
LGTM

aarzilli:
One of the tests failing on the teamcity builds seems semi-legitimate, on Go1.13:

```
Error compiling : exit status 2
# github.com/go-delve/delve/_fixtures/issue2319
loadinternal: cannot find runtime/cgo
main.cfunc: relocation target Cfunc not defined
```

aarzilli:
PS. if it's just that the test should be skipped on 1.13 the way we have to check for that is `if goversion.VersionAfterOrEqual(runtime.Version(), 1, 14) { ... }`

thanm:
> PS. if it's just that the test should be skipped on 1.13 the way we have to check for that is `if goversion.VersionAfterOrEqual(runtime.Version(), 1, 14) { ... }`

Hmm, hadn't expected a failure with 1.13 --  I'll take a closer look.

aarzilli:
> > PS. if it's just that the test should be skipped on 1.13 the way we have to check for that is `if goversion.VersionAfterOrEqual(runtime.Version(), 1, 14) { ... }`
> 
> Hmm, hadn't expected a failure with 1.13 -- I'll take a closer look.

Given that the problem is with compiling the test, not with delve, skipping the test would be perfectly fine.

thanm:
>Given that the problem is with compiling the test, not with delve, skipping the test would be perfectly fine.

yeah, I think that might be a better way to go. I had forgotten that back in 1.13 before the linker rewrite the linker only loads host objects in internal linking mode, so my hack for the test (the *.syso file) is not going to be viable. Probably not worth pursuing a workaround, so I'll update the test to skip on 1.13. Thanks.

thanm:
OK, skip added for 1.13 and below.

derekparker:
@thanm apologies for this if you've already done it, but would you mind rebasing this PR on master to pick up any CI related fixes? I want to make sure our CI run is happy with this, and it seems to be generating some noise.

thanm:
> would you mind rebasing this PR on master to pick up any CI related fixes? I want to make sure our CI run is happy with this, and it seems to be generating some noise.

Ok, I have rebased.  Sorry for the delays, I don't spend a lot of time with the github PR workflow.

thanm:
Looks like the TeamCity failure is:

```
...
integration1_test.go:757: Goroutine 1 not found
--- FAIL: Test1ClientServer_FullStacktrace (0.69s)
```

I ran this test on my workstation, seemed to pass ok. Of course, TeamCity failure is on Windows.

aarzilli:
That test is flaky on windows, we have to investigate why. Not a problem with this.

thanm:
OK, thanks.

thanm:
Let me know if there is anything else I need to do on this, thanks.

PR #2333, opened 2021-01-29
proc/core: off-by-one error reading ELF core files

aarzilli:
core.(*splicedMemory).ReadMemory checked the entry interval
erroneously when dealing with contiguous entries.


derekparker (approved):
LGTM

PR #2334, opened 2021-01-31
test: increase num of frames examined by TestClientServer_FullStacktrace

aarzilli:
When the runtime is allocating memory the stack trace needed to reach
user code might be greater than 10 frames.


derekparker (approved):
LGTM

PR #2335, opened 2021-02-02
service/dap: avoid double removal of temp built binary

polinasok:
Because `signalDisconnect` can be called (see function comments), sometimes we try to remove the built temp binary twice, resulting in errors like: `could not remove /Users/polina/delve/service/dap/__debug_bin: remove /Users/polina/delve/service/dap/__debug_bin: no such file or directory`. Fix this and add a test.


aarzilli:
We investigated the access is denied on delete on windows problem a while ago and couldn't figure out why it happened. I'm not sure if in this case it's the same thing or if it's a new problem. It seems too repeatable.

polinasok:
What do you suggest?

aarzilli:
Make sure Close is getting called before os.Remove and, if it is, disable the test on windows and open an issue.

polinasok:
> Make sure Close is getting called before os.Remove and, if it is, disable the test on windows and open an issue.

So I went down the rabbit hole of googling on this one. Found https://github.com/golang/go/issues/32088, related [SO post](https://stackoverflow.com/questions/63559737/golang-os-close-function-works-but-os-remove-function-does-not), etc. I think I have the right general background now.

That said, I am not sure what to make out of your comment in the context of this change. 

This change fixes the issue that arises out of trying to remove a file that was already removed by the shutdown sequence that has a couple of different entrance points and executes some parts of the code more than once. There is a comment explaining when that happens in the code. And I already put a similar safeguard for closing a chan in that code. This problem is not Windows specific. I observed it on my Mac.

If you are referring to the need to have a call to `Close` before the call to `gobuild.Remove` (which in turn calls `os.Remove`), then I am not sure how to accomplish that. I followed the pattern used in the rpc server - calling `gobuild.GoBuild` to generate the binary and then `gobuild.Remove` to remove it. Does `exec.Command("go",...)` + `Run()` leave the file open and in need of `Close()`? How do we do that without the open file descriptor? Where is this handled in the rpc server? Shouldn't this be handled generically somewhere in `gobuild` then?

Which test do you recommend that I disable on Windows?

aarzilli:
I meant BinaryInfo's Close not os.File's Close. Basically, just check that debugger.Debugger.Detach is called.

> Which test do you recommend that I disable on Windows?

The one added by this PR, that is failing.

service/dap/server.go
```diff
@@ -171,6 +171,7 @@ func (s *Server) signalDisconnect() {
 	}
 	if s.binaryToRemove != "" {
 		gobuild.Remove(s.binaryToRemove)
```
hyangah, line 173: The test is flaky. 

What I observed is the debugger.Debugger.Detach is called (most likely as a result of  handling `onDisconnectRequest`), but the process already exited so [the actual `Debugger.detach` isn't called](https://github.com/go-delve/delve/blob/c068861f95ec570fb7ea93638c87338f25c96af3/service/debugger/debugger.go#L418) .  But the test still fails.

I added 10ms Sleep before calling gobuild.Remove, and that made the test pass with -count=100. :-(

I think the race is somewhere deep down.

polinasok: Debugger.Detach is called twice actually (by disconnect and by server stop). This is by design because you could end up stopping as a result of an explicit disconnect or because of an error or an interrupt (ctrl-c), skipping the disconnect request.

In both cases, here the Detach method does not call the underlying target's detach because the target has terminated by the time it is called. This is expected for a test where the program runs to completion. We should be able to test such cases and not always disconnect tests forcefully in the middle. When a process exits, a terminate event is sent. And that triggers a disconnect request on the editor side. This is per DAP spec. So the sequence looks right to me here.

I would expect the debugger backend to release the binary when the program terminates. Does it do that? Does it do that too slowly somehow, so we end up trying to remove while the file is still in use? And if that is the case, this is not a flaky test problem, but a general problem because users could see the "could not remove" messages in the wild when the server cleans up when the program runs to completion.

polinasok:
> I meant BinaryInfo's Close not os.File's Close. Basically, just check that debugger.Debugger.Detach is called.

Detach is always called, but it never makes it to bi.Close() because the target already terminated by the time it is called.
I set a breakpoint in Close() and it is never called by this test by whatever code does ContinueOnce() and then detects that the process has exited. Should this be fixed? Should the internal readers be explicitly closed in that case as well?

> > Which test do you recommend that I disable on Windows?
> 
> The one added by this PR, that is failing.

I finally figured out how to navigate TeamCity results.

polinasok:
I wanted to debug this a bit more, but I don't have a Windows machine and the default backend is different on MacOS. Is it possible to enable TeamCity on my fork like I used to do with Travis?

service/dap/server.go
hyangah: @polinasok sorry if it's not clear - I didn't mean the sequence was wrong, but meant to reply to @aarzilli's previous comment about making sure Detach was called. And the test doesn't always fail and that's why I said 'flaky'. 

I agree that it looks like the problem is the debugger isn't releasing the binary as fast as we thought.

aarzilli: > I would expect the debugger backend to release the binary when the program terminates. Does it do that?

In theory it should be done by postExit, it is possible that we missed a spot. 

> I would expect the debugger backend to release the binary when the program terminates. Does it do that?

It is possible that this is the windows thing that we never figured out. It is possible that the thing that's keeping the file occupied is not delve but windows, because the program was executed recently.

aarzilli:
> Is it possible to enable TeamCity on my fork like I used to do with Travis?

I think that's complicated. When I need to do some tests I use the custom build. Click on the "..." next to the "Run" button, select "run as personal build" and "Upload patch". But maybe I have that because I am an admin and you don't. In that case maybe the easiest thing is to give you some permissions on teamcity. cc @nd

nd:
@polinasok I need an email to send you invite.

aarzilli:
@nd since there is no problem I've sent the invite myself.

service/dap/server.go
polinasok: So `postExit` does get called here. It must be Windows then :(

Here is the stacktrace:
```
github.com/go-delve/delve/pkg/proc/native.(*nativeProcess).postExit(0xc000a20100)
  Z:/buildAgent/work/3aca8f960d346a9c/pkg/proc/native/proc.go:319 +0x8a
github.com/go-delve/delve/pkg/proc/native.(*nativeProcess).trapWait(0xc000a20100, 0xffffffffffffffff, 0xc000965048, 0x1, 0x1)
  Z:/buildAgent/work/3aca8f960d346a9c/pkg/proc/native/proc_windows.go:378 +0x135
github.com/go-delve/delve/pkg/proc/native.(*nativeProcess).ContinueOnce(0xc000a20100, 0xc0002da000, 0x1, 0x1, 0xc000965170, 0x145b78)
  Z:/buildAgent/work/3aca8f960d346a9c/pkg/proc/native/proc.go:246 +0x15b
github.com/go-delve/delve/pkg/proc.(*Target).Continue(0xc00045fb00, 0x0, 0x0)
  Z:/buildAgent/work/3aca8f960d346a9c/pkg/proc/target_exec.go:72 +0x1ff
github.com/go-delve/delve/service/debugger.(*Debugger).Command(0xc0007486e0, 0xc0009658f0, 0x0, 0x0, 0x0)
  Z:/buildAgent/work/3aca8f960d346a9c/service/debugger/debugger.go:927 +0x1194
github.com/go-delve/delve/service/dap.(*Server).doCommand(0xc0005f4080, 0x474aeb, 0x8)
```

polinasok:
> @nd since there is no problem I've sent the invite myself.

It all worked. Thank you. I am able to run personal builds.

service/dap/server.go
hyangah: Isn't it better to retry to Remove a couple of times in case of failure if a bit of delay could make it succeed? It's annoying to see delve leaving binaries.

aarzilli (approved):
LGTM

service/dap/server.go
polinasok: https://github.com/go-delve/delve/pull/2351

service/dap/server_test.go
```diff
@@ -2040,15 +2041,38 @@ func runDebugSession(t *testing.T, client *daptest.Client, cmd string, cmdReques
...
+	os.Stderr = rescueStderr
+
+	rmErrRe, _ := regexp.Compile(`could not remove .*\n`)
+	rmErr := rmErrRe.FindString(string(err))
+	if rmErr != "" {
+		// On Windows, a file in use cannot be removed.
+		// When the process exits, Delve releases the binary by calling
+		// BinaryInfo.Close(), but it appears that it is still in use (by Windows?)
+		// shortly after, so this check can get flaky. Skip access errors on Windows.
+		if runtime.GOOS != "windows" || !strings.Contains(rmErr, "Access is denied") {
```
derekparker, line 2072: Do we need this guard now that https://github.com/go-delve/delve/pull/2351 is in?

derekparker:
@polinasok `TestLaunchDebugRequest` seems to now be failing on Windows.

polinasok:
> @polinasok `TestLaunchDebugRequest` seems to now be failing on Windows.

I saw that. It's back to access denied failure. Maybe the timeout is too small. At one point TeamCity tests were not running on new commits, so it was not clear either way. Looks like that's been fixed now. I will try another version.

polinasok:
TeamCity failures look unrelated now.

PR #2338, opened 2021-02-10
locspec,proc: improve documentation

aarzilli:
* Reformat pkg/locspec documentation so that it is formatted correctly
by godoc/pkgsite
* Unexport some types and variables in proc that don't need to be
exported.


derekparker (approved):
LGTM

PR #2343, opened 2021-02-17
proc/native: correctly read g address on linux/arm64

aarzilli:
When cgo is used the address of the g struct is saved on the special
register TPIDR_EL0. Because executing C code could overwrite the
contents of R28 that normally contains the address of g we should read
it from TPIDR_EL0 instead when runtime.iscgo is set.

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -61,6 +61,7 @@ type Target struct {
 	// have read and parsed from the targets memory.
 	// This must be cleared whenever the target is resumed.
 	gcache goroutineCache
+	iscgo  *bool
 }
 
 // ErrProcessExited indicates that the process has exited and contains both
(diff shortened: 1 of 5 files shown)


pkg/proc/target.go
```diff
@@ -181,6 +181,19 @@ func NewTarget(p Process, currentThread Thread, cfg NewTargetConfig) (*Target, e
 	return t, nil
 }
 
+// IsCgo returns the value of runtime.iscgo
+func (t *Target) IsCgo() bool {
+	scope := globalScope(t.BinInfo(), t.BinInfo().Images[0], t.Memory())
```
derekparker, line 186: Seems we only call this once, but maybe we should memoize the result just in case?

derekparker (approved):
LGTM

derekparker:
I want to wait to pull this in until we run this on the new arm64 builder. @aarzilli perhaps if you rebase and push it'll trigger?

aarzilli:
Done, all tests passed except TestCgoStacktrace and TestCgoEval, which we should probably disable on arm64.

PR #2347, opened 2021-02-18
service/dap: include optional breakpoint info in setBreakpoints responses

polinasok:
Updates https://github.com/golang/vscode-go/issues/1212

Following the same response pattern as the original vscode-go adapter seems to no longer result in the same vscode UI behavior - see BREAKPOINTS and margin marker. Adding optional information fixes that.

Before
image
After
image


aarzilli (approved):
LGTM

PR #2352, opened 2021-02-21
Documentation: update install instructions

aarzilli:
Remove OS specific install instructions that all repeat the same thing.

Since `go get ...` is being deprecated (to install executables)
recommend an install procedure that works on all supported versions of
go (`git clone && go install`) but also mention the new `go install`
mode in 1.16.


derekparker (approved):
LGTM

PR #2353, opened 2021-02-23
service/dap: detect and report unloaded variables

polinasok:
Updates #1515

This lays foundation and tests for where to add support for loading variables on the fly.


aarzilli (approved):
LGTM

PR #2358, opened 2021-02-24
TeamCity: fix go dist link on mac

nd:


aarzilli (approved):
LGTM

aarzilli:
The mac tests are still failing, looking at the logs I think that it might be because of debugserver's authentication dialog, or something like that (the gdbserver backend gets stuck in the Listen method).

nd:
it was indeed the dialog, fixed

PR #2360, opened 2021-02-24
service/dap: use specified working directory for launch requests

suzmue:
If a user specifies a working directory in the launch request,
then the process should use that working directory. This may
affect how the program runs and the user should be able to have
control over this.


polinasok (commented):
You can add a unittest to server_test.go using _fixtures/workdir.go.
At the very least let's add a bad case to [TestBadLaunchRequests](https://github.com/go-delve/delve/blob/2e80b32c417494ab3fabe929e7051076a7049f7c/service/dap/server_test.go#L2476)

polinasok:
I am looking at the old adapter and we set `--wd` in the context of dlv-attach well. But that shouldn't have any effect, should it?

service/dap/server.go
```diff
@@ -521,6 +521,19 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+	wd, ok := request.Arguments["cwd"]
+	if ok {
+		wdParsed, ok := wd.(string)
+		if !ok {
+			s.sendErrorResponse(request.Request,
+				FailedToLaunch, "Failed to launch",
+				fmt.Sprintf("'cwd' attribute '%v' in debug configuration is not a string.", wd))
+			return
+		}
+		s.config.Debugger.WorkingDir = wdParsed
```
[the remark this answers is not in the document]
suzmue: This change actually only uses the working directory from the launch arguments, not necessarily the workingDir passed on the command line. I think this warning still holds for the most part. Perhaps it needs to be slightly reworded.

suzmue:
@polinasok  thanks for the pointers on adding tests!

With regards to attach, the WorkingDir that we set for launch is specific to launch and does not need to be set for attach, since this is where the program should be launched from.

polinasok:
> @polinasok thanks for the pointers on adding tests!
> 
> With regards to attach, the WorkingDir that we set for launch is specific to launch and does not need to be set for attach, since this is where the program should be launched from.
 
True. Launch and attach do have separate request handlers and arguments. I was more thinking in terms of reminding myself when cwd is being used for other things (e.g. path mapping), which we would need to do consistently in both cases. Also, does cwd get used if the path to the binary is relative? I think in the existing adapter 'cwd' might be a superset of delve's `--wd` as it is used to configure other things, not just merely gets passed to `dlv --wd`, doesn't it? Just trying to think ahead if we will have to overload the meaning of the attribute as we move logic around (e.g. doing remote path mapping on the remote side). And in that case, we might want to rename the arg to avoid user confusion.

service/dap/server_test.go
```diff
@@ -1654,6 +1655,41 @@ func TestSetBreakpoint(t *testing.T) {
 	})
 }
 
+// TestWorkingDir executes to a breakpoint and tests that the specified
+// working directory is the one used to run the program.
+func TestWorkingDir(t *testing.T) {
+	runTest(t, "workdir", func(client *daptest.Client, fixture protest.Fixture) {
+		fixtureWorkingDir, err := filepath.Abs(filepath.Join(protest.FindFixturesDir(), "buildtest"))
+		if err != nil {
+			t.Fatalf("failed to get working directory: %e", err)
```
polinasok, line 1664: This message confused me. For a second this check was designed to look up the working directory of the fixture program running under runTest, which doesn't make sense until later in the test. Can we change it to "failed to get fixture directory" please?

service/dap/server_test.go
```diff
@@ -1654,6 +1655,41 @@ func TestSetBreakpoint(t *testing.T) {
...
+		}
+		runDebugSessionWithBPs(t, client, "launch",
+			// Launch
+			func() {
+				client.LaunchRequestWithArgs(map[string]interface{}{
+					"mode":        "exec",
+					"program":     fixture.Path,
+					"stopOnEntry": false,
+					"cwd":         fixtureWorkingDir,
```
polinasok, line 1673: What directory is this by default? A comment explaining the diff might be helpful.
Also you might want to match to this similar test:
https://github.com/go-delve/delve/blob/2e80b32c417494ab3fabe929e7051076a7049f7c/pkg/proc/proc_test.go#L2592-L2596

aarzilli:
TestWorkingDir fails on windows.

service/dap/server_test.go
suzmue: Sounds good. Will match to the proc test to make it clearer that this is not the default working dir.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: Like for build args it should say that launch args are to be used.
What happens if the program is not specified with the full path?

service/dap/server_test.go
```diff
@@ -1654,6 +1655,49 @@ func TestSetBreakpoint(t *testing.T) {
...
+				disconnect: false,
+			}})
+	})
+}
+
+// formatStringToVariable is a helper to format a string to the format it will
+// be in a variable. The entire string is in quotes and all backslashes need to be
+// escaped.
+func formatStringToVariable(str string) string {
+	return fmt.Sprintf(`"%s"`, strings.ReplaceAll(str, "\\", "\\\\"))
```
polinasok, line 1698: Use %q?

suzmue: Yes thank you!

suzmue:
> True. Launch and attach do have separate request handlers and arguments. I was more thinking in terms of reminding myself when cwd is being used for other things (e.g. path mapping), which we would need to do consistently in both cases. Also, does cwd get used if the path to the binary is relative? I think in the existing adapter 'cwd' might be a superset of delve's `--wd` as it is used to configure other things, not just merely gets passed to `dlv --wd`, doesn't it? Just trying to think ahead if we will have to overload the meaning of the attribute as we move logic around (e.g. doing remote path mapping on the remote side). And in that case, we might want to rename the arg to avoid user confusion.

Yes in the original mapping it does set delves '--wd' as well as is used in the auto remote path mapping as the local path.

This change sets the working directory to run the process in, but does not use the directory for anything else.

Should we use a different configuration name, and define the meaning very clearly? Or should we try to keep it from being necessary to do these mappings.

Does dlv dap accept relative paths for breakpoints, etc?

suzmue:
The TeamCity build failure does not seem to be related to this change.

aarzilli (approved):
LGTM

polinasok:
> > True. Launch and attach do have separate request handlers and arguments. I was more thinking in terms of reminding myself when cwd is being used for other things (e.g. path mapping), which we would need to do consistently in both cases. Also, does cwd get used if the path to the binary is relative? I think in the existing adapter 'cwd' might be a superset of delve's `--wd` as it is used to configure other things, not just merely gets passed to `dlv --wd`, doesn't it? Just trying to think ahead if we will have to overload the meaning of the attribute as we move logic around (e.g. doing remote path mapping on the remote side). And in that case, we might want to rename the arg to avoid user confusion.
> 
> Yes in the original mapping it does set delves '--wd' as well as is used in the auto remote path mapping as the local path.

If I remember correctly, the old adapter also determines and prints dlv's current working directory, which is not necessarily the same and is derived from the program's path.

Vscode's `cwd` is defined as `"Workspace relative or absolute path to the working directory of the program being debugged. Default is the current workspace."` But in reality its default is `"."` If `"."` is passed to `--wd` when dlv is run, it would be interpreted by delve as delve's working directory (program directory). But in the context of the extension, `"."` might mean something else (current workspace). There is also a predefined variable `${cwd} - the task runner's current working directory on startup`. That becomes the current workspace, which is not necessarily the same as the directory where the individual .go file is. And then which one will be used for mapping breakpoint paths? 

I have always found this confusing. Ideally we should be logging from the extension and from the adapter what all the different directories and path roots are, so there is no confusion on which process has which working directory and how relative paths are mapped to full path when things are built, breakpoints are set, etc.

> This change sets the working directory to run the process in, but does not use the directory for anything else.
> Should we use a different configuration name, and define the meaning very clearly? 

Since we might want delve to run in a different directory than the program, which might also be different from where the workspace files are, I think we should have a separate configuration attribute for this. To make it very clear what this is and to match dlv cli behavior, it makes sense to do what delve already does - name this `wd`, make it optional and provide no default value.

> Or should we try to keep it from being necessary to do these mappings.

Are you referring to path mapping for source files? My understanding is that those are different from where the program gets executed. Like running _fixtures/workdir.go in different directories. The source files will stay in the same place, but what's printed by the program will be different. And then dlv itself that might be building these source files can also run in different directories.

> Does dlv dap accept relative paths for breakpoints, etc?

Not yet. See #2391. My plan has been to get everything working with absolute paths first, then worry about interpreting relative paths while figuring out a cleaner way to differentiate between all of the above options than whatever the current adapter does.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: Could you please update commands.go as requested above?
It should say "Warning: working directory ignored with dap: specify via launch/attach request instead".

polinasok:
>  To make it very clear what this is and to match dlv cli behavior, it makes sense to do what delve already does - name this `wd`, make it optional and provide no default value.

Looking more closely at how this is treated when no explicit value is provided. I think dlv cli maps `""` to `"."` But in trying to make sense of what vscode-go does, I see to be deriving wd from the binary path. What should the default be?

polinasok:
Following up on our offline discussion:
- confirm the default WorkingDir setting if wd is not provided (i think we were leaning towards binary directory makes sense)
- decide if we want to accept relative paths and especially paths like ".", which mean different things in the context of running dlv and launch configuration specified within an editor

cmd/dlv/cmds/commands.go
```diff
@@ -422,7 +422,7 @@ func dapCmd(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(os.Stderr, "Warning: build flags ignored with dap; specify via launch/attach request instead\n")
 		}
 		if workingDir != "" {
-			fmt.Fprintf(os.Stderr, "Warning: working directory ignored with dap; launch requests must specify full program path\n")
+			fmt.Fprintf(os.Stderr, "Warning: working directory ignored with dap: specify via launch/attach request instead\n")
```
polinasok, line 425: actually attach doesn't apply here, does it?

derekparker: That's correct, it should not apply to Attach requests.

suzmue: Done.

derekparker (approved):
LGTM

polinasok:
@suzmue, looks like this got merged before we resolved the last two undecided issues that I mentioned in my last comment (the default and support/rejection of relative paths). 

In particular, current code uses the directory where the built binary is (dlv's cwd by default) as the default working directory for the debuggee while the legacy adapter uses the directory where the program source is. In an offline discussion @suzmue suggested that it `makes more sense to do it in the program field, so that you can then specify binary and wd to move where the binary lives and where it is executed without affecting the other.`

PR #2363, opened 2021-02-25
TeamCity: select latest patch version of go for mac builds

aarzilli:
Also adds a disabled tip builder for the next-version support branch.


derekparker (approved):
LGTM

PR #2364, opened 2021-03-01
service/dap: refactor mode validation on onLaunchRequest

lggomez:
This PR does the following actions on the dap/onLaunchRequest method:

- Refactor and move to the start of the method the launch mode argument validation
- Removes the redundant validation on https://github.com/go-delve/delve/blob/2e80b32c417494ab3fabe929e7051076a7049f7c/service/dap/server.go#L490

PTAL @polinasok

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -432,19 +432,25 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 const debugBinary string = "./__debug_bin"
 
 func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
-	// TODO(polina): Respond with an error if debug session is in progress?
+	// Validate launch request mode
+	mode, ok := request.Arguments["mode"]
+	if !ok || mode == "" {
+		mode = "debug"
+	}
+	if !isValidLaunchMode(mode) {
+		s.sendErrorResponse(request.Request,
+			FailedToLaunch, "Failed to launch",
+			fmt.Sprintf("Unsupported 'mode' value %q in debug configuration.", mode))
+		return
+	}
 
+	// TODO(polina): Respond with an error if debug session is in progress?
 	program, ok := request.Arguments["program"].(string)
 	if !ok || program == "" {
 		s.sendErrorResponse(request.Request,
 			FailedToLaunch, "Failed to launch",
 			"The program attribute is missing in debug configuration.")
 		return
-	}
-
-	mode, ok := request.Arguments["mode"]
-	if !ok || mode == "" {
-		mode = "debug"
 	}
 
 	if mode == "debug" || mode == "test" {
@@ -486,14 +492,6 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		s.binaryToRemove = debugname
 	}
 
-	// TODO(polina): support "remote" mode
-	if mode != "exec" && mode != "debug" && mode != "test" {
-		s.sendErrorResponse(request.Request,
-			FailedToLaunch, "Failed to launch",
-			fmt.Sprintf("Unsupported 'mode' value %q in debug configuration.", mode))
-		return
-	}
-
 	s.setLaunchAttachArgs(request)
 
 	var targetArgs []string
@@ -533,6 +531,16 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 	// will end the configuration sequence with 'configurationDone'.
 	s.send(&dap.InitializedEvent{Event: *newEvent("initialized")})
 	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+}
+
+// TODO(polina): support "remote" mode
+func isValidLaunchMode(launchMode interface{}) bool {
+	switch launchMode {
+	case "exec", "debug", "test":
+		return true
+	}
+
+	return false
 }
 
 // onDisconnectRequest handles the DisconnectRequest. Per the DAP spec,
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -2559,6 +2559,10 @@ func TestBadLaunchRequests(t *testing.T) {
 		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "buildFlags": "123"})
 		expectFailedToLaunch(client.ExpectErrorResponse(t)) // Build error
 
+		client.LaunchRequest("", fixture.Path, stopOnEntry)
+		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t),
+			"Failed to launch: Build error: exit status 1")
+
 		// We failed to launch the program. Make sure shutdown still works.
 		client.DisconnectRequest()
 		dresp := client.ExpectDisconnectResponse(t)


polinasok (commented):
Do I understand correctly that this PR accomplishes 2 things:
(1) Some renaming/refactoring for readability
(2) Change the default behavior where if mode is missing or is empty, we assume "debug"
?

polinasok:
> The handler will (maybe wrongly) assume/default to debug mode:
> https://github.com/go-delve/delve/blob/master/service/dap/server.go#L445-L448
Why is this wrong? This is what the current vscode-go adapter does. So far our approach has been to aim for feature parity unless there is an obvious bug in the legacy implementation.

If I understand correctly, you propose that mode is required and there is not default. A test "documenting" this feature would be helpful.  I should have added one for empty/missing myself.

service/dap/server.go
```diff
@@ -486,19 +500,11 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-			fmt.Sprintf("Unsupported 'mode' value %q in debug configuration.", mode))
-		return
-	}
-
 	s.setLaunchAttachArgs(request)
 
 	var targetArgs []string
-	args, ok := request.Arguments["args"]
-	if ok {
+	args, modeValOk := request.Arguments["args"]
```
polinasok, line 506: What's wrong with 'ok' (here and above)?  Isn't that a fairly standard go idiom that also follows the style preference for shorter names? I guess we could change it to `present` for just map lookup to be more consistent with https://golang.org/doc/effective_go, but for type conversion `ok` should definitely be ok. Also, I am not sure how `modeVal` part helps readability when you are not actually dealing with `mode`.

service/dap/server.go
```diff
@@ -430,21 +430,35 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 
 // Output path for the compiled binary in debug or test modes.
 const debugBinary string = "./__debug_bin"
+// TODO(polina): support "remote" mode
+var allowedModes = map[string]struct{}{
```
polinasok, line 434: I would use launchModes since there is also attach.

service/dap/server.go
```diff
@@ -486,19 +500,11 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		s.binaryToRemove = debugname
 	}
 
-	// TODO(polina): support "remote" mode
-	if mode != "exec" && mode != "debug" && mode != "test" {
```
polinasok, line 490: Good idea to move this check. It did accidentally got pushed to a suboptimal spot as the function evolved. Thank you for catching.

lggomez:
> Do I understand correctly that this PR accomplishes 3 things:
> (1) Some renaming/refactoring for readability
> (2) Moving the error check that can and should be done sooner
> (3) Change the default behavior where if mode is missing or is empty, we assume "debug"
> ?

Yeah, I'm in the middle of looking into how to port https://github.com/golang/vscode-go/pull/89/files into the DAP service to I thought this would come in handy as a separate PR because this validation will need to be extended

Now, about (3)...

> > The handler will (maybe wrongly) assume/default to debug mode:
> > https://github.com/go-delve/delve/blob/master/service/dap/server.go#L445-L448
> 
> Why is this wrong? This is what the current vscode-go adapter does. So far our approach has been to aim for feature parity unless there is an obvious bug in the legacy implementation.
> 
> If I understand correctly, you propose that mode is required and there is not default. A test "documenting" this feature would be helpful. I should have added one for empty/missing myself.

I can see you point now as it is in what the old dap does: https://github.com/golang/vscode-go/blob/cbb14795b88ad4e38011bee6e458120b89b8eb07/src/debugAdapter/goDebug.ts#L619. In this particular scenario the server could be more 'correct' on its data contract by enforcing a more thorough validation on the input instead of using fallback values as the client (and this would not break retrocompatiblity as vs code handles this). But, since this server is only intended for VS Code for the time being I'm completely impartial to achieve feature parity as you mention, so I will preserve that behavior

service/dap/server.go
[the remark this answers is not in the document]
lggomez: Yeah, the mutability is the big trade off for me in the lookup, under the assumption that no one else is going to access and manipulate this internal map. If you agree we can go with a switch on a method and thats it

service/dap/server_test.go
```diff
@@ -2502,6 +2502,10 @@ func TestBadLaunchRequests(t *testing.T) {
 		}
 
 		// Test for the DAP-specific detailed error message.
+		client.LaunchRequest("", fixture.Path, stopOnEntry)
```
polinasok, line 2505: As the comment on previous line states, this section is for dap specific messages that come from the dap package. I moved tests for error messages that come from elsewhere to a separate section to emphasize that sometimes I can't test for specific error message because it's OS-specific and in general out of our control. Could you please move this down to around line 2560. Or you might not need this at all given #2365.

service/dap/server.go
```diff
@@ -442,10 +447,18 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		return
 	}
 
-	mode, ok := request.Arguments["mode"]
-	if !ok || mode == "" {
+	// Validate launch request mode
+	modeArg := request.Arguments["mode"]
+	mode, ok := modeArg.(string)
+	if mode == "" {
```
polinasok, line 453: The new code doesn't make it obvious that missing value, not just empty value, is also treated as "debug". I guess you are implicitly relying on string conversion failing if the value is not present because modeArg is nil,  aren't you? But won't this happen if mode is some other type as well? Just like with operations that return non-nil err, I don't think you are supposed to rely on the default value of `mode` being "" if string conversion fails. You should be checking ok right away. Doesn't the check later invalidate treating nil as "debug"? I suggest going back to the original code and using interface keys in your map. If you think we should alter the default behavior, let's do that in a separate change against the new tests I added, so it's clear what's going on.

service/dap/server_test.go
polinasok: Just wanted to add that it is not obvious what's wrong with fixutre.Path from just reading the code (i.e. that it will result in `can't load package: package /var/folders/dn/8hszkj9x21n__1gr_p8qgjgr0007jz/T/increment.f7f7ea9f: cannot find package`), which is why some of the tests below use something a bit more obvious that also results in a build error.

service/dap/server.go
```diff
@@ -442,10 +447,18 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		return
 	}
 
-	mode, ok := request.Arguments["mode"]
-	if !ok || mode == "" {
+	// Validate launch request mode
```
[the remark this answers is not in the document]
lggomez: Indeed, the mode parameter is the base of all workflows so this will be a more optimal flow

service/dap/server.go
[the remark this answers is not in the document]
lggomez: Done!

service/dap/server.go
lggomez: Because of the change to the switch instead of the map these additional type checks are no longer necessary so I rolled back this

service/dap/server_test.go
lggomez: I moved the comment as requested to line 2560

polinasok: This was a misunderstanding. The comment needs to stay where it was. It was the test that I suggested moving. If my other PR goes in, we won't need it at all, btw.

service/dap/server.go
```diff
@@ -535,6 +533,16 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
 }
 
+// TODO(polina): support "remote" mode
+func isValidLaunchMode(launchMode interface{}) bool {
+	switch launchMode {
+	case "exec", "debug", "test":
```
polinasok, line 539: Looks great

service/dap/server_test.go
```diff
@@ -2553,6 +2556,7 @@ func TestBadLaunchRequests(t *testing.T) {
 		client.LaunchRequest("debug", fixture.Path+"_does_not_exist", stopOnEntry)
 		expectFailedToLaunch(client.ExpectErrorResponse(t)) // Build error
 
+		// Test for the DAP-specific detailed error message.
```
polinasok, line 2559: Please move this back. The earlier comment was about moving the new test.

lggomez: Now I got the clarification, switched back with the test

polinasok (approved):
Please update the PR title and description to match what it evolved into. Thank you for making the change and being responsive to the review commments.

polinasok:
Test failures don't look related. 
@aarzilli "Access denied" is back. I guess we need to adjust the the timeout on Windows even more and maybe just add the safeguards in the test anyway.

aarzilli (approved):
LGTM

aarzilli:
>  @aarzilli "Access denied" is back. I guess we need to adjust the the timeout on Windows even more and maybe just add the safeguards in the test anyway.

I noticed. I'd add an exception for windows on the test. Is there anything left for this PR or can it be merged?

polinasok:
I think this PR is good to go.

PR #2365, opened 2021-03-01
service/dap: clarify default attributes behavior with tests

polinasok:
Inspired by discussion in #2364


derekparker:
cc @lggomez (the @ didn't work in the above comment).

polinasok:
> cc @lggomez (the @ didn't work in the above comment).

Oops... Thanks!
Also, want to mention that the test failure looks unrelated. I don't see a way to rerun them though without pushing another commit.

aarzilli (approved):
LGTM

Issue #2346, opened 2021-02-18
Issue with variable value and stepping through the code on Apple arm64


zdima:
### What version of Delve are you using (`dlv version`)?
$ dlv version
Delve Debugger
Version: 1.6.0
Build: $Id: 8cc9751909843dd55a46e8ea2a561544f70db34d $

### What version of Go are you using? (`go version`)?
$ go version
go version go1.16 darwin/arm64

### What operating system and processor architecture are you using?
macOS 11.2 Apple M1

### What did you do?
$ git clone https://github.com/zdima/debug-test 
$ cd debug-test/
$ go build 
$ dlv debug debug-test 
(dlv) b main.callback
(dlv) c
Application start and provide a dialog with entry field and 2 buttons.
Enter text (test) into entry field, and click OK
The debugger will stop on callback function.
(dlv) p s
(dlv) n

### What did you expect to see?
The debugger for [p s] command should print value stored in variable s (test)
When command [n] entered, the debugger moved to next line 12.

### What did you see instead?
The result of [p s]
(unreadable could not read string at 0x4 due to protocol error E08 during memory read for packet $m4,40)
When [n] command entered, the program resume and stopped outside the callback function.
![Screen Shot 2021-02-17 at 10 14 59 PM](https://user-images.githubusercontent.com/298260/108300508-6f257880-716e-11eb-90e7-cdb157d2f488.png)
![Screen Shot 2021-02-17 at 10 21 29 PM](https://user-images.githubusercontent.com/298260/108300545-85333900-716e-11eb-991b-ebb7ea6dd6dd.png)

ebilling:
I've seen similar problems with the official 1.16 release.  I went back to 1.16rc1 and things worked fine.  Unfortunately, they've pulled the installer for rc1.

MindaWang:
I've seen similar problems with the official 1.16 rc1

alleindrach:
I've encountered the same problem on macbook pro m1

rahulkhairwar:
Facing a similar same issue on Macbook Pro M1, with Go1.16 and Go1.16rc1. During debug of a program, I see random behaviour for the values - sometimes they hold some random values(I've seen this behaviour for integers), sometimes they hold the error message `unreadable: protocol error E08 during memory read for packet $m40,20`.

lcwangchao:
Same issue with me

vince91:
I also have an issue with variable values using a specific module (go-ethereum), only on Apple arm64.

[...]

PR #2368, opened 2021-03-04
TeamCity: add mac/arm64 build

nd:


nd:
We finally added AppleM1 build agent.

nd:
@aarzilli It looks like automatic build triggering doesn't work because of the created VCS root copy. It looks like the old root is used in pull requests feature settings. I think it can be fixed by detaching the VCS root from Aggregator and enabling 'show changes from dependencies' option instead.

derekparker:
@nd that's my bad, I went around messing things up. I accidentally duplicated the VCS root in trying to update the user that is used for authentication. I created a new user in GH (DelveBot) and I want to use that for auth for TeamCity, but I think in doing so I've messed something up. I essentially want TeamCity to stop authing through my account, as it ends up using my profile picture in the aggregator response.

nd:
Switched to DelveBot for GH operations. Hope it has all necessary permissions.

derekparker:
@nd I've created an access token for DelveBot but I've forgotten where to input it in the TeamCity UI.

aarzilli:
It seems that now linux/amd64 agents aren't starting?

aarzilli:
PS. in any case the M1 builder is great news!

nd:
> I've created an access token for DelveBot but I've forgotten where to input it in the TeamCity UI.

It should be specified [here](https://delve.beta.teamcity.com/admin/editProject.html?projectId=Delve&tab=versionedSettings&subTab=tokens). Generate Token and paste the GH token, after that created credentialsJSON:xxx can be used in settings.kts.

derekparker:
@nd `credentialsJSON:a7e7526c-7195-4790-bbb6-9fb4692f92d0` should work then.

nd:
updated

nd:
> It seems that now linux/amd64 agents aren't starting?

not sure what caused it, but now they should start

derekparker:
@nd looks like it still not getting triggered (or at least not updating the status on this PR). Is there still more config to fix or update?

nd:
> looks like it still not getting triggered

TeamCity still uses old settings. Merging this PR into master should update them.

.teamcity/settings.kts
```diff
@@ -49,7 +49,9 @@ val targets = arrayOf(
         "windows/amd64/tip",
 
         "mac/amd64/1.16",
-        "mac/amd64/tip"
+        "mac/amd64/tip",
+
+        "mac/arm64/1.16"
```
derekparker, line 54: Can you please also add tip for the arm builder.

nd: done

aarzilli (approved):
LGTM

nd:
It looks like TC fails to publish status for PRs. @derekparker does DelveBot have repo:status and public_repo permissions?

nd:
You can open 'Commit status publisher' and test connection [here](https://delve.beta.teamcity.com/admin/editBuildFeatures.html?id=buildType:Delve_AggregatorBuild).

derekparker:
@nd thanks, updated.

aarzilli:
@nd do you know if it's possible to have a remote shell on the M1, I'd like to investigate #2346 but it's awkward without it.

nd:
> remote shell on the M1

Currently it is accessible from internal network only. Will try to organize an access from outside.

PR #2369, opened 2021-03-05
Continuous Integration cleanup

aarzilli:
* remove github workflow for testing macOS/amd64 that is now covered by
TeamCity
* fix DeepSource glob patterns to actually match what they are intended
to match (did the interpretation change?)
* disable some cgo tests on darwin/arm64


derekparker (approved):
LGTM

PR #2371, opened 2021-03-05
all: use golang.org/x/sys/execabs

zchee:
Use `golang.org/x/sys/execabs` for 
See https://blog.golang.org/path-security for more information.

Also, run `goimports -w -local=github.com/go-delve/delve $(go list -f='{{ .Dir }}' ./... | grep -v vendor)`.


aarzilli (commented):
There are multiple problems with this:

1. TestGeneratedDocs now fails on all platforms
2. Running goimports on everything makes reviewing what changed very hard
3. Like for 2, adding go1.17-style build constraints should be done separately
4. It is very likely that this actually breaks things, delve does have to sometimes execute files from the current directory
5. But most importantly, it doesn't fix any security bugs, if you can run delve you can execute arbitrary code in a multitude of different ways

zchee:
@aarzilli Ah, your right. sorry for the noisy. 
Close.

PR #2376, opened 2021-03-08
terminal/command: Add [linespec] argument to 'continue' command

### LINKED ISSUE #2373: Allow specifying location on continue
I usually always set a breakpoint on `main.main` and then continue to it.

```
dlv debug
(dlv) b main.main
(dlv) c
```

It would be nice if I could specify this in one command.

```
dlv debug
(dlv) c main.main
```

This would:

1. Create the breakpoint.
2. Continue until it hit the breakpoint.
3. Remove the breakpoint.

icholy:
This change allows specifying an optional linespec after the 'continue' command which sets a temporary breakpoint.

Fixed #2373

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -208,7 +208,12 @@ Run until breakpoint or program termination.
 
 	continue [<linespec>]
 
-Optional linespec argument allows you to set a temporary breakpoint.
+Optional linespec argument allows you to continue until a specific location is reached. The program will halt if a breakpoint is hit before reaching the specified location.
+
+For example:
+
+	continue main.main
+	continue encoding/json.Marshal
 
 
 Aliases: c
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -206,6 +206,11 @@ Defines <alias> as an alias to <command> or removes an alias.
 ## continue
 Run until breakpoint or program termination.
 
+	continue [<linespec>]
+
+Optional linespec argument allows you to set a temporary breakpoint.
+
+
 Aliases: c
 
 ## deferred
@@ -380,7 +385,7 @@ If regex is specified only local variables with a name matching it will be retur
 ## next
 Step over to next source line.
 
-	 next [count]
+	next [count]
 
 Optional [count] argument allows you to skip multiple lines.
 
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -152,12 +152,17 @@ A list of file redirections can be specified after the new argument list to over
 	2>error.txt	redirects the standard error of the target process to error.txt
 `},
 		{aliases: []string{"rebuild"}, group: runCmds, cmdFn: c.rebuild, allowedPrefixes: revPrefix, helpMsg: "Rebuild the target executable and restarts it. It does not work if the executable was not built by delve."},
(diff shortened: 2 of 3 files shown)


icholy:
@aarzilli I can't figure out why the TeamCity build is failing.

derekparker:
@icholy the TeamCity build is failing due to unrelated issues (notable CGO related failures on ARM64).

Documentation/cli/README.md
```diff
@@ -206,6 +206,11 @@ Defines <alias> as an alias to <command> or removes an alias.
 ## continue
 Run until breakpoint or program termination.
 
+	continue [<linespec>]
+
+Optional [<linespec>] argument allows you to set a temporary breakpoint.
```
icholy, line 211: Github markdown doesn't like `[<linespect>]`. It's getting rendered as `[]`.

aarzilli (approved):

Documentation/cli/README.md
icholy: done

pkg/terminal/command.go
```diff
@@ -152,12 +152,17 @@ A list of file redirections can be specified after the new argument list to over
 	2>error.txt	redirects the standard error of the target process to error.txt
 `},
 		{aliases: []string{"rebuild"}, group: runCmds, cmdFn: c.rebuild, allowedPrefixes: revPrefix, helpMsg: "Rebuild the target executable and restarts it. It does not work if the executable was not built by delve."},
-		{aliases: []string{"continue", "c"}, group: runCmds, cmdFn: c.cont, allowedPrefixes: revPrefix, helpMsg: "Run until breakpoint or program termination."},
+		{aliases: []string{"continue", "c"}, group: runCmds, cmdFn: c.cont, allowedPrefixes: revPrefix, helpMsg: `Run until breakpoint or program termination.
+
+	continue [<linespec>]
+
+Optional linespec argument allows you to set a temporary breakpoint.
```
derekparker, line 159: We need to expand on this documentation for sure. I would give at least one example of usage and also explain why you would want to use it (e.g. essentially a "continue until" command). I'd also explain that any breakpoint hit between current location and the one specified will still cause the program to halt, so the user might not end up where they want to be.

Once #2208 lands maybe we can use it to toggle other breakpoints if that's desirable.

derekparker (changes requested):
Needs some more documentation, otherwise this looks fine.

I am really excited about this feature though, way too often I set temp breakpoints that I don't really need only so I can 'continue' to them.

pkg/terminal/command.go
icholy: @derekparker is this enough?

```
Run until breakpoint or program termination.

	continue [<linespec>]

Optional linespec argument allows you to continue until a specific location is reached. The program will halt if a breakpoint is hit before reaching the specified location.

For example:

	continue main.main
	continue encoding/json.Marshal
```

derekparker: Yes, that works!

derekparker (approved):
LGTM

derekparker (changes requested):
Can you please add a test to `pkg/terminal/command_test.go` so we have coverage of this feature.

icholy:
@derekparker added

icholy:
Anything left on my end to get this over the finish line?

PR #2379, opened 2021-03-10
service/dap: add substitutePath configuration

suzmue:
Similar to substitute-path configuration in the dlv cli, substitutePath
in dap allows users to specify path mappings that are applied to the
source files in stacktrace and breakpoint requests.

Updates #2203


suzmue:
cc @polinasok

service/dap/server.go
```diff
@@ -1546,3 +1595,17 @@ func (s *Server) doCommand(command string) {
 			}})
 	}
 }
+
+func (s *Server) toLocalPath(path string) string {
```
polinasok, line 1599: Seeing these in the code, I am more convinced now that `toClientPath` and `toServerPath` would be more clear.

service/dap/server_test.go
```diff
@@ -1669,6 +1669,87 @@ func TestSetBreakpoint(t *testing.T) {
...
+	runTest(t, "loopprog", func(client *daptest.Client, fixture protest.Fixture) {
+		localDirPath := filepath.Join(string(filepath.Separator), "path", "that", "does", "not", "exist")
+		if runtime.GOOS == "windows" {
+			localDirPath = "C:" + localDirPath
+		}
+		localFilePath := filepath.Join(localDirPath, "loopprog.go")
+		debuggerDirPath := filepath.Dir(fixture.Source)
+
+		runDebugSessionWithBPs(t, client, "launch",
+			// Attach
```
polinasok, line 1685: Launch? :)

service/dap/server_test.go
```diff
@@ -1669,6 +1669,87 @@ func TestSetBreakpoint(t *testing.T) {
...
+// that does not exist and expects the substitutePath attribute
+// in the launch configuration to take care of the mapping.
+func TestLaunchSubstitutePath(t *testing.T) {
+	runTest(t, "loopprog", func(client *daptest.Client, fixture protest.Fixture) {
+		localDirPath := filepath.Join(string(filepath.Separator), "path", "that", "does", "not", "exist")
+		if runtime.GOOS == "windows" {
+			localDirPath = "C:" + localDirPath
+		}
+		localFilePath := filepath.Join(localDirPath, "loopprog.go")
+		debuggerDirPath := filepath.Dir(fixture.Source)
```
polinasok, line 1682: To improve readability, I would drop this var and rename the other one:
`"from": nonExistentDir, "to": filepath.Dir(fixture.Source)`
Also, I am confused why you have one as file path and another as dir path. If it doesn't exist, that won't matter either way, but what if it did? When does file vs dir matter?

service/dap/server_test.go
```diff
@@ -1669,6 +1669,87 @@ func TestSetBreakpoint(t *testing.T) {
...
+
+		// Start the program to attach to
+		cmd := execFixture(t, fixture)
+
+		runDebugSessionWithBPs(t, client, "attach",
```
polinasok, line 1722: How about putting shared code in a helper to avoid duplication and show how similar these tests are? Another option is to use the attach to test for bad substitute path where the mapping cannot be completed to see where that gets caughtt.

service/dap/server_test.go
```diff
@@ -2597,6 +2672,22 @@ func TestBadLaunchRequests(t *testing.T) {
...
+		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t),
+			"Failed to launch: 'substitutePath' attribute array element 0 '123' in debug configuration is not a map[string]interface{}")
+
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "substitutePath": []interface{}{map[string]interface{}{"to": "path2"}}})
+		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t),
+			"Failed to launch: 'from' in array element 0 for 'substitutePath' '<nil>' in debug configuration is not a string")
+
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "substitutePath": []interface{}{map[string]interface{}{"from": "path1", "to": 123}}})
+		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t),
+			"Failed to launch: 'to' in array element 0 for 'substitutePath' '123' in debug configuration is not a string")
```
polinasok, line 2689: maybe `'to' value 123 in mapping "path1"=>123 is no a string`?

polinasok (commented):
Thank you for adding many tests.

polinasok:
Following up on our offline discussion:
- let's add logging every time a mapping is applied
- let's clarify in tests/comments (and eventually user-facing documentation) how relative paths are treated

service/dap/server.go
```diff
@@ -132,6 +139,34 @@ func (s *Server) setLaunchAttachArgs(request dap.LaunchAttachRequest) {
 	if ok {
 		s.args.showGlobalVariables = globals
 	}
+	rules, ok := request.GetArguments()["substitutePath"]
+	if ok {
+		rulesParsed, ok := rules.([]interface{})
+		if !ok {
+			return fmt.Errorf("'substitutePath' attribute '%v' in debug configuration is not a []interface{}", rules)
```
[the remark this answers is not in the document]
suzmue: Unfortunately, swapping with map[string]interface or map[string]string does not work, it seems like the nesting is required.

I will change to having a single error message saying that the type is not correct instead of the more confusing intermediate steps.

service/dap/server.go
```diff
@@ -132,6 +139,34 @@ func (s *Server) setLaunchAttachArgs(request dap.LaunchAttachRequest) {
...
+		}
+		localToDebug := make([][2]string, 0, len(rulesParsed))
+		debugToLocal := make([][2]string, 0, len(rulesParsed))
+		for i, arg := range rulesParsed {
+			r, ok := arg.(map[string]interface{})
```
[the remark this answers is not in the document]
suzmue: You are correct that the path mapping order does matter. This is enforced by the order of the array and then preserved by the  slice type for the stored path mapping rules.  The map is for each individual path mapping rule where the launch.json looks like:

{
"from": "path", "to": "otherPath"
}

which is a map[string]string.

The overall substitutePath attribute is a list of these, which are ordered.

service/dap/server_test.go
suzmue: The path mappings have to be directories, but we use the file path to set the breakpoint.

service/dap/server_test.go
suzmue: Done. Will also add some tests for bad path mappings and ordering.

service/dap/server.go
suzmue: Will make the change.

service/dap/server.go
```diff
@@ -132,6 +139,34 @@ func (s *Server) setLaunchAttachArgs(request dap.LaunchAttachRequest) {
 	if ok {
 		s.args.showGlobalVariables = globals
 	}
+	rules, ok := request.GetArguments()["substitutePath"]
```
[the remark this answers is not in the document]
suzmue: Done

service/dap/server.go
```diff
@@ -492,7 +527,13 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		s.binaryToRemove = debugname
 	}
 
-	s.setLaunchAttachArgs(request)
+	err := s.setLaunchAttachArgs(request)
+	if err != nil {
+		s.sendErrorResponse(request.Request,
+			FailedToLaunch, "Failed to launch",
```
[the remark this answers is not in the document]
polinasok: Any thoughts?

suzmue:
> Why would we fail instead of ignoring an optional argument that's poorly configured? And if you think it is better to fail, should we do the same with other arguments? Or maybe continue while logging a warning?

I was following the same pattern from the other arguments where a malformed argument causes an error.  I think most of the arguments are checked and then sendErrorResponses like this.

service/dap/server.go
```diff
@@ -860,8 +864,11 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 		uniqueStackFrameID := s.stackFrameHandles.create(stackFrame{goroutineID, i})
 		stackFrames[i] = dap.StackFrame{Id: uniqueStackFrameID, Line: loc.Line, Name: fnName(loc)}
 		if loc.File != "<autogenerated>" {
-			localPath := s.toLocalPath((loc.File))
-			stackFrames[i].Source = dap.Source{Name: filepath.Base(localPath), Path: localPath}
+			clientPath := s.toClientPath((loc.File))
+			if clientPath != loc.File {
+				s.log.Debugf("server path=%s converted to client path=%s in stack trace request\n", loc.File, clientPath)
```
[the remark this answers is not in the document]
suzmue: Will do.

service/dap/server_test.go
```diff
@@ -1709,33 +1689,48 @@ func TestAttachSubstitutePath(t *testing.T) {
...
+	t.Helper()
+	nonexistentDir := filepath.Join(string(filepath.Separator), "path", "that", "does", "not", "exist")
+	if runtime.GOOS == "windows" {
+		nonexistentDir = "C:" + nonexistentDir
+	}
+
+	launchAttachConfig["stopOnEntry"] = false
+	launchAttachConfig["substitutePath"] = []map[string]string{
+		{"from": nonexistentDir, "to": filepath.Dir(fixture.Source)},
+		{"from": filepath.Dir(fixture.Source), "to": filepath.Dir(fixture.Source)}, // Since the path mappings are ordered, this should never apply.
```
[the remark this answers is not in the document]
suzmue: Ok I updated them with some more description. Let me know if thats any clearer.

service/dap/server.go
```diff
@@ -1529,12 +1525,20 @@ func (s *Server) toClientPath(path string) string {
 	if len(s.args.substitutePathServerToClient) == 0 {
 		return path
 	}
-	return locspec.SubstitutePath(path, s.args.substitutePathServerToClient)
+	clientPath := locspec.SubstitutePath(path, s.args.substitutePathServerToClient)
+	if clientPath != path {
+		s.log.Debugf("server path=%s converted to client path=%s in stack trace request\n", path, clientPath)
```
polinasok, line 1530: Please remove "in stacktrace request" since this is a general logging message and can be used in other contexts

service/dap/server.go
```diff
@@ -1529,12 +1525,20 @@ func (s *Server) toClientPath(path string) string {
...
 }
 
 func (s *Server) toServerPath(path string) string {
 	if len(s.args.substitutePathClientToServer) == 0 {
 		return path
 	}
-	return locspec.SubstitutePath(path, s.args.substitutePathClientToServer)
+	serverPath := locspec.SubstitutePath(path, s.args.substitutePathClientToServer)
+	if serverPath != path {
+		s.log.Debugf("client path=%s converted to server path=%s in set breakpoints request\n", path, serverPath)
```
polinasok, line 1541: Please remove "in breakpoints request" for same reasons as above

service/dap/server_test.go
```diff
@@ -1710,10 +1710,17 @@ func substitutePathTestHelper(t *testing.T, fixture protest.Fixture, client *dap
 	}
 
 	launchAttachConfig["stopOnEntry"] = false
+	// All paths used in the 'substitutePath' attribute must be absolute directory paths.
```
polinasok, line 1713: What happens if they are not? Should launch request handler catch this? Will this otherwise fail to map in subsequent stacktrace, breakpoint, etc requests that apply the mapping? Will this fail inside of Debugger command code? Have you given this a try already or this is another one of these things that we need to try and see (like the launch request program)?

polinasok:
> > Why would we fail instead of ignoring an optional argument that's poorly configured? And if you think it is better to fail, should we do the same with other arguments? Or maybe continue while logging a warning?
> 
> I was following the same pattern from the other arguments where a malformed argument causes an error. I think most of the arguments are checked and then sendErrorResponses like this.

There are actually two different patterns in the existing code for optional attributes. This was probably not the best choice on my part, so I was curious about your opinion about the overall discrepancy.

1. onLaunchRequest check if the supplied value is ill-formatted, and if it is, fails with an error.
Affected attributes: "buildFlags" (compound type, no default), "args" (compound type, no default).
2. setLaunchAttachArgs checks if the supplied value is ill-formatted and if it is, just ignores it falling back to the default.
Affected attributes: "stopOnEntry" (scalar type, default false), "stackTraceDepth" (scalar type, default 50), "showGlobalVariables" (scalar type, default false)

With your change setLaunchAttachArgs now parses 4 args, but detects and reports errors only for 1. That's why I asked if you think that maybe the rest of the args in that same helper should also trigger errors if they have the wrong type (not necessarily in this PR).

service/dap/server_test.go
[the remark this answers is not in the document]
polinasok: > Ok I updated them with some more description. Let me know if thats any clearer.

Not really, I think they are just saying in more words the same thing as before. But upon further thinking, I think I can articulate what was confusing to me.

I was confused because "from" and "to" were identical, so it did not seem like there was anything to apply. Plus it was the only rule with "from" of that value, so did this have to do with order?

But now that I think about this more, I realize that the mapping will be applied in the direction "from"=>"to" for setting a breakpoint, something helper explicitly tests for, AND in the direction "to" => "from" for stacktrace, which is something the helper does not explicitly test for, so I forgot it was even there. It would be nice to be explicit about both directions in the test. Otherwise, it really is not clear what you mean by "applied".

Finally, this made me realize how incredibly confusing "from" and "to" labels are because they make you think about only one direction in a bi-directional relationship. Can we use "client" and "server" in the config as well?

service/dap/server_test.go
aarzilli: I don't think this requirement exists (but I could be wrong). Paths in the executable can be relative because of `-trimpath`.

suzmue: Correct I think it can be relative, as long as all of the other paths that we use are also relative. The dap server does not currently have support for relative paths or trimmed paths which is why I included this comment. The substitute path rules are applied to the beginning of the path, so they would have to have the same part trimmed.

Additionally the current code requires the paths to be directories, since a separator is added to the end of the string if it is not already there. So if you have a file /dir/main.go, and you have a mapping "from": /dir/main.go, "to": "/other/dir/main.go", it will not apply because "/dir/main.go/" is not a prefix of "/dir/main.go".

We could add some code to catch the paths early by checking if they have certain file extensions, but since this could be remote, the paths may not exist on this machine to check.

suzmue:
> > > Why would we fail instead of ignoring an optional argument that's poorly configured? And if you think it is better to fail, should we do the same with other arguments? Or maybe continue while logging a warning?
> > 
> > 
> > I was following the same pattern from the other arguments where a malformed argument causes an error. I think most of the arguments are checked and then sendErrorResponses like this.
> 
> There are actually two different patterns in the existing code for optional attributes. This was probably not the best choice on my part, so I was curious about your opinion about the overall discrepancy.
> 
> 1. onLaunchRequest check if the supplied value is ill-formatted, and if it is, fails with an error.
>    Affected attributes: "buildFlags" (compound type, no default), "args" (compound type, no default).
> 2. setLaunchAttachArgs checks if the supplied value is ill-formatted and if it is, just ignores it falling back to the default.
>    Affected attributes: "stopOnEntry" (scalar type, default false), "stackTraceDepth" (scalar type, default 50), "showGlobalVariables" (scalar type, default false)
> 
> With your change setLaunchAttachArgs now parses 4 args, but detects and reports errors only for 1. That's why I asked if you think that maybe the rest of the args in that same helper should also trigger errors if they have the wrong type (not necessarily in this PR).

I think they probably should also have errors, so that we are not silently ignoring them. Will update in another PR.

service/dap/server_test.go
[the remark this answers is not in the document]
suzmue: I agree that it is confusing. I'll try to make it less confusing, by changing the target mapping to a different path.

From offline discussion, we are not set on client and server as this may be confusing if users use this for different things. I think for now, we should leave it as "from" and "to" in order to stay consistent with dlv cli, and we can continue this discussion separately.

polinasok (approved):
In the offline discussion we discussed server/client and compiled/source as alternatives to to/from. At the same time, if we can keep the names here consistent with what the old vscode adapter uses, then the users will be able to easily reuse the same debug configurations while switching between adapters. Let's get something in and we can iterate on this later.

aarzilli (approved):
LGTM

suzmue:
Fixed the merge conflict

derekparker (approved):
LGTM

PR #2383, opened 2021-03-12
service/dap: use (*api.Variable).SinglelinesString() for dap.Variable values

polinasok:
Take advantage of the function used with dlv cli `locals` command that inlines child values and can save users extra clicks to expand. Unlike `<type> `syntax used by vscode-go that inspired the first version of dap code, dlv cli's `(*type)(address)` syntax can be used as an expression to load variables that have not been fully loaded.

Updates https://github.com/golang/vscode-go/issues/1267


polinasok:
The only test failing (TestContinueUntil) is not a dap test.

aarzilli (approved):
LGTM

Issue #2384, opened 2021-03-12
examinemem with address arithmetic


felixge:
This is a feature request or maybe this feature already exists and I didn't find it.

I'd like to be able to examine memory locations as expressed using address arithmetic. Example:

```
x 0x000000c000046778 - 8
```

Currently this gives me an error like this:

```
Command failed: convert address into uintptr type failed, strconv.ParseUint: parsing "0x000000c000046778-8": invalid syntax
```

It'd be nice if `examinemem` instead evaluated the arguments as an address arithmetic expression in order to lookup `0x000000c000046770`.

If this is interesting, I might be able to contribute the code for this.

felixge:
This issue seems to be closely related to https://github.com/go-delve/delve/issues/2159. If `p` added support for hex output that would also help me. But perhaps this proposal would be simpler to implement.

felixge:
The longer I used the delve command line, the more I want this 🙈. So I just hacked it into my local fork and it seems to work well. The patch is here: https://github.com/go-delve/delve/pull/2385

Please let me know what you think : )

aarzilli:
Implemented by 7bf5482b328673e0b7dc3ebe7c6afe92c602aee2

PR #2385, opened 2021-03-13
examinememory: evaluate addr as expression

felixge:
This patch is a proof concept for #2384. If the proposal itself is accepted, I'd be happy to add test cases to this and do whatever else is needed before merging.

It makes it easy to read memory locations at an offset of a known
address, e.g.:

```
(dlv) x 0xc000046800 + 32
0xc000046820:   0x00
```


pkg/terminal/command.go
```diff
@@ -1667,14 +1668,16 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
 				return fmt.Errorf("size must be a positive integer (<=8)")
 			}
 		default:
-			if i != len(v)-1 {
-				return fmt.Errorf("unknown option %q", v[i])
+			expr := strings.Join(v[i:], " ")
```
aarzilli, line 1671: You can't do it this way because `examinemem m["one   two"]` won't work. You have to do something like this:

```
const (
    fmtPrefix = "-fmt "
    countPrefix = "-count "
)
for args != "" {
   switch {
   case strings.HasPrefix(args, fmtPrefix):
       args = args[len(fmtPrefix):...]
       // ... 
   case strings.HasPrefix(args, countPrefix):
       args = args[len(countPrefix):....]
       // ...
   default:
       // eval expression in args
      args = ""
   }
}
```

Also you should update the documentation of `examinemem` above.

pkg/terminal/command.go
```diff
@@ -1667,14 +1668,16 @@ func examineMemoryCmd(t *Term, ctx callContext, args string) error {
...
+			val, err := t.client.EvalVariable(ctx.Scope, expr, t.loadConfig())
+			if err != nil {
+				return err
 			}
-			// TODO, maybe we can support expression.
-			address, err = strconv.ParseUint(v[len(v)-1], 0, 64)
+			address, err = strconv.ParseUint(val.Value, 0, 64)
```
aarzilli, line 1676: I think this should take into account `val.RealType`, it's weird that `examinemem &x` wouldn't work.

pkg/terminal/command.go
aarzilli: On second thought it's probably simpler and better to have an explicit option `-x` that will interpret the rest of the args string as an expression. Still you will have to parse one argument at a time `strings.SplitN(args, " ", 2)` instead of just calling `strings.FieldFunc` like it does now.

felixge:
@aarzilli thanks for the feedback. Happy to adjust the patch as discussed. But just to be clear, you're happy with the general idea for adding this feature?

aarzilli:
> But just to be clear, you're happy with the general idea for adding this feature?

Yes.

pkg/terminal/command.go
felixge: Tried my best in 49cf77ca0f16909ad2c152ef8a4471c9e2ec4de0 - let me know what ya think.

pkg/terminal/command.go
felixge: Good point 49cf77ca0f16909ad2c152ef8a4471c9e2ec4de0 tries to handle different return types. I ended up getting it to work without looking at the `val.RealType` field, hope that's okay.

felixge:
@aarzilli this is ready for another look. docs are still missing, but I added some test cases. Let me know what you think about this iteration.

pkg/terminal/command.go
aarzilli: The way `-x` is implemented is fine but there's still the problem that the entire argument string is split in advance and `examinemem m["one  two"]` with two spaces between `one` and `two` will be mangled.

felixge: I'm trying to understand and reproduce the two-space issue you're describing. But I'm unable to see it.

Here is a quick test against `dlv` compiled from my local branch: https://gist.github.com/felixge/b4be92dbca9fbaba1a62488b2cdb9a83

aarzilli: You are right, with `strings.Split` (instead of strings.FieldsFunc) that problem doesn't exist but there's another problem:

```
(dlv) examinemem -fmt oct  -x &s1
```

(two spaces between `oct` and `-x`) fails with `Command failed: unknown option ""`

aarzilli: PS. the problem is: spaces shouldn't be significant before `-x` but can be significant after `-x`.

felixge: @aarzilli ah, got it! I think it all makes sense now, I'll update the patch to deal with this.

felixge: 3e16767894c3e70f203eeab071e1a43858b53928 should take care of double space handling. PTAL

aarzilli (approved):
LGTM

felixge:
@aarzilli thanks! This still needs docs, let's wait with merging this until I get a chance to send the changes for that, ok?

derekparker:
> @aarzilli thanks! This still needs docs, let's wait with merging this until I get a chance to send the changes for that, ok?

I'll wait for docs before merging, thanks for being considerate towards documentation!

derekparker:
@felixge when do you think you'll be able to send in docs with this patch?

pkg/terminal/command_test.go
```diff
@@ -1113,6 +1113,49 @@ func TestExamineMemoryCmd(t *testing.T) {
...
+	withTestTerminal("examinememory_expr", t, func(term *FakeTerminal) {
+		tests := []struct {
+			Expr string
+			Want int
+		}{
+			{Expr: "myInt", Want: 23},
+			{Expr: "&myInt", Want: 23},
+			// TODO: this fails, probably beause it's printing the value of myIntPtr
+			// rather than the value of the addr it is pointing to.
+			{Expr: "myIntPtr", Want: 23},
```
felixge, line 1126: I decided to add this new test to make sure the semantics of `examinememory <expression>` are well defined for all the standard Go types.

One thing I discovered is that my current implementation treats pointers differently depending on whether or not the expression itself is creating the pointer (`&myInt`) or if the variable itself already is a pointer (`myIntPtr`). IMO that's inconsistent and should be changed.

But the question is: How should this work? Should `examinememory &myInt` print the addr the pointer is pointing to, or the addr where the pointer is stored, i.e. the pointer value itself?

I can see both options to make some sense, so I figured I should see if y'all have opinions.

Alternatively we could reduce the scope of this patch to only support `addr + offset` kind of expressions officially, and declare any different kind of expression in `examinememory` as undefined behavior for now.

felixge (commented):
@derekparker sorry for the delay, got a bunch of things competing for my time right now, but will try to be more responsive in the next few days.

Anyway, I pushed some docs update in ddef59b4b002e847ae28c324b085e0f1f2bcaab5 and added 7c8891e40d50abda0e164f044d5649581a06f0a8 to discuss the semantics of this new feature a bit more. Maybe you and @aarzilli could take a look?

Thanks

_fixtures/examinememory_expr.go
```diff
@@ -0,0 +1,33 @@
+package main
```
aarzilli, line 1: I would prefer if we could avoid creating a new fixture for this. See testvariables2.go.

Documentation/cli/README.md
```diff
@@ -279,15 +279,16 @@ Aliases: ed
 ## examinemem
 Examine memory:
 
-	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <address>
+	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <expression>
```
aarzilli, line 282: This should be like this:

```
     	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <address>
     	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] -x <expression>
```

pkg/terminal/command_test.go
aarzilli: With the way you wrote the documentation you should probably only handle pointers (`val.Kind == reflect.Ptr`). You could also say that untyped integer expressions are also allowed and interpreted as addresses and do that too.

_fixtures/examinememory_expr.go
felixge: 👍 sure, on first glance this file looks like it should cover enough cases. But if it doesn't can I modify it?

Documentation/cli/README.md
felixge: 🙈 Ugh, sorry, yes. Will fix.

_fixtures/examinememory_expr.go
aarzilli: Yes.

pkg/terminal/command_test.go
felixge: "handle pointers" ... ok, but how? Should I print the memory at the addr the pointer is stored, or that the pointer is pointing to. And should `myIntPtr` be treated the same way as `&myInt`?

felixge:
@aarzilli thanks for the review. Left some comments and will update the PR based on your replied.

pkg/terminal/command_test.go
aarzilli: > "handle pointers" ... ok, but how? 

You should check that the type of the expression is a pointer and then use the value of the pointer, so &x and ptrToX would behave in the same way and `examinemem -x x` would not work.

derekparker:
@felixge what's the current status of this? I see @aarzilli has approved, but there's more conversation following.

pkg/terminal/command_test.go
felixge: Got it. Done!

_fixtures/examinememory_expr.go
felixge: Done . Got rid of this and used `testvariables2.go` instead.

Documentation/cli/README.md
felixge: Done.

felixge:
@aarzilli @derekparker this should be ready for another review now. The test failure seems unrelated as far as I can tell.

aarzilli (approved):

derekparker (approved):
LGTM

derekparker:
Thanks for this feature!

felixge:
@aarzilli @derekparker thanks for your patience and reviews as well as making delve awesome to begin with!

aarzilli:
You're welcome!

PR #2388, opened 2021-03-14
godwarf: assert children are not ignored

andreimatei:
The godwarf package provides two ways to turn a dwarf.Entry into a
godwarf.Tree: LoadTree and EntryToTree. The former doesn't handle
children - it doesn't advance a Reader past them (in fact, it doesn't
even know about a Reader). EntryToTree is only used for variables and
formal param DIEs, which don't have children, and it would very likely
be incorrect to use it for DIEs with children. This patch makes the
function panic if the entry can have children.

### DIFF
--- a/pkg/dwarf/godwarf/tree.go
+++ b/pkg/dwarf/godwarf/tree.go
@@ -75,7 +75,7 @@ func LoadTree(off dwarf.Offset, dw *dwarf.Data, staticBase uint64) (*Tree, error
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


andreimatei:
Please feel free to close if you don't like it.

If you do like it, another thing that `LoadTree` does which `EntryToTree` doesn't is dealing with PC ranges for the entry - `Tree.resolveRanges`. I could take a suggestion about how to assert that that part is not applicable when `EntryToTree` is called.

aarzilli (approved):
LGTM

PR #2389, opened 2021-03-14
proc/gdbserial: remove unnecessary conditional compilation

aarzilli:
Remove unnecessary conditionally compiled files introduced with the
darwin/arm64 port.


derekparker (approved):
LGTM

PR #2390, opened 2021-03-14
service: fix link to expr communication

andreimatei:
The wiki link was dangling. I think I got the right one instead.


aarzilli (approved):
LGTM

PR #2398, opened 2021-03-20
terminal: add optional format argument to print, display

### LINKED ISSUE #1038: Would it be possible to add a feature to dump structures in hexadecimal?

### LINKED ISSUE #1800: print can't output digits with HEX format.

### LINKED ISSUE #2159: Add ability to chose the base used to print integers
Delve 1.5.0
Go 1.14.7 linux/amd64
Gentoo

Printing a pointer variable from within delve gives a BEAUTIFUL pretty-print of the struct it points to, but does not display the address.  Attempting to print the address of the expression (double indirection), only gives the address of the pointer variable its self.

```
(dlv) p a
*main.A {
        number: 1,}
(dlv) p &a
(**main.A)(0xc000167ae8)
(dlv) p uintptr(a)
824633917608
```

824633917608 is 0xc0000300a8 in hex.  I suggest that printing any pointer variable display the address in additional to the pretty-print of the dereferenced value.

aarzilli:
Changes print so a format argument can be specified by using '%' as
prefix. For example:

print %x d

will print variable 'd' in hexadecimal. The interpretarion of the
format argument is the same as that of fmt's package.

Fixes #1038
Fixes #1800
Fixes #2159

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -242,7 +242,7 @@ Aliases: disass
 ## display
 Print value of an expression every time the program stops.
 
-	display -a <expression>
+	display -a [%format] <expression>
 	display -d <number>
 
 The '-a' option adds an expression to the list of expression printed every time the program stops. The '-d' option removes the specified expression from the list.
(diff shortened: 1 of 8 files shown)


aarzilli:
Done.

derekparker (approved):
LGTM

PR #2399, opened 2021-03-22
service/dap: truncate long compound map keys and use unique address suffix for uniqueness

polinasok:
When keys are truncated, there is risk of duplication in remaining strings. Using address keys as suffixes fixes that.
Removes suffixes in all other cases because after change #2383 keys that have not been truncated are unique inlined values rather than identical key type placeholders.

Updates https://github.com/golang/vscode-go/issues/1267

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -1190,7 +1190,7 @@ func TestScopesAndVariablesRequests2(t *testing.T) {
 						client.VariablesRequest(ref)
 						m3 := client.ExpectVariablesResponse(t)
 						expectChildren(t, m3, "m3", 2) // each key-value represented by a single child
-						ref = expectVarRegex(t, m3, 0, `main\.astruct {A: 1, B: 1}\(0x[0-9a-f]+\)`, `m3\[\(\*\(\*"main.astruct"\)\(0x[0-9a-f]+\)\)\]`, "42", hasChildren)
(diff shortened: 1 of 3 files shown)


service/dap/server_test.go
```diff
@@ -1207,6 +1207,14 @@ func TestScopesAndVariablesRequests2(t *testing.T) {
 							validateEvaluateName(t, client, m3kv1, 1)
 						}
 					}
+					// key - compound + truncated, value - scalar
+					ref = expectVarExact(t, locals, -1, "m5", "m5", `map[main.C]int [{s: "very long string 0123456789a0123456789b0123456789c0123456789d012...+73 more"}: 1, ]`, hasChildren)
+					if ref > 0 {
+						client.VariablesRequest(ref)
+						m5 := client.ExpectVariablesResponse(t)
+						expectChildren(t, m5, "m5", 1)
+						expectVarRegex(t, m5, 0, `main\.C {s: "very long string 0123456789a0123456789b0123456789c01\.\.\. @ 0x[0-9a-f]+`, `m5\[\(\*\(\*"main\.C"\)\(0xc[0-9a-f]+\)\)\]`, "1", hasChildren)
```
aarzilli, line 1216: Where it says `0xc` it should say `0x`, it fails on minor architectures because addresses don't start with `c`.

polinasok: An accidental copy-paste artifact. I got distracted before pushing a fix, oops.

polinasok:
The test that's failing now is unrelated - TestAttachDetach in proc.

aarzilli (approved):
LGTM

PR #2400, opened 2021-03-22
service/dap: supports noDebug launch requests

hyangah:
Following the DAP spec 

https://microsoft.github.io/debug-adapter-protocol/specification

Launch Request
This launch request is sent from the client to the debug adapter to 
start the debuggee with or without debugging (if ‘noDebug’ is true).

This is admittedly strange to be implemented inside delve (debugger),
but implementing this outside `dlv dap`  (from the editor plugin side)
implies the need for another DAP layer, which isn't desirable either.

If the launch requests has noDebug attribute set, run the built
binary directly. The launch request handler will block until
the binary terminates, so the editor won't send additional requests
like breakpoint setting etc. Still disconnect or restart requests
can flow in though and they should trigger killing of the target
process if it's still running.

In order to run the binary using os/exec on windows, the target
binary has to have .exe as its extension. So, add .exe to the default
output name if it is on windows. I am not sure though yet we want
to modify the user-specified output or not yet. Considering how
go commands behave (not automatically append .exe for 'go build -o')
I think respecting what user specified is right, but the failure
(file not exist) may be mysterious.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -16,10 +16,13 @@ import (
 	"io"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/go-delve/delve/pkg/gobuild"
 	"github.com/go-delve/delve/pkg/logflags"
@@ -67,6 +70,11 @@ type Server struct {
 	variableHandles *variablesHandlesMap
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
+
+	mu sync.Mutex
+
+	// noDebugProcess is set for the noDebug launch process.
+	noDebugProcess *exec.Cmd
 }
 
 // launchAttachArgs captures arguments from launch/attach request that
@@ -431,6 +439,13 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 // Output path for the compiled binary in debug or test modes.
 const debugBinary string = "./__debug_bin"
 
+func cleanExeName(name string) string {
+	if runtime.GOOS == "windows" && filepath.Ext(name) != ".exe" {
+		return name + ".exe"
+	}
+	return name
+}
+
 func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 	// Validate launch request mode
 	mode, ok := request.Arguments["mode"]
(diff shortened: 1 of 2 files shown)


hyangah:
Finally, all checks have passed. 
This brings in more changes than I hoped due to the windows issue. (i.e. os/exec.Command.Start can't locate the built binary unless it's .exe file. Alternative is to let the debugger Launch & continue, but not return until the debugee terminates. That will prevent the editor from sending breakpoints and other requests.)

@polinasok @suzmue

derekparker:
Can you bring some context here as to the motivation behind this change? I'm unsure of why we would want Delve to handle simple process creation without debugging that process at all.

hyangah:
@derekparker This is a strange side of DAP spec. :-(

https://microsoft.github.io/debug-adapter-protocol/specification
* Launch Request
This launch request is sent from the client to the debug adapter to start the debuggee with or without debugging (if ‘noDebug’ is true).
...
```
interface LaunchRequestArguments {
  /**
   * If noDebug is true the launch request should launch the program without
   * enabling debugging.
   */
  noDebug?: boolean;

  /**
   * Optional data from the previous, restarted session.
   * The data is sent as the 'restart' attribute of the 'terminated' event.
   * The client should leave the data intact.
   */
  __restart?: any;
}
Response to ‘launch’ request. This is just an acknowledgement, so no body field is required.

```

derekparker:
@hyangah ack, thanks for the explanation! Can you please rebase this PR to solve the conflicts.

hyangah:
@derekparker Thanks. Rebased.

service/dap/server.go
```diff
@@ -546,6 +572,52 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+	}
+	s.noDebugProcess = cmd
+	s.mu.Unlock() // allow disconnect or restart requests to call stopNoDebugProcess.
+
+	// block until the process terminates.
+	if err := cmd.Wait(); err != nil {
+		s.log.Errorf("process exited with %v", err)
+	}
+
+	s.mu.Lock()
```
derekparker, line 601: We never release this lock before returning, is this ok? What if we had a long running server with multiple such requests, this would block wouldn't it?

derekparker: (nevermind, there's the `defer s.mu.Unlock()` above I forgot about)

derekparker (approved):
LGTM

PR #2403, opened 2021-03-24
service/dap: switch goroutines when stepping

suzmue:
The step in, next, and step out requests all include a specified argument of a threadId.
This change will switch to the specified goroutine, if necessary, before executing
the step command.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -865,30 +865,20 @@ func (s *Server) onStepOutRequest(request *dap.StepOutRequest) {
 }
 
 func (s *Server) doStepCommand(command string, threadId int) {
-	// If we encounter an error, we will have to send a stopped event
-	// since we already sent the step response.
-	var sendErrorStoppedEvent = func(goroutineId int) {
+	// Use SwitchGoroutine to change the current goroutine.
+	state, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
+	if err != nil {
+		s.log.Errorf("Error switching goroutines while stepping: %e", err)
+		// If we encounter an error, we will have to send a stopped event
+		// since we already sent the step response.
 		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
 		stopped.Body.AllThreadsStopped = true
-		stopped.Body.ThreadId = goroutineId
+		if state.SelectedGoroutine != nil {
+			stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		}
 		stopped.Body.Reason = "error"
 		s.send(stopped)
-	}
-
-	state, err := s.debugger.State( /*nowait*/ true)
-	if err != nil {
-		s.log.Errorf("Error retrieving debugger state while stepping: %e", err)
-		sendErrorStoppedEvent(state.SelectedGoroutine.ID)
 		return
-	}
-	// Use SwitchGoroutine to change the current goroutine.
-	if state.SelectedGoroutine == nil || state.SelectedGoroutine.ID != threadId {
-		state, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
-		if err != nil {
-			s.log.Errorf("Error switching goroutines while stepping: %e", err)
-			sendErrorStoppedEvent(state.SelectedGoroutine.ID)
-			return
-		}
 	}
 	s.doCommand(command)
 }
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -881,16 +881,6 @@ func (s *Server) doStepCommand(command string, threadId int) {
 		sendErrorStoppedEvent(state.SelectedGoroutine.ID)
 		return
 	}
-	if state.NextInProgress {
-		s.log.Debug("NextInProgress when stepping")
-		err := s.debugger.CancelNext()
-		if err != nil {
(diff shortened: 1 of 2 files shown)


service/dap/server_test.go
```diff
@@ -2060,6 +2060,142 @@ func TestNextAndStep(t *testing.T) {
...
+				client.LaunchRequest("exec", fixture.Path, !stopOnEntry)
+			},
+			// Set breakpoints
+			fixture.Source, []int{15},
+			[]onBreakpoint{{ // Stop at line 15
+				execute: func() {
+					goroutineId := testStepParkedHelper(t, client, fixture)
+
+					client.StepOutRequest(goroutineId)
+					client.ExpectStepOutResponse(t)
```
aarzilli, line 2138: The goroutine selected here isn't guaranteed to finish stepping out before the program terminates, hence the intermittent test failure.

service/dap/server.go
```diff
@@ -846,28 +846,51 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
...
+	// since we already sent the step response.
+	var sendErrorStoppedEvent = func(goroutineId int) {
+		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
+		stopped.Body.AllThreadsStopped = true
+		stopped.Body.ThreadId = goroutineId
+		stopped.Body.Reason = "error"
+		s.send(stopped)
+	}
+
+	state, err := s.debugger.State( /*nowait*/ true)
```
aarzilli, line 878: I think just the call:

```
s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
```

is enough here. Any particular reason to do this check first?

service/dap/server_test.go
suzmue: Ah yes of course. Thanks for the tip! I will remove the test for stepOut here and see if there is a different fixture that would be easier to use for step out.

service/dap/server.go
suzmue: The initial reasoning behind doing the check was an optimization. I will change it to just do the call, and if we have performance issues this is an optimization we can make later.

service/dap/server.go
```diff
@@ -846,28 +846,41 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
...
+	if err != nil {
+		s.log.Errorf("Error switching goroutines while stepping: %e", err)
+		// If we encounter an error, we will have to send a stopped event
+		// since we already sent the step response.
+		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
+		stopped.Body.AllThreadsStopped = true
+		if state.SelectedGoroutine != nil {
+			stopped.Body.ThreadId = state.SelectedGoroutine.ID
+		}
+		stopped.Body.Reason = "error"
```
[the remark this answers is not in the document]
suzmue: Great idea! Will do.

service/dap/server.go
```diff
@@ -846,28 +846,41 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
...
+func (s *Server) doStepCommand(command string, threadId int) {
+	// Use SwitchGoroutine to change the current goroutine.
+	state, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
+	if err != nil {
+		s.log.Errorf("Error switching goroutines while stepping: %e", err)
+		// If we encounter an error, we will have to send a stopped event
+		// since we already sent the step response.
+		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
+		stopped.Body.AllThreadsStopped = true
+		if state.SelectedGoroutine != nil {
```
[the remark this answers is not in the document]
suzmue: You're right. I'll add the CurrentThread.GoroutineID

aarzilli (approved):
LGTM

polinasok (approved):
LGTM

derekparker (approved):
LGTM

polinasok:
Updates https://github.com/golang/vscode-go/issues/118

service/dap/server.go
```diff
@@ -846,28 +846,44 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
...
-	// This ignores threadId argument to match the original vscode-go implementation.
-	// TODO(polina): use SwitchGoroutine to change the current goroutine.
 	s.send(&dap.StepOutResponse{Response: *newResponse(request.Request)})
-	s.doCommand(api.StepOut)
+	s.doStepCommand(api.StepOut, request.Arguments.ThreadId)
+}
+
+func (s *Server) doStepCommand(command string, threadId int) {
+	// Use SwitchGoroutine to change the current goroutine.
+	state, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
```
polinasok, line 869: Are we guaranteed to get non-nil state here even in case of an error? We use state to populate a stopped event if err is not nil below.

polinasok: Added a check in https://github.com/go-delve/delve/pull/2422 after seeing "return nil, err" in Command code. That said, I have no idea how to trigger an error here. Tried switching to some random goroutine id, but it just quietly ignored my request. No error.

aarzilli: It's a bug. See #2424. The problem is that service/dap came to rely on it somewhat by treating the threadId argument as non-optional.

PR #2406, opened 2021-03-26
service: multiplex JSON-RPC and DAP on the same port [WIP]

aarzilli:
Changes service/rpccommon to serve DAP if the client appears to talk
that protocol.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -49,7 +49,7 @@ type Server struct {
 	// listener is used to accept the client connection.
 	listener net.Listener
 	// conn is the accepted client connection.
-	conn net.Conn
+	conn io.ReadWriteCloser
 	// stopChan is closed when the server is Stop()-ed. This can be used to signal
 	// to goroutines run by the server that it's time to quit.
 	stopChan chan struct{}
@@ -110,9 +110,11 @@ var DefaultLoadConfig = proc.LoadConfig{
 // via config and assumes its ownership. config.disconnectChan has to be set;
 // it will be closed by the server when the client disconnects or requests
 // shutdown. Once disconnectChan is closed, Server.Stop() must be called.
-func NewServer(config *service.Config) *Server {
+func NewServer(config *service.Config, debugger *debugger.Debugger) *Server {
 	logger := logflags.DAPLogger()
-	logflags.WriteDAPListeningMessage(config.Listener.Addr().String())
+	if debugger == nil {
+		logflags.WriteDAPListeningMessage(config.Listener.Addr().String())
+	}
 	logger.Debug("DAP server pid = ", os.Getpid())
 	return &Server{
 		config:            config,
@@ -122,6 +124,7 @@ func NewServer(config *service.Config) *Server {
 		stackFrameHandles: newHandlesMap(),
 		variableHandles:   newVariablesHandlesMap(),
 		args:              defaultArgs,
+		debugger:          debugger,
 	}
 }
 
@@ -213,9 +216,13 @@ func (s *Server) Run() {
 			s.signalDisconnect()
 			return
 		}
-		s.conn = conn
-		s.serveDAPCodec()
+		s.ServeConnection(conn)
 	}()
+}
+
+func (s *Server) ServeConnection(c io.ReadWriteCloser) {
+	s.conn = c
+	s.serveDAPCodec()
 }
 
 // serveDAPCodec reads and decodes requests from the client
@@ -448,6 +455,14 @@ func cleanExeName(name string) string {
 }
 
 func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
+	if s.debugger != nil {
+		//TODO: should we check something here?
+		s.send(&dap.InitializedEvent{Event: *newEvent("initialized")})
+		s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+		//TODO: should we send a status back here?
+		return
+	}
+
 	// Validate launch request mode
 	mode, ok := request.Arguments["mode"]
 	if !ok || mode == "" {
@@ -724,6 +739,7 @@ func (s *Server) onSetExceptionBreakpointsRequest(request *dap.SetExceptionBreak
 
 func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneRequest) {
 	if s.args.stopOnEntry {
+		//TODO: execute a blocking state call here
 		e := &dap.StoppedEvent{
 			Event: *newEvent("stopped"),
 			Body:  dap.StoppedEventBody{Reason: "entry", ThreadId: 1, AllThreadsStopped: true},
@@ -732,6 +748,7 @@ func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneReques
 	}
 	s.send(&dap.ConfigurationDoneResponse{Response: *newResponse(request.Request)})
 	if !s.args.stopOnEntry {
+		//TODO: it's probably a good idea to ignore this on a multiclient server
 		s.doCommand(api.Continue)
 	}
 }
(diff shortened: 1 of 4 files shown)


aarzilli:
This changes service/rpccommon so that a DAP client can connect directly to a headless instance of delve started in the "usual" way (`dlv --headless debug ...`). This doesn't handle special things,  `--continue` and `--accept-multiclient`, and someone else should finish it. Handling `--continue` seems fairly simple, just send a `continued` event in `onConfigurationDoneRequest` if the program is already running. `--accept-multiclient` is more effort intensive, one would need to send asynchronous `continued`, `stopped` and `breakpoint` events to clients as needed. I think the main problem is that vscode doesn't have the UI to say "disconnect from the server but leave the target running".

cc @polinasok @hyangah @suzmue

hyangah:
Thanks a lot, @aarzilli ! That will make it much easier for users to switch to use DAP.

For multiclient/continue support, I was wondering if it is feasible to make `dap.Server` to just work with a connection (maybe rename as `dap.Connection` or `dap.Session`) and let the `rpccommon.ServerImpl` handle the server layer (i.e. so it decides whether to terminate after a session ends, whether to accept multiple clients, or whether to switch back to the json mode, etc).

@polinasok is currently working on asynchronous event support (for continued, stopped, breakpoint). Once her work is done, I think we can see more clearly what can be done and need adjustment to support `--accept-multiclient` & `--continue`.

aarzilli:
> I was wondering if it is feasible to make dap.Server to just work with a connection

Yes.

> so it decides whether to terminate after a session ends, whether to accept multiple clients, or whether to switch back to the json mode

This is what this PR does.

hyangah:
> > so it decides whether to terminate after a session ends, whether to accept multiple clients, or whether to switch back to the json mode
> 
> This is what this PR does.

Yes, that's great.

>  I think the main problem is that vscode doesn't have the UI to say "disconnect from the server but leave the target running".

The DAP spec explicitly states that _"If the ‘attach’ request was used to connect to the debuggee, ‘disconnect’ does not terminate the debuggee."_, so if delve leaves the debugee as it is (I think that's already true for attach mode) and shuts down only the debug session (or connection) upon receiving a disconnect request, I don't think it is an issue for VSCode. (I am not sure about other editors though).

If a user attempts to "launch" with multi-client mode, I'd say the behavior is undefined.

aarzilli:
I'm going to close this now. When someone wants to finish this feel free to copy it.

service/dap/server.go
```diff
@@ -724,6 +739,7 @@ func (s *Server) onSetExceptionBreakpointsRequest(request *dap.SetExceptionBreak
 
 func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneRequest) {
 	if s.args.stopOnEntry {
+		//TODO: execute a blocking state call here
```
polinasok, line 742: Is this to check if it is already running or stopped? Why blocking? I think the idea would be to halt if it is running.

service/dap/server.go
```diff
@@ -732,6 +748,7 @@ func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneReques
 	}
 	s.send(&dap.ConfigurationDoneResponse{Response: *newResponse(request.Request)})
 	if !s.args.stopOnEntry {
+		//TODO: it's probably a good idea to ignore this on a multiclient server
```
polinasok, line 751: Why? How is this different from a user clicking continue manually? This is just a shortcut. (Also this is the default behavior in vscode)

polinasok (commented):
> > I think the main problem is that vscode doesn't have the UI to say "disconnect from the server but leave the target running".

Do you mean leave server + target running vs just target running?
I was unable to convince MS folks that this is worth supporting in DAP. Discussion is here:
https://github.com/microsoft/debug-adapter-protocol/issues/175 (and related https://github.com/microsoft/debug-adapter-protocol/issues/177)
 
> In VSCode, `--accept-multiclient` is mostly used for 'attach' mode.
> The DAP spec explicitly states that _"If the ‘attach’ request was used to connect to the debuggee, ‘disconnect’ does not terminate the debuggee."_, so if delve leaves the debugee as it is (I think that's already true for attach mode) and shuts down only the debug session (or connection) upon receiving a disconnect request, I don't think it is an issue for VSCode. (I am not sure about other editors though).
> 
> If a user attempts to "launch" with multi-client mode, I'd say the behavior is undefined.

The protocol and the terminateDebuggee default tell us to terminate. But ultimately the behavior of the code is defined by us as we write the handler. If we support multi-client mode and a mode where a process is attached to before launch/attach request, we will already be going beyond what the protocol recommends/expects. Then we might as well redefine the corresponding disconnect behavior as we see fit as well. The client can either send us an empty disconnect request (and then we decide when to terminate what) or we can get a request with terminateDebuggee (true by default for launch and false by default for attach) if we opt into this feature. Whether this is a launch or attach will determine what shows up in the UI by default - square for Stop (terminateDebuggee=true) or plugs for Disconnect (terminateDebuggee=false). The user can now flip between them in the new release. I am not sure if the extension can change the default icon for the session.

polinasok:
> This changes service/rpccommon so that a DAP client can connect directly to a headless instance of delve started in the "usual" way (`dlv --headless debug ...`). This doesn't handle special things, `--continue` and `--accept-multiclient`, and someone else should finish it. Handling `--continue` seems fairly simple, just send a `continued` event in `onConfigurationDoneRequest` if the program is already running. `--accept-multiclient` is more effort intensive, one would need to send asynchronous `continued`, `stopped` and `breakpoint` events to clients as needed.

Yes, this about sums it up. We would need all those features to make the servers compatible and mergeable. Except we won' be able to notify RPC clients of any changes in the state. In our previous discussion we concluded that supporting sequential multi-clients is critical, commonly used and well-oiled while parallel multi-clients is not. I was hoping to at least start by implementing the sequential case only with a separate flag and then see how much we would need the other tricker option, if at all. With dlv-dap that's easy because it is new and has no legacy requirements.

As we learned during the async code review, there are differences in how rpc and dap support async requests and requests during requests. We would likely need to align this better.

Also, there is breakpoint clearing. DAP clears all previous breakpoints with every request. If they are set in the same editor, they will be reset, but if you mix and match editors and protocols, RPC side would have a different behavior. This might notn be intuitive.

service/dap/server.go
aarzilli: Sure, that's also fine.

service/dap/server.go
aarzilli: If the program is already running this will wait until it stops and the restart it

aarzilli:
> > This changes service/rpccommon so that a DAP client can connect directly to a headless instance of delve started in the "usual" way (`dlv --headless debug ...`). This doesn't handle special things, `--continue` and `--accept-multiclient`, and someone else should finish it. Handling `--continue` seems fairly simple, just send a `continued` event in `onConfigurationDoneRequest` if the program is already running. `--accept-multiclient` is more effort intensive, one would need to send asynchronous `continued`, `stopped` and `breakpoint` events to clients as needed.
> 
> Yes, this about sums it up. We would need all those features to make the servers compatible and mergeable. Except we won' be able to notify RPC clients of any changes in the state. In our previous discussion we concluded that supporting sequential multi-clients is critical, commonly used and well-oiled while parallel multi-clients is not. I was hoping to at least start by implementing the sequential case only with a separate flag and then see how much we would need the other tricker option, if at all. With dlv-dap that's easy because it is new and has no legacy requirements.

IMHO you should just ignore the simultaneous multi client mode, just let it happen and it's the user's responsibility to deal with the synchronization. If it happens it isn't going to be multiple users but the same user connecting with, for example, vscode and dlv's command line client, they aren't going to issue a continue on both, for example.

Issue #2318, opened 2021-01-22
No-debug mode


polinasok:
This request is driven by a `Run Without Debugging` feature in vscode and `noDebug` launch flag in [DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Launch). It supported by the original vscode-go adapter that actually uses dlv to "run" programs in some cases. Please see https://github.com/golang/vscode-go/issues/1111 for details.

I can imagine that this can be relevant outside of DAP/vscode. If a user sets up a configuration (or even just a command) with various flags and options, it can be convenient to reuse it as-is with and without debugging and not have to copy all the settings in two places. As is explained in https://github.com/golang/vscode-go/issues/1111, "no debug" doesn't mean "no debugger", but to "run the program without hitting breakpoints and breaking on exceptions." 

Dlv has two hidden breakpoints. Looks like they are not subject to `clearall`, but can be cleared one-by-one.
```
$ dlv debug helloworld.go 
Type 'help' for list of commands.
(dlv) breakpoints
Breakpoint runtime-fatal-throw at 0x1033d10 for runtime.fatalthrow() /usr/local/go/src/runtime/panic.go:1162 (0)
Breakpoint unrecovered-panic at 0x1033d80 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1189 (0)
	print runtime.curg._panic.arg
(dlv) clearall
(dlv) breakpoints
Breakpoint runtime-fatal-throw at 0x1033d10 for runtime.fatalthrow() /usr/local/go/src/runtime/panic.go:1162 (0)
Breakpoint unrecovered-panic at 0x1033d80 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1189 (0)
	print runtime.curg._panic.arg
(dlv) clear unrecovered-panic
Breakpoint unrecovered-panic cleared at 0x1033d80 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1189
(dlv) breakpoints
Breakpoint runtime-fatal-throw at 0x1033d10 for runtime.fatalthrow() /usr/local/go/src/runtime/panic.go:1162 (0)
(dlv) 
```
Is that what the clients should do or should this be added as a command-line option?
Is there anything else that should be streamlined for a no-debugging experience?

aarzilli:
[...]

PR #2407, opened 2021-03-29
dap: change how noDebug launch request is served

hyangah:
PR #2400 added support for noDebug launch requests - that was done
by directly starting the target program using os/exec.Cmd and blocking
the launch request indefinitely until the program terminates or
is stopped with a disconnect request (when dlv dap can support it).
This PR updates the logic slightly - instead of skipping the launch response,
we now return the launch response. But still we don't send 'initialized' event
hoping that the client not to send any debug-related requests.

Still the handler goroutine will block in the onLaunchRequest handler until
the debugged program terminates.

Updates #2318

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -22,7 +22,6 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
-	"sync"
 
 	"github.com/go-delve/delve/pkg/gobuild"
 	"github.com/go-delve/delve/pkg/logflags"
@@ -71,10 +70,9 @@ type Server struct {
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
 
-	mu sync.Mutex
-
 	// noDebugProcess is set for the noDebug launch process.
-	noDebugProcess *exec.Cmd
+	noDebugProcess           *exec.Cmd
+	noDebugProcessTerminated bool
 }
 
 // launchAttachArgs captures arguments from launch/attach request that
@@ -166,6 +164,9 @@ func (s *Server) Stop() {
 				s.log.Error(err)
 			}
 		}
+	}
+	if s.noDebugProcess != nil {
+		s.stopNoDebugProcess()
 	}
 }
 
@@ -554,32 +555,32 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		s.config.Debugger.WorkingDir = wdParsed
 	}
 
-	noDebug, ok := request.Arguments["noDebug"]
-	if ok {
-		if v, ok := noDebug.(bool); ok && v { // noDebug == true
-			if err := s.runWithoutDebug(program, targetArgs, s.config.Debugger.WorkingDir); err != nil {
-				s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
-			} else { // program terminated.
-				s.send(&dap.TerminatedEvent{Event: *newEvent("terminated")})
-			}
+	noDebug := false
+	if noDebugArg, ok := request.Arguments["noDebug"]; ok {
+		if v, ok := noDebugArg.(bool); ok && v { // noDebug == true
+			noDebug = true
+		}
+	}
+
+	var err error
+	if noDebug {
+		if s.noDebugProcess, err = s.launchWithoutDebug(program, targetArgs, s.config.Debugger.WorkingDir); err != nil {
+			s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
+			return
+		}
+	} else {
+		if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
+			s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
 			return
 		}
 	}
-	var err error
-	if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
-		s.sendErrorResponse(request.Request,
-			FailedToLaunch, "Failed to launch", err.Error())
-		return
-	}
-
-	// Notify the client that the debugger is ready to start accepting
-	// configuration requests for setting breakpoints, etc. The client
-	// will end the configuration sequence with 'configurationDone'.
 	s.send(&dap.InitializedEvent{Event: *newEvent("initialized")})
 	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+	// TODO: if the client doesn't have supportConfigurationDone capability,
+	// start the debugging here.
 }
 
-func (s *Server) runWithoutDebug(program string, targetArgs []string, wd string) error {
+func (s *Server) launchWithoutDebug(program string, targetArgs []string, wd string) (*exec.Cmd, error) {
 	s.log.Debugln("Running without debug: ", program)
 
 	cmd := exec.Command(program, targetArgs...)
@@ -587,42 +588,25 @@ func (s *Server) runWithoutDebug(program string, targetArgs []string, wd string)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.Stdin = os.Stdin
-	cmd.Dir = s.config.Debugger.WorkingDir
+	cmd.Dir = wd
+	return cmd, nil
+}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	if s.noDebugProcess != nil {
-		return fmt.Errorf("previous process (pid=%v) is still active", s.noDebugProcess.Process.Pid)
-	}
-	if err := cmd.Start(); err != nil {
-		return err
-	}
-	s.noDebugProcess = cmd
-	s.mu.Unlock() // allow disconnect or restart requests to call stopNoDebugProcess.
-
-	// block until the process terminates.
-	if err := cmd.Wait(); err != nil {
-		s.log.Debugf("process exited with %v", err)
-	}
-
-	s.mu.Lock()
-	if s.noDebugProcess == cmd {
-		s.noDebugProcess = nil
-	}
-	return nil // Program ran and terminated.
+func (s *Server) runNoDebugProcess() error {
+	err := s.noDebugProcess.Run()
+	s.noDebugProcessTerminated = true
+	return err
 }
 
 func (s *Server) stopNoDebugProcess() {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	if s.noDebugProcess == nil {
+	if s.noDebugProcess == nil || s.noDebugProcessTerminated {
 		return
 	}
-
 	// TODO(hyangah): gracefully terminate the process and its children processes.
 	if err := s.noDebugProcess.Process.Kill(); err != nil {
 		s.log.Debugf("killing process (pid=%v) failed: %v", s.noDebugProcess.Process.Pid, err)
 	}
+	s.noDebugProcessTerminated = true
 }
 
 // TODO(polina): support "remote" mode
@@ -667,14 +651,16 @@ func (s *Server) onDisconnectRequest(request *dap.DisconnectRequest) {
 				s.log.Error(err)
 			}
 		}
-	} else {
-		s.stopNoDebugProcess()
 	}
 	// TODO(polina): make thread-safe when handlers become asynchronous.
 	s.signalDisconnect()
 }
 
 func (s *Server) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
+	if s.noDebugProcess != nil {
+		s.sendErrorResponse(request.Request, UnableToSetBreakpoints, "Unable to set or clear breakpoints", "running in noDebug mode")
+		return
+	}
 	// TODO(polina): handle this while running by halting first.
 
 	if request.Arguments.Source.Path == "" {
@@ -738,7 +724,7 @@ func (s *Server) onSetExceptionBreakpointsRequest(request *dap.SetExceptionBreak
 }
 
 func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneRequest) {
-	if s.args.stopOnEntry {
+	if s.debugger != nil && s.args.stopOnEntry {
 		e := &dap.StoppedEvent{
 			Event: *newEvent("stopped"),
 			Body:  dap.StoppedEventBody{Reason: "entry", ThreadId: 1, AllThreadsStopped: true},
@@ -746,8 +732,13 @@ func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneReques
 		s.send(e)
 	}
 	s.send(&dap.ConfigurationDoneResponse{Response: *newResponse(request.Request)})
-	if !s.args.stopOnEntry {
+	if s.debugger != nil && !s.args.stopOnEntry {
 		s.doCommand(api.Continue)
+	} else if s.noDebugProcess != nil {
+		if err := s.runNoDebugProcess(); err != nil { // block until program terminates.
+			s.log.Errorf("failed to start target: %v", err)
+		}
+		s.send(&dap.TerminatedEvent{Event: *newEvent("terminated")})
 	}
 }
 
@@ -767,7 +758,11 @@ func fnName(loc *proc.Location) string {
 
 func (s *Server) onThreadsRequest(request *dap.ThreadsRequest) {
 	if s.debugger == nil {
-		s.sendErrorResponse(request.Request, UnableToDisplayThreads, "Unable to display threads", "debugger is nil")
+		msg := "debugger is nil"
+		if s.noDebugProcess != nil {
+			msg = "running in noDebug mode"
+		}
+		s.sendErrorResponse(request.Request, UnableToDisplayThreads, "Unable to display threads", msg)
 		return
 	}
 	gs, _, err := s.debugger.Goroutines(0, 0)
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/go-delve/delve/pkg/gobuild"
 	"github.com/go-delve/delve/pkg/logflags"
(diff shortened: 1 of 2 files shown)


hyangah:
cc @polinasok

hyangah:
One drawback of this approach is that this will not allow users to run a program built without debug info. (VS Code currently allows to combine `exec` mode launch tasks with arbitrary preLaunch/postLaunch tasks). 
If there is a user of such arrangement, we can bring back `os/exec.Command`-based execution here in the future. For now, I hope this is simpler and minimizes further interruption to other ongoing work (i.e. async requests).

aarzilli:
Note that, in addition to not working when programs aren't built with debug info, there are other notable differences:

* on macOS and windows asyncpreempt will be disabled
* on darwin/arm64 ASLR will be disabled
* on macOS the use will sometimes be prompted for their password
* on macOS all versions of xcode that do not have the patch I submitted last year will not be able to run programs that `recover()` from a nil pointer dereference (and in general nil pointer dereferences will behave differently) -- I do not know if any released version of xcode contains the workaround
* running on architectures not supported by delve (for example windows/386) will not be possible

hyangah:
Thanks @aarzilli for the information. Some of them sound like a huge drawback.
Then, an alternative I can think of is to reenable the os/exec.Cmd logic back, but call its `Run` when ConfigurationDone message is received. I hope this doesn't interfere too much with the other planned work.

polinasok (commented):
Please add a reference to #2318 to the description. 

As for the new design, I found your original approach quite clever in that it disabled all but restart/disconnect controls in the UI. I suspect that this new one enables Pause even though we don't support it. The idea that I had to refine the first version was to send the launch response back, but not to send the initialize event, which is what signals to the editor that the debug adapter is ready to accept breakpoints and such. So if you skip it, then you should not get any such requests, including ConfigurationDone and ThreadsRequest.

We should still make sure that all theses debug request handlers fail gracefully in the absense of debugging. The evaluate request should already be responding with debugger-nil error. But we might want to make sure all others do the same and explicitly check for nil debugger. Even if the UI is disabled, one could trigger something unexpected with keyboard shortcuts.

service/dap/server_test.go
```diff
@@ -2377,44 +2377,37 @@ func TestLaunchRequestDefaults(t *testing.T) {
 
 func TestLaunchRequestDefaultsNoDebug(t *testing.T) {
```
polinasok, line 2378: Oh, and please remove "Defaults" from the name. Just like the duplicate tests below that you removed, it's an artifact from copying another test that tested for various defaults of the args in the launch arguments.

service/dap/server_test.go
```diff
@@ -2377,44 +2377,37 @@ func TestLaunchRequestDefaults(t *testing.T) {
...
 	runTest(t, "increment", func(client *daptest.Client, fixture protest.Fixture) {
-		runNoDebugDebugSession(t, client, "launch", func() {
```
polinasok, line 2380: how about runNoDebugSession? Two debugs are odd.

service/dap/server_test.go
```diff
@@ -2377,44 +2377,37 @@ func TestLaunchRequestDefaults(t *testing.T) {
...
+		runNoDebugDebugSession(t, client, func() {
 			client.LaunchRequestWithArgs(map[string]interface{}{
 				"noDebug": true,
```
polinasok, line 2382: Let's add some bad testing to TestBadLaunchRequests.

service/dap/server.go
```diff
@@ -549,75 +550,58 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+			return
+		}
 	}
-
-	// Notify the client that the debugger is ready to start accepting
-	// configuration requests for setting breakpoints, etc. The client
-	// will end the configuration sequence with 'configurationDone'.
 	s.send(&dap.InitializedEvent{Event: *newEvent("initialized")})
 	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+	// TODO: if the client doesn't have supportConfigurationDone capability,
```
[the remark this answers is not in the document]
hyangah: you're right. it's my misunderstanding. Removed.

service/dap/server.go
```diff
@@ -71,10 +70,9 @@ type Server struct {
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
 
-	mu sync.Mutex
-
 	// noDebugProcess is set for the noDebug launch process.
-	noDebugProcess *exec.Cmd
+	noDebugProcess           *exec.Cmd
+	noDebugProcessTerminated bool
```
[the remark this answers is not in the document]
hyangah: Placed back the mutex and removed bool.
I don't remember what made me to add the termination flag (there was some strange case), but thinking again, the underlying issue is the two goroutines racing, not the termination state.

service/dap/server.go
```diff
@@ -549,75 +550,58 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-		if v, ok := noDebug.(bool); ok && v { // noDebug == true
-			if err := s.runWithoutDebug(program, targetArgs, s.config.Debugger.WorkingDir); err != nil {
-				s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
-			} else { // program terminated.
-				s.send(&dap.TerminatedEvent{Event: *newEvent("terminated")})
-			}
-			return
+	noDebug := false
+	if noDebugArg, ok := request.Arguments["noDebug"]; ok {
+		if v, ok := noDebugArg.(bool); ok && v { // noDebug == true
```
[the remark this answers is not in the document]
hyangah: done.

service/dap/server.go
```diff
@@ -549,75 +550,58 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
 		}
 	}
+
 	var err error
-	if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
-		s.sendErrorResponse(request.Request,
-			FailedToLaunch, "Failed to launch", err.Error())
-		return
+	if noDebug {
+		if s.noDebugProcess, err = s.launchWithoutDebug(program, targetArgs, s.config.Debugger.WorkingDir); err != nil {
```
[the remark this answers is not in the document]
hyangah: now renamed to startNoDebugProcess, and there is stopNoDebugProcess.

service/dap/server.go
```diff
@@ -652,14 +636,16 @@ func (s *Server) onDisconnectRequest(request *dap.DisconnectRequest) {
 		if err != nil {
 			s.log.Error(err)
 		}
-	} else {
-		s.stopNoDebugProcess()
```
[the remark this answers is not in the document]
hyangah: I moved to Stop previously to match its debugger.Detach call. But now I am not sure about the teardown logic in general.
Changed to call it from both Stop and onDisconnectRequest for now (checking on s.noDebugProcess and setting it back to null will prevent double process kill)

service/dap/server.go
```diff
@@ -549,75 +550,58 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
 	s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+	// TODO: if the client doesn't have supportConfigurationDone capability,
+	// start the debugging here.
 }
 
-func (s *Server) runWithoutDebug(program string, targetArgs []string, wd string) error {
+func (s *Server) launchWithoutDebug(program string, targetArgs []string, wd string) (*exec.Cmd, error) {
 	s.log.Println("Running without debug: ", program)
 
 	cmd := exec.Command(program, targetArgs...)
```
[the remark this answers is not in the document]
hyangah: Updated

service/dap/server_test.go
hyangah: Done

service/dap/server_test.go
hyangah: Done

service/dap/server_test.go
hyangah: WDYT

hyangah:
> Please add a reference to #2318 to the description.
> 
> As for the new design, I found your original approach quite clever in that it disabled all but restart/disconnect controls in the UI. I suspect that this new one enables Pause even though we would not support it even with an async implementation. The idea that I had to refine the first version was to send the launch response back, but not to send the initialize event, which is what signals to the editor that the debug adapter is ready to accept breakpoints and such. So if you skip it, then you should not get any such requests, including ConfigurationDone and ThreadsRequest.

Thanks for the suggestion. That was my initial attempt (before sending the PR) and I changed my mind because I hoped to match in what handler the request blocks in when the program is running (i.e. ConfigurationDone request handler currently). I hoped matching noDebug and debug cases as much as possible could help classifying what request handlers can be potentially blocking,  but it sounds like that's not a huge deal and blocking from the launch request handling is ok. Now I changed my mind again and implemented what you described.

hyangah:
The test failure (on TeamCity/mac) seems unrelated to this change
```
------- Stdout: -------
=== RUN   TestContinue
dlv_test.go:241: API server listening at: 127.0.0.1:40573
dlv_test.go:241: hello world!
dlv_test.go:250: error detaching from headless instance: unexpected EOF
--- FAIL: TestContinue (1.23s)
```

aarzilli:
> The test failure (on TeamCity/mac) seems unrelated to this change

It is.

service/dap/server.go
```diff
@@ -550,58 +550,66 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-	if noDebugArg, ok := request.Arguments["noDebug"]; ok {
-		if v, ok := noDebugArg.(bool); ok && v { // noDebug == true
-			noDebug = true
-		}
-	}
-
-	var err error
-	if noDebug {
-		if s.noDebugProcess, err = s.launchWithoutDebug(program, targetArgs, s.config.Debugger.WorkingDir); err != nil {
+	if noDebug, _ := request.Arguments["noDebug"].(bool); noDebug {
```
polinasok, line 553: Isn't it the recommended practice to check err/ok instead of relying on the default/value not getting populated when an operation fails? Or does that not apply to scalars?
noDebug, ok := request.Arguments["noDebug"].(bool); ok && noDebug {...}

service/dap/server.go
```diff
@@ -550,58 +550,66 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-	} else {
-		if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
-			s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
-			return
+		// Skip 'initialized' event.
+		s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+
+		// Then, block until the program terminates or is stopped.
+		if err := cmd.Wait(); err != nil {
+			s.log.Errorf("program exited: %v", err)
```
polinasok, line 564: Based on your other CL, this should be Debug logging now :)

service/dap/server.go
```diff
@@ -550,58 +550,66 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+		s.mu.Unlock()
+
+		if !stopped {
+			s.send(&dap.TerminatedEvent{Event: *newEvent("terminated")})
+		}
+		return
+	}
+
+	var err error
+	if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
```
polinasok, line 580: does err := not work? (to avoid "var err error")

service/dap/server.go
```diff
@@ -550,58 +550,66 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+		s.send(&dap.LaunchResponse{Response: *newResponse(request.Request)})
+
+		// Then, block until the program terminates or is stopped.
+		if err := cmd.Wait(); err != nil {
+			s.log.Errorf("program exited: %v", err)
 		}
+
+		stopped := false
+		s.mu.Lock()
+		stopped = s.noDebugProcess == nil // if it was stopped, this should be nil.
```
polinasok, line 569: If you set it to nil only here, you could check Cmd.ProcessState here, couldn't you? And that would simplify the logic in stopNoDebugProcess too a bit.

service/dap/server.go
```diff
@@ -637,6 +645,8 @@ func (s *Server) onDisconnectRequest(request *dap.DisconnectRequest) {
 			s.log.Error(err)
 		}
 	}
+	s.stopNoDebugProcess()
```
polinasok, line 648: I know you safeguard against this being nil in the function, but I think the code was more clear when you had an else here. Either we have a debugger or a noDebug process.

service/dap/server_test.go
```diff
@@ -2713,6 +2706,14 @@ func TestBadLaunchRequests(t *testing.T) {
 
 		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "buildFlags": "bad flags"})
 		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t), "Failed to launch: Build error: exit status 1")
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "noDebug": true, "buildFlags": "bad flags"})
```
polinasok, line 2709: you could also add a test above where noDebug is something other than a bool

service/dap/server_test.go
```diff
@@ -2713,6 +2706,14 @@ func TestBadLaunchRequests(t *testing.T) {
...
+		expectFailedToLaunchWithMessage(client.ExpectErrorResponse(t), "Failed to launch: Build error: exit status 1")
+
+		// Bad "wd".
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "noDebug": false, "wd": "dir/invalid"})
+		expectFailedToLaunch(client.ExpectErrorResponse(t)) // invalid directory, the error message is system-dependent.
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "noDebug": true, "wd": "dir/invalid"})
+		expectFailedToLaunch(client.ExpectErrorResponse(t)) // invalid directory, the error message is system-dependent.
```
polinasok, line 2716: good new tests cases, thanks!

service/dap/server.go
```diff
@@ -550,58 +550,66 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+		cmd, err := s.startNoDebugProcess(program, targetArgs, s.config.Debugger.WorkingDir)
+		if err != nil {
 			s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
 			return
 		}
-	} else {
-		if s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs); err != nil {
-			s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())
-			return
+		// Skip 'initialized' event.
```
polinasok, line 559: I think it would be good to document here that by doing so we avoid debug requests.

service/dap/server.go
```diff
@@ -70,9 +71,10 @@ type Server struct {
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
 
+	mu sync.Mutex
```
polinasok: Let's call this noDebugProcessMux since we might end up with more locks and will need to differentiate.

service/dap/server.go
aarzilli: I'm not aware of anyone recommending this.

service/dap/server.go
aarzilli: `:=` only works with variable names on the left.

service/dap/server.go
polinasok: Good to know. I remember this coming up in some code review in my early Go days, but it was probably about structs, not scalars.

service/dap/server.go
polinasok: Ah, I see, "non-name ... on left side of :=". That's unfortunate.

service/dap/server.go
hyangah: I am using this multi-value returning type assertion here to avoid panic.

According to the language spec, when the type assertion fails, the value is a zero value (false, 0, empty struct). In this case, zero value (false) makes sense, so I don't think this form is violating the spec or any style suggestion. (for example, https://golang.org/src/net/http/server.go#L3147) When accessing a struct field, checking ok may be useful when accessing the zero value's field isn't desirable. Even when accessing a scalar field, checking ok may be necessary when distinction between zero value vs uninitialized value is necessary. All depend on the context.

For errors, yes, it's strongly recommended not to discard error (but as always, there are exceptions).

service/dap/server.go

service/dap/server.go

service/dap/server.go
hyangah: The program can terminate and we end up here because 1) the program exited by itself - due to normal termination or due to other signals (segfault, ...), or 2) the program was killed by Kill from stopNoDebugProcess. I don't see how to distinguish the two cases cleanly with Cmd.ProcessState when the program was killed by external processes.  In case of 2, we don't want to issue TerminatedEvent. Feel free to revisit/update it when the teardown logic & async request serving is settled down.

service/dap/server.go
hyangah: The offical name of := is a short variable declaration, and 's.debugger' isn't a variable.

service/dap/server.go
hyangah: done.
Strictly speaking, s.debugger != nil doesn't necessary mean noDebug mode (s.debugger could be in uninitialized state).

service/dap/server_test.go
hyangah: noDebug other than a bool will be ignored (not a bad request) and the launch requests should work as if it wasn't set. I see the other way - failing on non-bool type value - isn't wrong either. I am not too worried about it right now. If you want, feel free to add extra check and make it to the bad request case.
Added to the launch default test for now.

service/dap/server.go
polinasok: Based no the comment in the other PR, I understand now what your intension was here. Please disregard this comment.

derekparker (approved):
LGTM

service/dap/server.go
polinasok: Since debugger is a pointer and pointers have nil default values in Go (unlike C, as far as I know), you must be referring to a case of exiting early before neither debugger nor process have been launched. The else could be else-if instead or in addition to doing the check in stopNoDebugProcess()

service/dap/server_test.go
polinasok: Sure, looks good. I just meant we needed a test "documenting" behavior, whatever it tis.  
I did the same thing for a number of other flags - if they are bad, don't generate a launch error, but quietly use a default. Suzy and I had a discussion about this because she wanted to use the bad launch approach:
https://github.com/go-delve/delve/pull/2379#issuecomment-806053504

PR #2409, opened 2021-03-31
Documentation: mention developer group in macOS install instructions

aarzilli:
Some users seem to need to do this in order to use a debugger.

### DIFF
--- a/Documentation/installation/README.md
+++ b/Documentation/installation/README.md
@@ -38,6 +38,12 @@ If you didn't enable Developer Mode using Xcode you will be asked to authorize t
 sudo /usr/sbin/DevToolsSecurity -enable
 ```
 
+You might also need to add your user to the developer group:
+
+```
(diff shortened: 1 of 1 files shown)


Documentation/installation/README.md
```diff
@@ -38,6 +38,12 @@ If you didn't enable Developer Mode using Xcode you will be asked to authorize t
 sudo /usr/sbin/DevToolsSecurity -enable
 ```
 
+You might also need to add your user to the developer group:
+
+```
+sudo dscl . append /Groups/_developer GroupMembership your_username
+```
```
icholy, line 45: Can `your_username` be replaced with `$(whoami)`?

derekparker: Second this, seems cleaner from a copy/paste perspective for users.

aarzilli: Done.

derekparker (approved):
LGTM

PR #2412, opened 2021-04-02
service/dap: change dap error response logging to debug level

hyangah:
Error level logging shows up in the users' consoles/terminals
so be more conservative when logging. Move the followings to
Debug logging.

- DAP error reponses caused by invalid requests.
They are application level errors and DAP clients should handle them.
- Errors reported when debugee already exited.

Fixes golang/vscode-go#1392

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -159,7 +159,12 @@ func (s *Server) Stop() {
 	if s.debugger != nil {
 		kill := s.config.Debugger.AttachPid == 0
 		if err := s.debugger.Detach(kill); err != nil {
-			s.log.Error(err)
+			switch err.(type) {
+			case *proc.ErrProcessExited:
+				s.log.Debug(err)
+			default:
+				s.log.Error(err)
+			}
 		}
 	}
 }
@@ -575,7 +580,7 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 }
 
 func (s *Server) runWithoutDebug(program string, targetArgs []string, wd string) error {
-	s.log.Println("Running without debug: ", program)
+	s.log.Debugln("Running without debug: ", program)
 
 	cmd := exec.Command(program, targetArgs...)
 	// TODO: send stdin/out/err as OutputEvent messages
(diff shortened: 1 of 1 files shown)


hyangah:
cc @polinasok @suzmue

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

derekparker:
(re-running the test matrix to root out the flicker)

derekparker:
Nevermind, this failure is showing up on other PRs and should be investigated outside of this one.

service/dap/server.go
```diff
@@ -1429,7 +1444,7 @@ func (s *Server) sendErrorResponseWithOpts(request dap.Request, id int, summary,
 	er.Body.Error.Id = id
 	er.Body.Error.Format = fmt.Sprintf("%s: %s", summary, details)
 	er.Body.Error.ShowUser = showUser
-	s.log.Error(er.Body.Error.Format)
+	s.log.Debug(er.Body.Error.Format)
```
polinasok, line 1447: Why hide the error behind the logging flags?
Don't the users turn to the debug console when something doesn't work as expected because things are not always clearly reflected in the UI?

service/dap/server.go
```diff
@@ -1449,18 +1464,18 @@ func (s *Server) sendInternalErrorResponse(seq int, details string) {
 	er.Message = "Internal Error"
 	er.Body.Error.Id = InternalError
 	er.Body.Error.Format = fmt.Sprintf("%s: %s", er.Message, details)
-	s.log.Error(er.Body.Error.Format)
+	s.log.Debug(er.Body.Error.Format)
```
polinasok, line 1467: Ditto

service/dap/server.go
hyangah: Users are supposed to know the issue from the error response, not from the server-side logging.

service/dap/server.go
hyangah: Ditto.

PR #2413, opened 2021-04-03
service/dap: fix temp binary deletion race

### LINKED ISSUE #2411: service/dap: leaves the temporary binary behind
This is caused by race during Disconnect request handling.

1. Disconnect request is received, and [`onDisconnectRequest`](https://github.com/go-delve/delve/blob/b120b11cc383554064e3e2da36d29e4c5299a988/service/dap/server.go#L636-L659) starts the disconnect sequence, including detaching from the debugee.
2. `s.signalDisconnect` is called at the end.
3. `s.signalDisconnect` closes [`s.config.DisconnectChan`](https://github.com/go-delve/delve/blob/b120b11cc383554064e3e2da36d29e4c5299a988/service/dap/server.go#L189) before deleting the temporary  binary.
4. Closing of `s.config.DisconnectChan` unblocks [`waitForDisconnectSignal`](https://github.com/go-delve/delve/blob/b120b11cc383554064e3e2da36d29e4c5299a988/cmd/dlv/cmds/commands.go#L455).
5. [`server.Stop`](https://github.com/go-delve/delve/blob/b120b11cc383554064e3e2da36d29e4c5299a988/cmd/dlv/cmds/commands.go#L452) +[`os.Exit`](https://github.com/go-delve/delve/blob/b120b11cc383554064e3e2da36d29e4c5299a988/cmd/dlv/cmds/commands.go#L458) races with the remaining of `s.signalDisconnect`.

If `os.Exit` runs first, the temporary binary deletion by `s.signalDisconnect` won't complete. 

cc @polinasok @suzmue @aarzilli

polinasok:
Fixes #2411


polinasok:
cc @hyangah

service/dap/server.go
```diff
@@ -106,6 +106,28 @@ var DefaultLoadConfig = proc.LoadConfig{
 	MaxStructFields:    -1,
 }
 
+// tempBinary offers a thread-safe abstraction for
+// managing a temp compiled binary (if any).
+type tempBinary struct {
+	mux      sync.Mutex
```
hyangah, line 112: Is there a reason to have a separate mutex for binary? 
Given that the lifetime of the binary is bound to the debugger's, I think it makes more sense to share a single lock. (Server's mu, for example). I expect to see guard against other fields  in the feature (isn't s.debugger or others need protection too?). Fine-grained locks are hard to use and can lead to deadlock if misused, it's better to avoid them if they are really necessary.

polinasok: You are right. I could reuse mu. I will make the change.

polinasok:
PTAL

hyangah:
LGTM

The test failure looks like another unrelated flakiness to me.

```
TestAttachDetach github.com/go-delve/delve/pkg/proc

------- Stdout: -------
=== RUN   TestAttachDetach
proc_test.go:145: failed assertion at proc_test.go:2882: Page request after detach - Get "http://127.0.0.1:9191/nobp": read tcp 127.0.0.1:57450->127.0.0.1:9191: read: connection reset by peer
--- FAIL: TestAttachDetach (1.48s)
```

hyangah (approved):

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

derekparker:
Can be merged after rebase.

polinasok:
Unrelated TeamCity failure:
```TestFrameEvaluation github.com/go-delve/delve/pkg/proc```

PR #2414, opened 2021-04-04
service/dap: refine teardown logic

polinasok:
Updates #1515


service/dap/server.go
```diff
@@ -641,37 +651,32 @@ func isValidLaunchMode(launchMode interface{}) bool {
...
 	} else {
 		s.stopNoDebugProcess()
 	}
 	// TODO(polina): make thread-safe when handlers become asynchronous.
-	s.signalDisconnect()
+	s.triggerServerStop()
+}
+
+func (s *Server) stopDebugSession(killProcess bool) {
+	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
```
polinasok, line 668: Why doesn't rpc server halt in stop()?
https://github.com/go-delve/delve/blob/c223ef656da3ddbab3a58ea87108d4217731a01c/service/rpccommon/server.go#L91-L97? 
On a related note, the terminology collision here is unfortunate: https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#gracefully-ending-the-debug-session
From the RPC client point of view, it appears that
- Detach = detach debugger from the process (maybe kill it) + close client connection
- Disconnect = close client connection
But the client how-to uses a phrase "disconnecting a running program" as a case where Detach is called. And then we have DAP where disconnect request is sent to "disconnect from the debuggee and to terminate the debug adapter". Sigh.

polinasok:
Test failure looks unrelated:

polinasok:
cc @hyangah

service/dap/server.go
hyangah: Wild guess - isn't it because it's the rpc client who is supposed to call Halt (or Disconnect) command  before the server stops. Not sure about the case where the debugging stops because client unexpectedly crashes or gets disconnected.

service/dap/server.go
```diff
@@ -71,6 +87,8 @@ type Server struct {
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
 
+	// mu synchronizes access to objects set on start-up (from run goroutine)
+	// and stopped on teardown (from main goroutine)
```
hyangah, line 91: fyi - Since there is no mutex annotation in go, frequently used convention is to lay out the fields protected by a mutex after the mutex. For example, once you we start protecting debugger or conn with mu, place them after mu.

service/dap/server.go
```diff
@@ -144,51 +163,49 @@ func (s *Server) setLaunchAttachArgs(request dap.LaunchAttachRequest) {
...
-			default:
-				s.log.Error(err)
-			}
-		}
+		killProcess := s.config.Debugger.AttachPid == 0
+		s.stopDebugSession(killProcess)
+	} else {
+		s.stopNoDebugProcess()
 	}
+	s.mu.Unlock()
```
hyangah, line 187: can we use defer (after line 172)?

service/dap/server.go
```diff
@@ -640,38 +657,37 @@ func isValidLaunchMode(launchMode interface{}) bool {
...
-			}
-		}
+		killProcess := s.config.Debugger.AttachPid == 0 || request.Arguments.TerminateDebuggee
+		s.stopDebugSession(killProcess)
 	} else {
 		s.stopNoDebugProcess()
 	}
-	// TODO(polina): make thread-safe when handlers become asynchronous.
-	s.signalDisconnect()
+	s.mu.Unlock()
```
hyangah, line 671: Will the following work? (Sorry - it's hard for me to use github codereview ui :-( )

defer s.triggerServerStop()
s.mu.Lock()
defer s.mu.Unlock()
if s.debugger != nil {
   ... 
   s.stopDebugSession(killProcess)
 } else {
   s.stopNodebugProcess()
 }

service/dap/server.go
```diff
@@ -640,38 +657,37 @@ func isValidLaunchMode(launchMode interface{}) bool {
...
+	s.triggerServerStop()
+}
+
+func (s *Server) stopDebugSession(killProcess bool) {
+	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
+	if err == proc.ErrProcessDetached {
+		s.log.Debug(err)
+		return
+	} else if err == nil {
```
hyangah, line 680: s/else if/if

What other errors are possible and why is it ok not to Detach?

service/dap/server.go
polinasok: That's precisely the case I was thinking about. The client takes care of Halt before Detach. But if there is an error reading rpc messages from client, the server will go into shutdown mode and call Stop() without halting.

service/dap/server.go
```diff
@@ -35,13 +35,29 @@ import (
...
+// signal that can come from an OS interrupt (such as Ctrl-C) or
+// config.DisconnectChan (passed to NewServer()) as a result of
+// client connection failure or closure or a DAP disconnect request.
+//
+// (2) Run goroutine started from Run() that serves as both
+// a listener and a client goroutine. It accepts a client connection,
 // reads, decodes and processes each request, issuing commands to the
 // underlying debugger and sending back events and responses.
+// This gorouitne sends a stop-server signal via config.DisconnecChan
+// when encounering a clientt connection error or responding to
```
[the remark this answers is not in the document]
polinasok: done

service/dap/server.go
polinasok: done

service/dap/server.go
polinasok: done

service/dap/server.go
polinasok: done

service/dap/server.go
polinasok: done

polinasok (commented):
PTAL

polinasok:
Failing tests are unrelated:
TestIssue419github.com/go-delve/delve/pkg/proc
TestAttachDetachgithub.com/go-delve/delve/pkg/proc

service/dap/server.go
```diff
@@ -840,9 +865,10 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
 		s.config.Debugger.AttachPid = int(pid)
 		s.setLaunchAttachArgs(request)
 		var err error
+		s.mu.Lock()
+		defer s.mu.Unlock()
```
hyangah, line 869: This lock lives beyond the if block (so affect line 884-885) which isn't obvious. 
Consider either narrowing the scope of the lock, so lock is released before going out of the if block, or
handle if mode != "local" case first 

if mode != "local" {
  s.sendErrorResponse(...)
  return
}
// mode == "local"
...

hyangah (commented):
Thanks!
I don't know enough to review the Halt + Detach logic so still need review from delve maintainers. Other parts look good to me.

service/dap/server.go
polinasok: Will do. I am actually not happy how much logic ends up inside of lock in general :(

service/dap/server.go
```diff
@@ -607,11 +631,11 @@ func (s *Server) startNoDebugProcess(program string, targetArgs []string, wd str
 	return cmd, nil
 }
 
+// stopNoDebugProcess is called from Stop() (main goroutine) and
+// onDisconnectRequest (run goroutine) and requires holding mu lock.
 func (s *Server) stopNoDebugProcess() {
-	s.mu.Lock()
 	p := s.noDebugProcess
 	s.noDebugProcess = nil
-	defer s.mu.Unlock()
```
polinasok, line 614: @hyangah Was this meant to be regular unlock to minimize scope (hence the temp var p var)? With my new change, the entire function gets locked (so I shouldn't need `p` anymore). Not ideal? But at the same time, what else is there to do when we are existing?

hyangah: Heh, this is a bug - I meant s.mu.Unlock(), not defer s.mu.Unlock. :-)

Yeah, the intention was to reduce the scope - so the goroutine that's blocked onLaunchRequest (waiting for the termination of s.nodebugProcess) can proceed as soon as it's unblocked when p's killed. But you're right. We don't need to be too clever here.

service/dap/server.go
aarzilli: I don't think there's a particular reason, halting would probably be better. And the client howto should say detach where it says ""disconnecting a running program".

aarzilli (approved):
LGTM

service/dap/server.go
polinasok: Thanks for clarifying. I remove the temp var "p".

polinasok:
@suzmue 
TestStepOutPreservesGoroutine is failing because it hit program termination when it expected to stop after a step.
This is not related to my comment and nodebug changes. I can't reproduce it locally - a race in the test? I think it would help to enable --log-output=dap,debugger in these tests, so failed logs have more info.

polinasok:
> @suzmue
> TestStepOutPreservesGoroutine is failing because it hit program termination when it expected to stop after a step.
> This is not related to my comment and nodebug changes. I can't reproduce it locally - a race in the test? I think it would help to enable --log-output=dap,debugger in these tests, so failed logs have more info.

Fix at https://github.com/go-delve/delve/pull/2417

hyangah (approved):

derekparker:
Needs rebase, I'll TAL tomorrow.

polinasok:
> Needs rebase, I'll TAL tomorrow.

I am working on the merge.

service/dap/server.go
```diff
@@ -876,9 +913,13 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
 		s.config.Debugger.AttachPid = int(pid)
 		s.setLaunchAttachArgs(request)
 		var err error
-		if s.debugger, err = debugger.New(&s.config.Debugger, nil); err != nil {
-			s.sendErrorResponse(request.Request,
-				FailedToAttach, "Failed to attach", err.Error())
+		func() {
+			s.mu.Lock()
+			defer s.mu.Unlock() // Make sure to unlock in case of panic that will become internal error
```
polinasok, line 918: I am not sure why this did not show up before the merge, but there was a test with negative attach pid that resulted in a panic from debugger.New that skipped releasing the lock and then disconnect was forever waiting to acquire it. I changed onLaunchRequest in a similar way as well just in case.

polinasok:
Unrelated TeamCity failure:
```TestPackageRenamesgithub.com/go-delve/delve/service/test```

polinasok:
@derekparker Is there more to be done to get this PR in? It would be great to get it out of the way to resolve more conflicts in subsequent PRs. Thank you.

PR #2415, opened 2021-04-06
proc: remove duplicate Registers.Get implementations

aarzilli:
Moves the implementation of Registers.Get out of the backends and into
proc where it can be implemented only once per architecture.


derekparker (approved):
LGTM

derekparker:
Really nice refactor, thanks!

PR #2417, opened 2021-04-07
service/dap: handle terminated events in TestStepOutPreservesGoroutine

polinasok:
Based on [proc_test](https://github.com/go-delve/delve/blob/c223ef656da3ddbab3a58ea87108d4217731a01c/pkg/proc/proc_test.go#L4969-L4977), looks like termination is a possible and acceptable outcome (which makes sense since the breakpoint is hit at the very last goroutine, so all others might already be wrapping up or done).


aarzilli (approved):
LGTM

PR #2418, opened 2021-04-07
dap: use larger string type variable load limits in 'repl', 'variables' context

hyangah:
When evaluate requests are triggered in the context of 'repl' (DEBUG CONSOLE in VSCode) or 'variables' (Copy Value from VARIABLES section in VSCode), they are triggered by one-off human action and loading more data isn't too bad.

Variable auto-loading for strings or arrays is nice but currently it's unclear to me how this should be integrated in the DEBUG CONSOLE or with the Copy Value feature. Until we have better ideas and tools, let's go with these larger limits.

Unfortunately, the "Copy Value" from WATCH section triggers evaluate requests with "watch" context and we don't want to load large data automatically for "watch" yet. We can consider clipboard capability instead later. (Followup change)

Also applied a more generous max string limit when evaluating function calls - since the results can be volatile, it's not guaranteed that we'd be able to reload after the injected call is complete, and the current string limit (64) is too small in many cases.

Updates golang/vscode-go#1318
Updates golang/vscode-go#126


hyangah:
cc @polinasok @suzmue

service/dap/server.go
```diff
@@ -1159,18 +1171,18 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 // variables request can be issued to get the elements of the compound variable. As a
 // custom, a zero reference, reminiscent of a zero pointer, is used to indicate that
 // a scalar variable cannot be "dereferenced" to get its elements (as there are none).
-func (s *Server) convertVariable(v *proc.Variable, qualifiedNameOrExpr string) (value string, variablesReference int) {
-	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, false)
+func (s *Server) convertVariable(v *proc.Variable, qualifiedNameOrExpr string, loadOption proc.LoadConfig) (value string, variablesReference int) {
```
polinasok, line 1174: s/loadOption/loadConfig/

service/dap/server.go
```diff
@@ -1221,7 +1233,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 				// frame-independent address-based expression, not fully-qualified name.
 				loadExpr := fmt.Sprintf("*(*%q)(%#x)", typeName, v.Addr)
 				s.log.Debugf("loading %s (type %s) with %s", qualifiedNameOrExpr, typeName, loadExpr)
-				vLoaded, err := s.debugger.EvalVariableInScope(-1, 0, 0, loadExpr, DefaultLoadConfig)
+				vLoaded, err := s.debugger.EvalVariableInScope(-1, 0, 0, loadExpr, loadConfig)
```
polinasok, line 1236: It's interesting that the auto-loading limits change depending on whether you are expanding a Variable from Variables pane or from expression. This deserve a test case. 

I am also wondering now if we should treat all variable expansions this way. If a user manually tries to expand a variable, then that's a single human action as well. At the same time, if the user leaves a field expanded and steps, it stay expanded and will be automatically reloaded, so maybe this is more like a watch and we should be conservative.

polinasok (commented):
> When evaluate requests are triggered in the context of 'repl'
> (DEBUG CONSOLE in VSCode) or 'variables' (copy values from VARIABLES

s/copy values/"Copy Value"/

service/dap/server.go
```diff
@@ -95,16 +95,28 @@ var defaultArgs = launchAttachArgs{
...
+var (
+	DefaultLoadConfig = proc.LoadConfig{
+		FollowPointers:     true,
+		MaxVariableRecurse: 1,
+		MaxStringLen:       64,
+		MaxArrayValues:     64,
+		MaxStructFields:    -1,
+	}
+
+	InteractiveEvaluateLoadConfig = proc.LoadConfig{
```
aarzilli, line 112: These are **huge**. Consider what happens with this:
```
v := make([]interface{}, 4096)
for i := range v {
    v[i] = &v
}
```

service/dap/server_test.go
```diff
@@ -1756,96 +1755,109 @@ func TestEvaluateRequest(t *testing.T) {
...
-					// Struct member access (i.e. somevar.memberfield)
-					client.EvaluateRequest("ms.Nest.Level", 1000, "this context will be ignored")
-					got = client.ExpectEvaluateResponse(t)
-					expectEval(t, got, "1", noChildren)
+							// Variable lookup that's not fully loaded
+							client.EvaluateRequest("ba", 1000, evalContext)
+							got = client.ExpectEvaluateResponse(t)
+							want := "(loaded 64/200) []int len: 200, cap: 200, [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,...+136 more]"
+							if evalContext == "repl" || evalContext == "variables" { // larger load limit is applied.
+								want = "[]int len: 200, cap: 200, [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]"
```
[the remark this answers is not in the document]
hyangah: Yes, that's the only case. And, now I reverted that, but left the table-driven test change there to test with different contexts. To make sure behavior of other type evaluation wasn't affected.

service/dap/server.go
hyangah: obsolete.

service/dap/server.go
hyangah: Now narrowed the scope where this heuristic applies further and this applies only to the string type, and only from the evaluate request handler. So, convertVariableWithOpts isn't affected.

I'd go safe and stay with short limit now. I think, eventually users have to look into the data further through the repl mode. Allowing to sett different limits for the repl mode and fully supporting the dlv cli's expression syntaxes, and then ask users to use dlv expression to look into further from there may be more helpful and good for users for long term..

service/dap/server.go
hyangah: Point taken. Now applies only when the final value type is string.

service/dap/server.go
```diff
@@ -97,26 +97,26 @@ var defaultArgs = launchAttachArgs{
...
-	InteractiveEvaluateLoadConfig = proc.LoadConfig{
-		FollowPointers:     true,
-		MaxVariableRecurse: 64,
-		MaxStringLen:       4096,
-		MaxArrayValues:     4096,
-		MaxStructFields:    -1,
-	}
-)
+// For evaluate requests made in the context of 'repl' or 'variables', they are triggered
+// by explicit human action and the results can be displayed in a larger space (e.g. DEBUG CONSOLE)
```
polinasok, line 110: I think the reason is not the space, but the fact that this is a one-off request, not something that might be automatically triggered many times every time a breakpoint is hit. The user will know that the variable is large and be tolerant of the performance implications in this case.

service/dap/server.go
```diff
@@ -97,26 +97,26 @@ var defaultArgs = launchAttachArgs{
...
 
-	InteractiveEvaluateLoadConfig = proc.LoadConfig{
-		FollowPointers:     true,
-		MaxVariableRecurse: 64,
-		MaxStringLen:       4096,
-		MaxArrayValues:     4096,
-		MaxStructFields:    -1,
-	}
-)
+// For evaluate requests made in the context of 'repl' or 'variables', they are triggered
```
polinasok, line 109: Grammar is off here. Sorry for being nitpicky, but every time I read this comment, it throws me off. Should be "Evaluate requests made in the context of 'repl' or 'variables are triggered" by (drop "For" and ", they")

service/dap/server.go
```diff
@@ -97,26 +97,26 @@ var defaultArgs = launchAttachArgs{
...
+// by explicit human action and the results can be displayed in a larger space (e.g. DEBUG CONSOLE)
+// or copied in memory, so we use InteractiveEvaluateLoadConfig to apply larger limits
+// if the result is represented as a string. We won't apply this heuristic to `call` command.
+var InteractiveEvaluateLoadConfig = proc.LoadConfig{
+	FollowPointers:     true,
+	MaxVariableRecurse: 1,
+	MaxStringLen:       1024,
+	MaxArrayValues:     1024,
```
polinasok, line 117: The commit name says that you are applying larger limit to string only, but you are also applying it to arrays. Is this intentional? Also, there is no need to match them. Extra characters should be much cheaper than extra objects.

service/dap/server.go
```diff
@@ -1191,7 +1191,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 	}
 	value = api.ConvertVar(v).SinglelineString()
 	if v.Unreadable != nil {
-		return
+		return value, variablesReference
```
polinasok, line 1194: why?

service/dap/server.go
```diff
@@ -1376,23 +1371,32 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
 	} else { // {expression}
-		exprVar, err := s.debugger.EvalVariableInScope(goid, frame, 0, request.Arguments.Expression, loadConfig)
+		exprVar, err := s.debugger.EvalVariableInScope(goid, frame, 0, request.Arguments.Expression, DefaultLoadConfig)
 		if err != nil {
 			s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
 			return
 		}
+		// If we are in interactive mode (probably with sufficient space to display long strings),
+		// load the string with a more generous load config.
+		if ctxt := request.Arguments.Context; (ctxt == "repl" || ctxt == "variables") && exprVar.Kind == reflect.String {
```
polinasok, line 1389: Do you mean to do this for scalar string vars only? Weren't users who complained about this feature storing their strings in some kind of data structures? For example, https://github.com/golang/vscode-go/issues/1318. This might not be unintuitive. We would still be limited by recursion limit and array limits (if you don't change them). Have you actually tried testing this to see if it would be too much?

service/dap/server.go
```diff
@@ -1376,23 +1371,32 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+			if strVal := constant.StringVal(exprVar.Value); exprVar.Len > int64(len(strVal)) {
```
polinasok, line 1390: I think just len(exprVal.Value) should work.

service/dap/server_test.go
```diff
@@ -1893,6 +1884,121 @@ func TestEvaluateRequest(t *testing.T) {
...
+							wantLong:  fmt.Sprintf(`"%v"`, strings.Repeat("a", 1024)),
+						},
+						{
+							expr:      "string(bytes)",
+							wantShort: fmt.Sprintf(`"%v"`, strings.Repeat("a", 33)),
+							wantLong:  fmt.Sprintf(`"%v"`, strings.Repeat("a", 33)),
+						},
+						// EvalVariableInScope doesn't seem to use load config when
+						// evaluating []byte to string conversion. Probably not critical
+						// but figure out why.
```
polinasok, line 1960: Add TODO?

service/dap/server.go
```diff
@@ -1260,7 +1272,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 			variablesReference = maybeCreateVariableHandle(v)
 		}
 	}
-	return
+	return value, variablesReference
```
polinasok, line 1275: ditto

polinasok (changes requested):

service/dap/server.go
polinasok: I have been thinking more about this. I guess with proper documentation, we could advertise loading specific string fields with "Copy Value" and "Copy as Expression". And this would certainly make it safer performance-wise.

polinasok:
I found another issue for this: https://github.com/golang/vscode-go/issues/126

service/dap/server.go
hyangah: https://tour.golang.org/basics/7

"Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions."

service/dap/server.go
hyangah: I learned that expanded variables remain open and kept retrieved. In that case, loading all automatically may not be ideal. We need more careful look on the behavior of Variables section when implementing autoloading of strings and slices/arrays/large structs.

service/dap/server.go
hyangah: updated.

service/dap/server.go
hyangah: Outdated code.

service/dap/server.go
hyangah: This is focusing only on string value retrieved with evaluate + copy value context only.

service/dap/server_test.go
hyangah: []byte is not handled in this PR any more so the code is gone.

hyangah (commented):
Given that we recommended users to use the WATCH section to conveniently convert bytes to string or call .String or .Error methods, I feel that excluding evaluations triggered by WATCH isn't ideal. How bad would it be if we just retrieve all strings for 'evaluate' requests?

Alternatively, I can enable clipboard mode - with that vscode sends evaluate requests with 'clipboard' mode when user clicks Copy Value (both in VARIABLES and WATCH modes).

service/dap/server.go
hyangah: exprVal.Value is constant.Value.

service/dap/server.go
```diff
@@ -1646,7 +1662,16 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
...
 		}
 	case reflect.String:
-		// TODO(polina): implement auto-loading here
+		if evalContext == "repl" || evalContext == "variables" {
+			// TODO(hyangah): Copy Value from WATCH panel triggers an evaluate request in 'watch' mode,
+			// not in 'variables' mode. Either apply this in watch mode or after enabling the clipboard support,
+			// apply this in clipboard mode. When clipboard support is enabled, Copy Value from vscode VARIABLES
+			// panel also uses 'clipboard' mode. (Not sure where 'variables' mode will be used after then).
+			if strVal := constant.StringVal(v.Value); v.Len > int64(len(strVal)) {
+				value = reloadVariable(v, qualifiedNameOrExpr)
```
aarzilli, line 1671: Is this code path going to run on the variables returned by `s.debugger.FunctionArguments`? If it, is it will stop working really soon.

aarzilli (commented):
@hyangah you can't do any further loading of the return values of an injected call, by the time you get them they are gone. You get what you ask for in ReturnInfoLoadConfig and that's it.

service/dap/server.go
hyangah: Yes, I can imagine users would want to inspect function args further by querying from Debug Console.
What will be the new recommendation to achieve that in the future? 
Will the change affect other part of the reloadVariable too, or only be specific to string type args?

hyangah (commented):
> @hyangah you can't do any further loading of the return values of an injected call, by the time you get them they are gone. You get what you ask for in ReturnInfoLoadConfig and that's it.

That's what I am thinking too. I guess we should just set a different limit when evaluating function call requests if we don't want to allow users to configure that.

service/dap/server.go
aarzilli: It won't work for anything that is stored in a register, I'm not sure if there's any other code path that calls reloadVariable on a variable returned by `s.debugger.FunctionArguments`. It should use the `loadValue` method of `*proc.Variable` instead of calling EvalVariableInScope.

hyangah: This specific code path will work with the variable returned by `s.debugger.EvalVariableInScope` from https://github.com/go-delve/delve/blob/66782f9b27318b39130032f840cd684832ee05f8/service/dap/server.go#L1799. When evaluating a func arg name, does `EvalVariableInScope` do anything special to handle variable stored in a register?

`s.debugger.FunctionArguments` is used when handling the Scopes request (https://github.com/go-delve/delve/blob/66782f9b27318b39130032f840cd684832ee05f8/service/dap/server.go#L1314). It's possible that the user expands the args, editor sends Variables requests, `convertVariable` is called using the variable returned from the prior `FunctionArguments` call.  The string reload added in this PR will not be applied in this execution path.

TODO [here](https://github.com/go-delve/delve/blob/66782f9b27318b39130032f840cd684832ee05f8/service/dap/server.go#L1572 ) mentions that and it now sounds like a requirement for 1.17, not just an optimization.

aarzilli: > When evaluating a func arg name, does EvalVariableInScope do anything special to handle variable stored in a register?

Yes, a lot, and the address of the variable isn't going to be a real address so it can't be passed back to EvalVariableInScope like reloadVariable does.

aarzilli: > TODO here mentions that and it now sounds like a requirement for 1.17, not just an optimization.

My impression was that before this commit reloadVariable will never be called on variables returned by FunctionArgumens, only their children variables, which would be fine.

hyangah:
Moved the reloading part back to onEvaluateRequest to make it clear it doesn't use variables from `FunctionArguments`. 
Also, unconditionally set the string limit to a larger value when handling function call. Should we adjust other configuration to avoid accidentally loading too much data? 

Have no idea how this change makes other tests fail...  though.

service/dap/server.go
hyangah: Moved the logic out of this reloadVariable part to make it clear the code path added by this PR doesn't run as part of  general variable reloading.

aarzilli (commented):
TestAttachStopOnEntry has an unrelated bug that makes it fail occasionally (there's a PR to fix it), TestFrameEvaluation is flaky on arm64 due to cgo involvement.
I don't know about TestPreSetBreakpoint, someone familiar with it should take a look.

suzmue:
>TestAttachStopOnEntry has an unrelated bug that makes it fail occasionally (there's a PR to fix it), TestFrameEvaluation is flaky on arm64 due to cgo involvement.
I don't know about TestPreSetBreakpoint, someone familiar with it should take a look.


The TestPreSetBreakpoint test is flaky and needs to be fixed, but is unrelated to this change. The threads request arrived after the debugger had stopped so the test did not get the dummy response it was expecting.

service/dap/server.go
polinasok: I agree with narrowing the scope, starting small and considering expansions based on user feedback. I too was worried about the impact of reloading due to sticky expansion (see comment from Apr 6). Btw, we already load large structs completely because MaxStructFields defaults to -1.

service/dap/server.go
polinasok: Ah, thanks for the pointer. If readability is at stake here, then probably "return value, 0" would be even better.

polinasok (commented):
> Given that we recommended users to use the WATCH section to conveniently convert bytes to string or call .String or .Error methods, I feel that excluding evaluations triggered by WATCH isn't ideal. How bad would it be if we just retrieve all strings for 'evaluate' requests?
> 
> Alternatively, I can enable clipboard mode - with that vscode sends evaluate requests with 'clipboard' mode when user clicks Copy Value (both in VARIABLES and WATCH modes).

I am not familiar with clipboard mode. I am not sure how it is different from Copy Value and such.

> --
> Edit: I just did a quick test on how to load more data when a function call expression in WATCH section contains long strings. Currently proc.Variable with names like ~r1 or ~r2 don't get evaluateName, so they can't be reevaluated for Copy Value (VSCode just sends the value as an expression :-() Address based expression won't work reliably either - e.g. if there are multiple function calls registered. Some part of the return values are temporary, I don't know how loading/reevaluating on demand should be implemented.

That's right. Watch uses expressions, not variables, so no evaluateName and hence no features that it enables.

> Assuming users want injected function call to stringify some values, should we just go with a generous return value loading limit and present as much data as we could up front? What do you think?

I think this PR already applies relaxed limits to all call instances, including WATCH. I don't know how to determine what the magic value is for "as much data as we could". I did like your earlier suggestion to start small and expand based on user feedback. We could wait for this feedback to come on its own, or try to solicit it actively. It is always easier to relax things than to tighten them.

polinasok:
> > TestAttachStopOnEntry has an unrelated bug that makes it fail occasionally (there's a PR to fix it), TestFrameEvaluation is flaky on arm64 due to cgo involvement.
> > I don't know about TestPreSetBreakpoint, someone familiar with it should take a look.
> 
> The TestPreSetBreakpoint test is flaky and needs to be fixed, but is unrelated to this change. The threads request arrived after the debugger had stopped so the test did not get the dummy response it was expecting.

Yup, I saw that too. Are you looking into it or should I?

suzmue:
> > > TestAttachStopOnEntry has an unrelated bug that makes it fail occasionally (there's a PR to fix it), TestFrameEvaluation is flaky on arm64 due to cgo involvement.
> > I don't know about TestPreSetBreakpoint, someone familiar with it should take a look.
> > 
> > The TestPreSetBreakpoint test is flaky and needs to be fixed, but is unrelated to this change. The threads request arrived after the debugger had stopped so the test did not get the dummy response it was expecting.
> > 
> 

> Yup, I saw that too. Are you looking into it or should I?
> 

Sent a PR https://github.com/go-delve/delve/pull/2483

polinasok:
Oh and I think the PR description and title need to be updated to focus on strings only.

service/dap/server.go
```diff
@@ -1730,12 +1730,23 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 			s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
 			return
 		}
+
+		// The return values of injected function calls are volatile.
+		// Load as much useful data as possible.
+		// - String: a common use case of call injection is to stringify complex
+		// data conveniently. That can easily exceed the default limit, 64.
```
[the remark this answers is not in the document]
hyangah: IMO, it's better to relax the string type variable length limit for non-call evaluate requests sent with watch context too (as long as MaxVariableRecurse remains 1 and we relax the limit only on watch mode).

Watch is more like a short cut to automate the evaluation requests that users would've run through DEBUG CONSOLE if Watch wasn't available. It will be more convenient if they can see the full values without copying or running them again from DEBUG CONSOLE. Users can remove/edit them any time (unlike those in Variables section) if they find they are loading too much and slowing down the overall performance.

1K doesn't seem a lot, but one thing that we still need to address is to limit the size of the value string for compound types (https://github.com/golang/vscode-go/issues/1435). For map or struct with many long strings, I am afraid this can be huge. I think we can investigate it in a separate PR to address the issue and try to truncate the final value string.

hyangah:
> I am not familiar with clipboard mode. I am not sure how it is different from Copy Value and such.

Once we enable the clipboard mode, `Copy Value` will trigger clipboard context evaluate requests, Not 'watch' or 'variables' context any more.

> > Assuming users want injected function call to stringify some values, should we just go with a generous return value loading limit and present as much data as we could up front? What do you think?
> 
> I think this PR already applies relaxed limits to all call instances, including WATCH. I don't know how to determine what the magic value is for "as much data as we could". I did like your earlier suggestion to start small and expand based on user feedback. We could wait for this feedback to come on its own, or try to solicit it actively. It is always easier to relax things than to tighten them.

1K doesn't seem a lot. From google search, I see users choose values [300, ridiculously-large-value] ranges.

hyangah:
Looks like an unrelated test failure
Screen Shot 2021-05-18 at 11 09 14 AM

aarzilli (approved):

derekparker:
@polinasok I'll wait for your approval before final review / merge.

polinasok (approved):
LGTM

PR #2422, opened 2021-04-08
service/dap: fallback to CurrentThread if SelectedGouroutine is not set

polinasok:
Also add missing nil checks


aarzilli (approved):
LGTM but I'm pretty sure this makes no difference, it will always be 0.

Issue #2322, opened 2021-01-23
accept-multiclient: concurrent clients can get the server and each other stuck


polinasok:
When using `--accept-multiclient`, it is not immediately obvious that concurrent actions of other clients won't be properly reflected at each cli. In particular, when one client enters "continue", its prompt will disappear indicating a running state. But the other client will still have its prompt and can enter more commands that will get buffered and then cause unexpected behavior.

Adding event notifications or special handling in the server to reject client requests when in certain states is too much of an undertaking. I am not even sure how widely used this use case is. So I would propose highlighting this topic in the documentation and maybe printing warnings when a new clients connects that there is already another client connected and that care is needed to coordinate between clients.

Below is an example, where I ran into a situation where things got stuck. I played with this on my Mac, but I think the high-level concerns expressed below are universal, even if the exact specifics of Ctrl-C, etc. behavior vary from system to system.
```
$ dlv version
Delve Debugger
Version: 1.5.1
Build: $Id: bca418ea7ae2a4dcda985e623625da727d4525b5 $
$ go version
go version go1.14.3 darwin/amd64
```

1. Start long-running program (something with infinite loop works)
2. Attach server with `--headless --accept-multiclient`
3. The program is now suspended
4. Connect client A with` dlv connect`
5. Connect client B with `dlv connect`
6. Client A enters "c" to continue and waits for it to return
```
2021-01-22T17:45:52-08:00 debug layer=debugger continuing
```
7. Client B still shows a prompt as if the session is still suspended
8. Client B enters "c" to continue
9. Server receives 2nd continue and buffers it
9. Both clients are blocked with no prompt
10. Client B enters Ctrl-C to halt:
```  
Would you like to [p]ause the target (returning to Delve's prompt) or [q]uit this client (leaving the target running) [p/q]? p
```
11.  Buffered 2nd continue kicks in on the server side.
```
2021-01-22T17:48:35-08:00 debug layer=debugger halting
2021-01-22T17:48:35-08:00 debug layer=debugger continuing
```
12. Client B does not get its prompt back from the 2nd continue.
13. Client A does get its prompt back from the 1st continue.
13. Client A issues "locals" and blocks
14. Both clients are blocked. Ctrl-C in either window does nothing.
15. If I connect another client, I can suspended the execution and let all previous commands return. But even after that all the previous halt attempts kicked in and trying to navigate the prompts, I entered something that made one of the clients quit and the terminal enter some weird mode
[...]

PR #2423, opened 2021-04-08
service/dap: support running requests asynchronously

polinasok:
Currently the server processes one request at a time synchronously. That means that while continue (or next, or call, etc) are running, he server will not respond to any additional requests. This change updates the server to start a separate goroutine for running requests and resume the request loop to accept other requests. This will allow us to process requests for pause, setting breakpoints and disconnecting while the session is running, and return explicit errors for requests that cannot be processed while running (e.g. evaluating expressions).

Updates #1515
Updates https://github.com/golang/vscode-go/issues/1347


service/dap/server.go
```diff
@@ -433,7 +464,9 @@ func (s *Server) handleRequest(request dap.Message) {
 func (s *Server) send(message dap.Message) {
 	jsonmsg, _ := json.Marshal(message)
 	s.log.Debug("[-> to client]", string(jsonmsg))
+	s.sending.Lock()
 	_ = dap.WriteProtocolMessage(s.conn, message)
+	s.sending.Unlock()
```
polinasok, line 469: I went with simple lock here like in the rpc server. I think I need to tweak things a bit more to make sure it doesn't conflict with mu (e.g. in Stop()). An alternative would be to use a channel for all the sends and to have a dedicated goroutine that reads from that channel and sends over the connection. WDYT?

service/dap/server.go
```diff
@@ -677,6 +710,8 @@ func (s *Server) stopDebugSession(killProcess bool) {
 	if s.debugger == nil {
 		return
 	}
+	// Halting will stop any debugger command that's pending on another
+	// per-request goroutine, hence unblocking that goroutine to wrap-up and exit.
```
polinasok, line 714: Technically these additional goroutines could still be not done when this one is. Should we worry about goroutine leaks? Not worth adding a wait group?

service/dap/server_test.go
```diff
@@ -2172,6 +2196,8 @@ func testStepParkedHelper(t *testing.T, client *daptest.Client, fixture protest.
 	return goroutineId
 }
 
+// TestStepOutPreservesGoroutine is inspired by proc_test.TestStepOutPreservesGoroutine
+// and checks that StepOut preserves the currently selected goroutine.
```
polinasok, line 2200: This is a diff from https://github.com/go-delve/delve/pull/2417. I did not merge for the merge to show up as its own commit. Oops.

service/dap/daptest/client.go
```diff
@@ -50,18 +50,18 @@ func (c *Client) ReadMessage() (dap.Message, error) {
 	return dap.ReadProtocolMessage(c.reader)
 }
 
-func (c *Client) expectReadProtocolMessage(t *testing.T) dap.Message {
+func (c *Client) ExpectMessage(t *testing.T) dap.Message {
```
polinasok, line 53: This is a diff from https://github.com/go-delve/delve/pull/2421.

aarzilli:
It seems to me that this approach would lead to a lot of repeated code. Wouldn't it be easier to handle as much as possible of this inside (*Server).handleRequest?

For example:


```
func (s *Server) handleRequest(request dap.Message) {
	defer func() {
		// In case a handler panics, we catch the panic and send an error response
		// back to the client.
		if ierr := recover(); ierr != nil {
			s.sendInternalErrorResponse(request.GetSeq(), fmt.Sprintf("%v", ierr))
		}
	}()

	jsonmsg, _ := json.Marshal(request)
	s.log.Debug("[<- from client]", string(jsonmsg))
	
	// Requests that can be handled regardless of whether the targret is running
	switch request := request.(type) {
	case *dap.PauseRequest:
		//... do Command(api.Halt)...
		return
	}
	
	s.runningMu.Lock()
	defer s.runningMu.Unlock()
	
	if s.running {
		// Requests that can be handled while the target is running, or return an error
		switch request := request.(type {
		case *dap.ThreadsRequest:
			response := &dap.ThreadsResponse{
				Response: *newResponse(request.Request),
				Body:     dap.ThreadsResponseBody{Threads: []dap.Thread{{Id: 1, Name: "Dummy"}}},
			}
			s.send(response)
		default:
			s.sendErrorResponse(request.Request, DebuggeeIsRunningError, "Unable to produce stack trace", "debuggee is running")
		}
		return
	}
	
	// Requests that can only be handled while stopped
	switch request := request.(type) {
	case *dap.ContinueRequest:
		go s.onContinueRequest(request, resumeRequestLoop)
		<-resumeRequestLoop
		s.running = true // will be set back to false at the end of doCommand
	//... other request handers, mostly unchanged...
	}
}
```

I think this is also less error prone. For example, it seems to me, that this PR will accept two consecutive ContinueRequests and the second one will block until the first one terminates and then start immediately. Is that the intended behavior?

If at some point you decide to actually serve requests that happen while the target is running, for example if you want to make SetBreakpointsRequest:

- stop the target
- set breakpoints
- resume the target

the stop and resume logic can be handled inside the `if s.running` branch, and be the same for every request.

polinasok:
@hyangah

polinasok:
@suzmue

service/dap/server.go
polinasok: This will prevent any accidental lock conflicts and also protect us from very slow network sends.

hyangah (commented):
My current recommendation is to maintain the existing handleRequest structure (before this PR), but have a single separate, dedicated goroutine that processes the remaining synchronous operations sequentially. 

Looking into Debugger.Command implementation, I found `resumeNotify` is closed after acquiring `d.targetMutex` and `Debugger.Command` will still hold the lock until it returns. That means, even when the subsequent DAP request and the `Debugger.Command` or `Debugger.Goroutines` to serve the request runs in a separate goroutine, the goroutine will be blocked waiting for `d.targetMutex.Lock` and the handleRequest loop will be blocked on `resumeRequestLoop`. 

So, completely going asynchronous looks difficult to me. 

Moreover, there is no ordering guarantee when multiple goroutines are fighting over a mutex. I am not sure if that's desirable or not yet, but if we can preserve the ordering by enforcing those operations in a single goroutine, it would be nice and more readable.

Having only a bounded number of goroutines is also good when we need to do cleanup.

Btw I observed that `onEvaluateRequest` still runs in the Run goroutine after this PR, but if it's for `call`, this calls `Debugger.Command`. I expect the same for `onSetVariableRequest` too, and wonder if they are safe to run in the Run goroutine.

polinasok (commented):
> It seems to me that this approach would lead to a lot of repeated code. Wouldn't it be easier to handle as much as possible of this inside (*Server).handleRequest?

In theory, if the client is doing its job, most of the requests would never arrive while things are running. But I agree that it is safer and simpler to just handle all this in bulk for everything. 

> I think this is also less error prone. For example, it seems to me, that this PR will accept two consecutive ContinueRequests and the second one will block until the first one terminates and then start immediately. Is that the intended behavior?

This is the behavior in the rpc server, so I assumed that I should do the same at least to start with :) You can observe this with vscode. The continue button will be disabled while continuing, but you can still trigger it from the keyboard. Another way to do this is to have to clients with --accept-multiclient. We discussed this here: https://github.com/go-delve/delve/issues/2322.

But yes, I agree that we should just reject anything that doesn't make sense explicitly.

polinasok:
> My current recommendation is to maintain the existing handleRequest structure (before this PR), but have a single separate, dedicated goroutine that processes the remaining synchronous operations sequentially.

I looked into this and decided that it doesn't make sense. I added comments to HandleRequest, explaining why. Please let me know if I misunderstood something in your proposal.

> Btw I observed that onEvaluateRequest still runs in the Run goroutine after this PR, but if it's for call, this calls Debugger.Command. I expect the same for onSetVariableRequest too, and wonder if they are safe to run in the Run goroutine.

You are right. This one is a bit more nuanced than others, so I wanted to deal with it separately once we figure out the design with the simpler items. There is a related TODO under call that's also waiting for async support. But specifically about the issue you are asking about - yes the call will block processing any new requests. I think that's fine for assignments because they should be quick. But it is not fine if you call some long-running function because you won't be able to pause or even disconnect. Besides making it asynchronous, we would also need to issue a Continue event to let the editor know that we are executing, so it updates the UI accordingly. And in that case, I will need to add a proper stopped event when things stop.

polinasok:
The failure is related and I am looking into it.

service/dap/server.go
```diff
@@ -1673,13 +1830,16 @@ func (s *Server) resetHandlesForStoppedEvent() {
...
-		return
-	}
+// event needs to be sent to the client. asyncSetupDone is
+// a channel that will be closed to signal that an
+// asynchornous command has completed setup or was interrupted
+// due to an error, so the server is ready to receive new requests.
+func (s *Server) doCommand(command string, asyncSetupDone chan struct{}) {
+	defer func() {
+		s.asyncCommandDone(asyncSetupDone)
+	}()
```
hyangah, line 1840: defer s.asyncCommandDone(asyncSetupDone)

Wontt it be clear not to pass asyncSetupDone to asyncCommandDone given that it should be closed by s.debugger.Command in this case.

service/dap/server.go
```diff
@@ -325,53 +353,114 @@ func (s *Server) handleRequest(request dap.Message) {
...
+
+	switch request := request.(type) {
+	//--- Asynchronous requests ---
 	case *dap.ConfigurationDoneRequest:
 		// Optional (capability ‘supportsConfigurationDoneRequest’)
-		// Supported by vscode-go
-		s.onConfigurationDoneRequest(request)
+		go s.onConfigurationDoneRequest(request, resumeRequestLoop)
+		<-resumeRequestLoop
+		s.running = true
```
hyangah, line 443: stopOnEntry=true case?

service/dap/server.go
```diff
@@ -325,53 +353,114 @@ func (s *Server) handleRequest(request dap.Message) {
...
+	// --1-- Return a dummy response or an error right away.
+	// --2-- Halt execution, process the request, resume execution.
+	// TODO(polina): do this for setting breakpoints
+	// --3-- Handle such requests asynchronously and let them block until
+	// the process stops or terminates (e.g. using a channel and a single
+	// goroutine to preserve the order). This might not be appropriate
+	// for requests such as continue or step because they would skip
+	// the stop, resuming execution right away. Other requests
+	// might not be relevant anymore when the stop is finally reached, and
+	// state changed from the previous snapshot. The user might want to
```
hyangah, line 399: Thanks for the detailed explanation. The first part of the current version matches what I expected - processing the requests that can be served while the program is running. I wanted to avoid many goroutines fighting for the lock. But now it returns err for all requests (except threads and breakpoints in the future),  so there will be effectively only one or two extra goroutine outstanding in any given time and we don't have to worry about requests' ordering.

service/dap/server.go
```diff
@@ -325,53 +353,114 @@ func (s *Server) handleRequest(request dap.Message) {
...
+	// resume execution before the backlog of buffered request is cleared,
+	// so we would have to either cancel them or delay processing until
+	// the next stop. In addition, the editor itself might block waiting
+	// for these requests to return. We are not aware of any requests
+	// that would benefit from this approach at this time.
+	if s.running {
```
hyangah, line 405: @aarzilli the delve itself keeps track of this already. Is it not possible to expose IsRunning instead of keeping track of the state here with a separate set of lock?

polinasok: We could use State() or export isRunning() to avoid getting more state than we need. However, I think we need to lock more than just the bool, but also some related logic, so things cannot change after you flip the status, but before you do all the processing relying on the status being that. I am still revising the code to make sure I get this right. I don't think the current version is doing everything correctly.

service/dap/server.go
polinasok: It's handled inside of the handler. I am thinking of "running" as "running an async request', not just "running the target". And as I mentioned above, I still need to clean this up a bit.

service/dap/server.go
polinasok: Hmm, I am not sure anymore why I have it there. I think there was some intermediate version of this code, which had some error check and early return before Command was called. I think I lost nil debugger check in all the reshuffling. I should put it back, since it might be unsafe to rely on the editor not to issue certain commands when in noDebug mode or not past launch/attach.

polinasok (commented):
Thank you for a speedy review.

service/dap/server.go
polinasok: That would be equivalent to locking s.running when we read it, unlocking, then locking again when we write it, instead of locking the entire block of code that applies logic related to the status we are reading/writing. Is that sufficient? A sync request cannot be interrupted by a new request because it blocks the request loop. No issue there. But an async request can be interrupted by a sync request while running. So then we have the following
- request loop dispatches async request, sets running to true (locking more or less doesn't matter)
- things are running
- another request comes in
- option 1: lock, see that running==true, unlock
- option 2: lock, see that running==true, keep holding the lock
- async request stops running
- sync request still thinks things are running
- option 1:  async sends stopped event, locks, running=false, unlocks, sync sends is-running error (user experience is confusing - why am I getting an error when the session has clearly stopped?)
- option 2: async request tries to get the lock before sending stopped event AND setting running=false, lock is still held by the sync request until it is done sending the error response, so the error and the stopped event arrive in the right order
Or am I missing something?
That said, I have changed my code locally to do this and got new test failures that I am still trying to sort out.

aarzilli: From the point of view of a meat and bones user either is going to look like "I pressed a button just as the application was about to stop" and it's all fine. From a mechanical user point of view the observable sequence is either:

```
← send synchronous request
→ receive error "program is running"
→ receive stop message
```

or:

```
← send synchronous request
→ receive stop message
→ receive error "program is running"
```

you want to prevent the latter but you can't because of the delay between the time where the request is sent and the time it is received and acquires the lock. For that to be impossible the client would have to be holding a lock before sending the request.

polinasok: > you want to prevent the latter but you can't because of the delay between the time where the request is sent and the time it is received and acquires the lock. For that to be impossible the client would have to be holding a lock before sending the request.

But if the goroutine that wants to send a stopped event has acquired the running lock, then the request loop that receives the sync request would not be able to process it until the lock is released, and at that point "running" will be false, so it will not need to return an error.

aarzilli: I don't think it matters, you are seeing the answer as if the request happened after the stop event, which it didn't.

polinasok: Too late. I already updated the code :) Does this not matter enough not to have to change it back? 
I agree that this is not the biggest issue here though getting a slightly delayed useful response seems better to me than getting a confusing error that's contradicting what is being shown otherwise.

service/dap/server.go
polinasok: Sometimes debugger.Command doesn't close this channel.

polinasok:
FYI I fixed the failing test, but do not fully understand how [loopprog](https://github.com/go-delve/delve/blob/master/_fixtures/loopprog.go) that has an infinite loop would terminate on halt. I was not able to reproduce this on either Mac or Linux machines that I have. Only saw this via TeamCity.

aarzilli:
> FYI I fixed the failing test, but do not fully understand how [loopprog](https://github.com/go-delve/delve/blob/master/_fixtures/loopprog.go) that has an infinite loop would terminate on halt. I was not able to reproduce this on either Mac or Linux machines that I have. Only saw this via TeamCity.

Seems strange.

service/dap/server.go
hyangah: Having multiple locks in play makes me uncomfortable and hurts readability and long term maintainability. So, it would be better if we can avoid the lock.

polinasok (commented):
Thank you for the educational comments.

aarzilli (approved):
LGTM

service/dap/server.go
```diff
@@ -1042,9 +1198,10 @@ func (s *Server) doStepCommand(command string, threadId int) {
 		stopped.Body.Reason = "error"
 		stopped.Body.Text = err.Error()
 		s.send(stopped)
+		s.asyncCommandDone(asyncSetupDone)
```
hyangah, line 1201: how about calling this as the very first one after if err != nil so even when s.send panics, s.asyncCommandDone runs and unblocks the handler loop?

hyangah (commented):
Thanks for making this work.

LGTM

service/dap/server.go
polinasok: Looks like I missed this question earlier inside of the resolved conversation.

The goroutine does have a clear path to exit. But if the server is shut down at the same time, the main goroutine could be done before the goroutine checks state, sends stopped event, etc after the running command returns. This is in theory. I haven't actually observed this. If we run this in a test, we could end up with the goroutine still doing things when the test is done. A while back I was advised against this by @eliben. We only really have one such goroutine here with the current design though.

service/dap/server.go
polinasok: Oh darn panics. I keep forgetting about them. In that case its safest to just defer. And then we handle the panic from SwitchGoroutine as well. If it's already unblocked, it will do nothing.

hyangah:
@aarzilli @derekparker It would be great if this can be merged soon so VS Code Go extension can handle Delve DAP debug session teardown cleanly. Thanks for the review!

derekparker (approved):
LGTM

PR #2424, opened 2021-04-08
debugger: report error when switching goroutine is impossible

aarzilli:
Due to variable shadowing the SwitchGoroutine command never failed.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -934,14 +934,16 @@ func stoppedGoroutineID(state *api.DebuggerState) (id int) {
 
 func (s *Server) doStepCommand(command string, threadId int) {
 	// Use SwitchGoroutine to change the current goroutine.
-	state, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
+	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
 	if err != nil {
-		s.log.Errorf("Error switching goroutines while stepping: %e", err)
+		s.log.Errorf("Error switching goroutines while stepping: %v", err)
 		// If we encounter an error, we will have to send a stopped event
 		// since we already sent the step response.
 		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
 		stopped.Body.AllThreadsStopped = true
-		if state != nil {
+		if state, err := s.debugger.State(false); err != nil {
+			s.log.Errorf("Error retrieving state: %e", err)
+		} else {
 			stopped.Body.ThreadId = stoppedGoroutineID(state)
 		}
 		stopped.Body.Reason = "error"
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -2077,9 +2077,12 @@ func TestNextAndStep(t *testing.T) {
 					client.ExpectStepInResponse(t)
 					expectStop("main.inlineThis", 5)
 
-					client.NextRequest(-10000 /*this is ignored*/)
+					client.NextRequest(-1000)
 					client.ExpectNextResponse(t)
-					expectStop("main.inlineThis", 6)
+					if se := client.ExpectStoppedEvent(t); se.Body.Reason != "error" || se.Body.Text != "unknown goroutine -1000" {
+						t.Errorf("got %#v, want Reaspon=\"error\", Text=\"unknown goroutine -1000\"", se)
+					}
+					handleStop(t, client, 1, "main.inlineThis", 5)
 				},
 				disconnect: false,
 			}})
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -1095,7 +1095,8 @@ func (d *Debugger) Command(command *api.DebuggerCommand, resumeNotify chan struc
 		withBreakpointInfo = false
 	case api.SwitchGoroutine:
 		d.log.Debugf("switching to goroutine %d", command.GoroutineID)
-		g, err := proc.FindGoroutine(d.target, command.GoroutineID)
+		var g *proc.G
+		g, err = proc.FindGoroutine(d.target, command.GoroutineID)
 		if err == nil {
 			err = d.target.SwitchGoroutine(g)
 		}


service/dap/server.go
```diff
@@ -882,22 +882,30 @@ func (s *Server) onStepOutRequest(request *dap.StepOutRequest) {
...
-		// If we encounter an error, we will have to send a stopped event
-		// since we already sent the step response.
-		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
-		stopped.Body.AllThreadsStopped = true
-		if state.SelectedGoroutine != nil {
-			stopped.Body.ThreadId = state.SelectedGoroutine.ID
-		} else if state.CurrentThread != nil {
-			stopped.Body.ThreadId = state.CurrentThread.GoroutineID
+	var err error
+	if threadId != 0 { // TODO: threadId == 0 could be valid the functions calling this should actually check whether or not the client specified a threadId argument.
```
polinasok, line 886: I am not exactly sure what this TODO suggests. For what it's worth, this is not an optional arg in the DAP spec.

service/dap/server.go
```diff
@@ -882,22 +882,30 @@ func (s *Server) onStepOutRequest(request *dap.StepOutRequest) {
...
+		_, err = s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
+		if err != nil {
+			s.log.Errorf("Error switching goroutines while stepping: %e", err)
+			// If we encounter an error, we will have to send a stopped event
+			// since we already sent the step response.
+			state, err := s.debugger.State(false)
+			if err != nil {
+				s.log.Errorf("Error retrieving state: %e", err)
+				return
```
polinasok, line 895: You should not just return here on error. Then the client will think that you succeeded and are still running the step command - see comment on line 890. It will mark the thread as running in the stack trace. You must translate the error into a stopped event to let the client know that nothing is running.

polinasok (changes requested):

service/dap/server.go
aarzilli: You are right, I must have read the wrong request. I'll remove the comment.

service/dap/server.go
aarzilli: Done.

service/dap/server.go
```diff
@@ -882,22 +882,27 @@ func (s *Server) onStepOutRequest(request *dap.StepOutRequest) {
...
+	var err error
+	if threadId != 0 {
+		_, err = s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
+		if err != nil {
+			s.log.Errorf("Error switching goroutines while stepping: %e", err)
+			// If we encounter an error, we will have to send a stopped event
+			// since we already sent the step response.
+			stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
+			stopped.Body.AllThreadsStopped = true
+			if state, err := s.debugger.State(false); err != nil {
```
polinasok, line 894: why State(false)?

service/dap/server.go
```diff
@@ -882,22 +882,27 @@ func (s *Server) onStepOutRequest(request *dap.StepOutRequest) {
...
-		// If we encounter an error, we will have to send a stopped event
-		// since we already sent the step response.
-		stopped := &dap.StoppedEvent{Event: *newEvent("stopped")}
-		stopped.Body.AllThreadsStopped = true
-		if state.SelectedGoroutine != nil {
-			stopped.Body.ThreadId = state.SelectedGoroutine.ID
-		} else if state.CurrentThread != nil {
-			stopped.Body.ThreadId = state.CurrentThread.GoroutineID
+	var err error
+	if threadId != 0 {
```
polinasok, line 886: So what is the meaning of this check? That we assume that value of 0 means it was not set because delve would never return 0 as a meaningful value? Why would we not just handle 0 as any other value?

DAP spec includes required thread ids for all running requests. For continue they say if you must continue on all threads, then just reflect that in the response, but for next and step in/out they assume that it is possible to only advance on that one thread. Then UI then marks the one thread where you are stepping as running and the rest as paused with UI options to step on them.

service/dap/server.go
aarzilli: The Command call immediately preceeding this will block until the target stops, there is no reason to do a non-blocking state call here.

service/dap/server.go
aarzilli: > Why would we not just handle 0 as any other value?

Removed the check.

polinasok (commented):
>The Command call immediately proceeding this will block until the target stops, there is no reason to do a non-blocking state call here.

I feel like I am missing some nuance here. Are we relying on the fact that the previous call returned, so things are definitely not running? Would that assumption change if we are in some kind async multi-client mode? When would we want to use a non-blocking version? Would a non-blocking call return different information here? And what's the downside of using one vs the other?

aarzilli:
> > The Command call immediately proceeding this will block until the target stops, there is no reason to do a non-blocking state call here.
> 
> I feel like I am missing some nuance here. Are we relying on the fact that the previous call returned, so things are definitely not running? Would that assumption change if we are in some kind async multi-client mode? When would we want to use a non-blocking version? Would a non-blocking call return different information here? And what's the downside of using one vs the other?

There's almost never a good reason to call the non-blocking version of State. It isn't a synchronization primitive, either you know if the target is running because you are controlling it or non-blocking State won't help you. You can grep for GetStateNonBlocking to see acceptable uses of non-blocking State. Basically: when what you are about to do can be done regardless of what GetStateNonBlocking says, as a QoL improvement.

service/dap/server_test.go
```diff
@@ -2060,7 +2060,7 @@ func TestNextAndStep(t *testing.T) {
 					client.ExpectStepInResponse(t)
 					expectStop("main.inlineThis", 5)
 
-					client.NextRequest(-10000 /*this is ignored*/)
```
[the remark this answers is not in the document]
aarzilli: Done.

service/dap/server_test.go
```diff
@@ -2060,7 +2060,13 @@ func TestNextAndStep(t *testing.T) {
...
+					client.NextRequest(-1000)
+					client.ExpectNextResponse(t)
+					if se := client.ExpectStoppedEvent(t); se.Body.Reason != "error" || se.Body.Text != "unknown goroutine -1000" {
+						t.Errorf("got %#v, want Reaspon=\"error\", Text=\"unknown goroutine -1000\"", se)
+					}
+
+					client.NextRequest(1)
```
polinasok, line 2069: I think it is better not to add extra next and just change last expectStop to the previous line 5.

polinasok (approved):
Minor nit, but otherwise LGTM

polinasok:
> GetStateNonBlocking

Then it sounds like we need to revisit the rest of the uses of GetState in the dap server.

aarzilli:
> > GetStateNonBlocking
> 
> Then it sounds like we need to revisit the rest of the uses of GetState in the dap server.

They're fine. It's just that you could turn them all to blocking and it wouldn't change anything.

service/dap/server_test.go
polinasok: I would also add the follow to verify that we reported a stop on the same thread at the same function location
```handleStop(t, client, 1, "main.inlineThis", 6)```

aarzilli: This would fail without `client.NextRequest(1)`.

polinasok: How does it fail? It shouldn't fail as far as I can tell.
If you want, you can add a TODO for me to sort this out later.

polinasok: It should be handleStop(t, client, 1, "main.inlineThis", 5) - the line from the previous next stop, since this one did not succeed.

aarzilli: The stop event returned by NextRequest(-1000) is already consumed by the ExpectStoppedEvent that checks for the error.

polinasok: I think you are confusing two different helpers - expectStop and handleStop. expectStop = ExpectStoppedEvent+checks+handleStop. I am proposing that you add handleStop, which won't try to consume any additional stopped events.

aarzilli: Got it.

polinasok (approved):
Thank you.LGTM.

derekparker (approved):
LGTM

PR #2425, opened 2021-04-08
service: fix breakpoint IDs after Restart with disabled breakpoints

aarzilli:
When restarting we must take care of setting breakpoint IDs correctly
so that enabled breakpoints do not end up having the same ID as a
disabled breakpoint, also we must make sure that breakpoints created
after restart will not get an ID already used by a disabled breakpoint.


derekparker (approved):
LGTM

Issue #2245, opened 2020-12-02
service/dap: set `seq` field in responses


polinasok:
When studying vscode-go logs, I saw that `seq` was only used for requests and `request_seq` was used for responses while `seq` always stayed 0. For example:
```
To client: {"seq":0,"type":"response","request_seq":1,"command":"initialize","success":true,"body":{"supportsConditionalBreakpoints":true,"supportsConfigurationDoneRequest":true,"supportsSetVariable":true}}
To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"API server listening at: 127.0.0.1:38993\n"}}
To client: {"seq":0,"type":"event","event":"initialized"}
To client: {"seq":0,"type":"response","request_seq":2,"command":"launch","success":true}
```
It turns out that this is an artifact of a bug in the parent code (https://github.com/microsoft/vscode-debugadapter-node/issues/239) and in fact increasing `seq` numbrers are used. It doesn't seem like this is hurting anything, so this is likely just for bookkeeping on our end.

PR #2427, opened 2021-04-09
service/dap: delay disconnect response and log teardown progress

polinasok:
Updates https://github.com/golang/vscode-go/issues/1410

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -113,6 +113,8 @@ func runTest(t *testing.T, name string, test func(c *daptest.Client, f protest.F
 //                                 : 12 << continue
 // - Program runs to completion    :    << terminated event
 //                                 : 13 >> disconnect
+//                                 :    << output event (Process exited)
+//                                 :    << output event (Detaching)
 //                                 : 13 << disconnect
 // This test exhaustively tests Seq and RequestSeq on all messages from the
 // server. Other tests do not necessarily need to repeat all these checks.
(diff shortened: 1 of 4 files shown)


aarzilli (approved):
LGTM

polinasok:
@hyangah @suzmue

service/dap/server_test.go
```diff
@@ -222,6 +224,14 @@ func TestLaunchStopOnEntry(t *testing.T) {
 
 		// 13 >> disconnect, << disconnect
 		client.DisconnectRequest()
+		oep := client.ExpectOutputEventProcessExited(t, 0)
+		if oep.Seq != 0 || oep.Body.Category != "console" {
```
hyangah, line 228: (beyond the scope of this pr) I wonder why check on response's Seq is necessary. DAP isn't specifying the semantic of it, is it an important property that affects the behavior and requires checking? (Afraid this is another case of change-detector tests )

hyangah (approved):

service/dap/server_test.go
polinasok: This was for consistency with the rest of the test. The idea was to test that we set all the counters correctly. There are a couple of tests that check counters, and the rest do not. Technically we should be setting these (#2245). It did not seem like it mattered for vscode, but looks like other clients might care (https://github.com/mfussenegger/nvim-dap/pull/116). We could define a variable for counter and just increment it with every message instead of hardcoding the numbers.

derekparker (approved):
LGTM

PR #2428, opened 2021-04-09
service/dap: annotate shadowed variable names in variables request

suzmue:
In order to distinguish variables that are shadowed, this change
updates the names from 'name' to '(name)'. This is the same syntax
used in the terminal package.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -981,8 +981,6 @@ func (s *Server) onScopesRequest(request *dap.ScopesRequest) {
 	}
 	locScope := &fullyQualifiedVariable{&proc.Variable{Name: "Locals", Children: slicePtrVarToSliceVar(locals)}, "", true}
 
-	// TODO(polina): Annotate shadowed variables
-
 	scopeArgs := dap.Scope{Name: argScope.Name, VariablesReference: s.variableHandles.create(argScope)}
 	scopeLocals := dap.Scope{Name: locScope.Name, VariablesReference: s.variableHandles.create(locScope)}
 	scopes := []dap.Scope{scopeArgs, scopeLocals}
@@ -1135,8 +1133,17 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 				cfqname = "" // complex children are not struct fields and can't be accessed directly
 			}
 			cvalue, cvarref := s.convertVariable(c, cfqname)
+
+			// Annotate any shadowed variables to "(name)" in order
+			// to distinguish from non-shadowed variables.
+			name := c.Name
+			if c.Flags&proc.VariableShadowed == proc.VariableShadowed {
+				// If the variable is shadowed, update the name with parens
+				name = fmt.Sprintf("(%s)", name)
+			}
+
 			children[i] = dap.Variable{
-				Name:               c.Name,
+				Name:               name,
 				EvaluateName:       cfqname,
 				Value:              cvalue,
 				VariablesReference: cvarref,
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1138,7 +1138,6 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 			// to distinguish from non-shadowed variables.
 			name := c.Name
 			if c.Flags&proc.VariableShadowed == proc.VariableShadowed {
-				// If the variable is shadowed, update the name with parens
 				name = fmt.Sprintf("(%s)", name)
 			}
 
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -1562,6 +1562,43 @@ func TestGlobalScopeAndVariables(t *testing.T) {
 	})
 }
 
+// TestShadowedVariables executes to a breakpoint and checks the shadowed
+// variable is named correctly.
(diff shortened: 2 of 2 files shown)


service/dap/server.go
```diff
@@ -1135,8 +1133,16 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
...
+			// to distinguish from non-shadowed variables.
+			name := c.Name
+			if c.Flags&proc.VariableShadowed == proc.VariableShadowed {
+				name = fmt.Sprintf("(%s)", name)
+			}
+
 			children[i] = dap.Variable{
-				Name:               c.Name,
+				Name:               name,
 				EvaluateName:       cfqname,
```
hyangah, line 1146: doesn't EvaluateName need change too?

@aarzilli @derekparker we noticed that accessing shadowed variables (printing/setting) from dlv cli doesn't work. It looks like we can see them only by `locals`
```
(dlv) locals
(animal) = "dog"
animal = "cat"
(dlv) p (animal)
"cat"
```

Is it intended? Should we return an error if users try to evaluate scoped variables?

suzmue: I intentionally didn't change evaluate name in this PR. Adding parens does not actually affect the expression that is evaluated so it doesn't change it, so I thought keeping evaluateName the same will be clearer that the expression that will be evaluated is just "animal". If "(animal)" goes back to being not shadowed, then its name becomes "animal" again anyway.

I considered removing evaluateName, which in the vscode UI removes the 'Add to Watch' option, but a user may want to watch the variable when it is in scope later, so thought it best to keep the evaluateName without modification.

aarzilli: > we noticed that accessing shadowed variables (printing/setting) from dlv cli doesn't work

It's intentional, the argument of print is an expression that's evaluated similarly to the way Go does, so parenthesis do nothing. We could invent some way to access shadowed variables inside expressions but nobody ever asked for it before now.

aarzilli (approved):
LGTM

service/dap/server.go
polinasok: I don't think it's urgent and a TODO for now would work just fine, but I can't we use the magic scopeless syntax for shadowed variables and use that as evaluateName? Like here:
https://github.com/go-delve/delve/blob/8b206092276a50aaa7edd682fa22d95ca9ba8070/service/dap/server.go#L1220-L1222
or here
https://github.com/go-delve/delve/blob/8b206092276a50aaa7edd682fa22d95ca9ba8070/service/dap/server.go#L1056

service/dap/server_test.go
```diff
@@ -1562,6 +1562,43 @@ func TestGlobalScopeAndVariables(t *testing.T) {
...
+					client.ScopesRequest(1000)
+					scopes := client.ExpectScopesResponse(t)
+					expectScope(t, scopes, 0, "Arguments", 1000)
+					expectScope(t, scopes, 1, "Locals", 1001)
+					expectScope(t, scopes, 2, "Globals (package main)", 1002)
+
+					client.VariablesRequest(1001)
+					locals := client.ExpectVariablesResponse(t)
+
+					expectVarExact(t, locals, 0, "(a)", "a", "0", !hasChildren)
```
polinasok, line 1594: Let's add the result of getting their evaluateName and using that in an expression to the test.
For good name, you should be able to use validateEvaluateName helper.
For bad name, you could get an error, then switch scope (as if a user did that in the UI) and then try again.

service/dap/server.go
hyangah: One problem of allowing users to add the address-based expression to the watch option is the watch expressions are preserved across multiple debug sessions and become stale. I think not allowing users to access the shadowed variables but evaluate only within the current scope is intuitive.

aarzilli: It isn't guaranteed to work, the variable could have a fake address due to having been partially registerized. This will become more apparent after the regabi change.

service/dap/server_test.go
suzmue: By switching scope, do you mean running the program until after if statement so only one of the a's is in scope?

polinasok: It doesn't look like a separate scope object is created for the code the curlies, so running until after the curly would be the only option. Not really necessary. Without any extra special magic for evaluate expressions, you could just add two evaluate requests for "a" and "(a)" to show that they yield the same results. Not critical, but might be useful since we had the whole offline discussion if having () around the name makes a difference in expressions.

service/dap/server.go
polinasok: That makes sense. That's why Add to Watch expressions are meaningful and user-friendly fully qualified names and not address-based expressions. Staying consistent with this would mean not being able to evaluate shadowed variables. Agree that it's intuitive.

What does "partially registered" mean? There is an exception to the above for complex map keys and corresponding value because keys don't have their own names. Would this be an issue for those?

aarzilli: > What does "partially registered" mean?

Partially stored in a CPU register, instead of being in memory.

> There is an exception to the above for complex map keys and corresponding value because keys don't have their own names. Would this be an issue for those?

No.

derekparker (approved):
LGTM

PR #2429, opened 2021-04-10
proc: disable part of TestAttachDetach test on macOS

aarzilli:
There seems to be a problem where debugserver will leave a zombie
process instead of detaching correctly, we are sending the right
commands, it doesn't seem to be a problem with Delve.


derekparker (approved):
LGTM

PR #2433, opened 2021-04-12
service/dap: rename launch configuration 'wd' to 'cwd' 

suzmue:
Rename the launch configuration for specifying the working directory.


derekparker:
What's the purpose of this change? Does it adhere better to the DAP spec?

TeamCity build failure is unrelated.

aarzilli:
I think it's because the javascript adapter calls it cwd.

hyangah:
This is not part of DAP spec, but for historical reasons, a commonly used attribute among many debug adapters (like `program`, `env`, `port`, ...) https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes

> `cwd` - current working directory for finding dependencies and other files

hyangah (approved):
LGTM

aarzilli (approved):
LGTM

polinasok:
Just wanted to add some background. In the PR that added this flag, I was the one advocating for "wd" as the attribute name to match the name of delve's command-line flag. And in general I was under the impression that it was a good time to revisit and improve all attribute names since new adapter meant new configuration and a clean slate. At the same time, there was also value in having the user reuse existing configurations to debug with old vs new adapter by just changing a single setting. And at first we were just going to translate from legacy vscode name to new and improved dlv-dap name in the extension, but then realized that in that case the error messages from dlv-dap would not make sense to the users because they would be for flags they never set.  Thank you @suzmue for making the change.

PR #2435, opened 2021-04-13
service/dap: validate the client configurations in initialize request

suzmue:
The client can specify certain configurations in the initialize request.
For example, pathFormat determines the pathFormat. We do not currently
support configuring pathFormat, linesStartAt1, or columnsStartAt1, so
we report an error if the client attempts to set these to an
unsupported value.


suzmue:
cc @hyangah @polinasok

service/dap/server.go
```diff
@@ -433,6 +433,22 @@ func (s *Server) logToConsole(msg string) {
 }
 
 func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
+	if request.Arguments.PathFormat != "path" {
+		s.sendErrorResponse(request.Request, FailedToInitialize, "Failed to initialize",
+			fmt.Sprintf("Unsupported 'pathFormat' value '%v'.", request.Arguments.PathFormat))
```
hyangah, line 438: isn't %q not desirable?

service/dap/server_test.go
```diff
@@ -3060,6 +3067,71 @@ func TestBadAttachRequest(t *testing.T) {
...
+		}
+	}
+
+	runInitializeTest := func(args dap.InitializeRequestArguments, err string) {
+		// Only one initialize request is allowed, so use a new server
+		// for each test.
+		client, teardown := startDapServer(t)
+		defer teardown()
+		client.InitializeRequestWithArgs(args)
+		expectFailedToInitializeWithMessage(client.ExpectErrorResponse(t), err)
```
hyangah, line 3098: can we inline expectFailedToInitializeWithMessage & expectFailedToInitializeRequest?

service/dap/server.go
suzmue: I updated to '%s' since request.Arguments.PathFormat is a string

service/dap/server_test.go
```diff
@@ -74,13 +82,12 @@ func runTest(t *testing.T, name string, test func(c *daptest.Client, f protest.F
 	}()
 
 	client := daptest.NewClient(listener.Addr().String())
-	defer client.Close()
 
-	defer func() {
+	teardown := func() {
```
polinasok, line 86: I am actually removing code this in another PR: https://github.com/go-delve/delve/pull/2414/files#diff-7288f0435a51a46406796e0e110b3f8f1ea66e8c3e6b4209270c8f7f2ff89b9a
I don't think it is necessary (see comment). And even though there a stopOnce, I think it might have gotten flagged at one point for me with "go test -race".

polinasok (commented):
I have conflicting thoughts about this. What prompted this change? Unfortunately, the DAP spec doesn't make it clear if we should be responding with an error or just ignoring the arguments we do not support: https://microsoft.github.io/debug-adapter-protocol/specification. The editor itself, for example, just ignores quietly some of the things that the adapter supports, but the editor does not. I do see a mention here https://microsoft.github.io/debug-adapter-protocol/overview that "A debug adapter is expected to return error messages that honor this locale."

suzmue:
> What prompted this change?

This change was prompted by one of our tests in the vscode-go extension, which tests that we expect an error response when provided with an invalid 'pathFormat'. The vscode-debugadapter checks that pathFormat = 'path' and [returns an errror](https://github.com/microsoft/vscode-debugadapter-node/blob/cceda74da8bb8ed832246d15192e76a26cf12acc/adapter/src/debugSession.ts#L474) if it is something else.

I think this is a bit different from just ignoring something that we do not support, because if someone sets "pathFormat" to "url", they are saying that they plan to send urls instead of filepaths. Or if they set LinesStartAt1=false, then every line number they send to us and we send to them will be off by 1. This would result in a very broken experience for the user, that may not be obvious at first (for example when it comes to line numbers).

service/dap/server_test.go
polinasok: This PR got merged.

polinasok (approved):

service/dap/server_test.go
suzmue: Great! Merged it into this branch.

suzmue:
Just rebased this change.

@aarzilli @derekparker I don't think there is anything left to do on this could you please review? Thanks!

aarzilli (approved):
LGTM

PR #2437, opened 2021-04-17
_scripts: support macOS Big Sur

zchee:
Support macOS Big Sur for `make install`.

macOS has been increase major kernel version. Now Big Sur is `11.0`.
handle it and check whether the major version greater than `10`.

### DIFF
--- a/_scripts/make.go
+++ b/_scripts/make.go
@@ -221,13 +221,18 @@ func canMacnative() bool {
 	}
 
 	macOSVersion := strings.Split(strings.TrimSpace(getoutput("/usr/bin/sw_vers", "-productVersion")), ".")
+
+	major, err := strconv.ParseInt(macOSVersion[0], 10, 64)
+	if err != nil {
+		return false
+	}
 	minor, err := strconv.ParseInt(macOSVersion[1], 10, 64)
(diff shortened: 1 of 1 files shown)


zchee:
Should I handle also arm M1 (from TeamCity log)?

aarzilli:
Yeah, just disabling it on arm64 is ok.

zchee:
@aarzilli Fixed. but might be below is more readable?

```go
	if runtime.GOOS != "darwin" {
		return false
	}
	if runtime.GOARCH != "amd64" {
		return false
	}
```

Also, failed on windows_amd64, not sure why failed.
https://delve.beta.teamcity.com/buildConfiguration/Delve_windows_amd64_1_16/6922

aarzilli:
> @aarzilli Fixed. but might be below is more readable?

I think it's fine as is.

> Also, failed on windows_amd64, not sure why failed.
> https://delve.beta.teamcity.com/buildConfiguration/Delve_windows_amd64_1_16/6922

Not your fault.

zchee:
@aarzilli Thanks. PTAL.

aarzilli (approved):
LGTM

PR #2440, opened 2021-04-18
dap: handle SetVariable requests

hyangah:
The handler invokes debugger.SetVariableInScope, except for
string type variables. For which, we rely on the `call` command.
Moved the call expression handling logic to the new `doCall`
function, so it can be reused by the SetVariable request handler.

Earlier attempt of this PR tried to trigger a Stopped Event whenever
a variable is changed in order to force the editors to refetch data and
refresh UI. But that seems too expensive and some may think annoying
so we decided to not do that. Editors like VSCode tries to make a local
update for the variable after a successful set variable request. However,
if there are aliases, some info may look inconsistent.

```
  i1 := 1
  p1 :=&i1
  pp1 := &p1
``` 
For example, after updating *p1 for the code, users would see

Screen Shot 2021-05-07 at 1 47 52 PM

This inconsistency will be corrected as soon as users continue debugging
and the scope variable is reloaded behind the scene. If this becomes a problem
we need to find a more clever way.

Future improvement planned:
  - This PR returns the requested arg as the response. If the value string
  is a function call, another variable, etc, the value isn't necessarily correct.
  We need to reevaluate and return the value.
  - Maybe accepting arbitrary expression as the arg can be dangerous and
  we can consider to limit acceptable values. Also, currently call is used only
  for string type. If we decide to allow a function call expression for string,
  we can consider to open it to other types too. 

Also fixed a small bug in the call expression evaluation -
Previously, dlv dap returned an error "Unable to evaluate
expression: call stopped" if the call expression is for
variable assignment.  (e.g. call animal = "rabbit").

We met a bug in the injected call path used for assignment. Given the arrival
of the new version of injected call support, I don't know how much of effort
we want to make to chase the bug though.

Fixes golang/vscode-go#1173

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -108,6 +108,7 @@ type Server struct {
 	// mu synchronizes access to objects set on start-up (from run goroutine)
 	// and stopped on teardown (from main goroutine)
 	mu sync.Mutex
+
 	// conn is the accepted client connection.
 	conn net.Conn
 	// debugger is the underlying debugger service.
@@ -1368,7 +1369,7 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 		return
 	}
 
-	children, err := s.childrenOfVariables(v)
+	children, err := s.childrenToDAPVariables(v)
 	if err != nil {
 		s.sendErrorResponse(request.Request, UnableToLookupVariable, "Unable to lookup variable", err.Error())
 		return
@@ -1380,9 +1381,12 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 	s.send(response)
 }
 
-// childrenOfVariables returns the children of the referenced variable.
-func (s *Server) childrenOfVariables(v *fullyQualifiedVariable) ([]dap.Variable, error) {
-	children := make([]dap.Variable, 0)
+// childrenToDAPVariables returns the DAP presentation of the referenced variable's children.
+func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variable, error) {
+	// TODO(polina): consider convertVariableToString instead of convertVariable
+	// and avoid unnecessary creation of variable handles when this is called to
+	// compute evaluate names when this is called from onSetVariableRequest.
+	var children []dap.Variable
 
 	switch v.Kind {
 	case reflect.Map:
@@ -1684,13 +1688,16 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 	isCall, err := regexp.MatchString(`^\s*call\s+\S+`, request.Arguments.Expression)
 	if err == nil && isCall { // call {expression}
 		expr := strings.Replace(request.Arguments.Expression, "call ", "", 1)
-		retVars, err := s.doCall(goid, frame, expr)
+		_, retVars, err := s.doCall(goid, frame, expr)
+		// TODO(polina): We don't need to single out errTerminated case, and sending error
+		// response after 'terminated' event is fine. But TestEvaluateCallRequest is sensitive
+		// and fails without this.
 		if errors.Is(err, errTerminated) {
-			e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
-			s.send(e)
 			return
 		}
 		if err != nil {
+			// TODO(polina): once this is asynchronous, we could wait to reply until the user
+			// continues, call ends, original stop point is hit and return values are available.
 			s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
 			return
 		}
@@ -1711,10 +1718,6 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 		}
 	} else { // {expression}
 		exprVar, err := s.debugger.EvalVariableInScope(goid, frame, 0, request.Arguments.Expression, DefaultLoadConfig)
-		if errors.Is(err, errTerminated) {
-			e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
-			s.send(e)
-		}
 		if err != nil {
 			s.sendErrorResponseWithOpts(request.Request, UnableToEvaluateExpression, "Unable to evaluate expression", err.Error(), showErrorToUser)
 			return
@@ -1727,7 +1730,7 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 	s.send(response)
 }
 
-func (s *Server) doCall(goid, frame int, expr string) ([]*proc.Variable, error) {
+func (s *Server) doCall(goid, frame int, expr string) (*api.DebuggerState, []*proc.Variable, error) {
 	// This call might be evaluated in the context of the frame that is not topmost
 	// if the editor is set to view the variables for one of the parent frames.
 	// If the call expression refers to any of these variables, unlike regular
@@ -1735,12 +1738,15 @@ func (s *Server) doCall(goid, frame int, expr string) ([]*proc.Variable, error) 
 	// and the user will get an unexpected result or an unexpected symbol error.
 	// We prevent this but disallowing any frames other than topmost.
 	if frame > 0 {
-		return nil, fmt.Errorf("call is only supported with topmost stack frame")
+		return nil, nil, fmt.Errorf("call is only supported with topmost stack frame")
 	}
 	stateBeforeCall, err := s.debugger.State( /*nowait*/ true)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+	// TODO(polina): since call will resume execution of all goroutines,
+	// we should do this asynchronously and send a continued event to the
+	// editor, followed by a stop event when the call completes.
 	state, err := s.debugger.Command(&api.DebuggerCommand{
 		Name:                 api.Call,
 		ReturnInfoLoadConfig: api.LoadConfigFromProc(&DefaultLoadConfig),
@@ -1749,21 +1755,12 @@ func (s *Server) doCall(goid, frame int, expr string) ([]*proc.Variable, error) 
 		GoroutineID:          goid,
 	}, nil)
 	if _, isexited := err.(proc.ErrProcessExited); isexited || err == nil && state.Exited {
-		return nil, errTerminated
+		e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
+		s.send(e)
+		return nil, nil, errTerminated
 	}
 	if err != nil {
-		return nil, err
-	}
-
-	// Call commands used for variable assignments do not return a value.
-	// In go '=' is not an operator. Check if go/parser complains.
-	// If the above Call command passed but the expression is not a valid
-	// go expression, we just handled an assignment, and there is no
-	// return value for assignment.
-	if _, err := parser.ParseExpr(expr); err != nil {
-		s.resetHandlesForStoppedEvent()
-		s.sendStoppedEvent(state)
-		return nil, nil
+		return nil, nil, err
 	}
 
 	// After the call is done, the goroutine where we injected the call should
@@ -1771,31 +1768,40 @@ func (s *Server) doCall(goid, frame int, expr string) ([]*proc.Variable, error) 
 	// it is not guaranteed to be selected due to the possibility of the
 	// of simultaenous breakpoints. Therefore, we check all threads.
 	var retVars []*proc.Variable
+	found := false
 	for _, t := range state.Threads {
 		if t.GoroutineID == stateBeforeCall.SelectedGoroutine.ID &&
 			t.Line == stateBeforeCall.SelectedGoroutine.CurrentLoc.Line && t.CallReturn {
+			found = true
 			// The call completed. Get the return values.
 			retVars, err = s.debugger.FindThreadReturnValues(t.ID, DefaultLoadConfig)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 			break
 		}
 	}
-	// TODO(hyangah): call resetHandlesForStop&sendStoppedEvent unconditionally.
-	// States like variable objects aren't guaranteed to survive a continue
-	// and may be updated.
+	// Normal function calls expect return values, but call commands
+	// used for variable assignments do not return a value when they succeed.
+	// In go '=' is not an operator. Check if go/parser complains.
+	// If the above Call command passed but the expression is not a valid
+	// go expression, we just handled a variable assignment request.
+	isAssignment := false
+	if _, err := parser.ParseExpr(expr); err != nil {
+		isAssignment = true
+	}
 
-	if retVars == nil {
+	if !found || !isAssignment && retVars == nil {
 		// The call got interrupted by a stop (e.g. breakpoint in injected
 		// function call or in another goroutine).
 		s.resetHandlesForStoppedEvent()
 		s.sendStoppedEvent(state)
+
 		// TODO(polina): once this is asynchronous, we could wait to reply until the user
 		// continues, call ends, original stop point is hit and return values are available.
-		return nil, errors.New("call stopped")
+		return nil, nil, errors.New("call stopped")
 	}
-	return retVars, nil
+	return state, retVars, nil
 }
 
 func (s *Server) sendStoppedEvent(state *api.DebuggerState) {
@@ -1836,46 +1842,21 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
 	s.sendNotYetImplementedErrorResponse(request.Request)
 }
 
-func variableByName(v *proc.Variable, name string) (*proc.Variable, bool) {
-	if v == nil {
-		return nil, false
-	}
-	if v.Name == name {
-		return v, true
-	}
-
-	for _, child := range v.Children {
-		if child.Name == name {
-			return &child, true
-		}
-		// No need to look into the child's children - DAP clients are expected
-		// to provide the variable reference number, so the name is the name of
-		// the variable's direct child.
-	}
-	return nil, false
-}
-
-// nameToEvaluateName finds the variable of the specified name
-// from the given variable and its children, and returns its
-// evaluateName.
-func (s *Server) nameToEvaluateName(v *fullyQualifiedVariable, name string) (string, error) {
-	if v.Name == name && !v.isScope { // will this ever happen?
-		return v.fullyQualifiedNameOrExpr, nil
-	}
-	children, err := s.childrenOfVariables(v)
+// computeEvaluateName finds the named child, and computes its evaluate name.
+func (s *Server) computeEvaluateName(v *fullyQualifiedVariable, name string) (string, error) {
+	children, err := s.childrenToDAPVariables(v)
 	if err != nil {
 		return "", err
 	}
-	evaluateName := name
 	for _, c := range children {
 		if c.Name == name {
 			if c.EvaluateName != "" {
-				evaluateName = c.EvaluateName
+				return c.EvaluateName, nil
 			}
-			break
+			return "", errors.New("cannot set the variable without evaluate name")
 		}
 	}
-	return evaluateName, nil
+	return "", errors.New("failed to find the named variable")
 }
 
 // onSetVariableRequest handles 'setVariable' requests.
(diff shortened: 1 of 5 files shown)


hyangah:
The test failure is from arm64
```
------- Stdout: -------
=== RUN   TestSetVariable
DAP server listening at: [::]:49859
Connecting to server at: [::]:49859
server_test.go:2913: got &dap.ErrorResponse{Response:dap.Response{ProtocolMessage:dap.ProtocolMessage{Seq:0, Type:"response"}, RequestSeq:11, Success:false, Command:"setVariable", Message:"Unable to set variable"}, Body:dap.ErrorResponseBody{Error:dap.ErrorMessage{Id:2010, Format:"Unable to set variable: backend does not support function calls", Variables:map[string]string(nil), SendTelemetry:false, ShowUser:false, Url:"", UrlLabel:""}}}, want StoppedEvent
--- FAIL: TestSetVariable (0.47s)
```

Should we disable this test on arm64?
@polinasok @suzmue do you think dlv dap shouldn't set 'supportSetVariable' capability on arm64? I guess it's ok to set  'supportSetVariable' even on arm64 even with this limitation.

service/dap/server.go
```diff
@@ -1481,6 +1450,81 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+			if err != nil {
+				return nil, err
+			}
+			break
+		}
+	}
+	if retVars == nil {
+		// The call got interrupted by a stop (e.g. breakpoint in injected
+		// function call or in another goroutine).
+		s.resetHandlesForStop()
```
aarzilli, line 1511: I think this should be done unconditionally, every time continue is called. Variable objects in particular aren't guaranteed to survive a continue, especially now that SetVariable is supported: invalid Variables could lead to memory corruption.

service/dap/server.go
```diff
@@ -1511,10 +1555,105 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+	}
+	if v.Name == name {
+		return v, true
+	}
+
+	for _, child := range v.Children {
+		if child.Name == name {
+			return &child, true
+		}
+		if v, ok := variableByName(&child, name); ok {
```
aarzilli, line 1570: I don't understand why this recursion is needed.

aarzilli:
PS. arm64 never supported call injection.

service/dap/server.go
hyangah: It's from my misunderstanding of how DAP client would access nested fields. Removed.

service/dap/server.go
hyangah: Good point. Left a TODO. For other continues (doCommand is already doing the job I believe). For this call command, this is a pre-existing bug and I hope to address it separately.

suzmue:
> Should we disable this test on arm64?
@polinasok @suzmue do you think dlv dap shouldn't set 'supportSetVariable' capability on arm64? I guess it's ok to set 'supportSetVariable' even on arm64 even with this limitation.

If we don't support set variable on arm64 at all, I think it would probably be best if we didn't set 'supportSetVariable'. We can address that in a separate change tho.

hyangah:
@suzmue set variable works for scalar types except string on arm64. Similarly, on amd64, set variable for compound types or that requires memory allocation won't work anyway.

suzmue:
> @suzmue set variable works for scalar types except string on arm64. Similarly, on amd64, set variable for compound types or that requires memory allocation won't work anyway.

Ok then I think we should definitely still set 'supportsSetVariable'. It's probably worth checking the user experience when set variable fails.

aarzilli (approved):
LGTM

service/dap/server.go
```diff
@@ -1511,10 +1559,105 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+		// the variable's direct child.
+	}
+	return nil, false
+}
+
+// nameToEvaluateName finds the variable of the specified name
+// from the given variable and its children, and returns its
+// evaluateName.
+func (s *Server) nameToEvaluateName(v *fullyQualifiedVariable, name string) (string, error) {
+	if v.Name == name && !v.isScope { // will this ever happen?
```
polinasok, line 1585: I think per dap spec, you are not supposed to consider the variable (aka "container") itself, just its children. "Name - The name of the variable in the container". I think this is because the leaf variables won't have a reference, so the only way to identify them would be to express them as variables in a parent container with some reference.

service/dap/server.go
```diff
@@ -1511,10 +1559,105 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+		return
+	}
+	named, ok := variableByName(v.Variable, arg.Name)
+	if !ok {
+		s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", request.Arguments.VariablesReference))
+		return
+	}
+	// Assume the top-most frame & the current goroutine.
+	// It is not clear if DAP allows SetVariable requests
+	// for variables or memory location not visible from the scode.
```
polinasok, line 1630: what's "scode"?

I am not sure what DAP allows (since the protocol is not very detailed), but vscode would totally let you set a variable from whatever scope you are viewing in the UI, topmost or not. If for now we are assuming this relative to topmost scope only, we should debug log that we are making this assumption. And lets add a TODO to keep track of the frame for the variable, so we can send more explicit error messages for unsupported frames (like with call) or even allow alternative frames (since I think delve backed allows it).

service/dap/server.go
```diff
@@ -1481,6 +1450,85 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Call commands used for variable assignments do not return a value.
+	// In go '=' is not an operator. Check if go/parser complains.
+	// If the above Call command passed but the expression is not a valid
+	// go expression, we just handled an assignment, and there is no
+	// return value for assignment.
+	if _, err := parser.ParseExpr(expr); err != nil {
```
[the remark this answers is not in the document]
aarzilli: > But none of this is covered in the documentation, isn't it? It really should be.

The entire call injection thing is still marked experimental.

service/dap/server.go
```diff
@@ -482,8 +484,7 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 	response.Body.SupportsConditionalBreakpoints = true
 	response.Body.SupportsDelayedStackTraceLoading = true
 	response.Body.SupportTerminateDebuggee = true
-	// TODO(polina): support this to match vscode-go functionality
-	response.Body.SupportsSetVariable = false
+	response.Body.SupportsSetVariable = true
```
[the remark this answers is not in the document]
hyangah: There is a conflicting PR (https://github.com/go-delve/delve/pull/2441) 
Do you mind if we add the additional checks as a separate PR?

service/dap/server.go
```diff
@@ -1141,11 +1142,27 @@ func slicePtrVarToSliceVar(vars []*proc.Variable) []proc.Variable {
...
-	v, ok := s.variableHandles.get(request.Arguments.VariablesReference)
+	ref := request.Arguments.VariablesReference
+	v, ok := s.variableHandles.get(ref)
 	if !ok {
-		s.sendErrorResponse(request.Request, UnableToLookupVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", request.Arguments.VariablesReference))
+		s.sendErrorResponse(request.Request, UnableToLookupVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", ref))
 		return
 	}
+
+	children, err := s.childrenOfVariables(v)
```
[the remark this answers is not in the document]
hyangah: Done

service/dap/server.go
```diff
@@ -1469,6 +1434,10 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 		}
 	} else { // {expression}
 		exprVar, err := s.debugger.EvalVariableInScope(goid, frame, 0, request.Arguments.Expression, DefaultLoadConfig)
+		if errors.Is(err, errTerminated) {
```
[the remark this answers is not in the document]
hyangah: Added by mistake while resolving conflicts

service/dap/server.go
```diff
@@ -1481,6 +1450,85 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+	}
+	state, err := s.debugger.Command(&api.DebuggerCommand{
+		Name:                 api.Call,
+		ReturnInfoLoadConfig: api.LoadConfigFromProc(&DefaultLoadConfig),
+		Expr:                 expr,
+		UnsafeCall:           false,
+		GoroutineID:          goid,
+	}, nil)
+	if _, isexited := err.(proc.ErrProcessExited); isexited || err == nil && state.Exited {
+		return nil, errTerminated
```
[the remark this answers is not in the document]
hyangah: done

service/dap/server.go
```diff
@@ -1481,6 +1450,85 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	// Call commands used for variable assignments do not return a value.
```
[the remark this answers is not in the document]
hyangah: I don't see an obvious, meaningful way to do so yet.  You may have a better idea. Left TODO for you :-)

service/dap/server.go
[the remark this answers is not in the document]
hyangah: It's experimental and currently rarely used. Let's see how people use and what additional protection is possible/impossible from the runtime.

In another thread, we were talking about unconditionally resetting variable handles. I think it's the safest to reset variable handles given that we don't know what's cached or what's preloaded through alias, etc. Actually, in my first implementation I didn't reset the handle and I decided to refresh everything while writing tests. 

At least, it seems like we all agree that it's a limitation in DAP/VSCode and is not ideal. This is a rarely used feature, so let's see how people use this. Also note that call support is experimental in both delve and also go runtime.

service/dap/server.go
hyangah: done

service/dap/server_test.go
```diff
@@ -2123,7 +2123,17 @@ func TestEvaluateCallRequest(t *testing.T) {
 					if erres.Body.Error.Format != "Unable to evaluate expression: not enough arguments" {
 						t.Errorf("\ngot %#v\nwant Format=\"Unable to evaluate expression: not enough arguments\"", erres)
 					}
-					// Call can exit
+					// Assignment - expect a stopped event, but without an error.
+					client.EvaluateRequest("call one = two", 1000, "this context will be ignored")
+					stopped := client.ExpectStoppedEvent(t)
+					got = client.ExpectEvaluateResponse(t)
+					expectEval(t, got, "", noChildren)
+					handleStop(t, client, stopped.Body.ThreadId, "main.main", -1)
```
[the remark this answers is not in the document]
hyangah: done

service/dap/server_test.go
```diff
@@ -2866,6 +2876,220 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+				})
+			},
+			// Stop at the breakpoints set within the program, and after returning from barfoo (67)
+			fixture.Source, []int{67},
+			[]onBreakpoint{{
+				execute: func() {
+					startLineno := 66
+					if runtime.GOOS == "windows" && goversion.VersionAfterOrEqual(runtime.Version(), 1, 15) {
+						// Go1.15 on windows inserts a NOP after the call to
+						// runtime.Breakpoint and marks it same line as the
```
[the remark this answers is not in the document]
hyangah: removed 67 - no longer necessary.

service/dap/server.go
```diff
@@ -1511,10 +1559,105 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+
+	v, ok := s.variableHandles.get(arg.VariablesReference)
+	if !ok {
+		s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", request.Arguments.VariablesReference))
+		return
+	}
+	// We may need to translate the name to the evaluateName if the name
+	// refers to a field or element of a variable.
+	// https://github.com/microsoft/vscode/issues/120774
+	// Compute the evaluateName by running the same code the onVariablesRequest
```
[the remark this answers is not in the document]
hyangah: right. updated the comment

service/dap/server.go
hyangah: That's what I am thinking too. Do you want me to remove this?

hyangah: removed.

service/dap/server.go
```diff
@@ -1511,10 +1559,105 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+// onSetVariableRequest handles 'setVariable' requests.
+func (s *Server) onSetVariableRequest(request *dap.SetVariableRequest) {
+	arg := request.Arguments
+
+	v, ok := s.variableHandles.get(arg.VariablesReference)
+	if !ok {
+		s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", request.Arguments.VariablesReference))
+		return
+	}
+	// We may need to translate the name to the evaluateName if the name
```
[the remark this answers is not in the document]
hyangah: That shows how rarely this feature is used.

(I think I already sent this response. Not sure why github UI still shows it. :-( )

hyangah:
Again the DeepSource failure is not relevant to this. And I am not sure if the specific log security check is useful for the daptest package. Is there a way to disable the check for the package?

polinasok (commented):
Was working on another round of comments after thinking more deeply about the design for this. Looks like I did not make it before the new version. I will take a look now.

service/dap/server.go
```diff
@@ -1559,10 +1604,103 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+		// For the string case that was handled by s.doCall, s.doCall
+		// triggers StoppedEvent.
+		state, err := s.debugger.State(true)
+		if err != nil {
+			s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to reset state after setting variable", err.Error())
+		}
+		s.resetHandlesForStoppedEvent()
+		s.sendStoppedEvent(state)
+	}
+	// update the variable stored in variableHandles
```
polinasok, line 1700: This comment doesn't match the code

service/dap/server.go
```diff
@@ -1559,10 +1604,103 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+		if _, err := s.doCall(goid, frame, fmt.Sprintf("%v=%v", evaluateName, arg.Value)); err != nil {
+			s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to set variable", err.Error())
+			return
+		}
+	default:
+		if err := s.debugger.SetVariableInScope(goid, frame, 0, evaluateName, arg.Value); err != nil {
+			s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to set variable", err.Error())
+			return
+		}
+		// Send StoppedEvent so the client retrieve threads/scopes again.
```
polinasok, line 1687: I can swear I wrote a comment about this, but I don't see it anymore.

So in a similar fashion as with call assignments, I think resetting all threads and frames and sending a stopped event here might be an overkill. You might end up with a different stop reason because you pick it from the debugger and not based on doCommand code. You will have to wait to load all threads. Etc. Unfortunately here vscode doesn't help us and doesn't send a scopes request to refresh the variables. And ideally we do want to refresh all variables in the selected scope, both in the UI and in the handles map. And the reason why I say variables and not just this one variable is because ideally we want three things updated:
(1) the value of the thing you just set
(2) the value of any children that got reset because you reset the parent
(3) the inlined value within the value of the parent variable
I tried this with javascript. They update (1) and (2), but not (3). The best way to update all three would be to file an issue against vscode and tell them they need to send a scopes request like they do after evaluate requests. 

Settling for (1) and (2) is good enough for the time being. And I think the way to do that is to return new variables reference in the setVariable response. That's how you update the cached value. You just let the old one be (we can optimize this eventually and clear, but for now we just stop using its key), and you give the editor a new reference for the refreshed variable.

You should be able to test this with two pointers p1 and p2 of he same type and then set variable p1 to p2. What should we see? The expanded dereferenced value should not disappear and should change to whatever *p2 is. Does that work?

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					args := client.ExpectVariablesResponse(t)
+
+					// Test: try to set composite type 'bar'. That is not implemented.
+					expectVarExact(t, args, 1, "bar", "bar", `main.FooBar {Baz: 10, Bur: "lorem"}`, hasChildren)
+					client.SetVariableRequest(1000, "bar", `main.FooBar {Baz: 42, Bur: "ipsum"}`)
+					if got, want := client.ExpectErrorResponse(t), "*ast.CompositeLit not implemented"; !strings.Contains(got.Body.Error.Format, want) {
+						t.Errorf("got %#v, want error string containing %q", got, want)
+					}
+
+					// Test: try to set integer type of composite type var 'bar'. That is supported.
```
polinasok, line 2953: I think you mean try to set a nested field of a composite var.

polinasok, line 2953: You should also be testing other composite types:
- pointers that have children with no names - do we handle that?
- arrays that have children with index names 
- maps that have different child structure depending on whether keys and values are scalars (3 subcases actually) and I think set variable might not make sense in some of them
- interfaces: can a user set .(data)?

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+	protest.MustSupportFunctionCalls(t, testBackend)
+	runTest(t, "testvariables", func(client *daptest.Client, fixture protest.Fixture) {
+		runDebugSessionWithBPs(t, client, "launch",
+			func() {
+				client.LaunchRequestWithArgs(map[string]interface{}{
+					"mode": "exec", "program": fixture.Path, "showGlobalVariables": true,
+				})
+			},
+			// Stop at the breakpoints set within the program, and after returning from barfoo (67)
+			fixture.Source, []int{67},
```
polinasok, line 3021: I thought you removed this? Maybe that was the other test. Also it might be easier to just set 66  explicitly, then to rely on runtime breakpoint and then special case the line number. We did that in some test to test that we stop at correct lines, but you shouldn't need this here.

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					if got, want := client.ExpectSetVariableResponse(t), `"ipsum"`; got.Success != true || got.Body.Value != want {
+						t.Errorf("got %#v, want {Success=true, Body.Value=%q", got, want)
+					}
+					handleStop(t, client, 1, "main.foobar", startLineno)
+
+					client.VariablesRequest(1000)
+					args3 := client.ExpectVariablesResponse(t)
+					expectVarExact(t, args3, 1, "bar", "bar", `main.FooBar {Baz: 10, Bur: "ipsum"}`, hasChildren)
+
+					// Test: set integer 'a2'
```
polinasok, line 3072: That doesn't use call and you already cover this above.

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					locals2 := client.ExpectVariablesResponse(t)
+					expectVarExact(t, locals2, -1, "a2", "a2", "42", noChildren)
+
+					// Test: set integer 'a2' with an invalid value.
+					client.SetVariableRequest(1001, "a2", "false")
+					if got, want := client.ExpectErrorResponse(t), "can not convert false constant to int"; !strings.Contains(got.Body.Error.Format, want) {
+						t.Errorf("got %#v, want error string containing %q", got, want)
+					}
+
+					// TODO(hyangah): test with an integer field of a global variable (struct type).
```
polinasok, line 3095: That would be a regular set as well, not a call, won't it?

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+				},
+			}, {
+				// Stop at second breakpoint.
+				execute: func() {
+					handleStop(t, client, 1, "main.barfoo", -1)
+					// Test: set string 'a1' in main.barfoo.
+					// This shouldn't affect 'a1' in main.foobar -
```
polinasok, line 3102: You should load the scope for the other one and set again. That will impact the one in the topmost, won't it?

service/dap/server_test.go
```diff
@@ -2905,6 +2915,219 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					// we will check that in the next breakpoint.
```
polinasok, line 3103: You don't need to go to next breakpoint to check this. You can just load parent scope with a scopes request. That's what happens when users click on it in the UI to view all the vars. Then once you do that, you can try to set it again. That should impact the topmost scope because that's your default. If you switch back to it, you should see it.

service/dap/server_test.go
polinasok, line 2953: You need a test for field and parent having the same name. You used to have a bug for that case before in the code. Let's make sure it got fixed.

polinasok:
> That shows how rarely this feature is used.

Or it could be an indication that people don't even bother filing issues and just decide quietly to not not rely on the broken feature. Hopefully not.

service/dap/server.go
```diff
@@ -1519,6 +1514,8 @@ func (s *Server) doCall(goid, frame int, expr string) ([]*proc.Variable, error)
 		GoroutineID:          goid,
 	}, nil)
 	if _, isexited := err.(proc.ErrProcessExited); isexited || err == nil && state.Exited {
```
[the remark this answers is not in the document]
aarzilli: A cursory look at pkg/proc tells me that there are ways were we could return StopUnknown + ErrProcessExited. I think this is a bug, if the process exited there can not be any other reason for stopping. Probably adding a check inside (*Target).Continue after the call to ContinueOnce would fix it.

aarzilli:
> Again the DeepSource failure is not relevant to this.

It is not and it shouldn't be reported anyway because it isn't a new issue.

> And I am not sure if the specific log security check is useful for the daptest package. Is there a way to disable the check for the package?

No, just ignore it.

derekparker:
@polinasok you're still waiting on your review comments to be addressed, correct?

polinasok:
> @polinasok you're still waiting on your review comments to be addressed, correct?

I believe so.

derekparker:
@hyangah please fix merge conflicts.

@polinasok I haven't seen any updated addressing your comments so I'll hold off on final review and merge for now, unless I'm mistaken.

polinasok:
> @hyangah please fix merge conflicts.
> 
> @polinasok I haven't seen any updated addressing your comments so I'll hold off on final review and merge for now, unless I'm mistaken.

I would prefer if this did not get merged with the issues I pointed out without a further discussion. Thank you.
I believe @hyangah has been graciously waiting for the dust to settle on some of the other more urgent PRs to avoid unnecessary conflicts. As those get merged in, I expect there will be movement on this PR as well.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: What about StopCallReturned? I know the selected goroutine might not be the one where the call returned. Would it perhaps make sense to track stop reasons per thread/goroutine?

hyangah (commented):
Sorry that resolving merge conflict was painful and I had to rebase this. Github review UI doesn't allow me to respond to comments on outdated version. Let me know if I missed anything.

For the function name suggestions, I tried to come up with names that avoid 'get' prefixes (the recommendation is to avoid verb-prefixes like 'get' or 'make' when the following nouns are sufficient to carry the meaning).  Let me know if you have better name suggestions.

service/dap/server_test.go
```diff
@@ -3036,6 +3046,378 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					tester.failSetVariableAndStop(1001, "str", `callpanic()`, `callpanic panicked`)
+					handleStop(t, client, 1, "main.main", 197)
+
+					// breakpoint during a function call.
+					tester.failSetVariableAndStop(1001, "str", `callbreak()`, "call stopped")
+
+					// TODO(hyangah): continue after this causes runtime error while resuming
+					// unfinished injected call.
+					//   runtime error: can not convert %!s(<nil>) constant to string
+					// This can be reproducible with dlv cli. (`call str = callbreak(); continue`)
```
hyangah, line 3414: @aarzilli @derekparker

hyangah:
The TeamCity builder failed due to a failure in pkg/proc test (windows). I don't know how this PR is related.

```
=== RUN   TestFrameEvaluation
support.go:252: enabling recording for github.com/go-delve/delve/pkg/proc_test.TestFrameEvaluation
...
proc_test.go:1315: Goroutine 1 not found
--- FAIL: TestFrameEvaluation (0.08s)
```

service/dap/server.go
```diff
@@ -1669,69 +1687,18 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
-		}, nil)
-		if _, isexited := err.(proc.ErrProcessExited); isexited || err == nil && state.Exited {
-			e := &dap.TerminatedEvent{Event: *newEvent("terminated")}
-			s.send(e)
+		expr := strings.Replace(request.Arguments.Expression, "call ", "", 1)
+		_, retVars, err := s.doCall(goid, frame, expr)
+		// TODO(polina): We don't need to single out errTerminated case, and sending error
+		// response after 'terminated' event is fine. But TestEvaluateCallRequest is sensitive
+		// and fails without this.
+		if errors.Is(err, errTerminated) {
```
aarzilli, line 1695: I would prefer `errors.Is` wasn't used. It was introduced in 1.13 which is still pretty recent and every time we bump the minimum required version we get complaints.

service/dap/server_test.go
aarzilli: I don't know if understand, but this is just an instruction numbering problem, on windows the first instruction after returning from the call is a NOP that is numbered with the same line as the function call, on other operating systems it's the first instruction of the next line.

service/dap/server.go
```diff
@@ -1793,10 +1842,119 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
...
+			s.sendErrorResponse(request.Request, UnableToSetVariable, "Unable to set variable", err.Error())
+			return
+		}
+	}
+	// * Note on inconsistent state after set variable:
+	//
+	// The variable handles may be in inconsistent state - for example,
+	// let's assume there are two aliased variables pointing to the same
+	// memory and both are already loaded and cached in the variable handle.
+	// If set variable updates the memory through one of them,
```
aarzilli, line 1938: This is underselling it, the stack could have moved and setting a stale variable could corrupt the stack of a different goroutine.

hyangah: Agree. Updated the comment  (and a note in doCall) to include the discussion. PTAL.

polinasok:
> Sorry that resolving merge conflict was painful and I had to rebase this. Github review UI doesn't allow me to respond to comments on outdated version. Let me know if I missed anything.

I believe this happens because you rebase and force push. If you just push a new commit, then it would be possible to go back to previous commits and see and respond to the old comments. I am not sure how I will related the old comments to the new code now.

service/dap/server.go
```diff
@@ -1793,10 +1855,116 @@ func (s *Server) onReverseContinueRequest(request *dap.ReverseContinueRequest) {
 	s.sendNotYetImplementedErrorResponse(request.Request)
 }
 
-// onSetVariableRequest sends a not-yet-implemented error response.
-// Capability 'supportsSetVariable' is not set 'initialize' response.
-func (s *Server) onSetVariableRequest(request *dap.SetVariableRequest) { // TODO V0
-	s.sendNotYetImplementedErrorResponse(request.Request)
+// computeEvaluateName finds the named child, and computes its evaluate name.
+func (s *Server) computeEvaluateName(v *fullyQualifiedVariable, name string) (string, error) {
```
polinasok, line 1859: `name` seems like variable's, not child's name. How about ` cname`? `childName`? `child`? Or renaming v to `parent` or `container`?

service/dap/server_test.go
```diff
@@ -3036,6 +3046,377 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					tester := &helperForSetVariable{t, client}
+
+					handleStop(t, client, 1, "main.barfoo", -1)
+					// Test: set string 'a1' in main.barfoo.
+					// This shouldn't affect 'a1' in main.foobar - we will check that in the next breakpoint.
+					locals := tester.variables(1001)
+					expectVarExact(t, locals, -1, "a1", "a1", `"bur"`, noChildren)
+					tester.expectSetVariable(1001, "a1", `"fur"`)
+					tester.evaluate("a1", `"fur"`, noChildren)
+					// We will check a1 in main.foobar isn't affected from the next breakpoint.
```
polinasok, line 3368: Instead of advancing to the next breakpoint, you could just load another scope for the calling function. Users can do that in the UI by clicking on different functions in the stack.

service/dap/server_test.go
```diff
@@ -3036,6 +3046,377 @@ func TestUnupportedCommandResponses(t *testing.T) {
...
+					m1 := tester.variables(m1Ref)
+					elem1 := m1.Body.Variables[0]
+					tester.expectSetVariable(elem1.VariablesReference, "A", "-9999")
+					tester.expectSetVariable(elem1.VariablesReference, "B", "10000")
+					tester.evaluate(elem1.EvaluateName, "main.astruct {A: -9999, B: 10000}", hasChildren)
+
+					// map: struct -> int
+					m3Ref := expectVarExact(t, locals, -1, "m3", "m3", "map[main.astruct]int [{A: 1, B: 1}: 42, {A: 2, B: 2}: 43, ]", hasChildren)
+					tester.expectSetVariable(m3Ref, "main.astruct {A: 1, B: 1}", "8888")
+					// note: updating keys is possible, but let's not promise anything.
```
polinasok, line 3280: :)

polinasok (commented):
Thank you for expanding the test coverage and what the code handles.

derekparker:
@polinasok are you +1 on this?

polinasok (approved):
LGTM

aarzilli:
Merge conflicts.

hyangah:
Merged master, and passed tests. Thanks!

PR #2441, opened 2021-04-19
daptest/gen: autogenerate some assertion functions

hyangah:
daptest has type assertion functions that panic if the read response/event message is not
the expected type. This is not only against the recommended style guideline (Don't Panic, Useful
Test Failures, ...), but also it prevents from quickly diagnosing test failures occurred in remote CIs.

This PR changes the type assertion to the two return value type assertion, and t.Fatal with details
if the type is not expected. `service/dap/daptest/main.go` is a program that auto-generates those assertion functions in resp.go.

Run `go generate` from the service/dap directory to update resp.go.


suzmue:
LGTM.

The test failure is from service/dap/daptest/client.go:36 "Audit the logger for user defined data that could break log file pattern", which isn't part of the code that you changed.

aarzilli (approved):
LGTM

service/dap/daptest/gen/main.go
```diff
@@ -0,0 +1,99 @@
...
+
+	_ "github.com/google/go-dap"
+)
+
+var oFlag = flag.String("o", "", "output file name")
+
+var tmpl = template.Must(template.New("assert").Parse(`package daptest
+
+// Code generated by go generate; DO NOT EDIT.
+// The code generator program is in ./gen directory.
```
polinasok, line 24: Do you think perhaps this and the output library should be something shipped with go-dap from that repo?

service/dap/daptest/client.go
```diff
@@ -97,83 +99,9 @@ func (c *Client) ExpectUnsupportedCommandErrorResponse(t *testing.T) *dap.ErrorR
...
-
-func (c *Client) ExpectTerminatedEvent(t *testing.T) *dap.TerminatedEvent {
-	t.Helper()
-	return c.ExpectMessage(t).(*dap.TerminatedEvent)
-}
-
-func (c *Client) ExpectInitializeResponse(t *testing.T) *dap.InitializeResponse {
-	t.Helper()
-	initResp := c.ExpectMessage(t).(*dap.InitializeResponse)
-	if !initResp.Body.SupportsConfigurationDoneRequest {
```
polinasok, line 133: This check got lost in the shuffle. You need a wrapper around the other helper to test for this and other capabilities we expect to be set.

hyangah: Thanks. Nice catch. Added as ExpectInitializeResponseAndCapability (yeah, not a great name)

Curious why this capability check should be done for every initialize response - will the dap server change the set of supported capabilities across testing?

hyangah (commented):
> Do you think perhaps this and the output library should be something shipped with go-dap from that repo?

I don't think the generator code should be in go-dap. I see it specific to this service/dap test case.
Assertion test library is not recommended in general so it's better to avoid sharing further.

(Sorry, I can't respond to this comment directly. Github Review UI is strange.)

derekparker (approved):
LGTM

derekparker:
@polinasok any other thoughts here before merge?

service/dap/daptest/client.go
polinasok: It's fixed now. This might change with rr because it might or might not be supported on different architectures. We could just test this explicitly in some tests, but not others. It seemed easier to just hide this in a helper, even if that means rechecking the same (minor) thing every time.

service/dap/daptest/client.go
```diff
@@ -89,6 +89,15 @@ func (c *Client) expectErrorResponse(t *testing.T, id int, message string) *dap.
 	return er
 }
 
+func (c *Client) ExpectInitializeResponseAndCapability(t *testing.T) *dap.InitializeResponse {
```
polinasok, line 92: ExpectInitializeResponseWithDefaultCapabilities?
And please check also add the other missing capabilities that we set now. The previous helper never got updated as things evloved :(

hyangah: Done. Would be better if it was a purely refactoring cl.

polinasok (approved):

derekparker:
Just needs rebase.

hyangah:
Done.

polinasok (approved):
LGTM

PR #2443, opened 2021-04-20
service/dap: cancel next before sending stopped event

suzmue:
When 'NextInProgress' is set, sending a 'next', 'step', or 'stepOut'
request will result in a 'next while nexting' error. This change
cancels the 'NextInProgress' before sending the stopped event, so
users can continue execution as desired.

The other option for handling 'NextInProgress' would be to instead
skip any breakpoints that are encountered and notifying the user.
I chose to do cancel next instead because it is the simpler to
implement of the two choices and is easy to understand.

Updates golang/vscode-go#119


suzmue:
\cc @polinasok @hyangah

aarzilli:
The problem with both this and skipping breakpoint is that the user will not receive any indication that anything has gone wrong.

suzmue:
> The problem with both this and skipping breakpoint is that the user will not receive any indication that anything has gone wrong.

I can add an OutputEvent to the user that includes the error. Does that sound like an ok solution?

Ideally I would also like the description in the stopped event for the goroutine that next/step/stepout was referencing to also be 'next cancelled' or something else that indicates that the next/step/stepout was interrupted.

aarzilli:
I think that actually the other option is preferable:

> The other option for handling 'NextInProgress' would be to instead
> skip any breakpoints that are encountered and notifying the user.

The implementation is just a loop around the call to Command inside doCommand, where you check state.NextInProgress and send OutputEvents as appropriate.

aarzilli:
I think I should elaborate on my last message: I would prefer that there was a UI to notify the user of this event but there is no way to do it through DAP and vscode has been aware of this for a while now and nothing has been done about it, so the preferable solution is not on the table.

What I am worried about is that without a prominent indication that the current goroutine changed users that are mindlessly clicking the step over function will be confused that control flow will seem to jump around randomly.

Delve's command line client has implemented the autocontinue solution for many years and we haven't received many complaints about it (although still greater than zero). So doing the same thing in dap is somewhat safe and at least consistent with other parts of delve.

suzmue:
That makes a lot of sense. I think staying consistent with the command line client is a good choice too.

I do actually have a branch with that implemented but I will need to wait for @polinasok s async PR to get merged first before sending it to make sure it works. I'll close this PR for now and send the other later.

suzmue:
From #2530, there is still discussion about the right approach for breakpoints while nexting in DAP. Since this is designed to be used in an editor, which often has an easy way to disable all goroutines, it seems reasonable to allow breakpoints to interrupt goroutines.

If a user wants to not be interrupted, they may disable all breakpoints before stepping.

#2530 is being kept open to continue discussion about synchronizing between handleRequest and doRunCommand, which will be necessary for implementing log points.

polinasok:
Based on the offline discussion we had as a group, the decision is to neither cancel next nor skip the breakpoint and auto-continue. Instead we will go with the simplest and least invasive way of handling of this case - stop and don't cancel next -  PLUS increase communication to let the users know how to handle things from their end. 

Stopping at the breakpoint gives users the option to examine things. "Continue" lets them "skip" the breakpoint and finish "next". There is not a way to cancel next at this point, but let's see if we get any signals from the users that they need this option before investing into it.

In the meantime, a big user issue so far has been not knowing how to get out of the "next while nexting" error. Many users already thought that it killed their debug session. New enhanced error handling we have in dlv-dap might only make this worse.

So I propose that we do the following:
(1) Via Debug Console, communicate that next is in progress, another next cannot be performed, but Continue would finish next.
(2) Use a more descriptive stop reason - e.g.  PAUSED ON BREAKPOINT DURING STEP or STEP PAUSED BY BREAKPOINT
(3) Change "next while nexting" error to look less terminal. Instead of RUNTIME ERROR, we could call this "debug error", "command error", "invalid command". We could drop the stacktrace in the exception and print an explanation and a pointer to use Continue.

suzmue:
Closing based on @polinasok s comment above

PR #2444, opened 2021-04-20
service/dap: implement exception info

suzmue:
Implements the ExceptionInfo DAP request. This allows the server to provide additional
details about failures, such as fatalpanics, fatalthrows, and runtime errors.

Updates golang/vscode-go#644


suzmue:
cc @polinasok @hyangah

service/dap/server.go
```diff
@@ -1555,6 +1563,30 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
...
+// Capability 'supportsExceptionInfoRequest' is set 'initialize' response.
+func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
+	goroutineID := request.Arguments.ThreadId
+	body := s.exceptions[goroutineID]
+	frames, err := s.debugger.Stacktrace(goroutineID, s.args.stackTraceDepth, 0)
+	if err == nil && len(frames) > 0 {
+		apiFrames, err := s.debugger.ConvertStacktrace(frames, nil)
+		if err == nil {
+			var buf bytes.Buffer
+			terminal.PrintStack(func(path string) string {
```
aarzilli, line 1576: Pass s.toClientPath directly instead of the closure.

service/dap/server.go
```diff
@@ -1555,6 +1563,30 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
...
+				return s.toClientPath(path)
+			}, &buf, apiFrames, "\t", false)
+
+			body.Details.StackTrace = "Stack:\n" + buf.String()
```
aarzilli, line 1580: Do `fmt.Fprintf(&buf, "Stack:\n")` before the call to PrintStack instead of concatenating here.

service/dap/server.go
```diff
@@ -69,6 +71,9 @@ type Server struct {
 	variableHandles *variablesHandlesMap
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
+	// exceptions maps goroutineIds to info about the last exception that occured.
+	// Reset at every stop.
+	exceptions map[int]dap.ExceptionInfoResponseBody
```
aarzilli, line 76: IMHO it would be better to recover the informations about the exception inside `onExceptionInfoRequest` instead of having to keep another piece of global state correct. You can add a helper method to `debugger.(*Debugger)` to call `proc.FindGoroutine` and then the `Thread` member field will give you the thread of the goroutine with info about the current breakpoint.

suzmue: Yeah I agree I think that would be better. Do you think that would also work with the runtime errors?

I'll experiment a bit and see if I can make that work.

service/dap/server.go

service/dap/server.go

service/dap/server.go
aarzilli: It should, there is nothing special about them. The only thing I'm not sure about is if the current goroutine id is 0, but in that case you can just use CurrentThread.

suzmue: Thanks for the pointers!

I'm planning to address the issue with current goroutine id is 0 in a separate PR.  We are considering sending the current thread in addition to the goroutines in the threadsResponse when no goroutine is selected, so the exception info will correspond to a thread.

suzmue: Ok I was able to successfully move the handling of the two breakpoints to `onExceptionInfoRequest`, but the runtime errors as far as I can tell are not saved anywhere. We could either cache them on the debugger side (by saving them and then putting the result in the DebuggerState), or on the dap server side. Do you have any thoughts on which is better? Or of other ways to implement it that I may be missing?

aarzilli: No, that error isn't saved anywhere. It's weird to associate it with a specific thread (or goroutine), because usually it isn't. But the one exception, bad access, is -- and it's probably the only error that's ever going to happen in that branch. I'd say save it on DAP's side and return it in onExceptionInfoRequest but only for the current thread/goroutine.

service/dap/server.go
```diff
@@ -1567,7 +1566,39 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
 // Capability 'supportsExceptionInfoRequest' is set 'initialize' response.
 func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 	goroutineID := request.Arguments.ThreadId
-	body := s.exceptions[goroutineID]
+	body := dap.ExceptionInfoResponseBody{}
+	// Get the thread for the goroutine and the current state.
+	thread, _ := s.debugger.FindGoroutine(goroutineID)
```
aarzilli, line 1571: Can you call this variable `g`? We're using `thread` to mean "OS thread" everywhere else.

aarzilli: Sorry I didn't notice that this actually *is* a Thread. Please change FindGoroutine to return a goroutine.

service/dap/server.go
```diff
@@ -494,6 +498,8 @@ func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 	response.Body.SupportsReadMemoryRequest = false
 	response.Body.SupportsDisassembleRequest = false
 	response.Body.SupportsCancelRequest = false
+	// These requests are not yet supported by vscode-go
+	response.Body.SupportsExceptionInfoRequest = true
```
polinasok, line 502: nit: you should move this to the top, grouping together with all the other true capabilities. I don't think the comment is necessary.

service/dap/server.go
```diff
@@ -1555,6 +1562,64 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
 	s.sendNotYetImplementedErrorResponse(request.Request)
 }
 
+// onExceptionInfoRequest handles 'exceptionInfo' requests.
+// Capability 'supportsExceptionInfoRequest' is set 'initialize' response.
```
polinasok, line 1566: grammer: set in

service/dap/server.go
```diff
@@ -1555,6 +1562,64 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
...
+func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
+	goroutineID := request.Arguments.ThreadId
+	var body dap.ExceptionInfoResponseBody
+	// Get the goroutine and the current state.
+	g, _ := s.debugger.FindGoroutine(goroutineID)
+	state, _ := s.debugger.State( /*nowait*/ true)
```
polinasok, line 1572: why ignore the errors?

service/dap/server.go
```diff
@@ -1555,6 +1562,64 @@ func (s *Server) onCancelRequest(request *dap.CancelRequest) {
...
+	var bpState *proc.BreakpointState
+	if g != nil && g.Thread != nil {
+		// Check if this goroutine ID is stopped at a breakpoint.
+		bpState = g.Thread.Breakpoint()
+	}
+	if bpState != nil && bpState.Breakpoint != nil {
+		switch bpState.Breakpoint.Name {
+		case proc.FatalThrow:
```
polinasok, line 1580: @aarzilli Why isn't there a special stop reason for these?

service/dap/server.go
```diff
@@ -1684,10 +1757,10 @@ func (s *Server) doCommand(command string) {
 
 		// TODO(polina): according to the spec, the extra 'text' is supposed to show up in the UI (e.g. on hover),
```
polinasok, line 1758: Does text show up in some new way with the new changes? Do we need it since you do the same thing in exception info? This todo likely can go away now, right? You could turn this into just a comment explaining that we had to use exception info because text did not work with multiple goroutines.

service/dap/server.go
```diff
@@ -1684,10 +1757,10 @@ func (s *Server) doCommand(command string) {
...
 		// but so far I am unable to get this to work in vscode - see https://github.com/microsoft/vscode/issues/104475.
-		// Options to explore:
-		//   - supporting ExceptionInfo request
-		//   - virtual variable scope for Exception that shows the message (details here: https://github.com/microsoft/vscode/issues/3101)
 		// In the meantime, provide the extra details by outputing an error message.
+		// TODO(suzmue): the ExceptionInfo request is now supported and displays the error message, however if there
+		// is no threadId set in the stopped event, then the ExceptionInfo response is not visible. We should be able
+		// to delete this output event, once we solve that issue.
```
polinasok, line 1763: I remember that this happens, but I can't remember when. Could you remind me please?

service/dap/server_test.go
```diff
@@ -2417,13 +2417,20 @@ func TestBadAccess(t *testing.T) {
...
-							t.Errorf("\ngot  %#v\nwant ThreadId=1 Reason=\"runtime error\" Text=\"%s\"", se, errorPrefix)
+						if se.Body.ThreadId != 1 || se.Body.Reason != "exception" || se.Body.Description != "Paused on runtime error" || !strings.HasPrefix(se.Body.Text, errorPrefix) {
+							t.Errorf("\ngot  %#v\nwant ThreadId=1 Reason=\"exception\" Description=\"Paused on runtime error\" Text=\"%s\"", se, errorPrefix)
 						}
 						oe := client.ExpectOutputEvent(t)
 						if oe.Body.Category != "stderr" || !strings.HasPrefix(oe.Body.Output, "ERROR: "+errorPrefix) {
 							t.Errorf("\ngot  %#v\nwant Category=\"stderr\" Output=\"%s ...\"", oe, errorPrefix)
 						}
+
+						client.ExceptionInfoRequest(1)
```
polinasok, line 2428: So the client will send this in response to a stopped event marked as exception?

polinasok:
> No, that error isn't saved anywhere. It's weird to associate it with a specific thread (or goroutine), because usually it isn't. But the one exception, bad access, is -- and it's probably the only error that's ever going to happen in that branch. I'd say save it on DAP's side and return it in onExceptionInfoRequest but only for the current thread/goroutine.

@aarzilli That's the error that a call to debugger returns as err. Another kind of runtime error is "next while nexting". Why wouldn't it make sense to return the err from the last command inside of debugger State?  There is also 
```
// Filled by RPCClient.Continue, indicates an error
Err error `json:"-"`
```

polinasok (commented):
Could you please add some before and after screenshots to the PR for the different kinds of errors?

service/dap/server.go
aarzilli: The way I see it, they are normal breakpoints.

aarzilli:
> > No, that error isn't saved anywhere. It's weird to associate it with a specific thread (or goroutine), because usually it isn't. But the one exception, bad access, is -- and it's probably the only error that's ever going to happen in that branch. I'd say save it on DAP's side and return it in onExceptionInfoRequest but only for the current thread/goroutine.
> 
> @aarzilli That's the error that a call to debugger returns as err. Another kind of runtime error is "next while nexting". 

I think the "next while nexting" error should be handled in some other way. 

> Why wouldn't it make sense to return the err from the last command inside of debugger State?

I don't follow. Isn't it there already?

service/dap/server_test.go
suzmue: Yes exactly

service/debugger/debugger.go
```diff
@@ -967,6 +967,19 @@ func (d *Debugger) FindThread(id int) (proc.Thread, error) {
 	return nil, nil
 }
 
+// FindGoroutine returns the goroutine for the given 'id'.
+func (d *Debugger) FindGoroutine(id int) (*proc.G, error) {
+	d.targetMutex.Lock()
+	defer d.targetMutex.Unlock()
+
+	g, err := proc.FindGoroutine(d.target, id)
```
aarzilli, line 975: Just `return proc.FindGoroutine(...)` is enough here.

suzmue:
> > Why wouldn't it make sense to return the err from the last command inside of debugger State?
> 
> I don't follow. Isn't it there already?

In this code snippet:
`
state, err := debugger.Command(...)
`

`err` contains the errror, but `state` does not, unless `err.(proc.ErrProcessExited)`.

The `Err` field in the result of calling `debugger.State(...)` is also not set.

service/debugger/debugger.go
suzmue: Done.

service/dap/server.go
suzmue: Oh that actually wouldn't be an error, that would be stopped at a breakpoint. Not sure if there is a case where this comment applies actually.

aarzilli:
> > > Why wouldn't it make sense to return the err from the last command inside of debugger State?
> > 
> > 
> > I don't follow. Isn't it there already?
> 
> In this code snippet:
> `state, err := debugger.Command(...)`
> 
> `err` contains the errror, but `state` does not, unless `err.(proc.ErrProcessExited)`.
> 
> The `Err` field in the result of calling `debugger.State(...)` is also not set.

That field was added because of how the client works, debugger shouldn't set it at all that it does for process exited is an error. We could have always used it but we didn't and we couldn't change it now due to backwards compatibility.

service/dap/server.go

service/dap/server.go

service/dap/server.go
suzmue: I was originally ignoring them to do a best effort approach to filling out the exception info. Though if we can't verify that this goroutine had an exception, we should return an error. Fixed.

suzmue:
Here are examples of how the exception info is used in VS Code:

Bad access error:

Before:
Screen Shot 2021-04-28 at 3 02 56 PM
After:
Screen Shot 2021-04-28 at 3 04 54 PM


Panic:

Before:
Screen Shot 2021-04-28 at 3 03 16 PM
After:
Screen Shot 2021-04-28 at 3 04 37 PM

Concurrent Map Read/Write:

Before:
Screen Shot 2021-04-28 at 3 03 45 PM

After:
Screen Shot 2021-04-28 at 3 04 19 PM

polinasok:
> > > > Why wouldn't it make sense to return the err from the last command inside of debugger State?
> > > 
> > > 
> > > I don't follow. Isn't it there already?
> > 
> > 
> > In this code snippet:
> > `state, err := debugger.Command(...)`
> > `err` contains the errror, but `state` does not, unless `err.(proc.ErrProcessExited)`.
> > The `Err` field in the result of calling `debugger.State(...)` is also not set.
> 
> That field was added because of how the client works, debugger shouldn't set it at all that it does for process exited is an error. We could have always used it but we didn't and we couldn't change it now due to backwards compatibility.

Can another field be added? LastErr?

polinasok:
Screenshots look great. Thank you.

service/dap/server.go
polinasok: You never know. I always find new surprises :) You could check if goroutine is set and if not then send an output event.

aarzilli (approved):
LGTM

polinasok (approved):

PR #2446, opened 2021-04-22
proc: add way to use CPU registers in expressions

aarzilli:
Changes the expression evaluation code so that register names, when not
shadowed by local or global variables, will evaluate to the current
value of the corresponding CPU register.

This allows a greater flexibility with displaying CPU registers than is
possible with using the ListRegisters API call. Also it allows
debuggers users to view register values even if the frontend they are
using does not implement a register view.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -431,7 +431,7 @@ Print contents of CPU registers.
 
 	regs [-a]
 
-Argument -a shows more registers.
+Argument -a shows more registers. Individual registers can also be displayed by 'print' and 'display'. See [Documentation/cli/expr.md.](//github.com/go-delve/delve/tree/master/Documentation/cli/expr.md.)
 
 
 ## restart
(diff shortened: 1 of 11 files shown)


derekparker:
This is a cool idea, I like it a lot. My initial thoughts for the shadowing problem is why not `$RAX` or `%RAX` instead of `_RAX`?

aarzilli:
> This is a cool idea, I like it a lot. My initial thoughts for the shadowing problem is why not `$RAX` or `%RAX` instead of `_RAX`?

We'd have to write our own parser, instead of using `go/parser`. I don't think it's worth it.

derekparker (approved):
LGTM

vsivsi:
Does this work on darwin/amd64 ? 

I see:

```
(dlv) print %x RAX
c0001ec000
(dlv) print %x XMM0
Command failed: could not find symbol value for XMM0
```

vsivsi:
@aarzilli   A quick question about this... Is there some reason why this doesn't work?

```
(dlv) print %x RAX.uint8
[8]void [80,d1,16,1,0,0,0,0]
(dlv) print %x RAX.uint8[0]
Command failed: can not index "RAX.uint8"
```

Is it because the type is being seen as `[8]void` instead of `[8]uint8`?

Presumably there should be some way to "go all the way" and have these pseudo-array views of registers actually behave like arrays in expressions in display/watch interfaces?

PR #2450, opened 2021-04-23
service/dap: implement function breakpoints

suzmue:
The SetFunctionBreakpointsRequests allows users to set a breakpoint by specifying
a function name. These need to be distinguishable from breakpoints set using
the SetBreakpointsRequest. This change does this by setting the name of a function
breakpoint to a non-empty value.

We first find the location of the function using FindLocation. This allows us to support
users that want to set a breakpoint at "functionName" without having to specify
"main.functionName" for example. Then the address of the location is used to set the
breakpoint for the function.


service/dap/server.go
```diff
@@ -1541,10 +1528,103 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	response := &dap.SetFunctionBreakpointsResponse{Response: *newResponse(request.Request)}
+	response.Body.Breakpoints = make([]dap.Breakpoint, len(request.Arguments.Breakpoints))
+	for i, want := range request.Arguments.Breakpoints {
+		spec, err := locspec.Parse(want.Name)
+		if err != nil {
+			response.Body.Breakpoints[i].Verified = false
+			response.Body.Breakpoints[i].Message = err.Error()
+			continue
+		}
+		if _, ok := spec.(*locspec.RegexLocationSpec); ok {
```
aarzilli, line 1568: Are we ok with users being able to say, for example, `+1` when setting a function breakpoint and that creates a breakpoint on the next line that the editor thinks is a function breakpoint?

Or to type in `main.go:15` as a function breakpoint.

service/dap/server_test.go
```diff
@@ -1738,6 +1738,218 @@ func TestSetBreakpoint(t *testing.T) {
...
+					expectSetFunctionBreakpointsResponse([]Breakpoint{{-1, "", false, "Location \"String\" ambiguous"}})
+
+					client.SetFunctionBreakpointsRequest([]dap.FunctionBreakpoint{
+						{Name: "main.String"},
+					})
+					expectSetFunctionBreakpointsResponse([]Breakpoint{{-1, "", false, "Location \"main.String\" ambiguous"}})
+					client.SetFunctionBreakpointsRequest([]dap.FunctionBreakpoint{
+						{Name: "ReadFile"},
+					})
+					expectSetFunctionBreakpointsResponse([]Breakpoint{{-1, "", false, "Location \"ReadFile\" ambiguous"}})
```
aarzilli, line 1847: Apparently this isn't ambiguous on 1.14 and 1.15. See test failure in teamcity.

service/dap/server.go
```diff
@@ -1541,10 +1528,103 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+			response.Body.Breakpoints[i].Message = err.Error()
+			continue
+		}
+		if len(locs) == 0 {
+			response.Body.Breakpoints[i].Verified = false
+			response.Body.Breakpoints[i].Message = fmt.Sprintf("no location found for %q", want.Name)
+			continue
+		}
+		if len(locs) > 0 {
+			// TODO(suzmue): what should we do if there is more than one location?
```
aarzilli, line 1591: This never happens under the given circumstances, you can remove this comment.

service/dap/server.go
```diff
@@ -1541,10 +1528,103 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	// If a breakpoint:
+	// -- exists and not in request => ClearBreakpoint
+	// -- exists and in request => AmendBreakpoint
+	// -- doesn't exist and in request => SetBreakpoint
+
+	// Clear all existing function breakpoints in the file.
+	// Function breakpoints are set with the Name field set to be the
+	// function name. We cannot use FunctionName to determine this, because
+	// the debugger sets the function name for breakpoints.
+	if err := s.clearMatchingBreakpoints(func(bp *api.Breakpoint) bool { return bp.Name != "" }); err != nil {
```
aarzilli, line 1553: I think you should actually check that bp.Name starts with "functionBreakpoint"

service/dap/server_test.go
suzmue: Done. I removed this check, since it would depend on dependencies of "fmt" and "io/ioutil", which are subject to change.

service/dap/server.go

service/dap/server.go

service/dap/server.go
```diff
@@ -1541,10 +1528,103 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+		return
+	}
+
+	// Set all requested breakpoints.
+	response := &dap.SetFunctionBreakpointsResponse{Response: *newResponse(request.Request)}
+	response.Body.Breakpoints = make([]dap.Breakpoint, len(request.Arguments.Breakpoints))
+	for i, want := range request.Arguments.Breakpoints {
+		spec, err := locspec.Parse(want.Name)
+		if err != nil {
+			response.Body.Breakpoints[i].Verified = false
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -1541,10 +1528,103 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+			// body should be since there will be multiple breakpoints created.
+			response.Body.Breakpoints[i].Verified = false
+			response.Body.Breakpoints[i].Message = "regex function names are not supported"
+			continue
+		}
+
+		// Find the location of the function name. CreateBreakpoint requires the name to include the base
+		// (eg main.functionName is supported but not functionName).
+		// We first find the location of the function, and then set breakpoints for that location.
+		locs, err := s.debugger.FindLocation(-1, 0, 0, want.Name, true, s.args.substitutePathClientToServer)
```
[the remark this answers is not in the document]
suzmue: I ended up adding the FindLocationSpec signature. It is possible that loc.Find() will use locStr, so I continued to pass the original string as well.

service/dap/server.go
suzmue: I updated this check to only allow NormalLocationSpecs, which should limit this to `func:line`, `file:line`. I documented that `file:line` is accepted in the tests.

While `func:line` is what this request is required to support, I'm not sure it is worth making sure that `file:line` is not also supported. On the other hand, address, offset, and regex locations do not make sense in the context of function breakpoints. We could update the locspec package to have a function that will only check for functions that match the spec and not files (this could also maybe be done in a separate change).

@polinasok Do you want have any thoughts on this?

suzmue:
TeamCity failure looks like its an unrelated build system failure

service/dap/server.go
```diff
@@ -1543,10 +1530,98 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	// If a breakpoint:
+	// -- exists and not in request => ClearBreakpoint
+	// -- exists and in request => AmendBreakpoint
+	// -- doesn't exist and in request => SetBreakpoint
+
+	// Clear all existing function breakpoints in the file.
+	// Function breakpoints are set with the Name field set to be the
+	// function name. We cannot use FunctionName to determine this, because
+	// the debugger sets the function name for breakpoints.
+	if err := s.clearMatchingBreakpoints(func(bp *api.Breakpoint) bool { return strings.HasPrefix(bp.Name, "functionBreakpoint") }); err != nil {
```
aarzilli, line 1555: Declare the string "functionBreakpoint" as a constant and use it here and below, where the breakpoints are created.

service/dap/server.go
aarzilli: For the record: I don't have an opinion on this.

service/dap/server.go

aarzilli (approved):
LGTM, besides the issue with SetFunctionBreakpoints allowing non-function breakpoints to be set that needs a decision.

service/dap/server.go
```diff
@@ -1543,10 +1530,102 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	// According to the spec we should "Replaces all existing function
+	// breakpoints with new function breakpoints." The simplest way is
+	// to clear all and then set all.
+	//
+	// TODO(polina): should we optimize this as follows?
+	// See https://github.com/golang/vscode-go/issues/163 for details.
+	// If a breakpoint:
+	// -- exists and not in request => ClearBreakpoint
+	// -- exists and in request => AmendBreakpoint
+	// -- doesn't exist and in request => SetBreakpoint
```
hyangah, line 1549: Curious - does this optimization proposal still make sense? I don't see any performance issue report related to this, or an evidence that this added complexity would be make things faster. And, I don't see how this can be cheaply implemented for function breakpoints.

hyangah (commented):
no opinion on the filename:line format. 
funcname:line is also no

polinasok (commented):
How does a user set this when the capability is enabled? Could you please share a screenshot for some context? Thank you.

service/dap/server.go
aarzilli: I think we discussed this a while ago, I don't think this optimization ever did make sense.

suzmue: Will remove the comment.

suzmue:
> How does a user set this when the capability is enabled? Could you please share a screenshot for some context? Thank you.

In VS Code, the UI looks like this (click `+` or run command `Add Function Breakpoint`):

Screen Shot 2021-04-28 at 12 02 41 PM

After the breakpoint is set:

Screen Shot 2021-04-28 at 12 03 15 PM

The triangle indicates a function breakpoint in the second screenshot.

It does not actually check if the capability is set, before enabling the UI since you can set these before having an active debug session.

service/dap/server.go
```diff
@@ -1543,10 +1530,102 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	for i, want := range request.Arguments.Breakpoints {
+		spec, err := locspec.Parse(want.Name)
+		if err != nil {
+			response.Body.Breakpoints[i].Message = err.Error()
+			continue
+		}
+		if loc, ok := spec.(*locspec.NormalLocationSpec); !ok || loc.FuncBase == nil {
+			// These are likely to resolve to multiple places, so it is unclear what the response
+			// body should be since there will be multiple breakpoints created.
+			response.Body.Breakpoints[i].Message = fmt.Sprintf("breakpoint name %q could not be parsed as a function", want.Name)
```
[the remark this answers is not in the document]

service/dap/server_test.go
```diff
@@ -1738,6 +1738,244 @@ func TestSetBreakpoint(t *testing.T) {
...
+					client.SetFunctionBreakpointsRequest([]dap.FunctionBreakpoint{
+						{Name: "SomeType.SomeFunction"},
+					})
+					expectSetFunctionBreakpointsResponse([]Breakpoint{{22, filepath.Base(fixture.Source), true, ""}})
+
+					// Expect the next breakpoint to be at SomeType.SomeFunction.
+					client.ContinueRequest(1)
+					client.ExpectContinueResponse(t)
+
+					if se := client.ExpectStoppedEvent(t); se.Body.Reason != "breakpoint" || se.Body.ThreadId != 1 {
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -1543,10 +1530,96 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+
 func (s *Server) onSetFunctionBreakpointsRequest(request *dap.SetFunctionBreakpointsRequest) {
-	s.sendNotYetImplementedErrorResponse(request.Request)
+	if s.noDebugProcess != nil {
+		s.sendErrorResponse(request.Request, UnableToSetBreakpoints, "Unable to set or clear breakpoints", "running in noDebug mode")
+		return
+	}
+	// TODO(polina): handle this while running by halting first.
+
+	// According to the spec we should "Replaces all existing function
```
polinasok, line 1544: This looks like a copy of the regular breakpoint comment. To be more precise the spec says:
Replaces all existing function breakpoints with new function breakpoints

service/dap/server.go
```diff
@@ -1543,10 +1530,96 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+		if loc, ok := spec.(*locspec.NormalLocationSpec); !ok || loc.FuncBase == nil {
+			// Other locations do not make sense in the context of function breakpoints.
+			// Regex locations are likely to resolve to multiple places and offset locations
+			// are only meaningful at the time the breakpoint was created.
+			response.Body.Breakpoints[i].Message = fmt.Sprintf("breakpoint name %q could not be parsed as a function. name must be in the format 'funcName', 'funcName:line' or 'fileName:line'.", want.Name)
+			continue
+		}
+
+		// Find the location of the function name. CreateBreakpoint requires the name to include the base
+		// (eg main.functionName is supported but not functionName).
```
polinasok, line 1575: e.g.

service/dap/server.go
```diff
@@ -1543,10 +1530,96 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+		if err != nil {
+			response.Body.Breakpoints[i].Message = err.Error()
+			continue
+		}
+		if len(locs) == 0 {
+			response.Body.Breakpoints[i].Message = fmt.Sprintf("no location found for %q", want.Name)
+			continue
+		}
+		if len(locs) > 0 {
+			s.log.Debugf("multiple locations found for %s", want.Name)
```
polinasok, line 1587: Can a message be set when a breakpoint is verified? This might be a useful thing to note.

service/dap/server.go
```diff
@@ -1543,10 +1530,96 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
...
+	// TODO(polina): handle this while running by halting first.
+
+	// According to the spec we should "Replaces all existing function
+	// breakpoints with new function breakpoints." The simplest way is
+	// to clear all and then set all.
+
+	// Clear all existing function breakpoints in the file.
+	// Function breakpoints are set with the Name field set to be the
+	// function name. We cannot use FunctionName to determine this, because
+	// the debugger sets the function name for breakpoints.
```
polinasok, line 1551: do you mean for *all* breakpoints?

polinasok (approved):
I only have a couple of minor comments. Otherwise, looks great! Thank you, @suzmue and @aarzilli.

service/dap/server.go
suzmue: Yes this can be done.

service/dap/server.go

service/dap/server.go

service/dap/server.go
suzmue: Updated to make this read better.

service/dap/server_test.go
```diff
@@ -1738,6 +1738,244 @@ func TestSetBreakpoint(t *testing.T) {
...
+					// Test function names in imported package.
+					// Issue #275
+					client.SetFunctionBreakpointsRequest([]dap.FunctionBreakpoint{
+						{Name: "io/ioutil.ReadFile"},
+					})
+					expectSetFunctionBreakpointsResponse([]Breakpoint{{-1, "ioutil.go", true, ""}})
+
+					// Issue #296
+					client.SetFunctionBreakpointsRequest([]dap.FunctionBreakpoint{
+						{Name: "/io/ioutil.ReadFile"},
```
polinasok, line 1825: This and previous test case look exactly the same. Is that on purpose?

suzmue: They are very similar, but the second has a starting `/`. I added a comment indicating that these tests were inspired by one of the integration tests for the rpc server, which specifically checks for these different locations.

service/dap/server.go
```diff
@@ -1543,10 +1530,97 @@ func (s *Server) onRestartRequest(request *dap.RestartRequest) {
 	s.sendNotYetImplementedErrorResponse(request.Request)
 }
 
-// onSetFunctionBreakpointsRequest sends a not-yet-implemented error response.
-// Capability 'supportsFunctionBreakpoints' is not set 'initialize' response.
+// functionBpPrefix is the prefix of bp.Name for every breakpoint bp set
+// in this request.
+var functionBpPrefix = "functionBreakpoint"
```
derekparker, line 1535: Can you make this const please.

derekparker (changes requested):
One small change, otherwise looks good.

service/dap/server.go
suzmue: Done!

service/dap/server_test.go
polinasok: Ah, right! I totally missed it! Stared a them for a while too :) Sorry for the false alarm.

service/dap/server_test.go
```diff
@@ -1771,6 +1773,8 @@ func TestSetFunctionBreakpoints(t *testing.T) {
 								}
 								continue
 							}
+							// Some function breakpoints may be in packages that have been imported and we do not control, so
+							// we do not always want to check breakpoints.
```
polinasok, line 1777: do you mean breakpoint *lines*? hence the -1 used to ignore?

service/dap/server.go
polinasok: My only issue is relative paths. We have been questioning their usage and resolution for everything else. As long as we clearly log/document what the situation is here, this should be fine.

service/dap/server_test.go

suzmue:
>My only issue is relative paths. We have been questioning their usage and resolution for everything else. As long as we clearly log/document/test what the situation is here, this should be fine.

I updated this change to reject relative paths and included test cases.

suzmue:
Updates golang/vscode-go#1348

derekparker (approved):
LGTM

derekparker:
Please just rebase and I will merge today.

derekparker:
@suzmue CI seems pretty upset after the rebase.

suzmue:
@derekparker Thanks for the heads up! Seems happier now (though TeamCity is still running)

derekparker:
@suzmue `TestSetFunctionBreakpoints` is now failing on Windows / AMD64.

suzmue:
@derekparker I think I fixed the tests but there is an execution timeout now in the build

derekparker:
@suzmue re-running, let's see what happens!

PR #2453, opened 2021-04-26
dap: auto-load partially loaded nested variables

hyangah:
Keep track of whether a variable is partially loaded in the variable
handle. When a partial variable is requested (since user manually
expands the variable through UI or debug console), try to load with
the expression, and replace the partially loaded previous variable
with the newly updated one.

Unfortunately, the newly loaded variable may be still partial
if the struct has many fields, or the variable is a map, array, or
struct. They will be addressed separately - probably by utilizing
pagination UI.

Keep track of the goid and frame information, so they are
available when the partially loaded variable is auto-loaded.

Before: 
Screen Shot 2021-04-26 at 11 25 43 AM

After:
Screen Shot 2021-04-26 at 11 25 06 AM

### DIFF
--- a/service/dap/handles.go
+++ b/service/dap/handles.go
@@ -24,6 +24,12 @@ type fullyQualifiedVariable struct {
 	fullyQualifiedNameOrExpr string
 	// True if this represents variable scope
 	isScope bool
+	// Goroutine
+	goid int
+	// Frame
+	frame int
+	// True if the current variable is partial.
+	isPartial bool
 }
 
 func newHandlesMap() *handlesMap {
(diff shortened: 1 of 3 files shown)


hyangah:
cc @polinasok @suzmue 

Again - this is a low priority, non-v0 feature. So, review isn't urgent.

And, the TeamCity build failure looks like a build system failure.

```
mingw32-make : The term 'mingw32-make' is not recognized as the name of a cmdlet, function, script file, or operable
program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
  At Z:\buildAgent\work\3aca8f960d346a9c\_scripts\test_windows.ps1:69 char:1
  + mingw32-make test
      + CategoryInfo          : ObjectNotFound: (mingw32-make:String) [], CommandNotFoundException
      + FullyQualifiedErrorId : CommandNotFoundException
      ```

polinasok:
Hmm, I think we got our wires crossed somewhere because I see auto-loading as my deliverable on our quarterly plan. It marked as in progress because I have been already working on it. I took a quick look at your version and you are using a different approach, which I think is more complicated a the first glance and has a couple of bugs, but I need to take a closer look to confirm.  It also doesn't produce an accurate inlined parent value. My solution (which is already in place for interfaces in the code) has that covered. I was not done with it because like you said it was not high priority, but let me upload a PR showing what I have.

hyangah:
Yes, this PR didn't intend to address the map, etc. I will close this PR and wait for the other PR.

PR #2457, opened 2021-04-27
service/dap: log stacktrace when panic is recovered

polinasok:


service/dap/server.go
```diff
@@ -318,6 +319,7 @@ func (s *Server) handleRequest(request dap.Message) {
 		// In case a handler panics, we catch the panic and send an error response
 		// back to the client.
 		if ierr := recover(); ierr != nil {
+			s.log.Debugf("stacktrace from recovered panic: \n" + string(debug.Stack()))
```
hyangah, line 322: s.log.Debug or s.log.Debugf("... \n%s", debug.Stack())
And I guess this is a serious error, so why not s.log.Error?

polinasok: Oops, this clearly evolved into something weird :) Done.

As for the error, didn't we just do a round of cleanups to silence excessive error logging because its was always there? Maybe I am remembering wrong. Would you log just the error or the stack as well? Log or write to the debug console? The error is not a bad idea,, but the stacktrace is something we would need to diagnose this further, so probably not very useful for the user by default.

This does get sent back as an error response, so it might show up in other parts of the UI. But now that I am looking at it, the panic might happen after we sent the response, so it might be ignored, so extra logging or output event would give more info. Also, we would need another one of these for the async goroutines in the other PR, won't we?

hyangah: Maybe raise the severity level for the internal errors so that it can be surfaced.

Does the panic from the async goroutine get recovered? Otherwise, I suppose it will crash dlv and go runtime will print the useful stack trace. If the panic is recovered and swallowed, then, yes need some logging (even though IMO crashing is better)

hyangah (approved):

service/dap/server.go
aarzilli: > even though IMO crashing is better

it really isn't, because it's going to take down the target process too.

aarzilli (approved):

PR #2461, opened 2021-04-29
Slightly better version check for gdb server on newer macOS (arm64)

### LINKED ISSUE #2436: g command fails on most recent version of debugserver on darwin/arm64

1. What version of Delve are you using (`dlv version`)? Version 1.6.0
2. What version of Go are you using? (`go version`)? Go1.16.2
3. What operating system and processor architecture are you using? Macmini m1 chip
4. What did you do? 

- cd /Documents && mkdir test && cd test && touch main.go 

- use vscode set main.go like 

 `
package main 
import "fmt"
 
func main() {
	a := 1
 	fmt.Print("123")
 	fmt.Println(a)
}
`

- add  break point at line 6 and line 8
- go mod init test
- go mod tidy
- create default debug config ./vscode/launch.json
`{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch Package",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "${workspaceFolder}"
        }
    ]
}`
- click Launch Package

5. What did you expect to see? run debug success and go to breakpoint
6. What did you see instead?
image

无法加载源“”。means can not  load source

what should i do?

oxisto:
Instead of purely checking for a specific version of gdbserver (`"1200"`), we assume that all versions greater than 1200 are affected and thus `p.gcmdok` needs to be set to false.

Fixes #2436

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -330,10 +330,11 @@ func (p *gdbProcess) Connect(conn net.Conn, path string, pid int, debugInfoDirs 
 
 				// Workaround for darwin arm64. Apple's debugserver seems to have a problem
 				// with not selecting the correct thread in the 'g' command and the returned
-				// registers are empty / an E74 error is thrown.
+				// registers are empty / an E74 error is thrown. This was reported to LLVM
+				// as https://bugs.llvm.org/show_bug.cgi?id=50169
 				version, err := strconv.ParseInt(v[len("version:"):], 10, 32)
 
-				if err == nil && version >= 1200 && p.bi.Arch.Name == "arm64" {
+				if err == nil && version >= 1200 && version <= 1205 && p.bi.Arch.Name == "arm64" {
 					p.gcmdok = false
 				}
 			}
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -331,7 +331,9 @@ func (p *gdbProcess) Connect(conn net.Conn, path string, pid int, debugInfoDirs 
 				// Workaround for darwin arm64. Apple's debugserver seems to have a problem
 				// with not selecting the correct thread in the 'g' command and the returned
 				// registers are empty / an E74 error is thrown.
-				if v[len("version:"):] == "1200" && p.bi.Arch.Name == "arm64" {
+				version, err := strconv.ParseInt(v[len("version:"):], 10, 32)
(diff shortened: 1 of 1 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -331,7 +331,9 @@ func (p *gdbProcess) Connect(conn net.Conn, path string, pid int, debugInfoDirs
 				// Workaround for darwin arm64. Apple's debugserver seems to have a problem
 				// with not selecting the correct thread in the 'g' command and the returned
 				// registers are empty / an E74 error is thrown.
-				if v[len("version:"):] == "1200" && p.bi.Arch.Name == "arm64" {
+				version, err := strconv.ParseInt(v[len("version:"):], 10, 32)
+
+				if err == nil && version >= 1200 && p.bi.Arch.Name == "arm64" {
```
aarzilli, line 336: I don't think we should blacklist every version, I think apple should fix debugserver instead or change it to report the `g` command as unsupported. I support adding a check for `version == "1205"` as well as referencing the upstream bug: https://bugs.llvm.org/show_bug.cgi?id=50169

If they release another version without fixing the problem we'll blacklist it definitively.

oxisto: Suggestion: checking for `version >= 1200 && version <= 1205`?

aarzilli: Ok. But I wonder why no one complained about version 1201. Does it even exist?

oxisto: I think it basically refers to the Xcode version, and yes 12.1, 12.2 and so do exist for Xcode at least.

aarzilli: If that were the case, then why did no one complain?

aarzilli (approved):
LGTM, thank you.

ebilling:
Is this being merged soon?

derekparker (approved):
LGTM

MarcelXia:
The issue is back on the latest Go 1.16.5 :-( and since the code changes, I'm not able to patch it again.

oxisto:
> The issue is back on the latest Go 1.16.5 :-( and since the code changes, I'm not able to patch it again.

It is probably best to open up a new issue - but as far as I see, the original is problem is now worked around independently of the version by just disabling the 'g' command on all Mac OS systems if using the `debugserver` (which is the default).

PR #2463, opened 2021-05-01
proc: correctly truncate the result of binary ops on integers

### LINKED ISSUE #2454: The >> and << operators extend the integer size.
dlv version 1.6.0, Build: $Id: 8cc9751909843dd55a46e8ea2a561544f70db34d $
go version go1.16.3 linux/amd64
amd r 3600, linux 64 bits

I have a main.go like this:
```
$cat main.go
package main

import "fmt"

func main() {
	a := uint32(0xffffffff)
	x := a << 33
	fmt.Prinln(x) // 0
}
```
If you run this code you will see that the shift operator returns 0 as _shifts_.
However when I break at line `x := a << 33` and run `p a << 33` in the dlv debugger, it returns `36893488138829168640` as the number ware arbitrary size (or 64 bits).

This becomes a problem when comparing to 0, because if I add ==0 at the end of line 7 the program prints true, while dlv will print false for the same expression.

I expect the << and >> operators to not extend the size of the integer in dlv, like they do in go.

aarzilli:
Truncates the result of binary operations on integers to the size of
the resulting type.

Also rewrites convertInt to not require allocations.

Fixes #2454

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -3,7 +3,6 @@ package proc
 import (
 	"bytes"
 	"debug/dwarf"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"go/ast"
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM

JAicewizard:
thanks!

PR #2464, opened 2021-05-03
Documentation: correct the flag name used to configure log destination

hyangah:
The documentation mentions flags that do not exist. Replace them with
the existing, --log-dest flag.


hyangah:
The TeamCity build failure is the test infra failure, not due to this pr.

aarzilli (approved):

PR #2465, opened 2021-05-03
service/dap: add type information to dap variables

suzmue:
When the client declares the capability `supportsVariableType`, we can additionally send back types
information for the variables we send. We already have this information easily accessible through the
`.TypeString()` method. This change adds this to the variables response if `supportsVariableType=true`.


suzmue:
cc @polinasok @hyangah since we were talking about this earlier. This is low priority and can be put on hold until later.

Before:
image
image

After:
image
image

service/dap/server.go
```diff
@@ -1289,14 +1317,23 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
...
 			children[i] = dap.Variable{
 				Name:               name,
 				EvaluateName:       cfqname,
+				Type:               ctype,
 				Value:              cvalue,
 				VariablesReference: cvarref,
 			}
 		}
 	}
+	if !s.clientCapabilities.supportsVariableType {
```
polinasok, line 1330: How about a helper?

```
func getTypeIfSupported(var proc.Variable) string {
    if !s.clientCapabilities.supportsVariableType {
        return ""
    } else {
        return var.TypeString()
    }
}

... 
dap.Variable{
    ...
    Type: getTypeIfSupported(children[i])
    ...
}
```

service/dap/server.go
```diff
@@ -1233,6 +1258,7 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 				kvvar := dap.Variable{
 					Name:         key,
 					EvaluateName: valexpr,
+					Type:         valType, // The type of a key/value pair is the value type.
```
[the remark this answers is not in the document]
suzmue: I  think thats a great idea!

I had put just the value type because it is what I had seen from other debug adapters (python and typescript), but I think your idea would be more clear. 

I'll add a screenshot for this.

suzmue:
key/val types with the new type string:

Screen Shot 2021-05-06 at 1 22 11 PM

service/dap/server.go
```diff
@@ -1470,18 +1501,33 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
...
 		Body:     dap.VariablesResponseBody{Variables: children},
 	}
 	s.send(response)
 }
 
+func (s *Server) getTypeIfSupported(v *proc.Variable) string {
+	if !s.clientCapabilities.supportsVariableType {
+		return ""
+	}
+	return v.TypeString()
```
aarzilli, line 1528: Would you mind changing the definition of TypeString in proc to be like this:

```
func (v *Variable) TypeString() string {
	if v == nilVariable {
		return "nil"
	}
	if v.DwarfType == nil {
		return v.Kind.String()
	}
	if v.DwarfType.Common().Name != "" {
		return v.DwarfType.Common().Name
	}
	r := v.DwarfType.String()
	if r == "*void" {
		if i := v.DwarfType.Common().Index; i < len(v.bi.Images) && len(v.bi.Images[i].compileUnits) > 0 && v.bi.Images[i].compileUnits[0].isgo {
			r = "unsafe.Pointer"
		}
	}
	return r
}
```

aarzilli: PS. edited

service/dap/server.go
suzmue: Done.

service/dap/server.go
suzmue: Done!

PR #2467, opened 2021-05-06
TeamCity: fix Windows builds

aarzilli:
Bintray is shutting down and the URL we used to install mingw is no
longer available. Use chocolatey instead.


derekparker (approved):
LGTM

PR #2468, opened 2021-05-06
pkg/proc: Enable CGO Stacktrace tests on arm64

derekparker:
These seem to magically work again on my M1 Mac so, enabling them again.


derekparker:
Now it seems to be failing on Linux/ARM64.

aarzilli (approved):
LGTM

PR #2469, opened 2021-05-07
proc: add support for debugCallV2

mknyszek:
This change adds support for the new debug call protocol which had to
change for the new register ABI introduced in Go 1.17.

Summary of changes:
- Abstracts over the debug call version depending on the Go version
  found in the binary.
- Uses R12 instead of RAX as the debug protocol register when the binary
  is from Go 1.17 or later.
- Creates a variable directly from the DWARF entry for function
  arguments to support passing arguments however the ABI expects.
- Computes a very conservative stack frame size for the call when
  injecting a call into a Go process whose version is >=1.17.
- Fixes bug in fakeFunctionEntryScope.

### DIFF
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -42,8 +42,9 @@ import (
 const (
 	debugCallFunctionNamePrefix1 = "debugCall"
 	debugCallFunctionNamePrefix2 = "runtime.debugCall"
-	debugCallFunctionName        = "runtime.debugCallV1"
+	maxDebugCallVersion          = 2
 	maxArgFrameSize              = 65535
+	maxRegArgBytes               = 9*8 + 15*8 // TODO: Make this generic for other platforms.
 )
 
 var (
@@ -163,7 +164,7 @@ func EvalExpressionWithCalls(t *Target, g *G, expr string, retLoadCfg LoadConfig
 		return errFuncCallInProgress
 	}
 
-	dbgcallfn := bi.LookupFunc[debugCallFunctionName]
+	dbgcallfn, _ := debugCallFunction(bi)
 	if dbgcallfn == nil {
 		return errFuncCallUnsupported
 	}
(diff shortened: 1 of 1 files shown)


mknyszek:
Hey @aarzilli, this seems to pass at least some of the tests that were failing due to Go 1.17 issues, but there are still some problems. I'll get those fixed up and then we can try to figure out how to merge this with your PR.

Also, this builds on your change to make it easier to merge into your PR (including the `fakeFunctionEntryScope` fix).

pkg/proc/fncall.go
```diff
@@ -1016,3 +1074,35 @@ func findCallInjectionStateForThread(t *Target, thread Thread) (*G, *callInjecti
...
+		name := debugCallFunctionNamePrefix2 + "V" + strconv.Itoa(version)
+		fn, ok := bi.LookupFunc[name]
+		if ok && fn != nil {
+			return fn, version
+		}
+	}
+	return nil, 0
+}
+
+// debugCallProtocolReg returns the register ID (as defined in pkg/dwarf/regnum)
```
aarzilli, line 1093: s/register ID/DWARF register number/

pkg/proc/fncall.go
```diff
@@ -554,7 +560,16 @@ func funcCallCopyOneArg(scope *EvalScope, fncall *functionCallState, actualArg *
 	//TODO(aarzilli): autmoatic wrapping in interfaces for cases not handled
 	// by convertToEface.
 
-	formalArgVar := newVariable(formalArg.name, uint64(formalArg.off+int64(argFrameAddr)), formalArg.typ, scope.BinInfo, scope.Mem)
+	var formalArgVar *Variable
+	if formalArg.dwarfEntry != nil {
+		var err error
+		formalArgVar, err = extractVarInfoFromEntry(scope.BinInfo, scope.image(), scope.Regs, scope.Mem, formalArg.dwarfEntry)
```
aarzilli, line 566: I think this is wrong for arguments that end up on the stack. `scope` here is the scope of the call so it doesn't have the correct value for any of the registers, and in particular SP is wrong, it should be `cfa`. I think you need to make a new scope from the thread by calling `GoroutineScope(thread)` right before the call to `funcCallEvalArgs` (disclaimer: this advice is untested and could be wrong).

aarzilli: Actually it should be: before the call to `callOp` that comes right before the call to `funcCallEvalArgs`.

aarzilli: Actually, I'm not sure about which one is the correct one.

aarzilli:
I have incorporated this, with fixes, in https://github.com/go-delve/delve/pull/2451. Thank you for your help on this.

PR #2470, opened 2021-05-08
make: Add support for dlv-dap installation and update

lggomez:
This PR includes support for dlv-dap installation and update of the dlv-dap binary, as specified by https://github.com/golang/vscode-go/blob/master/docs/dlv-dap.md#updating-dlv-dap

PTAL @hyangah @aarzilli


_scripts/make.go
```diff
@@ -110,6 +113,36 @@ This option can only be specified if testset is basic or a single package.`)
 	return RootCommand
 }
 
+// copyDAPBinary replicates the dlv installation into dlv-dap binary
```
aarzilli, line 116: I think this should make a symlink with `os.Link` instead of copying it.

_scripts/make.go
```diff
@@ -52,9 +53,11 @@ func NewMakeCommands() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			tagFlag := prepareMacnative()
 			execute("go", "install", tagFlag, buildFlags(), DelveMainPackagePath)
+			dlvExecutablePath := installedExecutablePath()
 			if runtime.GOOS == "darwin" && os.Getenv("CERT") != "" && canMacnative() {
-				codesign(installedExecutablePath())
+				codesign(dlvExecutablePath)
 			}
+			copyDAPBinary(dlvExecutablePath)
```
aarzilli, line 60: I think this should only be done if a flag is passed to the install command, like `go run _scripts/make.go install -dap`. This thing is very specific to VSCode-go and Delve is used in other contexts too.

_scripts/make.go
lggomez: Done

_scripts/make.go
lggomez: PS: This is the way I found to do it since I'm a bit of a novice to makefiles, if you know a better solution I'd appreciate it

_scripts/make.go
```diff
@@ -110,6 +117,24 @@ This option can only be specified if testset is basic or a single package.`)
 	return RootCommand
 }
 
+// updateDAPSymlink replicates the dlv executable installation into a dlv-dap symlink
+func updateDAPSymlink(dlvExecutablePath string) {
+	dlvDAPExecutablePath := fmt.Sprintf("%s-dap", dlvExecutablePath)
+
+	if _, err := os.Lstat(dlvDAPExecutablePath); err == nil {
+		if err := os.Remove(dlvDAPExecutablePath); err != nil {
+			log.Println(err.Error())
```
aarzilli, line 126: return

_scripts/make.go
```diff
@@ -110,6 +117,24 @@ This option can only be specified if testset is basic or a single package.`)
...
+		}
+	} else if !os.IsNotExist(err) {
+		log.Println(err.Error())
```
aarzilli, line 129: return

Makefile
```diff
@@ -7,7 +7,7 @@ build:
 	@go run _scripts/make.go build
 
 install:
-	@go run _scripts/make.go install
+	@go run _scripts/make.go install $(ARGS)
```
aarzilli, line 10: I would have added a new target:

```
install-dap:
        @go run _scripts/make.go install --dap
```

but this is also fine.

_scripts/make.go
hyangah: VSCode Go's installing dlv@master as `dlv-dap` is a temporary workaround which we plan to get rid of as soon as dlv dap becomes relatively stable (at that point, we probably want to install one single 'dlv' binary used for both legacy/dlv-dap adapter modes). 

You won't need to use make to install your dev version of dlv. From the delve project directory,

go build -o <path_to_dlv_dap_dest> ./cmd/dlv

lggomez:
After seeing @hyangah's comment on the current state of the dlv-dap binary maybe it is overkill to have this feature since it is going to be deprecated and it may be the best to leave this PR out. I'll close it for the time being

PR #2472, opened 2021-05-10
service/dap: support setting breakpoints while running

polinasok:
Unlike vscode-go, this implementation does not resume execution after halting and setting breakpoints. Resuming automatically turns out to be much trickier than just skipping a stopped event and continuing. It is possible that the stop occurred due to reasons other than our manual stop (breakpoint, call return), so we would need to detect thtat and report any stops that the user might be interested in. In addition, next would be cancelled by the halt command, so to resume it, we would need to implement an alternative version of halt that doesn't cancel next. So for now, we will just be reporting the pause to the user and let them decide when and how to resume execution manually.

Updates #1515
Updates https://github.com/golang/vscode-go/issues/1475

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -431,7 +431,7 @@ func (s *Server) handleRequest(request dap.Message) {
 			if inProgress == api.Continue {
 				s.overrideStopReason <- skipStop
 			} else {
-				// This would one of the step commands, which get  cancelled by debugger.
+				// This would be one of the step commands, which get cancelled by debugger.
 				// We do not support any other running commands at this time, but if a new one
 				// gets added and this code doesn't get updated, the safest default behavior
 				// would be to declare the operation cancelled and let the user resume manually,
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -89,9 +89,6 @@ type Server struct {
 	listener net.Listener
 	// stopTriggered is closed when the server is Stop()-ed.
 	stopTriggered chan struct{}
-	// overrideStopReason allows the run goroutine to communicate to the
-	// command goroutine how to process a stop
-	overrideStopReason chan string
 	// reader is used to read requests from the connection.
 	reader *bufio.Reader
 	// log is used for structured logging.
@@ -172,14 +169,13 @@ func NewServer(config *service.Config) *Server {
 	logflags.WriteDAPListeningMessage(config.Listener.Addr().String())
 	logger.Debug("DAP server pid = ", os.Getpid())
 	return &Server{
-		config:             config,
-		listener:           config.Listener,
-		stopTriggered:      make(chan struct{}),
-		overrideStopReason: make(chan string, 1),
-		log:                logger,
-		stackFrameHandles:  newHandlesMap(),
-		variableHandles:    newVariablesHandlesMap(),
-		args:               defaultArgs,
+		config:            config,
+		listener:          config.Listener,
+		stopTriggered:     make(chan struct{}),
+		log:               logger,
+		stackFrameHandles: newHandlesMap(),
+		variableHandles:   newVariablesHandlesMap(),
+		args:              defaultArgs,
 	}
 }
 
@@ -343,8 +339,6 @@ func (s *Server) serveDAPCodec() {
 	}
 }
 
-const skipStop = ""
-
 // In case a handler panics, we catch the panic to avoid crashing both
 // the server and the target. We send an error response back, but
 // in case its a dup and ignored by the client, we also log the error.
@@ -389,15 +383,11 @@ func (s *Server) handleRequest(request dap.Message) {
 		return
 	}
 
-	// Non-blocking request handlers will signal when they are ready
-	// setting up for async execution, so more requests can be processed.
-	resumeRequestLoop := make(chan struct{})
-
 	// Most requests cannot be processed while the debuggee is running.
 	// We have a couple of options for handling these without blocking
 	// the request loop indefinitely when we are in running state.
 	// --1-- Return a dummy response or an error right away.
-	// --2-- Halt execution, process the request, resume execution.
+	// --2-- Halt execution, process the request, maybe resume execution.
 	// --3-- Handle such requests asynchronously and let them block until
 	// the process stops or terminates (e.g. using a channel and a single
 	// goroutine to preserve the order). This might not be appropriate
@@ -425,44 +415,13 @@ func (s *Server) handleRequest(request dap.Message) {
 			}
 			s.send(response)
 		case *dap.SetBreakpointsRequest:
-			// Halt running command
-			inProgress := s.debugger.RunningCommand()
-			// The channel is drained before running a command, but just in case drain again to avoid blocking
-			s.resetOverrideStopReason()
-			// TODO(polina): once api.Call is run asynchronously, treat it like api.Continue as well
-			if inProgress == api.Continue {
-				s.overrideStopReason <- skipStop
-			} else {
-				// This would be one of the step commands, which get cancelled by debugger.
-				// We do not support any other running commands at this time, but if a new one
-				// gets added and this code doesn't get updated, the safest default behavior
-				// would be to declare the operation cancelled and let the user resume manually,
-				s.overrideStopReason <- fmt.Sprintf("cancelled %s", inProgress)
-			}
 			s.log.Debug("halting execution to set breakpoints")
 			_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
 			if err != nil {
 				s.sendErrorResponse(request.Request, UnableToSetBreakpoints, "Unable to set or clear breakpoints", err.Error())
 				return
 			}
-			// Set breakpoints
 			s.onSetBreakpointsRequest(request)
-			// Resume or cancel command
-			if inProgress == api.Continue {
-				go func() {
-					defer s.recoverPanic(request)
-					s.log.Debug("resuming execution after setting breakpoints")
-					s.doRunCommand(api.Continue, resumeRequestLoop)
-				}()
-				<-resumeRequestLoop
-			} else {
-				s.send(&dap.OutputEvent{
-					Event: *newEvent("output"),
-					Body: dap.OutputEventBody{
-						Output:   fmt.Sprintf("WARNING: setting breakpoints halted execution and cancelled %q\n", inProgress),
-						Category: "stderr",
-					}})
-			}
 		default:
 			r := request.(dap.RequestMessage).GetRequest()
 			s.sendErrorResponse(*r, DebuggeeIsRunning, fmt.Sprintf("Unable to process `%s`", r.Command), "debuggee is running")
@@ -476,6 +435,10 @@ func (s *Server) handleRequest(request dap.Message) {
 	// to another goroutine. Please note that because of the running
 	// check above, there should be no more than one pending asynchronous
 	// request at a time.
+
+	// Non-blocking request handlers will signal when they are ready
+	// setting up for async execution, so more requests can be processed.
+	resumeRequestLoop := make(chan struct{})
 
 	switch request := request.(type) {
 	//--- Asynchronous requests ---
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -415,6 +424,37 @@ func (s *Server) handleRequest(request dap.Message) {
 				Body:     dap.ThreadsResponseBody{Threads: []dap.Thread{{Id: -1, Name: "Current"}}},
 			}
 			s.send(response)
+		case *dap.SetBreakpointsRequest:
+			// Halt running command
+			inProgress := s.debugger.RunningCommand()
+			if inProgress == api.Continue {
+				s.overrideStopReason <- skipStop
+			} else {
+				s.overrideStopReason <- fmt.Sprintf("cancelled %s", inProgress)
```
suzmue, line 433: could you include a comment here for why we need to cancel other requests?

service/dap/server_test.go
```diff
@@ -1855,14 +1856,79 @@ func TestSetBreakpoint(t *testing.T) {
...
+						t.Errorf("\ngot  %#v\nwant Reason='cancelled next' AllThreadsStopped=true ThreadId=0/1", se)
+					}
+					expectSetBreakpointsResponse(t, client, []Breakpoint{{11, fixture.Source, true, ""}})
+					oe := client.ExpectOutputEvent(t)
+					warning := "WARNING: setting breakpoints halted execution and cancelled \"next\"\n"
+					if oe.Body.Category != "stderr" || oe.Body.Output != warning {
+						t.Errorf("\ngot  %#v\nwant Category='stderr' Output=%q", oe, warning)
+					}
+
+					// We must continue since next got cancelled
```
suzmue, line 1901: This is a little confusing to me. Maybe: "We must continue since we are stopped where next was interrupted"

service/dap/server_test.go
```diff
@@ -1855,14 +1856,79 @@ func TestSetBreakpoint(t *testing.T) {
...
+					client.ExpectContinueResponse(t)
+
+					// Continue stops at line 11 (verifies that setting breakpoint during next worked)
+					se = client.ExpectStoppedEvent(t)
+					if se.Body.Reason != "breakpoint" || !se.Body.AllThreadsStopped || se.Body.ThreadId != 1 {
+						t.Errorf("\ngot  %#v\nwant Reason='breakpoint' AllThreadsStopped=true ThreadId=1", se)
+					}
+					handleStop(t, client, 1, "main.f", 11)
+
+					// We coninue and loop once more
```
suzmue, line 1912: s/coninue/continue

service/dap/server_test.go
```diff
@@ -1855,14 +1856,79 @@ func TestSetBreakpoint(t *testing.T) {
...
+					client.ContinueRequest(1)
+					client.ExpectContinueResponse(t)
+
+					// While its running, set breakpoint at line 10 and clear line 11
```
suzmue, line 1916: maybe add something about how we expect not to hit line 11 again before this request gets processed for the same reason as above (10 second sleep)

service/dap/server_test.go
```diff
@@ -1855,14 +1856,79 @@ func TestSetBreakpoint(t *testing.T) {
...
+					client.SetBreakpointsRequest(fixture.Source, []int{10})
+					expectSetBreakpointsResponse(t, client, []Breakpoint{{10, fixture.Source, true, ""}})
+
+					// Continue stops at line 10 (verifies that setting breakpoint during coninue worked)
```
suzmue, line 1920: s/coninue/continue

suzmue, line 1920: This could be either on the first or second loop around right? Maybe include that info in the comment about setting the breakpoint.

service/dap/server.go
```diff
@@ -415,6 +424,42 @@ func (s *Server) handleRequest(request dap.Message) {
...
+			}
+			s.log.Debug("halting execution to set breakpoints")
+			_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
+			if err != nil {
+				s.sendErrorResponse(request.Request, UnableToSetBreakpoints, "Unable to set or clear breakpoints", err.Error())
+			}
+			// Set breakpoints
+			s.onSetBreakpointsRequest(request)
+			// Resume or cancel command
+			if inProgress == api.Continue {
```
hyangah, line 448: Is it guaranteed that s.overrideStopReason is drained by now?
Won't it ever happen that the previous running command stops for other reasons before the message in line 432/438 was queued to the channel?

polinasok: That's a good point. I guess it could happen in theory that things stop running after we check isRunning, but before we do all this. We wouldn't want to go back to the idea of locking sections of code with a running lock, so that any related processing is grouped with the state that triggers it, would we? To avoid blocking on next write and to avoid next doCommand acting on a stale channel value, I think drain it at the beginning of doCommand should be sufficient. What do you think?

polinasok:
Failing test is TestAttachStopOnEntry, which I am hopefully fixing in the other PR.

service/dap/server.go
suzmue: I think we could also run into this scenario:

1. setBreakpointsRequest checks inProgressCommand == api.Continue
2. program stops at breakpoint
3. checks override reason and finds none.
4. sends stopped event for breakpoint
5. setBreakpointsRequest send overrideReason = skipStop
5. setBreakpointsRequest attempts to halt program.
6. setBreakpointsRequest sets breakpoints
7. setBreakpointsRequest continues

The UI would show that the program is stopped at a breakpoint when it is really running.

polinasok: This reminds me of the analysis I did in async PR (https://github.com/go-delve/delve/pull/2423#discussion_r620454562) when trying to lock sections of code to avoid things happening like this out of order. At the time I could only think of a case where a misleading is-running error would arrive after a stopped event. But the idea was the same: if you decouple checking is-running and acting on it, you might end up carrying out while-running actions when things are no longer running. I think I got this sorted out with a dedicated lock in some intermediate version of the previous PR, following the right lock/unlock order relative to the resumeLoop channel to avoid deadlocks. But the overall multi-lock design made people uncomfortable. What is the alternative? Maybe use this skipStop channel for two-way communication, so handling breakpoints checks on what doCommand reported (i.e. if it acted on that signal) before sending a follow-up continue?

polinasok: I can take a closer look tomorrow.

service/dap/server.go
```diff
@@ -415,6 +424,45 @@ func (s *Server) handleRequest(request dap.Message) {
...
 			}
 			s.send(response)
+		case *dap.SetBreakpointsRequest:
+			// Halt running command
+			inProgress := s.debugger.RunningCommand()
+			// The channel is drained before running a command, but just in case drain again to avoid blocking
+			s.resetOverrideStopReason()
+			// TODO(polina): once api.Call is run asynchronously, treat it like api.Continue as well
+			if inProgress == api.Continue {
+				s.overrideStopReason <- skipStop
```
aarzilli, line 434: I don't think this is a good way of doing this. By the time you send the halt command the target might be already in the process of stopping for a different reason, so you can't blindly skip the next stop, because it could be one that the user wants to see. You have to issue the halt command, examine the state that is returned to you to determine if you are the cause of the stop or not and then either:
- suppress the stop event
- set breakpoints
- automatically resume execution
or:
- propagate the stop event
- set breakpoints

Doing this correctly is complicated because you have to synchronize with both your own goroutines (which you have to do with mutexes and the likes) as well as with target threads (which you have to do by post-facto examination of its state). Personally, I wouldn't get into it, too complicated for such a marginally useful feature.

service/dap/server.go
hyangah: Is it too bad if setBreakpointsRequest triggers halt, and then fall through so it can be handled in the next loop like the other setBreakpointsRequests? Maybe the request can carry an extra info so after setBreakpointRequest is handled later, decide whether to continue or not depending on the stop reason. One drawback is the extra Stop event and Continue event, but will it be too visible for users?

service/dap/server.go
```diff
@@ -415,6 +424,45 @@ func (s *Server) handleRequest(request dap.Message) {
...
+			} else {
+				// This would be one of the step commands, which get cancelled by debugger.
```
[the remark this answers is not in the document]
suzmue: From my reading of the code in proc, when a step /next request is manually halted `ClearInternalBreakpoints` is always called.

It would be great if there is a way to be able to continue next requests, this is also a feature I am interested in for implementing logpoints.

service/dap/server.go
polinasok: Extra Continue event would solve the issue of the UI and state being out of sync, but it won't solve the issue that we should not have continued in the first place because the stopped happened for different reasons.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: We had this conversation a while back when trying to fix setting breakpoints in the old adapter:
https://github.com/golang/vscode-go/issues/787#issuecomment-713434803
I too saw `ClearInternalBreakpoints` and understood it as halt cancelling step.

service/dap/server.go
polinasok: The approach in this PR was inspired by how the original adapter handled this feature, which was clearly not quite right :)

I am hesitant not to support this at all because vscode users are used to this working.

Instead of asking the users to manually halt to set breakpoints, a compromise solution could be to support halt+setBreakpoints only and let the user continue as they see fit. This will save the users a step, and make it obvious what is going on.

As for understanding why the command stopped, this is something we might want to do a better job at anyway since we need to communicate the stop reasons and the details to the user and not just here, but in case of any stop. This could be done in a separate PR, but let me clarify the details here while we are on topic. Plus I could then add a TODO in this PR for what it would take to resume execution after setting a breakpoint while running.

If debugger.stopReason() is "manual", should we still be going over the threads and looking for any with a breakpoint? If the stop reason is anything other than "next finished", should we be checking NextInprogress==true and reporting (e.g. as a message in the Console) that a next operation is in progress and that hitting continue would complete it?

aarzilli: >If debugger.stopReason() is "manual", should we still be going over the threads and looking for any with a breakpoint? If the stop reason is anything other than "next finished", should we be checking NextInprogress==true and reporting (e.g. as a message in the Console) that a next operation is in progress and that hitting continue would complete it?

No. You should stop (i.e. send a stop event and not resume the process):

1. If StopReason is anything but manual
2. If any thread has a breakpoint or CallReturn set
3. If NextInProgress is false *and* the last command sent *by the user* is: next, step, stepOut, reverseNext, reverseStep or reverseStepOut

If you *do not* stop you should suppress the stop event that continue would send and execute the api.DirectionCongruentContinue command. This presumes that the "halt-without-cancelling-next" is implemented and used.
Letting the stop event go through and sending a new resume event doesn't work without explicit synchronization (the client could get them in the wrong order).

service/dap/server.go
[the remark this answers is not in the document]
aarzilli: What's a logpoint? Anyway it shouldn't be to hard to make a variant of halt that sets a flag so that ClearInternalBreakpoints isn't called in proc.(*Target).Continue.

service/dap/server.go
polinasok: Thank you, this is very helpful.
I have added a TODO for resuming execution with this information. The scope of this CL is resumed to just halt+set.
I will check what can be done to improve stopped event reporting in a different PR.

service/dap/server.go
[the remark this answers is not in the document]
suzmue: A log point is like a trace point but with user supplied message

From DAP spec for setBreakpoints arguments:
```
 /**
   * If this attribute exists and is non-empty, the backend must not 'break'
   * (stop)
   * but log the message instead. Expressions within {} are interpolated.
   * The attribute is only honored by a debug adapter if the capability
   * 'supportsLogPoints' is true.
   */
  logMessage?: string;
```

service/dap/server_test.go
```diff
@@ -1855,14 +1856,69 @@ func TestSetBreakpoint(t *testing.T) {
...
+			[]onBreakpoint{{
+				execute: func() {
+					// The program loops 3 times over lines 14-15-8-9-10-16
+					handleStop(t, client, 1, "main.main", 16) // Line that sleeps for 1 second
+
+					// We can set breakpoints while nexting
+					client.NextRequest(1)
+					client.ExpectNextResponse(t)
+					client.SetBreakpointsRequest(fixture.Source, []int{15}) // [16,] => [15,]
+					se := client.ExpectStoppedEvent(t)
```
aarzilli, line 1885: This is a race condition right? The response to set breakpoints could arrive before the stopped event?

aarzilli (approved):
LGTM

PR #2475, opened 2021-05-10
service/dap: warn users of debugging optimized functions

suzmue:
We want to be able to warn users about debugging optimized functions since variables
may not be what they expect. There is currently no way to show the users this warning
without returning an error to the entire request, so we can show this by annotating the
variable scopes instead.

Updates golang/vscode-go#1471


suzmue:
Screen shot of the variable pane:

Screen Shot 2021-05-10 at 2 17 05 PM

cc @polinasok @hyangah

aarzilli (approved):
LGTM

polinasok (approved):
LGTM

PR #2477, opened 2021-05-12
proc/native/linux: better handling of process death due to signals

aarzilli:
Handle the signaled status for the thread leader like we handle the
exited status, by returning ErrProcessExited and recording the killer
signal  in it.

Prior to this commit we would find out about the death of the thread
later in the loop, the condition would still be reported as
ErrProcessExited, but without recording the signal number anywhere.

Also fixes a bug in TestAttachStopOnEntry where the test would
inadvertently cause a SIGPIPE to be sent to the target process, making
it terminate early.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -364,39 +364,21 @@ func TestAttachStopOnEntry(t *testing.T) {
 		// 13 >> disconnect, << disconnect
 		client.DisconnectRequestWithKillOption(true)
 
-		// Both of these scenarios are somehow possible.
-		// Even though the program has an infininte loop,
-		// it apears that a halt can cause it to terminate.
-		// Since we are in async mode while running, we might receive messages in either order.
-		msg := client.ExpectMessage(t)
-		switch m := msg.(type) {
-		case *dap.StoppedEvent:
-			if m.Seq != 0 || m.Body.Reason != "pause" { // continue is interrupted
-				t.Errorf("\ngot %#v\nwant Seq=0 Reason='pause'", m)
-			}
-			oed := client.ExpectOutputEventDetachingKill(t)
-			if oed.Seq != 0 || oed.Body.Category != "console" {
-				t.Errorf("\ngot %#v\nwant Seq=0 Category='console'", oed)
-			}
-		case *dap.TerminatedEvent:
-			if m.Seq != 0 {
-				t.Errorf("\ngot %#v\nwant Seq=0'", m)
-			}
-			oep := client.ExpectOutputEventProcessExited(t, 0)
-			if oep.Seq != 0 || oep.Body.Category != "console" {
-				t.Errorf("\ngot %#v\nwant Seq=0 Category='console'", oep)
-			}
-			oed := client.ExpectOutputEventDetaching(t)
-			if oed.Seq != 0 || oed.Body.Category != "console" {
-				t.Errorf("\ngot %#v\nwant Seq=0 Category='console'", oed)
-			}
-		default:
-			t.Fatalf("got %#v, want StoppedEvent or TerminatedEvent", m)
+		msg := expectMessageFilterStopped(t, client)
+		if _, ok := msg.(*dap.OutputEvent); !ok {
+			// want detach kill output message
+			t.Errorf("got %#v, want *dap.OutputEvent", msg)
 		}
-		dResp := client.ExpectDisconnectResponse(t)
-		if dResp.Seq != 0 || dResp.RequestSeq != 13 {
-			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=13", dResp)
+		msg = expectMessageFilterStopped(t, client)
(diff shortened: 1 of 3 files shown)


aarzilli:
cc @polinasok I think there still are valid sequences of events that the server can send that TestAttachStopOnEntry can't handle but this is where the "halt kills the target process" thing was coming from. Cool bug.

aarzilli:
I'll do the pause/continue spam next.

polinasok:
Thank you for looking into this. 

> TestAttachStopOnEntry where the test would inadvertently cause a SIGPIPE to be sent to the target process

What exactly in the test caused SIGPIPE?

aarzilli:
> What exactly in the test caused SIGPIPE?

The stdout variable and any other variables holding a reference to the stdout pipe `*os.File` aren't alive for most of the test, the garbage collector is therefore allowed to collect the `os.File` corresponding to the pipe, this object has a finalizer that will close the file descriptor, writing to the write end of a pipe whose read end is closed causes the kernel to send a SIGPIPE, SIGPIPE isn't in the set of signals we are interested into so we forward it to the target process, killing it.

service/dap/server_test.go
```diff
@@ -397,6 +397,12 @@ func TestAttachStopOnEntry(t *testing.T) {
 		if dResp.Seq != 0 || dResp.RequestSeq != 13 {
 			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=13", dResp)
 		}
+		// If this call to KeepAlive isn't here there's a chance that stdout will
+		// be garbage collected (since it is no longer alive long before this
+		// point), when that happens, on unix-like OSes, the read end of the pipe
+		// will be closed by the finalizer and the target process will die by
+		// SIGPIPE, which the rest of this test does not expect.
+		runtime.KeepAlive(stdout)
```
hyangah, line 405: @bcmills noticed it can be actually the `cmd` leak. defer cmd.Wait would be one way to address that cleanly, but before calling cmd.Wait, stdout pipe needs to be drained or closed. https://golang.org/pkg/os/exec/#Cmd.StdoutPipe
It's strange that this issue is surfaced this way. :-(

aarzilli: There's a number of possible ways to fix this, anything that keeps a reference to the stdout pipe os.File alive will suffice. So, a goroutine that reads from scanOut, a deferred cmd.Wait, a deferred stdout.Close, passing cmd to fmt.Printf... the fact that cmd.Wait is never called doesn't really matter here, the problem is resources being released not resources being leaked.
Issues related to this get periodically reported to golang/go, the last instance I can find is https://github.com/golang/go/issues/41505 but there are much older versions.

service/dap/server_test.go
```diff
@@ -364,11 +364,11 @@ func TestAttachStopOnEntry(t *testing.T) {
 		// 13 >> disconnect, << disconnect
 		client.DisconnectRequestWithKillOption(true)
 
-		// Both of these scenarios are somehow possible.
```
polinasok, line 367: You removed the comment, but not the code for the alternative case, which shouldn't happen anymore.  I think in this case, it should go back to the original logic that this commit expanded.
https://github.com/go-delve/delve/pull/2423/commits/9c8e957134b1e271bb728f787712803831848385

service/dap/server_test.go
```diff
@@ -364,11 +364,11 @@ func TestAttachStopOnEntry(t *testing.T) {
...
-		// Even though the program has an infininte loop,
-		// it apears that a halt can cause it to terminate.
 		// Since we are in async mode while running, we might receive messages in either order.
 		msg := client.ExpectMessage(t)
+		if _, isOutput := msg.(*dap.OutputEvent); isOutput {
```
polinasok, line 369: What kind of message could that be?

polinasok (commented):
Thank you for the detailed explanation. I understand now. The SIGPIPE issue in the test helped uncover that the proc code could handle signals (not just SIGPIPE) better, reporting more the specific signal that caused the exit, not just the exit itself. In the meantime, the issue in the test is its own thing and needed to be handled by not having the signal at all, so termination happened when we asked for it with Detach, not before due to unrelated reasons.

service/dap/server_test.go
aarzilli: The `s.logToConsole("Detaching and terminating target process")`

service/dap/server_test.go
aarzilli: As far as I can see that wouldn't work because the stopped event can be received at any point (before the output, between the output and the disconnect or after the disconnect).

polinasok: Right, but there shouldn't be a TerminatedEvent in the mix anymore, which is gone now.

polinasok (approved):
LGTM

service/dap/server_test.go
aarzilli: I believe I have now removed that check.

derekparker (approved):
LGTM

PR #2480, opened 2021-05-14
service/dap: send 'continued' event on next/step/stepout requests

suzmue:
Next/step/stepout requests assume that execution is only resumed
on the thread that was specified. We cannot resume execution on
only a single thread, so we need to send a continued event to let
the client know that all threads have resumed.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1221,6 +1221,15 @@ func stoppedGoroutineID(state *api.DebuggerState) (id int) {
 // due to an error, so the server is ready to receive new requests.
 func (s *Server) doStepCommand(command string, threadId int, asyncSetupDone chan struct{}) {
 	defer s.asyncCommandDone(asyncSetupDone)
+	// All of the threads will be continued by this request, so we need to send
+	// a continued event so the UI can properly reflect the current state.
+	s.send(&dap.ContinuedEvent{
+		Event: *newEvent("continued"),
+		Body: dap.ContinuedEventBody{
+			ThreadId:            threadId,
+			AllThreadsContinued: true,
+		},
+	})
 	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.SwitchGoroutine, GoroutineID: threadId}, nil)
 	if err != nil {
 		s.log.Errorf("Error switching goroutines while stepping: %v", err)
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -1969,6 +1969,7 @@ func TestSetBreakpointWhileRunning(t *testing.T) {
 					// We can set breakpoints while nexting
 					client.NextRequest(1)
 					client.ExpectNextResponse(t)
+					client.ExpectContinuedEvent(t)
 					client.SetBreakpointsRequest(fixture.Source, []int{15}) // [16,] => [15,]
 					se, br := expectSetBreakpointsResponseAndStoppedEvent(t, client)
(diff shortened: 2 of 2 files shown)


suzmue:
Before:
Screen Shot 2021-05-14 at 12 12 51 PM
After:
Screen Shot 2021-05-14 at 12 18 42 PM

service/dap/server_test.go
```diff
@@ -1679,6 +1680,7 @@ func TestGlobalScopeAndVariables(t *testing.T) {
 					// Step into pkg.AnotherMethod()
 					client.StepInRequest(1)
 					client.ExpectStepInResponse(t)
+					client.ExpectContinuedEvent(t)
```
aarzilli, line 1683: Wouldn't it be better to change the code generator so that ExpectStepInResponse etc also always check for a continued event?

aarzilli (commented):
Is the failure with TestContinueOnEntry related?

suzmue:
The TestContinueOnEntry failure is unrelated (it looks like a similar bug to TestPreSetBreakpoint)

service/dap/server_test.go
polinasok: I think it is helpful to have explicit sequences in the tests. They serve as documentation of sorts.

service/dap/server.go
```diff
@@ -1221,6 +1221,15 @@ func stoppedGoroutineID(state *api.DebuggerState) (id int) {
 // due to an error, so the server is ready to receive new requests.
 func (s *Server) doStepCommand(command string, threadId int, asyncSetupDone chan struct{}) {
 	defer s.asyncCommandDone(asyncSetupDone)
+	// All of the threads will be continued by this request, so we need to send
+	// a continued event so the UI can properly reflect the current state.
```
polinasok, line 1225: It's good we can accomplish what we need this way, but seems like this was not the intended use based on the description in the spec. Do you think it would make more sense to have AllThreadsContinued added to Step/Next/etc responses?

polinasok (approved):

service/dap/server.go
suzmue: That would be great, since all of our commands that continue execution will resume all threads.

This would need to be added to a many responses, since only `ContinueRequest` includes the option and there are many flavors of commands that resume execution (ReverseContinue, StepOut, Next, StepInstruction, etc).

polinasok: https://github.com/microsoft/debug-adapter-protocol/issues/192

service/dap/server_test.go
aarzilli: If you insist. I think it's very repetitive. And besides, the specific sequence is an implementation quirk, it would still be compliant if they came in in the other order.

polinasok: I am hoping they will respond to my feature request and we wouldn't need any of them.

derekparker: I'm fine with the extra verbosity for now, but I agree with @aarzilli that this could be tedious / potentially fragile (testing the implementation rather than the behavior).

derekparker:
Before I merge @suzmue could you please rebase on master since https://github.com/go-delve/delve/pull/2489 has been merged and should fix the only failing test in this build.

service/dap/server_test.go
suzmue: I think I will make the change to the code generator, since this will be easier to remove if the dap change gets implemented and we can someday stop sending these events, instead of finding all the continuedevents that need to be removed.

suzmue: I can do that in a different PR since its more of a code health issue.

derekparker (approved):
LGTM

PR #2482, opened 2021-05-14
service/dap: move build error to output event

suzmue:
Since the build error may be long, including it in the error response may not be a good way to get the info to the user. This change moves that information to an output event. It also includes the build command that produced the error.

Updates golang/vscode-go#1425

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -723,17 +723,24 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		}
 
 		s.log.Debugf("building binary at %s", debugbinary)
+		var cmd string
 		var out []byte
 		switch mode {
 		case "debug":
-			out, err = gobuild.GoBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
+			cmd, out, err = gobuild.GoBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
 		case "test":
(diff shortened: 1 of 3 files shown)


aarzilli (approved):
LGTM

service/dap/server.go
```diff
@@ -723,17 +723,24 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
 		switch mode {
 		case "debug":
-			out, err = gobuild.GoBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
+			cmd, out, err = gobuild.GoBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
 		case "test":
-			out, err = gobuild.GoTestBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
+			cmd, out, err = gobuild.GoTestBuildCombinedOutput(debugbinary, []string{program}, buildFlags)
 		}
 		if err != nil {
+			s.send(&dap.OutputEvent{
```
polinasok, line 735: Does this always show up? I vaguely remember we had some issue with the early messages in the session not showing up in the Debug Console.

suzmue: I think once the initialize request is sent the rest of the messages get logged.

polinasok (approved):
LGTM

PR #2483, opened 2021-05-15
service/dap: fix flaky TestPreSetBreakpoint test

suzmue:
Fix the flaky test. The threads request may be processed after the program has already stopped.


aarzilli (approved):
LGTM

PR #2484, opened 2021-05-15
pkg/proc/native/linux: fix target crashes induced by RequestManualStop

aarzilli:
A RequestManualStop received while the target program is stopped can
induce a crash when the target is restarted.

This is caused by the phantom breakpoint detection that was introduced
in PR #2179 / commit e69d536.

Instead of always interpreting an unexplained SIGTRAP as a phantom
breakpoint memorize all possible unreported breakpoint hits and only
act on it when the thread hasn't moved from one.

Also clarifies the behavior of the halt command when it is received
while the target is stopped or in the process of stopping.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -5307,3 +5307,51 @@ func TestWatchpointCounts(t *testing.T) {
 		}
 	})
 }
+
+func TestManualStopWhileStopped(t *testing.T) {
+	// Checks that RequestManualStop sent to a stopped thread does not cause the target process to die.
+	withTestProcess("loopprog", t, func(p *proc.Target, fixture protest.Fixture) {
+		asyncCont := func(done chan struct{}) {
+			defer close(done)
+			err := p.Continue()
+			t.Logf("%v\n", err)
+			if err != nil {
+				panic(err)
+			}
+			for _, th := range p.ThreadList() {
+				if th.Breakpoint().Breakpoint != nil {
+					t.Logf("unexpected stop at breakpoint: %v", th.Breakpoint().Breakpoint)
+					panic("unexpected stop at breakpoint")
+				}
+			}
+		}
+
+		const (
+			repeatsSlow = 3
+			repeatsFast = 5
+		)
+
+		for i := 0; i < repeatsSlow; i++ {
+			t.Logf("Continue %d (slow)", i)
+			done := make(chan struct{})
+			go asyncCont(done)
+			time.Sleep(1 * time.Second)
+			p.RequestManualStop()
+			time.Sleep(1 * time.Second)
+			p.RequestManualStop()
+			time.Sleep(1 * time.Second)
+			<-done
+		}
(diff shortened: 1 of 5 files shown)


aarzilli:
cc @polinasok and this is why the pause/continue stuff crashed the target program. The good news is that this bug is only in the current version of Delve.

service/api/types.go
```diff
@@ -414,6 +414,9 @@ const (
 	// SwitchGoroutine switches the debugger's current thread context to the thread running the specified goroutine
 	SwitchGoroutine = "switchGoroutine"
 	// Halt suspends the process.
+	// The effect of Halt while the target process is stopped, or in the
+	// process of stopping, is operating system and timing dependent. It will
+	// either have no effect or cause the following resume to stop immediately.
```
polinasok, line 419: In that case, the dap pause should be checking is-running and not calling halt if not running, shouldn't it?

pkg/proc/proc_test.go
```diff
@@ -5307,3 +5307,51 @@ func TestWatchpointCounts(t *testing.T) {
...
+
+func TestManualStopWhileStopped(t *testing.T) {
+	// Checks that RequestManualStop sent to a stopped thread does not cause the target process to die.
+	withTestProcess("loopprog", t, func(p *proc.Target, fixture protest.Fixture) {
+		asyncCont := func(done chan struct{}) {
+			defer close(done)
+			err := p.Continue()
+			t.Logf("%v\n", err)
+			if err != nil {
+				panic(err)
```
polinasok, line 5319: Curious, why panic instead of t.Fatal?

polinasok:
@aarzilli And thank you for responding to my finding so quickly! I guess now I can put the offending test case from #2466 back into the mix.

service/api/types.go
aarzilli: Maybe. Doesn't the stop event disable the pause button anyway? The DAP specification doesn't say what the pause request does when the target is stopped anyway.

pkg/proc/proc_test.go
aarzilli: I thought t.Fatal didn't work inside a goroutine. Maybe it was t.Error?

aarzilli:
> @aarzilli And thank you for responding to my finding so quickly! I guess now I can put the offending test case from #2466 back into the mix.

I'd still say no. This test adds a test that basically does that but programmed directly to proc, instead of going through DAP. I'd rather not have tests for complicated logic in proc that go through DAP.

derekparker (approved):
LGTM

derekparker:
> > @aarzilli And thank you for responding to my finding so quickly! I guess now I can put the offending test case from #2466 back into the mix.
> 
> I'd still say no. This test adds a test that basically does that but programmed directly to proc, instead of going through DAP. I'd rather not have tests for complicated logic in proc that go through DAP.

I agree, this is a lower-level issue and I think this is testing it at the appropriate level.

PR #2485, opened 2021-05-15
dwarf/op,proc: fix handling of DW_OP_piece

aarzilli:
According to DWARFv4 section 2.6.1.3 having a DW_OP_piece when nothing
is on the stack is legal and represents uninitialized/unavailable
memory.


aarzilli:
This is from the Go 1.17 support branch, but it's a preexisting bug and that branch is already complicated without this.

derekparker (approved):
LGTM

PR #2486, opened 2021-05-15
proc: fix runtimeTypeToDIE setup

aarzilli:
The code populating runtimeTypeToDIE was incorrectly adding StaticBase
to the offset. We never noticed because on statically compiled
executables StaticBase is always zero and on PIE and plugins the
fallback code took care of the problem anyway.

A change in Go 1.17 broke the fallback code, making the issue apparent.

This commit fixes the setup of runtimeTypeToDIE and disables the
fallback code for Go 1.17 and later.

This change also fixes a rare failure in TestPluginVariables when PIE
is enabled.

### DIFF
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/go-delve/delve/pkg/dwarf/godwarf"
 	"github.com/go-delve/delve/pkg/dwarf/reader"
+	"github.com/go-delve/delve/pkg/goversion"
 )
 
 // The kind field in runtime._type is a reflect.Kind value plus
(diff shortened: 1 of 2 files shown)


aarzilli:
Like #2485 this is split from the Go 1.17 support branch because it's a preexisting bug. No test for this, it will be tested when we start testing Go 1.17 since TestPluginVariables would fail systematically when testing PIE (that test can also fail on < 1.17 but it's very rare).

derekparker (approved):
LGTM

pkg/proc/types.go
```diff
@@ -149,6 +150,10 @@ func runtimeTypeToDIE(_type *Variable, dataAddr uint64) (typ godwarf.Type, kind
 
 	// go1.7 to go1.10 implementation: convert runtime._type structs to type names
 
+	if goversion.ProducerAfterOrEqual(_type.bi.Producer(), 1, 17) {
```
derekparker, line 153: Could you add a comment here explaining why we no longer need this in >= 1.17 before I merge this, please.

derekparker (changes requested):
Changing review from approved to request changes based on my last comment.

aarzilli:
Done.

derekparker (approved):
LGTM

PR #2487, opened 2021-05-15
gdbserial: for the 'g' command on debugserver treat E74 as unsupported

aarzilli:
The maintainer of debugserver says he wants to fix the problem with the
'g' command but doesn't know when it will happen. Treat the error 'E74'
for the 'g' command on debugserver as if the server had returned an
unsupported error so that, for this specific problem, the error doesn't
resurface in the future.


derekparker (approved):
LGTM

Issue #2481, opened 2021-05-14
support breakpoints with hit count conditions


suzmue:
This is a feature request for Hit Condition support.

Similar to conditional breakpoints, users may want to set a breakpoint based on how many times that breakpoint has been hit.

Here is the VS Code UI for setting a hit condition:
![Screen Shot 2021-05-14 at 3 24 05 PM](https://user-images.githubusercontent.com/6634754/118319207-71c31000-b4c8-11eb-8d22-8cb03c763a81.png)

Possible HitConditions to support would be:
- `== expr`
- `>= expr`
- `<= expr`
- `% expr`

This is part of the DAP spec, and could be implemented on the DAP side but I think would be better if this could be supported / implemented where the breakpoint conditions are checked. Hit Conditions would probably need to be binary expressions where the TotalHitCount can be substituted in for `expr.X`

derekparker:
@suzmue is there anything more to do here with https://github.com/go-delve/delve/pull/2490 merged or can I close this out?

suzmue:
This can be closed! @aarzilli  added support for hit count conditions in the terminal package with #2518

PR #2490, opened 2021-05-17
pkg/proc: add support for hit count condition breakpoints

suzmue:
Similar to conditional breakpoints, users may want to set a breakpoint based on how many times a breakpoint has been hit. This change adds support for this feature to the proc layer by checking the hit condition after updating the hit count. The hit condition is implemented as a binary expression where the expr.X is replaced with the TotalHitCount for the breakpoint.

Updates #2481


pkg/proc/proc_test.go
```diff
@@ -1685,6 +1685,141 @@ func TestCondBreakpointError(t *testing.T) {
...
+func TestHitCondBreakpointEQ(t *testing.T) {
+	withTestProcess("break", t, func(p *proc.Target, fixture protest.Fixture) {
+		bp := setFileBreakpoint(p, t, fixture.Source, 6)
+		bp.HitCond = &ast.BinaryExpr{
+			Op: token.EQL,
+			X:  &ast.Ident{Name: "placeHolder"},
+			Y:  &ast.BasicLit{Kind: token.INT, Value: "3"},
+		}
+
+		err := p.Continue()
```
aarzilli, line 1697: `assertNoError(p.Continue(), ...)`

pkg/proc/breakpoints.go
```diff
@@ -195,6 +216,15 @@ func (bp *Breakpoint) CheckCondition(thread Thread) BreakpointState {
 	return bpstate
 }
 
+// checkHitCond evaluates bp's hit condition on thread.
+func (bp *Breakpoint) checkHitCond(bpstate BreakpointState, thread Thread) BreakpointState {
+	if bp.HitCond != nil {
+		bp.HitCond.X = astutil.Int(int64(bp.TotalHitCount))
+		bpstate.Active, bpstate.HitCondError = evalBreakpointCondition(thread, bp.HitCond)
```
aarzilli, line 223: Reuse `bpstate.CondError`, only set it if it's already non-nil.

pkg/proc/breakpoints.go
```diff
@@ -195,6 +216,15 @@ func (bp *Breakpoint) CheckCondition(thread Thread) BreakpointState {
 	return bpstate
 }
 
+// checkHitCond evaluates bp's hit condition on thread.
+func (bp *Breakpoint) checkHitCond(bpstate BreakpointState, thread Thread) BreakpointState {
+	if bp.HitCond != nil {
```
aarzilli, line 221: Either move this if to the caller or move all of the ifs around the call here. If you move them here invert the condition and return early.

pkg/proc/breakpoints.go
```diff
@@ -163,6 +168,22 @@ type returnBreakpointInfo struct {
...
+		bpstate.TotalHitCount++
+	}
+	//Check hit condition if this is a user breakpoint.
+	if bpstate.Active && !bpstate.Internal {
+		bpstate = bp.checkHitCond(bpstate, thread)
+	}
+	return bpstate
+}
+
+func (bp *Breakpoint) checkCond(bpstate BreakpointState, thread Thread) BreakpointState {
```
aarzilli, line 186: I think the correct receiver of this function and of checkHitCond is `*BreakpointState` and then they don't return anything.

pkg/proc/proc_test.go
```diff
@@ -1685,6 +1685,141 @@ func TestCondBreakpointError(t *testing.T) {
...
+			t.Fatalf("Unexpected error on first Continue(): %v", err)
+		}
+
+		bp.HitCond = &ast.BinaryExpr{
+			Op: token.EQL,
+			X:  &ast.Ident{Name: "placeHolder"},
+			Y:  &ast.BasicLit{Kind: token.INT, Value: "3"},
+		}
+
+		err = p.Continue()
```
aarzilli, line 1743: `assertNoError(p.Continue(), ...`

aarzilli (commented):
If you also want to propagate this through service/dap you can do it in the same PR.

pkg/proc/breakpoints.go
suzmue: Done

pkg/proc/breakpoints.go
suzmue: Done

pkg/proc/breakpoints.go
suzmue: Done

pkg/proc/proc_test.go
suzmue: Done

suzmue:
@aarzilli I propagated this through to `service/dap` but it required some changes to how we set breakpoints to do this correctly (using amend instead of clear/create). I can split this back out into a couple of PRs if that will make it easier to review.

pkg/proc/breakpoints.go
```diff
@@ -163,36 +172,76 @@ type returnBreakpointInfo struct {
...
+	// Evaluate the breakpoint condition and set CondError if it has not
+	// already been set.
+	var err error
+	bpstate.Active, err = evalBreakpointCondition(thread, bpstate.HitCond)
+	if bpstate.CondError == nil {
+		bpstate.CondError = err
+	}
+}
+
+func (bpstate *BreakpointState) setHitCondTotalHitCount() {
```
aarzilli, line 234: Is this to support things like `(hitcount >= 9) && (hitcount <= 15)`? It won't work because it will only replace the leftmost variable.

If this really needs arbitrary expressions involving the hitcount it should be a fake variable, like runtime.curg (see the ast.SelectorExpr in evalAST).

If it just needs to compare the hit count with an integer then maybe it doesn't need to be a BinaryExpr at all, just an operator ("<", ">=", etc.) and an integer argument. And it gets evaluated in a `switch` statement, without involving an EvalScope.

service/dap/server.go
```diff
@@ -1011,40 +1011,228 @@ func (s *Server) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
 	// If a breakpoint:
 	// -- exists and not in request => ClearBreakpoint
 	// -- exists and in request => AmendBreakpoint
 	// -- doesn't exist and in request => SetBreakpoint
 
-	// Clear all existing breakpoints in the file that are not function breakpoints.
-	if err := s.clearMatchingBreakpoints(func(bp *api.Breakpoint) bool { return bp.File == serverPath && bp.Name == "" }); err != nil {
+	// Get all existing breakpoints that match for this source.
+	sourceRequestPrefix := fmt.Sprintf("sourceBp Path=%s", request.Arguments.Source.Path)
+	existingBps := s.getMatchingBreakpoints(sourceRequestPrefix)
```
aarzilli, line 1025: I liked the old way of finding breakpoints better. This doesn't even work if a file's path happens to be the prefix of another file's path.

service/dap/server.go
```diff
@@ -1011,40 +1011,228 @@ func (s *Server) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
+	bpAdded := make(map[string]struct{}, len(existingBps))
+
+	// Amend existing breakpoints.
+	breakpoints := make([]dap.Breakpoint, len(request.Arguments.Breakpoints))
+	for i, want := range request.Arguments.Breakpoints {
+		reqString := fmt.Sprintf("%s Line=%d Column=%d", sourceRequestPrefix, want.Line, want.Column)
+		var err error
+		got, ok := existingBps[reqString]
+		if !ok {
+			// Skip if the breakpoint does not already exist.
```
aarzilli, line 1035: What kind of conflict would this be?

service/debugger/debugger.go
```diff
@@ -781,6 +785,28 @@ func copyBreakpointInfo(bp *proc.Breakpoint, requested *api.Breakpoint) (err err
 	if requested.Cond != "" {
 		bp.Cond, err = parser.ParseExpr(requested.Cond)
 	}
+	bp.HitCond = nil
+	if requested.HitCond != "" {
+		// A hit condition can be in the following formats:
```
aarzilli, line 790: This will need to be changed depending on what we decide to support in HitCond.

service/dap/server.go
suzmue: That makes sense. I can change the finding breakpoint logic back (or can use %q and add a " Line=" after Path which I think should resolve the Prefix issue

One thing I was thinking though for a method like this, is that we are guaranteed to only get breakpoints set by the DAP server, so if we eventually support parallel clients the DAP client won't clear all breakpoints set by an RPC user.

pkg/proc/breakpoints.go
```diff
@@ -70,10 +72,17 @@ type Breakpoint struct {
...
+	// true with the TotalHitCount.
+	HitCond *ast.BinaryExpr
 
 	// ReturnInfo describes how to collect return variables when this
 	// breakpoint is hit as a return breakpoint.
 	returnInfo *returnBreakpointInfo
+
+	// RequestString is a string used by the debug adapter to
+	// identify a breakpoint with a particular request.
+	RequestString string
```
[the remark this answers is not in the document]
suzmue: The name field has stricter rules about what characters are allowed. So to include the file path, we would have to sanitize the string somehow.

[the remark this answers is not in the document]
aarzilli: It would be reasonable to move those validity checks for the name out of debugger and into rpc2/server.go and rpc1/server.go

pkg/proc/breakpoints.go
suzmue: This was to support things like (`hitcount % 100 == 0`), though this could also be supported as `OP NUMBER`

suzmue: I'll update this logic to accept `NUMBER` or `OP NUMBER`! I think that makes more sense for now. 

If there is a use case to support other expressions we can change this at a later point after getting user feedback.

pkg/proc/breakpoints.go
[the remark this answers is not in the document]
suzmue: Ok sounds good. Will do that.

service/api/types.go
```diff
@@ -67,6 +67,8 @@ type Breakpoint struct {
 
 	// Breakpoint condition
 	Cond string
+	// Breakpoint hit count condition
+	HitCond string
```
aarzilli, line 71: Document the format.

service/dap/server.go
aarzilli: Ok.

service/api/types.go
suzmue: Done.

aarzilli (approved):
LGTM

service/dap/daptest/client.go
```diff
@@ -247,6 +247,27 @@ func (c *Client) SetConditionalBreakpointsRequest(file string, lines []int, cond
...
+// SetBreakpointsRequest sends a 'setBreakpoints' request with conditions.
+func (c *Client) SetHitConditionalBreakpointsRequest(file string, lines []int, conditions map[int]string) {
+	request := &dap.SetBreakpointsRequest{Request: *c.newRequest("setBreakpoints")}
+	request.Arguments = dap.SetBreakpointsArguments{
+		Source: dap.Source{
+			Name: filepath.Base(file),
+			Path: file,
+		},
+		Breakpoints: make([]dap.SourceBreakpoint, len(lines)),
+		//sourceModified: false,
```
derekparker, line 259: Please remove commented out code.

service/dap/server.go
derekparker: @aarzilli I assume by your LGTM this has been resolved?

pkg/proc/breakpoints.go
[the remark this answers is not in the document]
derekparker: Seems odd and not very intuitive to have different validity methods for the name field just depending on which API you're using. Maybe we relax the name check so that this new format is allowed as well while still validating the input?

service/dap/server.go
aarzilli: Yes.

derekparker:
CI failure:

```
make test
# github.com/go-delve/delve/service/dap
vet: service/dap/server_test.go:2387:6: undeclared name: handleStop
*** Error code 2
```

aarzilli:
I think this needs a rebase.

derekparker:
Re-running CI.

PR #2494, opened 2021-05-18
dap: add sameuser check

hyangah:
On linux, delve RPC server allows only connections from the same user
if --only-same-user is set (true, by default). Do the same for DAP
server.

Moved the sameuser check logic to service/internal/sameuser.
Considered importing service/rpccommon from the dap server,
but when we eventually migrate to multiplex rpc and dap from one
port, I am afraid that can cause cyclic imports.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -452,6 +452,7 @@ func dapCmd(cmd *cobra.Command, args []string) {
 				CheckGoVersion:       checkGoVersion,
 				TTY:                  tty,
 			},
+			CheckLocalConnUser: checkLocalConnUser,
 		})
 		defer server.Stop()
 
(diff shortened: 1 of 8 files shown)


hyangah:
cc @polinasok @suzmue 

The test failures are again "Detected possible usage of Logger for user defined data  GO-S0904" complaining about including the connection error message in the log.

aarzilli:
> The test failures are again "Detected possible usage of Logger for user defined data GO-S0904" complaining about including the connection error message in the log.

Yet again, betrayed by linters.

aarzilli (approved):

derekparker (approved):
LGTM

derekparker:
> > The test failures are again "Detected possible usage of Logger for user defined data GO-S0904" complaining about including the connection error message in the log.
> 
> Yet again, betrayed by linters.

🤣

PR #2498, opened 2021-05-19
gdbserial,regnum: support ZMM registers

### LINKED ISSUE #2479: No SSE/AVX/AVX-512 registers visible on darwin/amd64
v1.6 / latest master

v1.6.3

MacOS 10.15.7  darwin/amd64

4. What did you do?
Tried to inspect XMMn / YMMn / ZMMn / Kn register values using `regs -a` and the new 
`print <reg>` functionality (PR: https://github.com/go-delve/delve/pull/2446)

The current value(s) of these vector/mask registers. 

6. What did you see instead?

```
(dlv) print %x RAX
c0001ec000
(dlv) print %x XMM0
Command failed: could not find symbol value for XMM0
```

And

```
(dlv) regs -a
   Rip = 0x000000000116eff5
   Rsp = 0x000000c0000c8d10
   Rax = 0x000000c0001ec000
   Rbx = 0x0000000000000000
   Rcx = 0x0000000000000000
   Rdx = 0x000000000085c5bc
   Rsi = 0x00000000000000c0
   Rdi = 0x000000c0001ec0c0
   Rbp = 0x000000c0000c8ec8
    R8 = 0x00000000000000b0
    R9 = 0x0000000000000016
   R10 = 0x0000000000000000
   R11 = 0x0000000000000000
   R12 = 0x000000000880f606
   R13 = 0x0000000000000000
   R14 = 0x00000000000000c0
   R15 = 0x000000000000000c
 ST(0) = 0x7fff0000000000000000	-Inf
 ST(1) = 0x00000000000000000000	0
 ST(2) = 0x7fff0000000000000000	-Inf
 ST(3) = 0x7fff00000000fe9d66b0	-Inf
 ST(4) = 0x7fff0000000000000000	-Inf
 ST(5) = 0x7fff0000000000000000	-Inf
 ST(6) = 0x7fff00000000fffff85c	-Inf
 ST(7) = 0x7fff00000000ffffe822	-Inf
Rflags = 0x0000000000000246	[PF ZF IF IOPL=0]
    Cs = 0x000000000000002b
    Fs = 0x0000000000000000
    Gs = 0x0000000000000000
 MXCSR = 0x00001fa0	[RZ/RN=0 PM UM OM ZM DM IM PE]
```

aarzilli:
Fixes #2479


aarzilli:
@vsivsi PTAL.

vsivsi:
This looks great, I'd made essentially the same set of changes on an experiemental branch to generate in info I posted to #2479.

One issue I'm still seeing when testing this on a machine with AVX-512 is the integration tests don't pass yet:

```
--- FAIL: TestClientServer_FpRegisters (0.36s)
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/support.go:252: enabling recording for github.com/go-delve/delve/service/test_test.TestClientServer_FpRegisters
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1538: state after continue: &api.DebuggerState{Running:false, Recording:false, CoreDumping:false, CurrentThread:(*api.Thread)(0xc0001c6230), SelectedGoroutine:(*api.Goroutine)(0xc000374000), Threads:[]*api.Thread{(*api.Thread)(0xc0001c62a0), (*api.Thread)(0xc0001c6310), (*api.Thread)(0xc0001c6460), (*api.Thread)(0xc0001c6540), (*api.Thread)(0xc0001c65b0)}, NextInProgress:false, Exited:false, ExitStatus:0, When:"", Err:error(nil)}
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1547: avx2 variable: &api.Variable{Name:"avx2", Addr:0xc000112f36, OnlyAddr:false, Type:"bool", RealType:"bool", Flags:0x0, Kind:0x1, Value:"true", Len:0, Cap:0, Children:[]api.Variable{}, Base:0x0, Unreadable:"", LocationExpr:"[block] DW_OP_fbreg -0x52 ", DeclLine:22}
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1547: avx512 variable: &api.Variable{Name:"avx512", Addr:0xc000112f35, OnlyAddr:false, Type:"bool", RealType:"bool", Flags:0x0, Kind:0x1, Value:"true", Len:0, Cap:0, Children:[]api.Variable{}, Base:0x0, Unreadable:"", LocationExpr:"[block] DW_OP_fbreg -0x53 ", DeclLine:23}
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1580: register XMM1 expected "0x3ff66666666666663ff4cccccccccccd" got "0x3fe666663fd9999a3fcccccd3fc00000\tv2_int={ 3fcccccd3fc00000 3fe666663fd9999a }\tv4_int={ 3fc00000 3fcccccd 3fd9999a 3fe66666 }\tv8_int={ 0000 3fc0 cccd 3fcc 999a 3fd9 6666 3fe6 }\tv16_int={ 00 00 c0 3f cd cc cc 3f 9a 99 d9 3f 66 66 e6 3f }\tv2_float={ 0.22500005352776498 0.6999999281950295 }\tv4_float={ 1.5 1.6 1.7 1.8 }\n\t[YMM1h] 0x00000000000000000000000000000000\tv2_int={ 0000000000000000 0000000000000000 }\tv4_int={ 00000000 00000000 00000000 00000000 }\tv8_int={ 0000 0000 0000 0000 0000 0000 0000 0000 }\tv16_int={ 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 }\tv2_float={ 0 0 }\tv4_float={ 0 0 0 0 }\n\t[ZMM1hl] 0x3ff199999999999a3ff3333333333333\tv2_int={ 3ff3333333333333 3ff199999999999a }\tv4_int={ 33333333 3ff33333 9999999a 3ff19999 }\tv8_int={ 3333 3333 3333 3ff3 999a 9999 9999 3ff1 }\tv16_int={ 33 33 33 33 33 33 f3 3f 9a 99 99 99 99 99 f1 3f }\tv2_float={ 1.2 1.1 }\tv4_float={ 4.172325e-08 1.9 -1.5881868e-23 1.8874999 }\n\t[ZMM1hh] 0x00000000000000000000000000000000\tv2_int={ 0000000000000000 0000000000000000 }\tv4_int={ 00000000 00000000 00000000 00000000 }\tv8_int={ 0000 0000 0000 0000 0000 0000 0000 0000 }\tv16_int={ 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 }\tv2_float={ 0 0 }\tv4_float={ 0 0 0 0 }"
FAIL
FAIL	github.com/go-delve/delve/service/test	0.608s
FAIL
```

vsivsi:
Another suggestion: Please update the comment at the top of `amd64.go` here:

https://github.com/aarzilli/delve/blob/fix2479/pkg/dwarf/regnum/amd64.go#L8-L11

To point to some more current/authoritative version of this spec.

The official repo for the spec is here:
https://github.com/aarzilli/delve/blob/fix2479/pkg/dwarf/regnum/amd64.go#L8-L11

Unfortunately, it doesn't seem to host a PDF version of the document, but it is easy enough to close this repo and build one, or to just read the Latex source.

https://gitlab.com/x86-psABIs/x86-64-ABI/-/tree/master

Maybe the comment should read:

```golang
// The mapping between hardware registers and DWARF registers below is specified
// in the System V ABI AMD64 Architecture Processor Supplement, version 1.0,
// page 61, figure 3.36
// Source: https://gitlab.com/x86-psABIs/x86-64-ABI/-/tree/master
```

It's not ideal that there doesn't appear to be an official source for the PDF, but I think it's better 
to point to the official repo, than some random PDF on the internet that can change or disappear
without a trace.

aarzilli:
Changed the comment. The test failure could be related or could be unrelated. Can you post the output of `go test -run TestClientServer_FpRegisters -log-output=gdbwire` after changing the `gdbWireMaxLen` constant in gdbserver.go to something very large (like 10000, hopefully that's enough).

vsivsi:
@aarzilli   Here is the test output you requested above:

https://gist.github.com/vsivsi/6b61c1d45f1cb2a3da408a7732f66183

vsivsi:
Also, might I suggest increasing test coverage when `avx512 == true` to XMMn for n > 16 and the Kn mask registers?  That is, thread state that is not present in AVX and is newly enabled by this PR.

vsivsi:
Okay, now I've been trying to actually use this new functionality in some AVX-512 real code, and the vector register state output does not seem to be updating correctly. Other registers (including the AVX512 Kn masks) update correctly, but the vectors do not.

```
(dlv) s
> github.com/vsivsi/adder.varLenWrite() ./add_amd64.s:38 (PC: 0x11a6e89)
    33:		MOVQ         in_len+8(FP), DX
    34:		CMPQ         DX, $0x00
    35:		JE           fail
    36:		INCB         AL
    37:		MOVQ         out+24(FP), DX
=>  38:		VPXORQ       Z4, Z4, Z4          // This zeros all bits in Z4
    39:		VPXORQ       Z1, Z1, Z1
    40:		VMOVDQU32    og0<>+0(SB), Y1
    41:		KXNORB       K1, K1, K1
    42:		VPGATHERDQ   (DX)(Y1*1), K1, Z0
    43:		VPCMPUQ      $0x00, Z0, Z4, K1
(dlv) print XMM4
"00200100c000000080070000c00000000000000000000000000000000000000028e00000c000000060210100c000000000000000000000000000000000000000"
(dlv) print XMM4.uint32
[16]void [73728,192,1920,192,0,0,0,0,57384,192,74080,192,0,0,0,0]
(dlv) s
> github.com/vsivsi/adder.varLenWrite() ./add_amd64.s:39 (PC: 0x11a6e8f)
    34:		CMPQ         DX, $0x00
    35:		JE           fail
    36:		INCB         AL
    37:		MOVQ         out+24(FP), DX
    38:		VPXORQ       Z4, Z4, Z4           // This zeros all bits in Z4
=>  39:		VPXORQ       Z1, Z1, Z1
    40:		VMOVDQU32    og0<>+0(SB), Y1
    41:		KXNORB       K1, K1, K1
    42:		VPGATHERDQ   (DX)(Y1*1), K1, Z0
    43:		VPCMPUQ      $0x00, Z0, Z4, K1
    44:		KTESTB       K1, K1
(dlv) print XMM4
"00200100c000000080070000c00000000000000000000000000000000000000028e00000c000000060210100c000000000000000000000000000000000000000"
(dlv) print XMM4.uint32
[16]void [73728,192,1920,192,0,0,0,0,57384,192,74080,192,0,0,0,0]
(dlv) regs -a

<snip>

  XMM4 = 0x000000c000000780000000c000012000	v2_int={ 000000c000012000 000000c000000780 }	v4_int={ 00012000 000000c0 00000780 000000c0 }	v8_int={ 2000 0001 00c0 0000 0780 0000 00c0 0000 }	v16_int={ 00 20 01 00 c0 00 00 00 80 07 00 00 c0 00 00 00 }	v2_float={ 4.07423228292e-312 4.07423192814e-312 }	v4_float={ 1.03315e-40 2.69e-43 2.69e-42 2.69e-43 }
	[YMM4h] 0x00000000000000000000000000000000	v2_int={ 0000000000000000 0000000000000000 }	v4_int={ 00000000 00000000 00000000 00000000 }	v8_int={ 0000 0000 0000 0000 0000 0000 0000 0000 }	v16_int={ 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 }	v2_float={ 0 0 }	v4_float={ 0 0 0 0 }
	[ZMM4hl] 0x000000c000012160000000c00000e028	v2_int={ 000000c00000e028 000000c000012160 }	v4_int={ 0000e028 000000c0 00012160 000000c0 }	v8_int={ e028 0000 00c0 0000 2160 0001 00c0 0000 }	v16_int={ 28 e0 00 00 c0 00 00 00 60 21 01 00 c0 00 00 00 }	v2_float={ 4.07423220217e-312 4.074232284657e-312 }	v4_float={ 8.0412e-41 2.69e-43 1.03808e-40 2.69e-43 }
	[ZMM4hh] 0x00000000000000000000000000000000	v2_int={ 0000000000000000 0000000000000000 }	v4_int={ 00000000 00000000 00000000 00000000 }	v8_int={ 0000 0000 0000 0000 0000 0000 0000 0000 }	v16_int={ 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 }	v2_float={ 0 0 }	v4_float={ 0 0 0 0 }

<snip>

(dlv) 
```

vsivsi:
I haven't gotten to the bottom of the issues above yet, but it searching for trouble spots, I did find this `SetReg` code, which will certainly need updating.  

https://github.com/aarzilli/delve/blob/fix2479/pkg/proc/gdbserial/gdbserver.go#L1861-L1869

Should probably be something like:

```golang
	gdbreg, ok := t.regs.regs[regName]
	if !ok && strings.HasPrefix(regName, "xmm") {
		// XMMn and YMMn are the same amd64 register (in different sizes), if we
		// don't find XMMn try YMMn instead.
		gdbreg, ok = t.regs.regs["y"+regName[1:]]
		// If still not okay, then try ZMMn...
		if !ok {
			gdbreg, ok = t.regs.regs["z"+regName[1:]]
		}
	}
	if !ok {
		return fmt.Errorf("could not set register %s: not found", regName)
	}
```

aarzilli:
It does look like the output of the test matches what debugserver returns so that's "fine". I've changed the test to be more resilient towards the runtime overwriting our registers before we read them. I don't know what's causing your problem, I think it's fastest if you look for the cause yourself. I'll explain how this works.

We are communicating with a thing called debugserver, when you pass `-log-output=gdbwire` you can see the messages we exchange with it. Early on after connecting we send a bunch of `qRegisterInfo` requests that tell us what registers there are, how big they are (bitsize) and where they are (offset). The value of registers is retrieved using the `g` command, for example:

```
2021-05-20T11:29:05-07:00 debug layer=gdbconn <- $g;thread:6fc259;#2e
2021-05-20T11:29:05-07:00 debug layer=gdbconn -> $0102030405060708000000000000000000000000000000000000000000000....
```

here we are requesting the registers of thread `6fc259`. The response is hex encoded, two characters per byte, if the register at offset zero was 8 bytes long its value would be: `0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08`, with `0x01` being the least significant byte.

> Also, might I suggest increasing test coverage when avx512 == true to XMMn for n > 16 and the Kn mask registers? That is, thread state that is not present in AVX2 and is newly enabled by this PR.

I don't have AVX512 hardware so I won't be doing that but I wouldn't object to other people doing it.

aarzilli:
One thing that I wonder is that LLDB does not use the 'g' command, ever. There is another command to read registers (p) that it uses instead. The 'g' command has been broken in the past, maybe this is another case of 'g' being broken. You can test this by setting gcmdok to false in gdbserver.go

vsivsi:
Thanks for that explainer on how all of this works. I had inferred some it is from stepping through the code, but the overview helped greatly!

I think I've figured out what's happening, explains both the integration test failures I'm seeing and the apparent failure of ZMM registers to update while debugging interactively. 

**This bug scrambles the state of (X/Y/Z)0-15 when AVX512 is present. The only vector registers that properly reflect the thread state are XMM0, YMM0 and (X/Y/Z)16-31. All others are wrong.  That is, you don't need to be using AVX512 features to see incorrect results. Simply using SSE or AVX registers will produce incorrect results, as reflected in the failing integration tests.** 

So here's what I see happening on machines with AVX512:

The register state buffer being transmitted by `debugserver` is handling registers (X/Y/Z)MM0-15 incorrectly.  Specifically, it is not properly interleaving the so called `ZMM_Hi256 state` from the XSAVE extended region correctly. This state represents the upper 256 bits of ZMM0-ZMM15, stored as a 16 x 256 bit block. 

Note: ZMM16-ZMM31 are stored in a contiguous 16 x 512 bit block called `Hi16_ZMM state`, and this state does appear to be handled correctly.

The XMM0-XMM15 register state is also being advertised as a 16 x 512 bit block by `debugserver`:

```
XMM0 [ xmm0 128bits ] [ ymm0h 128 bits ] [ zmm0hl 128 bits ] [ zmm0hh 128 bits ] 
XMM1 [ xmm1 128bits ] [ ymm1h 128 bits ] [ zmm1hl 128 bits ] [ zmm1hh 128 bits ] 
... 
XMM15 [ xmm15 128bits ] [ ymm15h 128 bits ] [ zmm15hl 128 bits ] [ zmm15hh 128 bits ] 
```

But what I'm actually seeing is:

```
XMM0 [ xmm0 128bits ] [ ymm0h 128 bits ] [ xmm1 128 bits ] [ ymm1h 128 bits ] 
XMM1 [ xmm2 128bits ] [ ymm2h 128 bits ] [ xmm3 128 bits ] [ ymm3h 128 bits ] 
... 
XMM8 [ zmm0hl 128bits ] [ zmm0hh 128 bits ] [ zmm1hl 128bits ] [ zmm1hh 128 bits ] 
XMM9 [ zmm0hl 128bits ] [ zmm0hh 128 bits ] [ zmm1hl 128bits ] [ zmm1hh 128 bits ] 
...
XMM15 [ zmm14hl 128bits ] [ zmm14hh 128 bits ] [ zmm15hl 128bits ] [ zmm15hh 128 bits ] 
```

That is... rather than properly interleaving the `ZMM_Hi256 state` into the (already interleaved) SSE and AVX state, 
the debugger is just concatenating it onto the end and adjusting the offsets of XMM0-XMM15 to be 64 bytes each.

You can clearly see this in the failing integration test I cited upthread. XMM1 is failing because it actually has XMM2s value.

It would be easy enough to create a special case somewhere in delve to properly interleave the underlying buffer state to reflect the advertised offsets, and I've seen a fair amount of special cases in the code fixing up various upstream debugger issues, but I'm unsure what the best way to handle this is. It seems likely that some future update of `debugserver` will fix this pretty glaring error, and then what?  (or is this the specified behavior and delve is ignoring it? Idk.)

vsivsi:
The following code inserted into the `reloadRegisters` method defined in `gdbserver.go` between these lines:

https://github.com/aarzilli/delve/blob/fix2479/pkg/proc/gdbserial/gdbserver.go#L1582-L1583

```golang
	case "darwin":
		// For AVX512 on Darwin, need to interleave the ZMM_Hi256 state
		// (the upper 256 bits of ZMM0-ZMM15) into the AVX state (the lower)
		// 256 bits of those same registers.
		// The AVX State is 16 x 32 bytes, and so is the ZMM_Hi256 state.
		// This code assumes that these two buffers are packed contiguously
		// and interleaves them by taking the first 32 bytes of AVX state,
		// followed by the first 32 bytes of ZMM_Hi256, the next 32 bytes of
		// AVX, the next 32 of ZMM_Hi256, and so on.
		regInfoZmm0 := t.regs.regsInfo[47]
		if regInfoZmm0.Name == "zmm0" && regInfoZmm0.Bitsize == 512 {
			// The register values are replaced with slices into a new
			// buffer because the original buffer is used to *overwrite*
			// registers, so modifying that buffer leads to incorrect
			// values being written back. That is, the read and write
			// are symmetrical in assuming that the ZMM_Hi256 state is
			// not interleaved with the AVX state, but rather is
			// concatenated to it.
			dstBuf := make([]byte, 16*64)
			startOffset := regInfoZmm0.Offset
			srcBuf := t.regs.buf[startOffset : startOffset+16*64]
			for i := uint64(0); i < 16; i++ {
				copy(dstBuf[i*64:i*64+32], srcBuf[i*32:i*32+32])
				copy(dstBuf[i*64+32:i*64+64], srcBuf[16*32+i*32:16*32+i*32+32])
				regName := "zmm" + strconv.FormatUint(i, 10)
				regEntry := t.regs.regs[regName]
				regEntry.value = dstBuf[i*64 : i*64+64]
				t.regs.regs[regName] = regEntry
			}
		}
```

This solves the problem I am seeing on AVX512 hardware. It is admittedly a brittle hack (because it will likely break in the future with some update of `debugserver`, but it demonstrates that the unexpected behavior described above is indeed what is happening: The advertised register offsets + bit widths do not match the underlying buffer state for zmm0-zmm15.  

Curiously, `t.p.conn.writeRegisters` symmetrically assumes that the buffer is in this same configuration, so it is not possible to "fix up" the state in `regs.buf[]` directly, because that will lead to incorrect register values being written back! 

So the code above replaces the `value` slices for zmm0-zmm15 so they don't shadow the underlying buffer.

This almost certainly means that more code will need to be added someplace to "de-interleave" these values back into the main buffer when register values modified by delve.

aarzilli:
If we did something like this it would have to be tied to one specific version of debugserver. Two questions: have you reported this to apple? did you try to disable gcmdok and does this still happen?

vsivsi:
Hi, no I have not done either of those things. I had limited time over the weekend and my first priority has been to just get something working reliably for myself without going too far down the "debugging the debugger" rabbit hole!

Unfortunately, once I got the above "fix" implemented, I almost immediately hit upon the next issue, which is that when single-stepping through a Go assembly ".s" source file, Delve completely clobbers the AVX512 Kn mask register state.
This obviously modifies program behavior, making it impossible to reliably use. I haven't yet had time to try to figure out how/where this is happening.

For example, this very simple assembly function returns `0xff` when run normally, but incorrectly returns `0x00` when single stepping through the code:

```
// func maskTest() byte
// Requires: AVX512DQ
TEXT ·maskTest(SB), NOSPLIT, $0-1
	KXNORB K1, K1, K1
	KMOVB  K1, AX
	MOVB   AL, ret+0(FP)
	RET
```

This is what I see happening while single-stepping:

```
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:10 (PC: 0x11a6f40)
     5:	#include "textflag.h"
     6:	
     7:	// func maskTest() byte
     8:	// Requires: AVX512DQ
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
=>  10:		KXNORB K1, K1, K1
    11:		KMOVB  K1, AX
    12:		MOVB   AL, ret+0(FP)
    13:		RET
    14:	
    15:	// func add(x uint64, y uint64) uint64
(dlv) print %x K1
0
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:11 (PC: 0x11a6f44)
     6:	
     7:	// func maskTest() byte
     8:	// Requires: AVX512DQ
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
    10:		KXNORB K1, K1, K1
=>  11:		KMOVB  K1, AX
    12:		MOVB   AL, ret+0(FP)
    13:		RET
    14:	
    15:	// func add(x uint64, y uint64) uint64
    16:	TEXT ·add(SB), NOSPLIT, $0-24
(dlv) print %x K1
ff
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:12 (PC: 0x11a6f48)
     7:	// func maskTest() byte
     8:	// Requires: AVX512DQ
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
    10:		KXNORB K1, K1, K1
    11:		KMOVB  K1, AX
=>  12:		MOVB   AL, ret+0(FP)
    13:		RET
    14:	
    15:	// func add(x uint64, y uint64) uint64
    16:	TEXT ·add(SB), NOSPLIT, $0-24
    17:		MOVQ x+0(FP), AX
(dlv) print %x K1
0
(dlv) print %x RAX
0
```

That is: when restarting at line 11, Delve is overwriting the K1 mask with `0x00`, and this is a true thread state corruption, no just a display error, because register `AL` gets the incorrect value and it is subsequently written and returned on the stack!

vsivsi:
Curiously, this test code does not show the mask register corruption I observe when using Delve interactively.  A clue!

```
// line 113 of fputest_amd.s below
	KXNORQ K2, K2, K2
	KXNORD K3, K3, K3
	KXNORW K4, K4, K4
	KXNORB K5, K5, K5
```

```golang
func TestMaskRegSteppingBug(t *testing.T) {
	if runtime.GOARCH != "amd64" {
		t.Skip("test is valid only on AMD64")
	}

	protest.AllowRecording(t)
	withTestClient2Extended("fputest/", t, 0, [3]string{}, func(c service.Client, fixture protest.Fixture) {
		if testBackend == "rr" {
			_, err := c.CreateBreakpoint(&api.Breakpoint{File: filepath.Join(fixture.BuildDir, "fputest.go"), Line: 25})
			assertNoError(err, t, "CreateBreakpoint()")
		}

		_, err := c.CreateBreakpoint(&api.Breakpoint{File: filepath.Join(fixture.BuildDir, "fputest_amd64.s"), Line: 113})
		assertNoError(err, t, "CreateBreakpoint()")

		<-c.Continue()

		regs, err := c.ListThreadRegisters(0, true)
		assertNoError(err, t, "ListThreadRegisters()")
		for i := 0; i < 5; i++ {
			t.Logf("\nK2: %s, K3: %s, K4: %s, K5 %s\n", regs[64].Value, regs[65].Value, regs[66].Value, regs[67].Value)
			_, err = c.Step() // c.StepInstruction() makes no difference?
			assertNoError(err, t, "Step()")
			regs, err = c.ListThreadRegisters(0, true)
			assertNoError(err, t, "ListThreadRegisters()")
		}
		t.Fail()
	})
}
```

Outputs:

```
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/support.go:252: enabling recording for github.com/go-delve/delve/service/test_test.TestMaskRegSteppingBug
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1514: 
        K2: 0x0000000000000000, K3: 0x0000000000000000, K4: 0x0000000000000000, K5 0x0000000000000000
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1514: 
        K2: 0xffffffffffffffff, K3: 0x0000000000000000, K4: 0x0000000000000000, K5 0x0000000000000000
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1514: 
        K2: 0xffffffffffffffff, K3: 0x00000000ffffffff, K4: 0x0000000000000000, K5 0x0000000000000000
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1514: 
        K2: 0xffffffffffffffff, K3: 0x00000000ffffffff, K4: 0x000000000000ffff, K5 0x0000000000000000
    /Users/vsi/go/src/github.com/go-delve/delve/service/test/integration2_test.go:1514: 
        K2: 0xffffffffffffffff, K3: 0x00000000ffffffff, K4: 0x000000000000ffff, K5 0x00000000000000ff
```

Great!  So I guess I don't understand what is likely to be different between stepping in the integration test harness like this, versus interactively using the Delve via the `dlv` command line utility. And yes, I've rebuilt `dlv` with the codebase the test code is running in.

Any guidance on what else to look for here would be helpful!

vsivsi:
Another clue is the mask register clobbering only seems to happen before executing an AVX512 instruction (even one that doesn't have anything to do with a specific mask register):

```
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:10 (PC: 0x11a6f40)
     5:	#include "textflag.h"
     6:	
     7:	// func maskTest() byte
     8:	// Requires: AVX512BW, AVX512DQ, AVX512F
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
=>  10:		KXNORB K1, K1, K1
    11:		XORQ   AX, AX
    12:		VPXORQ Z0, Z0, Z0
    13:		KMOVB  K1, AX
    14:		KXNORQ K2, K2, K2
    15:		KXNORD K3, K3, K3
0: K1 = 0
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:11 (PC: 0x11a6f44)
     6:	
     7:	// func maskTest() byte
     8:	// Requires: AVX512BW, AVX512DQ, AVX512F
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
    10:		KXNORB K1, K1, K1
=>  11:		XORQ   AX, AX
    12:		VPXORQ Z0, Z0, Z0
    13:		KMOVB  K1, AX
    14:		KXNORQ K2, K2, K2
    15:		KXNORD K3, K3, K3
    16:		MOVQ   AX, CX
0: K1 = ff
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:12 (PC: 0x11a6f47)
     7:	// func maskTest() byte
     8:	// Requires: AVX512BW, AVX512DQ, AVX512F
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
    10:		KXNORB K1, K1, K1
    11:		XORQ   AX, AX
=>  12:		VPXORQ Z0, Z0, Z0
    13:		KMOVB  K1, AX
    14:		KXNORQ K2, K2, K2
    15:		KXNORD K3, K3, K3
    16:		MOVQ   AX, CX
    17:		KXNORW K4, K4, K4
0: K1 = ff
(dlv) s
> github.com/vsivsi/adder.maskTest() ./add_amd64.s:13 (PC: 0x11a6f4d)
     8:	// Requires: AVX512BW, AVX512DQ, AVX512F
     9:	TEXT ·maskTest(SB), NOSPLIT, $0-1
    10:		KXNORB K1, K1, K1
    11:		XORQ   AX, AX
    12:		VPXORQ Z0, Z0, Z0
=>  13:		KMOVB  K1, AX
    14:		KXNORQ K2, K2, K2
    15:		KXNORD K3, K3, K3
    16:		MOVQ   AX, CX
    17:		KXNORW K4, K4, K4
    18:		KXNORB K5, K5, K5
0: K1 = 0
```

vsivsi:
@aarzilli   I'm looking at the gdbwire log output for stepping through this function:

```
// func maskTest2() byte
// Requires: AVX512BW, AVX512F
TEXT ·maskTest2(SB), NOSPLIT, $0-1
	MOVQ   $0xfecaedfeefbeadde, CX
	KMOVQ  CX, K1
	XORQ   CX, CX
	VPXORQ Z0, Z0, Z0
	MOVB   AL, ret+0(FP)
	RET
```

Log Here:  https://gist.github.com/vsivsi/f0a389a81bac9c6265d58c8491252a04

And if I'm interpreting the log correctly, it appears that perhaps lldb is to blame for clobbering K1? 

Look for the string `deadbeeffeedcafe` in the register state buffers in the log. I may not be interpreting it correctly... Does it look to you like that value is just zeroed out after executing the `VPXORQ` instruction?  

I don't _think_ I can spot `dlv` writing the wrong value back over the wire before the incorrect value comes back from `lldb`.

If the issue is in `lldb`, then I think I'm sunk debugging with this on MacOS, unless there's an easy way to make Delve use a newer version of `lldb` that the one that comes with the MacOS command line dev tools from Apple.  It is easy enough to install an up-to-date current LLVM bundle using Homebrew, but what are the odds I can somehow point delve at that and expect it to "just work".  I assume that such bugs will not still exist in the latest versions of LLVM's toolchain.

aarzilli:
> I don't think I can spot dlv writing the wrong value back over the wire before the incorrect value comes back from lldb.

Same. But again, it could be that the g command specifically it's broken.

>  It is easy enough to install an up-to-date current LLVM bundle using Homebrew, but what are the odds I can somehow point delve at that and expect it to "just work".

I'd say they are pretty high. You can point delve to a different build of debugserver by setting the DELVE_DEBUGSERVER_PATH environment variable.

> I assume that such bugs will not still exist in the latest versions of LLVM's toolchain.

That's a strong assumption. Lldb doesn't use the g command to read registers so it's entirely possible that any bugs there are not noticed by anyone besides delve.

vsivsi:
@derekparker yes, I'll try that next, but can't get to it until later this afternoon.

vsivsi:
@aarzilli @derekparker Yes, setting `gcmdok` to `false` in the `newProcess` function in `gdbserver.go` seems to clear up the ZMM_Hi256 interleaving and Kn mask clobbering issues I've been reporting on this thread. 

So what does this mean for delve? I guess I'm unclear on the history of why these two mechanisms for communicating with debugserver have been maintained in the codebase, given that one of them appears to be badly broken (because its not really used/supported by lldb itself?)

aarzilli:
> So what does this mean for delve? I guess I'm unclear on the history of why these two mechanisms for communicating with debugserver have been maintained in the codebase, given that one of them appears to be badly broken (because its not really used/supported by lldb itself?)

Both mechanisms are optional, because of how we use it 'g' is more efficient, but clearly if it doesn't work we can't use it.

aarzilli:
Alright, I have disabled g/G on debugserver unconditionally.

aarzilli:
Alright, I have disabled g/G on debugserver unconditionally.

derekparker (approved):
LGTM

vsivsi:
@aarzilli  @derekparker   Hi, unfortunately, it seems there was a regression as the commits were squashed on this PR before it was merged into `master`.  

Specifically, the update to the `Slice` function in `gdbserver.go` to add a case for 512 bit registers was somehow dropped. [This comment from me](https://github.com/go-delve/delve/issues/2479#issuecomment-843693050) documents the issue.

The lost code should be here: https://github.com/go-delve/delve/blob/master/pkg/proc/gdbserial/gdbserver.go#L1924-L1925

So... without that code, merging this PR did not actually fix the visibility of the vector registers when AVX512 is present. 

Sorry to be the bearer of bad news...

vsivsi:
Adding this code at the position indicated above rescues the functionality when AVX512 is present:

```golang
		case reginfo.Bitsize == 512:
			if !strings.HasPrefix(strings.ToLower(reginfo.Name), "zmm") || !floatingPoint {
				continue
			}

			value := regs.regs[reginfo.Name].value
			xmmName := "x" + reginfo.Name[1:]
			ymmName := "y" + reginfo.Name[1:]
			r = proc.AppendBytesRegister(r, strings.ToUpper(xmmName), value[:16])
			r = proc.AppendBytesRegister(r, strings.ToUpper(ymmName), value[16:32])
			r = proc.AppendBytesRegister(r, strings.ToUpper(reginfo.Name), value[32:])
```

PR #2499, opened 2021-05-19
service/dap: set function breakpoints while running

suzmue:
Use the same logic to halt the program execution for setting function breakpoints as we do for breakpoints.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #2501, opened 2021-05-20
tests: increase stack depth for TestFrameEvaluation

aarzilli:
Sometimes on Windows TestFrameEvaluation fails because the stacktrace
doesn't look deep enough.


derekparker (approved):
LGTM

PR #2502, opened 2021-05-20
service/dap: remove unnecessary capability check

suzmue:
The Types field is only set in childrenToDAPVariables if the client
capability is supported. There is no need to check again after the
children have been calculated.


aarzilli (approved):
LGTM

PR #2503, opened 2021-05-21
service/dap: log calls to debugger.Command

suzmue:
Add additional logging to the server.


aarzilli (commented):
The debugger already logs calls to Command itself.

suzmue:
Oh thanks! We need to update which log output we are showing then.

PR #2504, opened 2021-05-22
terminal,service: Add filtering and grouping to goroutines command

aarzilli:
Adds filtering and grouping to the goroutines command.

The current implementation of the goroutines command is modeled after
the threads command of gdb. It works well for programs that have up to
a couple dozen goroutines but becomes unusable quickly after that.

This commit adds the ability to filter and group goroutines by several
different properties, allowing a better debugging experience on
programs that have hundreds or thousands of goroutines.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -364,7 +364,7 @@ Print out info for every goroutine. The flag controls what information is shown 
 	-t	displays goroutine's stacktrace (an optional depth value can be specified, default: 10)
 	-l	displays goroutine's labels
 
-If no flag is specified the default is -u.
+If no flag is specified the default is -u, i.e. the first frame within the first 30 frames that is not executing a runtime private function.
 
 FILTERING
 
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -394,16 +394,16 @@ To only display goroutines that are running (or are not running) on a OS thread,
 	goroutines -with running
 	goroutines -without running
 	
-To only display goroutines that are classified as system (or non-system), use:
+To only display user (or runtime) goroutines, use:
 
-	goroutines -with system
-	goroutines -without system
+	goroutines -with user
+	goroutines -without user
 
 GROUPING
 
-	goroutines -group (userloc|curloc|goloc|startloc|running|system)
+	goroutines -group (userloc|curloc|goloc|startloc|running|user)
 
-Groups goroutines by the given location, running status or system classification, up to 5 goroutines per group will be displayed as well as the total number of goroutines in the group.
+Groups goroutines by the given location, running status or user classification, up to 5 goroutines per group will be displayed as well as the total number of goroutines in the group.
 
 	goroutines -group label key
 
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -353,18 +353,62 @@ Aliases: gr
 ## goroutines
 List program goroutines.
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [-t (stack trace)] [-l (labels)]
+	goroutines [-u|-r|-g|-s] [-t [depth]] [-l] [-with loc expr] [-without loc expr] [-group argument]
 
 Print out info for every goroutine. The flag controls what information is shown along with each goroutine:
 
-	-u	displays location of topmost stackframe in user code
+	-u	displays location of topmost stackframe in user code (default)
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
 	-s	displays location of the start function
-	-t	displays goroutine's stacktrace
+	-t	displays goroutine's stacktrace (an optional depth value can be specified, default: 10)
 	-l	displays goroutine's labels
 
 If no flag is specified the default is -u.
+
+FILTERING
+
+If -with or -without are specified only goroutines that match the given condition are returned.
+
+To only display goroutines where the specified location contains (or does not contain, for -without and -wo) expr as a substring, use:
+
+	goroutines -with (userloc|curloc|goloc|startloc) expr
+	goroutines -w (userloc|curloc|goloc|startloc) expr
+	goroutines -without (userloc|curloc|goloc|startloc) expr
+	goroutines -wo (userloc|curloc|goloc|startloc) expr
+	
+To only display goroutines that have (or do not have) the specified label key and value, use:
+	
+
+	goroutines -with label key=value
+	goroutines -without label key=value
+	
+To only display goroutines that have (or do not have) the specified label key, use:
+
+	goroutines -with label key
+	goroutines -without label key
+	
+To only display goroutines that are running (or are not running) on a OS thread, use:
+
+
+	goroutines -with running
+	goroutines -without running
+	
+To only display goroutines that are classified as system (or non-system), use:
+
+	goroutines -with system
+	goroutines -without system
+
+GROUPING
+
+	goroutines -group (userloc|curloc|goloc|startloc|running|system)
+
+Groups goroutines by the given location, running status or system classification, up to 5 goroutines per group will be displayed as well as the total number of goroutines in the group.
+
+	goroutines -group label key
+
+Groups goroutines by the value of the label with the specified key.
+
 
 Aliases: grs
 
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -220,18 +220,62 @@ If called with the linespec argument it will delete all the breakpoints matching
 toggle <breakpoint name or id>`},
 		{aliases: []string{"goroutines", "grs"}, group: goroutineCmds, cmdFn: goroutines, helpMsg: `List program goroutines.
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [-t (stack trace)] [-l (labels)]
+	goroutines [-u|-r|-g|-s] [-t [depth]] [-l] [-with loc expr] [-without loc expr] [-group argument]
 
 Print out info for every goroutine. The flag controls what information is shown along with each goroutine:
 
-	-u	displays location of topmost stackframe in user code
+	-u	displays location of topmost stackframe in user code (default)
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
 	-s	displays location of the start function
-	-t	displays goroutine's stacktrace
+	-t	displays goroutine's stacktrace (an optional depth value can be specified, default: 10)
 	-l	displays goroutine's labels
 
-If no flag is specified the default is -u.`},
+If no flag is specified the default is -u.
+
+FILTERING
+
+If -with or -without are specified only goroutines that match the given condition are returned.
+
+To only display goroutines where the specified location contains (or does not contain, for -without and -wo) expr as a substring, use:
+
+	goroutines -with (userloc|curloc|goloc|startloc) expr
+	goroutines -w (userloc|curloc|goloc|startloc) expr
+	goroutines -without (userloc|curloc|goloc|startloc) expr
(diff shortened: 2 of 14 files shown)


aarzilli:
This is what grouped goroutines look like:

Example

Documentation/cli/README.md
```diff
@@ -340,19 +340,57 @@ Aliases: gr
...
+
+If -with or -without are specified only goroutines that match the given condition are returned.
+
+	goroutines -with (userloc|curloc|goloc|startloc) expr
+	goroutines -w (userloc|curloc|goloc|startloc) expr
+	goroutines -without (userloc|curloc|goloc|startloc) expr
+	goroutines -wo (userloc|curloc|goloc|startloc) expr
+	
+
+Only display goroutines where the specified location contains (or does not contain, for -without and -wo) expr as a substring.
```
derekparker, line 366: I think this would be clearer written before the example instead of after, and punctuated with `:` instead of a period.

pkg/terminal/command.go
```diff
@@ -704,89 +742,200 @@ const (
...
-			default:
-				return fmt.Errorf("wrong argument: '%s'", arg)
+		case "-w", "-with":
+			filter, err := readGoroutinesFilter(args, &i)
+			if err != nil {
+				return err
 			}
+			filters = append(filters, *filter)
+
+			//TODO: implement
```
derekparker, line 812: What's left to do here?

service/client.go
```diff
@@ -114,6 +114,8 @@ type Client interface {
 
 	// ListGoroutines lists all goroutines.
 	ListGoroutines(start, count int) ([]*api.Goroutine, int, error)
+	// ListGoroutinesWithFilter lists goroutines matching the filters
+	ListGoroutinesWithFilter(start, count int, filters []api.ListGoroutinesFilter, groupBy api.GoroutineField, groupByArg string, maxGroupExamples, maxGroups int) ([]*api.Goroutine, []api.GoroutineGroup, int, bool, error)
```
derekparker, line 118: I'd prefer a option struct argument instead of this large list of arguments.

service/debugger/debugger.go
```diff
@@ -1434,6 +1434,127 @@ func (d *Debugger) Goroutines(start, count int) ([]*proc.G, int, error) {
...
+// A maximum of maxGoroutinesPerGroup are saved in each group, but the total
+// number of goroutines in each group is recorded.
+func (d *Debugger) GroupGoroutines(gs []*proc.G, groupBy api.GoroutineField, groupByArg string, maxGoroutinesPerGroup, maxGroups int) ([]*proc.G, []api.GoroutineGroup, bool) {
+	if groupBy == api.GoroutineFieldNone {
+		return gs, nil, false
+	}
+	d.targetMutex.Lock()
+	defer d.targetMutex.Unlock()
+
+	examples := map[string][]*proc.G{}
```
derekparker, line 1513: Why call this examples? Not saying to change it just yet I just don't understand the naming. Is it because it's an example of the Goroutines which belong to this group (since not all may be included)?

derekparker (changes requested):

pkg/terminal/command.go
aarzilli: Nothing, left over note.

service/client.go
aarzilli: I think that it would be complicated to capture every possibility on a struct, for example:

```
goroutines -w label x -wo label x=b
```

will show all goroutines with a label with key 'x' that do not have value 'b'. Combining multiple with/without conditions on the same location could also be useful.

service/debugger/debugger.go
aarzilli: Yes. If you have a better name I'm open to changing it.

Documentation/cli/README.md
aarzilli: Done.

service/debugger/debugger.go
derekparker: Let's go with `members` or `groupMembers`.

Documentation/cli/README.md
```diff
@@ -340,19 +340,58 @@ Aliases: gr
...
 
 If no flag is specified the default is -u.
 
+FILTERING
+
+If -with or -without are specified only goroutines that match the given condition are returned.
+
+To only display goroutines where the specified location contains (or does not contain, for -without and -wo) expr as a substring, use:
+
+	goroutines -with (userloc|curloc|goloc|startloc) expr
```
polinasok, line 362: Is userloc equivalent of -u and curloc equivalent of -r?

Documentation/cli/README.md
```diff
@@ -340,19 +340,58 @@ Aliases: gr
...
 	-r	displays location of topmost stackframe (including frames inside private runtime functions)
 	-g	displays location of go instruction that created the goroutine
 	-s	displays location of the start function
-	-t	displays goroutine's stacktrace
+	-t	displays goroutine's stacktrace (an optional depth value can be specified, default: 10)
 	-l	displays goroutine's labels
 
 If no flag is specified the default is -u.
 
+FILTERING
```
polinasok, line 356: Is there a way to filter runtime goroutines vs user goroutines?

Documentation/cli/README.md
```diff
@@ -340,19 +340,58 @@ Aliases: gr
...
+	goroutines -with label key
+	goroutines -without label key
+	
+To only display goroutines that are running (or are not running) on a OS thread, use:
+
+
+	goroutines -with running
+	goroutines -without running
+
+GROUPING
```
polinasok, line 384: Would one be able to group as running vs not running?

Documentation/cli/README.md
```diff
@@ -340,19 +340,58 @@ Aliases: gr
...
+	
+To only display goroutines that have (or do not have) the specified label key, use:
+
+	goroutines -with label key
+	goroutines -without label key
+	
+To only display goroutines that are running (or are not running) on a OS thread, use:
+
+
+	goroutines -with running
```
polinasok, line 381: Would the sleepy goroutines in your test show up as running or not running?

Documentation/cli/README.md
aarzilli: Yes.

service/client.go
aarzilli: I'm not sure I understand, would this be:

```
type ListGoroutinesOptions struct {
    Filters []ListGoroutinesFilter
    ListGoroutinesGroupingOptions
}

type ListGoroutinesGroupingOptions struct {
    GroupBy GoroutineField
    GroupByArg string
    MaxGroupExamples int
    MaxGroups int
}
```

?

Documentation/cli/README.md
aarzilli: No, but we could add one by copying isSystemGoroutine.

Documentation/cli/README.md
aarzilli: Yes. I'll add it to the documentation.

Documentation/cli/README.md
aarzilli: Only the ones that have a thread id in parenthesis are considered to be running.

service/client.go
derekparker: I was thinking out loud and suggesting either what you mentioned _or_ just doing `ListGoroutinesGroupingOptions`, but since both filter and group are optional I think it makes sense to put all arguments related to those features in an optional struct. That way if the caller does not want to use those features they only have to pass through a single `nil` for that struct instead of the zero value of 4-5 arguments.

Documentation/cli/README.md
polinasok: That would be great. @suzmue was just looking into ways to hide those from the users.

aarzilli: Added.

service/client.go
aarzilli: Ok, done.

Documentation/cli/README.md
```diff
@@ -340,19 +340,63 @@ Aliases: gr
...
+	goroutines -without running
+	
+To only display goroutiens that are classified as system (or non-system), use:
```
polinasok, line 384: typo: goroutiens

polinasok, line 384: Is system different from runtime? What exactly does system cover?

Documentation/cli/README.md
```diff
@@ -340,19 +340,63 @@ Aliases: gr
 ## goroutines
 List program goroutines.
 
-	goroutines [-u (default: user location)|-r (runtime location)|-g (go statement location)|-s (start location)] [-t (stack trace)] [-l (labels)]
+	goroutines [-u|-r|-g|-s] [-t [depth]] [-l] [-with loc expr] [-without loc expr] [-group argument]
 
 Print out info for every goroutine. The flag controls what information is shown along with each goroutine:
 
-	-u	displays location of topmost stackframe in user code
+	-u	displays location of topmost stackframe in user code (default)
```
polinasok, line 347: How exactly is user vs non-user code defined?

polinasok (commented):
Thank you for making the changes. 

Bw, I find it helpful in code reviews when the new commit doesn't force-override the old one. Then you can see just the changes since the previous version and it is possible to hide resolved comments

Documentation/cli/README.md
aarzilli: Not a private runtime function.

Documentation/cli/README.md
aarzilli: >Is system different from runtime?
I don't know, what do you mean with runtime?
>What exactly does system cover?
Goroutines not started by user code, except for the main goroutine.

aarzilli: Fixed.

polinasok: Then perhaps it would be more user-friendly to have "with/without user" since I think that one is not controversial and is likely to be of the most interest to the users.

In some offline conversation recently there was a discussion about user vs runtime goroutines, where "runtime" was used to refer to anything that was not started by the user (plus main, of course).

Documentation/cli/README.md
polinasok: So then any standard library function would still be user stack frame, correct?

aarzilli: Yes.

Documentation/cli/README.md
aarzilli: Done.

polinasok:
#2522 was merged. It has a todo to share code with this PR.
https://github.com/go-delve/delve/pull/2522/files/1be89ec08d0cee550faefccd5d694b9d19ee7086#r650137044

aarzilli:
> #2522 was merged. It has a todo to share code with this PR.
> https://github.com/go-delve/delve/pull/2522/files/1be89ec08d0cee550faefccd5d694b9d19ee7086#r650137044

Done.

derekparker (approved):
LGTM

I'll wait for @polinasok to +1 this as well before merging since she has already left some feedback.

derekparker:
@polinasok mind giving this another look when you get a chance.

Documentation/cli/README.md
polinasok: Looks like I am not the only one wondering about the distinction: https://github.com/go-delve/delve/issues/2542#issuecomment-863005481. I think the clarification you provided with a mention of scripting alternatives might be useful in the doc as well.

service/dap/server.go
```diff
@@ -1539,11 +1539,7 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 	// Determine if the goroutine is a system goroutine.
 	// TODO(suzmue): Use the System() method defined in: https://github.com/go-delve/delve/pull/2504
```
polinasok, line 1540: This TODO comment should now be taken out.

polinasok (changes requested):

service/dap/server.go
aarzilli: Done.

Documentation/cli/README.md
aarzilli: Done.

polinasok (approved):
LGTM

PR #2505, opened 2021-05-25
_scripts: change gen-backend_test_health.go to be more stable

aarzilli:
Change gen-backend_test_health.go so that it only prints the number of
tests skipped and not the total or the percentage.

As it is it causes too many merge conflicts.


derekparker (approved):
LGTM

PR #2508, opened 2021-05-26
service/dap: test refactoring

polinasok:
Separates checking already read messages into separate helpers. Renames all helpers to follow the same convention:
- Expect = read + verify
- Check = only verify


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #2512, opened 2021-05-26
service/dap: implement array, slice, and map paging

suzmue:
If the client supports paging, we allow them to fetch array and slice items in chunks that we assume will be of a reasonable size. For example, VS Code requests indexed variables in chunks of 100.

Fixes golang/vscode-go#1518


suzmue:
Before:
Screen Shot 2021-05-26 at 2 31 36 PM

After (can inspect all elements):
Screen Shot 2021-05-26 at 2 30 35 PM

suzmue:
Map paging (smaller maps remain unchanged):

Before:
Screen Shot 2021-05-26 at 5 18 51 PM

After:
Screen Shot 2021-05-26 at 5 18 28 PM

Before:
Screen Shot 2021-05-26 at 5 18 56 PM

After:
Screen Shot 2021-05-26 at 5 18 14 PM

service/dap/server.go
```diff
@@ -1831,8 +1870,8 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 			}
 		}
 		// TODO(polina): as far as I can tell, evaluateName is ignored by vscode for expression variables.
-		// Should it be skipped altogether for all levels?
-		exprVal, exprRef := s.convertVariable(exprVar, fmt.Sprintf("(%s)", request.Arguments.Expression))
+		// Should it be skipped alltogether for all levels?
```
polinasok, line 1873: Hana just changed this with the last change. Should be one l.

service/dap/server.go
```diff
@@ -1831,8 +1870,8 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
...
+		exprVal, exprRef := s.convertVariableEvaluate(exprVar, fmt.Sprintf("(%s)", request.Arguments.Expression))
```
polinasok, line 1874: Indexed paging cannot be supported with expressions?

service/dap/server.go
```diff
@@ -1702,7 +1741,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 		if v.Len > int64(len(v.Children)) { // Not fully loaded
 			if v.Base != 0 && len(v.Children) == 0 { // Fully missing
 				value = reloadVariable(v, qualifiedNameOrExpr)
-			} else { // Partially missing (TODO)
+			} else if includeLoadedInfo {
```
polinasok, line 1744: FYI It would be more clear what's going on if you could reference the paging capability here directly. I see you have it as an option because of the expression evaluation. It's not clear to me why - please see question below.

Also, all the other reloading happens in this function. If you cannot do the same for arrays/slices, it would be helpful to add a comment here with some insights.

service/dap/server_test.go
```diff
@@ -1476,28 +1476,88 @@ func TestVariablesLoading(t *testing.T) {
 					// String partially missing based on LoadConfig.MaxStringLen
 					checkVarExact(t, locals, -1, "longstr", "longstr", "\"very long string 0123456789a0123456789b0123456789c0123456789d012...+73 more\"", "string", noChildren)
 
+					checkArrayChildren := func(t *testing.T, longarr *dap.VariablesResponse, parentName string, start int) {
+						t.Helper()
+						for i, child := range longarr.Body.Variables {
+							idx := i + start
```
polinasok, line 1482: start + i please

service/dap/server_test.go
```diff
@@ -1476,28 +1476,88 @@ func TestVariablesLoading(t *testing.T) {
...
 
 					// Map partially missing based on LoadConfig.MaxArrayValues
-					ref = checkVarRegex(t, locals, -1, "m1", "m1", `\(loaded 64/66\) map\[string\]main\.astruct \[.+\.\.\.\+2 more\]`, `map\[string\]main\.astruct`, hasChildren)
+					ref = checkVarRegex(t, locals, -1, "m1", "m1", `map\[string\]main\.astruct \[.+\.\.\.\+2 more\]`, `map\[string\]main\.astruct`, hasChildren)
 					if ref > 0 {
 						client.VariablesRequest(ref)
 						m1 := client.ExpectVariablesResponse(t)
 						checkChildren(t, m1, "m1", 64)
+
+						client.IndexedVariablesRequest(ref, 0, 66)
```
polinasok, line 1535: I am still a bit fuzzy on how exactly the client handles, but I imagine the decision to send an indexed request is based on whether the indexed property was set when the variable was returned as part of the children of the parent variable response? So shouldn't you mimic the same behavior here?

service/dap/server_test.go
```diff
@@ -1476,28 +1476,88 @@ func TestVariablesLoading(t *testing.T) {
...
+						checkChildren(t, longarr, "longslice", 100)
+						checkArrayChildren(t, longarr, "longslice", 0)
+
+						client.IndexedVariablesRequest(ref, 50, 50)
+						longarr = client.ExpectVariablesResponse(t)
+						checkChildren(t, longarr, "longslice", 50)
+						checkArrayChildren(t, longarr, "longslice", 50)
 					}
 
 					// Map partially missing based on LoadConfig.MaxArrayValues
```
polinasok, line 1528: Might be a good idea to update these comments with more details as to how we handle the missing values.

polinasok (commented):
Could you please provide a log what requests the client sends when paging is supported vs not?

Also, does this feature work with debug console, watch? What happens to features like "Copy Expression" and "Copy Value"?

service/dap/server.go
```diff
@@ -1591,6 +1613,17 @@ func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 				Value:              cvalue,
 				VariablesReference: cvarref,
 			}
+
+			if c.Kind == reflect.Array || c.Kind == reflect.Slice || c.Kind == reflect.Map {
+				children[i].IndexedVariables = int(c.Len)
+			}
+
+			// TODO(suzmue): Paging named variables are not yet supported by the VS Code UI
```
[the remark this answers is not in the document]
suzmue: I was actually thinking about this some more and I think it will make sense to use named variables to include things like "len", "cap", "string value" or whatever else and keep indexed for map values and arrays.

I'll remove this comment since it now seems like indexed is the right choice for maps.

service/dap/server_test.go
suzmue: I think likely that is true.

I was including this test to make sure we were still doing something reasonable if the client did not specify the indexed variables it wanted, without having to attach with a new client.

service/dap/server.go
suzmue: Evaluating an expression does not include support for paging. The request doesn't have context for how many children there may be so it cannot send a request that has a reasonable number. We could possibly parse slice expressions and set the LoadConfig to load the correct number of items, but this could be arbitrarily big defined by the user.

suzmue:
Logs without paging:

```
2021-05-27T12:57:51-04:00 debug layer=dap [<- from client]{"seq":13,"type":"request","command":"variables","arguments":{"variablesReference":1007,"format":{}}}
2021-05-27T12:57:51-04:00 debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":13,"success":true,"command":"variables","body":{"variables":[{...]}}
```

Logs with paging supported:

```
{"seq":12,"type":"request","command":"variables","arguments":{"variablesReference":1001,"format":{}}}
2021-05-27T12:48:49-04:00 debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":12,"success":true,"command":"variables","body":{"variables":[{"name":"myMap","value":"map[string]int [\"A\": 12, \"B\": 15, ]","type":"map[string]int","presentationHint":{},"evaluateName":"myMap","variablesReference":1002,"indexedVariables":1},{"name":"v","value":"[]uint8 len: 13, cap: 13, [102,111,111,98,97,114,98,97,122,98,97,98,97]","type":"[]uint8","presentationHint":{},"evaluateName":"v","variablesReference":1003,"indexedVariables":13},{"name":"animal","value":"\"dog\"","type":"string","presentationHint":{},"evaluateName":"animal","variablesReference":0},{"name":"longString","value":"\"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...+257 more\"","type":"string","presentationHint":{},"evaluateName":"longString","variablesReference":0},{"name":"m","value":"map[string]int [...]","type":"map[string]int","presentationHint":{},"evaluateName":"m","variablesReference":1004,"indexedVariables":150},{"name":"s","value":"\"goodbye\"","type":"string","presentationHint":{},"evaluateName":"s","variablesReference":0},{"name":"x","value":"main.hello {i: 5, s: \"hello\"}","type":"main.hello","presentationHint":{},"evaluateName":"x","variablesReference":1005},{"name":"m2","value":"map[main.hello]main.hello [...]","type":"map[main.hello]main.hello","presentationHint":{},"evaluateName":"m2","variablesReference":1006,"indexedVariables":150},{"name":"length","value":"300","type":"int","presentationHint":{},"evaluateName":"length","variablesReference":0},{"name":"longArr","value":"[]main.hello len: 300, cap: 300, [{i: 0, s: \"d\"},{i: 1, s: \"d\"},{i: 2, s: \"d\"},{i: 3, s: \"d\"},{i: 4, s: \"d\"},{i: 5, s: \"d\"},{i: 6, s: \"d\"},{i: 7, s: \"d\"},{i: 8, s: \"d\"},{i: 9, s: \"d\"},{i: 10, s: \"d\"},{i: 11, s: \"d\"},{i: 12, s: \"d\"},{i: 13, s: \"d\"},{i: 14, s: \"d\"},{i: 15, s: \"d\"},{i: 16, s: \"d\"},{i: 17, s: \"d\"},{i: 18, s: \"d\"},{i: 19, s: \"d\"},{i: 20, s: \"d\"},{i: 21, s: \"d\"},{i: 22, s: \"d\"},{i: 23, s: \"d\"},{i: 24, s: \"d\"},{i: 25, s: \"d\"},{i: 26, s: \"d\"},{i: 27, s: \"d\"},{i: 28, s: \"d\"},{i: 29, s: \"d\"},{i: 30, s: \"d\"},{i: 31, s: \"d\"},{i: 32, s: \"d\"},{i: 33, s: \"d\"},{i: 34, s: \"d\"},{i: 35, s: \"d\"},{i: 36, s: \"d\"},{i: 37, s: \"d\"},{i: 38, s: \"d\"},{i: 39, s: \"d\"},{i: 40, s: \"d\"},{i: 41, s: \"d\"},{i: 42, s: \"d\"},{i: 43, s: \"d\"},{i: 44, s: \"d\"},{i: 45, s: \"d\"},{i: 46, s: \"d\"},{i: 47, s: \"d\"},{i: 48, s: \"d\"},{i: 49, s: \"d\"},{i: 50, s: \"d\"},{i: 51, s: \"d\"},{i: 52, s: \"d\"},{i: 53, s: \"d\"},{i: 54, s: \"d\"},{i: 55, s: \"d\"},{i: 56, s: \"d\"},{i: 57, s: \"d\"},{i: 58, s: \"d\"},{i: 59, s: \"d\"},{i: 60, s: \"d\"},{i: 61, s: \"d\"},{i: 62, s: \"d\"},{i: 63, s: \"d\"},...+236 more]","type":"[]main.hello","presentationHint":{},"evaluateName":"longArr","variablesReference":1007,"indexedVariables":300}]}}
2021-05-27T12:48:58-04:00 debug layer=dap [<- from client]{"seq":13,"type":"request","command":"variables","arguments":{"variablesReference":1007,"filter":"indexed","count":100,"format":{}}}
2021-05-27T12:48:58-04:00 debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":13,"success":true,"command":"variables","body":{"variables":[...]}}
2021-05-27T12:49:02-04:00 debug layer=dap [<- from client]{"seq":14,"type":"request","command":"variables","arguments":{"variablesReference":1007,"filter":"indexed","start":200,"count":100,"format":{}}}
2021-05-27T12:49:02-04:00 debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":14,"success":true,"command":"variables","body":{"variables":[...]}}
```

service/dap/server.go
```diff
@@ -1591,6 +1613,17 @@ func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 				Value:              cvalue,
 				VariablesReference: cvarref,
 			}
+
+			if c.Kind == reflect.Array || c.Kind == reflect.Slice || c.Kind == reflect.Map {
```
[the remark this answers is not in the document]
suzmue: Oh oops this got leftover. Removing.

service/dap/server.go
suzmue: I'm sorry you are correct evaluate requests do support paging. I will either make the change or add a TODO.

service/dap/server.go
```diff
@@ -1456,6 +1456,27 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
...
 	}
 
+	if request.Arguments.Filter == "indexed" {
+		start, count := request.Arguments.Start, request.Arguments.Count
+		indexedLoadConfig := DefaultLoadConfig
+		indexedLoadConfig.MaxArrayValues = count
+		if v.Kind == reflect.Array || v.Kind == reflect.Slice {
+			newV, err := v.Reslice(int64(start), int64(start+count))
+			if err != nil {
+				s.sendErrorResponse(request.Request, UnableToLookupVariable, "Unable to lookup variable", fmt.Sprintf("unknown reference %d", ref))
```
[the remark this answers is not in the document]

service/dap/server.go
suzmue: Added support for evaluate.

pkg/proc/variables.go
```diff
@@ -1210,16 +1210,16 @@ func loadValues(vars []*Variable, cfg LoadConfig) {
 }
 
 // Extracts the value of the variable at the given address.
-func (v *Variable) loadValue(cfg LoadConfig) {
+func (v *Variable) LoadValue(cfg LoadConfig) {
```
derekparker, line 1213: Please fixup doc comment now that this is exported (e.g. `LoadValue extracts ...`).

derekparker (changes requested):
Just a few nits.

I thought a bit about the implications of exporting the fields from Variable because we typically try to shield the rest of the codebase from doing too much with proc because that's where the dragons roam, but I think in this case since you already have a reference to the variable, and the loading of variable information is pretty well defined through the internal API I don't think this will be an issue. My one hesitancy is that we're exposing some internal state via `Loaded`.

I know that `fullyQualifiedVariable` has `fullyQualifiedNameOrExpr` as well, another option instead of exposing these fields is to use `Debugger.EvalVariableInScope` with the proper expression / load config to reslice the slice or map: https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#loading-more-of-a-variable.

suzmue:
> Just a few nits.
> 
> I thought a bit about the implications of exporting the fields from Variable because we typically try to shield the rest of the codebase from doing too much with proc because that's where the dragons roam, but I think in this case since you already have a reference to the variable, and the loading of variable information is pretty well defined through the internal API I don't think this will be an issue. My one hesitancy is that we're exposing some internal state via `Loaded`.
> 
> I know that `fullyQualifiedVariable` has `fullyQualifiedNameOrExpr` as well, another option instead of exposing these fields is to use `Debugger.EvalVariableInScope` with the proper expression / load config to reslice the slice or map: https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md#loading-more-of-a-variable.

Thanks for the feedback @derekparker! I went this route because of the conversation on a [previous change](https://github.com/go-delve/delve/pull/2455#discussion_r622117378) between @aarzilli and @polinasok where this strategy was suggested.

I refactored the code a bit to expose less of the internal implementation and was able to avoid exporting any of the fields (by adding a single exported function to proc).

pkg/proc/variables.go
suzmue: no longer exporting this

pkg/proc/variables.go
```diff
@@ -116,11 +116,11 @@ type Variable struct {
 	closureAddr uint64
 
 	// number of elements to skip when loading a map
-	mapSkip int
+	MapSkip int
 
 	Children []Variable
 
-	loaded     bool
+	Loaded     bool
```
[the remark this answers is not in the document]
suzmue: no longer exporting loaded

pkg/proc/variables.go
```diff
@@ -116,11 +116,11 @@ type Variable struct {
 	closureAddr uint64
 
 	// number of elements to skip when loading a map
-	mapSkip int
+	MapSkip int
```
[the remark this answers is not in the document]
suzmue: no longer exporting this.

service/dap/server.go
suzmue: I added the reference to the paging capability here and added an explanation comment about

pkg/proc/eval.go
```diff
@@ -1950,6 +1950,26 @@ func (v *Variable) mapAccess(idx *Variable) (*Variable, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	case reflect.Map:
+		newV = v
+		newV.Children = nil
+		newV.loaded = false
+		newV.mapSkip = start
+	}
+	newV.loadValue(cfg)
```
aarzilli, line 1969: if v is neither an array, slice or map this will be a nil pointer dereference.

derekparker:
> I refactored the code a bit to expose less of the internal implementation and was able to avoid exporting any of the fields (by adding a single exported function to proc).

Thank you! Great refactor, I like this new approach.

pkg/proc/eval.go
```diff
@@ -1950,6 +1950,26 @@ func (v *Variable) mapAccess(idx *Variable) (*Variable, error) {
...
+func (v *Variable) Reslice(start, count int, cfg LoadConfig) (newV *Variable, err error) {
+	cfg.MaxArrayValues = count
+	switch v.Kind {
+	case reflect.Array, reflect.Slice:
+		newV, err = v.reslice(int64(start), int64(start+count))
+		if err != nil {
+			return nil, err
+		}
+	case reflect.Map:
+		newV = v
```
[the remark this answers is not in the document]
derekparker: You should use `v.clone()` here.

service/debugger/debugger.go
```diff
@@ -1414,6 +1414,19 @@ func (d *Debugger) EvalVariableInScope(goid, frame, deferredCall int, symbol str
 	return s.EvalVariable(symbol, cfg)
 }
 
+// Reslice will attempt to 'reslice' a map, array or slice so that the values
+// from [start:start+count] are loaded.
+func (d *Debugger) Reslice(v *proc.Variable, start, count int, cfg proc.LoadConfig) (*proc.Variable, error) {
```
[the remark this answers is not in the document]
derekparker: I prefer this approach simply due to the fact that the caller doesn't have to worry about locking target, that can remain an internal implementation detail of how `Debugger` interacts with `Target`. I know it adds another layer of indirection which can be a bit annoying, but it provides convenience for the caller.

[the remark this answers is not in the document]
aarzilli: I'm worried about the proliferation of debugger methods that just acquire a lock and call a function in proc and only exist to implement a feature of DAP.

pkg/proc/eval.go
```diff
@@ -1950,6 +1950,26 @@ func (v *Variable) mapAccess(idx *Variable) (*Variable, error) {
 	return nil, fmt.Errorf("key not found")
 }
 
+// Reslice returns a new array, slice or map that starts at index start and contains
+// up to count children.
+func (v *Variable) Reslice(start, count int, cfg LoadConfig) (newV *Variable, err error) {
```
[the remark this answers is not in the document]
suzmue: Great feedback. I changed this to load up to cfg.MaxArrayLength and removed the count arguments for both LoadResliced functions.

pkg/proc/eval.go
suzmue: Fixed.

aarzilli (approved):
LGTM

service/debugger/debugger.go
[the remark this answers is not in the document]
derekparker: @aarzilli I share that concern as well, and I'm open to suggestions for refactoring to remove some layers of abstraction that we may not need. But I think my original concern still stands that having the caller control the lock seems a bit less than ideal.

Another idea could be to have a helper function within the DAP code which calls `Debugger.LockTarger/UnlockTarget` within a single function call to emulate what's happening now but not make it part of the public Debugger API. Thoughts on that?

suzmue:
Pretty sure the test failure is unrelated

service/debugger/debugger.go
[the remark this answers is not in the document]
aarzilli: > But I think my original concern still stands that having the caller control the lock seems a bit less than ideal.

I don't care much about that since there's already a call to LockTarget/UnlockTarget in dap/server.go to convert goroutines efficiently.

> Another idea could be to have a helper function within the DAP code which calls Debugger.LockTarger/UnlockTarget within a single function call to emulate what's happening now but not make it part of the public Debugger API. Thoughts on that?

I don't have a preference between that and a helper in debugger.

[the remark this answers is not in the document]
derekparker: > > But I think my original concern still stands that having the caller control the lock seems a bit less than ideal.
> 
> I don't care much about that since there's already a call to LockTarget/UnlockTarget in dap/server.go to convert goroutines efficiently.

True, and yeah I know that exists. I just want to make it easy for the right thing to happen. That's a benefit of having an abstraction layer like `Debugger` provides, is the caller of those methods never has to think about locking/unlocking. That decision is made only once during implementation. More code == more bugs, so centralizing the mutex locking/unlocking seems safer.

> > Another idea could be to have a helper function within the DAP code which calls Debugger.LockTarger/UnlockTarget within a single function call to emulate what's happening now but not make it part of the public Debugger API. Thoughts on that?
> 
> I don't have a preference between that and a helper in debugger.

Alright, well I'm fine with this as it stands. Unless you have major concerns we can merge as is (after @polinasok +1).

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -1651,24 +1654,20 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
...
 // a scalar variable cannot be "dereferenced" to get its elements (as there are none).
-func (s *Server) convertVariableEvaluate(v *proc.Variable, qualifiedNameOrExpr string) (value string, variablesReference int) {
-	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, false, true)
-}
-
 func (s *Server) convertVariable(v *proc.Variable, qualifiedNameOrExpr string) (value string, variablesReference int) {
 	// If the client supports paging, we expect them to request the variables in reasonable chunks, so we do
 	// not want to display the loaded information.
-	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, false, !s.clientCapabilities.supportsVariablePaging)
+	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, false)
```
polinasok, line 1661: You no longer need the comment above this.

polinasok (changes requested):

service/dap/server.go
```diff
@@ -1702,7 +1741,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 		if v.Len > int64(len(v.Children)) { // Not fully loaded
 			if v.Base != 0 && len(v.Children) == 0 { // Fully missing
 				value = reloadVariable(v, qualifiedNameOrExpr)
-			} else { // Partially missing (TODO)
```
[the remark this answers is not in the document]
suzmue: I don't think that I actually addressed the reloading of the other variables in this function. I am going to leave it for now and can address this in a separate PR that looks at reloading a variable instead of calling EvalVariableInScope if thats alright.

service/dap/server.go

service/dap/handles.go
```diff
@@ -25,6 +25,7 @@ type fullyQualifiedVariable struct {
 	// True if this represents variable scope
 	isScope bool
 	// startIndex is the index of the first child for an array or slice.
+	// This variable represents a slice of the array.
```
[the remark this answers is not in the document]
suzmue: updated this to chunk.

service/dap/server.go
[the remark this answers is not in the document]
polinasok: Sorry I was not more clear. I thought it could be a good idea to update the how-to details of the TODO given what you went through exposing some of the internal proc hooks. I agree that any reloading should be done in a separate PR.

service/dap/server_test.go
```diff
@@ -2607,7 +2688,7 @@ func TestEvaluateRequest(t *testing.T) {
 					// Variable lookup that's not fully loaded
 					client.EvaluateRequest("ba", 1000, "this context will be ignored")
 					got = client.ExpectEvaluateResponse(t)
-					checkEval(t, got, "(loaded 64/200) []int len: 200, cap: 200, [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,...+136 more]", hasChildren)
+					checkEval(t, got, "[]int len: 200, cap: 200, [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,...+136 more]", hasChildren)
```
polinasok, line 2691: Should there also be a check for indexed children here and below? Looks like these are testing the feature as it would work with evaluated expressions (watch, repl, etc)

service/dap/server.go
```diff
@@ -1900,13 +1900,13 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 	// Those that are fully missing (e.g. due to hitting MaxVariableRecurse), can be reloaded in place.
 	// Those that are partially missing (e.g. MaxArrayValues from a large array), need a more creative solution
 	// that is still to be determined. For now, clearly communicate when that happens with additional value labels.
-	// TODO(polina): look into layered/paged loading for truncated strings, arrays, maps and structs.
```
polinasok, line 1860: Also don't need comments on the previous likes as well "need a more creative solution" for large arrays, etc.

polinasok:
> Also, does this feature work with debug console, watch? What happens to features like "Copy Expression" and "Copy Value"?

@suzmue Have you head a chance to check all these?

suzmue:
This works for hover as well!

Screen Shot 2021-06-07 at 4 48 16 PM

polinasok:
> This works for hover as well!

That's great news! Between Variables pane, hover and debug console, this will be quite handy.
Did you get a chance to check the right-click options as well?

polinasok:
LGTM

polinasok (approved):

PR #2513, opened 2021-05-26
dap: add 'clipboard' support, and truncate a long value

hyangah:
* add 'clipboard' capability
* apply a larger string limit for 'hover' and 'clipboard' context
* truncate the string representation of a compound type (or pointer of a compound type) variable if the length of the value is larger than 256. But when a variable is returned as a result of Evaluate Request for hover or copy value feature (clipboard or variables context), do not truncate the value.

Updates golang/vscode-go#1318
Fixes golang/vscode-go#1435

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1815,27 +1815,33 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 // custom, a zero reference, reminiscent of a zero pointer, is used to indicate that
 // a scalar variable cannot be "dereferenced" to get its elements (as there are none).
 func (s *Server) convertVariable(v *proc.Variable, qualifiedNameOrExpr string) (value string, variablesReference int) {
-	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, false, false)
+	return s.convertVariableWithOpts(v, qualifiedNameOrExpr, 0)
 }
 
 func (s *Server) convertVariableToString(v *proc.Variable) string {
-	val, _ := s.convertVariableWithOpts(v, "", true, false)
+	val, _ := s.convertVariableWithOpts(v, "", skipRef)
 	return val
 }
 
 // defaultMaxValueLen is the max length of a string representation of a compound or reference
 // type variable.
-const defaultMaxValueLen = 1 << 10 // 1K
+const defaultMaxValueLen = 1 << 8 // 256
+
+// Flags for convertVariableWithOpts option.
+const (
+	skipRef = 1 << iota
+	showFullValue
+)
 
 // convertVariableWithOpts allows to skip reference generation in case all we need is
 // a string representation of the variable. When the variable is a compound or reference
 // type variable and its full string representation can be larger than defaultMaxValueLen,
-// this returns a truncated value unless showFull is set.
-func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr string, skipRef, showFull bool) (value string, variablesReference int) {
+// this returns a truncated value unless showFull option flag is set.
+func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr string, opts int) (value string, variablesReference int) {
 	canHaveRef := false
 	maybeCreateVariableHandle := func(v *proc.Variable) int {
 		canHaveRef = true
-		if skipRef {
+		if opts&skipRef != 0 {
 			return 0
 		}
 		return s.variableHandles.create(&fullyQualifiedVariable{v, qualifiedNameOrExpr, false /*not a scope*/})
@@ -1971,7 +1977,8 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 			variablesReference = maybeCreateVariableHandle(v)
 		}
 	}
-	if len(value) > defaultMaxValueLen && !showFull && canHaveRef {
+	canTruncateValue := showFullValue&opts == 0
+	if len(value) > defaultMaxValueLen && canTruncateValue && canHaveRef {
 		value = value[:defaultMaxValueLen] + "..."
 	}
 	return value, variablesReference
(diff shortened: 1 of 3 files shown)


hyangah:
Arbitrarily chosen 1K, Still 1K is too large for my taste.

Before:
Screen Shot 2021-05-26 at 2 58 52 PM

After: 
Screen Shot 2021-05-26 at 2 59 42 PM

Still it's possible that we'd end up sending a large JSON message for variables response if the requested variable/scope contains many long variables. This PR doesn't mean to prevent that.

service/dap/server.go
```diff
@@ -1613,26 +1614,34 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
...
+// type variable.
+const defaultMaxValueLen = 1 << 10 // 1K
+
 // convertVariableWithOpts allows to skip reference generation in case all we need is
-// a string representation of the variable.
-func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr string, skipRef bool) (value string, variablesReference int) {
+// a string representation of the variable. When the variable is a compound or reference
+// type variable and its full string representation can be larger than defaultMaxValueLen,
+// this returns a truncated value unless showFull is set.
+func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr string, skipRef, showFull bool) (value string, variablesReference int) {
```
aarzilli, line 1633: I think this is enough boolean flags that it should be a bitfield.

derekparker:
> Arbitrarily chosen 1K, Still 1K is too large for my taste.

Would you like time to live with it and possibly change this before it gets submitted?

suzmue:
> Arbitrarily chosen 1K, Still 1K is too large for my taste.

I agree I think that these strings become unnecessarily long.

When I was testing paging, this was the cause of some things being so slow that it was unusable for investigating the values.

I would say drop it to ~100 (or even less) especially if it has children to expand to get more information.

service/dap/server.go
hyangah: done. PTAL

hyangah:
Lowered the max limit to 256. Tried 128 and it broke some tests and some didn't look nice with 128 (e.g. slice of pointers to a compound type, ...) due to all the type info and len/cap info before the actual data. :-(
I think a better solution is to truncate more intelligently, but I don't want to go too deep into it now.

service/dap/server.go
```diff
@@ -1814,26 +1815,40 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
...
+	val, _ := s.convertVariableWithOpts(v, "", skipRef)
 	return val
 }
 
+// defaultMaxValueLen is the max length of a string representation of a compound or reference
+// type variable.
+const defaultMaxValueLen = 1 << 8 // 256
+
+// Flags for convertVariableWithOpts option.
+const (
```
aarzilli, line 1831: I'd make these of a custom type: `type convertVariableFlags uint8`

hyangah:
@polinasok is this 256 limit ok? Or should we look into further to do better than SinglelineString and avoid truncating informational messages (e.g. failed load, or partial load...)?

aarzilli (approved):
LGTM, needs decision over the length constant.

suzmue:
LGTM

derekparker (approved):
LGTM

I'll wait to merge until the length is sorted out and we get an ack from @polinasok.

service/dap/server.go
```diff
@@ -2019,7 +2040,8 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 			return
 		}
 
-		if exprVar.Kind == reflect.String && (request.Arguments.Context == "repl" || request.Arguments.Context == "variables") {
+		ctxt := request.Arguments.Context
+		if exprVar.Kind == reflect.String && (ctxt == "repl" || ctxt == "variables" || ctxt == "hover" || ctxt == "clipboard") {
```
polinasok, line 2044: switch-case with multiple values for context might be easier to read here

polinasok (commented):
Could you remind me please what actions in the vscode UI generate evaluate requests with "hover" and "clipboard" context. I know that right click + "Copy to Value" is "variabels" and using Debug Console is "repl".

service/dap/server.go
```diff
@@ -2031,9 +2053,11 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 				}
 			}
 		}
-		// TODO(polina): as far as I can tell, evaluateName is ignored by vscode for expression variables.
-		// Should it be skipped altogether for all levels?
-		exprVal, exprRef := s.convertVariable(exprVar, fmt.Sprintf("(%s)", request.Arguments.Expression))
+		var opts convertVariableFlags
+		if ctxt == "variables" || ctxt == "hover" || ctxt == "clipboard" {
```
polinasok, line 2057: Sadly "Copy Value" from WATCH appears to result in "watch" context request.

polinasok (commented):
The PR description needs updating.

suzmue:
When the server declares the capability `supportsClipboardContext`, `Copy Value` from both the variable pane and WATCH use the clipboard context.

service/dap/server.go
suzmue: Are you still getting this when running a dap server with supportsClipboardContext?

hyangah:
@polinasok It looks like you accidentally edited [my comment](https://github.com/go-delve/delve/pull/2513#issuecomment-850876556) instead of responding to it.

> I am not sure what you mean by "truncating information messages".

- information other than the variable's value such as len/cap of slice/array, or type information or address expression
- information that indicates the loading states, such as " ... 2 more", "- FAILED TO LOAD: some error", ...

> Could you remind me please what actions in the vscode UI generate evaluate requests with "hover" and "clipboard" context. I know that right click + "Copy to Value" is "variabels" and using Debug Console is "repl". 

"hover" -> when you hover over the variable from the editor.
"clipboard" -> "Copy Value" context menu from VARIABLES or WATCH view

hyangah:
PR description is updated. Seems like we all agreed on 256 for now.
@suzmue has a prototype of removing type info and slice/array's len/cap info from the value presentation and making them available in a different way (type info is already supplied as the separate type attribute)

polinasok:
> "hover" -> when you hover over the variable from the editor.

I believe this is only true when you hover over the code, not in the Variables section.

> @suzmue has a prototype of removing type info and slice/array's len/cap info from the value presentation 
Should we consider smaller limit once the type info is dropped from the inlined value?

polinasok:
@hyangah Oh and sorry for editing the comment instead of replying.

hyangah:
>> "hover" -> when you hover over the variable from the editor.
>
> I believe this is only true when you hover over the code, not in the Variables section.

fyi @polinasok 'editor' refers to the main area where code/document is presented in vscode. 'views' or 'viewlets' are the sections presented in the side bar area.

PR #2514, opened 2021-05-26
dap: clean output executable name for windows

hyangah:
If the output binary name does not end with .exe, it can't run on Windows
in noDebug mode. If user-provided output name doesn't have file extension
name (.exe), append it.

Fixes golang/vscode-go#1473


hyangah:
`pkg/proc_test.TestFrameEvaluation` test on windows/amd64 failed. 
Looks unrelated.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM - re-running CI as sanity check.

PR #2516, opened 2021-05-27
gdbserial: actually fix ZMM register support

aarzilli:
PR #2498 / commit d552b338 left out some code due to a clerical error.

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -1859,8 +1859,11 @@ func (t *gdbThread) SetReg(regNum uint64, reg *op.DwarfRegister) error {
 	gdbreg, ok := t.regs.regs[regName]
 	if !ok && strings.HasPrefix(regName, "xmm") {
 		// XMMn and YMMn are the same amd64 register (in different sizes), if we
-		// don't find XMMn try YMMn instead.
+		// don't find XMMn try YMMn or ZMMn instead.
 		gdbreg, ok = t.regs.regs["y"+regName[1:]]
+		if !ok {
+			gdbreg, ok = t.regs.regs["z"+regName[1:]]
+		}
 	}
 	if !ok {
 		return fmt.Errorf("could not set register %s: not found", regName)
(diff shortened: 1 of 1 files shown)


aarzilli:
cc @vsivsi

derekparker:
> PR #2498 / commit [d552b33](https://github.com/go-delve/delve/commit/d552b33822c7d913110789a4a4ca282a4c5c93c6) left out some code due to a clerical error.

😆

derekparker:
I'll wait for @vsivsi to ack the fix.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1919,8 +1922,16 @@ func (regs *gdbRegisters) Slice(floatingPoint bool) ([]proc.Register, error) {
...
-			r = proc.AppendBytesRegister(r, strings.ToUpper(reginfo.Name), value[16:])
+			r = proc.AppendBytesRegister(r, strings.ToUpper(xmmName), value)
+
+		case reginfo.Bitsize == 512:
+			if !strings.HasPrefix(strings.ToLower(reginfo.Name), "zmm") || !floatingPoint {
+				continue
+			}
+
+			value := regs.regs[reginfo.Name].value
+			xmmName := "x" + reginfo.Name[1:]
```
derekparker, line 1933: Shouldn't this technically be `"z" + reginfo.Name[1:]` since we're referring to the full 512-bit width register? Similarly shouldn't above be `"y" + ...` and the one even before be `"x" + ...`? 

xmm refers to lower 128-bits, ymm refers to the 256-bit, and zmm is for AVX-512 referring to the full 512-bit width of the register, correct?

aarzilli: DWARF doesn't have separate register numbers for XMMx, YMMx and ZMMx, they all map to XMM. We could change amd64DwarfToName to accept all three variants, but it isn't worth it, and the correct name gets assigned by formatSSEReg later on anyway. This is just an internal thing.

derekparker: Ack, just wanted to make sure. Thanks.

derekparker:
Haven't heard back from @vsivsi but since this was just code left out of original PR anyways I'm going to merge it.

derekparker (approved):
LGTM

vsivsi:
@derekparker  Apologies, I lost track of this one for some reason. I just quickly tested out the current master commit https://github.com/go-delve/delve/commit/d9ba60dcfe6aee5998eada4c52def680b123b090 and it passes my smoke test.  Thanks!

PR #2517, opened 2021-05-29
proc: when converting registers to slices set Base address

aarzilli:
If the base address isn't set then indexing and slicing will not work.

Large floating point registers already had the base set but small
general purpose registers did not.


derekparker (approved):
LGTM

Issue #279, opened 2015-10-18
Add support for watchpoints


derekparker:
Add support for hardware / software watchpoints. Initially should only support watching a variable, but in the future should be expanded to support expressions.
- Watch a variable and break when the program reads it.
- Watch a variable and break when program writes it.
- Combination of both above.

mattkelly:
Any update on this? Is there a different thread I can follow along with?

glycerine:
Watchpoints are the reason I use a debugger at all. They are invaluable. If possible, please prioritize them.

cloehle:
Especially setting watchpoints on channels would be really useful to follow execution flow.

chainhelen:
I have a question here: are `hardware watchpoints` realized by CPU debug registers DR0 ~ DR3 on amd64?

aarzilli:
> Especially setting watchpoints on channels would be really useful to follow execution flow.

Writing to a channel is actually converted into a function call, so it would just be a regular conditional breakpoint on runtime.chansend/runtime.chanrecv (however it is currently impossible to write the appropriate condition).

> I have a question here: are hardware watchpoints realized by CPU debug registers DR0 ~ DR3 on amd64?

Yes, however things are complicated substantially by stack resize operations.

baiwfg2:
Hi. Is this being working on ? Right now I've found during VSCode + dlv debugging, I can see the watched variables' value, but can't see the program is stopped when the watched variables are changed, which seems not to match `watch` semantics. Better to behave like gdb watch

aarzilli:
Support for watchpoints was added to the extent that it is doable.

PR #2519, opened 2021-05-30
proc: allow multiple overlapping internal breakpoints

aarzilli:
Changes Breakpoint to allow multiple overlapping internal breakpoints
on the same instruction address.

This is done by changing the Breakpoint structure to contain a list of
"breaklets", each breaklet has a BreakpointKind and a condition
expression, independent of the other.

A breakpoint is considered active if any of its breaklets are active.

A breakpoint is removed when all its breaklets are removed.

We also change the terminology "internal breakpoint" to "stepping
breakpoint":

HasInternalBreakpoints -> HasSteppingBreakpoints
IsInternal -> IsStepping
etc...

The motivation for this change is implementing watchpoints on stack
variables.

Watching a stack variable requires also setting a special breakpoint to
find out when the variable goes out of scope. These breakpoints can not
be UserBreakpoints because only one user breakpoint is allowed on the
same instruction and they can not be internal breakpoints because they
should not be cleared when a next operation is completed (they should
be cleared when the variable watch is cleared).

Updates #279


derekparker (approved):
LGTM

derekparker:
I like the new distinction of "stepping" breakpoints instead of just "internal". Also, I think having the "breaklets" cleans things up a lot even without the addition of watchpoints.

derekparker:
All CI failures are known and unrelated.

PR #2521, opened 2021-05-31
proc,terminal,service: support stack watchpoints

aarzilli:
### proc,terminal,service: support stack watchpoints

Adds support for watchpoints on stack allocated variables.

When a stack variable is watched, in addition to the normal watchpoint
some support breakpoints are created:

* one breakpoint inside runtime.copystack, used to adjust the address
of the watchpoint when the stack is resized
* one or more breakpoints used to detect when the stack variable goes
out of scope, those are similar to the breakpoints set by StepOut.

Implements #279

### terminal,service: add way to see internal breakpoints

Now that Delve has internal breakpoints that survive for long periods
of time it will be useful to have an option to display them.

### proc: allow multiple overlapping internal breakpoints

Changes Breakpoint to allow multiple overlapping internal breakpoints
on the same instruction address.

This is done by changing the Breakpoint structure to contain a list of
"breaklets", each breaklet has a BreakpointKind and a condition
expression, independent of the other.

A breakpoint is considered active if any of its breaklets are active.

A breakpoint is removed when all its breaklets are removed.

We also change the terminology "internal breakpoint" to "stepping
breakpoint":

HasInternalBreakpoints -> HasSteppingBreakpoints
IsInternal -> IsStepping
etc...

The motivation for this change is implementing watchpoints on stack
variables.

Watching a stack variable requires also setting a special breakpoint to
find out when the variable goes out of scope. These breakpoints can not
be UserBreakpoints because only one user breakpoint is allowed on the
same instruction and they can not be internal breakpoints because they
should not be cleared when a next operation is completed (they should
be cleared when the variable watch is cleared).

Updates #279

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -1,31 +1,31 @@
 Tests skipped by each supported backend:
 
-* 386 skipped = 6
+* 386 skipped = 7
 	* 1 broken
 	* 3 broken - cgo stacktraces
-	* 2 not implemented
-* arm64 skipped = 4
+	* 3 not implemented
+* arm64 skipped = 5
 	* 1 broken
 	* 1 broken - global variable symbolication
-	* 2 not implemented
-* darwin skipped = 2
-	* 2 not implemented
+	* 3 not implemented
+* darwin skipped = 3
+	* 3 not implemented
 * darwin/arm64 skipped = 1
 	* 1 broken - cgo stacktraces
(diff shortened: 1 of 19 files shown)


aarzilli:
Depends on #2519

derekparker:
Needs rebase now that the overlapping breakpoints PR has landed.

aarzilli:
Rebased.

derekparker (approved):
LGTM

PR #2522, opened 2021-06-02
service/dap: deemphasize internal runtime stack frames

suzmue:
Apply a presentation hint to the internal runtime stack frames, so that these can be deemphasized in the UI. This should allow
users to more easily inspect their own code, and will keep the option to view those frames if they choose to.


suzmue:
Before:
Screen Shot 2021-06-02 at 10 26 01 PM
Screen Shot 2021-06-02 at 10 26 04 PM
After:
Screen Shot 2021-06-02 at 10 27 15 PM
Screen Shot 2021-06-02 at 10 27 07 PM

pkg/terminal/command.go
```diff
@@ -2353,13 +2353,17 @@ func PrintStack(formatPath func(string) string, out io.Writer, stack []api.Stack
...
 	for i := range stack {
+		if !filter(stack[i]) {
+			continue
+		}
 		if stack[i].Err != "" {
 			fmt.Fprintf(out, "%serror: %s\n", s, stack[i].Err)
 			continue
 		}
-		fmt.Fprintf(out, fmtstr, ind, i, stack[i].PC, stack[i].Function.Name())
+		fmt.Fprintf(out, fmtstr, ind, printed, stack[i].PC, stack[i].Function.Name())
```
aarzilli, line 2366: Wouldn't it be better to leave the stack frame number unaltered? So that it's going to be obvious that something was elided.

service/dap/server.go
```diff
@@ -1534,6 +1542,16 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 			stackFrames[i].Source = dap.Source{Name: filepath.Base(clientPath), Path: clientPath}
 		}
 		stackFrames[i].Column = 0
+
+		g, err := s.debugger.FindGoroutine(goroutineID)
```
aarzilli, line 1546: You shouldn't call this in a loop.

service/dap/server.go
```diff
@@ -2375,7 +2393,13 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 		if err == nil {
 			var buf bytes.Buffer
 			fmt.Fprintln(&buf, "Stack:")
-			terminal.PrintStack(s.toClientPath, &buf, apiFrames, "\t", false)
+			userLoc := g.UserCurrent()
+			userFuncPkg := fnPackageName(&userLoc)
+			terminal.PrintStack(s.toClientPath, &buf, apiFrames, "\t", false, func(s api.Stackframe) bool {
+				// Include all stack frames if the stack trace is for a system goroutine,
+				// otherwise, skip runtime stack frames.
+				return userFuncPkg == "runtime" || !strings.HasPrefix(s.Location.Function.Name(), "runtime")
```
aarzilli, line 2401: s.Location.Function could be nil. Also it should be `"runtime."` so that it doesn't elide cgo frames.

pkg/terminal/command.go
polinasok: +1

pkg/terminal/command.go
```diff
@@ -2353,13 +2353,17 @@ func PrintStack(formatPath func(string) string, out io.Writer, stack []api.Stack
 	d := digits(len(stack) - 1)
 	fmtstr := "%s%" + strconv.Itoa(d) + "d  0x%016x in %s\n"
 	s := ind + strings.Repeat(" ", d+2+len(ind))
+	printed := 0
 
 	for i := range stack {
+		if !filter(stack[i]) {
```
polinasok, line 2359: Are we filtering in or out? I would suggest names like include/exclude to make this clearer.

service/dap/server.go
```diff
@@ -1534,6 +1542,16 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 			stackFrames[i].Source = dap.Source{Name: filepath.Base(clientPath), Path: clientPath}
 		}
 		stackFrames[i].Column = 0
+
+		g, err := s.debugger.FindGoroutine(goroutineID)
+		if err == nil {
+			userLoc := g.UserCurrent()
+			userLocPackageName := fnPackageName(&userLoc)
+			packageName := fnPackageName(loc)
+			if userLocPackageName != "runtime" && packageName == "runtime" {
```
polinasok, line 1551: Perhaps this deserves a helper? Also, how does this relate to System() from #2504?

service/dap/server_test.go
```diff
@@ -3270,6 +3270,19 @@ func TestPanicBreakpointOnContinue(t *testing.T) {
 					if eInfo.Body.ExceptionId != "panic" || eInfo.Body.Description != "\"BOOM!\"" {
 						t.Errorf("\ngot  %#v\nwant ExceptionId=\"panic\" Description=\"\"BOOM!\"\"", eInfo)
 					}
+
+					client.StackTraceRequest(se.Body.ThreadId, 0, 20)
+					st := client.ExpectStackTraceResponse(t)
+					for i, frame := range st.Body.StackFrames {
+						if strings.HasPrefix(frame.Name, "runtime") {
```
polinasok, line 3277: runtime. (based on above comment)

polinasok (changes requested):

service/dap/server.go
suzmue: This is trying to extract that same information. I can leave a TODO to use the check from that change.

service/dap/server_test.go

suzmue:
Here is a screenshot with the deemphasized stack frames and using `i` as the index.
Screen Shot 2021-06-07 at 5 23 11 PM

service/dap/server.go
suzmue: done.

service/dap/server.go
suzmue: Moved it out of the loop.

pkg/terminal/command.go
suzmue: Yeah this makes sense, especially since they are still in the call stack frame at this time.

aarzilli (approved):

polinasok (approved):

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -1524,6 +1532,15 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 		return
 	}
 
+	// Determine if the goroutine is a system goroutine.
+	// TODO(suzmue): Use the System() method defined in: https://github.com/go-delve/delve/pull/2504
```
polinasok, line 1536: Now that this is merged, #2504 should address this TODO

ansiwen:
I had to switch to legacy mode because of this change for stepping through runtime code, because vscode always jumps out of the deemphasized code into the user code after each step. I couldn't find a way to tell vscode to ignore the deemphasizing.

polinasok:
Thank you for the input @ansiwen and for filing the vscode-go issue. This is exactly the kind of feedback we are looking for. We will be removing this feature. Please stay tuned.

PR #2524, opened 2021-06-02
service/dap: add throw reason to exception info

suzmue:
We can get the throw reason by looking at the argument "s" in runtime.throw. This is not currently working in Go 1.16 or Go 1.17 (see golang/go#46425), but does work in Go 1.15 and Go 1.14

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2366,7 +2366,9 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 			// Attempt to get the value of the throw reason.
 			// This is not currently working for Go 1.16 or 1.17: https://github.com/golang/go/issues/46425.
 			handleError := func(err error) {
-				body.Description = fmt.Sprintf("Error getting throw reason: %s", err.Error())
+				if err != nil {
+					body.Description = fmt.Sprintf("Error getting throw reason: %s", err.Error())
+				}
 				if goversion.VersionAfterOrEqual(s.debugger.TargetGoVersion(), 1, 16) {
 					body.Description = "Throw reason unavailable, see https://github.com/golang/go/issues/46425"
 				}
@@ -2376,7 +2378,7 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 			if err == nil {
 				if exprVar.Value != nil {
 					body.Description = exprVar.Value.String()
-				} else if exprVar.Unreadable != nil {
+				} else {
 					handleError(exprVar.Unreadable)
 				}
 			} else {
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2368,7 +2368,7 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 			exprVar, err := s.debugger.EvalVariableInScope(goroutineID, 1, 0, "s", DefaultLoadConfig)
 			if err == nil {
 				body.Description = exprVar.Value.String()
-			} else if goversion.VersionAfterOrEqual(runtime.Version(), 1, 16) {
+			} else if goversion.VersionAfterOrEqual(s.debugger.TargetGoVersion(), 1, 16) {
 				body.Description = "Throw reason unavailable, see https://github.com/golang/go/issues/46425"
 			} else {
 				body.Description = fmt.Sprintf("Error getting throw reason: %s", err.Error())
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 
 	"github.com/go-delve/delve/pkg/gobuild"
+	"github.com/go-delve/delve/pkg/goversion"
 	"github.com/go-delve/delve/pkg/locspec"
 	"github.com/go-delve/delve/pkg/logflags"
 	"github.com/go-delve/delve/pkg/proc"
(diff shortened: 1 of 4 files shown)


suzmue:
Screen shot of fatal throw reason:
Screen Shot 2021-06-02 at 5 57 55 PM

service/dap/server.go
```diff
@@ -2336,6 +2336,12 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 		case proc.FatalThrow:
 			// TODO(suzmue): add the fatal throw reason to body.Description.
```
aarzilli, line 2337: This comment can be removed now, right?

suzmue: Yes, thanks for catching this.

service/dap/server_test.go
```diff
@@ -3335,6 +3334,9 @@ func TestFatalThrowBreakpoint(t *testing.T) {
 					if se.Body.Reason != "exception" || se.Body.Description != "Paused on fatal error" {
 						t.Errorf("\ngot  %#v\nwant Reason=\"exception\" Description=\"Paused on fatal error\"", se)
 					}
+
+					// TODO(suzmue): Get the exception info for the thread and check the description
+					// includes "all goroutines are asleep - deadlock!".
```
derekparker, line 3339: You should be able to test this, just gate the assertion on the Go compiler being version 1.14 or 1.15 (which we still test via CI).

derekparker (changes requested):

service/dap/server_test.go
suzmue: The actual issue that is breaking this test is that the deadlock is not associated with a goroutine ID, so we can't request the exception info. This needs to be fixed too, but makes sense to do separately because we will likely want to add the thread that is running to display the stack trace etc.

aarzilli (approved):

service/dap/server_test.go
polinasok: What about _fixtures/fatalerror.go? Does that one result in a selected goroutine, so you have an id linked to a description that you can check?

service/dap/server.go
```diff
@@ -2334,8 +2334,13 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 	if bpState != nil && bpState.Breakpoint != nil && (bpState.Breakpoint.Name == proc.FatalThrow || bpState.Breakpoint.Name == proc.UnrecoveredPanic) {
 		switch bpState.Breakpoint.Name {
 		case proc.FatalThrow:
-			// TODO(suzmue): add the fatal throw reason to body.Description.
 			body.ExceptionId = "fatal error"
+			// Attempt to get the value of the throw reason.
+			// This is not currently working for Go 1.16 or 1.17: https://github.com/golang/go/issues/46425.
```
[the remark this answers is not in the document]
suzmue: That's a good idea! I will add that as the reason if its on Go 1.16 or Go 1.17.

I have been considering these additional exception info descriptions best effort, if they don't work, we just don't include them. I think it would be good now to include these extra info messages so that it can help us know if it is no longer working easier.

service/dap/server_test.go
suzmue: I don't see that fixture. I could add a fixture to do a concurrent map read / write and I could test it that way.

service/dap/server.go
```diff
@@ -2361,14 +2362,25 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
...
+			exprVar, err := s.debugger.EvalVariableInScope(goroutineID, 1, 0, "s", DefaultLoadConfig)
+			if err == nil {
+				body.Description = exprVar.Value.String()
+			} else if goversion.VersionAfterOrEqual(runtime.Version(), 1, 16) {
```
aarzilli, line 2371: I think you have to check the version of the target executable, not of delve, see `proc.BinaryInfo.Producer`.

service/dap/server_test.go
polinasok: Huh! That's because it was in my local client. I was going to add it at some point, but never did.  This is easier than concurrent map operations:
```
func main() {
    var f func()
    go f()
}
```

suzmue: Added the fixture and the test

aarzilli:
See test failures.

polinasok (approved):

service/dap/server.go
```diff
@@ -2361,14 +2362,28 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
 	if bpState != nil && bpState.Breakpoint != nil && (bpState.Breakpoint.Name == proc.FatalThrow || bpState.Breakpoint.Name == proc.UnrecoveredPanic) {
 		switch bpState.Breakpoint.Name {
 		case proc.FatalThrow:
-			// TODO(suzmue): add the fatal throw reason to body.Description.
 			body.ExceptionId = "fatal error"
+			// Attempt to get the value of the throw reason.
+			// This is not currently working for Go 1.16 or 1.17: https://github.com/golang/go/issues/46425.
+			exprVar, err := s.debugger.EvalVariableInScope(goroutineID, 1, 0, "s", DefaultLoadConfig)
+			if err == nil {
```
aarzilli, line 2369: I think you need to check exprVar.Unreadable also, see the test failures on 1.16.

suzmue: Done.

service/dap/server_test.go
suzmue: The exception info is shown in the source code, so if its not associated with a thread with a stack trace, I think there would need to be a separate UI for it.

polinasok:
@suzmue Is this PR ready for @derekparker to take another look for final approval?

polinasok:
DeepSource nag sounds reasonable. Probably a good idea to fix it.

suzmue:
> @suzmue Is this PR ready for @derekparker to take another look for final approval?

Yes I think its ready.

> DeepSource nag sounds reasonable. Probably a good idea to fix it.

This is unrelated to this change, so I will just send a quick follow up PR to fix it.

derekparker (approved):
LGTM

PR #2527, opened 2021-06-04
proc: give unique addresses to registerized variables

aarzilli:
We told clients that further loading of variables can be done by
specifying a type cast using the address of a variable that we
returned.

This does not work for registerized variables (or, in general,
variables that have a complex location expression) because we don't
give them unique addresses and we throw away the compositeMemory object
we made to read them.

This commit changes proc so that:

1. variables with location expression divided in pieces do get a unique
memory address
2. the compositeMemory object is saved somewhere
3. when an integer is cast back into a pointer type we look through our
saved compositeMemory objects to see if there is one that covers the
specified address and use it.

The unique memory addresses we generate have the MSB set to 1, as
specified by the Intel 86x64 manual addresses in this form are reserved
for kernel memory (which we can not read anyway) so we are guaranteed
to never generate a fake memory address that overlaps a real memory
address of the application.

The unfortunate side effect of this is that it will break clients that
do not deserialize the address to a 64bit integer. This practice is
contrary to how we defined our types and contrary to the specification
of the JSON format, as of json.org, however it is also fairly common,
due to javascript itself having only 53bit integers.

We could come up with a new mechanism but then even more old clients
would have to be changed.


hyangah:
This is a nice feature if works reliably. Can we safely assume the same mechanism would work in various arch/os?
If API 3 is ever a possibility :-), can we consider string type for address (DAP uses string)?

aarzilli:
> Can we safely assume the same mechanism would work in various arch/os?

AFAIK this is fine everywhere

>  If API 3 is ever a possibility :-), can we consider string type for address (DAP uses string)?

Representing an address as a string would go against everything I believe in.

polinasok:
> > If API 3 is ever a possibility :-), can we consider string type for address (DAP uses string)?
> 
> Representing an address as a string would go against everything I believe in.

@hyangah Is this what you were referring to?

From [dap spec](https://microsoft.github.io/debug-adapter-protocol/specification):
```/**
     * Optional memory reference to a location appropriate for this result.
     * For pointer type eval results, this is generally a reference to the
     * memory address contained in the pointer.
     * This attribute should be returned by a debug adapter if the client has
     * passed the value true for the 'supportsMemoryReferences' capability of
     * the 'initialize' request.
     */
    memoryReference?: string;
```

From C logs:
```
"memoryReference":"0x00007ffeefbfee90"
```

hyangah:
@polinasok And `ReadMemoryResponse`, `DisassembledInstruction`, ..  many address related stuffs are defined as string, not number.

polinasok:
Is it possible to share an example with dlv cli, where this did not work before and does work now for illustration purposes?

I am also curious what impact it could have on the end user (as opposed to the client developer). 
Do we expect them to use the addresses as well?
We recently realized that we don't display addresses for the user with dlv-dap:
`&myint` prints `*0` instead of ```(*int)(0xc000016078)```

aarzilli:
See TestLongStringArg and it's fixture. Evaluating `(*(*string)(address_of_s))[64:]` will not load more of 's' on Go 1.17 because 's' is in a register and doesn't have a real address. It can also happen < 1.17 but only on optimized binaries and there's no way to tell in advance which variables will be affected.
This doesn't affect the cli much because you would always just use the variable name rather than its address with a typecast.

Jccp1983 (commented):
Read Me

pkg/proc/bininfo.go
```diff
@@ -2196,6 +2202,76 @@ func (bi *BinaryInfo) ListPackagesBuildInfo(includeFiles bool) []*PackageBuildIn
...
+	fakeAddressUnresolv = 0xbeed000000000000 // this address never resloves to memory
+)
+
+// newCompositeMemory creates a new compositeMemory object and registers it.
+// If the same composite memory has been created before it will return a
+// cached object.
+// This caching is primarily done so that registerized variables don't get a
+// different address every time they are evaluated, which would be confusing
+// and leak memory.
+func (bi *BinaryInfo) newCompositeMemory(mem MemoryReadWriter, regs op.DwarfRegisters, pieces []op.Piece, descr *locationExpr) (int64, *compositeMemory, error) {
```
derekparker, line 2216: I think this should live on `Target` or stay within `pkg/proc/mem.go`. `BinaryInfo` is intended to contain information we've parsed from the binary on disk, memory is more of a running process concern.

pkg/proc/mem.go
```diff
@@ -79,8 +79,6 @@ func cacheMemory(mem MemoryReadWriter, addr uint64, size int) MemoryReadWriter {
 // fakeAddress used by extractVarInfoFromEntry for variables that do not
```
derekparker, line 79: You should take this comment away now.

pkg/proc/target.go
```diff
@@ -232,9 +232,10 @@ func (t *Target) SupportsFunctionCalls() bool {
 	return t.Process.BinInfo().Arch.Name == "amd64"
 }
 
-// ClearAllGCache clears the internal Goroutine cache.
+// ClearStopCaches clears internal caches that should not survive a restart.
 // This should be called anytime the target process executes instructions.
-func (t *Target) ClearAllGCache() {
+func (t *Target) ClearStopCaches() {
```
derekparker, line 237: nit: I think just `ClearCaches` is ok here.

derekparker (changes requested):

pkg/proc/bininfo.go
aarzilli: Done. The outcome is less than ideal since that code path ends up getting called by the breakpoint setting code, which does not have access to the Target object.

pkg/proc/variables.go
```diff
@@ -1111,7 +1111,7 @@ func readVarEntry(entry *godwarf.Tree, image *Image) (name string, typ godwarf.T
 
 // Extracts the name and type of a variable from a dwarf entry
 // then executes the instructions given in the  DW_AT_location attribute to grab the variable's address
-func extractVarInfoFromEntry(bi *BinaryInfo, image *Image, regs op.DwarfRegisters, mem MemoryReadWriter, entry *godwarf.Tree) (*Variable, error) {
+func extractVarInfoFromEntry(tgt *Target, bi *BinaryInfo, image *Image, regs op.DwarfRegisters, mem MemoryReadWriter, entry *godwarf.Tree) (*Variable, error) {
```
derekparker, line 1114: We can remove the `bi *BinaryInfo` argument here now that we're passing in `Target`.

aarzilli: We can't, at the moment. When it gets called to evaluate a breakpoint condition tgt will be nil because that code doesn't have access to a target object. I was thinking of moving the breakpoint condition evaluation code into `(*Target).Continue` but it's too late in the release cycle and it overlaps too much with #2519. I think we have to live with this uglyness for a while.

derekparker: Ack, alright. I'm on board with shifting around the breakpoint condition eval code as well.

derekparker (approved):
LGTM

derekparker:
Just needs rebase, then we can rebase the 1.17 support PR and I will start reviewing that in more depth.

PR #2529, opened 2021-06-07
service/dap: truncate value for expandable types in hover

suzmue:
On hovers, including the full value for complex types is unnecessary, since the user will likely need to expand the children to inspect the values. Additionally, this can really slow down getting the hover values. This change will allow truncating the value
in hover if the variable has a variables reference.

Updates golang/vscode-go#1435

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1832,10 +1832,7 @@ type convertVariableFlags uint8
 
 const (
 	skipRef convertVariableFlags = 1 << iota
-	// A variable is considered basic if it cannot be expanded to get additional
-	// values.
-	showFullValueBasic
-	showFullValueAll
+	showFullValue
 )
 
 // convertVariableWithOpts allows to skip reference generation in case all we need is
@@ -1982,7 +1979,10 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 			variablesReference = maybeCreateVariableHandle(v)
 		}
 	}
-	canTruncateValue := showFullValueAll&opts == 0 && (showFullValueBasic&opts == 0 || variablesReference != 0)
+
+	// By default, only values of variables that have children can be truncated.
+	// If showFullValue is set, then all value strings are not truncated.
+	canTruncateValue := showFullValue&opts == 0
 	if len(value) > defaultMaxValueLen && canTruncateValue && canHaveRef {
 		value = value[:defaultMaxValueLen] + "..."
 	}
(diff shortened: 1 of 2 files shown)


suzmue:
Hover over `x.s` (truncates the value of hover):
Screen Shot 2021-06-07 at 6 27 27 PM

Hover over `s` (stays the same as currently):
Screen Shot 2021-06-07 at 6 27 21 PM

service/dap/server.go
```diff
@@ -1979,7 +1982,7 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 			variablesReference = maybeCreateVariableHandle(v)
 		}
 	}
-	canTruncateValue := showFullValue&opts == 0
+	canTruncateValue := showFullValueAll&opts == 0 && (showFullValueBasic&opts == 0 || variablesReference != 0)
```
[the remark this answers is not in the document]
polinasok: noChildren would be even more meaningful than no ref.
And there seems to be the does-have-ref and can-have-ref nuance here.
This is indeed hard to follow.
Also it is confusing that we  don't check if we even need to truncate (i.e. value is indeed to long) before going through all the trouble.

service/dap/server.go
```diff
@@ -2057,8 +2060,11 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 			}
 		}
 		var opts convertVariableFlags
-		if ctxt == "variables" || ctxt == "hover" || ctxt == "clipboard" {
-			opts |= showFullValue
+		if ctxt == "clipboard" || ctxt == "variables" {
```
polinasok, line 2063: With clipboard option on, when do we get "variables"? Is this in case other editors don't use clipboard?

service/dap/server.go
```diff
@@ -1832,7 +1832,10 @@ type convertVariableFlags uint8
 
 const (
 	skipRef convertVariableFlags = 1 << iota
-	showFullValue
+	// A variable is considered basic if it cannot be expanded to get additional
+	// values.
+	showFullValueBasic
+	showFullValueAll
```
polinasok, line 1838: how about:
showFullValue and showFullValueIfNoChildren

polinasok (changes requested):

service/dap/server.go
suzmue: This is to support clients that do not use clipboard context

service/dap/server.go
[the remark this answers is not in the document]
suzmue: With `canHaveRef` helping decide if the value can be truncated, the default behavior is to only truncate if it has children.

I will simplify this code and add a comment describing this.

service/dap/server.go
suzmue: Not applicable anymore.

aarzilli (approved):

service/dap/server.go
```diff
@@ -2061,10 +2061,7 @@ func (s *Server) onEvaluateRequest(request *dap.EvaluateRequest) {
 		}
 		var opts convertVariableFlags
 		if ctxt == "clipboard" || ctxt == "variables" {
```
polinasok, line 2063: Let's add a comment that these contexts are usually used to copy the value, which is why we want it in full.

polinasok (approved):

service/dap/server.go
derekparker: I'll wait to merge until this is addressed.

derekparker (approved):
LGTM

PR #2370, opened 2021-03-05
service: serialize calls to Command API

aarzilli:
Wait until the target process has resumed before accepting new calls to
Command. Before this if a 'continue' was immediately followed by a
'halt' the 'halt' could be processed before the 'continue'.

Fixes #1608
Fixes #2216

Issue #2317, opened 2021-01-21
pkg/terminal: add option to disable auto-continue during next/step/stepout


hzzb:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
```
localhost:~ xxx$ dlv version
Delve Debugger
Version: 1.2.0
Build: 86d4b0b9915d55833a99bfb4c8b3a138f03ab802
localhost:~ xxx$ 
```

2. What version of Go are you using? (`go version`)?
```
localhost:~ xxx$ go version
go version go1.12.8 darwin/amd64
localhost:~ xxx$ 
```

3. What operating system and processor architecture are you using?
```
localhost:~ xxx$ uname -a
Darwin Hanss-MacBook-Pro.local 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64
localhost:~ xxx$ sysctl -a | grep machdep.cpu 
machdep.cpu.max_basic: 20
machdep.cpu.max_ext: 2147483656
machdep.cpu.vendor: GenuineIntel
machdep.cpu.brand_string: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
machdep.cpu.family: 6
machdep.cpu.model: 61
localhost:~ xxx$ 
```

4. What did you do?
```Golang
// file main.go
package main

import (
	"time"
)

func main() {
	go b()
	time.Sleep(time.Second)
	time.Sleep(time.Second)
}

func b() {
	time.Sleep(time.Second)    // breakpoint  2
}
```

```
localhost:~ xxx$ dlv debug main.go
Type 'help' for list of commands.
(dlv) b main.main
Breakpoint 1 set at 0x105882f for main.main() ./main.go:7
(dlv) b main.b
Breakpoint 2 set at 0x105888f for main.b() ./main.go:13
(dlv) c
(dlv) n
```
In main func, repeat `n` command.

5. What did you expect to see?
After some one particular `n` command, dlv stops at breakpoint 2.

6. What did you see instead?

The dlv debugger didn't catch breakpoint 2.

Will dlv switch goroutine automatically to catch breakpoints on different goroutines?   I found gdb works as expected.

aarzilli:
[...]

PR #2530, opened 2021-06-07
service/dap: complete step requests by skipping bps in other goroutines

suzmue:
Other goroutines may encounter breakpoints during a step request on
a certain goroutine. When this happens, state.NextInProgress is set
to true. If another step request is issued, then we get a "next while
nexting" error from the debugger. Instead we have two options:

1. skip all breakpoints and send output events to notify the user.
2. cancel next and notify user when the breakpoint is hit.

The terminal package implements (1), and this makes sense for dlv dap
as well because when users hit next, they are choosing to focus on a
particular goroutine. In order to get the behavior where they stop when
a breakpoint is hit, they can simply add a breakpoint to the next line,
and then run continue.

"Pause" requests may come in while we are processing the breakpoint that was
hit. If this happens, the pause simply does not take effect and the user will have to
try again. The VS Code UI does NOT update to have the threads be stopped once
it receives the "pause" response.

If a set breakpoints request is being processed concurrently with the stopped event,
then we want to make sure that we do not resume execution before it has finished.
This also applies to any other synchronous requests that expect the state to be
stopped. The continueMu lock is used to protect sections of code that rely on another
goroutine not continuing the program execution.

Updates golang/vscode-go#119


service/dap/server.go
```diff
@@ -378,6 +385,8 @@ func (s *Server) recoverPanic(request dap.Message) {
 
 func (s *Server) handleRequest(request dap.Message) {
 	defer s.recoverPanic(request)
+	s.resumeMu.Lock()
```
polinasok, line 388: The PR comment seems to indicate that you don't want to interrupt finishing next with a halt. But this would block all other requests as well, including disconnect.

service/dap/server.go
```diff
@@ -2490,6 +2508,35 @@ func (s *Server) resetHandlesForStoppedEvent() {
...
+	defer s.asyncCommandDone(resumeRequestLoop)
+
+	go func() {
+		defer s.resumeMu.Unlock()
+		<-resumeRequestLoop
+	}()
+
+	// Check for a manual halt.
+	select {
+	case <-s.overrideStopReason:
```
polinasok, line 2523: It would be more readable/reliable if this was called haltRequested or you explicitly read "manual halt" from the channel instead of relying on the comment and the fact that it's the only value ever sent over the channel.

That said, I am confused as to why this is here. Is this a manual halt that happened while next was going from one breakpoint to another? And why can't the debugger's stop reason be used?

polinasok (changes requested):
I think I will need to go over this one more time. Still a bit confused about the desired synchronization.

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+			var continued bool
+			continued, state, err = s.runContinueCommand()
+			if continued {
+				continue
+			}
+			// Otherwise, a manual halt was attempted while the program was paused, so we should stop.
+			stopReason = proc.StopManual
+		}
+
+		// There was no breakpoint to skip, we should cancel next and process this stop.
```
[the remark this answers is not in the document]
suzmue: I don't know of any scenario where this would be hit, but I figured if it is possible it is better to handle it.

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+			}
+			s.log.Debugf("goroutine %d hit breakpoint (id: %d, loc: %s:%d) during %s\n", id, bp.ID, bp.File, bp.Line, command)
+			s.send(&dap.OutputEvent{
+				Event: *newEvent("output"),
+				Body: dap.OutputEventBody{
+					Output:   fmt.Sprintf("goroutine %d hit breakpoint (id: %d, loc: %s:%d) during %s\n", id, bp.ID, bp.File, bp.Line, command),
+					Category: "console",
+				}})
+			var continued bool
+			continued, state, err = s.runContinueCommand()
```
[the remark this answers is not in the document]
suzmue: I expect this to also be used with log points in the future, so it may be resuming a continue command as well.

service/dap/server.go
suzmue: -> doRunCommand calls and returns from next (stopped on a breakpoint)
-> halt request is issued (debuggee is not running)
-> halt request succeeds
-> doRunCommand sees that it should continue the next request and resumes execution.

The client thinks that the program has halted, but it is actually running.

Perhaps alternatively we make sure to send a continued event so that the client knows execution was resumed? In case it thought it had halted.

That could be a simpler way to deal with this.

service/dap/server.go
```diff
@@ -2490,6 +2508,35 @@ func (s *Server) resetHandlesForStoppedEvent() {
 	s.exceptionErr = nil
 }
 
+func (s *Server) runContinueCommand() (bool, *api.DebuggerState, error) {
+	s.resumeMu.Lock()
+	resumeRequestLoop := make(chan struct{})
```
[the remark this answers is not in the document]
suzmue: Updated the name of the channel to be more descriptive.

service/dap/server.go
```diff
@@ -640,6 +649,15 @@ func (s *Server) handleRequest(request dap.Message) {
 	}
 }
 
+func (s *Server) manualHalt() error {
+	select {
+	case s.overrideStopReason <- "manual halt":
```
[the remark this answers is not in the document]
suzmue: Removed references to manual halt.

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+	for err == nil && !state.Exited && state.CurrentThread != nil && state.NextInProgress {
+		// If there is a NextInProgress, we want to notify the user that a breakpoint
+		// was hit and then continue.
+		if bp := state.CurrentThread.Breakpoint; bp != nil {
+			id := 0
+			if state.SelectedGoroutine != nil {
+				id = state.SelectedGoroutine.ID
+			}
+			s.log.Debugf("goroutine %d hit breakpoint (id: %d, loc: %s:%d) during %s\n", id, bp.ID, bp.File, bp.Line, command)
+			s.send(&dap.OutputEvent{
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+				Event: *newEvent("output"),
+				Body: dap.OutputEventBody{
+					Output:   fmt.Sprintf("goroutine %d hit breakpoint (id: %d, loc: %s:%d) during %s\n", id, bp.ID, bp.File, bp.Line, command),
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
 	// asyncSetupDone (e.g. when having an error next while nexting).
 	// So we should always close it ourselves just in case.
 	defer s.asyncCommandDone(asyncSetupDone)
+	// Clear the manual halt check. handleRequest is blocked until asyncSetupDone
+	// is closed so this will make it empty when we start.
+	s.resetOverrideStopReason()
 	state, err := s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
+	// Make sure that asyncSetupDone is closed, so that handleRequest
+	// will release resumeMu.
+	s.asyncCommandDone(asyncSetupDone)
```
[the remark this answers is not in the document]
suzmue: Done.

service/dap/server.go
suzmue: I updated the description of this PR to describe the issues that we need to guard from.

This change deals with making sure that doRunCommand does not resume execution while the handleRequest goroutine expects the program to be stopped.

polinasok (changes requested):
I am still thinking how to handle this better.

aarzilli:
This seems much too complicated compared to continueUntilCompleteNext, I was thinking more of something along the lines of:

```
func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
	defer s.asyncCommandDone(asyncSetupDone)
        var state api.DebuggerState
        var err error
        for {
		state, err = s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
		if !(state /= nil && state.NextInProgress) {
			break
		}
		asyncSetupDone = nil
		command = api.DirectionCongruentContinue
		// ... log breakpoints hit ...
	}
	// ... rest of doRunCommand, unchanged ...
```

A pause request is going to clear the internal breakpoints anyway, so it doesn't matter. You could also have a flag, protected by a mutex, to make the for loop stop regardless of the return value of `s.debugger.Command`.

service/dap/server.go
polinasok: I understand now that you will only block while auto-continue is starting up, akin to blocking with resumeRequestLoop when starting up async requests.

polinasok:
> This seems much too complicated compared to continueUntilCompleteNext, I was thinking more of something along the lines of:
> 
> ```
> func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
> 	defer s.asyncCommandDone(asyncSetupDone)
>         var state api.DebuggerState
>         var err error
>         for {
> 		state, err = s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
> 		if !(state != nil && state.NextInProgress) {
> 			break
> 		}
> 		asyncSetupDone = nil

I think there are some corner cases in debugger.Command that don't release the channel. That's why the code has defer, which won't help if you reset this.

Also, don't we want the subsequent auto-continue command not to get interrupted by another incoming command prematurely? Isn't this what this relatively new chan arg is for? Added in #2370. 

> 		command = api.DirectionCongruentContinue

I realized that we added a TODO as a result of previous similar discussion on how to auto-continue here that might be useful to revisit:
https://github.com/go-delve/delve/blob/688f94a4f8383e22972df8da8631eb282dac1c8c/service/dap/server.go#L451

> 		// ... log breakpoints hit ...
> 	}
> 	// ... rest of doRunCommand, unchanged ...
> ```

> A pause request is going to clear the internal breakpoints anyway, so it doesn't matter. You could also have a flag, protected by a mutex, to make the for loop stop regardless of the return value of `s.debugger.Command`.

Does pause clear internal breakpoints (and by extension NextInProgress) if it is handled while stopped?
What if the execution is already stopping because of a non-halt reason when halt arrives? Command will return NextInProgress=true and not break out of the loop. Then halt will succeed. Then we will still auto-continue. So if we want to recognize halt reliably, we must track that explicitly (maybe with a chan instead of flag+lock?).

polinasok:
Darn, I just realized that an overview comment that I wrote yesterday that took me forever to express never posted and is not in any of my tabs :( How frustrating! Let me try to recreate:

> -> doRunCommand calls and returns from next (stopped on a breakpoint)
> -> halt request is issued (debuggee is not running)
> -> halt request succeeds
> -> doRunCommand sees that it should continue the next request and resumes execution.
>
> The client thinks that the program has halted, but it is actually running.
>
> Perhaps alternatively we make sure to send a continued event so that the client knows execution was resumed? In case it > thought it had halted.

I think the UI is only refreshed from running to paused if you send a stopped event. Since the code doesn't send a stopped event when NextInProgress is on, even if halt comes and succeeds, the UI will match auto-continue, so a continued event should not be needed.

However, I don't think it is a good idea to ignore halt. If a user explicitly asks us to stop, we should not be auto-continuing. One might say that next is generally quick, so it won't be too noticeable. But next might be stepping over a long-running function and not be quick. And the goal is to come up with a general approach that would work for logpoints and then we are dealing with a continue with an indefinitely stopping point.

So how do we recognize halt? At the high level there seem to be two options:
1) Delay halt until after auto-continue is in progress by blocking the request loop. Then let halt interrupt auto-continue, which will clear NextInProgress for us, so we break out of the finish-next loop.
2) Track that halt was requested and use that signal to not auto-continue at all. 

The problem with 1) is that even if halt is issued while we are executing, execution might already be stopping for different reasons. So we might swallow halt anyway and not notice it and still issue an auto-continue when we are not supposed to.

Then there is also the issue of other requests.

Disconnect = halt + detach. We don't want to get halt before auto-continue and detach after as it would hang. Locking halt+detach (as the code does now) should solve that. If we auto-continue after, it would just fail. No harm done.

SetBreakpoint while running = halt + set. With the lock that shouldn't get interrupted by auto-continue as well, so that's good. But auto-continue would be inconsistent with how setting breakpoints while running works otherwise (it is essentially another form of halt as of now).

Other requests. If they arrive while we are continuing, there will be an is-running error. If they arrive while at temp stop, they will get processed. And because request loop holds the lock, auto-continue will not interrupt them. That would be ok for blocking requests like evaluating an expression. But wouldn't that be a problem for async requests? What if we schedule another continue? We should not auto-continue right after, should we? So maybe it makes more sense to handle these as if things are running even during temp stops? Then we would need an is-running flag+lock at the server level (something we decided against in favor of proc's is-running) in the first async PR.

aarzilli:
> > This seems much too complicated compared to continueUntilCompleteNext, I was thinking more of something along the lines of:
> > ```
> > func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
> > 	defer s.asyncCommandDone(asyncSetupDone)
> >         var state api.DebuggerState
> >         var err error
> >         for {
> > 		state, err = s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
> > 		if !(state != nil && state.NextInProgress) {
> > 			break
> > 		}
> > 		asyncSetupDone = nil
> > ```
> 
> I think there are some corner cases in debugger.Command that don't release the channel. That's why the code has defer, which won't help if you reset this.

There's already a defer at the start of the function, no need to pass it multiple times to Command.

> Also, don't we want the subsequent auto-continue command not to get interrupted by another incoming command prematurely? Isn't this what this relatively new chan arg is for? Added in #2370.

Because it could arrive during the window where IsRunning returns false? It doesn't seem worth the extra complexity to me. The continue button wouldn't even be enabled if you didn't send a stopped event, right?

> Does pause clear internal breakpoints (and by extension NextInProgress) if it is handled while stopped?

No, but depending on the operating system it will stop the next resume. Or not, hence the flag to stop anyways.

polinasok:
> > > This seems much too complicated compared to continueUntilCompleteNext, I was thinking more of something along the lines of:
> > > ```
> > > func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
> > > 	defer s.asyncCommandDone(asyncSetupDone)
> > >         var state api.DebuggerState
> > >         var err error
> > >         for {
> > > 		state, err = s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
> > > 		if !(state != nil && state.NextInProgress) {
> > > 			break
> > > 		}
> > > 		asyncSetupDone = nil
> > > ```
> > 
> > I think there are some corner cases in debugger.Command that don't release the channel. That's why the code has defer, which won't help if you reset this.
> 
> There's already a defer at the start of the function, no need to pass it multiple times to Command

That defer won't be of much help if the loop sets the channel to nil. I guess the assumption is that that non-nil state means command succeeded and closed the channel, so defer won't be needed.

> > Also, don't we want the subsequent auto-continue command not to get interrupted by another incoming command prematurely? Isn't this what this relatively new chan arg is for? Added in #2370.
> 
> Because it could arrive during the window where IsRunning returns false? It doesn't seem worth the extra complexity to me. The continue button wouldn't even be enabled if you didn't send a stopped event, right?

Yes, the continue button won't be shown. And I doubt users will try to do some crazy with keyboard shortcuts. But other commands can be an issue as well. We need to make sure that disconnect is not interleaved with auto-continue or detach might happen while running, which is not ok, right? Same with other requests. We might get an evaluate request and end up having auto-continue overlap with some part of responding to it when a blocking API call.

So far all the blocking handles have been written with the assumption that state is constant. If we allow auto-continue to kick in freely, that assumption will no longer hold. 

> > Does pause clear internal breakpoints (and by extension NextInProgress) if it is handled while stopped?
> 
> No, but depending on the operating system it will stop the next resume. Or not, hence the flag to stop anyways.

polinasok:
I have been wondering actually: is skipping breakpoints the right thing to do? It seems to me that when users set breakpoints, they want to actually stop at them and poke around. I know the PR description says:
> In order to get the behavior where they stop when a breakpoint is hit, they can simply add a breakpoint to the next line,
> and then run continue
but that's something they have to remember and do in advance. Once they skip some breakpoints, there is no going back. But if we stop at those breakpoints, the users have the option to explore them or to skip them manually.

I think the main issue with next in progress, is that people hit next again and then see "next while nexting" and assume that the error, like a panic or fatal throw has killed their debug session. If we could educate them that they can finish next with Continue, they might not mind this. We could make the stop reason be "paused on breakpoint (next in progress)" and the Debug Console print "next in progress, use continue to complete next". And if they do get "next while nexting", then show "use continue instead" along side the error message.

Alternatively I don't see a big problem with cancelling next either, as long as we explicitly report that. It's much easier to switch back to the file of interest in a graphical UI than in a terminal. If the user really doesn't care about the breakpoint, they would just go back and set a breakpoint in the other file. They can also highlight the goroutine/frame they care about and resume there.

Another option is to let Next finish next and Continue just continue. Or the other way around, so the user has both the option to cancel and to finish at their fingertips.

polinasok:
> I have been wondering actually: is skipping breakpoints the right thing to do? It seems to me that when users set breakpoints, they want to actually stop at them and poke around.

Hmm, looks like I am not the only one who thought this: #2317

aarzilli:
> That defer won't be of much help if the loop sets the channel to nil

That's not how defer works: https://play.golang.org/p/ArBkOGruSN7

> We need to make sure that disconnect is not interleaved with auto-continue or detach might happen while running, which is not ok, right?

Yes. I think this feature in general, speaks in favor of having a running flag maintained by dap/server, instead of using the one used by debugger. That way the handle loop will know if the target process is just stopped temporarily and will be autocontinued.

> I think the main issue with next in progress, is that people hit next again and then see "next while nexting" and assume that the error, like a panic or fatal throw has killed their debug session. If we could educate them that they can finish next with Continue, they might not mind this.

That could be. Note that they would have no option to cancel the next operation in progress because there is no UI for that.

> Alternatively I don't see a big problem with cancelling next either, as long as we explicitly report that. It's much easier to switch back to the file of interest in a graphical UI than in a terminal. If the user really doesn't care about the breakpoint, they would just go back and set a breakpoint in the other file. They can also highlight the goroutine/frame they care about and resume there.

It's actually hard to find the goroutine you were looking for unless you specifically took note of the goroutine ID beforehand and setting a breakpoint like next does (consider recursive functions).

Also I think users will just mash through with the step over button and not notice at all.

service/dap/server.go
```diff
@@ -385,8 +384,8 @@ func (s *Server) recoverPanic(request dap.Message) {
 
 func (s *Server) handleRequest(request dap.Message) {
 	defer s.recoverPanic(request)
-	s.resumeMu.Lock()
-	defer s.resumeMu.Unlock()
+	s.continueMu.Lock()
```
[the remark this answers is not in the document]
suzmue: That example was from when I thought that the UI would assume the halt succeeded when it receives a response, which it does not. Because of this the current code does not actually address this.

Though maybe it needs to depending on if Halt while paused clears the internal breakpoints.

polinasok:
> > That defer won't be of much help if the loop sets the channel to nil
> 
> That's not how defer works: https://play.golang.org/p/ArBkOGruSN7

Good point.

> > We need to make sure that disconnect is not interleaved with auto-continue or detach might happen while running, which is not ok, right?
> 
> Yes. I think this feature in general, speaks in favor of having a running flag maintained by dap/server, instead of using the one used by debugger. That way the handle loop will know if the target process is just stopped temporarily and will be autocontinued.

Everybody felt pretty strongly against this in the original PR (see [removing running flag](https://github.com/go-delve/delve/pull/2423/commits/33461eb65a5a3e7490da89d9dd8aed9992643257)), but there seems to be now way around this if we want to extend the definition of running to actually running + any related handling.
 
> > I think the main issue with next in progress, is that people hit next again and then see "next while nexting" and assume that the error, like a panic or fatal throw has killed their debug session. If we could educate them that they can finish next with Continue, they might not mind this.
> 
> That could be. Note that they would have no option to cancel the next operation in progress because there is no UI for that.

@suzmue Is this something we can address with those custom requests you uncovered? Maybe there can be a menu to cancel or continue next when user choses the next running command?

Although personally, I would imagine that when a user clicks "continue", they actually want to continue, not finish "next". We can train them to think about it differently, but it doesn't seem very intuitive be default.
 
> > Alternatively I don't see a big problem with cancelling next either, as long as we explicitly report that. It's much easier to switch back to the file of interest in a graphical UI than in a terminal. If the user really doesn't care about the breakpoint, they would just go back and set a breakpoint in the other file. They can also highlight the goroutine/frame they care about and resume there.
> 
> It's actually hard to find the goroutine you were looking for unless you specifically took note of the goroutine ID beforehand and setting a breakpoint like next does (consider recursive functions).
> 
> Also I think users will just mash through with the step over button and not notice at all.

That's a good point. That is how they get "next while nexting". But while going to the goroutine might be tricky, going to a file and setting a breakpoint in place of the removed internal breakpoint is still an option.

aarzilli:
> Everybody felt pretty strongly against this in the original PR (see removing running flag), but there seems to be now way around this if we want to extend the definition of running to actually running + any related handling.

It is true that we were against it during the original PR, but no one brought up this use case. Personally, I can live knowing that I was wrong.

polinasok:
@aarzilli @suzmue
It seems to me that we are having 2 conversations in parallel:
1) How to synchronize things to support skipping breakpoints
2) If we should use (1) in case of next.

I am happy to continue discussing (1) to arrive at a sound solution since we also need it for logpoints.

But I am also not convinced that this is the best way to support interrupted next (see https://github.com/golang/vscode-go/issues/119#issuecomment-861671563, #2317). @aarzilli What was the original reasoning behind implementing (1) in the terminal? I am wondering if I am missing some important consideration.

aarzilli:
> What was the original reasoning behind implementing (1) in the terminal?

IIRC we assumed that users that used next or step wanted to focus on one goroutine, it was a decision taken a long time ago.

> I am wondering if I am missing some important consideration.

I don't think so.

suzmue:
> It seems to me that we are having 2 conversations in parallel:
> 
> 1. How to synchronize things to support skipping breakpoints
> 2. If we should use (1) in case of next.

I agree that these are two separate conversations.

I chose to start with next while nexting for the skipping breakpoints implementation because there was less other code to worry about (parsing log messages etc), but if we choose to go another route with next, then we can reuse the solution for log points.

I tried to rework the solution to track the running separately for the debugger and ended up adding a layer to wrap debugger.Debugger. If we do not like that I can put the code back in the Server. I think adding some sort of layer makes sense to avoid having to think about the synchronization at each place that we call into the debugger.

aarzilli (commented):
IMHO this is too heavy handed, there's only two places that need to be concerned with the running flag: handleRequest and doRunCommand.

aarzilli:
PS. it's also wrong, doCall should definitely not start doing auto-continues, it's already dicey as is.

hyangah:
> I am happy to continue discussing (1) to arrive at a sound solution since we also need it for logpoints.

Re: logpoint -> Doesn't delve's tracepoint support have the same synchronization issue discussed here? Or, is it simpler because it doesn't currently evaluate an expression? It looks like tracepoint/logpoint are now a not-too-new concept in modern debuggers. I wonder if it can be supported from the lower layer than dap where the debugger state is kept track of.

suzmue:
> PS. it's also wrong, doCall should definitely not start doing auto-continues, it's already dicey as is.

Good point, I moved the synchronization logic back to the DAP server (it's the same code just within Server instead of Debugger).

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+			continued, state, err = s.runContinueCommand()
+			if continued {
+				continue
+			}
+			// Otherwise, a manual halt was attempted while the program was paused, so we should stop.
+			stopReason = proc.StopManual
+		}
+
+		// There was no breakpoint to skip, we should cancel next and process this stop.
+		if cancelNextErr := s.debugger.CancelNext(); cancelNextErr != nil {
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+			s.log.Error(cancelNextErr)
+		} else {
+			s.log.Debug("next cancelled\n")
+			s.send(&dap.OutputEvent{
```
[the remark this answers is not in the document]

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+	// Clear the manual halt check. handleRequest is blocked until asyncSetupDone
+	// is closed so this will make it empty when we start.
+	s.resetOverrideStopReason()
 	state, err := s.debugger.Command(&api.DebuggerCommand{Name: command}, asyncSetupDone)
+	// Make sure that asyncSetupDone is closed, so that handleRequest
+	// will release resumeMu.
+	s.asyncCommandDone(asyncSetupDone)
+
+	var stopReason = proc.StopUnknown
+	for err == nil && !state.Exited && state.CurrentThread != nil && state.NextInProgress {
```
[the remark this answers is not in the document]
suzmue: Changed to only check one.

service/dap/server.go
```diff
@@ -2501,13 +2548,62 @@ func (s *Server) doRunCommand(command string, asyncSetupDone chan struct{}) {
...
+				Body: dap.OutputEventBody{
+					Output:   fmt.Sprintf("goroutine %d hit breakpoint (id: %d, loc: %s:%d) during %s\n", id, bp.ID, bp.File, bp.Line, command),
+					Category: "console",
+				}})
+			var continued bool
+			continued, state, err = s.runContinueCommand()
+			if continued {
+				continue
+			}
+			// Otherwise, a manual halt was attempted while the program was paused, so we should stop.
```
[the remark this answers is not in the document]
suzmue: StopReason did not get updated if the program was already halted. In my latest change, I updated this to actually set the reason even if it had stopped before. If we keep that logic maybe we can use that to check instead.

service/dap/server_test.go
```diff
@@ -3011,6 +3011,79 @@ func TestNextAndStep(t *testing.T) {
 	})
 }
 
+func TestIssue387(t *testing.T) {
```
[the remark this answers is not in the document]

pkg/proc/target.go
```diff
@@ -108,6 +108,13 @@ func (sr StopReason) String() string {
 	}
 }
 
+func (t *Target) StopReason() StopReason {
```
aarzilli, line 111: I don't think that a method meant to inspect the reason why the target stopped should retroactively change it based on things that happened after the target stopped.

service/dap/server.go
```diff
@@ -128,6 +128,11 @@ type Server struct {
 	// sendingMu synchronizes writing to net.Conn
 	// to ensure that messages do not get interleaved
 	sendingMu sync.Mutex
+
+	runningMutex sync.Mutex
```
aarzilli, line 132: runningMu and stopMu, for naming consistency.

service/dap/server.go
```diff
@@ -2570,6 +2575,49 @@ func (s *Server) resetHandlesForStoppedEvent() {
 	s.exceptionErr = nil
 }
 
+func (s *Server) setRunning(running bool) {
+	s.runningMutex.Lock()
+	s.running = running
+	s.runningMutex.Unlock()
+}
+
+func (s *Server) IsRunning() bool {
```
aarzilli, line 2584: This can be private for now.

service/dap/server.go
```diff
@@ -2570,6 +2575,49 @@ func (s *Server) resetHandlesForStoppedEvent() {
...
+// halt is a helper function to allow dap to make sure that a halt
+// is processed. It is possible that this halt request will come in while a
+// a breakpoint is being processed. If the debugger decides to resume execution
+// after processing the breakpoint, the halt request would be skipped. Additional
+// synchronization is required between "resume" and "halt" to make sure this does
+// not happen.
+func (s *Server) halt() (*api.DebuggerState, error) {
+	s.stopMutex.Lock()
+	defer s.stopMutex.Unlock()
+	s.setRunning(false)
```
aarzilli, line 2599: I think the running flag should be fully controlled by the doRunCommand goroutine. Between the point where this line is executed and the point where doRunCommand finishes the value of the running flag is lying. Maybe it doesn't have any adverse effects but it bothers me.

service/dap/server_test.go
```diff
@@ -3109,6 +3109,79 @@ func TestNextAndStep(t *testing.T) {
 	})
 }
 
+func TestNextInterrupted(t *testing.T) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("test is not valid on ARM64")
```
aarzilli, line 3114: Why not? I understand the freebsd backend, but this is suspicious.

suzmue: Hmmmm I don't know where this came from... removing.

service/dap/server.go
suzmue: I think that makes sense.

If we keep the StoppedReason to not update after the fact, then we would need to call CheckAndClearManualHalt() from the dap code to check if a manual halt was issued OR add another tracking mechanism within DAP server to check if a halt was issued. I think the second would be too many different things to track when the target is already tracking it.

service/dap/server.go

service/dap/server.go

pkg/proc/target.go
suzmue: I was thinking about it as "even if it hadn't stopped for t.stopReason, it would have stopped because a manual halt was issued".  The rest of the state is still there and we can surface the breakpoints, etc.

I agree, it does seem a little weird though. I will remove this and have DAP server check this separately from checking the StopReason.

service/dap/server.go
aarzilli: I think tracking it explicitly is the better option, someone reading dap/server.go will not have to understand that StopReason always reports StopManual if a manual stop request was issued, especially since we make no such guarantee in the documentation.

service/dap/server.go
```diff
@@ -2569,6 +2574,48 @@ func (s *Server) resetHandlesForStoppedEvent() {
...
+// not happen.
+func (s *Server) halt() (*api.DebuggerState, error) {
+	s.haltMu.Lock()
+	defer s.haltMu.Unlock()
+	return s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
+}
+
+func (s *Server) resume() (*api.DebuggerState, error) {
+	resumeNotify := make(chan struct{}, 1)
+	s.haltMu.Lock()
```
aarzilli, line 2603: I don't think you need to acquire haltMu anymore, nor do you need to pass a non-nil resumeNotify to continue.

aarzilli: In fact what is haltMu protecting now?

suzmue: haltMu is protecting the invariant that if a halt request is issued, the process will either halt or remain halted.

If we do not include this, it is possible we could get a halt request right before continuing execution. If this halt came from a setBreakpoints request, the set breakpoints request would hang trying to change breakpoints while targetMutex is held by the continue command.

suzmue:
Closing this to keep the history readily available, but will open a new PR with this synchronization logic but for the log points use case.

PR #2535, opened 2021-06-14
sameuser: log the uid/remoteUID when they are not matching

hyangah:
Sometimes we observe failure in the same user check (not a file read/parse error).
Log uid/remoteUID, and the info from /proc/net/tcp{6} file.

Updates golang/vscode-go#1555

### DIFF
--- a/service/internal/sameuser/sameuser_linux.go
+++ b/service/internal/sameuser/sameuser_linux.go
@@ -1,4 +1,5 @@
-//+build linux
+//go:build linux
+// +build linux
 
 package sameuser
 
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM does this happen every time or only sometimes?

hyangah:
So far I observed it only once (and surprised...).

hyangah:
@polinasok hit this issue. But not sure if there is any actionable (but disabling the same user check in our integration test).

```
2021/07/23 07:06:24 sameuser_linux.go:62: connection from different user (remote: 0, local: 1001) detected:   
    60: 0100007F:DD82 0100007F:962D 06 00000000:00000000 03:00000133 00000000     0        0 0 3 0000000000000000   
2021/07/23 07:06:24 sameuser_linux.go:119: closing connection from different user (127.0.0.1:56706): 
    connections to localhost are only accepted from the same UNIX user for security reasons
```

aarzilli:
The 06 in the state column is TIME_WAIT, why is it saying that we closed the connection before we closed it?

polinasok:
> The 06 in the state column is TIME_WAIT, why is it saying that we closed the connection before we closed it?

Logs are [here](https://github.com/golang/vscode-go/issues/1555#issuecomment-887217767). I hit this once during an automated test run of a pending CL.

PR #2537, opened 2021-06-15
terminal,api: move PrintStack function

aarzilli:
Commit 30cdedae6910f5e9af6739845bacfd5b8778e745 introduced a dependency
from service/dap to pkg/terminal to call a stack printing function,
it's weird to have code that implements the DAP protocol depend on the
code for the JSON-RPC client.

Move PrintStack to a different package that can be called by both.


hyangah:
cc @suzmue

polinasok (approved):
Thank you!

derekparker (approved):
LGTM

PR #2540, opened 2021-06-16
proc: remove stack barrier support

derekparker:
Stack barriers were removed way back in Go 1.9 so it's safe to
eliminate and clean up this code now.


aarzilli (approved):
LGTM

PR #2543, opened 2021-06-17
proc,terminal: Ensure correct exit status

derekparker:
Ensure that any command executed after the process we are trying to
debug prints a correct and consistent exit status.

Previously the exit code was being lost after the first time we printed
that a process has exited. Additionally, certain commands would print
the PID of the process and other would not. This change makes everything
more correct and consistent.

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -213,8 +213,7 @@ func (t *Target) IsCgo() bool {
 func (t *Target) Valid() (bool, error) {
 	ok, err := t.proc.Valid()
 	if !ok && err != nil {
-		var pe ErrProcessExited
-		if errors.As(err, &pe) {
+		if pe, ok := err.(ErrProcessExited); ok {
 			pe.Status = t.exitStatus
 			err = pe
 		}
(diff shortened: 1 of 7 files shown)


pkg/proc/target.go
```diff
@@ -203,6 +207,21 @@ func (t *Target) IsCgo() bool {
...
 }
 
+// Valid returns true if this Process can be used. When it returns false it
+// also returns an error describing why the Process is invalid (either
+// ErrProcessExited or ErrProcessDetached).
+func (t *Target) Valid() (bool, error) {
+	ok, err := t.proc.Valid()
+	if !ok && err != nil {
+		var pe ErrProcessExited
+		if errors.As(err, &pe) {
```
aarzilli, line 217: `errors.As` has been introduced fairly recently, technically it's in all versions we support but every time we bump the minimum version required to compile delve we get complaints. Since it isn't that complicated to do this without `errors.As` (Valid doesn't chain ErrProcessExited with anything), I'd do it without it.

hyangah: Or, maybe golang.org/x/xerrors (intended for transition golang.org/design/29934-error-values)

derekparker: Done.

aarzilli (approved):
LGTM

PR #2544, opened 2021-06-18
proc,dwarf: Improve DWARF v5 support

derekparker:
While Go still mostly uses DWARF v4, newer versions of GCC will emit
DWARF v5 by default. This patch improves support for DWARF v5 by parsing
the .debug_line_str section and using that during file:line lookups.

### DIFF
--- a/pkg/dwarf/line/line_parser.go
+++ b/pkg/dwarf/line/line_parser.go
@@ -81,6 +81,9 @@ func ParseAll(data []byte, debugLineStr []byte, logfn func(string, ...interface{
 func Parse(compdir string, buf *bytes.Buffer, debugLineStr []byte, logfn func(string, ...interface{}), staticBase uint64, normalizeBackslash bool, ptrSize int) *DebugLineInfo {
 	dbl := new(DebugLineInfo)
 	dbl.Logf = logfn
+	if logfn == nil {
+		dbl.Logf = func(string, ...interface{}) {}
+	}
 	dbl.staticBase = staticBase
 	dbl.ptrSize = ptrSize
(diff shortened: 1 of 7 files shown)


pkg/dwarf/line/line_parser.go
```diff
@@ -174,11 +177,12 @@ func parseIncludeDirs5(info *DebugLineInfo, buf *bytes.Buffer) bool {
 		for dirEntryFormReader.next(buf) {
 			switch dirEntryFormReader.contentType {
 			case _DW_LNCT_path:
-				if dirEntryFormReader.formCode != _DW_FORM_string {
+				if dirEntryFormReader.formCode == _DW_FORM_string {
 					info.IncludeDirs = append(info.IncludeDirs, dirEntryFormReader.str)
 				} else {
```
aarzilli, line 182: I think this should be a switch:

```
switch dirEntryFormReader {
case _DW_FORM_string:
    //...
case <whatever the debug_line_string form is, DW_FORM_line_strp?>:
   //...
default:
   // the call to info.Logf
}
```

pkg/dwarf/line/line_parser.go
```diff
@@ -277,11 +281,11 @@ func parseFileEntries5(info *DebugLineInfo, buf *bytes.Buffer) bool {
 
 			switch fileEntryFormReader.contentType {
 			case _DW_LNCT_path:
-				if fileEntryFormReader.formCode != _DW_FORM_string {
+				if fileEntryFormReader.formCode == _DW_FORM_string {
```
aarzilli, line 284: Same as above.

aarzilli (approved):
LGTM

PR #2546, opened 2021-06-21
service/dap: increase default string loading limit and refactor

polinasok:
This PR proposes a limit of 512 with a TODO to consider 1024. 
Vscode truncates display values at 1023, so any higher won't give additional viewing power.
For additional inspiration, the C/C++ vscode adapter appears to be limiting string loads at 1024.

updates https://github.com/golang/vscode-go/issues/1318

### DIFF
--- a/_fixtures/longstrings.go
+++ b/_fixtures/longstrings.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"runtime"
+)
+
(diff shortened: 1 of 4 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #2547, opened 2021-06-23
test: disable avx512 test when backend is 'rr'

aarzilli:
The rr backend does not report avx512 registers back to us.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -1565,7 +1565,7 @@ func TestClientServer_FpRegisters(t *testing.T) {
 			if regtest.name == "XMM11" && !avx2 {
 				continue
 			}
-			if regtest.name == "XMM12" && !avx512 {
+			if regtest.name == "XMM12" && (!avx512 || testBackend == "rr") {
 				continue
 			}
 			found := false


derekparker:
Is this across the board, only with certain versions, or a bug upstream? FWIW you can explicitly disable recording of AVX512 registers so I'd imagine we could get that back from RR somehow: https://github.com/rr-debugger/rr/blob/878f756966c660a6e838cbf3516f953e586b9edc/src/RecordCommand.cc#L35.

aarzilli:
I don't know, but they are not in the target description files: https://github.com/rr-debugger/rr/tree/878f756966c660a6e838cbf3516f953e586b9edc/third-party/gdb so they can't be returned to anyone.

derekparker:
This test skip won't show up in our generated document to keep track of which backends skip which tests. I think we should break this test up into sub tests (https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) in order to `t.Skip` the AVX512 tests on RR and track it properly.

aarzilli:
That document doesn't check anything but proc_test.go, it wouldn't show up there anyway, without some further changes.

derekparker:
> That document doesn't check anything but proc_test.go, it wouldn't show up there anyway, without some further changes.

Ack you're right, forgot about that. This is fine then.

derekparker (approved):
LGTM

PR #2548, opened 2021-06-23
terminal: add prompt when breakpoint is hit during next/step/stepout

### LINKED ISSUE #2317: pkg/terminal: add option to disable auto-continue during next/step/stepout

aarzilli:
Adds a prompt asking the user what to do when a breakpoint is hit on a
different goroutine while next/step/stepout is being executed.

Gives the user the opportunity to cancel next/step/stepout, continue
once skipping the breakpoint or automatically skipping all other
concurrent breakpoints until next/step/stepout is finished.

Fixes #2317

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1436,8 +1436,28 @@ func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string, sho
 		}
 		return nil
 	}
+	autoContinue := false
 	for {
-		fmt.Printf("\tbreakpoint hit during %s, continuing...\n", op)
+		fmt.Printf("\tbreakpoint hit during %s", op)
+		if !autoContinue {
+			fmt.Printf("\n")
+			answer, err := promptAutoContinue(t, op)
+			switch answer {
+			case "f": // finish next
+				autoContinue = true
+				fallthrough
+			case "c": // continue once
+				fmt.Printf("continuing...\n")
+			case "s": // stop and cancel
+				fallthrough
+			default:
+				t.client.CancelNext()
+				printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
+				return err
+			}
+		} else {
+			fmt.Printf(", continuing...\n")
+		}
 		stateChan := t.client.DirectionCongruentContinue()
 		var state *api.DebuggerState
 		for state = range stateChan {
@@ -1450,6 +1470,20 @@ func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string, sho
 		if !state.NextInProgress {
 			printfile(t, state.CurrentThread.File, state.CurrentThread.Line, true)
 			return nil
+		}
+	}
+}
+
+func promptAutoContinue(t *Term, op string) (string, error) {
+	for {
+		answer, err := t.line.Prompt(fmt.Sprintf("[c] continue [s] stop her and cancel %s, [f] finish %s skipping all breakpoints? ", op, op))
+		if err != nil {
+			return "", err
+		}
+		answer = strings.ToLower(strings.TrimSpace(answer))
+		switch answer {
+		case "f", "c", "s":
+			return answer, nil
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/terminal/command.go
```diff
@@ -1454,6 +1474,20 @@ func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string, sho
 	}
 }
 
+func promptAutoContinue(t *Term, op string) (string, error) {
+	for {
+		answer, err := t.line.Prompt(fmt.Sprintf("[c] continue [s] stop her and cancel %s, [f] finish %s skipping all breakpoints? ", op, op))
```
derekparker, line 1479: s/her/here/

derekparker (changes requested):
One minor typo change.

pkg/terminal/command.go
```diff
@@ -1436,8 +1436,28 @@ func continueUntilCompleteNext(t *Term, state *api.DebuggerState, op string, sho
 		}
 		return nil
 	}
+	autoContinue := false
```
polinasok, line 1439: suggest renaming to skipBreakpoints

pkg/terminal/command.go
polinasok: It is not clear that "continue" will hit the internal breakpoint (unless it counters others along the way). Maybe "continue %s hitting all breakpoints"?

aarzilli: > s/her/here/

Done.

> It is not clear that "continue" will hit the internal breakpoint (unless it counters others along the way). Maybe "continue %s hitting all breakpoints"?

That makes it sound like it does the same thing as 'finish' IMO. It would have to be "continue and try to finish %s but repeat this prompt if a breakpoint is hit" but I think that's too verbose. I think just saying 'continue' here is fine, there's three options and with the other two it should be immediately obvious whether they are right or wrong, it should be clear by exclusion if you want 'continue' and it's a safe option to experiment with.

pkg/terminal/command.go
aarzilli: Done.

pkg/terminal/command.go
derekparker: I agree with @aarzilli here that the wording as is should be ok for users. If we get reports on confusion from a UX perspective we can reevaluate.

derekparker (approved):
LGTM

pkg/terminal/command.go
polinasok: That's fair.
Also, this could be explained in more detail in the documentation without polluting the prompt.

polinasok (approved):

PR #2549, opened 2021-06-23
service/dap: fix error string format

suzmue:
Fix style issues that are causing tests to fail (See go-delve/delve#2524)

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -614,7 +614,7 @@ func traceCmd(cmd *cobra.Command, args []string) {
 }
 
 func isBreakpointExistsErr(err error) bool {
-	return strings.Contains(err.Error(), "breakpoint exists")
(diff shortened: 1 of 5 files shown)


aarzilli (approved):

cmd/dlv/cmds/commands.go
```diff
@@ -614,7 +614,7 @@ func traceCmd(cmd *cobra.Command, args []string) {
 }
 
 func isBreakpointExistsErr(err error) bool {
-	return strings.Contains(err.Error(), "Breakpoint exists")
+	return strings.Contains(err.Error(), "breakpoint exists")
```
aarzilli, line 617: I don't think we should change this, the JSON-RPC API returns strings for errors and older clients may be relying on this formatting.

suzmue: Ok I reverted the changes that are not in delve dap.

aarzilli (approved):

PR #2550, opened 2021-06-23
service/dap: set hit breakpoint ids

suzmue:
By setting the breakpoint id for the breakpoint that was hit, the dap client can show the user
which breakpoint was hit by this stopped event. VS Code does this by shifting focus in the
breakpoints view to the breakpoint with the returned id.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -2058,7 +2058,7 @@ func TestHitBreakpointIds(t *testing.T) {
 				execute: func() {
 					checkStop(t, client, 1, "main.main", 30)
 
-					// Set two breakpoints at the next two lines in main
+					// Set two source breakpoints and two function breakpoints.
 					client.SetBreakpointsRequest(fixture.Source, []int{23, 33})
 					sourceBps := client.ExpectSetBreakpointsResponse(t).Body.Breakpoints
 					checkBreakpoints(t, client, []Breakpoint{{line: 23, path: fixture.Source, verified: true}, {line: 33, path: fixture.Source, verified: true}}, sourceBps)
@@ -2094,7 +2094,6 @@ func TestHitBreakpointIds(t *testing.T) {
 					checkHitBreakpointIds(t, se, "function breakpoint", functionBps[1].Id)
 					checkStop(t, client, 1, "main.anotherFunction", 27)
 				},
-				// The program has an infinite loop, so we must kill it by disconnecting.
 				disconnect: true,
 			}})
 	})
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -1947,7 +1947,7 @@ func checkBreakpoints(t *testing.T, client *daptest.Client, bps []Breakpoint, br
 	}
 	for i, bp := range breakpoints {
 		if bps[i].line < 0 && !bps[i].verified {
-			if bp.Verified != bps[i].verified || !strings.Contains(bp.Message, bps[i].msgPrefix) {
+			if bp.Verified != bps[i].verified || !stringContainsCaseInsensitive(bp.Message, bps[i].msgPrefix) {
 				t.Errorf("got breakpoints[%d] = %#v, \nwant %#v", i, bp, bps[i])
 			}
 			continue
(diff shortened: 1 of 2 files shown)


aarzilli (approved):

derekparker (approved):
LGTM

service/dap/server_test.go
```diff
@@ -2037,6 +2037,73 @@ func TestSetBreakpoint(t *testing.T) {
...
+				client.LaunchRequest("exec", fixture.Path, !stopOnEntry)
+			},
+			// Set breakpoints
+			fixture.Source, []int{30}, // b main.main
+			[]onBreakpoint{{
+				execute: func() {
+					checkStop(t, client, 1, "main.main", 30)
+
+					// Set two breakpoints at the next two lines in main
+					client.SetBreakpointsRequest(fixture.Source, []int{23, 33})
```
polinasok, line 2062: "Next two lines" - isn't this supposed to be 32 then?

service/dap/server_test.go
```diff
@@ -2037,6 +2037,73 @@ func TestSetBreakpoint(t *testing.T) {
...
+					checkHitBreakpointIds(t, se, "function breakpoint", functionBps[0].Id)
+					checkStop(t, client, 1, "main.anotherFunction", 26)
+
+					client.ContinueRequest(1)
+					client.ExpectContinueResponse(t)
+					se = client.ExpectStoppedEvent(t)
+					checkHitBreakpointIds(t, se, "function breakpoint", functionBps[1].Id)
+					checkStop(t, client, 1, "main.anotherFunction", 27)
+				},
+				// The program has an infinite loop, so we must kill it by disconnecting.
```
polinasok, line 2097: This looks like a left-over comment from copy-paste from some other test.

service/dap/server_test.go
```diff
@@ -2037,6 +2037,73 @@ func TestSetBreakpoint(t *testing.T) {
...
 
+func checkHitBreakpointIds(t *testing.T, se *dap.StoppedEvent, reason string, id int) {
+	if se.Body.ThreadId != 1 || se.Body.Reason != reason || len(se.Body.HitBreakpointIds) != 1 || se.Body.HitBreakpointIds[0] != id {
+		t.Errorf("got %#v, want Reason=%q, ThreadId=1, HitBreakpointIds=[]int{%d}", se, reason, id)
+	}
+}
+
+// TestHitBreakpointIds executes to a breakpoint and tests that
+// the breakpoint ids in the stopped event are correct.
+func TestHitBreakpointIds(t *testing.T) {
```
polinasok, line 2048: This feature should apply to all breakpoints, right? So why not check an additional setting no the breakpoints int the other breakpoint-dedicated tests we already have?

polinasok (changes requested):
I am not sure what you mean by "the dap client can show the user which breakpoint was hit". Doesn't it already show that based on the file/line info? 
image
How exactly is this new feature impacting the UI?

service/dap/server_test.go
suzmue: Sorry left over comment.

service/dap/server_test.go
suzmue: thanks for catching this. I was initially using "loopsprog" (and actually forgot I had switched the fixture).

suzmue:
Its pretty subtle in the VS Code UI:

Before:
Screen Shot 2021-06-25 at 12 07 02 PM

After:
Screen Shot 2021-06-25 at 12 08 06 PM

service/dap/server_test.go
suzmue: This should apply to all breakpoints, but I thought checking that each ID lined up would be too much of a hassle to do in every test, since we have to then indicated which breakpoint we wanted to hit for every ID. 

I guess this could be done with a search in the SetBreakpointsResponse, but I thought it would add unnecessary complexity to the other tests.

polinasok: I did not mean doing this in all test functions that use breakpoints to test other things. I meant adding a couple of extra lines to the test functions dedicated to breakpoints: TestPreSetBreakpoint, TestSetBreakpoint, TestSetFunctionBreakpoint.

polinasok:
> Its pretty subtle in the VS Code UI.

I was wondering if it would somehow impact the Breakpoints viewlet. That's pretty nice actually. Too bad it doesn't display the id there or let people name their breakpoints.

polinasok (approved):

PR #2553, opened 2021-06-25
service/debugger: improve CreateBreakpoint docs

derekparker:
The existing documentation was a little light on details. This patch
gives more insight into how the argument to this function is used and
interpreted.

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -622,7 +622,15 @@ func (d *Debugger) state(retLoadCfg *proc.LoadConfig) (*api.DebuggerState, error
 
 // CreateBreakpoint creates a breakpoint using information from the provided `requestedBp`.
 // This function accepts several different ways of specifying where and how to create the
-// breakpoint that has been requested.
+// breakpoint that has been requested. Any error encountered during the attempt to set the
+// breakpoint will be returned to the caller.
+//
+// The ways of specifying a breakpoint are listed below in the order they are considered by
+// this function:
+//
+// - If requestedBp.TraceReturn is true then it is expected that
+// requestedBp.Addrs will contain the list of return addresses
+// supplied by the caller.
 //
 // - If requestedBp.File is not an empty string the breakpoint
 // will be created on the specified file:line location
@@ -634,11 +642,11 @@ func (d *Debugger) state(retLoadCfg *proc.LoadConfig) (*api.DebuggerState, error
 // - If requestedBp.Addrs is filled it will create a logical breakpoint
 // corresponding to all specified addresses.
 //
-// - If requestedBp.TraceReturn is true then it is expected that
-// requestedBp.Addrs will contain the list of return addresses
-// supplied by the caller.
-//
 // - Otherwise the value specified by arg.Breakpoint.Addr will be used.
+//
+// Note that this method will use the first successful method in order to
+// create a breakpoint, so mixing different fields will not result is multiple
+// breakpoints being set.
 func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -620,7 +620,17 @@ func (d *Debugger) state(retLoadCfg *proc.LoadConfig) (*api.DebuggerState, error
 	return state, nil
 }
 
-// CreateBreakpoint creates a breakpoint.
+// CreateBreakpoint creates a breakpoint using information from the provided `requestedBp`.
+// This function accepts several different ways of specifying where and how to create the
+// breakpoint that has been requested.
+//
+// 1. You may provide a file:line combination and the debugger will take care of associating that with
+// the correct address in which to place the breakpoint in the running program.
+// 2. You may provide the name of a function. In this case the debugger will resolve the function name
+// to the correct address and place a breakpoint at the entry point of the function after the function prologue.
+// 3. You may provide a memory address (or addresses) directly and the debugger will set a breakpoint directly
+// on the addresses provided. Note that if you are setting a breakpoint of type `TraceReturn` it is expected that
+// the `requestedBp` contains the list of addresses representing each location that a function may return.
 func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint) (*api.Breakpoint, error) {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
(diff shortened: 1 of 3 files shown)


service/debugger/debugger.go
```diff
@@ -620,7 +620,17 @@ func (d *Debugger) state(retLoadCfg *proc.LoadConfig) (*api.DebuggerState, error
 	return state, nil
 }
 
-// CreateBreakpoint creates a breakpoint.
+// CreateBreakpoint creates a breakpoint using information from the provided `requestedBp`.
+// This function accepts several different ways of specifying where and how to create the
+// breakpoint that has been requested.
+//
+// 1. You may provide a file:line combination and the debugger will take care of associating that with
+// the correct address in which to place the breakpoint in the running program.
```
polinasok, line 628: It might be useful to mention behavior in error cases. For example, if you specify a blank line, it won't place a breakpoint at a statement immediately following, but return an error.

hyangah: api.Breakpoint doc is giving some hints about requests. What I want to learn is the relationship between requested breakpoints and returned breakpoints/errors (yes, the error cases @polinasok  mentioned), and whether all the fields in api.Breakpoint are relevant and can be mixed.

aarzilli:
This is currently documented by `(*rpc2.RPCServer).CreateBreakpoint`. This comment has some things the one in rpc2 doesn't and vice versa (the one that exists says how it works if multiple fields are specified). I think we should have only one, either in debugger or rpc2, it should be a combination of both and the other function should point to it.

aarzilli:
PS. IMHO the documentation should stay on the rpc2 method because we've always pointed people that want to use the JSON-RPC API to that documentation page. So the method in debugger should reference it.

derekparker:
> This is currently documented by `(*rpc2.RPCServer).CreateBreakpoint`. This comment has some things the one in rpc2 doesn't and vice versa (the one that exists says how it works if multiple fields are specified). I think we should have only one, either in debugger or rpc2, it should be a combination of both and the other function should point to it.

That's true, I think we should reference other docs where appropriate.

> PS. IMHO the documentation should stay on the rpc2 method because we've always pointed people that want to use the JSON-RPC API to that documentation page. So the method in debugger should reference it.

This seems backwards to me. The implementation, the actual logic that is being documented, is happening in `debugger.CreateBreakpoint`. The JSON-RPC API is no longer the only consumer of this method within the codebase. The DAP server uses it as well. I think the method that implements the logic should be documented directly and any callers of the method should reference it instead.

Consider the discoverability of the documentation as well. If you were a new contributor to this project and wanted to update the documentation for a change you've made, would you look right above the method you've modified or go search in a completely different package for the documentation?

service/debugger/debugger.go
derekparker: I agree, I'll add in some words about the error handling and the interactions between the different fields and also some more information on the returned breakpoint.

aarzilli:
> This seems backwards to me.

It is backwards. The logic is: the API in debugger is not stable, the only stable API is the one in rpc2, so we point users to the documentation of rpc2 which should be self contained. All users of debugger are assumed to be internal and if we change debugger we update them as well in the same commit. External users, if they exist, will find the comment inside debugger that points to the documentation on rpc2.

Another alternative is to put this in the comment for the api.Breakpoint type: "when an object of this type is passed to CreateBreakpoint the breakpoint is created (follows explanation...)".

derekparker:
> > This seems backwards to me.
> 
> It is backwards. The logic is: the API in debugger is not stable, the only stable API is the one in rpc2, so we point users to the documentation of rpc2 which should be self contained. All users of debugger are assumed to be internal and if we change debugger we update them as well in the same commit. External users, if they exist, will find the comment inside debugger that points to the documentation on rpc2.
> 
> Another alternative is to put this in the comment for the api.Breakpoint type: "when an object of this type is passed to CreateBreakpoint the breakpoint is created (follows explanation...)".

Hah, I didn't expect it but in a way I'm glad this change is sparking a bit of a discussion on how we document the codebase.

Currently, as you mentioned, we document things within the JSON-RPC API, except that we don't document the client library directly, we put all documentation in `rpc2/server.go`. However if we stop to think about it, is that really correct? Consider a brand new user trying to use our client library. One would expect to be able to read the documentation for the client library itself and be productive based off that. As of right now, there is almost no documentation at all attached to the client library itself (see https://pkg.go.dev/github.com/go-delve/delve/service/rpc2#RPCClient.CreateBreakpoint). Thinking of how I personally use other peoples code, I would expect the client library to be heavier on documentation since that's the first entry point from a users perspective. I wouldn't expect to look at the server documentation in order to know how to use the client library. Does that make sense?

Also, I'm well aware we have https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md so it's not like our client is undocumented, it's actually very well documented (thanks @aarzilli), but I think we should add documentation to the client lib itself which also further references the more extensive documentation (with examples, etc) that's found in the `ClientHowto.md` doc.

Here's my documentation philosophy thoughts:

1. Documentation of logic should live closest to where the behavior it describes is implemented (locality of documentation).
2. Our user-facing client libraries should be documented at a high level and then reference any further documentation the user might need to be able to successfully use the client lib.
3. The server should be documented at an even higher level and maybe reference the client lib or perhaps other internal documentation as well. In fact, I think the server documentation should be more "our" documentation (in other words, for contributors to the project and considered more as "internal" documentation rather than user-facing).

I wonder if godoc is smart enough to link types found in doc comments to their documentation?

aarzilli:
Documenting the client library is only immediately useful for clients written in Go which, at least going by published clients, is a minority.

I don't know, there's four places where any given piece of user relevant documentation can go (client, server, debugger, api). Either one is ok, there's pro and cons to all of them. Picking the server is consistent with what we've done so far and most useful to JSON-RPC clients not written in Go. Other choices have different advantages and disadvantages.

> I wonder if godoc is smart enough to link types found in doc comments to their documentation?

AFAIK, no.

derekparker:
> Documenting the client library is only immediately useful for clients written in Go which, at least going by published clients, is a minority.

This is a very good point, but I'd imagine folks would still browse through the official client when implementing their own to see how it interacts with the server.

> I don't know, there's four places where any given piece of user relevant documentation can go (client, server, debugger, api). Either one is ok, there's pro and cons to all of them. Picking the server is consistent with what we've done so far and most useful to JSON-RPC clients not written in Go. Other choices have different advantages and disadvantages.

True, there are a few layers at play here and it can be difficult to determine where is the best place. That is why it seems to me the easiest choice is to document the code where it's implemented and link to that from other relevant parts of the code. How would you feel about this PR if I kept the docs in the debugger package, updated the server docs to reflect there's additional information in the documentation for `debugger.CreateBreakpoint` and then update both client / server with a link to that documentation?

I think that's the easier and most complete way to cover all the bases. Client / Server are still documented for users and the actual logic is documented where it lives.

aarzilli:
> How would you feel about this PR if I kept the docs in the debugger package, updated the server docs to reflect there's additional information in the documentation for debugger.CreateBreakpoint and then update both client / server with a link to that documentation?

Ok. But also integrate the documentation of debugger.CreateBreakpoint with the stuff that was on RPCServer.CreateBreakpoint.

PR #2557, opened 2021-06-28
_scripts: auto check that pkg/proc and runtime are synchronized

aarzilli:
Adds a script that check automatically that the the assumptions that
pkg/proc makes about runtime are met by the actual runtime, using a
combination of parsing and magic comments.

Also emits a file describing all the struct fields, constants and
variables of the runtime that we use in pkg/proc.

### DIFF
--- a/_scripts/rtype-out.txt
+++ b/_scripts/rtype-out.txt
@@ -1,0 +1,78 @@
+var firstmoduledata moduledata
+
+var debug anytype
+
+type _defer struct {
+	fn anytype
+	pc uintptr
+	sp uintptr
+	siz int32 (optional)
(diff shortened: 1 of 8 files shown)


aarzilli:
Something is missing here, stand by.

aarzilli:
It's good now.

aarzilli:
PS. this was discussed last friday, the file _scripts/rtype-out.txt contains a list of everything from the runtime that we use (bar anything I have missed). It's autogenerated by _scripts/rtype.go and the same script also automatically checks that everything is still valid.

derekparker (approved):
LGTM

PR #2559, opened 2021-06-30
service/dap: add panic and throw text to stopped event

suzmue:
We can add more information to the stopped events on errors using
the `Text` field in the stopped event. We already use this to display
the runtime errors. Adding this information to the stopped reason will
also help to show the user additional info when a stopped event is not
associated with a particular goroutine.


suzmue:
Example for deadlock:
 
Screen Shot 2021-06-30 at 3 31 33 PM

pkg/proc/eval.go
```diff
@@ -67,16 +67,18 @@ func ConvertEvalScope(dbp *Target, gid, frame, deferCall int) (*EvalScope, error
...
-		return ThreadScope(ct)
-	}
 
 	var opts StacktraceOptions
 	if deferCall > 0 {
 		opts = StacktraceReadDefers
 	}
 
-	locs, err := g.Stacktrace(frame+1, opts)
+	var locs []Stackframe
```
aarzilli, line 76: I'm guessing that the reason for this change is that there is a bug where the frame parameter is ignored if we can't find the requested goroutine. I don't think we should sneak this bug fix in here without any comment.
It should be in a separate PR, if possible with a test in pkg/proc/proc_test.go.

suzmue: Done in #2563

aarzilli (approved):
LGTM

derekparker (changes requested):
Blocking this from merge until https://github.com/go-delve/delve/pull/2563 is pulled in and this is rebased.

suzmue:
> Blocking this from merge until #2563 is pulled in and this is rebased.

This branch has been rebased.

service/dap/server.go
```diff
@@ -2509,34 +2509,19 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
...
-			// Attempt to get the value of the throw reason.
-			// This is not currently working for Go 1.16 or 1.17: https://github.com/golang/go/issues/46425.
-			handleError := func(err error) {
-				if err != nil {
-					body.Description = fmt.Sprintf("Error getting throw reason: %s", err.Error())
-				}
+			body.Description, err = s.throwReason(goroutineID)
+			if err != nil {
+				body.Description = fmt.Sprintf("Error getting throw reason: %s", err.Error())
+				// This is not currently working for Go 1.16 or 1.17: https://github.com/golang/go/issues/46425.
```
polinasok, line 2515: This issue was closed 7 days ago.

service/dap/server.go
```diff
@@ -2593,6 +2578,25 @@ func (s *Server) onExceptionInfoRequest(request *dap.ExceptionInfoRequest) {
...
+func (s *Server) throwReason(goroutineID int) (string, error) {
+	return s.getExprString("s", goroutineID, 1)
+}
+
+func (s *Server) panicReason(goroutineID int) (string, error) {
+	return s.getExprString("(*msgs).arg.(data)", goroutineID, 0)
+}
+
+func (s *Server) getExprString(expr string, goroutineID, frame int) (string, error) {
+	exprVar, err := s.debugger.EvalVariableInScope(goroutineID, frame, 0, expr, DefaultLoadConfig)
```
polinasok, line 2590: Do we need a larger string limit here?

suzmue: The throw reasons are all pretty short, the only concern is a long panic message. I think the default of 512 will probably be ok for now? We can definitely increase it if we need to.

service/dap/server.go
suzmue: Removed comment about 1.17

polinasok (approved):

derekparker (approved):
LGTM

PR #2561, opened 2021-06-30
service/dap: send dap error response on dap error

suzmue:
The code previously expected the server to close when receiving
a message that it could not decode. However, there may be changes to
the spec that make new requests that we have not handled yet. In
that case, it would be preferable to return an error.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -357,7 +357,7 @@ func (s *Server) serveDAPCodec() {
 	s.reader = bufio.NewReader(s.conn)
 	for {
 		request, err := dap.ReadProtocolMessage(s.reader)
-		// Handle errors gracefully by responding with an ErrorResponse.
+		// Handle unknown command errors gracefully by responding with an ErrorResponse.
 		// -- "Request command 'foo' is not supported" means we
 		// potentially got some new DAP request that we do not yet have
 		// decoding support for, so we can respond with an ErrorResponse.
@@ -366,12 +366,12 @@ func (s *Server) serveDAPCodec() {
 			case <-s.stopTriggered:
 			default:
 				if err != io.EOF {
-					// Send an error response to the users if we were unable to process the message.
-					var seq int
 					if decodeErr, ok := err.(*dap.DecodeProtocolMessageFieldError); ok {
-						seq = decodeErr.Seq
+						// Send an error response to the users if we were unable to process the message.
+						s.sendDapErrorResponse(decodeErr.Seq, err.Error())
+						continue
 					}
-					s.sendDapErrorResponse(seq, err.Error())
+					s.log.Error("DAP error: ", err)
 				}
 				s.triggerServerStop()
 			}
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -372,7 +372,7 @@ func (s *Server) serveDAPCodec() {
 				if err != io.EOF {
 					if decodeErr, ok := err.(*dap.DecodeProtocolMessageFieldError); ok {
 						// Send an error response to the users if we were unable to process the message.
-						s.sendDapErrorResponse(decodeErr.Seq, err.Error())
+						s.sendInternalErrorResponse(decodeErr.Seq, err.Error())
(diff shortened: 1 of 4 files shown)


polinasok:
There is a TODO for what you did in this PR right above the code you touched. Please update it.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -357,19 +357,21 @@ func (s *Server) serveDAPCodec() {
...
 			case <-s.stopTriggered:
 			default:
 				if err != io.EOF {
-					s.log.Error("DAP error: ", err)
+					// Send an error response to the users if we were unable to process the message.
+					var seq int
+					if decodeErr, ok := err.(*dap.DecodeProtocolMessageFieldError); ok {
+						seq = decodeErr.Seq
+					}
+					s.sendDapErrorResponse(seq, err.Error())
```
polinasok: Actually, this is exactly the kind of thing that sendInternalErrorResponse was added for. It is used for similar things already in the code (except for the one instance in onLaunchRequest, where it is used incorrectly, but that one is going away as I am refactoring this code right now).

service/dap/server_test.go
```diff
@@ -4811,5 +4811,9 @@ func TestBadlyFormattedMessageToServer(t *testing.T) {
 		if err.Body.Error.Format != "DAP Error: Request command 'unknown' is not supported (seq: 1)" || err.Seq != 1 {
 			t.Errorf("got %v, want  Seq=1 Error=\"DAP Error: Request command 'unknown' is not supported (seq: 1)\"", err)
 		}
+
+		// Make sure that the unknown request did not kill the server.
```
polinasok, line 4815: That's a good idea.

service/dap/server.go
```diff
@@ -357,18 +357,20 @@ func (s *Server) serveDAPCodec() {
 	s.reader = bufio.NewReader(s.conn)
 	for {
 		request, err := dap.ReadProtocolMessage(s.reader)
-		// TODO(polina): Differentiate between errors and handle them
```
polinasok, line 360: Clarification about graceful error handling. There were 2 original todos that now got removed. The 2nd one was specifically for what you are adding, so makes sense. I am wondering about the 1st one. It was referring to the fact that there might be other kinds of errors here that we might want to handle gracefully. "Unknown command" was just one of them.  Technically, the way your code is written right now it would cover more than just unknown request commands , so "for example" that got removed is still accurate. Those other ones are just not at all likely (like unknown type). But in addition to the DecodeProtocolMessageFieldError, there could other kind of decoding errors. Maybe also network errors? Makes sense to make those fatal, but what bout thing like requests that are too large or unmarshalling errors? Would we want to handle those gracefully? Not that you need to add any code now, jus wondering if it makes sense to keep and revisit the TODO.

service/dap/server_test.go
```diff
@@ -4805,15 +4804,16 @@ func TestBadInitializeRequest(t *testing.T) {
 
 func TestBadlyFormattedMessageToServer(t *testing.T) {
 	runTest(t, "increment", func(client *daptest.Client, fixture protest.Fixture) {
-		// Send a badly formatted message to the server, and expect it to close the
```
polinasok, line 4808: Let's not take out a valid test case since there are still cases when we close the connection. You just need a different way of triggering it from the client (unmarshall error?), but the rest of the logic should still stand. In some cases you don't die (new test case) and in some cases you do (old test case modified).

polinasok (changes requested):

service/dap/server_test.go
suzmue: Ok added a test case back in.

service/dap/server.go
suzmue: Good point. I will add more details back into this comment.

service/dap/server.go

polinasok (approved):

PR #2562, opened 2021-07-01
service/dap: pretty print type for variables

suzmue:
Add a function to pretty print the type of a variable to add more
information in the type field in the variables response.

Example:
- interface{} -> interface{}(int)

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1947,7 +1947,7 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 	if !s.clientCapabilities.supportsVariableType {
 		return ""
 	}
-	return v.TypeString()
+	return api.ConvertVar(v).TypeString()
 }
 
 // convertVariable converts proc.Variable to dap.Variable value and reference
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1944,21 +1944,32 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 	if !s.clientCapabilities.supportsVariableType {
 		return ""
 	}
-	return prettyTypeString(v)
+	typ := maybeAddParens(underlyingTypeString(v))
+	if typ == maybeAddParens(v.TypeString()) {
+		return v.TypeString()
+	}
+	return fmt.Sprintf("%s %s", v.TypeString(), typ)
 }
 
-func prettyTypeString(v *proc.Variable) string {
+func underlyingTypeString(v *proc.Variable) string {
 	switch v.Kind {
 	case reflect.Ptr:
 		if len(v.Children) > 0 {
-			return "*" + prettyTypeString(&v.Children[0])
+			return "*" + underlyingTypeString(&v.Children[0])
 		}
 	case reflect.Interface:
 		if len(v.Children) > 0 && v.Children[0].Kind != reflect.Invalid {
-			return fmt.Sprintf("%s(%s)", v.TypeString(), v.Children[0].TypeString())
+			return maybeAddParens(v.Children[0].TypeString())
 		}
 	}
 	return v.TypeString()
+}
+
+func maybeAddParens(typ string) string {
+	if !strings.HasPrefix(typ, "(") {
+		typ = fmt.Sprintf("(%s)", typ)
+	}
+	return typ
 }
 
 // convertVariable converts proc.Variable to dap.Variable value and reference
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1944,8 +1944,19 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 	if !s.clientCapabilities.supportsVariableType {
 		return ""
 	}
-	if v.Kind == reflect.Interface && len(v.Children) > 0 {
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -1947,7 +1947,7 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 	if !s.clientCapabilities.supportsVariableType {
 		return ""
 	}
-	return v.TypeString()
+	return api.ConvertVar(v).TypeString()
```
aarzilli, line 1950: `ConvertVar` does a bunch of extra things, I think you should just do, in service/dap:

```
if v.Kind == reflect.Interface && len(v.Children) > 0 {
    return fmt.Sprintf("%s(%s)", v.TypeString(), v.Children[0].TypeString())
}
return v.TypeString()
```

Note that you have to check that there are children, which you didn't in your method. Also I'm not sure about the reflect.Ptr case. Why just pointers? What about slices, arrays and maps?

polinasok:
> Example:
> 
> * interface{} -> interface{}(int)

How are pointers impacted? It seems that only interfaces look different in the unittests. What has changed for the pointers?

suzmue:
> > Example:
> > 
> > * interface{} -> interface{}(int)
> 
> How are pointers impacted? It seems that only interfaces look different in the unittests. What has changed for the pointers?

Pointers to interface types are actually what is impacted by the pointer case.

For example `ptrinf` in the tests:
*interface{} -> *interface{}(**interface{})

service/dap/server.go
suzmue: Pointers point to a single address, whereas slices, arrays, and maps may have different underlying types for each interface, so looking for the concrete type does not make sense.

service/dap/server.go
```diff
@@ -1944,6 +1944,20 @@ func (s *Server) getTypeIfSupported(v *proc.Variable) string {
 	if !s.clientCapabilities.supportsVariableType {
 		return ""
 	}
+	return s.prettyTypeString(v)
+}
+
+func (s *Server) prettyTypeString(v *proc.Variable) string {
```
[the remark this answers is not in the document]

aarzilli (commented):
A potentially unintended side effect of this is that, in this program:

```
package main

import (
	"fmt"
)

type intptr *int

func main() {
	var p intptr = nil
	fmt.Printf("%#v\n", p)
}
```

The type of `p` becomes `*int` instead of `main.intptr`.

suzmue:
Thanks for pointing that out @aarzilli. I have updated this change to include both v.TypeString() and the new printed format. 

One of the things I was looking to explore with this PR was a way to remove the extra type info from the pretty printed string without losing that information.

Updated changes from this PR:
- `main.intptr` -> `main.intptr (*int)` // original pretty print `*int`
- `*int` -> `*int` // original pretty print `*int`
- `interface {}`-> `interface {} (int)` // original pretty print `interface {}(int)`
- `*interface {}` -> `*interface {} (*(int))` // original pretty print `*interface {}(int)`
- `*interface {}` -> `*interface {} (*(**interface {}))` // original pretty print `*interface {}(**interface {})`

aarzilli:
I'll leave the decision on this to others. IMO this is all too complicated: is `A (*B)` an interface of type A and concrete type *B or a typedef of *B called A? Why does `type C *D` get the parenthesized version but not `type C []D`? How is a user supposed to know what `*(int)` means?

suzmue:
> I'll leave the decision on this to others. IMO this is all too complicated: is `A (*B)` an interface of type A and concrete type *B or a typedef of *B called A? Why does `type C *D` get the parenthesized version but not `type C []D`? How is a user supposed to know what `*(int)` means?

You're right maybe this isn't the right place to put this. Perhaps it would make more sense as a child of the variable, where a name could be used to describe what its actually supposed to mean / present them in separate places.

If this is too complicated, we can hold off on adding this additional information, and consider the most appropriate place to include this after its been decided that it is necessary.

polinasok:
I can see that the type info included into the value is more detailed than what we get form type mouse-over in some cases (e.g. interfaces). I haven't checked all of them for a full comparison. Bu with the type in the value, the mouse-over is unlikely to get a lot of usage. Are there any specific types where you don't think there is enough detail in the inlined type+value and the mouse-over could be a good fallback?

I am not convinced we want to remove the type from the value string. It can be nice to see things at glance without having to hover and in the right spot too. I believe other typed languages also include type info.

If we were to remove the type info from the value string, wouldn't it make more sense to extend the existing pretty-printer to give us more options (e.g. return type and value separately) instead of redefining pretty-printing in the  dap server?

suzmue:
> I can see that the type info included into the value is more detailed than what we get from type mouse-over in some cases (e.g. interfaces). I haven't checked all of them for a full comparison. But with the type in the value, the mouse-over is unlikely to get a lot of usage. Are there any specific types where you don't think there is enough detail in the inlined type+value and the mouse-over could be a good fallback?

This change was mostly in preparation of removing that information from the inlined type.

Other dap editors besides VS Code may also choose to display the type info differently, which is why I was thinking it could be useful to add more detail here anyway.

It sounds like we think this is too complicated for now, especially because we aren't prepared to remove the type info from the value string. So I will close this PR.

PR #2563, opened 2021-07-01
pkg/proc: fix bug where frame parameter is ignored if no goroutine

suzmue:
ConvertEvalScope() attempts to find the scope for the specified
goroutine id and frame index. If the goroutine that is found is nil,
then it falls back to the threads stack trace to find the scope.
This fix makes sure that the frame id is taken into account for
thread strack traces as well.


aarzilli (approved):
LGTM, thank you.

derekparker (approved):
LGTM

derekparker:
@suzmue Please fix conflicts.

PR #2565, opened 2021-07-02
Update README.md

vadasambar:
- add info about frame error

Added a note in README around frames because I couldn't find much information around why I couldn't execute `next/step/stepout` commands in frames other than my deepest frame. 

[Slack thread](https://gophers.slack.com/archives/C07EXM5E0/p1625163851093300?thread_ts=1625153812.092900&cid=C07EXM5E0)

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -328,7 +328,7 @@ Set the current frame, or execute command on a different frame.
 The first form sets frame used by subsequent commands such as "print" or "set".
 The second form runs the command on the given frame.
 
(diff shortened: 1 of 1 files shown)


aarzilli:
I don't think this is necessary. The commands already returns an error when used incorrectly. If we linked the pull request for every feature we add delve would print nothing but pull request urls. The comment about how the numbering works is evident by the numbers printed by stack (and it's also how the frame command works in other debuggers).

derekparker:
The document you've added to is automatically generated and should not be edited directly. Perhaps adding something here https://github.com/go-delve/delve/blob/master/Documentation/cli/getting_started.md would be helpful. Otherwise, if you think the error message the Delve outputs could be more helpful we could investigate that as well. However, all that being said I think this PR as-is will not be merged.

PR #2566, opened 2021-07-02
dap: handle reverse request flow (e.g. RunInTerminal)

hyangah:
This PR adds support for reverse requests that are sent 
from the debug adapter to the client. Currently, RunInTerminal 
is the only such reverse request.

This is a pre-work for the 'console' support (golang/vscode-go#124).
We plan to use RunInTerminal to ask the editor to run a command 
in the integrated or external terminal while handling a launch request
with the console attribute.

Launch request was classified as a synchronous request
and was blocking ServeDAPCodec loop, which means the
response message from the client couldn't be read until
onLaunchRequest returns. This PR adds two goroutines -
one to handle requests from the client (loopHandleRequests),
and another to handle responses (loopHandleResponses).
serveDAPCodec will keep read DAP messages from the net.Conn,
but delegate message handling to the goroutines through
buffered channels.

Alternatively, I tried to avoid introducing goroutines
by making onLaunchRequest asynchronously complete the launch
if integrated or external console mode is chosen. I.e.,
make onLaunchRequest return without sending LaunchResponse,
free the ServeDAPCodec loop, and return LaunchResponse (or
ErrorResponse) when RunInTerminal response is received.
But it was hard to follow, didn't look like a typical
go program, and wasn't extensible when DAP adds more
reverse requests or we ever want to utilize RunInTerminal
while handling other requests.

For reverse requests, we maintain a pendingReverseReq map for
requests sent to the client. When response messages arrive
we look up the map, and notify the waiters using a buffered
channel.

onLaunchRequest uses RunInTerminal if the launch request
has "integrated" or "external" console attribute and asks
the client to run a bogus command - this is for testing.
The follow up PRs will implement the real command that
starts a delve and use the command from the integrated
or external terminal.


service/dap/server.go
```diff
@@ -376,7 +422,20 @@ func (s *Server) serveDAPCodec() {
...
 			return
 		}
-		s.handleRequest(request)
+
+		jsonmsg, _ := json.Marshal(message)
+		s.log.Debug("[<- from client]", string(jsonmsg))
+
+		switch m := message.(type) {
+		case dap.RequestMessage:
+			requests <- m
```
aarzilli, line 431: Is this a potential deadlock? If the request handling loop is waiting for a specific response but the server sends too many requests before the response.

service/dap/server.go
```diff
@@ -351,13 +361,49 @@ func (s *Server) Run() {
...
+			return
+		}
+	}
+}
+
 // serveDAPCodec reads and decodes requests from the client
 // until it encounters an error or EOF, when it sends
 // a disconnect signal and returns.
 func (s *Server) serveDAPCodec() {
+	requests := make(chan dap.RequestMessage, 10)
```
aarzilli, line 396: const maxNumQueuedRequests = 10 (same below)

service/dap/server.go
```diff
@@ -746,6 +864,35 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+				fmt.Sprintf("invalid value for console attribute: %q", console))
+			return
+		}
+		if !s.clientCapabilities.supportsRunInTerminalRequest {
+			s.sendErrorResponse(request.Request,
+				FailedToLaunch, "Failed to launch",
+				fmt.Sprintf("client does not support RunInTerminal feature necessary for console=%q", console))
+			return
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
aarzilli, line 882: The timeout should also be a constant.

hyangah: done.

service/dap/server.go
hyangah: done - or do you prefer them to be defined in the package level?

service/dap/server.go
hyangah: True. Added comments to explain the case and potential solutions. 🤞

aarzilli: I was thinking more along the lines of doing the send inside a select with a default statement and logging the error if the send can't be done.

aarzilli: The infinitely growing queue isn't a solution because infinite doesn't exist and by logging dropped messages you'll at least know that you have this problem.

aarzilli (approved):
LGTM

hyangah:
Rebased.

service/dap/daptest/client.go
```diff
@@ -539,3 +539,40 @@ func (c *Client) newRequest(command string) *dap.Request {
...
+
+// ExpectRunInTerminalRequest reads a protocol message from the connection
+// and fails the test if the read message is not *RunInTerminalRequest.
+func (c *Client) ExpectRunInTerminalRequest(t *testing.T) *dap.RunInTerminalRequest {
+	t.Helper()
+	m := c.ExpectMessage(t)
+	return c.CheckRunInTerminalRequest(t, m)
+}
+
+// CheckRunInTerminalResponse fails the test if m is not *RunInTerminalResponse.
```
suzmue, line 561: Should these say RunInTerminalRequest instead of RunInTerminalResponse?

suzmue (approved):
LGTM

polinasok:
Just wanted to mention that @hyangah is currently exploring alternatives design, so this PR is on hold for now until further notice.

PR #2567, opened 2021-07-02
service/dap: add backend launch/attach attribute

polinasok:
updates golang/vscode-go/issues/1591

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -836,6 +836,18 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		}
 	}
 
+	backend, ok := request.Arguments["backend"]
+	if ok {
+		backendParsed, ok := backend.(string)
+		if !ok {
+			s.sendErrorResponse(request.Request,
+				FailedToLaunch, "Failed to launch",
+				fmt.Sprintf("'backend' attribute '%v' in debug configuration is not a string.", backend))
+			return
+		}
+		s.config.Debugger.Backend = backendParsed
+	}
+
 	s.config.ProcessArgs = append([]string{program}, targetArgs...)
 	s.config.Debugger.WorkingDir = filepath.Dir(program)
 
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -836,6 +836,18 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		}
 	}
 
+	backend, ok := request.Arguments["backend"]
+	if ok {
+		backendParsed, ok := backend.(string)
+		if !ok {
+			s.sendErrorResponse(request.Request,
+				FailedToLaunch, "Failed to launch",
+				fmt.Sprintf("'backend' attribute '%v' in debug configuration is not a string.", backend))
```
hyangah, line 845: %q instead of '%v'

service/dap/server.go
```diff
@@ -836,6 +836,18 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		}
 	}
 
+	backend, ok := request.Arguments["backend"]
```
hyangah, line 839: is it simpler if use

backend, ok := request.Arguments["backend"].(string)
if !ok {
   ...
}

I think ok will be false if backend attribute doesn't exist so request.Arguments["backend"] is nil.

hyangah (commented):
LGTM

service/dap/server.go
polinasok: We combine the checks like this for required args, but split them for optional ones. This way if the arg is not specified, it won't be an error and a default will be used instead. (In this case, the default is supplied in command.go, but I can make it explicit in the server like with output). And if the arg is specified, then we only accept valid values and complain about anything malformed. This code is getting quite repetitive. I have been playing with refactoring and helpers in another PR.

hyangah: fyi - for config parsing I am proposing a different approach in #2571.

polinasok: I have updated the code to make it clear that missing value is not an error and a default is used instead. 
I don't think it is worth massaging this code to be more succinct since you have a PR that impacts this code anyway.

polinasok:
The Aggregator failure doesn't look related.

derekparker (approved):
LGTM

derekparker:
Will merge in the 1.7.1 window unless this is urgent enough to fit into the 1.7.0 release.

polinasok:
Users have not way to specify dlv-dap backend right now from vscode-go. The default is "default". "rr" is not yet available. Do they ever need to switch between native and lldb from the defaul?

derekparker:
> Users have not way to specify dlv-dap backend right now from vscode-go. The default is "default". "rr" is not yet available. Do they ever need to switch between native and lldb to a value other than implied by "default"?

I'm not sure what you mean here. Default should always do the right thing.

aarzilli (approved):

PR #2569, opened 2021-07-04
proc/native,proc/gdbserial: detect and complain about Rosetta

aarzilli:
Delve does not run under Rosetta. Detect this condition and point
confused users towards the solution.


derekparker (approved):
LGTM

PR #2571, opened 2021-07-04
dap: define LaunchConfig/AttachConfig types

hyangah:
Formally define these types and document their meaning.
We will auto-generate the dlv-dap documentation from these Go type doc.

mapToStruct is a helper that sets the given struct's fields with the
info in map[string]interface{} (launch/attach's Arguments). We achieve
this by reencoding map[string]interface{} to json and decoding back to
the target struct. If go-dap left the implementation-specific arguments
as json.RawMessage and let the implementation decode as needed, maybe
this reencoding could've been avoided.

encoding/json itself does not have mean to enforce required fields.
There was a test case that checks substitutePath elements must set
both from/to fields. Path.UnmarshalJSON implements the check.
I am not yet sure about the need for distinction between missing
'from/to' and empty strings yet. (empty value is useful when dealing with
a binary built with trimpath, right?)

A minor behavior change - previously, if noDebug is not a boolean type,
we ignored the attribute silently. Since we use json decoding, any
mismatched types will cause an error and this non-boolean type noDebug
attribute will result in launch failure.


service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Output path for the binary of the debugee.
+	Output string `json:"output,omitempty"`
+
+	// Path to the program folder (or any go file within that folder)
+	// when in `debug` or `test` mode, and to the pre-built binary file
+	// to debug in `exec` mode.
+	// If it is not an absolute path, it will be interpreted as a path
+	// relative to the working directory of the delve process.
+	Program string `json:"program,omitempty"`
```
hyangah, line 40: maybe required fields shouldn't be omitempty?
OTOH, I think it's ok to rely on checking zero value in most cases.
(StackTraceDepth was a concern but the current code already ignores '0')

service/dap/server.go
```diff
@@ -735,9 +710,15 @@ func cleanExeName(name string) string {
 }
 
 func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
+	var args LaunchConfig
+	if err := mapToStruct(request.Arguments, &args); err != nil {
```
aarzilli, line 714: I think that the logicaly thing would be to change dap.LaunchRequest.Arguments to be a json.RawMessage.

hyangah: Proposed https://github.com/google/go-dap/pull/57

service/dap/server.go
```diff
@@ -735,9 +710,15 @@ func cleanExeName(name string) string {
...
+		s.sendErrorResponse(request.Request,
+			FailedToLaunch, "Failed to launch", fmt.Sprintf("invalid debug configuration - %v", err))
+		return
+	}
 	// Validate launch request mode
-	mode, ok := request.Arguments["mode"]
-	if !ok || mode == "" {
+	mode := args.Mode
+	if mode == "" {
```
polinasok, line 721: What about defining a LaunchConfig constant with the defaults? This will both document what they are and allow you to unmarshall into a pre-populated struct. This way the defaults will stay if json doesn't specify alternatives.

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// NoDebug is used to run the program without debugging.
+	NoDebug bool `json:"noDebug,omitempty"`
+
+	LaunchAttachCommonConfig
+}
+
+// LaunchAttachCommonConfig is the attributes common in both launch/attach requests.
+type LaunchAttachCommonConfig struct {
```
polinasok, line 49: I think backend belongs here as well.

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+// LaunchConfig is the collection of launch request attributes recognized by delve DAP implementation.
+type LaunchConfig struct {
+	// Command line arguments passed to the debugged program.
+	Args []string `json:"args,omitempty"`
```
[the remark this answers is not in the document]
hyangah: Done.

ps. As you guessed, I generated this file with a go program based on vscode-go's settings.json - preserving map/object key orders is tricky so I went with alphabetic order.

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Backend used by delve. Maps to `dlv`'s `--backend` flag.
+	// Allowed Values: `"default"`, `"native"`, `"lldb"`
+	Backend string `json:"backend,omitempty"`
+
+	// Build flags, to be passed to the Go compiler. Maps to dlv's `--build-flags` flag.
+	BuildFlags string `json:"buildFlags,omitempty"`
+
+	// Absolute path to the working directory of the program being debugged
+	// if a non-empty value is specified. If not specified or empty,
+	// the working directory of the delve process will be used.
```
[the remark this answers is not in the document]
hyangah: fixed.

service/dap/types.go
hyangah: done

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+	// NoDebug is used to run the program without debugging.
+	NoDebug bool `json:"noDebug,omitempty"`
+
+	LaunchAttachCommonConfig
+}
+
+// LaunchAttachCommonConfig is the attributes common in both launch/attach requests.
+type LaunchAttachCommonConfig struct {
+	// Configuration name.
+	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
hyangah: Removed - I planned to use this when configuring the console (integrated or external terminal) with 'console' param. 
But I am still not sure whether using a fixed title for those terminals like debugpy is good enough as the first step.
Will add it if we really need it.

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Allowed Values: `"default"`, `"native"`, `"lldb"`
+	Backend string `json:"backend,omitempty"`
+
+	// Build flags, to be passed to the Go compiler. Maps to dlv's `--build-flags` flag.
+	BuildFlags string `json:"buildFlags,omitempty"`
+
+	// Absolute path to the working directory of the program being debugged
+	// if a non-empty value is specified. If not specified or empty,
+	// the working directory of the delve process will be used.
+	Cwd string `json:"cwd,omitempty"`
```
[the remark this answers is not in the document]
hyangah: done.

hyangah (commented):
Sigh I lost my draft responses. I dislike github codereview ux.

service/dap/types.go
hyangah: I don't see the need of distinguishing "" and omitted value in case of replay mode either. 
So, I am voting for just treating zero value and unset value in the same way, and leaving this as omitempty except fields we are sure we know omitting and "" need to be handled differently.

Let me know if there are fields where omitted value and zero value need to be interpreted differently.
(Is it catastrophic to handle substitutePath's missing from/to fields as if they were set to ""?)

service/dap/server.go
hyangah: once the pr is merged and go-dap is released, I will update this pr to use the new api.

service/dap/server_test.go
```diff
@@ -4562,41 +4555,41 @@ func TestBadLaunchRequests(t *testing.T) {
 			"Failed to launch: The program attribute is missing in debug configuration.")
 
 		// Bad "args"
-		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "exec", "program": fixture.Path, "args": nil})
+		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "exec", "program": fixture.Path, "args": "foobar"})
```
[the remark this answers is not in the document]
hyangah: nil is a perfectly fine (zero) value for []string type so this will be accepted as a valid launch config.
Is there a reason that args: nil should be considered a bad value?

service/dap/server_test.go
```diff
@@ -4562,41 +4555,41 @@ func TestBadLaunchRequests(t *testing.T) {
...
 		// Bad "buildFlags"
 		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "buildFlags": 123})
 		checkFailedToLaunchWithMessage(client.ExpectInvisibleErrorResponse(t),
-			"Failed to launch: 'buildFlags' attribute '123' in debug configuration is not a string.")
+			"Failed to launch: invalid debug configuration - cannot unmarshal number into \"buildFlags\" of type string")
 
 		client.LaunchRequestWithArgs(map[string]interface{}{"mode": "debug", "program": fixture.Source, "substitutePath": 123})
 		checkFailedToLaunchWithMessage(client.ExpectInvisibleErrorResponse(t),
-			"Failed to launch: 'substitutePath' attribute '123' in debug configuration is not a []{'from': string, 'to': string}")
+			"Failed to launch: invalid debug configuration - cannot unmarshal number into \"substitutePath\" of type []dap.SubstitutePath")
```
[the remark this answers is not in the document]
hyangah: Agree. Added a special casing for substitutePath and named the helper to be launch/attach request-specific (now unmarshalLaunchAttachArgs) because substitutePath is only for launch/attach requests.

fyi I am not sure how much more effort we want to make here without understanding how/why people would end up with such misconfiguration (assigning a completely wrong type to substitutePath) in practice. I think, it's better for us to focus on improving the documentation of how to use substitutePath (e.g. what does "" mean? Is it prefix of the path? should it use platform-specific paths? ....), and then as we gather more common misconfiguration cases over time we tweak the error messages to help minimizing confusion.

I think error messages when users specify non-string array value as args (e.g. If user uses [1, 2] instead of ["1", "2"], the error message is `cannot unmarshal number into \"args\" of type string`) is also confusing (encoding/json error message doesn't keep the full context of parsing and there are multiple issues open about improving json error messages) and needs improvement if it turns out many users end up with misconfiguration :-(

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// LaunchConfig is the collection of launch request attributes recognized by delve DAP implementation.
+type LaunchConfig struct {
+	// Command line arguments passed to the debugged program.
+	Args []string `json:"args,omitempty"`
+
+	// Backend used by delve. Maps to `dlv`'s `--backend` flag.
+	// Allowed Values: `"default"`, `"native"`, `"lldb"`
+	Backend string `json:"backend,omitempty"`
```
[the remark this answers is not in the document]
hyangah: The more I think, `dlv dap` should ignore `--backend` flag and other per-debug session flags - except the case where the full debug commands are supplied to the delve command (like dlv debug, test ...) and the debugger is setup without a launch/attach request.

Same as wd and buildFlags. What do you think?

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Build flags, to be passed to the Go compiler. Maps to dlv's `--build-flags` flag.
```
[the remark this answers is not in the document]
hyangah: Yeah, copy from vscode-go doesn't work again.
Changed to the description of --build-flags doc.

BTW what do you think about changing flags to []string?

service/dap/types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Absolute path to the working directory of the program being debugged
+	// if a non-empty value is specified. If not specified or empty,
+	// the working directory of the delve process will be used.
+	Cwd string `json:"cwd,omitempty"`
+
+	// Environment variables passed to the program.
+	Env map[string]string `json:"env,omitempty"`
+
+	// One of `debug`, `test`, `exec`.
```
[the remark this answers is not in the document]
hyangah: Done - DefaultLaunchMode = "".

What do you think about making this required and rejecting an empty string?
Do you know any client that depends on an empty mode?

service/dap/server.go
hyangah: Good idea. Added LaunchMode and defined MarshalJSON. Same of AttachMode.
I think we can apply a similar trick for backend (but as a string type).

Still torn between 'int' and 'string' as their underlying type - Use of 'int' (and make "" and "debug" map to 0) helps avoiding this ""  (string zero value when mode is omitted) -> "debug" mapping code. On the other hand, this string in json <->int in the program mapping may not look obvious.

service/dap/types.go
hyangah: Done

service/dap/server.go
```diff
@@ -743,7 +736,7 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		return
 	}
 
-	if mode == "debug" || mode == "test" {
+	if mode := args.Mode; mode == DebugLaunchMode || mode == TestLaunchMode {
```
polinasok, line 739: I know in general constants are considered a good practice because they code more readable, easier to change by only updating one place, etc. But in this case, I actually think they are not helpful :(

service/dap/types.go
polinasok: I have no issues with having program as omitempty. Just wanted to highlight that it is not always a required field, so no special treatment is needed.

SubstitutePath of "" is not very useful. If omitted or empty, the mapping will be equally useless.

I am not sure about cwd. I guess we are free to define it whichever way we see fit. But as a point of reference, I think --wd of "" is a valid input too the backend with rpc setup.

service/dap/types.go
```diff
@@ -97,20 +128,65 @@ func (m *SubstitutePath) UnmarshalJSON(data []byte) error {
...
+		return err
+	}
+	if err := json.Unmarshal(buf.Bytes(), config); err != nil {
+		if uerr, ok := err.(*json.UnmarshalTypeError); ok {
+			// Format json.UnmarshalTypeError error string in our own way. E.g.,
+			//   "json: cannot unmarshal number into Go struct field LaunchArgs.substitutePath of type dap.SubstitutePath"
+			//   => "cannot unmarshal number into 'substitutePath' of type {from:string, to:string}"
+			//   "json: cannot unmarshal number into Go struct field LaunchArgs.program of type string" (go1.16)
+			//   => "cannot unmarshal number into 'program' of type string"
+			if strings.HasPrefix(uerr.Value, "substitutePath") {
```
polinasok, line 184: Why has prefix instead of full mach?

polinasok (changes requested):

service/dap/types.go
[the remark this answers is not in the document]
hyangah: done.

service/dap/server.go
hyangah: Agree. :-(
Changed to "debug" and "test', const literals.
Still have LaunchMode and AttachMode consts in types.go - not sure if they are good for documentation purpose.
What do you think? Should I remove them too and go back to 'string'?

service/dap/types.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Absolute path to the working directory of the program being debugged
+	// if a non-empty value is specified. If not specified or empty,
+	// the working directory of the delve process will be used.
+	// This is similar to delve's `-wd` flag.
+	Cwd string `json:"cwd,omitempty"`
+
+	// Output path for the binary of the debugee.
+	// This is deleted after the debugsession ends.
+	Output string `json:"output,omitempty"`
```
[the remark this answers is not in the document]
hyangah: Done

service/dap/types.go
```diff
@@ -0,0 +1,192 @@
...
+	if err := json.Unmarshal(buf.Bytes(), config); err != nil {
+		if uerr, ok := err.(*json.UnmarshalTypeError); ok {
+			// Format json.UnmarshalTypeError error string in our own way. E.g.,
+			//   "json: cannot unmarshal number into Go struct field LaunchArgs.substitutePath of type dap.SubstitutePath"
+			//   => "cannot unmarshal number into 'substitutePath' of type {from:string, to:string}"
+			//   "json: cannot unmarshal number into Go struct field LaunchArgs.program of type string" (go1.16)
+			//   => "cannot unmarshal number into 'program' of type string"
+			if strings.HasPrefix(uerr.Value, "substitutePath") {
+				return fmt.Errorf("cannot unmarshal %v into 'substitutePath' of type {from:string, to:string}", uerr.Value)
+			}
```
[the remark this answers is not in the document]
hyangah: done.

service/dap/types.go
```diff
@@ -0,0 +1,192 @@
...
+	BuildFlags string `json:"buildFlags,omitempty"`
+
+	// Absolute path to the working directory of the program being debugged
+	// if a non-empty value is specified. If not specified or empty,
+	// the working directory of the delve process will be used.
+	// This is similar to delve's `-wd` flag.
+	Cwd string `json:"cwd,omitempty"`
+
+	// Output path for the binary of the debugee.
+	// This is deleted after the debugsession ends.
```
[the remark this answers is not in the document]
hyangah: Done

service/dap/server.go
hyangah: Saw the other comment. Yes, it's my misunderstanding. Removed the types and reverted back to plain string.

service/dap/types.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+type LaunchMode string
```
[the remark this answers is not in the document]
hyangah: I see. Removed the mode types.

Initializing non-zero default fields before json unmarshal is smart.
Currently the fields that need non-zero values as default are scalar types, so initializing with narrow copy will be safe.

But let's not do that now. There are only three fields that have non-zero values as default - `Backend`, `StackTraceDepth`, and `Output`. And, we still need to validate those fields after Unmarshal - for example, if the client does not omit empty/zero values but sends `{ mode:"", output:"", backend:"", stackTraceDepth:0 }`, they will overwrite the fields. Then we are now facing again whether to treat these different from those unset fields cases...

service/dap/types.go
hyangah: Ooops. I meant uerr.Field, not uerr.Value here.

Changed to the exact matching. Before implementing SubstitutePath custom unmarshaller, encoding/json may report 'substitutePath.from' if decoding error occurs while processing 'from'.

service/dap/types.go
```diff
@@ -1,44 +1,74 @@
...
+	TestLaunchMode LaunchMode = "test"
+	// "exec": executes a precompiled binary and begin a debug session.
+	ExecLaunchMode LaunchMode = "exec"
+)
+
+func (m *LaunchMode) UnmarshalJSON(data []byte) error {
+	var mode string
+	if err := json.Unmarshal(data, &mode); err != nil {
+		if _, ok := err.(*json.UnmarshalTypeError); ok {
+			return fmt.Errorf(`cannot unmarshal '%s' into 'mode' of type string`, data)
```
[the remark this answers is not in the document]
hyangah: Good idea. After removing Mode types, now there are only two places. Let's consider that when one more rewrite appears. :-)

service/dap/server_test.go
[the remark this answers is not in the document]
hyangah: There was a bug in unmarshalLaunchAttachArgs (should've checked Field instead of Value of json UnmarshalTypeError) and fixed.

service/dap/types.go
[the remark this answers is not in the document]
hyangah: Just saw https://github.com/mfussenegger/nvim-dap/issues/225 where users use launch config that is missing "mode" attribute. So, reverted the change that rejects requests without "mode" attribute, and applied Polina's comment to initialize the fields with non-zero default values.

I still think requests with explicit `{ mode: "" }` (empty value) is a strange case, so still treat them as errors. If you think it's safer to just "" as the default, let me know.

service/dap/types.go
```diff
@@ -14,13 +14,34 @@ func isValidLaunchMode(mode string) bool {
 	return false
 }
 
+// Default values for Launch/Attach configs.
+// Used to initialize configuration variables before decoding
+// arguments in launch/attach requests.
+var (
+	defaultLaunchAttachCommonConfig = LaunchAttachCommonConfig{
+		Backend:         "default",
+		StackTraceDepth: 50,
```
polinasok, line 23: We don't need defaultArgs in server.go anymore, do we?

service/dap/types.go
```diff
@@ -59,11 +80,11 @@ type LaunchAttachCommonConfig struct {
 	StopOnEntry bool `json:"stopOnEntry,omitempty"`
 
 	// Backend used by delve. See `dlv help backend` for allowed values.
```
polinasok, line 82: `dlv backend` also works

service/dap/server.go
```diff
@@ -214,43 +214,18 @@ func NewServer(config *service.Config) *Server {
...
-			}
-			to, ok := pathMapping["to"].(string)
-			if !ok {
-				return typeMismatchError
-			}
-			clientToServer = append(clientToServer, [2]string{from, to})
-			serverToClient = append(serverToClient, [2]string{to, from})
+func (s *Server) setLaunchAttachArgs(args LaunchAttachCommonConfig) error {
+	s.args.stopOnEntry = args.StopOnEntry
+	if depth := args.StackTraceDepth; depth > 0 {
```
polinasok, line 219: Is this logic needed now that you populate args with defaults too?

service/dap/server_test.go
```diff
@@ -4703,7 +4686,12 @@ func TestBadLaunchRequests(t *testing.T) {
 		}
 		checkFailedToLaunch(client.ExpectInvisibleErrorResponse(t))
 
-		client.LaunchRequest("" /*debug by default*/, fixture.Path+"_does_not_exist", stopOnEntry)
```
polinasok, line 4706: doesn't just changing "" to "debug" in LaunchRequest() work without doing the entire args map?

service/dap/types.go
```diff
@@ -6,39 +6,68 @@ import (
...
+	// This is similar to delve's `--wd` flag.
 	Cwd string `json:"cwd,omitempty"`
 
-	// Environment variables passed to the program.
-	Env map[string]string `json:"env,omitempty"`
-
-	// One of `debug`, `test`, `exec`.
-	Mode string `json:"mode,omitempty"`
+	// Build flags, to be passed to the Go compiler.
+	// For example, "-tags=integration -mod=vendor -cover -v".
```
polinasok, line 64: Like --build-flags

service/dap/types.go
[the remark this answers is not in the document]
polinasok: This is still inaccurate, right?

polinasok (changes requested):

service/dap/types.go
hyangah: NewServer uses defaultArgs. Not sure if it does anything useful but I feel that the cleanup belongs to a separate change.

service/dap/server.go
hyangah: yes, it is needed in case the client sends 0 or a negative value.

service/dap/types.go
hyangah: Done.

Side note: I really wish we could accept []string  and skip delve's custom build flag parsing if possible. There are subtle differences between dlv and go's flag parsing.

service/dap/server_test.go
hyangah: I reverted the change that rejects an empty mode, so reverted this test change too. Obsolete

service/dap/types.go
polinasok: NewServer uses them just to have defaults in case user supplies nothing. Your new code accomplishes that now instead, so the original defaults should no longer be needed. I think this is very much in scope for this change. But at the very least please add a TODO.

polinasok (changes requested):
Only a couple minor nits. Otherwise, this is ready for @derekparker final stamp of approval.

service/dap/types.go
```diff
@@ -6,14 +6,42 @@ import (
...
+//      Optional args with default: output, cwd, noDebug
+//      Optional args: buildFlags, args
+// -- "test" - builds and launches debugger for specified test (similar to 'dlv test')
+//      same args as above
+// -- "exec" - launches debugger for precompiled binary (similar to 'dlv exec')
+//      Required args: program
+//      Optional args with default: cwd, noDebug
+//      Optional args: args
+// -- replay: skips program build and sets the Debugger.CoreFile property based on the
+//		Required args: coreFilePath
```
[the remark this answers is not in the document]
hyangah: done

service/dap/types.go
hyangah: Added TODO. Please do the cleanup as you wish. Thanks.

service/dap/types.go
```diff
@@ -56,12 +86,12 @@ type LaunchConfig struct {
 
 	// Absolute path to the working directory of the program being debugged
 	// if a non-empty value is specified. If not specified or empty,
-	// the working directory of the delve process will be used.
+	// the program or the program's directory will be used.
```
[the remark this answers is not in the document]
hyangah: Rephrased it to be the built program. And I believe that behavior is a bug and needs to be fixed.

service/dap/types.go
```diff
@@ -85,12 +86,12 @@ type LaunchConfig struct {
...
 	// This is similar to delve's `--wd` flag.
 	Cwd string `json:"cwd,omitempty"`
 
 	// Build flags, to be passed to the Go compiler.
-	// For example, "-tags=integration -mod=vendor -cover -v".
+	// For example, "-tags=integration -mod=vendor -cover -v" like delve's `--build-flags`.
```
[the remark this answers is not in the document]
hyangah: Done

service/dap/types.go
```diff
@@ -56,12 +86,12 @@ type LaunchConfig struct {
...
 	BuildFlags string `json:"buildFlags,omitempty"`
 
 	// Output path for the binary of the debugee.
```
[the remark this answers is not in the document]
hyangah: Done.

hyangah:
I think I addressed all the issues but not sure how to get rid of the "change requested" mark. Is there anything missing or does it need @polinasok's review? Github review UI is strange.

polinasok (approved):

aarzilli (approved):

derekparker (approved):
LGTM

polinasok:
Please see errors.

PR #2572, opened 2021-07-04
all: update github.com/spf13/cobra to v1.1.3

hyangah:
This removes indirect dependencies from go.mod, and
includes the fix for the missing -help flag info.

The latest cobra release is v1.2.1. Given that there were
minor security-related dependency cleanup during v1.2 release,
I was tempted to pick up the latest version, but that caused
dependency updates in golang.org/x/sys and golang.org/x/tools
which may be too recent (golang.org/x/* follow the go's release
support policy, so recent versions may not be compatible with
go versions beyond go's official version support policy).

Verified that dlv still builds with go1.12.x.
(go1.12 is the oldest version of go that can build the latest delve already).

$ go get -d github.com/spf13/cobra@v1.1.3
$ go mod tidy
$ go mod vendor
$ go run _scripts/gen-usage-docs.go


aarzilli:
motivation?

hyangah:
Sorry I was too optimistic. 

The original motivation was to update the md file generator (I didn't like the dlv_dap.md change in https://github.com/go-delve/delve/pull/2568/files ) Obviously updating to v1.2.1 is too ambitious - that brings in recent golang.org/x/* modules.

The current dependency (2017) is rather old and that's a version before having go.mod. Maybe time to consider updating? 
v1.1.3 seems to be the latest one that doesn't cause too many updates in go.mod.

derekparker:
@hyangah could you please rebase this.

hyangah:
Rebased.
The test failures are due to https://github.com/go-delve/delve/issues/2602

derekparker (approved):
LGTM

PR #2575, opened 2021-07-08
service/dap: add modes comments (TODO from PR/2367)

polinasok:


service/dap/server.go
```diff
@@ -932,16 +932,34 @@ func (s *Server) stopNoDebugProcess() {
...
+func isValidLaunchMode(mode interface{}) bool {
+	switch mode {
 	case "exec", "debug", "test":
 		return true
 	}
-
 	return false
 }
 
+// Atach debug seesions support the following modes:
```
aarzilli, line 955: Attach and sessions

polinasok: Oops. Done.

service/dap/server.go
```diff
@@ -932,16 +932,34 @@ func (s *Server) stopNoDebugProcess() {
...
 		return true
 	}
-
 	return false
 }
 
+// Attach debug sessions support the following modes:
+// -- [DEFAULT] "local" -- attaches debugger to a local running process
+//      Required args: processId
+// TODO(polina): support "remote" mode
```
hyangah, line 958: Is it still a todo? what does this mean?

polinasok: I moved the old TODO to a different place. We discussed having a mode where the debugger is already debugging a process (either started from cli without a client or left over from another client session in a multi-session mode), so the attach request won't be attaching to a new process, but will just "attach" to an already running debug session. This will be a mode different from "local". The legacy name for that is "remote", but we could also choose something else.

aarzilli (approved):

service/dap/server.go
lggomez: Yup, the old remote mode!. Attach sounds better although remote also made sense in the context of attaching to a remote docker container. The old adapter had a plethora of issues pre and mid-modules rollout which I guess this implementation won't have but there were still several border cases of folks trying to attach to docker containers via WSL which I bet will be interesting to try here (for reference, [here's](https://github.com/microsoft/vscode-go/issues?q=is%3Aissue+is%3Aclosed+remote) the query for this mode's issues)

derekparker (approved):
LGTM

derekparker:
I'm waiting for the 1.7.1 merge window to open before this goes in.

Issue #2552, opened 2021-06-25
dap: inconsistencies with how signals are forwarded to process


polinasok:
The original plan was to have dlv-dap server behave as would a headless dlv-rpc server. Thus, `service.Config.Foreground` was always set to true. This got changed when `tty` flag was introduced: https://github.com/go-delve/delve/commit/3e60ae202b7d7c67fe7ea5fffafc92c7e39e4b67#r47528045. And while `tty` is not a global flag and doesn't show up in dlv-dap mode, the change also made the code depend on the value of `--headless`. As a result the foreground behavior changes if `--headless` is changed, even though code prints a warning that this flag is ignored because dap mode is always headless. I thought this was merely cosmetic and never got around to correcting this. But now I am realizing that started relying on SIGINT not being forwarded as part of clean shutdown (see #2445). Perhaps it is worth adding an explicit flag to forward or not forward the signals? Some users do need to forward the signals while others appreciate the simplicity of killing the server with Ctrl-C.

```
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"
)

func main() {

	sigs := make(chan os.Signal, 1)
	done := make(chan bool, 1)

	signal.Notify(sigs, syscall.SIGINT)

	go func() {
		sig := <-sigs
		fmt.Println()
		fmt.Println(sig)
		done <- true
	}()

	fmt.Println("awaiting signal")
	<-done
	fmt.Println("got signal: exiting...")
}
```

```
$ go run sigint.go
awaiting signal
^C
interrupt
got signal: exiting...

$ dlv debug sigint.go ############# Does not forward signals
Type 'help' for list of commands.
(dlv) c
awaiting signal
received SIGINT, stopping process (will not forward signal)
Stopped at: 0x7fff20323cde
=>   1:	no source available
(dlv) exit

$ dlv debug sigint.go --headless ############# Forwards signals
API server listening at: 127.0.0.1:58781
debugserver-@(#)PROGRAM:LLDB  PROJECT:lldb-1205.0.27
 for x86_64.
Got a connection, launched process /Users/polina/delve/__debug_bin (pid = 81180).
awaiting signal
^C
interrupt
got signal: exiting...

$ ./dlv dap --listen=:54321 ############ Foreground=false
DAP server listening at: [::]:54321
awaiting signal
^C

$ ./dlv dap --listen=:54321 --headless=false ########### Foreground=false
DAP server listening at: [::]:54321
awaiting signal
^C

[...]

PR #2578, opened 2021-07-12
cmd/dap: add dlv dap --foreground

polinasok:
updates #2552


aarzilli:
Wouldn't it be better to just have it always pass true? Or at least have the foreground flag default to true?

derekparker:
Should this simply be closed since https://github.com/go-delve/delve/pull/2589 was merged? The diff on this looks wrong too, it's not taking into account the PR, it seems.

polinasok:
> Should this simply be closed since #2589 was merged? The diff on this looks wrong too, it's not taking into account the PR, it seems.

Yes, this needed a merge.

derekparker:
@polinasok can you run `go run _scripts/gen-usage-docs.go`

hyangah:
If this is useful for delve dap users, can this be applied to other headless commands?

polinasok:
> If this is useful for delve dap users, can this be applied to other headless commands?

I think there isn't general consensus that this is useful, which is why I only added it to dlv dap.

polinasok:
Any additional thoughts? Votes for or against?

aarzilli:
Generally against, I don't see many people asking for this and it doesn't seem useful.

PR #2580, opened 2021-07-12
cmd: treat SIGTERM as a server disconnect signal

polinasok:
SIGINT might get forwarded to the target process, so we need another way to request clean shutdown.

**Before**
```
DAP server listening at: [::]:54321
2021-07-12T03:00:21-07:00 debug layer=dap DAP server pid = 19852
Terminated: 15
```
```
API server listening at: [::]:54321
2021-07-12T03:03:23-07:00 debug layer=rpc API server pid = 20146
2021-07-12T03:03:23-07:00 info layer=debugger launching process with args: [/Users/polina/delve/__debug_bin]
debugserver-@(#)PROGRAM:LLDB  PROJECT:lldb-1205.0.27
 for x86_64.
Got a connection, launched process /Users/polina/delve/__debug_bin (pid = 20161).
Terminated: 15
Exiting.
```
**After**
```
DAP server listening at: [::]:54321
2021-07-12T03:00:53-07:00 debug layer=dap DAP server pid = 19889
2021-07-12T03:01:04-07:00 debug layer=dap DAP server stopping...
2021-07-12T03:01:04-07:00 debug layer=dap DAP server stopped
```
```
API server listening at: [::]:54321
2021-07-12T03:02:19-07:00 debug layer=rpc API server pid = 20013
2021-07-12T03:02:19-07:00 info layer=debugger launching process with args: [/Users/polina/delve/__debug_bin]
debugserver-@(#)PROGRAM:LLDB  PROJECT:lldb-1205.0.27
 for x86_64.
Got a connection, launched process /Users/polina/delve/__debug_bin (pid = 20028).
2021-07-12T03:02:46-07:00 debug layer=debugger detaching
```

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -685,14 +685,14 @@ func connectCmd(cmd *cobra.Command, args []string) {
 }
 
 // waitForDisconnectSignal is a blocking function that waits for either
-// a SIGINT (Ctrl-C) signal from the OS or for disconnectChan to be closed
-// by the server when the client disconnects.
+// a SIGINT (Ctrl-C) or SIGTERM (kill -15) OS signal or for disconnectChan
+// to be closed by the server when the client disconnects.
 // Note that in headless mode, the debugged process is foregrounded
 // (to have control of the tty for debugging interactive programs),
 // so SIGINT gets sent to the debuggee and not to delve.
 func waitForDisconnectSignal(disconnectChan chan struct{}) {
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGINT)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
 	if runtime.GOOS == "windows" {
 		// On windows Ctrl-C sent to inferior process is delivered
 		// as SIGINT to delve. Ignore it instead of stopping the server


cmd/dlv/cmds/commands.go
```diff
@@ -685,14 +685,14 @@ func connectCmd(cmd *cobra.Command, args []string) {
...
 // Note that in headless mode, the debugged process is foregrounded
 // (to have control of the tty for debugging interactive programs),
 // so SIGINT gets sent to the debuggee and not to delve.
 func waitForDisconnectSignal(disconnectChan chan struct{}) {
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGINT)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
 	if runtime.GOOS == "windows" {
 		// On windows Ctrl-C sent to inferior process is delivered
 		// as SIGINT to delve. Ignore it instead of stopping the server
```
polinasok, line 698: Should I worry about SIGTERM behavior on Windows? (I don't have a system to test this on).

aarzilli:
I'm not sure what's the connection between this and #2552. Signals sent to the delve process never get forwarded to the target process, there is no mechanism that does this, unless the cli frontend runs. What the `Foreground` flag does is give access to stdin to the target process and make it the foreground process for the terminal, the latter means that it will be the process that receives SIGINT when ^C is pressed, SIGTSTP when ^Z is pressed, etc.

On macos things are more complicated because there is debugserver between us and the target process, `Foreground` makes debugserver the foreground process.

So #2552 is talking about the Foreground flag but no matter how that is set a SIGINT sent directly to delve will do the same thing, OTOH there is no way to generate a SIGTERM with a key combination by default (I think?) so what good does adding it as a termination signal do?

polinasok:
Thank you for the detailed explanation. Looks like I did not realize that foregrounding was accomplished like with "fg" in a terminal. Somehow I was under the impression that the debugger received and dispatched anything incoming either to itself or the target. In that case, we can indeed continue sending SIGINT to delve from the extension for clean shutdown regardless of how #2552 bug gets resolved. 

Given that SIGTERM is the "nice" program termination signal, why is it not treated as such and responded to with a clean shutdown?

aarzilli:
> Given that SIGTERM is the "nice" program termination signal, why is it not treated as such and responded to with a clean shutdown?

Because we never did. We could start. I was thinking that the advantage with this is that SIGTERM is the signal kill sends by default, so it could be useful.

aarzilli (approved):

cmd/dlv/cmds/commands.go
derekparker: I don't believe so, the code that follows will still work even for a forwarded SIGTERM.

derekparker (approved):
LGTM

PR #2582, opened 2021-07-13
[Draft] Support env from dap request

weihanglo:
This PR adds `env` arguments support from DAP requests, which is quite handy for some IDE/Editor's configuration files, such like VSCode's launch.json.

If this feature is acceptable, I'll drive it to finish. Thanks.

### DIFF
--- a/_fixtures/testenv_from_dap.go
+++ b/_fixtures/testenv_from_dap.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"os"
+)
+
+func main() {
+	// this test expects two ENV: VAR1=foo VAR2=bar.
+	if os.Getenv("VAR1") != "foo" {
+		panic("VAR1 is not foo!")
+	}
+	if os.Getenv("VAR2") != "bar" {
+		panic("VAR2 is not bar!")
+	}
+}
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -428,7 +428,7 @@ func getLdEnvVars() []string {
 // LLDBLaunch starts an instance of lldb-server and connects to it, asking
 // it to launch the specified target program with the specified arguments
 // (cmd) on the specified directory wd.
-func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs []string, tty string, redirects [3]string) (*proc.Target, error) {
+func LLDBLaunch(cmd, environ []string, wd string, flags proc.LaunchFlags, debugInfoDirs []string, tty string, redirects [3]string) (*proc.Target, error) {
 	if runtime.GOOS == "windows" {
 		return nil, ErrUnsupportedOS
 	}
@@ -523,9 +523,7 @@ func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs [
 		process.SysProcAttr = sysProcAttr(foreground)
 	}
 
-	if runtime.GOOS == "darwin" {
-		process.Env = proc.DisableAsyncPreemptEnv()
-	}
+	process.Env = proc.MergeInheritedEnviron(environ)
 
 	if err = process.Start(); err != nil {
 		return nil, err
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -90,7 +90,15 @@ func withTestProcess(name string, t testing.TB, fn func(p *proc.Target, fixture 
 	withTestProcessArgs(name, t, ".", []string{}, 0, fn)
 }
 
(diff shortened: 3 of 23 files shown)


hyangah:
FYI VS Code Go extension already handles `env` argument by launching dlv dap with the environment. So, I am curious about the motivation of this PR. 

- Is this to handle the case where the dlv dap isn't launched by the extension?
- In the current VS Code Go + Delve DAP integration, the `env` arg in VSCode's launch.json applies to both building of the target binary and running the binary. I think ability of separating them is useful but first we need to decide the meaning of `env` argument.
- `env` should be applied to `noDebug` case. 

I also wanted `dlv dap` to recognize the env arguments but that is purely due to a cosmetic reason (when launching a delve command through RunInTerminal mode).

weihanglo:
> FYI VS Code Go extension already handles env argument by launching dlv dap with the environment.

That's true, but in reality there are other DAP clients don't support environment out-of-box. It would be great if Delve supports it.

> but first we need to decide the meaning of env argument.

Oh sorry. I didn't aware of that. The PR is focused on **running the binary (debugee)**, which plays a similar role like `args`. IMO it'd be better to separate build env from execution env, so if we could add another arg `buildEnv` if needed.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -523,9 +523,7 @@ func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs [
 		process.SysProcAttr = sysProcAttr(foreground)
 	}
 
-	if runtime.GOOS == "darwin" {
```
aarzilli, line 526: I do not like that you moved the calls to DisableASyncPreemptEnv out of the backends and into `proc.MergeInheritedEnviron`. It is error prone, as evidenced by the fact you made several errors doing it and now it doesn't get called everywhere it should be called.

I think it would be better to leave them where they are but pass an environment to it to mutate.

pkg/proc/proc_test.go
```diff
@@ -90,7 +90,15 @@ func withTestProcess(name string, t testing.TB, fn func(p *proc.Target, fixture
 	withTestProcessArgs(name, t, ".", []string{}, 0, fn)
 }
 
+func withTestProcessEnv(name string, t testing.TB, environ []string, fn func(p *proc.Target, fixture protest.Fixture)) {
```
aarzilli, line 93: Just add an environment argument to `withTestProcessArgs` instead of adding a new function.

_fixtures/testenv_from_dap.go
```diff
@@ -0,0 +1,15 @@
+package main
```
aarzilli, line 1: why not use the same test fixture for dap?

pkg/proc/target.go
```diff
@@ -144,15 +145,26 @@ type NewTargetConfig struct {
...
 			// Go 1.14 asynchronous preemption mechanism is incompatible with
 			// debuggers, see: https://github.com/golang/go/issues/36494
 			env[i] += ",asyncpreemptoff=1"
 		}
 	}
+}
+
+// MergeInheritedEnviron merges given environment variables with inherited ones,
+// in the form "key=value".
+func MergeInheritedEnviron(environ []string) []string {
```
aarzilli, line 160: I think it's weird that this doesn't work like the `Env` field of `os/exec.Cmd`, i.e. when it's nil it uses `os.Environ()` otherwise it uses whatever it's passed.

aarzilli (commented):
I think there is a general problem with this. The only reason such an extended change would be if there was a way, within the same run of delve, to restart the target process with a different environment.

Right now this ability does not exist. Which means the parent process of delve can always control exactly the environment of the target process of delve by controlling the environment of delve itself. Do we think we want to do that? Because if we don't then the only reason for this is to support the `env` in launch.json which can be accomplished by:

```
for _, x := range request.Arguments["env"].([]string) {
    os.Setenv(x)
}
```

at an appropriate place (after the gobuild calls) inside `dap/server.go`.

service/debugger/debugger.go
```diff
@@ -244,7 +249,7 @@ func (d *Debugger) TargetGoVersion() string {
 }
 
 // Launch will start a process with the given args and working directory.
-func (d *Debugger) Launch(processArgs []string, wd string) (*proc.Target, error) {
+func (d *Debugger) Launch(processArgs, environ []string, wd string) (*proc.Target, error) {
```
aarzilli, line 252: You are not saving environ in d.environ but later you expect it to be filled.

weihanglo:
Thanks for all your review and comments!
I am going to close this since I have no use case now. If someone wants to work on it, feel free to take it.

PR #2584, opened 2021-07-13
service/dap: add len as metadata for map

suzmue:
This change adds an additional metadata variable for maps. There is currently no place that the size of a map is included. To make this information easier for users to access, we can add it as a named variable to be returned as a child of a map.

Map:
Screen Shot 2021-07-13 at 12 32 40 PM

Paged Indexed:
Screen Shot 2021-07-13 at 12 32 46 PM

Empty map:
Screen Shot 2021-07-13 at 12 36 14 PM

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1903,8 +1903,7 @@ func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 
 func getNamedVariableCount(v *proc.Variable) int {
 	namedVars := 0
-	switch v.Kind {
-	case reflect.Map:
+	if v.Kind == reflect.Map {
 		// len
 		namedVars += 1
 	}
@@ -1921,8 +1920,7 @@ func getNamedVariableCount(v *proc.Variable) int {
 func (s *Server) metadataToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variable, error) {
 	var children []dap.Variable
 
-	switch v.Kind {
-	case reflect.Map:
+	if v.Kind == reflect.Map {
 		children = append(children, dap.Variable{
 			Name:         "len()",
 			Value:        fmt.Sprintf("%d", v.Len),
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1903,7 +1903,7 @@ func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 
 func getNamedVariableCount(v *proc.Variable) int {
 	namedVars := 0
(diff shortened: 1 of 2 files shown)


service/dap/server_test.go
```diff
@@ -1281,8 +1281,9 @@ func TestScopesAndVariablesRequests2(t *testing.T) {
 					if ref > 0 {
 						client.VariablesRequest(ref)
 						m2 := client.ExpectVariablesResponse(t)
-						checkChildren(t, m2, "m2", 1) // each key-value represented by a single child
-						ref = checkVarExact(t, m2, 0, "1", "m2[1]", "*main.astruct {A: 10, B: 11}", "int: *main.astruct", hasChildren)
+						checkChildren(t, m2, "m2", 2) // each key-value represented by a single child
+						checkVarExact(t, m2, 0, "len()", "len(m2)", "1", "int", false)
```
polinasok, line 1285: s/false/noChildren/

service/dap/server.go
```diff
@@ -1915,6 +1921,16 @@ func getNamedVariableCount(v *proc.Variable) int {
 func (s *Server) metadataToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variable, error) {
 	var children []dap.Variable
 
+	switch v.Kind {
+	case reflect.Map:
+		children = append(children, dap.Variable{
```
polinasok, line 1924: Shouldn't you be checking that number of children is not 0?

polinasok (changes requested):

service/dap/server.go
suzmue: Yes I will explicitly check this here. If it has no children, then the parent variable will not have a variables reference, so the children are never requested.

polinasok: Shouldn't ListOfBytesOrRunes use v.Len as well instead of length of children (which might fail to load although unlikely)?

polinasok: Another thought is to just check Children in the outer function once for all types. It will be more obvious that metadata is skipped for empty child-less variables.

suzmue: These are actually two different scenarios. For ListOfBytesOrRunes, we need to have an actual child to inspect to determine if it is the write kind of slice, but with a map, even if it isn't loaded at the time, we can still show the length.

We could attempt to load the slice if its Len > 0 but len(Children) == 0, but I'm not sure if thats worthwhile.

polinasok: Ah, good point.

polinasok (approved):

derekparker (approved):
LGTM

PR #2585, opened 2021-07-13
pkg/proc: update check for system goroutine

suzmue:
The finalizer goroutine can be either a system goroutine or a user goroutine. It is considered a user goroutine only when it calls back to user code. This change adopts the approximation done by src/cmd/trace/trace.go which only special cases "runtime.main", and always considers the finalizer to be a system goroutine.


pkg/proc/variables.go
```diff
@@ -534,8 +534,19 @@ func (g *G) System(tgt *Target) bool {
...
-	case "runtime.main", "runtime.handleAsyncEvent", "runtime.runfinq":
+	case "runtime.main", "runtime.handleAsyncEvent":
 		return false
+	case "runtime.runfinq":
+		s, err := goroutineEvalScope(tgt, g, 0, 0)
+		if err != nil {
+			return true
+		}
+		v, err := s.EvalVariable("fingRunning", loadSingleValue)
+		if err != nil {
```
aarzilli, line 545: `|| v.Unreadable != nil`

suzmue: Done.

aarzilli:
Can you do a rebase of this I want to see if any of the test failures are because of this or old ones (I think it's the latter).

pkg/proc/variables.go
```diff
@@ -534,8 +534,19 @@ func (g *G) System(tgt *Target) bool {
 		return false
 	}
 	switch loc.Fn.Name {
-	case "runtime.main", "runtime.handleAsyncEvent", "runtime.runfinq":
+	case "runtime.main", "runtime.handleAsyncEvent":
 		return false
+	case "runtime.runfinq":
+		s, err := goroutineEvalScope(tgt, g, 0, 0)
+		if err != nil {
+			return true
```
derekparker, line 542: The default before was to return `false`, this new change is essentially "assume it's a system goroutine unless we can prove otherwise" correct?

suzmue: Correct.

It seems to be much more rare that it is a user goroutine than a system goroutine. And the approximation in src/cmd/trace/trace.go would treat the finalizer as a system goroutine always.

suzmue:
@derekparker @aarzilli 

I updated this change to implement the simpler solution of always treating the finalizer goroutine as a system goroutine. We plan to expose the goroutineFilters to DAP users and by default only show user goroutines. Having this finalizer goroutine appear as a user goroutine would cause more confusing I believe than having it appear as a system goroutine.

aarzilli (approved):

derekparker (approved):
LGTM

derekparker:
Re-running CI before merge.

PR #2587, opened 2021-07-13
service/dap: send terminated event when disconnecting

suzmue:
From the DAP spec:

> Terminated Event indicates that debugging of the debuggee has terminated. This does not mean that the debuggee itself has exited.

We should send a terminated event when we send a disconnect response since we are terminating the debugging of the debuggee.


polinasok:
Please update dlv_test.go

service/dap/server.go
```diff
@@ -999,7 +1002,6 @@ func (s *Server) stopDebugSession(killProcess bool) error {
 		switch err.(type) {
 		case proc.ErrProcessExited:
 			exited = err
```
polinasok: We could also send ExitedEvent for all the case where we detect that process has Exited, but I have no idea how that would change/help. The event takes an exit code and I thought maybe vscode would reflect that somehow, but in my experiments I did not see any impact. I also checked a number of implementations form the dap spec page and did not find any usage there or in the mock debugger.

suzmue:
> Please update dlv_test.go

Done.

service/dap/server.go
suzmue: There is no corresponding UI for exited events (at least in VS Code) at this point. If this is something important to Go users we should probably file an issue with them.

polinasok: We already print to the debug console when process exited, so I don't think it is very important to have special support in the UI.

polinasok (approved):

derekparker (approved):
LGTM

polinasok:
@derekparker I am glad this got merged in, thank you, but curious why this made it into 1.7.0.

derekparker:
> @derekparker I am glad this got merged in, thank you, but curious why this made it into 1.7.0.

I included it into 1.7.0 because it seemed like a very low-risk change on our end and (more importantly) seemed to me like a bugfix patch where perhaps the VSCode (or other DAP) client expects this response and it was not being provided. I suppose I should have clarified the latter assumption before merging, however it seemed very low-risk either way.

PR #2588, opened 2021-07-13
Update README.md

saihemanth-cloudera:


derekparker:
This change is not correct. Go is deprecating use of `go get`. See this commit which introduced the change you're trying to revert: https://github.com/go-delve/delve/commit/780bcac9ccbc1711b4310b9cae653e759a23c1c0#diff-881d56b5ef5874b314fbef016c7057f4b41f564130a44ebf156a95a83639973d.

PR #2589, opened 2021-07-14
cmd/dap: server - always headless, target - always foregrounded

polinasok:
updates #2552

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -188,6 +188,7 @@ While --continue is not supported, stopOnEntry launch/attach attribute can be us
 execution is resumed at the start of the debug session.`,
 		Run: dapCmd,
 	}
+	// TODO(polina): support --tty when dlv dap allows to launch a program from command-line
 	rootCommand.AddCommand(dapCommand)
 
 	// 'debug' subcommand.
@@ -411,6 +412,8 @@ func dapCmd(cmd *cobra.Command, args []string) {
 
 		if headless {
 			fmt.Fprintf(os.Stderr, "Warning: dap mode is always headless\n")
+		} else {
+			headless = true
 		}
 		if acceptMulti {
 			fmt.Fprintf(os.Stderr, "Warning: accept-multiclient mode not supported with dap\n")
@@ -446,10 +449,9 @@ func dapCmd(cmd *cobra.Command, args []string) {
 			DisconnectChan: disconnectChan,
 			Debugger: debugger.Config{
 				Backend:              backend,
-				Foreground:           headless && tty == "",
+				Foreground:           headless,
 				DebugInfoDirectories: conf.DebugInfoDirectories,
 				CheckGoVersion:       checkGoVersion,
-				TTY:                  tty,
 			},
 			CheckLocalConnUser: checkLocalConnUser,
 		})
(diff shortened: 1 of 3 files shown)


cmd/dlv/cmds/commands.go
```diff
@@ -411,6 +412,8 @@ func dapCmd(cmd *cobra.Command, args []string) {
 
 		if headless {
 			fmt.Fprintf(os.Stderr, "Warning: dap mode is always headless\n")
+		} else {
+			headless = true
```
hyangah, line 416: how about unconditionally set to headless. (dap mode is always headless)

hyangah (approved):
LGTM

cmd/dlv/cmds/commands.go
polinasok: Do you mean change the error message?

hyangah:
```
if headless {
   fmt.Fprintf(...)
}
headless = true
```

Removing the error message completely is another thing to consider - currently this doesn't provide a warning for  `dlv dap --headless=false`.

polinasok: I removed the error message and added more info to he help documentation.

The idea was that the default is false, so when people don't specify the flag (knowing that the server is always headless), then are not spammed. Is there a way to tell if false is coming from default or from a value the user specified explicitly?

derekparker: You can use the following to see if the flag has been set explicitly: https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#FlagSet.Changed.

cmd/dlv/cmds/commands.go
```diff
@@ -409,9 +410,7 @@ func dapCmd(cmd *cobra.Command, args []string) {
 		}
 		defer logflags.Close()
 
-		if headless {
-			fmt.Fprintf(os.Stderr, "Warning: dap mode is always headless\n")
-		}
+		headless = true
```
derekparker, line 413: I don't think this temp var is used anywhere except when initializing the `debugger.Config` below so let's just get rid of it and set the value to `true` below.

cmd/dlv/cmds/commands.go
```diff
@@ -409,9 +410,7 @@ func dapCmd(cmd *cobra.Command, args []string) {
 		}
 		defer logflags.Close()
 
-		if headless {
-			fmt.Fprintf(os.Stderr, "Warning: dap mode is always headless\n")
```
derekparker, line 413: I think we should use the `FlagSet.Changed` method to see if the user tried to explicitly set this to `false` and print a warning that it will be ignored.

derekparker (changes requested):

cmd/dlv/cmds/commands.go
polinasok: Just what I was looking for! Done!

cmd/dlv/cmds/commands.go
polinasok: Heh, full circle to what the Foreground was before we started messing with it :) Done.

derekparker (approved):
LGTM

derekparker:
Will merge when the 1.7.1 window opens.

PR #2592, opened 2021-07-15
service/dap: variables response must not have null variables array

### LINKED ISSUE #2590: dap variables response can be `null` but must be empty
Some users of nvim-dap reported that they get an error when using `dlv dap` as debug adapter.

You can find the full issue here: https://github.com/mfussenegger/nvim-dap/issues/225

We managed to figure out that they do get a `null` variables response, and that trips up some code in nvim-dap that doesn't expect a null value - because according to the debug adapter protocol specification, the `variables` result must not be `null`:

    debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":7,"success":true,"command":"variables","body":{"variables": null}}

The odd thing is that I wasn't able to reproduce it, and instead get a correct empty variables response:

    debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":7,"success":true,"command":"variables","body":{"variables":[]}}

I had a quick look at the code - including it's history - and didn't find anything suspicious:

> https://github.com/go-delve/delve/blob/5dd4b7df9da1770edb4c2fd86535b58b1df294e7/service/dap/server.go#L1566-L1570

But I'm also not too familiar with Go.

```
Version: 1.6.1
Build: $Id: 114218c22f3791287c4bc2f4ff35a846a1416ee9 $
```

Unfortunately I didn't ask the issue reporters in the original issue.

With the following version it works for me:

```
> go version
go version go1.16.5 linux/amd64
```

Unknown

- Works on Linux/ x86_64
- Original issue reporters are on MacOS

4. What did you do?

Set a breakpoint in the following code in the `Println` Line. `nvim-dap` automatically retrieves the scopes and variables if a breakpoint is hit.

```go
package main

import "fmt"


func main() {
    words := "world"
    fmt.Println("hello" + words)
}
```

6. What did you see instead?

- See description above

Maybe you have an idea what's going on - if not feel free to close, I will be merging a workaround - although I think it ought to be fixed in delve given that it is not compliant with the specification.

polinasok:
Fixes #2590

The above issue breaks clients that don't expect null as per spec.

From [dap spec](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Variables) ([json](https://microsoft.github.io/debug-adapter-protocol/debugAdapterProtocol.json)), 
```
variables: Variable[];
```
```
"variables": {
        "type": "array",  <========== no "null" option
```

Technically, only top-level scope variables (Arguments, Locals, etc) would result variables requests and could trigger this issue. The rest of childless variables are mapped to a reference of 0 and therefore do not offer an option for further expansion to view the children that result in a variable request/response. But just in case I am updating children initialization for all types and cases.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1716,7 +1716,7 @@ func (s *Server) onVariablesRequest(request *dap.VariablesRequest) {
 		}
 	}
 
-	var children []dap.Variable
+	children := []dap.Variable{} // must return empty array, not null, if no children
 	if request.Arguments.Filter == "named" || request.Arguments.Filter == "" {
 		named, err := s.metadataToDAPVariables(v)
 		if err != nil {
@@ -1769,7 +1769,7 @@ func (s *Server) childrenToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 	// TODO(polina): consider convertVariableToString instead of convertVariable
 	// and avoid unnecessary creation of variable handles when this is called to
 	// compute evaluate names when this is called from onSetVariableRequest.
-	var children []dap.Variable
+	children := []dap.Variable{} // must return empty array, not null, if no children
 
 	switch v.Kind {
 	case reflect.Map:
@@ -1917,7 +1917,7 @@ func getNamedVariableCount(v *proc.Variable) int {
 // metadataToDAPVariables returns the DAP presentation of the referenced variable's metadata.
 // These are included as named variables
 func (s *Server) metadataToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variable, error) {
-	var children []dap.Variable
+	children := []dap.Variable{} // must return empty array, not null, if no children
 
 	if isListOfBytesOrRunes(v.Variable) {
 		// Return the string value of []byte or []rune.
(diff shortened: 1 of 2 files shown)


polinasok:
@derekparker Can we still get this fix into milestone 1.7.0?

polinasok:
The failure doesn't look related. I have never seen one like this before.
```
2021-07-15T07:34:34Z debug layer=dap halting execution to set breakpoints
2021-07-15T07:34:34Z debug layer=debugger halting
2021-07-15T07:34:34Z debug layer=dap "next" command stopped - reason "manual", location ?:-1
2021-07-15T07:34:34Z error layer=dap runtime error: thread 3012 does not exist
2021-07-15T07:34:34Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"stopped","body":{"reason":"exception","description":"runtime error","text":"thread 3012 does not exist","allThreadsStopped":true}}
```

hyangah (approved):
LGTM
Thanks for providing the fix quickly.

I wonder if it's possible that go-dap layer offers some degree of extra type handling/checking by letting the type implement json.Marshaler eventually (e.g. translating nil to [] if nil isn't acceptable in spec).

polinasok:
> LGTM
> Thanks for providing the fix quickly.
> 
> I wonder if it's possible that go-dap layer offers some degree of extra type handling/checking by letting the type implement json.Marshaler eventually (e.g. translating nil to [] if nil isn't acceptable in spec).

Yes, I am working on that as well. There are other fields that need this check.

derekparker (approved):
LGTM

PR #2594, opened 2021-07-16
service/dap: send continued event before step response

suzmue:
Send the continued event before the step response to make sure that there is no time where the client believes that only a single thread is running.

Fixes golang/vscode-go#1629
Updates golang/vscode-go#1617

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1458,6 +1458,15 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
 // onNextRequest handles 'next' request.
 // This is a mandatory request to support.
 func (s *Server) onNextRequest(request *dap.NextRequest, asyncSetupDone chan struct{}) {
+	// All of the threads will be continued by this request, so we need to send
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -1458,20 +1458,47 @@ func (s *Server) onAttachRequest(request *dap.AttachRequest) {
 // onNextRequest handles 'next' request.
 // This is a mandatory request to support.
 func (s *Server) onNextRequest(request *dap.NextRequest, asyncSetupDone chan struct{}) {
+	// All of the threads will be continued by this request, so we need to send
```
polinasok, line 1461: how about a helper, so you don't repeat same code and comments 3 times?

suzmue:
Fixes golang/vscode-go#1629

polinasok (approved):

polinasok:
@derekparker Are we too late for 1.7.0 release? Vscode rolled out some UI fine tuning that exposed a timing bugs in our adapter that can be pretty annoying and confusing to the users.

aarzilli (approved):

PR #2595, opened 2021-07-16
service/dap: limit the number of goroutines returned from threads req

suzmue:
This adds a cap and a log message if there are many goroutines. This will help
prevent the debugger from freezing, but does not yet address making sure the
interesting goroutines are the ones that are returned.

Updates golang/vscode-go#129

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -189,6 +189,8 @@ const (
 	// what is presented. A common use case of a call injection is to
 	// stringify complex data conveniently.
 	maxStringLenInCallRetVars = 1 << 10 // 1024
+	// Max number of goroutines that we will return.
+	maxGoroutines = 1 << 10
 )
 
 // NewServer creates a new DAP Server. It takes an opened Listener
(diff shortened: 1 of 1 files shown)


aarzilli (approved):

derekparker (approved):
LGTM

polinasok (commented):
LGTM

polinasok:
Btw, is a risk of loosing the current goroutine if we just truncate? `Goroutines` does enforce that the current goroutine is first, does it?

suzmue:
> Btw, is a risk of loosing the current goroutine if we just truncate? `Goroutines` does enforce that the current goroutine is first, does it?

There is definitely a risk of losing the current goroutine, and I did not attempt to address that in this PR. It's possible to get around this by checking the threads and finding the IDs of the goroutines that are running, and appending them to the end of the list if they are not already included.

If we allow users to specify filters on what goroutines they want displayed, we can make it easier for them to only see the goroutines they want to inspect.

polinasok:
Can't you just rely on debugger state to get current goroutine and make sure it is in the returned list?

suzmue:
> Can't you just rely on debugger state to get current goroutine and make sure it is in the returned list?

Sure that sounds like it would work. Looking at the threads would allow us to make sure that any running goroutines were included if there was more than one.

polinasok:
FYI making sure that selected routine is included in the returned list was added in #2683

PR #2596, opened 2021-07-16
dwarf/line: Fix AllPCsBetween returning invalid PCs

### LINKED ISSUE #2593: DLV debugger assigns incorrect value as error (unreadable invalid interface type: key not found)
Version: 1.6.1
Build: $Id: 114218c22f3791287c4bc2f4ff35a846a1416ee9 $

go version go1.16.5 windows/amd64

Windows 10 x64

4. What did you do?
A simple call that return error = nil

error being nil

6. What did you see instead?
it is not nil and propagates as non-nil further when debugging
The error is
`(unreadable invalid interface type: key not found)`
The debugger becomes a bit "unstable", I don't fully understand why it jumps between multiple "return" from the caller. But 2 caller above the stack, it "recovers" as err == nil.

Using Visual Studio Code 1.58. I had problems with proc.go always opening, so I followed a recommendation and installed dlv for debugging and set this in my settings.
```
    "go.delveConfig": {
        "debugAdapter": "dlv-dap",
    }
```
This shows the error being non-nil
image

If I set the error with `var a error` instead of declaration assignation operator, it works.
image

This is using the old "legacy" debugger without the settings "debugAdapter". It works fine.
image

derekparker:
The is_stmt state machine register is set *after* the address, file, and
line gets updated in the line number program. Ensure that AllPCsBetween
correctly takes this into account so that we do not accidentally set
breakpoints on non-statement instructions.

Fixes #2593


derekparker:
@nd @artspb it seems like TeamCity is not picking up the latest https://github.com/go-delve/delve/blob/master/.teamcity/settings.kts. It's still running 1.14 tests and it's not including 1.17.

nd:
Probably due to missing comma: #2598

nd:
I've noticed that when one runs a build in aggregator its dependencies are reused and not re-executed. Is it convenient for you? If not it can be changed to always rerun a whole build chain.

cc @aarzilli

nd:
Temcity still doesn't apply settings, investigating

nd:
Ok now it does. I usually fix it by disabling versioned settings and then enabling them again. TC detect that there are settings in .teamcity dir in the repository and asks whether it should apply them. Accepting this fixed the issue.

aarzilli:
I need to look more into this but this looks wrong to me. If we had the sequence `DW_LNS_copy; DW_LNS_negate_stmt` and stmt starts with the value true the `DW_LNS_copy` should add the current PC address to the table with `stmt = true`, it shouldn't be skipped because it's followed by `DW_LNS_negate_stmt`, which should only affect the opcodes that follow it, not the ones that precede it. I think there's a different problem here, maybe we aren't tracking state validity correctly.

derekparker:
> which should only affect the opcodes that follow it, not the ones that precede it. I think there's a different problem here, maybe we aren't tracking state validity correctly.

Are you sure? The Dwarf standard appendix of line number information doesn't contain an example of is_stmt (at least in v4), however the standard states (emphasis mine):

> A boolean indicating that the *current instruction* is a recommended breakpoint location. A recommended breakpoint location is intended to “represent” a line, a statement and/or a semantically distinct subpart of a statement.

To me "current instruction" means the instruction the state machine is currently at, not the next one we're about to parse.

derekparker:
I’ll look into the CI failures on Monday.

derekparker:
Actually, the more I think about it and observe the change in behavior with this patch I believe you're right. I think the way we're currently tracking `is_stmt` is correct and we're just hitting the odd line number bug in the compiler.

I'm going to close this out in favor of an upstream fix.

PR #2597, opened 2021-07-16
service/dap: page stack frames

suzmue:
Returning monotonically increasing totalFrames values for subsequent requests can be used to enforce paging in the client. If we are not at the end of the stackFrames that are returned, we return a higher total frames to suggest to the client that they should request more frames.


pkg/proc/stack.go
```diff
@@ -93,19 +93,29 @@ func (frame *Stackframe) FramePointerOffset() int64 {
...
+	frames, _, err := ThreadStackframes(thread, 0, depth)
+	return frames, err
+}
+
+// ThreadStackframes returns the stack trace for thread.
+// Note the locations in the array are return addresses not call addresses.
+// ThreadStackframes also returns the next index to be used as 'start' argument
+// while scanning for all available stack frames, or -1 if there was an error
+// or if the index already reached the last possible value.
+func ThreadStackframes(thread Thread, start, depth int) ([]Stackframe, int, error) {
```
aarzilli, line 105: I don't like the Stacktrace/Stackframes split where one is a wrapper for calling the other. I think there should be only one of each pair. But also, why are you doing the thing with `next` instead of just keeping the iterator?

service/dap/server.go
```diff
@@ -1549,13 +1553,80 @@ type stackFrame struct {
 // As per DAP spec, this request only gets triggered as a follow-up
 // to a successful threads request as part of the "request waterfall".
 func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
+	// In order to support paging of stack frames, we always check an additional 10
+	// beyond the number of stack frames that were requested. Since the total frames
+	// will be larger than the number of frames loaded, the client will continue to
```
aarzilli, line 1558: What does "continue to request them" entail, precisely? Does the use have to do something to get more or will the client continue to ask for more until there aren't any more automatically? Because it could be infinite.

service/dap/server.go
```diff
@@ -1549,13 +1553,80 @@ type stackFrame struct {
...
+		totalFrames += 20
+	}
+	response := &dap.StackTraceResponse{
+		Response: *newResponse(request.Request),
+		Body:     dap.StackTraceResponseBody{StackFrames: stackFrames, TotalFrames: totalFrames},
+	}
+	s.send(response)
+}
+
+func (s *Server) loadFrames(goroutineID, start, levels int) ([]dap.StackFrame, int, bool, error) {
```
polinasok, line 1588: A function comment would be helpful. For example, is the returned int - total loaded frames (which would be either levels or less) or total frames in the entire stacktrace, including that are yet to be loaded?

service/dap/server.go
```diff
@@ -1549,13 +1553,80 @@ type stackFrame struct {
...
+		s.nextStackFrame[goroutineID][totalFrames] = next
+	}
+
+	if (start - startFrameIndex) < len(frames) {
+		frames = frames[(start - startFrameIndex):]
+	} else {
+		frames = frames[:0]
+	}
+
+	// Convert frames to DAP stack frames.
```
polinasok, line 1624: The name of the function says exactly that. I don't think you need the comment.

polinasok (changes requested):

service/dap/server.go
aarzilli: Ok.

pkg/proc/stack.go
aarzilli: IMHO this is introducing a lot of machinery, both in service/dap and in stack.go for what seems like a meager optimization. You could call `Stacktrace(goroutineID, request.Arguments.StartFrame + request.Arguments.Levels, 0)` and get the same effect. Since the stacktrace always starts from the top you are saving very little work (just loading inlined frames, of which there will be none in non-optimized code) and some memory (not saving frames before the start).

I think there are three better solutions possible:

1. Add a `skip` member to stackIterator, keep `proc.(*stackIterator).stacktrace` and `proc.(*stackIterator).appendInlinedCalls` identical except for the `append`s, when `it.skip > 0` instead of `append`ing it will decrement `it.skip`. This removes the nextStackFrame map from service/dap and the weird new loop in `(*stackIterator).stacktrace`, the `Stackframe/Stacktrace` split and the new return value they would need. You lose a small amount of work saved (which will be nothing anyway on optimized code) but keep the memory save and the `proc` functions have a reasonable APIs.
2. Keep the stackIterator. This gets rid of the `Stackframe/Stacktrace` split and you actually make stack loading not be O(n**2) (although, it probably doesn't matter given that it's driven by user input and will produce unreasonably long stacks long before it becomes slow)
3. Do not care about the performance and just call `Stacktrace(StartFrame+Levels)` always, maybe add a limit to `Levels - StartFrame` or something.

BTW you don't need to ask for more frames to determine if you have reached the bottom of the stack, if the last frame is an error or has `Bottom` set, then it's the bottom frame. You can just send `len(frames)+arbitraryNumber` to the client if the last frame isn't the bottom of the stack.

service/dap/server.go
```diff
@@ -1549,13 +1553,80 @@ type stackFrame struct {
...
+	}
+
+	stackFrames, totalFrames, hasMore, err := s.loadFrames(goroutineID, start, levels)
 	if err != nil {
 		s.sendErrorResponse(request.Request, UnableToProduceStackTrace, "Unable to produce stack trace", err.Error())
 		return
 	}
 
+	if hasMore {
+		totalFrames += 20
```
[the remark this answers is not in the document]
suzmue: Changed it to be s.args.stackTraceDepth instead. The client will request up to the number of frames that the server says it has remaining. If that number is very small, it will take many many requests to load all the stack frames.

pkg/proc/stack.go
suzmue: I made the change to load all of the frames up to start + levels and added the check if Bottom is set.

suzmue:
Here is a screen capture of stack frame paging:


https://user-images.githubusercontent.com/6634754/126826199-a540cbc6-bbcf-4ed3-84ce-02bd6f4968af.mov

service/dap/server_test.go
```diff
@@ -1997,7 +2078,7 @@ func TestLaunchRequestWithStackTraceDepth(t *testing.T) {
 				execute: func() {
 					client.StackTraceRequest(1, 0, 0)
 					stResp = client.ExpectStackTraceResponse(t)
-					checkStackFrames(t, stResp, "main.Increment", 8, 1000, 2 /*returned*/, 2 /*available*/)
+					checkStackFrames(t, stResp, "main.Increment", 8, 1000, 1 /*returned*/, 2 /*available*/)
```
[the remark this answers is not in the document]
suzmue: Changed back. The number of levels to be loaded should be s.args.stackTraceDepth + 1, not s.args.stackTraceDepth.

aarzilli:
TestLaunchRequestWithStackTraceDepth fails:

```
=== RUN   TestLaunchRequestWithStackTraceDepth
DAP server listening at: [::]:39689
2021-07-23T18:59:19Z debug layer=dap DAP server pid = 16389
Connecting to server at: [::]:39689
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":1,"type":"request","command":"initialize","arguments":{"adapterID":"go","locale":"en-us","linesStartAt1":true,"columnsStartAt1":true,"pathFormat":"path","supportsVariableType":true,"supportsVariablePaging":true,"supportsRunInTerminalRequest":true}}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":1,"success":true,"command":"initialize","body":{"supportsConfigurationDoneRequest":true,"supportsFunctionBreakpoints":true,"supportsConditionalBreakpoints":true,"supportsEvaluateForHovers":true,"supportsSetVariable":true,"supportsExceptionInfoRequest":true,"supportTerminateDebuggee":true,"supportsDelayedStackTraceLoading":true,"supportsClipboardContext":true}}
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":2,"type":"request","command":"launch","arguments":{"mode":"exec","program":"/tmp/increment.a443c8eb","stackTraceDepth":0}}
2021-07-23T18:59:19Z debug layer=dap debug backend is ''
2021-07-23T18:59:19Z debug layer=dap running program in /tmp

2021-07-23T18:59:19Z info layer=debugger launching process with args: [/tmp/increment.a443c8eb]
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"initialized"}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":2,"success":true,"command":"launch"}
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":3,"type":"request","command":"setBreakpoints","arguments":{"source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"breakpoints":[{"line":8}]}}
2021-07-23T18:59:19Z info layer=debugger created breakpoint: &api.Breakpoint{ID:1, Name:"sourceBp Path=\"/delve/_fixtures/increment.go\"  Line=8 Column=0", Addr:0x4ba494, Addrs:[]uint64{0x4ba494}, File:"/delve/_fixtures/increment.go", Line:8, FunctionName:"main.Increment", Cond:"", HitCond:"", Tracepoint:false, TraceReturn:false, Goroutine:false, Stacktrace:0, Variables:[]string(nil), LoadArgs:(*api.LoadConfig)(nil), LoadLocals:(*api.LoadConfig)(nil), WatchExpr:"", WatchType:0x0, HitCount:map[string]uint64{}, TotalHitCount:0x0, Disabled:false}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":3,"success":true,"command":"setBreakpoints","body":{"breakpoints":[{"id":1,"verified":true,"source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"line":8}]}}
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":4,"type":"request","command":"configurationDone","arguments":{}}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":4,"success":true,"command":"configurationDone"}
2021-07-23T18:59:19Z debug layer=debugger continuing
2021-07-23T18:59:19Z debug layer=dap "continue" command stopped - reason "breakpoint", location /delve/_fixtures/increment.go:8
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"stopped","body":{"reason":"breakpoint","threadId":1,"allThreadsStopped":true,"hitBreakpointIds":[1]}}
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":5,"type":"request","command":"stackTrace","arguments":{"threadId":1,"format":{}}}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":5,"success":true,"command":"stackTrace","body":{"stackFrames":[{"id":1000,"name":"main.Increment","source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"line":8,"column":0},{"id":1001,"name":"main.Increment","source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"line":11,"column":0},{"id":1002,"name":"main.Increment","source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"line":11,"column":0},{"id":1003,"name":"main.main","source":{"name":"increment.go","path":"/delve/_fixtures/increment.go"},"line":17,"column":0},{"id":1004,"name":"runtime.main","source":{"name":"proc.go","path":"/usr/local/go/go1.15.14/src/runtime/proc.go","presentationHint":"deemphasize"},"line":204,"column":0},{"id":1005,"name":"runtime.goexit","source":{"name":"asm_amd64.s","path":"/usr/local/go/go1.15.14/src/runtime/asm_amd64.s","presentationHint":"deemphasize"},"line":1374,"column":0}],"totalF
rames":6}}
    server_test.go:2098: 
        got  len(StackFrames)=6
        want 1
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":6,"type":"request","command":"continue","arguments":{"threadId":1}}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":6,"success":true,"command":"continue","body":{"allThreadsContinued":true}}
2021-07-23T18:59:19Z debug layer=debugger continuing
4
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"terminated","body":{}}
2021-07-23T18:59:19Z debug layer=dap [<- from client]{"seq":7,"type":"request","command":"disconnect","arguments":{"terminateDebuggee":true}}
2021-07-23T18:59:19Z debug layer=debugger halting
2021-07-23T18:59:19Z debug layer=dap halt returned state: Process 16772 has exited with status 0
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"output","body":{"category":"console","output":"Process 16772 has exited with status 0\n","source":{}}}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"output","body":{"category":"console","output":"Detaching\n","source":{}}}
2021-07-23T18:59:19Z debug layer=debugger detaching
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":7,"success":true,"command":"disconnect"}
2021-07-23T18:59:19Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"terminated","body":{}}
2021-07-23T18:59:19Z debug layer=dap DAP server stopping...
2021-07-23T18:59:19Z debug layer=dap DAP server stopped
--- FAIL: TestLaunchRequestWithStackTraceDepth (0.42s)
```

Also:

```
2021-07-23T19:02:20Z error layer=dap recovered panic: runtime error: index out of range [0] with length 0
goroutine 693 [running]:
runtime/debug.Stack(0xc000e870d8, 0x835160, 0xc000434a38)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/go-delve/delve/service/dap.(*Server).recoverPanic(0xc000ae00e0, 0x8d8408, 0xc000982840)
	/tmp/cirrus-ci-build/service/dap/server.go:398 +0x76
panic(0x835160, 0xc000434a38)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/go-delve/delve/service/debugger.(*Debugger).Launch(0xc000538f20, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x1, 0x1)
	/tmp/cirrus-ci-build/service/debugger/debugger.go:248 +0x6f6
github.com/go-delve/delve/service/debugger.New(0xc0006f5200, 0x0, 0x0, 0x0, 0xc000af3300, 0x809ba0, 0x1)
	/tmp/cirrus-ci-build/service/debugger/debugger.go:190 +0x5e5
github.com/go-delve/delve/service/dap.(*Server).onAttachRequest.func1(0xc000ae00e0, 0xc000e87790)
	/tmp/cirrus-ci-build/service/dap/server.go:1561 +0x89
github.com/go-delve/delve/service/dap.(*Server).onAttachRequest(0xc000ae00e0, 0xc000982840)
	/tmp/cirrus-ci-build/service/dap/server.go:1562 +0x3cc
github.com/go-delve/delve/service/dap.(*Server).handleRequest(0xc000ae00e0, 0x8d8408, 0xc000982840)
	/tmp/cirrus-ci-build/service/dap/server.go:575 +0x1b8a
github.com/go-delve/delve/service/dap.(*Server).serveDAPCodec(0xc000ae00e0)
	/tmp/cirrus-ci-build/service/dap/server.go:389 +0x105
github.com/go-delve/delve/service/dap.(*Server).Run.func1(0xc000ae00e0)
	/tmp/cirrus-ci-build/service/dap/server.go:354 +0xda
created by github.com/go-delve/delve/service/dap.(*Server).Run
	/tmp/cirrus-ci-build/service/dap/server.go:333 +0x3f
```

aarzilli (approved):

service/dap/server_test.go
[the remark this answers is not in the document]
suzmue: Ok I'm reconsidering changing it back.

Do we think setting s.args.stackTraceDepth to 1 should mean we get 1 frame back, or 2 frames back? I thought that getting 1 frame back would give the user the most control over how many frames they load, since they cannot set the stackTraceDepth to 0.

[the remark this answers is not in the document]
polinasok: I am pretty sure the math was based on what the legacy adapter had. And in that case we should have accepted 0 to result in 1 frame. But you are right, its likely more intuitive to just have depth be the number of frames.

service/dap/server.go
```diff
@@ -1688,36 +1700,35 @@ func (s *Server) onStackTraceRequest(request *dap.StackTraceRequest) {
 		isSystemGoroutine = g.System(s.debugger.Target())
 	}
 
-	stackFrames := make([]dap.StackFrame, len(frames))
-	for i, frame := range frames {
+	var stackFrames []dap.StackFrame
```
polinasok, line 1703: We had a null bug when variables array was no longer explicitly set to empty (fix at #2592). Can something like this happen here? I guess we don't accept 0 depth, but can we get an empty stack without an error? Might be safest to just initialize to empty just in case.

polinasok (approved):

derekparker (approved):
LGTM

PR #2600, opened 2021-07-19
service/dap: fix TestPreSetBreakpoint

aarzilli:
TestPreSetBreakpoint sometimes fail because a goroutine is stopped in
some runtime/internal package instead of just 'runtime'.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -515,7 +515,7 @@ func TestPreSetBreakpoint(t *testing.T) {
 		wantMain := dap.Thread{Id: 1, Name: "* [Go 1] main.Increment (Thread ...)"}
 		wantRuntime := dap.Thread{Id: 2, Name: "[Go 2] runtime.gopark"}
 		for _, got := range tResp.Body.Threads {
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

service/dap/server_test.go
```diff
@@ -515,7 +515,7 @@ func TestPreSetBreakpoint(t *testing.T) {
 		wantMain := dap.Thread{Id: 1, Name: "* [Go 1] main.Increment (Thread ...)"}
 		wantRuntime := dap.Thread{Id: 2, Name: "[Go 2] runtime.gopark"}
 		for _, got := range tResp.Body.Threads {
-			if got.Id != 1 && !reMain.MatchString(got.Name) && !strings.Contains(got.Name, "runtime.") {
+			if got.Id != 1 && !reMain.MatchString(got.Name) && !(strings.Contains(got.Name, "runtime.") || strings.Contains(got.Name, "runtime/")) {
```
polinasok, line 518: Maybe just it would make sense to drop both "." and "/" and just stick with "runtime"?

PR #2605, opened 2021-07-20
terminal: config -list should print strings in quotes

aarzilli:
Commit 8e91d3b0b added a number of configuration options to control the
colors of sytnax highlighting, unfortunately 'config -list' will print
all of those to stdout without quoting them, resulting in the color of
the last one being applied to all subsequent text.

Change 'config -list' to print strings in quotes so that we don't
accidentally send escape sequences to the terminal.

### DIFF
--- a/pkg/terminal/config.go
+++ b/pkg/terminal/config.go
@@ -81,13 +81,16 @@ func configureList(t *Term) error {
 			continue
 		}
 
-		if field.Kind() == reflect.Ptr {
+		switch field.Kind() {
+		case reflect.Ptr:
 			if !field.IsNil() {
 				fmt.Fprintf(w, "%s\t%v\n", fieldName, field.Elem())
 			} else {
 				fmt.Fprintf(w, "%s\t<not defined>\n", fieldName)
 			}
-		} else {
+		case reflect.String:
+			fmt.Fprintf(w, "%s\t%q\n", fieldName, field)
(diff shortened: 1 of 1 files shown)


aarzilli:
@nd do you have any clue why teamcity isn't publishing the PR build status anymore? 

```

Commit Status Publisher error. GitHub status: 'Failure'. Publisher: githubStatusPublisher(https://api.github.com). Failed to complete request to GitHub. Please check if the error is not returned by a proxy or caused by the lack of permissions. Status: HTTP/1.1 404 Not Found
```

nd:
@aarzilli no clue so far, is the teamcity access token still valid in github (https://github.com/settings/applications)?

aarzilli:
@derekparker can you check teamcity's token?

nd:
Derek changed the token in UI, but push of updated settings failed (looks like token specified in the VCS root is either invalid or doesn't have push permissions). Here is the update for dsl: #2607.

derekparker:
As mentioned elsewhere I updated the token at GitHubs request (new token format, more secure, etc, etc). I updated it in the UI but I seem to break things every time I touch the UI... as the config is now read only I think updating the settings.kts file will be better going forward. At least that way I can have @nd check things out before I break everything.

derekparker (approved):
LGTM

PR #2607, opened 2021-07-21
teamcity: specify working github token in dsl

nd:


aarzilli (approved):

nd:
I'm not sure what's going on, but now token specified in pull requests build feature fails with unauthorized error. Maybe tokens are changed in github settings?

aarzilli:
This changed only one token but there are three places that have a token, is that relevant?

nd:
I've changed only one token because others were working fine. If token in commit status publisher has necessary permissions in github, I can specify it in the other feature as well.

PR #2608, opened 2021-07-21
Make teamcity ui settings readonly

nd:
It looks like editing settings both in UI and in git leads to problems. This change disables settings editing in UI. If applied, all changes should be made by editing settings.kts. Feel free to reject if you want edit settings in UI, but in this case it is better to give teamcity permissions to push changes in settings to the repository, currently push fails. Otherwise every failed push disables settings synchronization.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -68,6 +68,9 @@ project {
         }))
     }
     buildType(AggregatorBuild(tests))
+    params {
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

derekparker:
Thanks. I had regenerated the token at GitHubs request and updated through the UI but I've always messed things up that way. Doing it this way will be better.

PR #2613, opened 2021-07-22
proc: make moduleDataToImage more robust

### LINKED ISSUE #2611: TestPluginVariables fails on go 1.17 linux/386

aarzilli:
Conversion form a moduledata object into an image object was
implemented by looking for a function covering the start address of the
text section of the moduledata object, and then converting that into
its corresponding image.

Unfortunately this seems to not always work. In particular it does not
work on linux/386 with go1.17 (but it might also fail on other
combinations): the start address of the text section is, for whatever
reason, not part of any function.

As a fallback simply scan all images we know of and return the closest
one that has start address less than or equal to the start address of
the text section we are looking for.

Fixes TestPluginVariables on go1.17/linux/386.

Fixes #2611


aarzilli:
Looks like it's breaking something else, TBD

aarzilli:
Fixed.

derekparker (approved):
LGTM

PR #2616, opened 2021-07-23
pkg/proc: Prefer throw instead of fatalthrow

### LINKED ISSUE #2602: TestFatalThrowBreakpoint systematically fails on Go 1.17
Is the version test for 1.16 supposed to be extended to 1.17? cc @suzmue

derekparker:
Currently there is a breakpoint set at runtime.fatalthrow to catch any
situation where the runtime crashes (e.g. deadlock).
When we do this, we go up a frame in order to parse the crash reason.
The problem is that there is no guarentee the "s" variable we attempt to
parse will still be considered "live".
Since runtime.fatalthrow is never called directly, set a breakpoint on
runtime.throw instead and prevent having
to search up a stack frame in order to
get the throw reason.

Fixes #2602

### DIFF
--- a/pkg/proc/amd64_arch.go
+++ b/pkg/proc/amd64_arch.go
@@ -219,7 +219,7 @@ func amd64SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
 		return true
 
 	default:
-		if it.systemstack && it.top && it.g != nil && strings.HasPrefix(it.frame.Current.Fn.Name, "runtime.") && it.frame.Current.Fn.Name != "runtime.throw" {
+		if it.systemstack && it.top && it.g != nil && strings.HasPrefix(it.frame.Current.Fn.Name, "runtime.") && it.frame.Current.Fn.Name != "runtime.throw" && it.frame.Current.Fn.Name != "runtime.fatalthrow" {
 			// The runtime switches to the system stack in multiple places.
 			// This usually happens through a call to runtime.systemstack but there
 			// are functions that switch to the system stack manually (for example
--- a/pkg/proc/amd64_arch.go
+++ b/pkg/proc/amd64_arch.go
@@ -219,7 +219,7 @@ func amd64SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
 		return true
 
 	default:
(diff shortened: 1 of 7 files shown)


derekparker:
The test in question is still failing on TeamCity because they are still using the Beta (`go1.17beta1 linux/amd64`), not the RC, and the beta does not have the relevant upstream fix.

I'm not sure about the other failures though, I'll investigate tomorrow.

pkg/proc/amd64_arch.go
```diff
@@ -219,7 +219,7 @@ func amd64SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
 		return true
 
 	default:
-		if it.systemstack && it.top && it.g != nil && strings.HasPrefix(it.frame.Current.Fn.Name, "runtime.") && it.frame.Current.Fn.Name != "runtime.fatalthrow" {
+		if it.systemstack && it.top && it.g != nil && strings.HasPrefix(it.frame.Current.Fn.Name, "runtime.") && it.frame.Current.Fn.Name != "runtime.throw" {
```
aarzilli, line 222: I think this shouldn't change. Just the place where we set the breakpoint.

derekparker: Without this change `TestCgoStacktraces2` will fail, we still need to prevent switching to the goroutine stack, and since we will no longer see `runtime.fatalthrow` in the stacktrace we have to check for `runtime.throw` instead. I could keep `runtime.fatalthrow` and just add this as an OR condition.

derekparker:
@nd it looks like the 1.17 TC builders are using the Go 1.17 beta1 release. Is there any way to force them to use the RC?

nd:
@derekparker hopefully #2619 does that

derekparker:
This patch fixes TestFatalThrowBreakpoint on 1.17 but it seems we have some flakiness in the Windows port on 1.17 as well.

pkg/proc/amd64_arch.go
aarzilli: Yes, I think adding an OR is the right solution. Even though we don't stop on runtime.fatalthrow deliberately anymore the program could end up stopped there anyway.

aarzilli (approved):
LGTM

aarzilli:
@polinasok @suzmue server/dap changes ok?

suzmue:
LGTM. 

Thanks for doing this fix!

polinasok (approved):

PR #2367, opened 2021-03-04
service/dap: Support for replay and core modes

lggomez:
This PR aims to add support for rr replay and core actions from the DAP layer. This basically encloses the following:

## New launch modes: replay and core

The following modes are added:

- **replay**: Replays an rr trace, allowing backwards flows (reverse continue and stepback). Requires a `traceDirPath` property on `launch.json` pointing to a valid rr trace directory. 
Equivalent to `dlv replay <tracedir>` command.
- **core**: Replays a core dump file, showing its callstack and the file matching the callsite. Requires a `coreFilePath` property on `launch.json` pointing to a valid coredump file.
Equivalent to `dlv core <exe> <corefile>` command.

## Dependencies

To achieve this the following additional changes were made:

- Implement the `onStepBackRequest` and `onReverseContinueRequest` methods on service/dap
- Adapt `onLaunchRequest` with the requried validations and logic for these new modes
- Use `CapabilitiesEvent` responses to enable the StepBack controls on the supported scenarios (see [dicussion here](https://github.com/go-delve/delve/pull/2367#discussion_r590975026))
- Add the corresponding launch.json support on vs code: https://github.com/golang/vscode-go/pull/1268

PR #2618, opened 2021-07-23
service/dap: fix backend parsing in replay mode

polinasok:
Replay mode configuration suffered during a rebase, and the issue was not caught by the unittests because there were none testing that replay mode args get set up and passed to `debugger.New` correctly. Because we added backend parsing (#2567) in parallel with replay/core support (#2367), `CoreFile` was set in `replay` mode, but `rr` backend was replaced with `"default"`, so this was treated as a `core` mode instead:
https://github.com/go-delve/delve/blob/658d36cb1909cd4267e538b10f2f84df8f752216/service/debugger/debugger.go#L167-L177

@lggomez

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -772,6 +772,20 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		return
 	}
 
+	backend, ok := request.Arguments["backend"]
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

PR #2621, opened 2021-07-26
proc: use signed comparison when searching image for module data

### LINKED ISSUE #2620: Flickering tests on Windows and Go 1.17
There are several flickering tests on Windows for Go 1.17. The tests usually fail with something like: `unreadable could not resolve interface type`.

aarzilli:
StaticBase is the difference between the entry point declared in the
image file and the entry point as loaded in memory, since this
difference could be a negative number we have to use a signed
comparison when searching for a mapping.

Causes intermittent test failures on windows when resolving interface
types for position independent executables.

Fixes #2620


derekparker (approved):
LGTM

PR #2622, opened 2021-07-26
service/dap: make next while nexting error more clear

suzmue:
To make it more clear that the user can resume the program when they encounter a next while nexting error, make the exception information have instructions for resuming the program. This implements the suggestions outlined by @polinasok in https://github.com/go-delve/delve/pull/2443#issuecomment-869455021.

Screen Shot 2021-07-26 at 5 20 28 PM

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2837,7 +2837,8 @@ func newEvent(event string) *dap.Event {
 
 const BetterBadAccessError = `invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation]
 Unable to propagate EXC_BAD_ACCESS signal to target process and panic (see https://github.com/go-delve/delve/issues/852)`
-const BetterNextWhileNextingError = "Step was interrupted by a breakpoint, cannot issue another step request. Use `Continue` to resume the program."
+const BetterNextWhileNextingError = `Unable to step while the previous step is interrupted by a breakpoint.
+Use 'Continue' to resume the original step command.`
 
 func (s *Server) resetHandlesForStoppedEvent() {
 	s.stackFrameHandles.reset()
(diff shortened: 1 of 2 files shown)


aarzilli (approved):

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -2819,6 +2837,7 @@ func newEvent(event string) *dap.Event {
 
 const BetterBadAccessError = `invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation]
 Unable to propagate EXC_BAD_ACCESS signal to target process and panic (see https://github.com/go-delve/delve/issues/852)`
+const BetterNextWhileNextingError = "Step was interrupted by a breakpoint, cannot issue another step request. Use `Continue` to resume the program."
```
polinasok, line 2840: Suggest slightly different wording and formatting with multiple lines:
```
`Unable to step while the previous step is interrupted by a breakpoint.
Use `Continue` to resume the original step command.`
```

polinasok (changes requested):

polinasok (approved):
This looks good.
Do you think we should also print a message to the console when the breakpoint is hit that step was interrupted and that Continue can be used to resume, so people see this before they get an exception? I guess we don't keep track of which operation was interrupted, so this might not be straightforward.

suzmue:
> Do you think we should also print a message to the console when the breakpoint is hit that step was interrupted and that Continue can be used to resume, so people see this before they get an exception? I guess we don't keep track of which operation was interrupted, so this might not be straightforward.

Added a message when the step is interrupted by a breakpoint (state.NextInProgress is true when this is the case).

PR #2625, opened 2021-07-27
*: Initial eBPF tracing support

derekparker:
This patch adds support for using eBPF (https://ebpf.io/) as the underlying technology for the `dlv trace` subcommand. This initial patch adds support only for Linux, but Windows can be supported as well eventually. This eBPF based implementation is *significantly* faster than the ptrace based one.

For example with a local test program:

Program execution: 23.7µs
w/ eBPF trace: 683.1µs
w/ ptrace tracing: 2.3s

Since this patch relies on an external dependency which needs a C lib it is currently gated behind a build tag and must be explicitly enabled by the user. The feature is also incomplete and should be considered in the "experimental" stage for now.

The eBPF based tracer knows how to parse certain function arguments such as basic ints, floats, etc as well as strings. Future patches will add improved support for slices, structs, interfaces, etc. Another limitation of this current patch is it does not support return arguments. I plan on supporting return arguments in the future via uretprobes, but this involves monitoring the runtime to be sure we remove / reset uretprobes when the runtime grows and copies the stack (see https://github.com/golang/go/issues/22008). This will involve a good amount of new code, and as this patch was already growing large I wanted to submit a first working implementation and simply build off of that in future patches.

The current patch supports both the stack based *and* the register based ABI for Go for function arguments.


derekparker:
There is still more to be added in terms of documentation, but I figured I'd submit this now since it's working so I can start gathering feedback.

derekparker:
(cc @grantseltzer since we had spoken about this)

pkg/proc/breakpoints.go
```diff
@@ -384,6 +390,60 @@ func (t *Target) SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr)
...
+	for _, entry := range varEntries {
+		isret, _ := entry.Val(dwarf.AttrVarParam).(bool)
+		if isret {
+			continue
+		}
+		_, dt, err := readVarEntry(entry.Tree, fn.cu.image)
+		if err != nil {
+			return err
+		}
+		offset, pieces, _, err := t.BinInfo().Location(entry, dwarf.AttrLocation, fn.Entry, *dregs)
```
aarzilli, line 430: I think you should be able to pass `op.DwarfRegisters{}` here. That's better because we are not using the registers of the current thread that have nothing to do with what we are doing anyway.

pkg/proc/eval.go
```diff
@@ -427,6 +427,8 @@ func (scope *EvalScope) LocalVariables(cfg LoadConfig) ([]*Variable, error) {
 }
 
 // FunctionArguments returns the name, value, and type of all current function arguments.
+// If the cfg LoadConfig parameter is nil the values of the function arguments will not
```
aarzilli, line 430: I think this is a leftover change, cfg can't be nil.

pkg/proc/core/core.go
```diff
@@ -269,6 +270,13 @@ func (p *process) Checkpoints() ([]proc.Checkpoint, error) { return nil, nil }
 // ClearCheckpoint clears a checkpoint, but will only return an error for core files.
 func (p *process) ClearCheckpoint(int) error { return errors.New("checkpoint not found") }
 
+func (p *process) SupportsBPF() bool {
+	return false
+}
+
+func (dbp *process) SetUProbe(fnName string, args []ebpf.UProbeArgMap) {
```
aarzilli, line 277: `panic("not implemented")` just in case?

pkg/proc/internal/ebpf/helpers.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/go-delve/delve/pkg/dwarf/op"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+	"github.com/aquasecurity/libbpfgo/helpers"
+)
+
+//go:embed trace_probe/trace.o
+var TraceProbeBytes []byte
+
+const FakeAddressBase = 0xbeef000000000000
```
aarzilli, line 27: For consistency with the other code this should be fakeAddressUnresolv (0xbeed000000000000) because we are using the 0xbeef000000000000 to describe registerized variables whose CompositeMemory is cached.

pkg/proc/internal/ebpf/helpers_disabled.go
```diff
@@ -0,0 +1,47 @@
+//go:build !ebpf
+// +build !ebpf
+
+package ebpf
+
+// #include "./trace_probe/function_vals.bpf.h"
```
aarzilli, line 6: Do we need this? I wouldn't expect to need a C compiler if eBPF was disabled.

pkg/proc/internal/ebpf/trace_probe/trace.bpf.c
```diff
@@ -0,0 +1,165 @@
...
+    ret = bpf_probe_read_user(&param->val, param->size, (void *)(addr));
+    if (ret < 0) {
+        return 1;
+    }
+    return 0;
+}
+
+__always_inline
+void get_value_from_register(struct pt_regs *ctx, void *dest, int reg_num) {
+    switch (reg_num) {
```
aarzilli, line 44: Are the other registers not there?

pkg/proc/mem.go
```diff
@@ -92,6 +96,18 @@ type compositeMemory struct {
...
 }
 
+// CreateCompositeMemory created a new composite memory type using the provided MemoryReadWriter as the
+// underlying memory buffer.
+func CreateCompositeMemory(mem MemoryReadWriter, arch *Arch, regs op.DwarfRegisters, pieces []op.Piece) (*compositeMemory, error) {
+	// This is basically a small wrapper to avoid having to change all callers
+	// of newCompositeMemory since it existed first.
+	cm, err := newCompositeMemory(mem, arch, regs, pieces)
+	if cm != nil {
+		cm.base = FakeAddressBase
```
aarzilli, line 106: Same as above, it should be fakeAddressUnresolv.

pkg/proc/native/proc_linux.go
```diff
@@ -45,6 +47,14 @@ const (
 // process details.
 type osProcessDetails struct {
 	comm string
+
+	ctx *ebpf.EBPFContext
```
aarzilli, line 51: Given where this is place it should have a more descriptive name, like ebpfCtx or just ebpf.

pkg/proc/native/proc_linux.go
```diff
@@ -183,9 +193,17 @@ func initialize(dbp *nativeProcess) error {
 		comm = match[1]
 	}
 	dbp.os.comm = strings.ReplaceAll(string(comm), "%", "%%")
+
 	return nil
 }
 
+func (dbp *nativeProcess) GetBufferedTracepoints() []ebpf.RawUProbeParams {
+	if dbp.os.ctx == nil {
+		return make([]ebpf.RawUProbeParams, 0)
```
aarzilli, line 202: Why not nil?

pkg/proc/native/proc_linux.go
```diff
@@ -689,6 +706,46 @@ func (dbp *nativeProcess) EntryPoint() (uint64, error) {
 	return linutil.EntryPointFromAuxv(auxvbuf, dbp.bi.Arch.PtrSize()), nil
 }
 
+func (dbp *nativeProcess) SupportsBPF() bool {
+	return true
```
aarzilli, line 710: This return value should depend on the ebpf tag.

pkg/proc/native/proc_linux.go
```diff
@@ -689,6 +706,46 @@ func (dbp *nativeProcess) EntryPoint() (uint64, error) {
...
+	// We only allow up to 6 args for a BPF probe.
+	// Return early if we have more.
+	if len(args) > 6 {
+		return
+	}
+
+	debugname := dbp.bi.Images[0].Path
+	offset, err := ebpf.SymbolToOffset(debugname, fnName)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
aarzilli, line 728: I think it would be better to return the error rather than calling os.Exit.

pkg/proc/native/proc_linux.go
```diff
@@ -689,6 +706,46 @@ func (dbp *nativeProcess) EntryPoint() (uint64, error) {
...
+		os.Exit(-1)
+	}
+	err = dbp.os.ctx.AttachUprobe(dbp.Pid(), debugname, offset)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+	fn, ok := dbp.bi.LookupFunc[fnName]
+	if !ok {
+		panic("could not find function")
```
aarzilli, line 738: Same as above, better to return an error than panicking.

pkg/proc/variables.go
```diff
@@ -734,6 +734,11 @@ func resolveTypedef(typ godwarf.Type) godwarf.Type {
 	}
 }
 
+// SetMemory sets the memory for the variable.
+func (v *Variable) SetMemory(mem MemoryReadWriter) {
```
aarzilli, line 738: Is this a leftover? I can't find where it's called.

service/debugger/debugger.go
```diff
@@ -757,6 +757,13 @@ func isBreakpointExistsErr(err error) bool {
 	return r
 }
 
+func (d *Debugger) CreateTracepoint(fnName string) error {
```
aarzilli, line 760: I'm not sure about the name of this function, it seems confusing with the old meaning of Tracepoint. `CreateBPFTracepoint`, `CreateBufferedTracepoint`, `CreateAsyncTracepoint`, `CreateFastTracepoint`? Same for `SetTracepoint`.

aarzilli (commented):
procNative for darwin and freebsd are missing the GetBufferedTracepoints method, hence the compilation errors

> I plan on supporting return arguments in the future via uretprobes, but this involves monitoring the runtime to be sure we remove / reset uretprobes when the runtime grows and copies the stack

This was also needed for stack watchpoints, see `WatchOutOfScopeBreakpoint` code in the stack watchpoint PR.

pkg/proc/breakpoints.go
derekparker: Done.

pkg/proc/eval.go
derekparker: That's correct. Removed.

pkg/proc/core/core.go
derekparker: Done.

pkg/proc/internal/ebpf/helpers.go
derekparker: Done.

pkg/proc/internal/ebpf/helpers_disabled.go
derekparker: We still need the definition of `C.function_parameter_list_t` for the `CreateFunctionParameterList` function.

pkg/proc/internal/ebpf/trace_probe/trace.bpf.c
derekparker: They are, I was just being lazy and selfishly implemented only the ones I needed ;). Added the rest.

pkg/proc/mem.go
derekparker: Done.

pkg/proc/native/proc_linux.go
derekparker: Done.

pkg/proc/native/proc_linux.go
derekparker: Done.

pkg/proc/native/proc_linux.go
derekparker: Done.

pkg/proc/native/proc_linux.go
derekparker: Done.

pkg/proc/native/proc_linux.go
derekparker: Done.

pkg/proc/variables.go
derekparker: It is, removed.

service/debugger/debugger.go
derekparker: Done.

derekparker:
> > I plan on supporting return arguments in the future via uretprobes, but this involves monitoring the runtime to be sure we remove / reset uretprobes when the runtime grows and copies the stack
> 
> This was also needed for stack watchpoints, see `WatchOutOfScopeBreakpoint` code in the stack watchpoint PR.

Ack, that's right. I'll take another look at that code and then I can probably get uretprobes working with this CL unless you'd rather that still be a follow up if it's make it easier to review.

aarzilli:
> Ack, that's right. I'll take another look at that code and then I can probably get uretprobes working with this CL unless you'd rather that still be a follow up if it's make it easier to review.

I think it's better to merge it as is and then do a followup, less danger of merge conflicts. Eventually it will also need some tests.

pkg/proc/internal/ebpf/helpers_disabled.go
aarzilli: What if it was called internally by UpdateArgMap so it didn't need to be in the disabled version?

derekparker:
> > Ack, that's right. I'll take another look at that code and then I can probably get uretprobes working with this CL unless you'd rather that still be a follow up if it's make it easier to review.
> 
> I think it's better to merge it as is and then do a followup, less danger of merge conflicts. Eventually it will also need some tests.

Sounds good to me. And yeah, I'll be including tests in the next patch.

service/debugger/debugger.go
aarzilli: I meant the name in the JSON-RPC API should also change.

pkg/proc/internal/ebpf/helpers_disabled.go
derekparker: That's a great refactor. Done.

aarzilli:
Is the binary file trace.o being committed on purpose? Otherwise LGTM.

aarzilli (approved):

derekparker:
> Is the binary file trace.o being committed on purpose? Otherwise LGTM.

No, it can be build locally and `make build-bpf` will build it so I think it's fine not to commit it to the repo. In the next patch set I'll add documentation on installing deps necessary to build it.

PR #2627, opened 2021-07-28
Fix crashes on Go dev.typeparams (soon to be Go main branch)

aclements:
We're making a few cleanups to the _defer struct on the dev.typeparams branch that currently cause Delve to crash. The plan is to merge dev.typeparams to the Go main branch as soon as 1.18 development opens, so this PR is getting ahead of that (also, we like to be able to use Delve ourselves :). It inspects the struct's type, so it should adapt smoothly to these changes without having to be keyed off the Go version.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #2628, opened 2021-07-29
proc: move breakpoint condition evaluation out of backends

aarzilli:
Moves breakpoint condition evaluation from the point where breakpoints
are set, inside ContinueOnce, to (*Target).Continue.

This accomplishes three things:

1. the breakpoint evaluation method needs not be exported anymore
2. breakpoint condition evaluation can be done with a full scope,
containing a Target object, something that wasn't possible before
because ContinueOnce doesn't have access to the Target object.

3. moves breakpoint condition evaluation out of the critical section
where some of the threads of the target process might be still
running.

### DIFF
--- a/_scripts/make.go
+++ b/_scripts/make.go
@@ -353,10 +353,30 @@ func testStandard() {
 		fmt.Println("\nTesting RR backend")
 		testCmdIntl("basic", "", "rr", "normal")
 	}
-	if TestIncludePIE && (runtime.GOOS == "linux" || (runtime.GOOS == "windows" && goversion.VersionAfterOrEqual(runtime.Version(), 1, 15))) {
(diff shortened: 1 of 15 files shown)


aarzilli:
Looks like it's causing tests on linux/386 to fail. Will look into it.

aarzilli:
The change in timing uncovered some problems with the way we stop threads on Windows, which I fixed on a separate commit. Also TestIssue1101 sometimes fails on linux, this is also not caused by this change but it makes it more frequent, I can't reproduce it locally and from investigating it in teamcity it doesn't look like we can do anything about it.

derekparker (approved):
LGTM

derekparker:
I'll ignore the deepsource errors.

PR #2629, opened 2021-07-29
cmd: refactor building logic into a helper

polinasok:
This refactoring reduces code duplication and makes the helper available to be used by dlv-dap command in the future.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -472,18 +472,30 @@ func dapCmd(cmd *cobra.Command, args []string) {
 	os.Exit(status)
 }
 
+func buildBinary(cmd *cobra.Command, args []string, isTest bool) string {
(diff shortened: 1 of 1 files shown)


polinasok:
I am looking into the test failure.

cmd/dlv/cmds/commands.go
```diff
@@ -472,18 +472,30 @@ func dapCmd(cmd *cobra.Command, args []string) {
 	os.Exit(status)
 }
 
+func buildBinary(cmd *cobra.Command, args []string, isTest bool) string {
```
hyangah, line 475: How about making this to return (string, bool) or (string, error)? 
https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors

polinasok: Good point. Done.

aarzilli (approved):

derekparker (approved):
LGTM

PR #2632, opened 2021-07-30
pkg/proc: add tests for next interrupted by breakpoint

suzmue:
Adds tests that make sure that when a next request is interrupted by a breakpoint, the stepping breakpoints are cleared.


derekparker (approved):
LGTM

PR #2633, opened 2021-07-30
pkg/terminal: Allow fuzzy searching tab completions

derekparker:
This patch implements fuzzy searching for tab completions in the
terminal client. Under the hood it is using a trie data structure
(https://en.wikipedia.org/wiki/Trie) to perform very fast prefix / fuzzy
searches.

https://user-images.githubusercontent.com/1314079/127693894-b53c1cf0-bedb-4a49-909b-b3d7e5d5555d.mp4


pkg/terminal/terminal.go
```diff
@@ -210,22 +211,36 @@ func (t *Term) Run() (int, error) {
...
+		fns.Add(fn, nil)
+	}
+	for _, cmd := range t.cmds.cmds {
+		for _, alias := range cmd.aliases {
+			cmds.Add(alias, nil)
+		}
+	}
+
 	t.line.SetCompleter(func(line string) (c []string) {
+		prefix := ""
```
aarzilli, line 227: I think we should look use the code in `Commands.Find` to find out which command it is, like CallWithContext does. And then use the `command` object to decide what to do.

pkg/terminal/terminal.go
```diff
@@ -210,22 +211,36 @@ func (t *Term) Run() (int, error) {
...
 		if strings.HasPrefix(line, "break ") || strings.HasPrefix(line, "b ") {
```
aarzilli, line 228: We should also check for `trace`, it didn't do it before but clearly it's wrong.

pkg/terminal/terminal.go
```diff
@@ -210,22 +211,36 @@ func (t *Term) Run() (int, error) {
...
+			prefix = "break "
+		}
+		if strings.HasPrefix(line, "continue ") || strings.HasPrefix(line, "c ") {
+			prefix = "continue "
+		}
+		if prefix != "" {
 			filter := line[strings.Index(line, " ")+1:]
```
aarzilli, line 235: IMHO we should make this work better:

```
if spc := strings.LastIndex(line, " "); spc > 0 {
    prefix := line[:spc]
    funcs := fns.FuzzySearch(line[spc+1:])
    for _, f := range funcs {
         c = append(c, prefix+f)
    }
}

```

aarzilli: This way it will also work with `break bpname fmt.Pri<TAB>`, now it doesn't.

derekparker: Done.

pkg/terminal/terminal.go
derekparker: Done.

pkg/terminal/terminal.go
derekparker: I'm not sure what this buys us. `Commands.Find` returns a function object, so I suppose we could type switch on that?

aarzilli: I was thinking of changing `Commands.Find` to return the `command` object and then look into that (or make another method that gets called both by `Commands.Find` and here).

Regarding what this buys us is that it will work with user defined aliases instead of a hardcoded list.

derekparker: Ah, I see now. Done.

aarzilli (approved):

PR #2635, opened 2021-08-02
pkg/proc: configure target to not clear stepping breakpoints

suzmue:
In order for DAP to support halting the program (either manually or on a breakpoint) performing some action and then resuming execution, there needs to be a way to stop the program without clearing the internal breakpoints. This is necessary for log points and stopping the program to set breakpoints.

The debugging UI makes it seem like a user should be able to set or clear a breakpoint at any time. Adding this ability to complete synchronous requests while the program is running is thus important to create a seamless user experience.

This change just adds a configuration to determine whether the target should clear the stepping breakpoints, and changes the server to use this new mode. Using the new mode means that the DAP server must determine when it expect the next to be canceled and do this manually.


pkg/proc/proc_test.go
```diff
@@ -2562,6 +2562,31 @@ func TestStepOutBreakpoint(t *testing.T) {
 	})
 }
 
+func TestStepOutBreakpointKeepsSteppingBreakpoints(t *testing.T) {
```
aarzilli, line 2565: This is testing the exact same code as TestNextBreakpointKeepsSteppingBreakpoints, remove.

pkg/proc/target.go
```diff
@@ -46,6 +46,10 @@ type Target struct {
 	// CanDump is true if core dumping is supported.
 	CanDump bool
 
+	// KeepsSteppingBreakpoints is true if the target will not clear
```
aarzilli, line 49: s/if/then/
s/unless/until/

pkg/proc/target_exec.go
```diff
@@ -89,7 +93,7 @@ func (dbp *Target) Continue() error {
 			}
 			return err
 		}
-		if dbp.StopReason == StopLaunched {
+		if dbp.StopReason == StopLaunched && !dbp.KeepsSteppingBreakpoints {
```
aarzilli, line 96: The behavior of this is going to be very weird. This code is executed in response to a reverse step request, which results in the recording reaching the beginning of the process execution. 
The reverse step request can never finish if this happens.

service/dap/server.go
```diff
@@ -1411,6 +1411,12 @@ func (s *Server) onConfigurationDoneRequest(request *dap.ConfigurationDoneReques
 		}
 		s.send(e)
 	}
+	func() {
```
aarzilli, line 1414: It is not necessary to lock the target for this here, I think.

pkg/proc/target_exec.go
```diff
@@ -1014,8 +1018,8 @@ func stepOutReverse(p *Target, topframe, retframe Stackframe, sameGCond ast.Expr
 	return err
 }
 
-// onNextGoroutine returns true if this thread is on the goroutine requested by the current 'next' command
-func onNextGoroutine(thread Thread, breakpoints *BreakpointMap) (bool, error) {
+// OnNextGoroutine returns true if this thread is on the goroutine requested by the current 'next' command
+func OnNextGoroutine(thread Thread, breakpoints *BreakpointMap) (bool, error) {
```
aarzilli, line 1022: I'm not fond of exposing this function. Maybe KeepsSteppingBreakpoints should be more granular than just a boolean.

suzmue: Good point.

The two cases for DAP that we want to keep the stepping breakpoint are:
1. halt request
2. log points (which can be marked as tracepoints)

I can change this to be a setting that is more granular to cover these two cases.

Though I wonder if tracepoints should actually never clear stepping breakpoints?

pkg/proc/target_exec.go
suzmue: I think your point about making KeepsSteppingBreakpoints more granular is a good one. I can change this to only apply for halt and tracepoint and take out this check here.

pkg/proc/proc_test.go
suzmue: Done.

pkg/proc/target_exec.go
aarzilli: So:

```
type KeepSteepingBreakpoints uint8
const` (
    HaltKeepsSteppingBreakpoints KeepSteppingBreakpoints = 1 << iota
    TracepointKeepsSteppingBreakpoints
    // others?
)
```

> Though I wonder if tracepoints should actually never clear stepping breakpoints?

I think we could do that. The argument for keeping it this way is that it's been like this for a long time, but I doubt it would hurt anything.

aarzilli (approved):

service/dap/server.go
derekparker: I see you've approved this with this comment outstanding. Would you like this changed? Can anything ever race for this?

pkg/proc/target_exec.go
suzmue: I switched it to preserver the old behavior for tracepoints.  and updated the configuration to match @aarzilli comment above.

service/dap/server.go
aarzilli: Yes, I still think it's unnecessary.

suzmue: There shouldn't be an issue with adjusting this without locking, so I'll remove the lock.

derekparker (approved):
LGTM

PR #2640, opened 2021-08-03
Documentation: update installation doc

hyangah:
Specify the required go version.
Utilize go in modules mode.
Separate stable vs unstable version installaiton steps.

### DIFF
--- a/Documentation/installation/README.md
+++ b/Documentation/installation/README.md
@@ -1,9 +1,5 @@
 # Installation
 
-## Installing the latest stable version
-
-The latest stable version requires go 1.12 or a later version.
-
 With go version 1.16 or later:
 
 ```
@@ -12,12 +8,19 @@ $ go install github.com/go-delve/delve/cmd/dlv@latest
 
 With go version 1.15 or older:
 
-Change to any directory that is both outside of your `GOPATH`
-and outside of a module (e.g. a temp directory), and run:
+Clone the git repository and build:
 
 ```
-$ GO111MODULE=on go get github.com/go-delve/delve/cmd/dlv@latest
+$ git clone https://github.com/go-delve/delve
+$ cd delve
+$ go install github.com/go-delve/delve/cmd/dlv
 ```
+
+Note that the above instruction for go 1.15 or older will
+build `dlv` from the master branch and the produced binary
+will not contain complete build information.
+When you report an issue, make sure to check if the
+problem can be reproducible from the current master.
 
 See `go help install` for details on where the `dlv` executable is saved.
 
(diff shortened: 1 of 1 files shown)


aarzilli:
I'd rather wait 6 months before doing this, installing by cloning the repository and building works on all versions of go, the `install latest` version is still too recent and recommending `GO111MODULES=on go get` requires the caveat of having to move to a directory that doesn't contain a module.

In 6 months we won't officially support any version that doesn't support the global install command so this is more tenable.

Documentation/installation/README.md
```diff
@@ -1,20 +1,25 @@
 # Installation
 
-Clone the git repository and build:
+## Installing the latest stable version
+
+The latest stable version requires go 1.12 or a later version.
```
aarzilli, line 5: The documentation intentionally omits to specify the minimum version of go because historically we forget to update it when it changes.

aarzilli:
PS. since this is probably related to the recent build breakage on master I wanted to point out that: (1) it has only happened a few times in 6 years so it's not that bad and (2) two out of three people reporting the problem were not following the install instructions anyway.

Documentation/installation/README.md
```diff
@@ -1,20 +1,25 @@
...
 
-On Go version 1.16 or later, this command will also work:
+With go version 1.15 or older:
+
+Change to any directory that is both outside of your `GOPATH`
+and outside of a module (e.g. a temp directory), and run:
 
 ```
-$ go install github.com/go-delve/delve/cmd/dlv@latest
+$ GO111MODULE=on go get github.com/go-delve/delve/cmd/dlv@latest
```
hyangah, line 19: This works for all go 1.12+ and produces reproducible builds. Go 1.12 was released in early 2019, which was 2+ years ago and the method using `git clone` won't work with earlier versions of go any way.

hyangah:
> PS. since this is probably related to the recent build breakage on master I wanted to point out that: (1) it has only happened a few times in 6 years so it's not that bad and (2) two out of three people reporting the problem were not following the install instructions anyway.

I got a separate internal report about build breakage (cloud code) and it wasn't their intention to test the bleeding edge version of dlv :-). Or, is it intentional to promote usage of dlv built from master rather than the tagged version? If so, I think the `go install` command should be changed to pick the `master`. e.g. `go install github.com/go-delve/delve/cmd/dlv@master`.

Many people use `go get` to build dlv and in the current page. Showing them how to use the command correctly is beneficial in my opinion.

aarzilli:
> Or, is it intentional to promote usage of dlv built from master rather than the tagged version?

I'm personally indifferent to this.

> Many people use go get to build dlv and in the current page. Showing them how to use the command correctly is beneficial in my opinion.

Isn't go get going to be deprecated anyway in the future? Why should we keep telling people about an install method that will not continue to work and, depending where it's used, doesn't work anyway.

hyangah:
Yes, `go get` for tool install will be deprecated in go1.18 in favor of `go install`.

Unfortunately, `git clone + go install` does not produce dlv with the real dlv version info. As seen in multiple issues before, we know `dlv version` is misleading and `dlv version -v` or `go version -m` won't do a better job for the binary built in this way. 

It sounds like reproducible build has not been a big concern in maintaining delve. I tried to mention the caveat here in the revision but just keep asking users to rebuild dlv from the master before reporting an issue (maybe add to the issue template?) may not add too much work either. I will close this PR.

PR #2641, opened 2021-08-04
service/dap: fix race condition in TestContinueOnEntry

aarzilli:
The Threads request sent immediately after the ConfigurationDone
request can be processed in two possible states:

* while the program is being executed, when it will return Id=1,
Name=Dummy
* after the program terminates, when it will return Id=-1, Name=Current

additionally the response could be received in any order with the
Terminate event.

Remove the problematic Threads request from the test.


derekparker (approved):
LGTM

polinasok (approved):

PR #2642, opened 2021-08-04
service: fix sameuser check

aarzilli:
Change the socket search to check both the remote and local fields of the
socket match the socket we want to find.

Sockets are identified by the 4-uple

local_addr, local_port, remote_addr, remote_port

Two socket can differ by a single one of this four elements.

It is possible for the same local_port to be used by two different sockets,
as long as they are connecting to different remote addresses (or remote
ports).

An example of this bug in action can be seen at:

https://github.com/golang/vscode-go/runs/3141270564?check_suite_focus=true

There the server starts listening on 127.0.0.1:46011 and rejects a valid
client connection by finding the following socket:

60: 0100007F:DD82 0100007F:962D 06 00000000:00000000 03:00000133 00000000     0        0 0 3 0000000000000000

the local address of this socket is 0100007F:DD82 (127.0.0.1:56706), and the
remote address is 0100007F:962D (127.0.0.1:38445).

The reported error is:

closing connection from different user (127.0.0.1:56706): connections to localhost are only accepted from the same UNIX user for security reasons

note how the local port does match the socket line (56706) but the remote
port is wrong (38445 instead of 46011).

Note also that the state of this socket is 06, or TIME_WAIT, which would be
impossible if this was the right socket, since the right socket would still
be open.

Fixes https://github.com/golang/vscode-go/issues/1555

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -342,7 +342,7 @@ func (s *Server) Run() {
 			return
 		}
 		if s.config.CheckLocalConnUser {
-			if !sameuser.CanAccept(s.listener.Addr(), conn.RemoteAddr()) {
+			if !sameuser.CanAccept(s.listener.Addr(), conn.LocalAddr(), conn.RemoteAddr()) {
 				s.log.Error("Error accepting client connection: Only connections from the same user that started this instance of Delve are allowed to connect. See --only-same-user.")
 				s.triggerServerStop()
 				return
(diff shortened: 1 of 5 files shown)


derekparker (approved):
LGTM

derekparker:
@aarzilli looks like we're getting a Windows failure. Do we just need a rebase to pick up a fix?

PR #2643, opened 2021-08-04
service/dap: fix TestNextParked/TestStepInParked

aarzilli:
The loop searching for a suitable goroutine is not guaranteed to ever
find it, and failing to find one is not an error.

Changes testStepParkedHelper to match the behavior of TestNextParked in
proc_test.go.

Deletes TestStepInParked because it does not test anything meaningful
beyond what's already tested by TestNextParked.


derekparker (approved):
LGTM

PR #2644, opened 2021-08-04
tests: check presence of gcc for cgo tests

aarzilli:
The install of gcc sometimes fails on our CI, it is not an error if the
tests for cgo can not run because there's no C compiler.


derekparker (approved):
LGTM

PR #2648, opened 2021-08-06
service/dap: support starting Debugger in Run and remote attach mode

polinasok:
New command-line interface coming in a separate PR

### DIFF
--- a/service/dap/daptest/client.go
+++ b/service/dap/daptest/client.go
@@ -136,14 +136,9 @@ func (c *Client) ExpectOutputEventRegex(t *testing.T, want string) *dap.OutputEv
 	return e
 }
 
-func (c *Client) ExpectOutputEventProcessExited(t *testing.T, status int) *dap.OutputEvent {
+func (c *Client) ExpectOutputEventProcessExited(t *testing.T, status int, detaching string) *dap.OutputEvent {
 	t.Helper()
-	return c.ExpectOutputEventRegex(t, fmt.Sprintf(`Process [0-9]+ has exited with status %d\n`, status))
-}
-
-func (c *Client) ExpectOutputEventDetaching(t *testing.T) *dap.OutputEvent {
-	t.Helper()
-	return c.ExpectOutputEventRegex(t, `Detaching\n`)
+	return c.ExpectOutputEventRegex(t, fmt.Sprintf(`Process [0-9]+ has exited with status %d\n%s`, status, detaching))
 }
 
 func (c *Client) ExpectOutputEventDetachingKill(t *testing.T) *dap.OutputEvent {
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -108,7 +108,7 @@ type Server struct {
 	// See also comment for convertVariable.
 	variableHandles *variablesHandlesMap
 	// args tracks special settings for handling debug session requests.
-	args launchAttachArgs
+	args sessionArgs
 	// exceptionErr tracks the runtime error that last occurred.
 	exceptionErr error
 	// clientCapabilities tracks special settings for handling debug session requests.
@@ -132,9 +132,9 @@ type Server struct {
 	sendingMu sync.Mutex
 }
 
-// launchAttachArgs captures arguments from launch/attach request that
-// impact handling of subsequent requests.
-type launchAttachArgs struct {
+// sessionArgs captures arguments from launch/attach request that
+// impact handling of subsequent requests of the debug session.
+type sessionArgs struct {
 	// stopOnEntry is set to automatically stop the debugee after start.
 	stopOnEntry bool
 	// stackTraceDepth is the maximum length of the returned list of stack frames.
@@ -150,7 +150,7 @@ type launchAttachArgs struct {
 }
 
 // defaultArgs borrows the defaults for the arguments from the original vscode-go adapter.
-var defaultArgs = launchAttachArgs{
+var defaultArgs = sessionArgs{
 	stopOnEntry:                  false,
 	stackTraceDepth:              50,
 	showGlobalVariables:          false,
@@ -217,8 +217,8 @@ func NewServer(config *service.Config) *Server {
 }
 
 // If user-specified options are provided via Launch/AttachRequest,
-// we override the defaults for optional args.
-func (s *Server) setLaunchAttachArgs(request dap.LaunchAttachRequest) error {
+// we override the defaults for optional session args.
+func (s *Server) setSessionArgs(request dap.LaunchAttachRequest) error {
 	stop, ok := request.GetArguments()["stopOnEntry"].(bool)
 	if ok {
 		s.args.stopOnEntry = stop
@@ -326,10 +326,26 @@ func (s *Server) triggerServerStop() {
 // and starts processing requests from it. Use Stop() to close connection.
 // The server does not support multiple clients, serially or in parallel.
 // The server should be restarted for every new debug session.
-// The debugger won't be started until launch/attach request is received.
+// The debugger starts if the config has process info when Run is called.
+// Otherwise, the debugger won't be started until launch/attach request
+// is received with the complete configuration info.
 // TODO(polina): allow new client connections for new debug sessions,
 // so the editor needs to launch delve only once?
 func (s *Server) Run() {
+	// TODO(polina): support replay/core configuration as well?
+	if len(s.config.ProcessArgs) > 0 || s.config.Debugger.AttachPid > 0 {
+		debugger, err := debugger.New(&s.config.Debugger, s.config.ProcessArgs) // Process is halted on entry
+		// TODO(polina): support continue on entry
+		if err != nil {
+			s.log.Errorf("Error starting debugger: %s\n", err)
+			s.triggerServerStop()
+			return
+		}
+		s.mu.Lock()
+		s.debugger = debugger
+		s.mu.Unlock()
+	} // else wait for launch/attach request to start debugging
+
 	go func() {
 		conn, err := s.listener.Accept() // listener is closed in Stop()
 		if err != nil {
@@ -692,6 +708,18 @@ func (s *Server) logToConsole(msg string) {
 		}})
 }
 
+// Guard against logging to a nil client connection
+// as some of the shutdown/initialization code might
+// get executed ooutside of a client connection.
+// Caller must hold mu lock.
+func (s *Server) logToConsoleIfConn(msg string) {
+	if s.conn != nil {
+		s.logToConsole(msg)
+	} else {
+		s.log.Debug(msg)
+	}
+}
+
 func (s *Server) onInitializeRequest(request *dap.InitializeRequest) {
 	s.setClientCapabilities(request.Arguments)
 	if request.Arguments.PathFormat != "path" {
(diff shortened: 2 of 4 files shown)


polinasok:
What does this error mean?
```
DAP server listening at: [::]:57492
Connecting to server at: [::]:57492
2021-08-06T07:13:51Z error layer=dap runtime error: ptraceGetLwpInfo err no such process -1
--- FAIL: TestAttachRemoteContinueOnEntry (0.18s)
    server_test.go:5097: got &dap.StoppedEvent{Event:dap.Event{ProtocolMessage:dap.ProtocolMessage{Seq:0, Type:"event"}, Event:"stopped"}, Body:dap.StoppedEventBody{Reason:"exception", Description:"runtime error", ThreadId:0, PreserveFocusHint:false, Text:"ptraceGetLwpInfo err no such process -1", AllThreadsStopped:true, HitBreakpointIds:[]int(nil)}}, want *dap.TerminatedEvent
DAP server listening at: [::]:57494
2021-08-06T07:13:51Z error layer=dap detach returned error: device busy

```

aarzilli:
I was under the impression that for this feature was going to be implemented by letting users start delve in the old way (`dlv debug --headless ...`) with the argument behind the decision being that it wouldn't break existing documentation. It seems that this PR is going into a different direction. What's the overriding reason for doing so? It also seems that this changeset also highlights another problem: a dap.Server object can now exist without a connected client, every `s.send` call is suspicious and needs to be audited to determine whether it can happen without a client, hence the `logToConsoleIfConn` and presumably more IfConns to come if the continueOnEntry is also implemented.

polinasok:
> I was under the impression that for this feature was going to be implemented by letting users start delve in the old way (`dlv debug --headless ...`) with the argument behind the decision being that it wouldn't break existing documentation. It seems that this PR is going into a different direction. What's the overriding reason for doing so?

`dlv debug --headless` is indeed the final destination, but I wanted to get there gradually, with the dlv-dap command without multi-client and such being an intermediate version that can highlight any discrepancies we must address (e.g. on shutdown) and can also unblock [Cloud Code](https://cloud.google.com/code) to try out dlv-dap while we work on the full merger.

> It also seems that this changeset also highlights another problem: a dap.Server object can now exist without a connected client, every `s.send` call is suspicious and needs to be audited to determine whether it can happen without a client, hence the `logToConsoleIfConn` and presumably more IfConns to come if the continueOnEntry is also implemented.

This is only true for shutdown because the same shutdown code is currently reused no matter how things are shut down. We report explicit progress to the user via the Debug Console because we have had confusion as to whether things shut down behind the scenes or not. One might argue that even after `dlv --headless` merger it would be useful to have such shutdown logging available as well if the dap client is still connected. But if that is a concern, we can have separate shutdown code depending on how things were launched. We need to sort how Stop will be called with the merger anyway. And on a related note, why is rpc's `Stop()` only calling `Detach` and not `Halt+Detach`?

Anything else that the server does can only happen in response to a request from a client, so unless we also want to safeguard for unexpected connection interruptions, not additional auditing is needed. As for continue on entry, it would be nice to have it available without multi-client, by just calling the continue command after attaching the debugger.

aarzilli:
> > I was under the impression that for this feature was going to be implemented by letting users start delve in the old way (`dlv debug --headless ...`) with the argument behind the decision being that it wouldn't break existing documentation. It seems that this PR is going into a different direction. What's the overriding reason for doing so?
> 
> `dlv debug --headless` is indeed the final destination, but I wanted to get there gradually, with the dlv-dap command without multi-client and such being an intermediate version that can highlight any discrepancies we must address (e.g. on shutdown) and can also unblock [Cloud Code](https://cloud.google.com/code) to try out dlv-dap while we work on the full merger.

I don't think this is true, that this is a gradual path to there. If dap.Server never starts a dabug session without a client connection then it doesn't even need to exist unless a client is connected. Also the entire initialization of the debugger object would happen elsewhere and you would end up with three different initialization paths for dap.Server:

* not connected to a client, but has a debugger object (this PR)
* connected to a client, no debugger object: it will be created via a Launch request (normal mode)
* connected to a client, with a debugger object, Launch request needs to be ignored (`dlv --headless`)

It's tricky to get it right. FWIW I think the current PR does not, it calls logToConsoleIfConn but later in the same method calls s.send with a nil conn, AFAICT either the logToConsoleIfConn is unnecessary or the send will SIGSEGV.

> > It also seems that this changeset also highlights another problem: a dap.Server object can now exist without a connected client, every `s.send` call is suspicious and needs to be audited to determine whether it can happen without a client, hence the `logToConsoleIfConn` and presumably more IfConns to come if the continueOnEntry is also implemented.
> 
> This is only true for shutdown because the same shutdown code is currently reused no matter how things are shut down.

Not if the `TODO(polina): support continue on entry` is implemented.

> We report explicit progress to the user via the Debug Console and via logging because we have had confusion as to whether things shut down behind the scenes or not. One might argue that even after `dlv --headless` merger it would be useful to have such shutdown logging both on client and server sides. But if that is a concern, we can have separate shutdown code depending on how things were launched. We need to sort out how Stop will be called with the merger anyway. And on a related note, why is rpc's `Stop()` only calling `Detach` and not `Halt+Detach`?

AFAIK it only gets there if the target process is stopped or terminated.

polinasok:
> If dap.Server never starts a dabug session without a client connection then it doesn't even need to exist unless a client is connected. 

I am not sure what you mean.

> Also the entire initialization of the debugger object would happen elsewhere and you would end up with three different initialization paths for dap.Server.

Yes, that would be the case.

> connected to a client, with a debugger object, Launch request needs to be ignored (dlv --headless)

Is that the case though? I planned both (1) and (3) to behave the same during client initialization with the debugger already present.

- launch + program => start dlv, tell it to launch program
- launch + program + host/port => connect to dlv, tell it to launch program
- attach + host/port => connect to dlv, expect it to already have running program 

> It's tricky to get it right. FWIW I think the current PR does not, it calls logToConsoleIfConn but later in the same method calls s.send with a nil conn, AFAICT either the logToConsoleIfConn is unnecessary or the send will SIGSEGV.

Which method are you referring to? I tested this both manually with kill and added TestNoClient to server_test.go. Where do you expect this to crash?

> > And on a related note, why is rpc's Stop() only calling Detach and not Halt+Detach?

> AFAIK it only gets there if the target process is stopped or terminated.

Not if Stop is triggered when the server is force-killed with a signal while the program is running.

aarzilli:
> > If dap.Server never starts a dabug session without a client connection then it doesn't even need to exist unless a client is connected.
> 
> I am not sure what you mean.

With the `--headless` mode serving DAP and JSON-RPC multiplexed it would work like this:

* rpccommon/server.go` listen on the socket and accepts connections from clients
* after a client is connected it looks at the first byte sent on the connection to determine which protocol is used
* if it's DAP it starts dap.Server

dap.Server wouldn't start the debugger and wouldn't be even instantiated until there's a connection from a client to serve.

> 
> > Also the entire initialization of the debugger object would happen elsewhere and you would end up with three different initialization paths for dap.Server.
> 
> Yes, that would be the case.
> 
> > connected to a client, with a debugger object, Launch request needs to be ignored (dlv --headless)
> 
> Is that the case though? I planned both (1) and (3) to behave the same during client initialization with the debugger already present.

For responding to the launch request, yes, but in (1) you have a Server object without a connection to a client and in (3) you don't.

> > It's tricky to get it right. FWIW I think the current PR does not, it calls logToConsoleIfConn but later in the same method calls s.send with a nil conn, AFAICT either the logToConsoleIfConn is unnecessary or the send will SIGSEGV.
> 
> Which method are you referring to? I tested this both manually with kill and added TestNoClient to server_test.go. Where do you expect this to crash?

You're right, I was just reading the diff wrong, there should be no possible sends in the path that has logToConsoleIfConn, but it doesn't matter it's an exceptional state that the code can be in that will need to be accounted for on every change.

> > > And on a related note, why is rpc's Stop() only calling Detach and not Halt+Detach?
> 
> > AFAIK it only gets there if the target process is stopped or terminated.
> 
> Not if Stop is triggered when the server is force-killed with a signal while the program is running.

Why should it work differently?

hyangah:
> > > If dap.Server never starts a dabug session without a client connection then it doesn't even need to exist unless a client is connected.
> > 
> > 
> > I am not sure what you mean.
> 
> With the `--headless` mode serving DAP and JSON-RPC multiplexed it would work like this:
> 
> * rpccommon/server.go` listen on the socket and accepts connections from clients
> * after a client is connected it looks at the first byte sent on the connection to determine which protocol is used
> * if it's DAP it starts dap.Server
> 
> dap.Server wouldn't start the debugger and wouldn't be even instantiated until there's a connection from a client to serve.
> 
> > > Also the entire initialization of the debugger object would happen elsewhere and you would end up with three different initialization paths for dap.Server.
> > 
> > 
> > Yes, that would be the case.
> > > connected to a client, with a debugger object, Launch request needs to be ignored (dlv --headless)
> > 
> > 
> > Is that the case though? I planned both (1) and (3) to behave the same during client initialization with the debugger already present.
> 
> For responding to the launch request, yes, but in (1) you have a Server object without a connection to a client and in (3) you don't.
> 
> > > It's tricky to get it right. FWIW I think the current PR does not, it calls logToConsoleIfConn but later in the same method calls s.send with a nil conn, AFAICT either the logToConsoleIfConn is unnecessary or the send will SIGSEGV.
> > 
> > 
> > Which method are you referring to? I tested this both manually with kill and added TestNoClient to server_test.go. Where do you expect this to crash?
> 
> You're right, I was just reading the diff wrong, there should be no possible sends in the path that has logToConsoleIfConn, but it doesn't matter it's an exceptional state that the code can be in that will need to be accounted for on every change.
> 
> > > > And on a related note, why is rpc's Stop() only calling Detach and not Halt+Detach?
> > 
> > 
> > > AFAIK it only gets there if the target process is stopped or terminated.
> > 
> > 
> > Not if Stop is triggered when the server is force-killed with a signal while the program is running.
> 
> Why should it work differently?

Confusing. When the delve headless server is force-killed with a signal, how is the target program getting killed?
(OTOH, I guess killing non-dap headless server while target program is running is a rare event given that headless server runs in background and most well-behaved clients (dlv terminal or other clients) would send halt first before disconnecting)

polinasok:
> a dap.Server object can now exist without a connected client, every `s.send` call is suspicious and needs to be audited to determine whether it can happen without a client, hence the `logToConsoleIfConn` and presumably more IfConns to come if the continueOnEntry is also implemented.

Interestingly, in the rpc server with --continue, the handler also can call send without a connected client on forced shutdown. But it just results in a write error because there used to be a client.
```
2021-08-27T01:06:19-07:00 error layer=rpc writing response:write tcp 127.0.0.1:58310->127.0.0.1:58311: use of closed network connection
```

aarzilli:
> It also adds logic to Run that will not apply because rpccommon Run will be used instead. So that part won't get reused here

So let's not add it.

> I don't see what the harm is in having this mode as well, especially while the other version is still being worked out.

The harm is more special cases.

polinasok:
All right. Let's see where the other version takes me.

soffieswan043:
#2648 (comment)

soffieswan043 (commented):
#2648 (comment)

PR #2649, opened 2021-08-06
fix/cmds: use simple chan read instead of select-case

hitzhangjie:
Use simple chan read instead of select-case on only one chan.

When run `golangci-lint run .`, it reports several chan receive problems that use select-case on only one chan. We can use simple chan receive instead of select-case.


hitzhangjie:
Sorry, I don't know how to solve the Aggregator problem here.

image

aarzilli (approved):

derekparker (approved):
LGTM

PR #2650, opened 2021-08-10
proc: do not assign temporary breakpoint IDs

aarzilli:
Internal breakpoints do not need IDs and assigning them from a counter
separate from the user ID counter can be a cause of confusion.

If a user breakpoint is overlayed on top of a pre-existing internal
breakpoint the temporary ID will be surfaced as if it was a user ID,
possibly conflicting with another user ID.

If a temporary breakpoint is overlayed on top of a pre-existing user
breakpoint and the user breakpoint is first deleted and then
re-created, the user ID will be resurrected along with the breakpoint,
instead of allocating a fresh one.

This change removes internal breakpoint IDs entirely, only user
breakpoints receive an ID.

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -27,6 +27,8 @@ const (
 
 	unrecoveredPanicID = -1
 	fatalThrowID       = -2
+
+	NoLogicalID = -1000 // Logical breakpoint ID for breakpoints internal breakpoints.
 )
 
 // Breakpoint represents a physical breakpoint. Stores information on the break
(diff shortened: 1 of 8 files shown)


pkg/proc/breakpoints.go
```diff
@@ -609,11 +620,22 @@ func (t *Target) setBreakpointInternal(addr uint64, kind BreakpointKind, wtype W
...
 		newBreaklet.HitCount = map[int]uint64{}
+		bpmap.breakpointIDCounter++
+		newBreaklet.LogicalID = bpmap.breakpointIDCounter
 	}
 	if bp, ok := bpmap.M[addr]; ok {
 		if !bp.canOverlap(kind) {
 			return bp, BreakpointExistsError{bp.File, bp.Line, bp.Addr}
 		}
+		if kind == UserBreakpoint {
+			bp.Tracepoint = false
```
derekparker, line 631: Why set all these here?

derekparker (changes requested):

pkg/proc/breakpoints.go
aarzilli: They are only relevant for user breakpoints if they are set it means they had been set for an old user breakpoint that was removed and kept around by overlapping non-user breakpoints.

derekparker (approved):
LGTM

PR #2653, opened 2021-08-11
service/dap: clarify handling of relative program path

### LINKED ISSUE #2392: dlv dap: (maybe) support relative paths in launch requests
This might already be working (since dlv allows relative paths on the command-line, and we use the same backend). In either case, we need to "document" the current behavior with unittests and then user facing documentation (e.g. `--help`). If it is not working or should not (yet?) be supported, then an explicit error at the time of parsing launch requests with a request to only specify full paths might be desirable.

The existing unittests currently always specify the full fixture source path. For example: 
https://github.com/go-delve/delve/blob/333e84a0cb209a9ec762b753cbd4ecbc679104d4/service/dap/server_test.go#L2282

We should also try the test with a relative path. For example, if we assume that dlv's working directory is delve/service/dap in the context of dap tests, then a relative path like `../../_fixtures/foo.go` could be used in the test. This should work with launch-debug tests because they put the binary in the same directory where dlv is running. With exec, I think the fixtures get built into some temp location with a path that doesn't overlap.

Another thing to try is testing this with packages and not individual go files.

\cc @suzmue

polinasok:
Closes #2392
Updates golang/vscode-go#1680
Updates golang/vscode-go#1348

This PR adds tests, logging and `--help` documentation note confirming that relative program paths are indeed handled and are interpreted relative to server's working directory.

### DIFF
--- a/Documentation/usage/dlv_dap.md
+++ b/Documentation/usage/dlv_dap.md
@@ -7,13 +7,15 @@
 [EXPERIMENTAL] Starts a headless TCP server communicating via Debug Adaptor Protocol (DAP).
 
 The server is always headless and requires a DAP client like vscode to connect and request a binary
-to be launched or process to be attached to. The following modes are supported:
+to be launched or process to be attached to. The following modes can be specified via client's launch config:
 - launch + exec (executes precompiled binary, like 'dlv exec')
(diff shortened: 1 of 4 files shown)


aarzilli (approved):

derekparker (approved):
LGTM

PR #2654, opened 2021-08-11
pkg/proc: Parse Goroutine ID in eBPF tracer

derekparker:
This patch enables the eBPF tracer backend to parse the ID of the
Goroutine which hit the uprobe. This implementation is specific to AMD64
and will have to be generalized further in order to be used on other
architectures.

This patch also adds a test for the eBPF tracer backend. The test will skip
if it is not being run as root because we need SYS_CAP_ADMIN in order to
load the BPF program.

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -463,15 +463,51 @@ func (t *Target) SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr) 
 	return t.setBreakpointInternal(addr, kind, 0, cond)
 }
 
+// SetEBPFTracepoint will attach a uprobe to the function
+// specified by 'fnName'.
 func (t *Target) SetEBPFTracepoint(fnName string) error {
+	// Not every OS/arch that we support has support for eBPF,
+	// so check early and return an error if this is called on an
+	// unsupported system.
 	if !t.proc.SupportsBPF() {
 		return errors.New("eBPF is not supported")
 	}
+	// Start putting together the argument map. This will tell the eBPF program
+	// all of the arguments we want to trace and how to find them.
 	var args []ebpf.UProbeArgMap
 	fn, ok := t.BinInfo().LookupFunc[fnName]
 	if !ok {
 		return fmt.Errorf("could not find function %s", fnName)
 	}
+
+	// Get information on the Goroutine so we can tell the
+	// eBPF program where to find it in order to get the
+	// goroutine ID.
+	rdr := t.BinInfo().Images[0].DwarfReader()
+	rdr.SeekToTypeNamed("runtime.g")
+	var goidOffset int64
+	for {
+		ent, err := rdr.Next()
+		if err != nil {
+			return err
+		}
+		name, ok := ent.Val(dwarf.AttrName).(string)
+		if !ok {
+			continue
+		}
+		if name == "goid" {
+			off, ok := ent.Val(dwarf.AttrDataMemberLoc).(int64)
+			if !ok {
+				return errors.New("invalid type case for member location offset")
+			}
+			goidOffset = off
+			break
+		}
+	}
+
+	// Start looping through each argument / return parameter for the function we
+	// are setting the uprobe on. Parse location information so that we can pass it
+	// along to the eBPF program.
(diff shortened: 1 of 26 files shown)


derekparker:
Converting this to a draft PR. I will need to either commit the `.o` file or build it in CI to run the test. I'll work on fixing this tomorrow.

derekparker:
@nd is it possible to run this build in TeamCity and have it pick up the changes I've made to the settings.kts file? I want to test using ubuntu 21.04 instead of the 20.04 LTS to pick up the newer kernel.

nd:
@derekparker yes. Use the 'Run...' button on the build page. It will open custom build dialog. On the 'Changes' tab select `Use settings: from VCS`

Screenshot (219)

It doesn't work for all changes in dsl. As far as I can see you changed build steps, such a change should work.

derekparker:
@nd I'm kind of at a loss here. The `darwin/amd64` builds keep getting `signal: killed`, but I don't understand why. I've tried this build on both amd64 and m1 based macs and I haven't run into the same issue. Is there something going on with the darwin/amd64 builders?

I can't get dmesg output, and nothing I've tried has revealed what is causing issues here.

nd:
Looks like rebooting the builder helped. You can do that from UI on the build agent page ('Reboot agent machine' link in the bottom) .

derekparker:
@nd thanks for your help! One quick question, what is the host linux system the containers are running on? I'm curious how the host kernel is configured and knowing the distro will help.

@aarzilli this is ready for review. I've added a test but disabled it in CI because I believe we're likely running on a kernel that isn't configured properly for loading the eBPF program. So, there's a test that can be ran locally and eventually we can enable it in CI once we get the build environment conducive to it.

nd:
They use Ubuntu 20.04.2 LTS. Kernel versions for arm64 is5.4.0-1045-aws, for x86 and x64 - 5.8.0-1041-aws. The easiest way to get more details is to run something like 

    cat /boot/config-`uname -r` 

as a first build step.

pkg/proc/breakpoints.go
```diff
@@ -463,15 +463,51 @@ func (t *Target) SetBreakpoint(addr uint64, kind BreakpointKind, cond ast.Expr)
...
 		return errors.New("eBPF is not supported")
 	}
+	// Start putting together the argument map. This will tell the eBPF program
+	// all of the arguments we want to trace and how to find them.
 	var args []ebpf.UProbeArgMap
 	fn, ok := t.BinInfo().LookupFunc[fnName]
 	if !ok {
 		return fmt.Errorf("could not find function %s", fnName)
 	}
+
```
aarzilli, line 482: I think `findType` and then iterating through the `Field` slice is better.

pkg/proc/internal/ebpf/trace_probe/trace.bpf.c
```diff
@@ -140,6 +140,42 @@ int parse_param(struct pt_regs *ctx, function_parameter_t *param) {
...
+    }
+    // The bpf_get_current_task helper returns us the address of the task_struct in
+    // kernel memory. Use the bpf_probe_read_kernel helper to read the struct out of
+    // kernel memory.
+    bpf_probe_read_kernel(task, sizeof(struct task_struct), (void*)(task_ptr));
+
+    // Get the Goroutine ID which is stored in thread local storage.
+    __u64  goid;
+    size_t g_addr;
+    bpf_probe_read_user(&g_addr, sizeof(void *), (void*)(task->thread.fsbase-8));
```
aarzilli, line 169: The goroutine address offset isn't always -8, it's BinaryInfo.gStructOffset. Can it be passed down here?

pkg/proc/breakpoints.go
derekparker: Done.

pkg/proc/internal/ebpf/trace_probe/trace.bpf.c
derekparker: Done.

derekparker:
Not sure why we're getting a bunch of FreeBSD failures all of a sudden.

aarzilli (approved):

Issue #2507, opened 2021-05-25
dap: support `buildDir` attribute in the Launch request


hyangah:
Like `dlv debug` or `dlv test` commands, `dlv dap` builds the target program/test 
from the current directory where the dlv command is invoked. 
There is `cwd` attribute in the launch argument, but that corresponds to `--wd` flag 
and determines only the working directory of the target program.

We assumed the client can run `dlv dap` from the correct directory required for build
until now. However, we recently heard of some cases where launching `dlv dap` from
the right directory is not easy.

* vimspector always launches the debug adapter process from the workspace root. 
Currently building targets outside of the workspace, or outside of the main module isn't
possible.
* some clients may want to lazily determine the launch target after starting `dlv dap`
process. 
* if we ever make a dlv dap process to serve multiple launch requests (sequentially or not)
building the build directory and the dlv's process working directory isn't a good idea.

So, allow a new, optional attribute that specifies the build directory
(where `go build` command runs).

cc @polinasok @suzmue

polinasok:
Please help me understand the uses cases better: 

> * vimspector always launches the debug adapter process from the workspace root.
>   Currently building targets outside of the workspace, or outside of the main module isn't
>   possible.

Do you have a reference to the discussion with vimspector? I am curious to understand why they always start the process from the workspace root only as opposed to what vscode-go does.

> * some clients may want to lazily determine the launch target after starting `dlv dap`
>   process.

Do you have a specific client/use case in mind?

> * if we ever make a dlv dap process to serve multiple launch requests (sequentially or not)

Is this so we can have a long-running debug adapter server that the same client can start once and direct all debug sessions to like in #2329)

>   building the build directory and the dlv's process working directory isn't a good idea.

"build the build directory and" - do you mean combining (because dlv's working directory, i.e. where dlv is launched, will be determined by the first launched request while the next launch request might use a target from a different location?

hyangah:
Note: we ended up implementing `delveCwd` instead of `buildDir`. The attribute changes Delve's working directory using `os.Chdir`.

I realized I didn't answer to Polina's question here

> Do you have a reference to the discussion with vimspector? I am curious to understand why they always start the process from the workspace root only as opposed to what vscode-go does.

vimspector is a [general dap client](https://github.com/puremourning/vimspector/tree/726a06e0d043e5897ade0624da2ecbbceab53484/python3/vimspector) for all languages, not specialized for any particular language. 
[...]

Issue #2551, opened 2021-06-24
Attach and debug any child process fork'ed by the target process


derekparker:
Currently Delve does not use `PTRACE_O_TRACEFORK` or any other mechanism to begin debugging a process created when the target process (the original process being debugged) `fork`s. This is an issue for tracking and discussing this new potential feature.

This would require changes to the API and to the `debugger` layer to allow for controlling and manipulating multiple processes simultaneously. It also requires some thought on how such a situation should behave from a users perspective.

Open questions / items to consider:

* Amount of work needed to make the `debugger` layer be able to handle multiple processes. We abstract a lot away via the `Target` type so the debugger would essentially need to initialize and keep track of N different child `Target`s.
* What new functionality will be needed from an API perspective? We would obviously need a way to switch between which is the "current" process which would affect any command used to manipulate or explore a process being debugged.
* How do we coordinate the two processes? Stop both when one hits a breakpoint or only the process which hit it? What happens when both programs hit a breakpoint, which do we show / prioritize for the user?

There's certainly a lot more to consider here but this is at least a brief list to get the topic started.

mdempsky:
Something else not explicitly stated is that the subprocess might exec, so ideally delve would be able to debug multiple processes running different executables, including setting break points in processes+programs that aren't yet running. (Maybe this is implicitly understood as a requirement though.)

As a concrete use case to consider: I work on the Go compiler, and often I'm running something like `go build std cmd` or `go run run.go` (where run.go runs all of the tests in [$GOROOT/test](https://github.com/golang/go/tree/master/test)). Usually I care about some individual compile process deeply nested within this process tree. A limiting factor in my use of Delve is that I need to first reconstruct the exact environment and command-line that was used to invoke the failing compile process, and then use that to invoke Delve. This is a large hassle currently, so I haven't attempted it many times.

If I could just run the entire cmd/go process under Delve, set a breakpoint somewhere in cmd/compile, and then let delve handle all the details in letting me debug from there, that would be great.

--

As for implementation details, for reference, gdb supports multi-process debugging with "follow-forks-mode": https://sourceware.org/gdb/onlinedocs/gdb/Forks.html

Also relevant: https://sourceware.org/gdb/onlinedocs/gdb/Inferiors-Connections-and-Programs.html

[...]

PR #2660, opened 2021-08-16
dap: support delveCwd and use a temporary file as default debug binary

hyangah:
The directory where a go command runs often determine the go command's behavior. 
For example, in the module mode, go commands need to run from the right main module
to build the target correctly. Delve traditionally uses delve's current working directory
as the working directory for the go commands it invokes to build the debugged package.
Users are expected to run delve from the right directory depending on the target they want
to build.

In Delve DAP server mode, however,  the target may be determined after the Delve server
starts. For example, some editors launch debug adapters from the workspace root directory
and then send the launch request. If the launch request requires to build a file in a different
module inside subdirectory, the go build command will fail.

This PR provides an option (`delveCwd`) to change the working directory of the Delve process.
When receiving a launch request with this property, we will simply call os.Chdir to the directory.
This has a side-effect, though. We hoped to implement a long-lived Delve DAP server and
support less-expensive debug session restart, but, because the delve process's working
directory is changed, we need extra care when the day comes.
 
Fixes https://github.com/go-delve/delve/issues/2507

This PR also fixes:

* Instead of creating the debug output in the current working directory when `Output`
is not specified, create it in a temp directory.
* In 'test' launch mode, use the package directory as the default `cwd`. That matches
`go test` and `dlv test` behavior.
* We no longer infer the default `cwd` based on the `program` path. The default `cwd`
for non-test mode is Delve's working directory. That matches other dlv command's behavior.
* Clarify the meaning of the relative paths

Updates https://github.com/golang/vscode-go/issues/1348


service/dap/server.go
```diff
@@ -819,16 +819,17 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 		}
 
 		buildFlags := args.BuildFlags
+		buildDir := args.BuildDir
```
aarzilli, line 822: Why not just `if args.BuildDir != "" { os.Chdir(args.BuildDir) }`?

hyangah: That limits DAP package to work only with one debug session and potentially adds further confusion to handling of the relative paths.

aarzilli: > That limits DAP package to work only with one debug session

I was looking at this for #2551, there are very deeply rooted limitations around this. It might look like you can just instantiate multiple instances of Debugger but they won't work except in select circumstances.

> potentially adds further confusion to handling of the relative paths.

I was actually thinking the opposite. The PR, as it is, adds one more possible directory that a relative path can be relative to.

hyangah: Good to know that such limitation already exists. Adding one more blocker doesn't change the game.

My original plan was to interpret all the paths as paths relative to the workspace root because that's what we tell users. 
os.Chdir makes relative paths in `cwd`. But I guess vscode-go can avoid this by always explicitly setting the cwd.

I left the output to be computed based on the original working directory path for now. Should it be relative to the updated process working directory too? It's a temporary file, so either seems fine to me.

hyangah: Because of test failures switched to recover to the original wd after build. 
(debug server is running in process, so we need to make sure to switch back to the original directory for each test. I figured it's easier to just affect only build step).
But still contained changes to only service/dap package.

Once build is done, cwd is going to be used. So, keeping the original working directory shouldn't be confusing.

aarzilli: If you do it this way you might as well just pass it inside exec.Command. The point of doing the `os.Chdir` would be to change to it as early as possible (so it would be the first thing you do after getting the launch request and never switch back, so that the original working directory can not be accidentally used by any of the other code.

The tests that intentionally change build directory would be responsible for deferring an os.Chdir to undo the change.

polinasok:
Please rebase. It is hard to tell what is new and what is old code.

polinasok (commented):
In light of the os.Chdir discussion, I reread #2507. I wonder if I do not fully understand the use case for this change and hence have a hard time evaluating the proposed alternative.

Currently we have two different roots that we are dealing with:
1. dlv wd - where dlv command was run; used to resolve relative paths (program, output, wd, traceDir, coreDumpPath, what else?)
2. target wd (aka (c)wd) - can specify to be different from dlv wd (dlv wd if unspecified)

If we want to build in a certain directory, we must launch dlv there. And in that case not only build will be done there, but all relative paths will use that build directory as root. All or nothing.

When dlv is started on user's behalf from a client, everything is supposed to revolve around the workspace root. Yet users are allowed to specify targets outside of that root. So to decouple building and path resolution vscode-go runs dlv where the program is and resolves paths on its own prior to running delve. Except it only does this for program, cwd, and output, not trace and core files, for example. 

--buildDir gives a mechanism for clients to decouple dlv-wd from build-dir, so they don't have to keep track of paths-to-resolve and can let dlv handle them consistently. I am not sure when direct users of dlv would use this feature or if it will be mostly hidden from the users by clients. To avoid any path-relative-to-what confusion it could be strictly for build only, so all paths, including target and output (inputs and outputs of build), will still be relative to dlv wd.

Will os.Chdir done before building and resolving paths be any different than switching directories before running dlv? If a client wanted relative-root to be different from build-root, wouldn't they have to massage paths externally anyway?

service/dap/server.go
hyangah: I reverted back to the original version of this pr - that passes the build directory down to the build package. We thought about it a lot and we think it's better to limit the scope of buildDir to be only the build stage and leave the delve's working directory untouched.

Made `output` to be buildDir-relative. I have two conflicting thought about this - if the user tries to build a target outside the workspace (e.g. a binary that depends on the libraries in the workspace), writing a file outside the workspace is a bit strange. On the other hand, by defaulting to the package directory (most likely the buildDir), when users start multiple debug sessions concurrently, we will have less chance of multiple delves attempting to work with the same  `__debug_bin` path.

hyangah:
> In light of the os.Chdir discussion, I reread #2507. I wonder if I do not fully understand the use case for this change and hence have a hard time evaluating the proposed alternative.
> 
> Currently we have two different roots that we are dealing with:
> 
> dlv wd - where dlv command was run; used to resolve relative paths (program, output, wd, traceDir, coreDumpPath, what else?)
> target wd (aka (c)wd) - can specify to be different from dlv wd (dlv wd if unspecified)
> If we want to build in a certain directory, we must launch dlv there. And in that case not only build will be done there, but all relative paths will use that build directory as root. All or nothing.
> 
> When dlv is started on user's behalf from a client, everything is supposed to revolve around the workspace root. Yet users are allowed to specify targets outside of that root. So to decouple building and path resolution vscode-go runs dlv where the program is and resolves paths on its own prior to running delve. Except it only does this for program, cwd, and output, not trace and core files, for example.
> 
> --buildDir gives a mechanism for clients to decouple dlv-wd from build-dir, so they don't have to keep track of paths-to-resolve and can let dlv handle them consistently. I am not sure when direct users of dlv would use this feature or if it will be mostly hidden from the users by clients.

To clarify - this PR is for DAP only, and does not propose `--buildDir` flag. I don't believe that is useful as a general command line flag. It's useful in DAP because, like many other parameters, build directory is determined later after the delve dap command starts to run.

> To avoid any path-relative-to-what confusion it could be strictly for build only, so all paths, including target and output (inputs and outputs of build), will still be relative to dlv wd.

The actual use case is general-purpose debug adapter plugins such as vimspector that starts debug adapters from the workspace root or whatever it decides to run for all languages. VS Code Go could avoid the issue because it is Go-specific. 

> 
> Will os.Chdir done before building and resolving paths be any different than switching directories before running dlv? If a client wanted relative-root to be different from build-root, wouldn't they have to massage paths externally anyway?

In theory, users can specify all the attributes explicitly to make everything deterministic. But it would be nice if delve dap can provide a reasonable default behavior and users don't have to set everything by themselves. 

Anyway, rebased, updated the types.go to reflect my current thought.

The DeepSource failure is annoying and I am not sure how useful that check is. (Can we turn it off or, completely switch to staticcheck or something less noisy?)

service/dap/server.go
aarzilli: Maybe we should finally fix the default path of __debug_bin to be a randomly generated one with ioutil.TempFile.

hyangah: Took a stab at it. Turned out a bit complicated than I thought and I needed to fix the bug [that infers cwd from the `program`](https://github.com/go-delve/delve/blob/09798471003d55918581b8b835c0046e0c2f560a/service/dap/types.go#L108). 
If you think this is better for a separate PR, I am happy to split it.

This change will break user workflow if the test/program to debug has to run in the package directory but users didn't set 'cwd' nor run dlv dap from the package directory. VS Code Go plugin currently runs `dlv dap` from the debugged test or program's directory, so I think it will be ok.

For `test` mode, we may consider to make dlv figure out the package directory (go list?) and use the directory as the default working directory, like `go test`. That will allow `dlv test`  and DAP launch to accept a package path outside the main module (e.g. dependency's test!), but I guess that's quite a niche use case.

service/dap/server.go
```diff
@@ -735,9 +736,21 @@ func (s *Server) setClientCapabilities(args dap.InitializeRequestArguments) {
 	s.clientCapabilities.supportsVariableType = args.SupportsVariableType
 }
 
-// Default output file pathname for the compiled binary in debug or test modes,
-// relative to the current working directory of the server.
-const defaultDebugBinary string = "./__debug_bin"
+func tempDebugBinary() (string, error) {
```
aarzilli, line 739: Instead of returning the error just log it and return `"__debug_bin"`

service/dap/server.go
aarzilli: > For test mode, we may consider to make dlv figure out the package directory (go list?) and use the directory as the default working directory, like go test

`dlv test` already does this, see testCmd in cmd/dlv/cmds/commands.go.

So, now that the output file is in the temp directory, why should buildDir be applied only to the build command?

hyangah: That's great. (Sorry I missed it).
Made a change to infer the package directory as the default cwd for test mode.

Can we use buildDir as the cwd for test mode?

For test mode, gopls simulates the `go.work` mechanics in older versions of go by creating [a workspace module in a temp directory](https://github.com/golang/tools/blob/master/gopls/doc/workspace.md#workspace-module-experimental) and build all targets from the workspace module. The working directory of the test, in this case, is the package directory seen by the go command run from the temporary workspace module directory (buildDir).

For non-test mode, we cannot assume buildDir as the default cwd.

service/dap/server.go

service/dap/server.go
aarzilli: > Can we use buildDir as the cwd for test mode?

This is an argument for doing `os.Chidr(buildDir)` as soon as possible and then forgetting about it.

> For non-test mode, we cannot assume buildDir as the default cwd.

Why?

aarzilli: > We don't want to do that. Cwd and all related args are can be exposed as 'workspace root'-relative.

The original motivation for this PR was:

> vimspector always launches the debug adapter process from the workspace root.
> Currently building targets outside of the workspace, or outside of the main module isn't
> possible.

This sounds to me like vimspector lacks a feature (letting user specify the working directory for the debugger) and we are writing a workaround for it. So it should minimally do that.
Right now we have that "the working directory = where the go command can be run", this introduces an exception (as well as one more thing that relative paths can be relative to, which we already have too many of). Exceptions are bad, how likely is it that we are going to remember of this particular thing the next time we write something that shells to go?

Even for the relative paths alone, how weird is it that now in launch.json you have some paths that are relative to buildDir (program), some paths that are relative to the working directory of delve (cwd, output, traceDirPath and coreFilePath) and some paths that are relative to cwd (args)?

> In the gopls' experimental module workspace mode, we want to build the test target from the module in the temp directory which users wouldn't know the location, and then switch back to either the package directory (for test) or back to the workspace root or user-specified cwd (for non-test).

This sounds like gopls should either use the `-modfile` compile flag or handle the compilation itself. Delve has never been a fully general compiler wrapper.

hyangah: > The original motivation for this PR was:
> 
> > vimspector always launches the debug adapter process from the workspace root.
> Currently building targets outside of the workspace, or outside of the main module isn't
> possible.
> 
> This sounds to me like vimspector lacks a feature (letting user specify the working directory for the debugger) and we are writing a workaround for it. So it should minimally do that.
> Right now we have that "the working directory = where the go command can be run", this introduces an exception (as well as one more thing that relative paths can be relative to, which we already have too many of). Exceptions are bad, how likely is it that we are going to remember of this particular thing the next time we write something that shells to go?
> 
> Even for the relative paths alone, how weird is it that now in launch.json you have some paths that are relative to buildDir (program), some paths that are relative to the working directory of delve (cwd, output, traceDirPath and coreFilePath) and some paths that are relative to cwd (args)?

I understand your concern about tech debt.

From editor's perspective, workspace root is the directory users are primarily working with by default. buildDir is a hack to work around rare cases where we cannot run `go build` from the workspace root - e.g. multi-module workspace or case-insensitive filesystem or symlink madness. Ideally, most users shouldn't need to worry about `buildDir`.

`cwd`, `traceDirPath`, `coreFilePath` are relevant to the mode where the build step is unnecessary. I found it difficult to explain these without using the workspace root. But `workspace root` is irrelevant to delve. The closest is the delve's process working directory assuming the editor starts delve from the working directory. 

If we do os.Chdir(buildDir), I think the documentation presented to users will be

* `cwd` is interpreted as a path relative to delve's working directory.
* `buildDir` specifies the directory the go command runs  in `debug`, `test` modes. It is irrelevant to other modes. When this is used, it also changes the delve's working directory and, as a result, `cwd` that is computed based on delve's working directory will be affected. 

Note that we still cannot avoid dealing with three directories.

If we take this path (os.Chdir) I think `buildDir` is not the right name because it will be doing beyond what it looks.
We should rename it (maybe `delveWD`? I admit this can be confused with delve's --wd. Other suggestions are welcome), and apply it to all modes. Is it ok with you?

> 
> > In the gopls' experimental module workspace mode, we want to build the test target from the module in the temp directory which users wouldn't know the location, and then switch back to either the package directory (for test) or back to the workspace root or user-specified cwd (for non-test).
> 
> This sounds like gopls should either use the -modfile compile flag or handle the compilation itself. Delve has never been a fully general compiler wrapper.

gopls already utilizes -modfile. The question is how to make Delve behave consistent with gopls. An alternative is to pass -modfile flag as buildFlags and make sure delve use this flag appropriately when shelling out the go command. I think this is as messy as buildDir if not harder.

pkg/gobuild/gobuild.go
```diff
@@ -61,9 +61,9 @@ func GoBuild(debugname string, pkgs []string, buildflags string) error {
 
 // GoBuildCombinedOutput builds non-test files in 'pkgs' with the specified 'buildflags'
 // and writes the output at 'debugname'.
-func GoBuildCombinedOutput(debugname string, pkgs []string, buildflags string) (string, []byte, error) {
+func GoBuildCombinedOutput(debugname string, pkgs []string, buildflags, buildDir string) (string, []byte, error) {
```
polinasok, line 64: Please add a reference to the new arg to the function comment. Is debugname expected to be an absolute path at this point or will it be resolved relative to buildDir?

service/dap/server.go
```diff
@@ -735,10 +736,29 @@ func (s *Server) setClientCapabilities(args dap.InitializeRequestArguments) {
...
+	}
+	f, err := ioutil.TempFile("", binaryPattern)
+	if err != nil {
+		s.log.Errorf("failed to create a temporary binary (%v), falling back to %q", err, defaultDebugBinary)
+		return defaultDebugBinary
+	}
+	name := f.Name()
+	if err := f.Close(); err != nil {
+		s.log.Errorf("failed to create a temporary binary (%v), falling back to %q", err, defaultDebugBinary)
+		return defaultDebugBinary
```
polinasok, line 757: It's confusing that you use exe with the temp file name, but not default one here. How about using cleanExeName at the end only.

service/dap/server.go
```diff
@@ -811,28 +831,34 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+		debugbinary := args.Output
+		if debugbinary == "" {
+			debugbinary = s.tempDebugBinary()
 		}
 
-		buildFlags := args.BuildFlags
+		if !filepath.IsAbs(debugbinary) {
+			o, err := filepath.Abs(debugbinary)
+			if err != nil {
+				s.sendInternalErrorResponse(request.Seq, err.Error())
```
polinasok, line 842: this should `s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to launch", err.Error())`

service/dap/server.go
```diff
@@ -811,28 +831,34 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
-			s.sendInternalErrorResponse(request.Seq, err.Error())
-			return
+		debugbinary := args.Output
+		if debugbinary == "" {
+			debugbinary = s.tempDebugBinary()
 		}
 
-		buildFlags := args.BuildFlags
+		if !filepath.IsAbs(debugbinary) {
+			o, err := filepath.Abs(debugbinary)
```
polinasok, line 840: Won't this be a no-op if the path is already absolute? Why have the extra if?

service/dap/server.go
```diff
@@ -811,28 +831,34 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+		if buildDir == "" {
+			buildDir = "."
+		}
 
 		var cmd string
 		var out []byte
-		wd, _ := os.Getwd()
-		s.log.Debugf("building program '%s' in '%s' with flags '%v'", program, wd, buildFlags)
+		var err error
+		s.log.Debugf("building program %q in %q with flags '%v'", program, args.BuildDir, args.BuildFlags)
```
polinasok, line 856: s/args.BuildDir/buildDir/

service/dap/server.go
```diff
@@ -858,10 +890,7 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 	}
 
 	s.config.ProcessArgs = append([]string{program}, args.Args...)
-	s.config.Debugger.WorkingDir = filepath.Dir(program)
-	if args.Cwd != "" {
-		s.config.Debugger.WorkingDir = args.Cwd
-	}
+	s.config.Debugger.WorkingDir = args.Cwd // cwd: "" == "."
```
polinasok, line 893: Why not make "." explicit instead of relying on the comment?

service/dap/server.go
```diff
@@ -847,6 +873,12 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
 			return
 		}
 		program = debugbinary
+
+		if mode == "test" && args.Cwd == "" {
```
polinasok, line 877: I think this belongs below where you handle Cwd, so all logic for it is in one place.

service/dap/server.go
```diff
@@ -847,6 +873,12 @@ func (s *Server) onLaunchRequest(request *dap.LaunchRequest) {
...
+			// In test mode, run the test binary from the package directory
+			// like in `go test` and `dlv test` by default.
+			args.Cwd = getPackageDir(args.Program, buildDir)
```
polinasok, line 880: Why would the fallback be buildDir and not server's cwd?
This is part of running, not part of building.

service/dap/server_test.go
```diff
@@ -4411,16 +4406,93 @@ func TestLaunchRequestWithRelativeExecPath(t *testing.T) {
...
-		fixtures := protest.FindFixturesDir()
-		testdir, _ := filepath.Abs(filepath.Join(fixtures, "buildtest"))
-		client.LaunchRequestWithArgs(map[string]interface{}{
-			"mode": "test", "program": testdir, "output": "__mytestdir"})
-	})
-	<-serverStopped
+	orgWD, _ := os.Getwd()
+	fixtures := protest.FindFixturesDir() // relative to current working directory.
+	fixturesDir, _ := filepath.Abs(fixtures)
+	pkgDir, _ := filepath.Abs(filepath.Join(fixtures, "buildtest"))
```
polinasok, line 4412: s/pkgDir/absPkgDir/

polinasok:
It would be helpful if the top-level PR description always summarized whatever the current state of the PR is. For example, output is resolved relative to buildDir, but what about program?

service/dap/server.go
hyangah: That's ioutil.TempFile's search pattern to avoid conflicts. https://pkg.go.dev/io/ioutil#TempFile

service/dap/server.go
hyangah: Why is it user's error? (previously it was internal error).

service/dap/server.go
hyangah: Hoping that user specified the package directory or file directory as buildDir, so even when `go list` fails for some reason, that can be a closer approximation to how users would run `dlv test`.

service/dap/server.go
polinasok: Previous code was wrong. Failure to launch is not a user error. It's an error to satisfy the launch request for any reason (bad user input, system malfunction, etc).  The internal errors were add for cases where we could not associate the error with the specific request (when not being able to parse the command or panic'ing).

service/dap/server.go
polinasok: Then you should use cleanExeName here on the default binary. And I wonder if you should use it on the user-specified value.

service/dap/server.go
polinasok: That's confusing.
Do we expect users to use buildDir or the client?

service/dap/server.go
aarzilli: > If we take this path (os.Chdir) I think buildDir is not the right name because it will be doing beyond what it looks.

Yes, it would be `delveWorkDir`, `delveCurrentDir` or something similar, and it would be documented as "directory where Delve will run" and it would be always applied and `cwd`, `traceDirPath`, `coreFilePath` and `program` would always be relative to it.

polinasok: If we go with this approach I think it should be`dlvCwd` to match `cwd`.

To confirm dlv internally will switch dlv to that directory, build in it and resolve all relative paths other than `args` against that directory.
This will help clients that cannot control where dlv is started to have dlv "relaunch" itself in a different directory.
This will NOT help clients that want to resolve paths against the workspace while building outside of the workspace. In particular, vscode-go extension will need to translate relative paths for `cwd`, `program`, `output`,`traceDirPath`, `corFilePath` into absolute workspace-based paths before starting dlv in a directory where we want to build things.

Does this sum up the solution?

aarzilli: Yes.

hyangah: gopls has nothing to do with launch.json. It runs bunch of `go list` commands underneath.

gopls had been creating its own world when the experimental workspace module feature was enabled. It merges all the go.mod files found in the workspace into one, create a temporary directory, place the merged go.mod (gopls.mod) in the temporary directory, and invoke all go commands from the temporary directory that contains the unified go.mod file.

This is problematic because other go tools including dlv and go test run by vscode-go extension don't know about it. So, the plan was initially to get the gopls workspace module directory location from gopls (gopls has a command) and make other go tools (including `dlv debug/test`, `go test`, `go list`) run from the gopls workspace module directory when building a go command is necessary. For `dlv`, we were thinking to run `dlv debug/test` from the directory with the right `-wd` flag.

(I think gopls uses `-modfile` for something else (e.g. to avoid corrupting the go.mod file in users' workspace) but I am not 100% certain.)

Good new is: eventually this experimental workspace module got evolved to the go.work proposal. Once `go.work` is adopted (go1.18), I think all of this is unnecessary. We just need to make sure all the commands to run from the directory under the `go.work` file exists. For older versions of Go, I don't know how it would work yet - gopls team is talking about emulating go.work feature from the gopls side. But, I don't know how this will be implemented, and how we address the same problem (i.e. dlv and other tools don't know about this go.work emulation). 

Ability to change the build directory feature will be still useful, when working with `dlv dap` server externally launched.

BTW Will update this PR to  use `dlvCWD`.

pkg/gobuild/gobuild.go
hyangah: Obsolete

service/dap/server.go

service/dap/server.go

service/dap/server.go

service/dap/server.go
hyangah: Now fallback to . (the delve's working directory)
Most users shouldn't set delveCwd.

service/dap/server_test.go
hyangah: done.

service/dap/server.go
hyangah: Obsolete

service/dap/server.go
hyangah: Abs is not a no-op, strictly speaking. It also calls filepath.Clean. But I guess that's harmless in most cases. Done.

service/dap/server.go
We end up here only if there is no user-specified value.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM - @polinasok feel free to merge after your +1.

hyangah:
Merged master. I hope this can be reviewed soon - so I don't have to handle merge conflicts again.

polinasok (commented):
Thank you for reworking build dir into delve cwd! LGTM

PR #2666, opened 2021-08-19
Documentation: add notes on porting Delve to other architectures

aarzilli:
Once in a while we get asked about porting Delve, I thought we should write it down somewhere.


derekparker (approved):
LGTM

PR #2670, opened 2021-08-23
pkg/proc: attempt to leverage debuginfod-find

derekparker:
If we cannot find debug information locally on the system, try and
invoke debuginfod-find in order to ask the linux distribution we're on
to find the debug information for us.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -14,6 +14,7 @@ import (
 	"go/token"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


derekparker:
cc @fche

fche:
(nice and simple looking!  note debuginfod-find can also fetch source code!)

pkg/proc/bininfo.go
```diff
@@ -1111,8 +1109,20 @@ func (bi *BinaryInfo) openSeparateDebugInfo(image *Image, exe *elf.File, debugIn
...
+	// We cannot find the debug information locally on the system. Try and see if we're on a system that
+	// has debuginfod so that we can use that in order to find any relevant debug information.
 	if debugFilePath == "" {
-		return nil, nil, ErrNoDebugInfoFound
+		const debuginfodFind = "debuginfod-find"
+		if _, err := exec.LookPath(debuginfodFind); err == nil {
+			cmd := exec.Command(debuginfodFind, "debuginfo", desc1+desc2)
+			out, err := cmd.CombinedOutput()
+			if err != nil {
+				return nil, nil, err
```
aarzilli, line 1120:
```
return nil, nil, fmt.Errorf("could not open debug info, additionally debuginfod-find returned %v %q", err, string(out))
```

otherwise the error is going to be "exit status 1", or something like that.

aarzilli: or maybe we should just return ErrNoDebugInfoFound and log the output of debuginfod-find instead. It's more likely that we end up in this path because the executable is stripped accidentally than because debuginfod-find could have helped but didn't.

derekparker: Done.

fche:
BTW, Re. source file fetching .... `pkg/terminal/command.go`  `printfile()` could be the right place to call out to `debuginfod-find source BUILDID /foo/bar.go`, if it has available to it the buildid of the target thread's current executable.

aarzilli (approved):

PR #2671, opened 2021-08-24
service/dap: set non-build launch/attach errors as visible

polinasok:
In general, we deal with 2 types of launch/attach errors:
1. Errors printed to Debug Console via OutputEvent that DO NOT need visible reporting from client (`showUser=false`)
2. Errors NOT printed to Debug Console that DO need visible reporting from client (`showUser=true`).

However, vscode hasn't been honoring the protocol: it triggered a modal alert for launch/attach errors regardless of the `showUser` value and showed a duplicate visible error with `showUser=true`. Therefore, we have been setting `showUser=false` for all launch/attach errors, which is omitted during json conversion as per types generated via [go-dap](https://github.com/google/go-dap).

microsoft/vscode#128484 fixes this behavior. New editor version has been released and now has:
* no `showUser` => visible error (for backward compatibility)
* `showUser=false` => no visible error
* `showUser=true` => visible error

After this change we will have:
* non-build errors: `showUser=true` - 1 pop-ups with new vscode version, 2 pop-ups with old vscode version
* build errors (details in Console): `showUser=false` - 1 pop-up with any vscode version

After https://github.com/google/go-dap/pull/60 is pulled in, we will have:
* build errors (details in Console): `showUser=false` - 1 pop-up with old vscode version, no pop-up with new vscode version (our goal)

Updates golang/vscode-go#1375


polinasok:
The vscode version that we are tailoring this for has been released, so this change can go in now. I am going to release a new version of go-dap as well, but would prefer to pull that in as a separate change.

polinasok:
@suzmue @hyangah

suzmue (approved):
LGTM.

hyangah:
The choice of word - long errors vs short errors - is a bit strange. It's more about whether to show the modal window, isn't it? Whether the error message is printed in DEBUG CONSOLE is not relevant - IMO the error message shown in the modal window printed in DEBUG CONSOLE as well is not bad. Or, is this long error vs short error part of the 'protocol'? 

Other than that, LGTM. Thanks for fixing the issue.

polinasok:
> The choice of word - long errors vs short errors - is a bit strange.

It's not the best. And I am open to suggestions. But the length of the message (and also whether it was readable when centered) was the main reason why we changed build errors from reporting details in the modal pop-up to printing them to Debug Console. And that had the added benefit of editing code while still having the error details on the screen. That would not have been so important for a short message that the user can just remember.

> It's more about whether to show the modal window, isn't it?

Whether a modal window or something else is used to display visible errors is not part of the protocol. It's up to the clients how they make visible errors visible.

To me this is about visibility. We want all error messages to be visible to the user. Silent failures are not helpful. We can either rely on the default visibility mechanism or bypass it and use an OutputEvent. AFAIK we made that choice best on message length.

> Whether the error message is printed in DEBUG CONSOLE is not relevant - IMO the error message shown in the modal window printed in DEBUG CONSOLE as well is not bad. 

I am not sure what you mean. We had the users specifically complain that they did not want the visible modal window because it was redundant and annoying and the useful information was in the DEBUG CONSOLE anyway. So whether the error message was printed there was relevant. And if you don't use the modal nor the Debug Console, you will end up with a confusing silent failure. So I think the relationship between the two mechanisms is very relevant.

> Or, is this long error vs short error part of the 'protocol'?

Long vs short is not part of the protocol. But as main editor implementing the protocol did not support long messages well, we had to improvise.

hyangah:
> > The choice of word - long errors vs short errors - is a bit strange.
> 
> It's not the best. And I am open to suggestions. But the length of the message (and also whether it was readable when centered) was the main reason why we changed build errors from reporting details in the modal pop-up to printing them to Debug Console. And that had the added benefit of editing code while still having the error details on the screen. That would not have been so important for a short message that the user can just remember.
> > Whether the error message is printed in DEBUG CONSOLE is not relevant - IMO the error message shown in the modal window printed in DEBUG CONSOLE as well is not bad.
> 
> I am not sure what you mean. We had the users specifically complain that they did not want the visible modal window because it was redundant and annoying and the useful information was in the DEBUG CONSOLE anyway.

golang/vscode-go#1375 users were complaining because the modal window interrupts the workflow (they need to click the button to let it go away), not because it was redundant. Yes, long error messages aren't desirable for modal window because modal is not the best display area for (literally) long messages. However, there could be a type of error where the detailed error message is presented in the DEBUG CONSOLE, while a short summary message suitable for a modal window can be shown at the same time. I don't mean we should implement it though.

>  So whether the error message was printed there was relevant. And if you don't use the modal nor the Debug Console, you will end up with a confusing silent failure. So I think the relationship between the two mechanisms is very relevant.
> 

I meant they don't need to be exclusive to each other. You can use BOTH for certain types of errors. 
Personally I think the previous behavior was perfectly fine for some type of users who do not continuously trigger debug sessions (i.e. build error details in DEBUG CONSOLE, and short message as a modal). 

> > Or, is this long error vs short error part of the 'protocol'?
> 
> Long vs short is not part of the protocol. But as main editor implementing the protocol did not support long messages well, we had to improvise.

That's what I thought but "vscode hasn't been honoring the protocol" - this sounded as if vscode isn't following a certain protocol that it is supposed to do.  Maybe that's just me.

polinasok:
> > To me this is about visibility. We want all error messages to be visible to the user. Silent failures are not helpful. We can either rely on the default visibility mechanism or bypass it and use an OutputEvent. AFAIK we made that choice based on message length.
> 
> I originally filed [golang/vscode-go#1502](https://github.com/golang/vscode-go/issues/1502) not because of the message length, but because that build failure log is suitable for debug output. Editing code while still having the error details on the screen is not the added benefit, but the main issue I hoped the fix to address.

I was referring to #2482, which happened earlier. We were not thinking about needing the error on the screen or that having to close the modal for consecutive compile errors adds a substantial click overhead. When errors are really short, you can just remember them. For example, it might be nice to see what the bad attribute while you fix it, but it is not critical.

I don't know why vscode chose this modal window that requires clicking. It would have been better to have a pop-up on the side that goes away once you restart the session. The other error pop-ups we get (e.g. for dlvLoadConfig) don't have the size nor the center problem. If vscode used those, we would not be making any of these changes.

> > > Whether the error message is printed in DEBUG CONSOLE is not relevant - IMO the error message shown in the modal window printed in DEBUG CONSOLE as well is not bad.
> > 
> > whether the error message was printed there was relevant. And if you don't use the modal nor the Debug Console, you will end up with a confusing silent failure. So I think the relationship between the two mechanisms is very relevant.
> 
> I meant they don't need to be exclusive to each other. You can use BOTH for certain types of errors.

Yes, I agree that we can have both. But we should not skip both. It is only ok to turn off `showUser` if we find another way to communicate the error. Having modal doesn't prevent us from also printing to Debug Console. Having Debug Console error allows us to skip the modal.

> > > Or, is this long error vs short error part of the 'protocol'?
> > 
> > Long vs short is not part of the protocol. But as main editor implementing the protocol did not support long messages well, we had to improvise.
> 
> That's what I thought but "vscode hasn't been honoring the protocol" - this sounded as if vscode isn't following a certain protocol that it is supposed to do. Maybe that's just me.

Before the change vscode was not honoring `showUser=false`. The protocol gave us a choice if we wanted the ErrorResponse to be shown or not, but the editor did not. When to use that option (for long vs short or critical vs not messages) is the implementers' decision.

hyangah:
> I was referring to #2482, which happened earlier. We were not thinking about needing the error on the screen or that having to close the modal for consecutive compile errors adds a substantial click overhead. When errors are really short, you can just remember them. For example, it might be nice to see what the bad attribute while you fix it, but it is not critical.

FYI https://github.com/go-delve/delve/pull/2482 (created on May 14) is a fix for golang/vscode-go#1502 (filed on May 13 as a feedback for the initial fix for https://github.com/golang/vscode-go/issues/1425) even though it was not linked. 

> I don't know why vscode chose this modal window that requires clicking. It would have been better to have a pop-up on the side that goes away once you restart the session. The other error pop-ups we get (e.g. for dlvLoadConfig) don't have the size nor the center problem. If vscode used those, we would not be making any of these changes.

IIUC the launch failure error is special and most likely needs attention from users, so the modal window is not unreasonable. VS Code uses the notification popups (those appear on the side) for errors with `showUser` when handling other types of requests. (example: https://github.com/golang/vscode-go/issues/143)  We have this issue just because the extension let delve handle compilation too unlike other extensions that handle target build through other means (tasks, etc). 

> Yes, I agree that we can have both. But we should not skip both. It is only ok to turn off showUser if we find another way to communicate the error. Having modal doesn't prevent us from also printing to Debug Console. Having Debug Console error allows us to skip the modal.

No body suggested we skip both. 

Seems like the discussion derailed. I think this PR itself is good to go.

polinasok:
I updated the PR description to remove long/short details.

aarzilli (approved):

derekparker (approved):
LGTM

PR #2673, opened 2021-08-24
tests: correct checks that depend on regabi (fixes Cirrus-CI builds)

aarzilli:
The tests incorrectly assumed that regabi was enabled everywhere on Go1.17/amd64, in fact it is not enabled on freebsd.


derekparker (approved):
LGTM

PR #2674, opened 2021-08-24
service/debugger: return correct exit status on manual halt

suzmue:
Make sure that the exit status for the process is correctly output.

Fixes golang/vscode-go#1717


polinasok:
Why does this trigger with dap but not rpc (based on the logs in the issue)?

aarzilli (commented):
Shouldn't it be doRunCommand that sends the output even after the process exits?

suzmue:
This is triggered by DAP because the output message is coming from the disconnect request, not from doRunCommand (which is where the exited error first appears).

I could also rework where the message is coming from, but I will have to make sure we won't be getting duplicate messages (even though we are getting duplicate terminated events, these are not user facing so are more acceptable).

polinasok (approved):

polinasok:
We could consider moving the exited message to doRunCommand... But would the reason be because it is better design or because we would rather not touch the exit code propagation code? Even outside of that decision, it does seem like the right thing to get the same code when querying for debugging state more than once. Also, this refactor would make the order of messages a bit more unpredictable for testing, won't it?

polinasok:
@aarzilli Is there a  still a concern around the proposed approach?

aarzilli:
I think service/dap should print the process exited message when the process exits not when it tries to manipulate it after it has exited.

suzmue:
> I think service/dap should print the process exited message when the process exits not when it tries to manipulate it after it has exited.

I agree with this. Would it be alright to leave this as a TODO for now?

aarzilli (approved):

derekparker (approved):
LGTM

PR #2294, opened 2021-01-09
terminal: Go syntax highlighting for listings

aarzilli:
Fixes #1273

PR #2675, opened 2021-08-24
Avoid bright colors in default syntax highlighting

krobelus:
These two were hard to read on default xterm or on Konsole's "White on
Black" theme. The non-bright colors still work on dark backgrounds.

See https://github.com/go-delve/delve/pull/2294#discussion_r691076091


aarzilli (approved):
LGTM but how do we know someone else won't complain?

PR #2676, opened 2021-08-24
Prefer string variant for source-list-line-color config option

krobelus:
This commit allows to use either of the following two lines in
~/.config/dlv/config.yml to highlight keywords in red

	source-list-keyword-color: "\x1b[31m"
	source-list-keyword-color: 31

This was already possible for "source-list-line-color", which was also
in the default config with the integer syntax.

I don't have a strong opinion on whether we should support both, but
it took me a bit to figure out the correct syntax for "\x1b[31m"
(Are quotes required? Can I write \033 or \e instead of \x1b?)
so it seems alright to support bare numbers.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -53,9 +53,8 @@ type Config struct {
 	// expression for its argument.
 	ShowLocationExpr bool `yaml:"show-location-expr"`
 
-	// Source list line-number color (3/4 bit color codes as defined
-	// here: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors),
-	// or a string containing a terminal escape sequence.
+	// Source list line-number color, as a terminal escape sequence.
+	// For historic reasons, this can also be an integer color code.
 	SourceListLineColor interface{} `yaml:"source-list-line-color"`
 
 	// Source list arrow color, as a terminal escape sequence.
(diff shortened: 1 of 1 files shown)


aarzilli (commented):
The documentation needs to be changed, both of the fields of Config as well as in writeDefaultConfig.

> I don't have a strong opinion on whether we should support both, but it took me a bit to figure out the correct syntax for "\x1b[31m" (Are quotes required? Can I write \033 or \e instead of \x1b?) so it seems alright to support bare numbers.

This is all documented by the YAML specification.

krobelus:
> The documentation needs to be changed, both of the fields of Config as well as in writeDefaultConfig.

Wow, I didn't realize that the default config actually contained examples with `"\x1b[..."`, because I had a config file written by `config -save` - that should probably include the helpful defaults but that's another issue.

With that in mind I think there's no need for the integer variant; I have updated the patch to simply remove the mention of the integer option from the default config.

aarzilli (approved):

aarzilli:
> Wow, I didn't realize that the default config actually contained examples with `"\x1b[..."`, because I had a config file written by `config -save` - that should probably include the helpful defaults but that's another issue.

That's a good idea.

derekparker (approved):
LGTM

PR #2678, opened 2021-08-28
service/dap: add test coverage for Stop()

polinasok:


aarzilli (approved):

polinasok:
I am not even sure what DeepSource is unhappy about, but since it is not any code I touched in this PR, I can ignore it, right?

aarzilli:
> I am not even sure what DeepSource is unhappy about, but since it is not any code I touched in this PR, I can ignore it, right?

Yes. Also that's bad advice and you should generally ignore it even if it was code that you wrote.

aarzilli:
PS. same thing with the other error they added recently where they recommend using `_` for unused parameter.

derekparker (approved):
LGTM

PR #2682, opened 2021-08-31
tests: fix TestIssue419

aarzilli:
errChan should not be closed otherwise the function that requests the
manual stop might find it closed.


derekparker (approved):
LGTM

PR #2683, opened 2021-08-31
service/dap: include selected goroutine in threads request

suzmue:
When we set a limit on the number of threads that would be
returned, it was possible that the selected thread was not
included in the list of threads. This could cause issues
because the stopped reason is associated with the selected
goroutine, so users could be missing out on important info.

This change makes sure that the selected goroutine is included
by adding it to the end of the list.


aarzilli (approved):

derekparker (approved):
LGTM

PR #2684, opened 2021-09-01
pkg/goversion: visibly warn the user with --check-go-version=false

polinasok:
Since Delve doesn't forcefully take out support for no-longer-supported versions, users and clients might want to turn off the version check to continue debugging with older versions of Go. Add a warning to notify them of the incompatibility in this case as well.  This will apply to too-new Go versions as well. Also, log the Go version.

Before
```
$ dlv debug _fixtures/increment.go 
Version of Go is too old for this version of Delve (minimum supported version 1.15, suppress this error with --check-go-version=false)
$ dlv debug _fixtures/increment.go --check-go-version=false
(dlv)
```

After
```
$ ./dlv debug _fixtures/increment.go
Go version 1.14.15 is too old for this version of Delve (minimum supported version 1.15, suppress this error with --check-go-version=false)
$ ./dlv debug _fixtures/increment.go --check-go-version=false
WARNING: undefined behavior - Go version 1.14.15 is too old for this version of Delve (minimum supported version 1.15)
(dlv)
```

Updates #golang/vscode-go/issues/1716

### DIFF
--- a/pkg/goversion/compat.go
+++ b/pkg/goversion/compat.go
@@ -2,6 +2,7 @@ package goversion
 
 import (
 	"fmt"
+	"os"
 )
 
 var (
@@ -10,20 +11,30 @@ var (
 	MaxSupportedVersionOfGoMajor = 1
 	MaxSupportedVersionOfGoMinor = 17
 	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	goTooOldWarn                 = fmt.Errorf("Warning: version of Go is too old for this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
 	dlvTooOldErr                 = fmt.Errorf("Version of Delve is too old for this version of Go (maximum supported version %d.%d, suppress this error with --check-go-version=false)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
+	dlvTooOldWarn                = fmt.Errorf("Warning: version of Delve is too old for this version of Go (maximum supported version %d.%d)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
 )
 
 // Compatible checks that the version specified in the producer string is compatible with
 // this version of delve.
-func Compatible(producer string) error {
+func Compatible(producer string, warnonly bool) error {
 	ver := ParseProducer(producer)
 	if ver.IsDevel() {
 		return nil
 	}
 	if !ver.AfterOrEqual(GoVersion{MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor, -1, 0, 0, ""}) {
+		if warnonly {
+			fmt.Fprintf(os.Stderr, "%v\n", goTooOldWarn)
+			return nil
+		}
 		return goTooOldErr
 	}
 	if ver.AfterOrEqual(GoVersion{MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor + 1, -1, 0, 0, ""}) {
+		if warnonly {
+			fmt.Fprintf(os.Stderr, "%v\n", dlvTooOldWarn)
+			return nil
+		}
 		return dlvTooOldErr
 	}
 	return nil
--- a/pkg/goversion/compat.go
+++ b/pkg/goversion/compat.go
@@ -10,10 +10,10 @@ var (
 	MinSupportedVersionOfGoMinor = 15
 	MaxSupportedVersionOfGoMajor = 1
 	MaxSupportedVersionOfGoMinor = 17
-	goTooOldErr                  = fmt.Errorf("Version of Go may be too old to work with this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
-	goTooOldWarn                 = fmt.Errorf("Warning: version of Go may be too old to work with this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
-	dlvTooOldErr                 = fmt.Errorf("Version of Delve is too old for this version of Go (maximum supported version %d.%d, suppress this error with --check-go-version=false)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
-	dlvTooOldWarn                = fmt.Errorf("Warning: version of Delve is too old for this version of Go (maximum supported version %d.%d)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
+	goTooOldErr                  = fmt.Sprintf("Go version %%s is too old for this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	goTooOldWarn                 = fmt.Sprintf("WARNING: undefined behavior - Go version %%s is too old for this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	dlvTooOldErr                 = fmt.Sprintf("Version of Delve is too old for Go version %%s (maximum supported version %d.%d, suppress this error with --check-go-version=false)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
+	dlvTooOldWarn                = fmt.Sprintf("WARNING: undefined behavior - version of Delve is too old for Go version %%s (maximum supported version %d.%d)", MaxSupportedVersionOfGoMajor, MaxSupportedVersionOfGoMinor)
 )
 
 // Compatible checks that the version specified in the producer string is compatible with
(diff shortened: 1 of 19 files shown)


polinasok:
cc @hyangah

aarzilli (commented):
We shouldn't do this.

hyangah:
Can't this warning message be logged from the DAP layer, instead of changing all dlv command's behavior?

aarzilli:
> We shouldn't do this.

After looking at the diff:

> The general consensus appears to be that we can turn off the check

was this intended to be "we can not turn off the check"?

pkg/goversion/compat.go
```diff
@@ -10,20 +11,30 @@ var (
...
 // this version of delve.
-func Compatible(producer string) error {
+func Compatible(producer string, warnonly bool) error {
 	ver := ParseProducer(producer)
 	if ver.IsDevel() {
 		return nil
 	}
 	if !ver.AfterOrEqual(GoVersion{MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor, -1, 0, 0, ""}) {
+		if warnonly {
+			fmt.Fprintf(os.Stderr, "%v\n", goTooOldWarn)
```
aarzilli, line 28: This should go to logflags.logOut if not nil (same with the other one).

pkg/goversion/compat.go
```diff
@@ -10,20 +11,30 @@ var (
 	MaxSupportedVersionOfGoMajor = 1
 	MaxSupportedVersionOfGoMinor = 17
 	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	goTooOldWarn                 = fmt.Errorf("Warning: version of Go is too old for this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
```
aarzilli, line 14: I don't think the hint about how to disable the error should be removed.

polinasok:
> > We shouldn't do this.
> 
> After looking at the diff:
> 
> > The general consensus appears to be that we can turn off the check
> 
> was this intended to be "we can not turn off the check"?

I think I wrote it right. No "not". What we gathered from the discussion is that Delve does not aggressively clean up the code for older versions, so it is fair to expect them to work for a while longer after being dropped. And that is why there is a flag to bypass the check that will otherwise prevent the user from proceeding.

So right now there are two options:
1. Check and die
2. Turn off (1) and completely hide the problem from the user

If the user is the one doing (2), hopefully they still remember what they did and have context if something behaves oddly and remember to revisit their tool setup at the right time. But if (2) is done for the user by the editor, they will have no idea there is anything wrong with their Go version. I think they deserve a warning in that case. And not something hidden behind an optional log flag, only in dap mode, but a visible and reliable warning. What is the downside of such a warning?

pkg/goversion/compat.go
polinasok: The existing hint wording makes it sound like you can disable the warning by using the false value when in fact the false value is what triggers the warning. Unless we want to have more modes with another flag like -- warn-go-version.

hyangah:
I think the general consensus is that
  1) delve won't turn off the check by default.
  2) clients that run delve on behalf of end users can choose to turn off the checks depending on their policy.
   
 Empirically and historically, delve may continue to work with go versions just dropped from the supported version list, but there is no guarantee from delve.  Taking a risk is up to the clients. VS Code Go decided to take a risk. 

And I think it's good to have a warning even when users choose to turn off the check with the flag. Similar to the warning when delve works with an optimized binary. So, I hope this PR that prints the warning even when users disable the check to go in. If people are concerned the extra warning will bother existing users, I think at least we hope to enable this warning for the sessions launched by DAP, which is relatively new.

aarzilli:
> > > We shouldn't do this.
> > 
> > 
> > After looking at the diff:
> > > The general consensus appears to be that we can turn off the check
> > 
> > 
> > was this intended to be "we can not turn off the check"?
> 
> I think I wrote it right. No "not". What we gathered from the discussion is that Delve does not aggressively clean up the code for older versions, so it is fair to expect them to work for a while longer after being dropped. And that is why there is a flag to bypass the check that will otherwise prevent the user from proceeding.
> 
> So right now there are two options:
> 
>     1. Check and die
> 
>     2. Turn off (1) and completely hide the problem from the user
> 
> 
> If the user is the one doing (2), hopefully they still remember what they did and have context if something behaves oddly and remember to revisit their tool setup at the right time. But if (2) is done for the user by the editor, they will have no idea there is anything wrong with their Go version. I think they deserve a warning in that case. And not something hidden behind an optional log flag, only in dap mode, but a visible and reliable warning. What is the downside of such a warning?

I interpreted "The general consensus appears to be that we can turn off the check" as "The general consensus appears to be that delve can turn off the check by default" before looking at the diff. I see that you meant "we = users of delve". In general this is fine.

pkg/goversion/compat.go
aarzilli: Ok.

briandealwis:
I would be very much in favour to emitting a warning.  I'd suggest a slight change of wording in this case:
```
$ ./dlv exec loopprog --check-go-version=false
Warning: Go version 1.14 may not work with this version of Delve (minimum supported version 1.15)
```

polinasok:
> I interpreted "The general consensus appears to be that we can turn off the check" as "The general consensus appears to be that delve can turn off the check by default" before looking at the diff. I see that you meant "we = users of delve". In general this is fine.

Oops, sorry for the confusion. I should have been more clear which "we" hat (user of delve or delve dev) I was wearing when I wrote the comment. Yes, users/clients of delve have the option to turn off the check with a false flag. And in that case, I would like the problem still to be known to the end user with a warning.

polinasok:
Agree with @briandealwis that less damning and at the same time more clarifying wording of the warning and perhaps the hard error would be helpful:
- 1.14 may not work with this version of Delve
- 1.14 is not guaranteed to work with this version of Delve

pkg/goversion/compat.go
```diff
@@ -2,28 +2,39 @@ package goversion
...
 )
 
 var (
 	MinSupportedVersionOfGoMajor = 1
 	MinSupportedVersionOfGoMinor = 15
 	MaxSupportedVersionOfGoMajor = 1
 	MaxSupportedVersionOfGoMinor = 17
-	goTooOldErr                  = fmt.Errorf("Version of Go is too old for this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	goTooOldErr                  = fmt.Errorf("Version of Go may be too old to work with this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
+	goTooOldWarn                 = fmt.Errorf("Warning: version of Go may be too old to work with this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
```
briandealwis, line 14: Since we're erroring in this case, why not just say it's unsupported
```suggestion
	goTooOldErr                  = fmt.Errorf("Version of Go is not supported by this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
	goTooOldWarn                 = fmt.Errorf("Warning: version of Go may be too old to work with this version of Delve (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
```

polinasok: I wanted to keep the wording consistent. The level of support doesn't change based on the flag value, only how we treat the check. If error/warning consistency is not critical, then we might as well just leave the original error as-is.

briandealwis: It seems odd to error on a "may be".  The key point is that it's an unsupported version, and it's unsupported because it's too old.  WDYT:
```suggestion
	goTooOldErr                  = fmt.Errorf("Version of Go is not supported by this version of Delve (minimum supported version %d.%d, suppress this error with --check-go-version=false)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
	goTooOldWarn                 = fmt.Errorf("Warning: version of Go is not supported by this version of Delve and may not function correctly (minimum supported version %d.%d)", MinSupportedVersionOfGoMajor, MinSupportedVersionOfGoMinor)
```

pkg/goversion/compat.go
polinasok: There are other warnings that don't use logOut and go to os.Stderr. What makes this one different?
Wouldn't it potentially send the message to a file with log-dest along with other logging? I don't think that's what desirable here.

pkg/goversion/compat.go
polinasok: I revered to the original "too old" wording, but with a clarification about undefined behavior in the warning version. Also added Go version number to the message.

pkg/goversion/compat.go
aarzilli: For the most part if we print to stderr we also exit, otherwise it wouldn't be possible to easily distinguish output from delve from the output of the program. I think all the exceptions are in pkg/config/config.go, which should be changed.

aarzilli (approved):

aarzilli:
LGTM

polinasok:
Test failures don't look related

derekparker (approved):
LGTM

derekparker:
Re-running FreeBSD builder as a sanity check, TeamCity failures seem unrelated.

Issue #2685, opened 2021-09-02
dap: Pause request while stopped prevents forward progress on following Continue


suzmue:
1. What version of Delve are you using (`dlv version`)?
at tip
2. What version of Go are you using? (`go version`)?
`go version go1.17 linux/amd64`
3. What operating system and processor architecture are you using?
linux/amd64
4. What did you do?
- Debug a process.
- Send a manual halt request
- Continue

5. What did you expect to see?
I expected the process to continue beyond the location it was stopped at
6. What did you see instead?
Process is still stopped at same location 

From my investigation it seems that the signal sent to halt the process, is being received by the Continue call, blocking the program from making forward progress

suzmue:
This is causing the test failures on linux in PR #2634

aarzilli:
This is working as intended.

suzmue:
> This is working as intended.

Updated the issue to address the issue in DAP only.

polinasok:
This can be closed now, right? Halt while halted may or may not be applied to the Continue that follows, so we just make sure not to issue double halts.

PR #2686, opened 2021-09-02
service/dap: do not send a halt request if the debuggee is not running

suzmue:
This change makes sure to avoid sending a pause request while the debuggee is not running, to try to protect
the user from the undefined Halt behavior:

```
	// Halt suspends the process.
	// The effect of Halt while the target process is stopped, or in the
	// process of stopping, is operating system and timing dependent. It will
	// either have no effect or cause the following resume to stop immediately.
```

It is not possible to fully protect the dap client, and the halt request may still have the same issue.

This adds test to service/dap to test this issue.

Updates go-delve/delve#2685

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1654,10 +1654,12 @@ func (s *Server) doStepCommand(command string, threadId int, asyncSetupDone chan
 // onPauseRequest handles 'pause' request.
 // This is a mandatory request to support.
 func (s *Server) onPauseRequest(request *dap.PauseRequest) {
-	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
-	if err != nil {
-		s.sendErrorResponse(request.Request, UnableToHalt, "Unable to halt execution", err.Error())
-		return
+	if s.debugger.IsRunning() {
+		_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
+		if err != nil {
+			s.sendErrorResponse(request.Request, UnableToHalt, "Unable to halt execution", err.Error())
+			return
+		}
 	}
 	s.send(&dap.PauseResponse{Response: *newResponse(request.Request)})
 	// No need to send any event here.
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -36,10 +36,6 @@ type nativeProcess struct {
 	// signalled to stop as a result of a Halt API call. Used to disambiguate
 	// why a thread is found to have stopped.
 	manualStopRequested bool
-	// manualStopPending is set if there is a signal sent by a Halt API call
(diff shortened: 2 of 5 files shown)


suzmue:
@aarzilli @derekparker If there is a better way to go about fixing this, please feel free to re-use my tests in a different PR

aarzilli:
See test failures. In general: I don't think there is a way to do what you want to do. The behavior of RequestManualStop is documented in service/api/types.go:

```
	// Halt suspends the process.
	// The effect of Halt while the target process is stopped, or in the
	// process of stopping, is operating system and timing dependent. It will
	// either have no effect or cause the following resume to stop immediately.
```

For #2634 I would suggest writing the tests in a way that does not rely on undefined behavior.

aarzilli:
PS. the code that you are changing is our critical section, it's very easy to introduce race conditions with the kernel there that happen maybe 1 or 2% of the times our tests are run. Usually when I change anything in there, unless it's a completely trivial change, I'll run all the tests in pkg/proc 200 times on all affected architectures and the specific test I'm looking for 1000 times.

suzmue:
> See test failures. In general: I don't think there is a way to do what you want to do. The behavior of RequestManualStop is documented in service/api/types.go:
> 
> ```
> 	// Halt suspends the process.
> 	// The effect of Halt while the target process is stopped, or in the
> 	// process of stopping, is operating system and timing dependent. It will
> 	// either have no effect or cause the following resume to stop immediately.
> ```
> 
> For #2634 I would suggest writing the tests in a way that does not rely on undefined behavior.

Thanks for pointing me to that. For some reason I thought that Halt while stopped was a no-op, I see it is clearly defined as otherwise.

I can instead make this a DAP issue and try to minimize the likelihood of hitting the race condition (and write tests that do not rely on this.)

aarzilli (approved):

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -1654,10 +1654,12 @@ func (s *Server) doStepCommand(command string, threadId int, asyncSetupDone chan
 // onPauseRequest handles 'pause' request.
 // This is a mandatory request to support.
 func (s *Server) onPauseRequest(request *dap.PauseRequest) {
-	_, err := s.debugger.Command(&api.DebuggerCommand{Name: api.Halt}, nil)
-	if err != nil {
-		s.sendErrorResponse(request.Request, UnableToHalt, "Unable to halt execution", err.Error())
-		return
+	if s.debugger.IsRunning() {
```
polinasok, line 1657: Should we be doing the same in StopDebugSession (like is now done in rpccommon::Stop) or it doesn't matter because there will not be another run request to halt after the end?

suzmue: Yeah it should not matter because we won't be resuming execution.

polinasok (approved):
LGTM

PR #2687, opened 2021-09-03
proc: only apply regabiMallocgcWorkaround to first image

aarzilli:
The workaround for runtime.mallocgc with regabi should only be applied
to the first image.


aarzilli:
This was merged together with the Go 1.18 changes.

PR #2693, opened 2021-09-09
service/dap: return selected goroutine first

suzmue:
Often in the VS Code UI, it can be tough to find the selected goroutine, even though it is already expanded. Putting the selected goroutine at the top will make it much easier for users to quickly find the interesting goroutine with the relevant stop reason.


suzmue:
cc @hyangah @polinasok

service/dap/server_test.go
```diff
@@ -967,7 +967,14 @@ func TestStackTraceRequest(t *testing.T) {
 	})
 }
 
-func TestSelectedThreadsRequest(t *testing.T) {
+func TestSelectedGoroutine(t *testing.T) {
+	saved := os.Getenv("GOMAXPROCS")
+	defer func() {
```
aarzilli, line 972: No need to wrap this in `func() { }()`

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

hyangah (approved):

service/dap/server.go
```diff
@@ -1389,30 +1389,34 @@ func (s *Server) onThreadsRequest(request *dap.ThreadsRequest) {
...
-				var selectedFound bool
-				for _, g := range gs {
-					if g.ID == state.SelectedGoroutine.ID {
-						selectedFound = true
-						break
-					}
+		}
+		// Make sure the selected goroutine is included in the list of threads
+		// to return.
+		selectedIdx := -1
```
polinasok, line 1395: For a moment I thought this was the goid. Maybe use -100 or constant IndexNotFound, so this is not confused with -1 for selected goroutine?

polinasok:
Does this rearrange the goroutine list to move the selected one to the top or duplicates it at the top?
I remember that we discussed that it might be odd if the position of the goroutine jumps around. Somebody might have their view scrolled to a specific goroutine and then have it disappear because it was bumped to the top. Have you had a chance to compare both approaches?

suzmue:
> Does this rearrange the goroutine list to move the selected one to the top or duplicates it at the top (with unique id -1)?
> I remember that we discussed that it might be odd if the position of the goroutine jumps around. Somebody might have their view scrolled to a specific goroutine and then have it disappear because it was bumped to the top. Have you had a chance to compare both approaches?

This rearranges by moving the selected goroutine to the top.

With duplicating the goroutine I worry a bit about artificially creating additional goroutines, which they can step on either to go through the program.

The VS Code UI does snap to the goroutine id returned by the stopped event. I am not certain that doing this is necessary, and will perhaps add conclusion. So I am going to close this for now unless we have a more strong signal that this would be helpful (and which of the two solutions would be more helpful)

polinasok:
We could also add a flag to control this and then users could configure which way they want. Too much?

polinasok:
@suzmue Checked this again. The editor does snap to the selected goroutine, so I think we don't need to do anything with ordering to highlight it. Why did we think a while back that we did need to do something? Did it not snap before? Or was this because it can be annoying to visually separate the goroutine stuck in the middle of many pre-expanded goroutines from previous stops?

PR #2694, opened 2021-09-09
service/dap: show decimal and hex values for uint

suzmue:
As discussed in golang/vscode-go#1459, users may want to see different representations of bytes and runes.

This change displays the hex value for all uints alongside the decimal value.
In addition to the variables response, this will impact the values in all evaluate contexts.

Examples:
```
myuint8: 255 = 0xff
myuint16: 65535 = 0xffff
myuint32: 4294967295 = 0xffffffff
myuint64: 18446744073709551615 = 0xffffffffffffffff
```

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2016,20 +2016,10 @@ func (s *Server) metadataToDAPVariables(v *fullyQualifiedVariable) ([]dap.Variab
 
 func isListOfBytesOrRunes(v *proc.Variable) bool {
 	if len(v.Children) > 0 && (v.Kind == reflect.Array || v.Kind == reflect.Slice) {
-		return isByte(&v.Children[0]) || isRune(&v.Children[0])
+		childKind := v.Children[0].RealType.Common().ReflectKind
+		return childKind == reflect.Uint8 || childKind == reflect.Int32
 	}
 	return false
-}
-
-func isByte(v *proc.Variable) bool {
-	return v.RealType.Common().ReflectKind == reflect.Uint8
-}
-
-func isRune(v *proc.Variable) bool {
-	if v.RealType == nil || v.RealType.Common() == nil {
-		return false
-	}
-	return v.RealType.Common().ReflectKind == reflect.Int32
 }
 
 func (s *Server) getTypeIfSupported(v *proc.Variable) string {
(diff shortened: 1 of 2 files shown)


suzmue:
@hyangah @polinasok

service/dap/server.go
```diff
@@ -2107,6 +2108,9 @@ func (s *Server) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr s
 	}
 
 	switch v.Kind {
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		n, _ := strconv.ParseUint(api.ConvertVar(v).Value, 10, 64)
+		value = fmt.Sprintf("%s = 0x%x", value, n)
```
aarzilli, line 2113: `%#x`

aarzilli (commented):
what happens if users that want to see unsigned integers as decimals exist (you wouldn't know if they did because they wouldn't have had a reason to speak until now)?

suzmue:
> what happens if users that want to see unsigned integers as decimals exist (you wouldn't know if they did because they wouldn't have had a reason to speak until now)?

This change does not remove the decimal value, but instead displays both.

Screen Shot 2021-09-13 at 12 18 49 PM

 If we decide to display just one, we would definitely put this behind a configuration for these users.

service/dap/server.go

aarzilli (approved):

hyangah:
@suzmue Can you capture and share the screenshot of arrays? And this does not affect "Copy Value" and for Copy Value (clipboard context evaluate req) this will be still decimal, right?

suzmue:
This is what it looks like in arrays:

Screen Shot 2021-09-14 at 3 42 38 PM

It currently applies to all contexts, but we could definitely change this to not include clipboard context.

polinasok:
I suggest updating the PR description with the information about what will show up via Copy Value, hover, etc. 
And in case of uint8, we will have both the string child  and this feature?

polinasok (approved):

suzmue:
Screen Shot 2021-09-15 at 2 27 26 PM

From offline discussion, we will for now leave both values in all contexts.

polinasok:
> From offline discussion, we will for now leave both values in all contexts.

I updated the top PR comment.

PR #2699, opened 2021-09-14
pkg/proc: merge register data in (*compositeMemory).WriteMemory

### LINKED ISSUE #2698: pkg/proc: failed to call value method on struct type: wrong number of bytes for register Rax

```
Delve Debugger
Version: 1.7.1
Build: $Id: 3bde2354aafb5a4043fd59838842c4cd4a8b6f0b $
```

Go 1.17.

`linux/amd64`

4. What did you do?

Take the following program:

```
package main

import "fmt"

type T struct {
	a uint32
	b uint8
	c uint8
	d uintptr
}

func (t T) String() string {
	return fmt.Sprintf("%d %d %d %d", t.a, t.b, t.c, t.d)
}

func main() {
	t := T{
		a: 1,
		b: 2,
		c: 3,
		d: 4,
	}
	printT(t)
}

//go:noinline
func printT(t T) {
	fmt.Println(t.String())
}
```
(https://play.golang.org/p/PDweiOxPae_1)

and ran

```
dlv debug
(dlv) break main.go:23
(dlv) c
(dlv) call t.String()
```

Debug call prints `1 2 3 4`.

6. What did you see instead?

```
(dlv) call t.String()
> main.main() ./main.go:23 (PC: 0x498d04)
Command failed: wrong number of bytes for register Rax (4)
```

CC @heschi @aarzilli

mknyszek:
Right now, if (*compositeMemory).WriteMemory needs to write a value to
a register that's smaller than the full size of the register (say, a
uint32 being passed as an argument), then (*AMD64Registers).SetReg can
later fail a sanity check that ensures the passed DwarfRegister is a
full size register.

Fix this by reading the old value of the register and overwriting just
the relevant parts with the new register. For the purposes of an
argument, it would probably be fine to just pad with zeroes, but merging
with the existing value is more generally correct.

Fixes #2698

### DIFF
--- a/pkg/proc/mem.go
+++ b/pkg/proc/mem.go
@@ -179,7 +179,9 @@ func (mem *compositeMemory) WriteMemory(addr uint64, data []byte) (int, error) {
 
 			switch piece.Kind {
 			case op.RegPiece:
-				err := mem.regs.ChangeFunc(piece.Val, op.DwarfRegisterFromBytes(pieceMem))
+				oldReg := mem.regs.Reg(piece.Val)
+				newReg := op.DwarfRegisterFromBytes(pieceMem)
+				err := mem.regs.ChangeFunc(piece.Val, oldReg.Overwrite(newReg))
 				if err != nil {
 					return donesz, err
 				}
--- a/_fixtures/fncall.go
+++ b/_fixtures/fncall.go
@@ -176,6 +176,17 @@ func regabistacktest2(n1, n2, n3, n4, n5, n6, n7, n8, n9, n10 int) (int, int, in
 	return n1 + n2, n2 + n3, n3 + n4, n4 + n5, n5 + n6, n6 + n7, n7 + n8, n8 + n9, n9 + n10, n10 + n1
 }
 
+type Issue2698 struct {
+	a uint32
+	b uint8
+	c uint8
+	d uintptr
+}
+
+func (i Issue2698) String() string {
+	return fmt.Sprintf("%d %d %d %d", i.a, i.b, i.c, i.d)
+}
+
 func main() {
 	one, two := 1, 2
 	intslice := []int{1, 2, 3}
@@ -192,6 +203,12 @@ func main() {
 	var pable_pa PRcvrable = pa
 	var x X = 2
 	var x2 X2 = 2
+	issue2698 := Issue2698{
+		a: 1,
+		b: 2,
+		c: 3,
+		d: 4,
+	}
 
 	fn2clos := makeclos(pa)
 	fn2glob := call1
@@ -208,5 +225,5 @@ func main() {
 	d.Method()
 	d.Base.Method()
 	x.CallMe()
-	fmt.Println(one, two, zero, call, call0, call2, callexit, callpanic, callbreak, callstacktrace, stringsJoin, intslice, stringslice, comma, a.VRcvr, a.PRcvr, pa, vable_a, vable_pa, pable_pa, fn2clos, fn2glob, fn2valmeth, fn2ptrmeth, fn2nil, ga, escapeArg, a2, square, intcallpanic, onetwothree, curriedAdd, getAStruct, getAStructPtr, getVRcvrableFromAStruct, getPRcvrableFromAStructPtr, getVRcvrableFromAStructPtr, pa2, noreturncall, str, d, x, x2.CallMe(5), longstrs, regabistacktest, regabistacktest2)
+	fmt.Println(one, two, zero, call, call0, call2, callexit, callpanic, callbreak, callstacktrace, stringsJoin, intslice, stringslice, comma, a.VRcvr, a.PRcvr, pa, vable_a, vable_pa, pable_pa, fn2clos, fn2glob, fn2valmeth, fn2ptrmeth, fn2nil, ga, escapeArg, a2, square, intcallpanic, onetwothree, curriedAdd, getAStruct, getAStructPtr, getVRcvrableFromAStruct, getPRcvrableFromAStructPtr, getVRcvrableFromAStructPtr, pa2, noreturncall, str, d, x, x2.CallMe(5), longstrs, regabistacktest, regabistacktest2, issue2698.String())
 }
--- a/pkg/dwarf/op/regs.go
+++ b/pkg/dwarf/op/regs.go
@@ -164,3 +164,20 @@ func (reg *DwarfRegister) FillBytes() {
 	reg.Bytes = make([]byte, 8)
 	binary.LittleEndian.PutUint64(reg.Bytes, reg.Uint64Val)
 }
+
+// Overwrite takes the contents of reg1 and overwrites them with the contents
+// of reg2 in little-endian order, returning a new register. The new register
+// will always contain the complete contents of both registers, so if reg2 is
+// larger than reg1, the final register will be reg2's size.
+func (reg1 *DwarfRegister) Overwrite(reg2 *DwarfRegister) *DwarfRegister {
+	reg1.FillBytes()
+	reg2.FillBytes()
+	width := len(reg1.Bytes)
+	if len(reg2.Bytes) > len(reg1.Bytes) {
+		width = len(reg2.Bytes)
+	}
+	b := make([]byte, width)
+	copy(b, reg1.Bytes)
+	copy(b, reg2.Bytes)
+	return DwarfRegisterFromBytes(b)
+}
--- a/pkg/proc/dwarf_expr_test.go
+++ b/pkg/proc/dwarf_expr_test.go
@@ -245,7 +245,7 @@ func TestDwarfExprComposite(t *testing.T) {
 	if changeCalls[0] != "3 - 2f00000000000000" {
 		t.Errorf("wrong call to SetReg (Rbx)")
 	}
-	if changeCalls[1] != "2 - 0c00" {
+	if changeCalls[1] != "2 - 0c00000000000000" {
 		t.Errorf("wrong call to SetReg (Rcx)")
 	}
 	if mem.data[0x10] != 13 || mem.data[0x11] != 0x00 {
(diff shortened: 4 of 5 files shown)


mknyszek:
This probably needs a test? But I'm not sure where to put it.

heschi:
FWIW, I still don't understand why merging the data is correct. Does the register ABI guarantee that if an int32 is stored in a 64-bit register, the upper bits will be ignored? I don't see anything to that effect in the [spec](https://go.googlesource.com/go/+/refs/heads/dev.regabi/src/cmd/compile/internal-abi.md). Otherwise, it seems like leaving old data in the register is actively dangerous.

mknyszek:
> FWIW, I still don't understand why merging the data is correct. Does the register ABI guarantee that if an int32 is stored in a 64-bit register, the upper bits will be ignored? I don't see anything to that effect in the [spec](https://go.googlesource.com/go/+/refs/heads/dev.regabi/src/cmd/compile/internal-abi.md). Otherwise, it seems like leaving old data in the register is actively dangerous.

Thinking about this again, yeah in the context of like, a load instruction, the upper bits are usually zeroed to avoid a dependency in the CPU hardware (thwarting register renaming). I'm not sure I agree that it's dangerous, maybe a bit harder to debug? That is, unless the compiler is assuming that the upper bits of the register _are_ zeroed at the other end of the call. I'm really not sure.

I'm not familiar enough with DWARF location lists to say for sure how this should work. I imagine if a location list can define multiple values that go into the same register, then merging actually is the right call, because then this code would overwrite the value completely each time (I think? IDK, it doesn't seem like there's a way to say "just put this byte at the 7th byte in the register" or something).

My reasoning behind this was that this method, `WriteMemory`, needed to be very general. Anyway, the zeroing version is even simpler. Just `FillBytes` before passing it to `ChangeFunc`. It also works. I'm open to changing the CL to work either way.

mknyszek:
> Thinking about this again, yeah in the context of like, a load instruction, the upper bits are usually zeroed to avoid a dependency in the CPU hardware (thwarting register renaming). I'm not sure I agree that it's dangerous, maybe a bit harder to debug? That is, unless the compiler is assuming that the upper bits of the register _are_ zeroed at the other end of the call. I'm really not sure.

Actually, I think it could. Imagine a case where you pass a `uint32` and its promoted to a `uint64` on the other side that's then stored on the stack or heap or something. I could see the compiler turning that into a no-op (no zero-extension, just a store).

I'm inclined to agree with Heschi on this.

mknyszek:
Updated the code to reflect this.

aarzilli:
> > Thinking about this again, yeah in the context of like, a load instruction, the upper bits are usually zeroed to avoid a dependency in the CPU hardware (thwarting register renaming). I'm not sure I agree that it's dangerous, maybe a bit harder to debug? That is, unless the compiler is assuming that the upper bits of the register _are_ zeroed at the other end of the call. I'm really not sure.
> 
> Actually, I think it could. Imagine a case where you pass a `uint32` and its promoted to a `uint64` on the other side that's then stored on the stack or heap or something. I could see the compiler turning that into a no-op (no zero-extension, just a store).
> 
> I'm inclined to agree with Heschi on this.

The problem with this is that the same code gets executed for `set`, so if the user does:

```
(dlv) set s.x = 3
```

And x happens to be `uint32` stored in a register then the upper half of the register will be zeroed, which could be wrong because the compiler could be using the upper half for something else.
I don't think this is ever the case in Go, if that's the case maybe just add

```
TODO(aarzilli):` must mix bytes here if called by (*EvalScope).SetVariable
```

to WriteMemory.

Otherwise, if it is possible, a zeroReg flag must be added to WriteMemory and to `(*EvalScope).setValue)` which is set to true when fncall.go calls setValue and to false everywhere else.

pkg/proc/mem.go
```diff
@@ -179,7 +179,14 @@ func (mem *compositeMemory) WriteMemory(addr uint64, data []byte) (int, error) {
 
 			switch piece.Kind {
 			case op.RegPiece:
-				err := mem.regs.ChangeFunc(piece.Val, op.DwarfRegisterFromBytes(pieceMem))
+				newReg := op.DwarfRegisterFromBytes(pieceMem)
+				// Make sure the register is full-width, otherwise ChangeFunc could fail.
+				// Zero-extension is fine, because that's what most load instructions do,
+				// and this is effectively emulating a load to a register, not a store to
+				// memory.
+				newReg.Bytes = nil
```
aarzilli, line 187: I'm worried about registers larger than 8 bytes, with this code. I think something like this would be better:

```
oreg := mem.regs.Reg(piece.Val)
b := pieceMem
if oreg != nil {
    if oreg.Bytes != nil {
        b = make([]byte, len(oreg.Bytes))
    } else {
        b = make([]byte, mem.arch.PtrSize())
    }
    //TODO(aarzilli): mix with original bytes from oreg if called through (*Eval).SetVariable
    copy(b, pieceMem)
}
newReg := op.DwarfRegisterFromBytes(b)
```

aarzilli (commented):
TestDwarfExprComposite is failing because it explicitly looks for a 2 bytes call to SetReg, the test should be updated.

> This probably needs a test? But I'm not sure where to put it.

It should be easy to add a test for this to TestCallFunction in service/test/variables_test.go

pkg/proc/mem.go
mknyszek: I don't believe the compiler does any vectorization at this point, and there aren't any vector primitives, so I didn't really consider that. Do you just want to future-proof it?

I'm happy to make this change, just checking.

mknyszek:
> TestDwarfExprComposite is failing because it explicitly looks for a 2 bytes call to SetReg, the test should be updated.

Yeah, I noticed, but hadn't gotten around to debugging it. Thanks!
> 
> > This probably needs a test? But I'm not sure where to put it.
> 
> It should be easy to add a test for this to TestCallFunction in service/test/variables_test.go

Sounds good. I'll do that.

pkg/proc/mem.go
aarzilli: > Do you just want to future-proof it?

Yes.

mknyszek:
> The problem with this is that the same code gets executed for `set`, so if the user does:
> 
> ```
> (dlv) set s.x = 3
> ```
> 
> And x happens to be `uint32` stored in a register then the upper half of the register will be zeroed, which could be wrong because the compiler could be using the upper half for something else.

Ah, OK. That was my original theory; that this was used by something else that needed to preserve the upper bits.

> I don't think this is ever the case in Go, if that's the case maybe just add
> 
> ```
> TODO(aarzilli):` must mix bytes here if called by (*EvalScope).SetVariable
> ```
> 
> to WriteMemory.
> 
> Otherwise, if it is possible, a zeroReg flag must be added to WriteMemory and to `(*EvalScope).setValue)` which is set to true when fncall.go calls setValue and to false everywhere else.

I think as you say that the compiler never actually does preserve upper bits like this. It would be weird to do this, I think, even in the `SetVariable` case.

Though, it occurs to me that this may also all be context dependent. What if the value needs to be sign-extended in a register? Then both zeroes, and preserving the original bits is incorrect. Perhaps we should loop in someone that knows more about the compiler.

aarzilli:
> Though, it occurs to me that this may also all be context dependent. What if the value needs to be sign-extended in a register?

Mmmh I didn't think of that. I wonder what gdb does.

mknyszek:
According to @cherrymui, the compiler doesn't care what goes in the upper bits. I still think we should zero them.

aarzilli:
> Mmmh I didn't think of that. I wonder what gdb does.

For the record the byte mixing is consistent with what gdb does:

```
type astruct {
    	x int16
	y int32
}

func testfn(s astruct) {
...
}
```

with gdb:

```
(gdb) p s
$2 = {x = 2, y = 3}
(gdb) info registers
rax            0x2                 2
rbx            0x3                 3
...
(gdb) set s.x = 10         // x is rax, y is rbx
(gdb) info reg
rax            0xa                 10
rbx            0x3                 3
(gdb) set s.x = -1        // no sign extension beyond the size of the variable
(gdb) info reg
rax            0xffff              65535
rbx            0x3                 3
...
(gdb) set $rax = 0xdeadbeef00000002     // write garbage to rax
(gdb) info reg
rax            0xdeadbeef00000002  -2401053092612145150
rbx            0x3                 3
...
(gdb) set s.x = -2       // mixing bytes, also no sign extension
(gdb) info reg
rax            0xdeadbeef0000fffe  -2401053092612079618
rbx            0x3                 3
...
```

we can still do things differently for Go but there is no type information at that point so a different approach would be needed.

mknyszek:
> > Mmmh I didn't think of that. I wonder what gdb does.
> 
> For the record the byte mixing is consistent with what gdb does:
> 
> ```
> type astruct {
>     	x int16
> 	y int32
> }
> 
> func testfn(s astruct) {
> ...
> }
> ```
> 
> with gdb:
> 
> ```
> (gdb) p s
> $2 = {x = 2, y = 3}
> (gdb) info registers
> rax            0x2                 2
> rbx            0x3                 3
> ...
> (gdb) set s.x = 10         // x is rax, y is rbx
> (gdb) info reg
> rax            0xa                 10
> rbx            0x3                 3
> (gdb) set s.x = -1        // no sign extension beyond the size of the variable
> (gdb) info reg
> rax            0xffff              65535
> rbx            0x3                 3
> ...
> (gdb) set $rax = 0xdeadbeef00000002     // write garbage to rax
> (gdb) info reg
> rax            0xdeadbeef00000002  -2401053092612145150
> rbx            0x3                 3
> ...
> (gdb) set s.x = -2       // mixing bytes, also no sign extension
> (gdb) info reg
> rax            0xdeadbeef0000fffe  -2401053092612079618
> rbx            0x3                 3
> ...
> ```
> 
> we can still do things differently for Go but there is no type information at that point so a different approach would be needed.

Oof, OK. That's somewhat convincing to me. @heschi what are your thoughts?

heschi:
Clearly there's no perfect answer so I'm fine with whatever. But for me zeroing is still better because it's deterministic and at least it's fairly clear what's going on if it goes wrong. I wouldn't want to be in a situation where you can call `foo.String` unless there happens to be a negative number in some register, for example. I have no idea how likely that is.

aarzilli (approved):
LGTM, I assume that the compiler people are ok with this also.

mknyszek:
@heschi I think despite the risks it still makes sense to merge the values just because gdb does it. In particular (as @cherrymui pointed out to me) it makes the behavior consistent when debugging a C/Go binary.

@aarzilli AFAICT, they're OK with it.

derekparker (approved):
LGTM

PR #2703, opened 2021-09-17
Go 1.18 support

aarzilli:


aarzilli:
cc @heschi I think you wanted to see this
1. tests on arm64 are disabled, tracking the regabi conversion as it's happening seems like too much of a hassle
2. generic support is included, except for setting a breakpoint on a generic function **by function name**, because of https://github.com/golang/go/issues/48032

heschi:
Thank you very much. We're working on an internal generics preview program and having Delve work would be of great help.

pkg/proc/bininfo.go
```diff
@@ -149,21 +149,156 @@ func (err *ErrFunctionNotFound) Error() string {
...
+		for _, pc := range bi.inlinedCallLines[fileLine{filename, lineno}] {
+			pcs = append(pcs, line.PCStmt{PC: pc, Stmt: true})
+		}
 	}
+
+	if len(pcs) == 0 {
+		return nil, &ErrCouldNotFindLine{fileFound, filename, lineno}
+	}
+
+	// 2. assign all occurences of filename:lineno to their containing function
```
derekparker, line 186: I don't see a `1.`, but there's a `2.` and `.3`, etc... in this comment list.

_scripts/test_linux.sh
```diff
@@ -21,7 +21,10 @@ function getgo {
 }
 
 if [ "$version" = "gotip" ]; then
-	exit 0
+	# TODO: remove this
```
derekparker, line 24: This will be removed once the regabi stuff is a little less chaotic I assume?

derekparker (approved):
LGTM—just a few nits and questions.

pkg/proc/bininfo.go
aarzilli: Added.

_scripts/test_linux.sh
aarzilli: Yes, I'm thinking of doing it on the timeline we'd normally support the next version of Go, remove this and fix the test failures.

PR #2704, opened 2021-09-17
[wip] eBPF tracing backend return value parsing

derekparker:


aarzilli:
I was thinking that the current code is setting the tracepoint at function entry, which means that sometimes it will fire multiple times even if the function is only called once, because of the prologue. It's probably worth remembering that it needs to be fixed.

derekparker:
> I was thinking that the current code is setting the tracepoint at function entry, which means that sometimes it will fire multiple times even if the function is only called once, because of the prologue. It's probably worth remembering that it needs to be fixed.

Ack, yeah, it should definitely be moved past the prologue.

derekparker:
@aarzilli fixed everything except for the prologue bit. I'll fix that in a follow up PR.

PR #2710, opened 2021-09-21
go.mod: update google/go-dap to 0.6.0

polinasok:
Besides the dap protocol refresh, this is part 2 of the fix for the issue below. Part 1 is at #2671. They must be committed in the defined order.

Fixes https://github.com/golang/vscode-go/issues/1375


aarzilli (commented):
LGTM (but if I understand correctly this should not be merged yet?)

polinasok:
> LGTM (but if I understand correctly this should not be merged yet?)

Should be merged after  #2671.

PR #2713, opened 2021-09-22
service/dap: allow "instruction" stepping granularity

suzmue:
If the step granularity is "instruction" the program should step forward a single instruction.


service/dap/daptest/client.go
```diff
@@ -303,17 +303,41 @@ func (c *Client) NextRequest(thread int) {
 	c.send(request)
 }
 
+// NextRequest sends a 'next' request.
```
polinasok, line 307: copy/paste comment inconsistency

service/dap/daptest/client.go
```diff
@@ -303,17 +303,41 @@ func (c *Client) NextRequest(thread int) {
...
+
 // StepInRequest sends a 'stepIn' request.
 func (c *Client) StepInRequest(thread int) {
-	request := &dap.NextRequest{Request: *c.newRequest("stepIn")}
+	request := &dap.StepInRequest{Request: *c.newRequest("stepIn")}
+	request.Arguments.ThreadId = thread
+	c.send(request)
+}
+
+// NextRequest sends a 'next' request.
```
polinasok, line 322: same

service/dap/server.go
```diff
@@ -1563,6 +1563,15 @@ func (s *Server) doStepCommand(command string, threadId int, asyncSetupDone chan
 		s.send(stopped)
 		return
 	}
+
+	if granularity == "instruction" {
```
polinasok, line 1568: What does it mean to next vs step vs step out with instruction granularity?
What exactly does this look like in the UI?

aarzilli: The specification of DAP doesn't say but I think that Next + instruction granularity is meant to step over a `CALL` instruction, like an assembly level debugger would do. But we don't have that feature.

suzmue: I think @aarzilli covered it, the details are not explicitly written out in the DAP spec, so I think we are free to make our own decisions, we can definitely look in to adding support for different behavior in each. I can add a TODO to investigate this at a future point.

The UI for all of these disassembly features is very new. Currently, if you are looking at disassembled code, then clicking the regular next, step in, step out buttons will include granularity="instruction".

service/dap/daptest/client.go
suzmue: done.

service/dap/daptest/client.go
suzmue: done.

polinasok:
Could you please paste a screenshot of what the UI looks like with the capability enabled? Thank you.

suzmue:
This change by itself does not enable this in VS Code.  All three of these capabilities need to be true for the UI to change:
```
	response.Body.SupportsDisassembleRequest = true
	response.Body.SupportsSteppingGranularity = true
	response.Body.SupportsInstructionBreakpoints = true
```

If the disassembly view is open, then the toolbar will step by instruction instead of line, but there is no visible change to the UI currently.
Screen Shot 2021-09-28 at 5 10 48 PM

aarzilli:
Please rebase / merge with master.

polinasok (approved):

derekparker:
Re-running CI.

PR #2715, opened 2021-09-22
service/dap: replace math/rand with crypto/rand

suzmue:
Use crypto/rand instead of math/rand as recommended by deepsource checks. Although the security provided by crypto/rand for this test is not necessary.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -2,11 +2,12 @@ package dap
 
 import (
 	"bufio"
+	"crypto/rand"
 	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
+	"math/big"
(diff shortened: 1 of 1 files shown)


aarzilli:
I think we just remove DeepSource instead. This is one of several silly suggestions that DeepSource has been making recently.

suzmue:
> I think we just remove DeepSource instead. This is one of several silly suggestions that DeepSource has been making recently.

Sounds good to me. For recent dap changes, DeepSource hasn't been a very good signal since it flags many existing things in our tests.

aarzilli:
@derekparker is the one who can do it.

derekparker:
I've removed DeepSource from the Delve org, we're going to run staticcheck instead.

PR #2716, opened 2021-09-22
service/dap: set instruction breakpoints

suzmue:
Support setting instruction breakpoints, this is newly supported in VS Code for the disassembly view.


service/dap/server.go
```diff
@@ -1254,6 +1258,81 @@ func (s *Server) onSetFunctionBreakpointsRequest(request *dap.SetFunctionBreakpo
...
+
+func (s *Server) onSetInstructionBreakpointsRequest(request *dap.SetInstructionBreakpointsRequest) {
+	// Get all existing function breakpoints.
+	existingBps := s.getMatchingBreakpoints(instructionBpPrefix)
+	bpAdded := make(map[string]struct{}, len(existingBps))
+	for _, bp := range existingBps {
+		existingBps[bp.Name] = bp
+	}
+
+	breakpoints := make([]dap.Breakpoint, len(request.Arguments.Breakpoints))
```
aarzilli, line 1271: It seems to me that this loop as well as the next exists almost identically also in onSetFunctionBreakpointsRequest and onSetBreakpointsRequest. Wouldn't it be better to refactor it to a common function, like:

```
breakpoints := updateExistingBreakpoints(len(request.Arguments.Brekapoints), func(i int) (name, cond, hitCond string) {
	want := &request.Arguments.Breakpoints[i]
	return fmt.Sprintf("%s PC=%s", instructionBpPrefix, want.InstructionReference), want.Cond, want.HitCond
})
```

where breakpoints is `[]*dap.Breakpoint` and is nil for breakpoints that the third, final loop needs to add (because they are new).

There is one difference that I noticed that this and SetFunctionBreakpointsRequest calls toClientPath but SetBreakpointsRequest doesn't.

service/dap/server.go
```diff
@@ -1254,6 +1258,81 @@ func (s *Server) onSetFunctionBreakpointsRequest(request *dap.SetFunctionBreakpo
...
+		reqString := fmt.Sprintf("%s PC=%s", instructionBpPrefix, want.InstructionReference)
+		var err error
+		got, ok := existingBps[reqString]
+		if !ok {
+			// Skip if the breakpoint does not already exist.
+			// These will be created after deleting existing
+			// breakpoints to avoid conflicts.
+			continue
+		}
+		if _, ok := bpAdded[reqString]; ok {
```
aarzilli, line 1283: Isn't this always false?

suzmue: This could be true if there are multiple instruction breakpoint requests with the same arguments.

service/dap/server.go
suzmue: There is a lot of shared code here. Thanks for the suggestion on how to refactor this. I did a slightly modified refactor to allow the three functions to share all of the logic (besides location computation).

aarzilli (approved):
LGTM

service/dap/server_test.go
```diff
@@ -2296,6 +2296,108 @@ func TestSetBreakpoint(t *testing.T) {
...
+					st = client.ExpectStackTraceResponse(t)
+					if len(st.Body.StackFrames) < 1 {
+						t.Fatalf("\ngot  %#v\nwant len(stackframes) => 1", st)
+					}
+					pc9 := st.Body.StackFrames[0].InstructionPointerReference
+
+					// Clear the source breakpoints.
+					// TODO(suzmue): there is an existing issue that breakpoints with identical locations
+					// from different setBreakpoints, setFunctionBreakpoints, setInstructionBreakpoints
+					// requests will prevent subsequent ones from being set.
```
polinasok, line 2346: What do you mean? Does this mean that if a person sets a breakpoint at a location and then changes it to be a different kind of breakpoint, it won't work?

service/dap/server.go
```diff
@@ -1196,94 +1200,128 @@ func (s *Session) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
+			got.Cond = want.condition
+			got.HitCond = want.hitCondition
+			got.Tracepoint = want.logMessage != ""
+			got.UserData = want.logMessage
 			err = s.debugger.AmendBreakpoint(got)
-			bpAdded[reqString] = struct{}{}
 		}
-
-		updateBreakpointsResponse(breakpoints, i, err, got, clientPath)
+		bpAdded[want.name] = struct{}{}
```
polinasok, line 1280: Why do this use empty struct here as opposed to bool?

service/dap/server.go
```diff
@@ -1196,94 +1200,128 @@ func (s *Session) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
 	// -- exists and not in request => ClearBreakpoint
 	// -- exists and in request => AmendBreakpoint
 	// -- doesn't exist and in request => SetBreakpoint
 
-	// Get all existing breakpoints that match for this source.
-	sourceRequestPrefix := fmt.Sprintf("sourceBp Path=%q ", request.Arguments.Source.Path)
-	existingBps := s.getMatchingBreakpoints(sourceRequestPrefix)
+	// Get all existing breakpoints matching the prefix.
+	existingBps := s.getMatchingBreakpoints(prefix)
 	bpAdded := make(map[string]struct{}, len(existingBps))
```
polinasok, line 1256: What are bpAdded? The name is confusing.

polinasok (changes requested):

service/dap/server_test.go
suzmue: Already if a user sets a function breakpoint at name=myFunction which is defined in main.go:17 and a source breakpoint at file=main.go line=17. And lets say the function breakpoint is set first. Then if the function breakpoint is removed, there will no longer be a breakpoint set at main.go:17, since the source breakpoint will have failed to set because there was already a breakpoint there.

service/dap/server.go
```diff
@@ -1196,94 +1200,128 @@ func (s *Session) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
+	for _, bp := range existingBps {
+		existingBps[bp.Name] = bp
```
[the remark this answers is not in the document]
suzmue: This is actually already done in getMatchingBreakpoints. Removing.

service/dap/server.go
suzmue: maps of empty structs use less memory than bools when implementing a set. If we want to change it for readability, these sets are never that large so probably have minimal space and time savings.

service/dap/server.go
suzmue: Added a comment to explain.

service/dap/server.go
```diff
@@ -1196,94 +1200,128 @@ func (s *Session) onSetBreakpointsRequest(request *dap.SetBreakpointsRequest) {
...
+	Addrs []uint64
+}
+
+// setBreakpoints is a helper function for setting source, function and instruction
+// breakpoints. It takes the prefix of the name for all breakpoints that should be
+// included, the total number of breakpoints, and functions for computing the metadata
+// and the location. The location is computed separately because this may be more
+// expensive to compute and may not always be necessary.
+func (s *Session) setBreakpoints(prefix string, totalBps int, metadataFunc func(i int) *bpMetadata, locFunc func(i int) (*bpLocation, error)) []dap.Breakpoint {
 	// See https://github.com/golang/vscode-go/issues/163 for details.
```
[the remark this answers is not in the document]
suzmue: Removed.

polinasok (approved):

PR #2717, opened 2021-09-23
service/dap: merge arguments and local scopes

suzmue:
Separating the args and the locals scopes in DAP currently provides a worse user experience, since VS Code only auto expands the first non-expensive scope (microsoft/vscode#93230). To handle this, we combine the args and the locals into a single scope, with the arguments appearing first.

Fixes golang/vscode-go#1798

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -28,6 +28,8 @@ import (
 const stopOnEntry bool = true
 const hasChildren bool = true
 const noChildren bool = false
+const localsScope = 1000
+const globalsScope = 1001
 
 var testBackend string
 
@@ -593,21 +595,16 @@ func TestPreSetBreakpoint(t *testing.T) {
 
 		client.ScopesRequest(1000)
 		scopes := client.ExpectScopesResponse(t)
-		if len(scopes.Body.Scopes) > 2 {
-			t.Errorf("\ngot  %#v\nwant len(Scopes)=2 (Arguments & Locals)", scopes)
+		if len(scopes.Body.Scopes) > 1 {
+			t.Errorf("\ngot  %#v\nwant len(Scopes)=1 (Locals)", scopes)
 		}
-		checkScope(t, scopes, 0, "Arguments", 1000)
-		checkScope(t, scopes, 1, "Locals", 1001)
+		checkScope(t, scopes, 0, "Locals", 1000)
 
-		client.VariablesRequest(1000) // Arguments
+		client.VariablesRequest(localsScope) // Locals
 		args := client.ExpectVariablesResponse(t)
-		checkChildren(t, args, "Arguments", 2)
+		checkChildren(t, args, "Locals", 2)
 		checkVarExact(t, args, 0, "y", "y", "0 = 0x0", "uint", noChildren)
 		checkVarExact(t, args, 1, "~r1", "", "0 = 0x0", "uint", noChildren)
-
-		client.VariablesRequest(1001) // Locals
-		locals := client.ExpectVariablesResponse(t)
-		checkChildren(t, locals, "Locals", 0)
 
 		client.ContinueRequest(1)
 		ctResp := client.ExpectContinueResponse(t)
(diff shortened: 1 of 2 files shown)


service/dap/server_test.go
```diff
@@ -593,22 +595,17 @@ func TestPreSetBreakpoint(t *testing.T) {
...
 		client.ScopesRequest(1000)
 		scopes := client.ExpectScopesResponse(t)
-		if len(scopes.Body.Scopes) > 2 {
-			t.Errorf("\ngot  %#v\nwant len(Scopes)=2 (Arguments & Locals)", scopes)
+		if len(scopes.Body.Scopes) > 1 {
+			t.Errorf("\ngot  %#v\nwant len(Scopes)=1 (Locals)", scopes)
 		}
-		checkScope(t, scopes, 0, "Arguments", 1000)
-		checkScope(t, scopes, 1, "Locals", 1001)
+		checkScope(t, scopes, 0, "Locals", 1000)
```
polinasok, line 601: Shouldn't this use your constant?

service/dap/server_test.go
```diff
@@ -1354,21 +1338,14 @@ func TestScopesAndVariablesRequests2(t *testing.T) {
...
-
-					client.VariablesRequest(1000)
-					args := client.ExpectVariablesResponse(t)
-					checkChildren(t, args, "Arguments", 0)
+					checkScope(t, scopes, 0, "Locals", 1000)
 
 					// Locals
 
-					client.VariablesRequest(1001)
+					client.VariablesRequest(1000)
```
polinasok, line 1348: can use constant

service/dap/server_test.go
```diff
@@ -1354,21 +1338,14 @@ func TestScopesAndVariablesRequests2(t *testing.T) {
...
 					}
-					checkScope(t, scopes, 0, "Arguments", 1000)
-					checkScope(t, scopes, 1, "Locals", 1001)
-
-					// Arguments
-
-					client.VariablesRequest(1000)
-					args := client.ExpectVariablesResponse(t)
-					checkChildren(t, args, "Arguments", 0)
+					checkScope(t, scopes, 0, "Locals", 1000)
```
polinasok, line 1344: can use constant

service/dap/server_test.go
```diff
@@ -4694,40 +4664,37 @@ func TestSetVariable(t *testing.T) {
 
 					checkStop(t, client, 1, "main.foobar", startLineno)
 
-					// Args of foobar(baz string, bar FooBar)
-					args := tester.variables(1000)
+					// Local variables
```
polinasok, line 4667: I think the original comment was more useful
You can move next line to before the comment since it is used by all checks now

service/dap/server_test.go
```diff
@@ -4883,23 +4850,19 @@ func TestSetVariableWithCall(t *testing.T) {
...
```
polinasok, line 4853: same as above

polinasok (changes requested):

service/dap/server_test.go
suzmue: Done.

service/dap/server_test.go
suzmue: Done.

service/dap/server_test.go
suzmue: Done

service/dap/server_test.go

service/dap/server_test.go

service/dap/server_test.go
```diff
@@ -2279,7 +2250,7 @@ func TestSetBreakpoint(t *testing.T) {
 					client.ExpectContinueResponse(t)
 					client.ExpectStoppedEvent(t)
 					checkStop(t, client, 1, "main.loop", 8)
-					client.VariablesRequest(1001) // Locals
+					client.VariablesRequest(localsScope) // Locals
```
hyangah, line 2253: I think we don't need comments like  '// Locals' any more

hyangah (commented):
Thanks for delivering the fix so quickly

polinasok (approved):

service/dap/server_test.go
suzmue: done.

PR #2723, opened 2021-09-28
*: clean up staticcheck problems

aarzilli:
Fix problems that can be fixed, ignore the ones that don't make sense


derekparker (approved):
LGTM

PR #2725, opened 2021-09-28
refactor: move from io/ioutil to io and os package

kekeimiku:
ref https://golang.org/doc/go1.16#ioutil.


aarzilli:
We can't do this because it doesn't compile in 1.15, which we still support. At minimum it will have to wait until march but since there's no material benefit it's better to wait a couple of years at least.

PR #2726, opened 2021-09-28
service/dap: auto resume execution when setting breakpoints

suzmue:
A set breakpoints request could come in while the program is running. For a seamless user experience, the server should set the breakpoint and then continue program execution.

Updates golang/vscode-go#1676

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -483,20 +483,6 @@ func (s *Server) handleRequest(request dap.Message) {
 				return
 			}
 			s.onSetBreakpointsRequest(request)
-			// TODO(polina): consider resuming execution here automatically after suppressing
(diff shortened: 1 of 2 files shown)


service/dap/server.go
```diff
@@ -1120,8 +1104,10 @@ func (s *Session) stopDebugSession(killProcess bool) error {
 
 // halt sends a halt request if the debuggee is running.
 // changeStateMu should be held when calling (*Server).halt.
-func (s *Session) halt() (*api.DebuggerState, error) {
-	s.setHaltRequested(true)
+func (s *Session) halt(shouldResume bool) (*api.DebuggerState, error) {
```
aarzilli, line 1107: I think that instead of having a bool argument here, callers should just call setHaltRequested directly when they would pass false to shouldResume.

PR #2728, opened 2021-09-30
service/dap: support disassemble request

suzmue:
This adds support for disassembling a region of code using the instructionPointerReference value returned in the stack trace request.


suzmue:
Here are some screenshots from the new VS Code Disassembly UI:

Screen Shot 2021-09-30 at 3 07 50 PM
Screen Shot 2021-09-30 at 3 08 00 PM

service/dap/server.go
```diff
@@ -2667,10 +2668,170 @@ func (s *Server) onReadMemoryRequest(request *dap.ReadMemoryRequest) {
...
+		s.sendErrorResponse(request.Request, UnableToDisassemble, "Unable to disassemble", err.Error())
+		return
+	}
+
+	start := uint64(addr)
+	maxInstructionLength := s.debugger.Target().BinInfo().Arch.MaxInstructionLength()
+	offset := request.Arguments.InstructionOffset * maxInstructionLength
+	// Adjust the offset to include instructions before the requested address.
+	if offset < 0 {
+		start = uint64(addr + int64(offset))
```
aarzilli, line 2688: Unfortunately wn can not do this. For many (all?) ISA with variable instruction length the instruction encoding is not [self-synchronizing](https://en.wikipedia.org/wiki/Self-synchronizing_code), which means that this will usually produce a non-sensical disassembly.
There's also the problem that both start and end, as they are calculated, could cross a page bounary into unallocated memory causing the entire disassembly to fail even if there are enough instructions to satisfy the request.

The way the spec is pretty hard to satisfy here, the fully general interpretation would require backward disassembly, which is impossible to do. 
The easiest thing to do would be to call `Disassemble(-1, addr, 0)` which will disassemble the function around `addr` and then pad with undefined instructions in both directions. I don't know if this works with the requests sent by VSCode.

service/dap/server.go
```diff
@@ -2667,10 +2668,170 @@ func (s *Server) onReadMemoryRequest(request *dap.ReadMemoryRequest) {
...
+
+	// Disassemble the instructions
+	procInstructions, err := s.debugger.Disassemble(-1, start, end)
+	if err != nil {
+		s.sendErrorResponse(request.Request, UnableToDisassemble, "Unable to disassemble", err.Error())
+		return
+	}
+
+	// Find the section of instructions that were requested.
+	startIdx, endIdx, err := findRange(procInstructions, addr, (request.Arguments.InstructionOffset), (request.Arguments.InstructionCount))
```
aarzilli, line 2709: why the superfluous parenthesis?

service/dap/server.go
```diff
@@ -2667,10 +2668,170 @@ func (s *Server) onReadMemoryRequest(request *dap.ReadMemoryRequest) {
...
+		if instruction.Loc.File != lastFile || instruction.Loc.Line != lastLine {
+			instructions[i].Location = dap.Source{Path: instruction.Loc.File}
+			instructions[i].Line = instruction.Loc.Line
+			lastFile, lastLine = instruction.Loc.File, instruction.Loc.Line
+		}
+	}
+	response := &dap.DisassembleResponse{
+		Response: *newResponse(request.Request),
+		Body: dap.DisassembleResponseBody{
+			Instructions: instructions,
```
aarzilli, line 2735: The spec says:

> An adapter must return exactly this number of instructions - any
> unavailable instructions should be replaced with an implementation-defined
> 'invalid instruction' value.

I think this can return fewer.

aarzilli (approved):
LGTM

service/dap/server_test.go
```diff
@@ -5744,3 +5743,311 @@ func TestBadlyFormattedMessageToServer(t *testing.T) {
...
+				// Stop at line 17
+				execute: func() {
+					checkStop(t, client, 1, "main.main", 17)
+
+					client.StackTraceRequest(1, 0, 1)
+					st := client.ExpectStackTraceResponse(t)
+					if len(st.Body.StackFrames) < 1 {
+						t.Fatalf("\ngot  %#v\nwant len(stackframes) => 1", st)
+					}
+					// Request the single instruction that the program is stopped at.
```
polinasok, line 5766: Should we test for 0 instructions as well?

service/dap/server.go
polinasok: What do you mean? The specification specifically says "exactly this number of instructions".

service/dap/server.go
suzmue: These were left over from a previous refactor. Removed.

service/dap/server.go
suzmue: My initial implementation returned < request.Arguments.InstructionCount if we reached the end of the region that contains instructions for the program. It has been updated to always return request.Arguments.InstructionCount.

service/dap/server.go
```diff
@@ -2668,9 +2668,29 @@ func (s *Server) onReadMemoryRequest(request *dap.ReadMemoryRequest) {
...
+var invalidInstruction = dap.DisassembledInstruction{
+	Address:     "out of bounds",
+	Instruction: "invalid instruction",
+}
+
 // onDisassembleRequest handles 'disassemble' requests.
 // Capability 'supportsDisassembleRequest' is set in 'initialize' response.
 func (s *Server) onDisassembleRequest(request *dap.DisassembleRequest) {
+	// If the requested memory address is an invalid location, return all invalid instructions.
+	if request.Arguments.MemoryReference == invalidInstruction.Address {
```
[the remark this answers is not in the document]
suzmue: This is a made up invalid instruction. The disassembly view is implemented as an infinite scroll in VS Code, and thus may continue requesting invalid instructions as we scroll. Maybe we could provide a decreasing address so they can be used to get back in bounds. I'll add a TODO to do this, but I don't think its urgent.

service/dap/server_test.go
suzmue: done.

aarzilli:
Needs a rebase.

PR #2729, opened 2021-10-01
service/dap: refactor the server into two layers

polinasok:
This change splits `dap.Server`
- `Server` - starts run goroutine, accept a client, starts `Session` for client connection and shuts down on stop signal
- `Session` - handles the traffic for an open generalized `io.ReadWriteCloser` connection and triggers server's `Stop` on user disconnect or connection error

This will allow `rpccommon.Server` to use `dap.Session.serveDAPCodec` as a `rpccommon.ServerImpl.ServeJSONCodec` alternative.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -92,15 +92,20 @@ import (
 // wrap-up and exit.
 type Server struct {
 	// config is all the information necessary to start the debugger and server.
-	config *service.Config
+	config *Config
 	// listener is used to accept the client connection.
 	listener net.Listener
-	// stopTriggered is closed when the server is Stop()-ed.
-	stopTriggered chan struct{}
-	// reader is used to read requests from the connection.
-	reader *bufio.Reader
-	// log is used for structured logging.
-	log *logrus.Entry
+	// session is the debug session that comes with an client connection.
+	session   *Session
+	sessionMu sync.Mutex
+}
+
+// Session is an abstraction for serving and shutting down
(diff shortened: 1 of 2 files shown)


service/dap/server.go
```diff
@@ -92,15 +92,20 @@ import (
...
 	// listener is used to accept the client connection.
 	listener net.Listener
-	// stopTriggered is closed when the server is Stop()-ed.
-	stopTriggered chan struct{}
-	// reader is used to read requests from the connection.
-	reader *bufio.Reader
-	// log is used for structured logging.
-	log *logrus.Entry
+	// session is the debug session that comes with an client connection.
+	session   *Session
```
aarzilli, line 99: A bit weird that Server has a config but also a session that also has a config. Is it just to save keystrokes? Could the session object just be an embed?

polinasok: I was kind of modeling after rpc1/2 that also take configs. 
What exactly are you suggesting? Drop "session", so Session is embedded into Server and config field is no longer needed? In this design session is only created set if there is an open connection.

polinasok:
I am looking into the test failures. Everything was passing for me locally :(

service/dap/server.go
aarzilli: Ok. I see now.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

Issue #2323, opened 2021-01-24
dlv dap: --accept-multiclient and --continue support


polinasok:
With headless dlv, the `--accept-multiclient` feature allows for multiple clients to connect to a server debugging a single process. It is used in 3 scenarios:
1. To launch/attach to a remote process and use `--continue` (triggered by a quick behind-the-scenes client)
2. To connect, disconnect and connect again to a debugging server, usually in a remote container, for repeated debugging sessions or accidental dropped connections
3. To connect multiple clients concurrently to debug the same process. 
  This comes into play when a user hits a limitation in vscode UI. In that case, they might also connect with dlv-connect or even GoLand to take advantage of their additional features (e.g. to use `config` command, print variables in flattened text form, etc). We recommend this method when users ask to expose dlv cli features via the Debug Console (which is non-trivial due to the split architecture). And we have been considering streamlining the double client option via the extension to help with this use case (as opposed to reimplementing the dlv cli on the adapter/debugger side). This mode comes with limitations and risks due to lack of client UI synchronization - see #2322.

In all cases, delve starts debugging a process as part of server start-up outside of any client. And with this option, the server doesn't automatically exit when any and all client connections are gone, but can continue running, with the target suspended or running, waiting for another client to connect. 

Unfortunately, this model was not considered by DAP designers (as recently confirmed with @weinand offline). First of all, DAP clients are supposed to always send a `launch`/`attach` request to start the debug session, which doesn't make sense if the debugger already launched/attached to a process. Secondly, there is no DAP request to add a breakpoint to the list of already set breakpoints. There is only a [`setBreakpoint`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_SetBreakpoints) request, which clears all previous breakpoints and sets all specified breakpoints. That means that concurrent and sequential clients will clear each other's breakpoints. Thirdly, it will be tricky if not impossible to keep the clients' UIs in sync with what is actually happening, which might lead to untimely requests and issues like in #2322. We might be able to accomplish some UI synchronization with optional DAP events like [`continued` event](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Continued), but with DAP not actively supporting this use case, there will likely be situations where the right events are not available.

[...]

PR #2731, opened 2021-10-04
service/dap: support accept-multiclient shutdown in remote attach mode

polinasok:
The dap server doesn't support multiple clients, but the inner connect layer (see #2729) is planned to be used by rpccommon server, which does.

#updates #2323


aarzilli:
> TestAttachRequest error doesn't look related, but what can be causing it?

We don't clean up breakpoints before detaching, it probably runs a bit between detach and the point where it gets killed. `nativeProcess.detach` should pass the SIGINT through PTRACE_DETACH instead of killing the process separately.

PR #2732, opened 2021-10-04
service/dap: log parsed and applied launch configs

polinasok:
Launch/attach requests contain way more fields than the dap server actually cares about, including sensitive information that users like to scrub out of the logs (e.g. golang/vscode-go#1701). Logging the fields that we use will help both users and maintainers make sense of what defaults are used, how relative paths are expanded, and so on.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -819,6 +819,7 @@ func cleanExeName(name string) string {
 }
 
 func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
+	var err error
 	if s.debugger != nil {
 		s.sendShowUserErrorResponse(request.Request, FailedToLaunch,
 			"Failed to launch", "debugger already started - use remote attach to connect to a server with an active debug session")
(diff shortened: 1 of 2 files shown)


polinasok:
@suzmue @hyangah

service/dap/server.go
```diff
@@ -831,83 +832,79 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 			FailedToLaunch, "Failed to launch", fmt.Sprintf("invalid debug configuration - %v", err))
 		return
 	}
+	s.config.log.Debug("parsed launch config: ", prettyPrint(args))
```
suzmue, line 835: Is there any reason to do this before setting args.Mode to debug below?

service/dap/server.go
```diff
@@ -831,83 +832,79 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
 		s.sendShowUserErrorResponse(request.Request, FailedToLaunch, "Failed to launch",
 			"The program attribute is missing in debug configuration.")
 		return
 	}
-
-	if backend := args.Backend; backend != "" {
-		s.config.Debugger.Backend = backend
-	} else {
-		s.config.Debugger.Backend = "default"
+	args.Program, err = filepath.Abs(args.Program)
```
suzmue, line 851: Could this change the meaning of the launch configuration?

service/dap/server.go
polinasok: My goal was to log what the user supplied and how we interpreted it. It's not exact because a couple of defaults are populated before the parsing, but this on is not one of them (and I cannot remember why.)

service/dap/server.go
polinasok: I don't think so because that's what the underlying backend is supposed to do and that's what we promise to the user - interpreting the program relative to the server's working directory. I just made it explicit, so we can log it and it is clear from the code what is going on.

polinasok: Also, that's what the tests are for. Do you think we are missing something key to verify such things?

polinasok: Hmm, now there is cwd that we could also expand before passing it into the debugger to be expanded there. What do you think?

aarzilli:
> @derekparker @aarzilli Something changed with how interface type is printed, which is not only triggering pretty-printing tests, but also tests where the type is used as part of expressions.

Correct.

suzmue:
LGTM

PR #2734, opened 2021-10-06
service/dap: rename delveCwd to dlvCwd and clarify comments

polinasok:
Let's use `dlvCwd` to match the tool's name as well as other launch attributes like `dlvFlags` or `dlvLoadConfig`.

### DIFF
--- a/service/dap/types.go
+++ b/service/dap/types.go
@@ -52,7 +52,7 @@ var (
 	}
 )
 
-// LaunchConfig is the collection of launch request attributes recognized by delve DAP implementation.
+// LaunchConfig is the collection of launch request attributes recognized by DAP implementation.
 type LaunchConfig struct {
 	// Acceptable values are:
 	//   "debug": compiles your program with optimizations disabled, starts and attaches to it.
@@ -68,7 +68,7 @@ type LaunchConfig struct {
 	// when in `debug` or `test` mode, and to the pre-built binary file
 	// to debug in `exec` mode.
 	// If it is not an absolute path, it will be interpreted as a path
-	// relative to Delve's working directory.
+	// relative to dlv's working directory.
 	// Required when mode is `debug`, `test`, `exec`, and `core`.
 	Program string `json:"program,omitempty"`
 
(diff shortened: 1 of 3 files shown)


service/dap/types.go
```diff
@@ -117,19 +117,18 @@ type LaunchConfig struct {
...
+	// When dlv needs to build the program (in debug/test modes),
+	// it will run the go command from this directory as well.
 	//
-	// If a relative path is provided as DelveCwd, it will be
-	// interpreted as a path relative to Delve's current working
+	// If a relative path is provided as DlvCwd, it will be
+	// interpreted as a path relative to dlv's current working
 	// directory.
-	DelveCwd string `json:"delveCwd,omitempty"`
+	DlvCwd string `json:"dlvCwd,omitempty"`
```
hyangah, line 131: Another option is DelveCwd (DelveCWD?) as a Go struct field name, and `dlvCwd` as json field name. Up to you.

hyangah (approved):
LGTM

I think `dlvCwd` (or `dlvCWD`?) as the json attribute makes sense. Not sure about `DlvCwd`.

The reason that I preferred 'Delve' in the documentation is 1) Delve is the official name; even the command `dlv` official doc calls itself Delve, 2) service/dap package itself is a library that can be used to implement any command line tool other than `dlv`. But if you prefer dlv to Delve, I am fine with the change.

While we are discussing naming... what do you think about considering CWD, DAP instead of Cwd, Dap in go code?
(https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

polinasok:
> The reason that I preferred 'Delve' in the documentation is 1) Delve is the official name; even the command `dlv` official doc calls itself Delve, 2) service/dap package itself is a library that can be used to implement any command line tool other than `dlv`. But if you prefer dlv to Delve, I am fine with the change.

I am fine with Delve. I just noticed we had dlv, delve and Delve in the same file, so when it came up in my search/replace, I decided to keep it for consistency as dlv everywhere. I can change it to Delve or delve - I have seen both in the code comments in this repo (although delve seems more common). 

> While we are discussing naming... what do you think about considering CWD, DAP instead of Cwd, Dap in go code? (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

CWD somehow looks odd to me. Maybe I am just not used to it. os.Getwd, not os.GetWD. We could use dlvcwd or dlvwd, but we can't use cwd because we need to capitalize to export, so I don't know. Dap really should be DAP in both code and comments. I don't know why we ended up with Dap. But honestly, I can pretty much live with whatever as long as it is consistent.

hyangah:
> I am fine with Delve. I just noticed we had dlv, delve and Delve in the same file, so when it came up in my search/replace, I decided to keep it for consistency as dlv everywhere. I can change it to Delve or delve - I have seen both in the code comments in this repo (although delve seems more common).

Let's go with Delve for documentation for consistency with other documentation.

> 
> > While we are discussing naming... what do you think about considering CWD, DAP instead of Cwd, Dap in go code? (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
> 
> CWD somehow looks odd to me. Maybe I am just not used to it. os.Getwd, not os.GetWD. We could use dlvcwd or dlvwd, but we can't use cwd because we need to capitalize to export, so I don't know. Dap really should be DAP in both code and comments. I don't know why we ended up with Dap. But honestly, I can pretty much live with whatever as long as it is consistent.

OK, then, let's just capitalize DAP.

Many of names in os reflect preexisting posix names (`getwd`, `getpagesize`) so no camel casing in word boundary. And they were added before style guide was established. 

Thanks.

PR #2738, opened 2021-10-06
_fixtures/loopprog: print less often

polinasok:


polinasok:
So I just ran into a test that uses loopprog, which reminded me why we use it. It's because you can easily set a breakpoint that will be hit (as opposed to having the program stuck on Listen and requiring a client to unblock). And there is a simple variables at that breakpoint that can be tested.

PR #2742, opened 2021-10-11
service/dap: add registers configuration for variables response

suzmue:
Like with showGlobalVariables, provide the option to see the register contents.

Screen Shot 2021-10-11 at 7 07 38 PM

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -89,7 +89,8 @@ var (
 
 	allowNonTerminalInteractive bool
 
-	conf *config.Config
+	conf        *config.Config
+	loadConfErr error
 )
 
 const dlvCommandLongDesc = `Delve is a source level debugger for Go programs.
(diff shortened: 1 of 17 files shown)


aarzilli:
> > LGTM. Are all those options documented anywhere that a user would find?
> 
> The launch config options are documented in [package.json](https://github.com/golang/vscode-go/blob/1f4e83b5ab749e1c32925df892596a875e303e82/package.json#L964) in vscode-go, which then becomes this [doc table](https://github.com/golang/vscode-go/blob/master/docs/debugging.md#configuration). Those doc strings show up on hover in vscode when users edit the config. And after Hana's refactoring we now also have them in [godoc](https://pkg.go.dev/github.com/go-delve/delve/service/dap#LaunchConfig). It might be a good idea to add some explicit details and/or pointers to the documentation in this repo as well.

Given that other projects are starting to use this besides vscode-go, should that be file be referenced from delve's repository?

derekparker:
> The remaining test failure is: TestScopePrefix github.com/go-delve/delve/pkg/terminal on Windows, which I have been seeing on a lot of issues and is unrelated (this change doesn't touch any code that pkg/terminal uses).

Ack, thanks for the quick fixup! Please just rebase and this will be good to go.

PR #2747, opened 2021-10-14
service/dap: allow expression evaluation in log messages

suzmue:
From the DAP spec:   
```
If this attribute exists and is non-empty, the backend must not 'break' (stop) but log the message instead.
Expressions within {} are interpolated.
```

This change parses the log messages and stores the parsed values as a format string and list of expressions to evaluate and get the string value of.

Updates golang/vscode-go#123


derekparker (approved):
LGTM

PR #2748, opened 2021-10-15
service/dap: fix goroutine id selection for hardcoded breakpoints

suzmue:
Determining the stopped goroutine id on a breakpoint required
checking for breakpoints since some may be tracepoints. However,
there may be goroutines stopped on hardcoded breakpoints with
no breakpoint. We fix this by checking for runtime.breakpoint or
StopReason=proc.StopHardcodedBreakpoint.


derekparker (approved):
LGTM

Issue #2328, opened 2021-01-27
dlv dap: remote connect (aka remote attach)


polinasok:
**Background**
The following [debug configurations](https://github.com/golang/vscode-go/blob/bc8b63e62e6dcbfbd60e6e9f71eec7aff9ae367a/src/goDebugConfiguration.ts#L31-L114) are offered to users in vscode-go.
   * (1) Launch file/package/test: `request='launch' mode='debug'/'test'/'exec' program=...` (Local Debugging)
   * (2) Attach to process: `request='attach' mode='local' processId=...` (Local Debugging)
   * (3) Connect to server: `request='attach' mode='remote' host=... port=...` (Remote Debugging)

With traditional `dlv`, (1) and (2) configuration attributes are translated into cli arguments:
  * (1) `dlv debug/exec/test <program> --headless`
  * (2) `dlv attach <pid> --headless`
  
With `dlv dap`, the server command takes no args, but waits for the DAP client to tell it what to debug via
 * (1) [launch](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Launch) request arguments
 * (2) [attach](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Attach) request arguments

This difference is hidden under the hood and is irrelevant to the local debugging users - options (1) and (2) as the debug configuration is the only thing a user needs to provide.

In case of Remote Debugging, option (3), the user first starts the debug server on their own at host/port, locally or remotely, and then the client connects to it for the debug session as per the host/port in the debug configuration. Because the user deals with the server command directly, the differences in how program/pid are supplied come into light. In addition, remote debugging comes with the use case of connecting to the server with multiple client (tracked under #2323).

**Remote Option 1** (Done)
If we follow the traditional DAP initialization sequence, the user would need to launch the server outside of the client, then use the client to start a debug session and tell the server what to launch/attach to (options (1) and (2) above). So essentially this means introducing new client-side configurations to express (3)+(1) and (3)+(2). This could look like:
   - (3-1) Connect to server and launch: `host=... port=... request='launch' mode='debug'/'test'/'exec' program=...`
   - (3-2) Connect to server and attach:  `host=... port=... request='attach' mode='local' processId=...`

Host and port will be used to connect to dlv-dap. Request mode, etc will drive the launch/attach request sent to it just like they do in (1) and (2). Any additional configuration processing would need to be done on the client side. **Dlv-dap can already support this bearing the considerations below.**

**Remote Option 2** (TODO)
[...]

PR #2749, opened 2021-10-15
service/dap: improve shutdown logic and test coverage

polinasok:
This high-level goal is to push more shutdown logic into `Session` and not rely `Server.Stop()` because `dap`'s version of `Stop()` is different from `rpccommon`. That means triggering the same shutdown steps in multiple places, so additional safeguards are added to make some of the duplicate steps no-ops to avoid errors, extra logging, etc.

This includes the following changes:
1. close client conn when request loop in serveDAPCodec exits (to match `rpccommon`)
2. exit request loop after processing a disconnect request (instead of relying on next read to fail because conn got closed by triggered `Stop()`, which is skipped in case of `accept-multiclient` and is not closed in `Stop()` in `rpccommon`)
3. reset `debugger` to nil to avoid shutting it down more than once and causing duplicate logging (in case `stopDebugSession` is called from `onDisconnectRequest` and `Server.Stop` or `Session.Close` as things shut down)
4. reset `binaryToRemove` to "" upon removal to avoid duplicate error (in case `Close()` is called more than once outside of `Stop()`, which technically is not the case right now)
4. expand testing for all possible server shutdown triggers
5. testing for `Session`-only shutdown as it will be integrated into `rpccommon` without `dap.Server` wrapper

Fixes https://github.com/golang/vscode-go/issues/1830
Updates #2328

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -16,9 +16,8 @@ Tests skipped by each supported backend:
 	* 4 not implemented
 * linux/386/pie skipped = 1
 	* 1 broken
-* linux/arm64 skipped = 4
+* linux/arm64 skipped = 1
 	* 1 broken - cgo stacktraces
-	* 3 not implemented
 * pie skipped = 2
 	* 2 upstream issue - https://github.com/golang/go/issues/29322
 * windows skipped = 2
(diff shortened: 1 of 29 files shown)


polinasok:
@suzmue @hyangah

PR #2750, opened 2021-10-15
service/dap: add config expressions to evaluate requests

suzmue:
We want to provide more flexibility for users to make changes to their configurations while the debug session is running. This could be accomplished by creating a custom request, but that were require a new UI as well, and every client of dlv dap to provide its own UI for this. By using the evaluate context, users can use the already existing debug console to change their configurations.

This change includes a refactor of the terminal code in order to share the code with the dap package.

This change provides a very similar to UI as the terminal package, but there are different configuration options that are DAP specific. We plan to use this same mechanism to expose a few other commands including "sources" to help users debug an ineffective substitutePath configuration.


polinasok (approved):

polinasok:
LGTM

derekparker (approved):
LGTM

PR #2759, opened 2021-10-26
service/dap: support goroutine filters in dap

suzmue:
This change refactors the goroutine filter parsing from the terminal package into service/api so it can be shared by the terminal client and the dap server. This supports setting goroutine filters using the launch configuration.


aarzilli:
> > Is this useful given that it can't be used interactively? Even with the `dlv config` hack the goroutine list won't reload until the execution is resumed, right?
> 
> It is not particularly useful if it can't be used interactively. When we allow it to be updated interactively, we can send an "invalidated" event, which will trigger a request from the dap client to get all the information again.

I think we should wait for that.

polinasok:
> > Is this useful given that it can't be used interactively? Even with the `dlv config` hack the goroutine list won't reload until the execution is resumed, right?
> 
> It is not particularly useful if it can't be used interactively. When we allow it to be updated interactively, we can send an "invalidated" event, which will trigger a request from the dap client to get all the information again.



> > > Is this useful given that it can't be used interactively? Even with the `dlv config` hack the goroutine list won't reload until the execution is resumed, right?
> > 
> > 
> > It is not particularly useful if it can't be used interactively. When we allow it to be updated interactively, we can send an "invalidated" event, which will trigger a request from the dap client to get all the information again.
> 
> I think we should wait for that.

@suzmue Is `dlv config` supposed to be that interactive mechanism that triggers the invalidated event? Now that the config change is merged, can this be done in this PR?

service/dap/server.go
```diff
@@ -1601,11 +1607,19 @@ func (s *Session) onThreadsRequest(request *dap.ThreadsRequest) {
...
-		if err == nil && s.args.hideSystemGoroutines {
-			gs = s.debugger.FilterGoroutines(gs, []api.ListGoroutinesFilter{{
-				Kind:    api.GoroutineUser,
-				Negated: false,
-			}})
+		if err == nil {
+			// Parse the goroutine arguments.
+			filters, _, _, _, _, _, parseErr := api.ParseGoroutineArgs(s.args.goroutineFilters)
+			if parseErr != nil {
+				s.logToConsole(parseErr.Error())
```
polinasok, line 1614: Should this log to stderr instead of console since it's an error, not an informational FYI?

service/dap/types.go
```diff
@@ -154,9 +154,13 @@ type LaunchAttachCommonConfig struct {
 	ShowRegisters bool `json:"showRegisters,omitempty"`
 
 	// Boolean value to indicate whether system goroutines
-	// should be should be hidden from the call stack view.
+	// should be hidden from the call stack view.
 	HideSystemGoroutines bool `json:"hideSystemGoroutines,omitempty"`
 
+	// String value to indicate which system goroutines
```
polinasok, line 160: You might want to link to FILTERING for documentation
https://github.com/go-delve/delve/blob/master/Documentation/cli/README.md#goroutines

service/dap/server_test.go
```diff
@@ -805,6 +805,110 @@ func checkStackFramesExact(t *testing.T, got *dap.StackTraceResponse,
...
+			wantLen: 1,
+		},
+		{
+			name:    "no match to user and not user",
+			filter:  "-with user -without user",
+			want:    []string{"Dummy"},
+			wantLen: 1,
+		},
+	}
+	for _, tc := range tt {
```
aarzilli, line 868: move the for loop inside runDebugSessionWithBPs and change the filter through the config command.

aarzilli (approved):

service/dap/config.go
```diff
@@ -27,11 +27,6 @@ func configureSet(sargs *launchAttachArgs, args string) (bool, string, error) {
 		return false, "", fmt.Errorf("%q is not a configuration parameter", cfgname)
 	}
 
-	// If there were no arguments provided, just list the value.
```
polinasok, line 30: This is an unfortunate change. Adding -list is much less intuitive and more typing for the user.

service/dap/server_test.go
```diff
@@ -3970,7 +3973,7 @@ Type help followed by a command for full documentation.
 					checkEval(t, got, formatConfig(50, false, false, "", false, [][2]string{}), noChildren)
 
 					// Read and modify showGlobalVariables.
-					client.EvaluateRequest("dlv config showGlobalVariables", 1000, "repl")
+					client.EvaluateRequest("dlv config -list showGlobalVariables", 1000, "repl")
```
polinasok, line 3976: This is an unfortunate change. Adding -list is much less intuitive and more typing for the user.  But adding special empty filter  to separate look-up and reset seems complicated too :(

polinasok (approved):

service/dap/server_test.go
suzmue: Yeah I agree that its not ideal :( Though it lines up better with the dlv cli now.

derekparker (approved):
LGTM

Issue #2762, opened 2021-10-28
service/dap: TestNextParked flaky


aarzilli:
https://delve.teamcity.com/viewLog.html?buildId=27717

polinasok:
@aarzilli In the failure log the process runs to completion, but exits with status -1 instead of 0. Would could be causing flakiness in the termination status of [parallel_next.go](https://github.com/go-delve/delve/blob/master/_fixtures/parallel_next.go)? Is that expected under certain conditions or a bug somewhere? Would skipping the status check to unflake the test just paper over a problem that should be investigated?

aarzilli:
My guess is https://github.com/go-delve/delve/blob/9013a121d8bc6359018bccfaa1f256a8870d01c7/pkg/proc/native/proc_linux.go#L442, we get some stop reason, try to resume the main thread and find that it's gone. I don't know why that happens but I think it does sometimes, In those cases we don't have the exit status.
It could also be that the target process got a SIGHUP, but I don't see why that would be the case.

polinasok:
Hmm, it thinks it hasn't exited?
https://github.com/go-delve/delve/blob/9013a121d8bc6359018bccfaa1f256a8870d01c7/vendor/golang.org/x/sys/unix/syscall_linux.go#L320-L325
We get this status when we get debugger state after the command that returned exited status. We have a TODO to refactor that code. Would that help?

aarzilli:
>  We get this status when we get debugger state after the command that returned exited status. We have a TODO to refactor that code. Would that help?

I think this all happens inside ContinueOnce.

PR #2765, opened 2021-10-28
service/dap: expect TestNextParked to finish terminated

suzmue:
Updates TestNextParked to always run until program terminates.

Updates #2762

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -4280,8 +4280,7 @@ func TestNextParked(t *testing.T) {
 			fixture.Source, []int{15},
 			[]onBreakpoint{{ // Stop at line 15
 				execute: func() {
-					if goroutineId := testStepParkedHelper(t, client, fixture); goroutineId >= 0 {
-
+					if goroutineId, terminated := testStepParkedHelper(t, client, fixture); !terminated {
 						client.NextRequest(goroutineId)
 						client.ExpectNextResponse(t)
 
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

service/dap/server_test.go
```diff
@@ -4291,64 +4290,66 @@ func TestNextParked(t *testing.T) {
 						}
 					}
 				},
-				disconnect: false,
+				terminated: true,
```
polinasok, line 4293: Will this work? The harness only looks at terminated if disconnect is true.

suzmue: will add disconnect: true just in case.

service/dap/server_test.go
```diff
@@ -4291,64 +4290,67 @@ func TestNextParked(t *testing.T) {
...
+				goroutineId = g.Id
 				break
 			}
 		}
+		if goroutineId >= 0 {
+			break
+		}
+	}
+	if goroutineId < 0 {
+		goroutineId = se.Body.ThreadId
```
polinasok, line 4346: It's a bit odd that instead of searching for a parked goroutine like before, you now give up and just test switching to the same goroutine you are already one.

polinasok (changes requested):
The description says "Updates TestNextParked to always run until program terminates", but the helper function changes took out the for-loop that could continue to completion. Moreover, your original test was already written in a way that expected the program to run to termination on its own (as opposed to getting stuck in an infinite loop or permanent waiting state), so a forceful disconnect+kill (`disconnect: true`) was not needed at the tested breakpoint. The failure log linked in the bug showed that the process continued until a terminated event, which was followed by a disconnect sequence, expecting a success exit status. But the status code was unexpectedly -1 instead of 0.

So to me it looks like the problem is not lack of termination, but that `proc.ErrProcessExited` returned after termination had an unexpected status. This reminds me of the earlier change where you had to propagate the status correctly. Did it not work? Or did it not land? Could there be another bug in the Debugger that causes wrong status codes? (And yes, we do need to follow up on the TODO to move the process exit printing to earlier in the code right after it happens.)

polinasok:
Just wanted to add that if you do rewrite the helper to always terminate, so disconnect+terminated always applies, you will be hitting the same kind of check for 0 exit status that triggered the failure due to -1 exist status. I think the only way to avoid the exit status check is to make sure that the process is always killed with disconnect. That would fix the flakiness, but paper over whatever is causing an unexpected exit status.

polinasok:
The PR proposes an alternative way to hide the -1 exit status issue: #2784. Another option would be to set a breakpoint at the end of main, so you never terminate and can use disconnect. I am not sure how to reproduce and debug -1 status.

suzmue:
Closing this in favor of #2784

Issue #2763, opened 2021-10-28
service/dap: race condition in TestNoDebug_GoodExitStatus


aarzilli:
https://delve.teamcity.com/viewLog.html?buildId=27603

cc @polinasok

polinasok:
Hit this again:
```
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":2,"success":true,"command":"launch"}
4
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"output","body":{"category":"console","output":"Process 20633 has exited with status 0\n","source":{}}}
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"terminated","body":{}}
2021-12-30T20:37:08Z debug layer=dap [<- from client]{"seq":3,"type":"request","command":"disconnect","arguments":{"terminateDebuggee":true}}
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"output","body":{"category":"console","output":"Terminating process 20633\n","source":{}}}
2021-12-30T20:37:08Z debug layer=dap noDebug process killed
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":3,"success":true,"command":"disconnect"}
2021-12-30T20:37:08Z debug layer=dap [-> to client]{"seq":0,"type":"event","event":"terminated","body":{}}
2021-12-30T20:37:08Z debug layer=dap DAP server stopping...
2021-12-30T20:37:08Z debug layer=dap DAP server stopped
    server_test.go:5316: got &dap.OutputEvent{Event:dap.Event{ProtocolMessage:dap.ProtocolMessage{Seq:0, Type:"event"}, Event:"output"}, Body:dap.OutputEventBody{Category:"console", Output:"Terminating process 20633\n", Group:"", VariablesReference:0, Source:dap.Source{Name:"", Path:"", SourceReference:0, PresentationHint:"", Origin:"", Sources:[]dap.Source(nil), AdapterData:interface {}(nil), Checksums:[]dap.Checksum(nil)}, Line:0, Column:0, Data:interface {}(nil)}}, want *dap.DisconnectResponse
--- FAIL: TestNoDebug_GoodExitStatus (0.29s)
```

PR #2766, opened 2021-10-28
service/dap: fix race in no debug test

suzmue:
Close the exited channel first before sending status events.  This way when we disconnect after all the status messages, we know for sure the exit signal has been propagated and shutdown won't try to kill the process.

Updates go-delve/delve#2763


service/dap/server_test.go
```diff
@@ -5026,7 +5026,16 @@ func runNoDebugSession(t *testing.T, client *daptest.Client, launchRequest func(
 	client.ExpectOutputEventProcessExited(t, exitStatus)
 	client.ExpectTerminatedEvent(t)
 	client.DisconnectRequestWithKillOption(true)
-	client.ExpectDisconnectResponse(t)
+	m := client.ExpectMessage(t)
+	switch m.(type) {
+	case *dap.DisconnectResponse:
+	case *dap.OutputEvent:
+		// There may be a race between closing the exited channel, and trying
+		// to kill the process, so we may get a "Terminating process" output
```
polinasok, line 5034: Kill-after-undetected-exit doesn't hurt, does it? So the only problem is inconsistent messages?
What about closing the channel before issuing process existed message and the terminated event? Then this case won't flake. And the case where the user clicks disconnect after process exists, but before everything is reflected is very unlikely.

suzmue: Done.

polinasok:
I am not sure how I missed when this PR got updated and why the approval fell through the cracks. Approving now. 
I think the new version does the right thing to fix the race and should fix the flakiness that appears to be more common now.

- First close the exited channel
- Then send a terminated event
- This will trigger a disconnect request from the client
- The handler will now find a reliably closed exited channel (so no more flaky "terminating" output event)
- and just send a disconnect response back

polinasok (approved):
LGTM

Issue #2768, opened 2021-10-29
pkg/proc/native: intermittent test failures in TestWatchpointsBasic


aarzilli:
Example: https://delve.teamcity.com/viewLog.html?buildId=27817

For whatever reason it seems that the hardware breakpoint state gets restored somehow on some threads, sometimes, after a hardware breakpoint has been disabled.

PR #2769, opened 2021-10-29
proc/native: disable watchpoints on Windows

aarzilli:
There are persistent issues with watchpoints on Windows, it is not
clear whether it's a problem with the VM running the tests or if there
is a real bug in our implementation of hardware breakpoints on Windows.

Until the cause can be determined watchpoints on Windows will be
disabled.

Updates #2768


derekparker (approved):
LGTM

PR #2770, opened 2021-10-30
proc/gdbserial: workaround for debugserver register set bug

### LINKED ISSUE #2767: No error on attempt to call optimized function
```
Delve Debugger
Version: 1.7.0
Build: b41e47a30539294818283b89dae4d4879242e0c5
```

```
go version go1.17.2 darwin/amd64
```

darwin/amd64

4. What did you do?

Debugged the program

```
package main

func main() {
  res := math.Floor(2.5)
  println(res)
}
```

and called `math.Floor(2.5)` at line with println.

Result 2 or error why result cannot be computed

6. What did you see instead?

Result 0 and no error.

Command-line delve client shows the error saying that calling of optimized function is not possible. In go1.16.5 same call produces 'too many function arguments' error.

Ide log: https://gist.github.com/nd/65887e7c494da72ce768a9ef040d4752
Delve command-line client log: https://gist.github.com/nd/a6a15b6cdaeaf03eb46e60d2408dd874

aarzilli:
Debugserver has a bug where writing to a AVX-2 or AVX-512 register does
not work unless it is followed by at least a write to a AVX (not 2 or
512) register.

See also: https://bugs.llvm.org/show_bug.cgi?id=52362

Fixes #2767

### DIFF
--- a/_fixtures/setymmreg/main.go
+++ b/_fixtures/setymmreg/main.go
@@ -1,0 +1,7 @@
+package main
+
+func asmFunc()
+
+func main() {
+	asmFunc()
+}
--- a/_fixtures/setymmreg/setymmreg_amd64.s
+++ b/_fixtures/setymmreg/setymmreg_amd64.s
@@ -1,0 +1,9 @@
+#include "textflag.h"
+
+TEXT ·asmFunc(SB),0,$0-0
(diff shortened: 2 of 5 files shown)


derekparker (approved):
LGTM

PR #2771, opened 2021-10-31
*: Replace libbpfgo with cilium/ebpf

derekparker:
This change has a few implications, the biggest of which is the fact
that we no longer need to rely on a C library (libbpf) to build the eBPF
enabled version of Delve. This makes Delve with eBPF support go get-able
which will allow us to more easily enable it by default once it has
complete parity with the ptrace based version.


derekparker:
Sorry for the weird diff, I originally had the go mod / vendor changes in a different commit but git did something odd during the rebase onto master and combined them.

derekparker:
> Sorry for the weird diff, I originally had the go mod / vendor changes in a different commit but git did something odd during the rebase onto master and combined them.

Nevermind, I rebased it to make it easier to review.

derekparker:
Not sure why linux/386 staticcheck test is failing. I don't have access to a 386 machine, but I've set `GOARCH` and ran staticcheck but I actually get a different diff than what's happening on CI.

@aarzilli any thoughts?

aarzilli:
> Not sure why linux/386 staticcheck test is failing. I don't have access to a 386 machine, but I've set `GOARCH` and ran staticcheck but I actually get a different diff than what's happening on CI.
> 
> @aarzilli any thoughts?

Weird, it seems to complain that it doesn't build but clearly it is, otherwise the tests wouldn't be running.

aarzilli:
Actually, it's not weird. It's correct. It's because we are always checking the amd64 build in TestStaticcheck and because that's a crosscompile in linux/386 it disables cgo. The file `helpers.go` should add `&& cgo` to the `go:build` directive, and `helpers_disabled.go` should add `|| !cgo`.

pkg/proc/native/proc_linux.go
```diff
@@ -735,10 +735,10 @@ func (dbp *nativeProcess) SetUProbe(fnName string, goidOffset int64, args []ebpf
 	}
 
 	debugname := dbp.bi.Images[0].Path
-	offset, err := ebpf.SymbolToOffset(debugname, fnName)
-	if err != nil {
-		return err
-	}
+	// offset, err := ebpf.SymbolToOffset(debugname, fnName)
```
aarzilli, line 738: leftover?

pkg/proc/internal/ebpf/helpers.go
```diff
@@ -1,13 +1,12 @@
-//go:build ebpf
-// +build ebpf
+//go:build linux && amd64 && go1.16 && go1.17
+// +build linux,amd64,go1.16,go1.17
```
aarzilli, line 2: Isn't it sufficient to have go1.17 here? Won't go1.17 always imply go1.16? Also why is go1.17 needed?

pkg/proc/native/proc_ebpf_linux.go
```diff
@@ -1,6 +1,3 @@
-//go:build ebpf
-// +build ebpf
-
 package native
```
aarzilli, line 1: shouldn't this have the same build tags as the rest of the ebpf stuff (linux && amd64 && go1.17 && cgo)?

aarzilli (commented):
pkg/proc/internal/ebpf/trace_bpfel_x86.o is committed, I think accidentally?

pkg/proc/native/proc_linux.go
derekparker: Yes, removed.

pkg/proc/internal/ebpf/helpers.go
derekparker: We really just need > go1.15 because of the embed feature.

pkg/proc/native/proc_ebpf_linux.go
derekparker: Yeah, you're right.

derekparker:
> pkg/proc/internal/ebpf/trace_bpfel_x86.o is committed, I think accidentally?

I actually committed it on purpose. It's an artifact of `go generate` and must be in the repo for Delve w/ eBPF support to be go get-able.

derekparker:
Fun new failures on Windows:

```
------- Stdout: -------
=== RUN   TestCgoSources
    proc_test.go:117: Launch(): fork/exec Z:\buildAgent\temp\buildTmp\cgostacktest.a0e0eca0: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
--- FAIL: TestCgoSources (3.05s)
```

aarzilli:
> Fun new failures on Windows:
> 
> ```
> ------- Stdout: -------
> === RUN   TestCgoSources
>     proc_test.go:117: Launch(): fork/exec Z:\buildAgent\temp\buildTmp\cgostacktest.a0e0eca0: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
> --- FAIL: TestCgoSources (3.05s)
> ```

Well, maybe you should stop trying to add viruses to delve.

aarzilli (approved):

derekparker:
> > Fun new failures on Windows:
> > ```
> > ------- Stdout: -------
> > === RUN   TestCgoSources
> >     proc_test.go:117: Launch(): fork/exec Z:\buildAgent\temp\buildTmp\cgostacktest.a0e0eca0: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
> > --- FAIL: TestCgoSources (3.05s)
> > ```
> 
> Well, maybe you should stop trying to add viruses to delve.

🤣

PR #2773, opened 2021-11-02
delve: support linux-loong64 native debug

abner-chenc:
delve now can be built (-mod=vendor) to loong64-arch and running on linux-loong64 OS. 
The following commands work correctly on linux-loong64.
  1.  running the program: continue, next, restart, step-instruction
  2.  manipulating breakpoints: break, breakpoints, clear, clearall
  3.  viewing program variables and memory: regs
  4.  viewing the call stack and selecting frames: stack, up, down
  5.  other commands: disassemble(GNU), funcs, list, sources

LoongArch documents:
 1.  https://github.com/loongson/LoongArch-Documentation

Patches submitted by related projects for loong64:
 1.  https://go-review.googlesource.com/c/arch/+/358854
 2.  https://go-review.googlesource.com/c/sys/+/355550
 3.  https://go-review.googlesource.com/c/go/+/342303
 4.  https://github.com/loongson/go (branch: loong64-go1.15.6)
 5.  https://github.com/creack/pty (already merged)


chenguoqi-cc:
@XiaodongLoong @limeidan

aarzilli:
See [Delve's porting guide](https://github.com/go-delve/delve/blob/master/Documentation/internal/portnotes.md). Also see the build failures.

aarzilli:
In general this seems mostly fine. The main two problems are that (a) it doesn't compile on anything but go1.18 and (b) we need to do CI for it.

chenguoqi-cc:
Ths

chenguoqi-cc:
> In general this seems mostly fine. The main two problems are that (a) it doesn't compile on anything but go1.18 and (b) we need to do CI for it.

Thank you！


**CI macheine**： We can provide machines with loongarch64 instruction architecture that can be accessed remotely. Of course, we can also provide physical machines for CI.    What resources do we need to provide to access CI？


**Golang compiler**：    This  is go1.15.6 （https://github.com/loongson/go， branch loong64-go1.15.6） that we have transplanted. It can work normally at present。   The patch on go loong64 for master has been submitted to the golang community (see  https://go-review.googlesource.com/c/go/ +/342303）。 We are doing our best to merge the patch into the master branch of go as soon as possible.

aarzilli:
> We can provide machines with loongarch64 instruction architecture that can be accessed remotely. Of course, we can also provide physical machines for CI. What resources do we need to provide to access CI？

Can they run the JVM? These are the instructions for adding an agent: https://www.jetbrains.com/help/teamcity/setting-up-and-running-additional-build-agents.html
cc @nd anything we should know about our setup regarding this?

> Golang compiler

I don't think you understand. It is fine if delve needs 1.18 to compile on loong64, because that's the first version of go that supports loong64, but on other architectures it needs to compile all the way down to 1.15, at least.
It seems to me that it is mostly because of `elf.EM_LOONGARCH` not being defined, if that's the only thing you should simply define a _elf_EM_LOONGARCH constant locally and use that instead of `elf.EM_LOONGARCH`.

Also the part of the original message where you say "The following commands work correctly on linux-loong64" is slightly concerning. Supporting a few commands isn't a sufficient condition to get merged, it needs to actually pass our test suite (see the porting notes for exceptions).

nd:
> anything we should know about our setup regarding this?

It should work as described. Let me know in case of problems.

derekparker:
Any updates on this work?

PR #2775, opened 2021-11-03
TeamCity: disable Windows Defender

aarzilli:
Shaves off 3 minutes out of every windows build.

### DIFF
--- a/_scripts/test_windows.ps1
+++ b/_scripts/test_windows.ps1
@@ -3,6 +3,8 @@ param (
     [Parameter(Mandatory = $true)][string]$arch
 )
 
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

derekparker:
I know we're technically in a freeze here. The eBPF change made it in (although it's gated by a feature flag so no big concern there). However this only touches CI and get's us green again so I'm going to merge it.

aarzilli:
> I know we're technically in a freeze here. The eBPF change made it in

ah sorry I forgot about the freeze.

Also this doesn't get us back on the green, the Windows failure was a transitory, when I re-ran the tests it didn't happen again.

PR #2778, opened 2021-11-05
service/dap: misc remote attach improvements

polinasok:
- Improve mismatched launch/attach request/mode errors
- Log if process was halted/running on client exit
- Update attach config comments to include new remote mode

Updates #2328

### DIFF
--- a/service/dap/daptest/client.go
+++ b/service/dap/daptest/client.go
@@ -173,9 +173,11 @@ func (c *Client) ExpectOutputEventTerminating(t *testing.T) *dap.OutputEvent {
 	return c.ExpectOutputEventRegex(t, `Terminating process [0-9]+\n`)
 }
 
-func (c *Client) ExpectOutputEventClosingClient(t *testing.T) *dap.OutputEvent {
+const ClosingClient = "Closing client session, but leaving multi-client DAP server at .+:[0-9]+ with debuggee %s\n"
+
+func (c *Client) ExpectOutputEventClosingClient(t *testing.T, status string) *dap.OutputEvent {
 	t.Helper()
-	return c.ExpectOutputEventRegex(t, `Closing client session, but leaving multi-client DAP server running at .+:[0-9]+\n`)
+	return c.ExpectOutputEventRegex(t, fmt.Sprintf(ClosingClient, status))
 }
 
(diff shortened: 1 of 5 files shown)


service/dap/server.go
```diff
@@ -904,8 +914,8 @@ func cleanExeName(name string) string {
 func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 	var err error
 	if s.debugger != nil {
-		s.sendShowUserErrorResponse(request.Request, FailedToLaunch,
-			"Failed to launch", "debugger already started - use remote attach to connect to a server with an active debug session")
+		s.sendShowUserErrorResponse(request.Request, FailedToLaunch, "Failed to launch",
```
aarzilli, line 917: What's the point of printing the server address here? Clearly the user already knows it.

polinasok: Not if they accidentally used the wrong configuration. This happened to me. I had a dlv-mux server running, which I forgot to take down in some tab. Then to check something else I switched to my "Launch" configuration, forgetting that it had a port set. So instead of having vscode-go start dlv-dap, it connected to my mux server. It took me a moment to make sense of the error message. I thought this tweak could help clarify what happened.

aarzilli (approved):

service/dap/daptest/client.go
```diff
@@ -173,9 +173,11 @@ func (c *Client) ExpectOutputEventTerminating(t *testing.T) *dap.OutputEvent {
 	return c.ExpectOutputEventRegex(t, `Terminating process [0-9]+\n`)
 }
 
-func (c *Client) ExpectOutputEventClosingClient(t *testing.T) *dap.OutputEvent {
+const ClosingClient = "Closing client session, but leaving multi-client DAP server at .+:[0-9]+ with debuggee %s\n"
```
derekparker, line 176: Do we need to export this?

polinasok: They are used directly in some server_test tests in the dap package. I can add helpers to hide them if you think it's worth it.

polinasok: I mean additional helpers.

derekparker (approved):
LGTM

Issue #2757, opened 2021-10-25
service/dap: close of closed channel


aarzilli:
Full log: https://delve.teamcity.com/viewLog.html?buildId=27460

```
  27	=== RUN   TestStepInstruction
    	DAP server listening at: [::]:55577
    	2021-10-25T16:12:08+02:00 warning layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
    	2021-10-25T16:12:08+02:00 debug layer=dap DAP server pid = 3940
    	Connecting to server at: [::]:55577
    	2021-10-25T16:12:08+02:00 debug layer=dap DAP connection started
    	2021-10-25T16:12:08+02:00 debug layer=dap [<- from client]{"seq":1,"type":"request","command":"initialize","arguments":{"adapterID":"go","locale":"en-us","linesStartAt1":true,"columnsStartAt1":true,"pathFormat":"path","supportsVariableType":true,"supportsVariablePaging":true,"supportsRunInTerminalRequest":true}}
    	2021-10-25T16:12:08+02:00 debug layer=dap [-> to client]{"seq":0,"type":"response","request_seq":1,"success":true,"command":"initialize","body":{"supportsConfigurationDoneRequest":true,"supportsFunctionBreakpoints":true,"supportsConditionalBreakpoints":true,"supportsEvaluateForHovers":true,"supportsSetVariable":true,"supportsExceptionInfoRequest":true,"supportTerminateDebuggee":true,"supportsDelayedStackTraceLoading":true,"supportsLogPoints":true,"supportsDisassembleRequest":true,"supportsClipboardContext":true,"supportsSteppingGranularity":true,"supportsInstructionBreakpoints":true}}
    	2021-10-25T16:12:08+02:00 debug layer=dap [<- from client]{"seq":2,"type":"request","command":"launch","arguments":{"mode":"exec","program":"/Users/builduser/agent/temp/buildTmp/testvariables.86246a30","request":"launch","stopOnEntry":false}}
    	2021-10-25T16:12:08+02:00 debug layer=dap parsed launch config: {
    		"mode": "exec",
    		"program": "/Users/builduser/agent/temp/buildTmp/testvariables.86246a30",
    		"backend": "default",
    		"stackTraceDepth": 50
    	}
    	2021-10-25T16:12:08+02:00 debug layer=dap launching binary '/Users/builduser/agent/temp/buildTmp/testvariables.86246a30' with config: {
    		"mode": "exec",
[...]

PR #2781, opened 2021-11-08
server/dap: stop running command if conn closed - fixes nil dereference bug

polinasok:
If connection is closed while the debuggee is running, any pending goroutine handling the running command and analyzing its stop reasons should stop right away without calling any debugger methods as the debugger pointer would have already been cleared at session closure.

This fixes the following failure that especially impacts repeated remote attach to a multi-client server, but can also show up when server is shutting down:
```
2021-11-08T00:12:34-08:00 error layer=dap recovered panic: runtime error: invalid memory address or nil pointer dereference
goroutine 50 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/go-delve/delve/service/dap.(*Session).recoverPanic(0xc0002421c0, {0x16c43e0, 0xc000ea32f0})
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:516 +0x66
panic({0x1570180, 0x1a37de0})
	/usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/go-delve/delve/service/debugger.(*Debugger).StopReason(0xc000b41b40)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/debugger/debugger.go:2056 +0x38
github.com/go-delve/delve/service/dap.(*Session).stoppedGs(0xc0002421c0, 0xc000eb87e0)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:3589 +0xab
github.com/go-delve/delve/service/dap.(*Session).resumeOnceAndCheckStop(0xc0002421c0, {0x160a819, 0xc00024227c}, 0xc000dc5bf8)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:3556 +0x94
github.com/go-delve/delve/service/dap.glob..func1(0xb, {0x160a819, 0xc000d91518}, 0xc000099bc0)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:3543 +0x1e
github.com/go-delve/delve/service/dap.(*Session).runUntilStop(0xc0002421c0, {0x160a819, 0x8}, 0x18)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:3535 +0x107
github.com/go-delve/delve/service/dap.(*Session).runUntilStopAndNotify(0xc0002421c0, {0x160a819, 0x8}, 0x0)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:3421 +0x46
github.com/go-delve/delve/service/dap.(*Session).onConfigurationDoneRequest(0xc0002421c0, 0xc000ea32f0, 0xc000099aa0)
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:1577 +0x305
github.com/go-delve/delve/service/dap.(*Session).handleRequest.func1()
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:653 +0x85
created by github.com/go-delve/delve/service/dap.(*Session).handleRequest
	/Users/polina/go/pkg/mod/github.com/go-delve/delve@v1.7.3-0.20211103164926-183bb8dffd9f/service/dap/server.go:651 +0x10c5
```

Updates #2328

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -849,11 +849,25 @@ func TestRemoteDAPClientAfterContinue(t *testing.T) {
 
 	go func() { // Capture logging
 		for scanErr.Scan() {
-			t.Log(scanErr.Text())
+			text := scanErr.Text()
+			if strings.Contains(text, "Internal Error") {
+				t.Error("ERROR", text)
+			} else {
+				t.Log(text)
+			}
 		}
 	}()
 
 	c := newDAPRemoteClient(t, listenAddr)
+	c.ContinueRequest(1)
+	c.ExpectContinueResponse(t)
+	c.DisconnectRequest()
+	c.ExpectOutputEventClosingClient(t)
+	c.ExpectDisconnectResponse(t)
+	c.ExpectTerminatedEvent(t)
+	c.Close()
+
+	c = newDAPRemoteClient(t, listenAddr)
 	c.DisconnectRequestWithKillOption(true)
 	c.ExpectOutputEventDetachingKill(t)
 	c.ExpectDisconnectResponse(t)
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -112,6 +112,8 @@ type Server struct {
 type Session struct {
 	config *Config
 
+	id int
+
 	// stackFrameHandles maps frames of each goroutine to unique ids across all goroutines.
 	// Reset at every stop.
 	stackFrameHandles *handlesMap
@@ -131,7 +133,7 @@ type Session struct {
 	mu sync.Mutex
 
 	// conn is the accepted client connection.
-	conn io.ReadWriteCloser
+	conn *connection
 	// debugger is the underlying debugger service.
 	debugger *debugger.Debugger
 	// binaryToRemove is the temp compiled binary to be removed on disconnect (if any).
@@ -169,6 +171,29 @@ type Config struct {
 	// StopTriggered is closed when the server is Stop()-ed.
 	// Can be used to safeguard against duplicate shutdown sequences.
 	StopTriggered chan struct{}
+}
+
+type connection struct {
+	io.ReadWriteCloser
+	closed chan struct{}
+}
+
+func (c *connection) Close() error {
+	select {
+	case <-c.closed:
+	default:
+		close(c.closed)
+	}
+	return c.ReadWriteCloser.Close()
+}
+
+func (c *connection) isClosed() bool {
+	select {
+	case <-c.closed:
+		return true
+	default:
+		return false
+	}
 }
 
 type process struct {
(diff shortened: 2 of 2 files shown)


aarzilli (commented):
Why not leave s.debugger set?

polinasok:
> Why not leave s.debugger set?

A couple of reasons. 
(1) It won't help and just shift the issue in case of back-to-back sessions. New auto-resume logic relies on locking state changes and tracking if there is a pending running command or "real" halt. Those mechanisms are per session. When a new session starts up, it halts execution. Auto-resuming goroutine from the previous session pauses and checks if it needs to restart. It doesn't register that it is an intentional interruption because there is no shared state between this and the other session.*
(2) In several cases, shutdown code runs more than once. If debugger is not nil, we halt+detach again and again and print duplicate logging messages.

* I think there is still risk that halting from a new (dap or rpc) client will interrupt the auto-resume loop from the previous client at the wrong time and go unregistered or try to halt while halted because separate clients don't synchronize between each other. But that's out of scope for this PR. And yet another reason to move auto-resume logic inside of the debugger.

polinasok:
@suzmue

service/dap/server.go
```diff
@@ -171,6 +173,29 @@ type Config struct {
 	StopTriggered chan struct{}
 }
 
+type connection struct {
+	io.ReadWriteCloser
+	closed chan struct{}
```
derekparker, line 178: Is this preferable to a simple `bool` because of the necessary "close once and only once" semantics? E.g., bool can be switched from true/false but a chan can only be closed once?

cmd/dlv/dlv_test.go
```diff
@@ -849,11 +849,25 @@ func TestRemoteDAPClientAfterContinue(t *testing.T) {
 
 	go func() { // Capture logging
 		for scanErr.Scan() {
-			t.Log(scanErr.Text())
+			text := scanErr.Text()
+			if strings.Contains(text, "Internal Error") {
+				t.Error("ERROR", text)
```
derekparker, line 854: Is this just for visibility?

derekparker (commented):
A few questions.

cmd/dlv/dlv_test.go
polinasok: Yes, otherwise it looked exactly like the rest of logging

service/dap/server.go
polinasok: I guess I was modeling it after stopTriggered (aka stopChan in rpccommon).

derekparker: Ack, makes sense.

cmd/dlv/dlv_test.go
derekparker: Ack, sgtm.

derekparker (approved):
LGTM—also awaiting @aarzilli review.

aarzilli (approved):

polinasok:
Thank you, both!

polinasok:
But... I have been thinking more about this. Yes, I was modeling after `stopTriggered`, but that channel is only closed in a single function/goroutine - during `Stop`, which is not supposed to be called more than once. And connection can be closed more than once, hence the select guard. But since then it can be closed both from serve goroutine and from main goroutine that calls Stop, will this lead to an issue like #2757?

If we use a bool like Derek suggested, we can set it to `closed=true` multiple times without a problem. And we will never set it to false other than at creation, so there will be no competition between multiple goroutines writing, no locking is necessary, right? And if I always set it to false preemptively before closing, then even if the two events get interrupted and we bail out of the running command slightly prematurely, it's ok. We just can't do it the other way and have the running command think based on the not-yet-updated flag that things are still active.

Would this be better? #2783

PR #2783, opened 2021-11-09
service/dap: use bool flag instead of channel to detect closed connection

polinasok:
Follow-up on #2781

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -175,25 +175,14 @@ type Config struct {
 
 type connection struct {
 	io.ReadWriteCloser
-	closed chan struct{}
+	isClosed bool
 }
 
 func (c *connection) Close() error {
-	select {
-	case <-c.closed:
-	default:
-		close(c.closed)
-	}
+	// Set the flag first in case we get interrupted
+	// by a check on another goroutine.
+	c.isClosed = true
 	return c.ReadWriteCloser.Close()
-}
-
-func (c *connection) isClosed() bool {
-	select {
-	case <-c.closed:
-		return true
-	default:
-		return false
-	}
 }
 
 type process struct {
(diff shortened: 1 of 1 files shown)


aarzilli:
What's synchronizing this with the goroutine that checks isClosed? isClosed gets called by the goroutine spawned by `dap.ContinueRequest` but closed is set by the goroutine in the handlerLoop, right?

polinasok:
> What's synchronizing this with the goroutine that checks isClosed? isClosed gets called by the goroutine spawned by `dap.ContinueRequest` but closed is set by the goroutine in the handlerLoop, right?

IsClosed is checked on a goroutine started by any of the running requests (continue, configurationDone with !stopOnEntry, etc). It is closed by the main goroutine in Stop or loop goroutine in ServeDAPCodec on error (before detach) or disconnect (after detach) or in DisconnectRequest in case of accept-multiclient (when server goes on). I don't know what I was thinking. Derek asked about bool flag, and I wanted too see what it would look like given the double channel closure concern, but that's no different than just checking directly if debugger is nil before calling it. With the channel, or bool, or direct debugger check we are risking that somebody will interfere in the middle of the check-then-act logic.

I need to think about this more :(

aarzilli:
Yes, I think both this and #2781 are broken.
There's nothing synchronizing the `s.debugger = nil` with the runUntilStop goroutine, that goroutine could be doing anything and have the rug pulled from under it.
Maybe the disconnect-and-continue case should actually stop the target and then resume it with a reduced version of the resume loop, that doesn't do logpoints or anything else weird.
Either that or disconnect and runUntilStop need to synchronize somehow.

Anyway, I think we made a change in haste and more thought should go into this. I'd leave it as is for 1.7.2, it's probably good enough.

derekparker:
Should we revert the last commit before 1.7.3? What's the overall confidence level?

aarzilli:
The contention window is fairly small in normal conditions, IMO it should stay.

polinasok:
It's not an ideal solution and can race and run into issues in theory, but it does solve the very obvious and common crash in practice.

polinasok:
> It's not an ideal solution and can race and run into issues in theory, but it does solve the very obvious and common crash in practice.

And I will keep working on this to have a better solution soon. But yeah, agree with @aarzilli that it is worth keeping it.

derekparker:
Any updates on this?

polinasok:
I still plan to finish this, but that won't happen for a couple of weeks.

polinasok:
I have something else in the works, so closing this.

PR #2785, opened 2021-11-10
_scripts: fix test_linux.sh version check

aarzilli:
The version check to disable the buildvcs option is wrong, it doesn't
matter now but it will start failing when 1.18.1 is released.


derekparker (approved):
LGTM

PR #2789, opened 2021-11-14
version: fix build version using buildInfo

aarzilli:
In go1.18 buildInfo will include the git revision hash, use that to fix
the Build field of Version so that it is correct even if Delve wasn't
built using make.go.


derekparker (approved):
LGTM

PR #2790, opened 2021-11-14
proc/*: minor miscellaneous code cleanups

aarzilli:
* made Pid a field of Target instead of a method of Process
* changed argument of NewTarget to ProcessInternal, since that's the
interface that backends have to implement
* removed warnings about ProcessInternal since there is no way for
users of pkg/proc to access those methods anyway
* made RecordingManipulation an optional interface for backends, Target
supplies its own dummy implementation when the backend doesn't
* inlined small interfaces that only existed to be inlined in
proc.Process anyway
* removed unused function findExecutable in the Windows and no-native
darwin backends
* removed (*EvalScope).EvalVariable, an old synonym for EvalExpression

### DIFF
--- a/pkg/proc/interface.go
+++ b/pkg/proc/interface.go
@@ -13,9 +13,20 @@ import (
 // There is one exception to this rule: it is safe to call RequestManualStop
 // concurrently with ContinueOnce.
 type Process interface {
-	Info
-	ProcessManipulation
-	RecordingManipulation
+	// ResumeNotify specifies a channel that will be closed the next time
+	// ContinueOnce finishes resuming the target.
+	ResumeNotify(chan<- struct{})
+	BinInfo() *BinaryInfo
+	EntryPoint() (uint64, error)
+
+	// RequestManualStop attempts to stop all the process' threads.
+	RequestManualStop() error
+	// CheckAndClearManualStopRequest returns true the first time it's called
+	// after a call to RequestManualStop.
+	CheckAndClearManualStopRequest() bool
+
+	FindThread(threadID int) (Thread, bool)
+	ThreadList() []Thread
 
 	Breakpoints() *BreakpointMap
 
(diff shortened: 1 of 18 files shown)


pkg/proc/target.go
```diff
@@ -36,8 +36,12 @@ const (
 // Target represents the process being debugged.
 type Target struct {
 	Process
+	RecordingManipulation
 
-	proc ProcessInternal
+	proc   ProcessInternal
+	recman RecordingManipulationInternal
+
+	Pid int
```
derekparker, line 44: I'd actually prefer this remain unexported and we expose it via a method. It makes it clear that it should never be modified (and makes it impossible to modify) via any consumer package.

pkg/proc/interface.go
```diff
@@ -95,31 +112,3 @@ type Checkpoint struct {
 	When  string
 	Where string
 }
-
-// Info is an interface that provides general information on the target.
-type Info interface {
```
derekparker, line 100: These interfaces were more for organization than to be directly used. I guess I'm ok with removing them, but how would you feel about simple unexporting them to make clear they aren't to be used outside this package?

derekparker (changes requested):

pkg/proc/target.go
aarzilli: Done.

pkg/proc/interface.go
aarzilli: IMO there's so few methods that they make things harder to understand rather than clearer. But I don't feel stronger about it so if you want I'll revert.
Making them unexported would be worse because they wouldn't be shown in the godoc documentation.

derekparker: Ack, I don't feel strongly so SGTM.

derekparker (approved):

Issue #2786, opened 2021-11-11
Functions not found in program with CGO on windows


nd:
1. What version of Delve are you using (`dlv version`)?

Delve Debugger
Version: 1.7.2
Build: $Id: 5b6f24e7fbcad3fe3bc574d3763b2f20afa1d6a1 $

2. What version of Go are you using? (`go version`)?

go version go1.17.3 windows/amd64

3. What operating system and processor architecture are you using?

windows 10/amd64 

4. What did you do?

Compiled program with CGO (with gcc in path, gcc version: gcc (Rev2, Built by MSYS2 project) 11.2.0) using 'go build -o exe -gcflags "all=-N -l" cgo'

```
package main

/*
 const int a=1;
*/
import "C"

func F() {
	println("ok")
}

func main() {
	println("ok")
}
```

Ran 'dlv exec exe' and executed funcs command.

5. What did you expect to see?

At least main and F

6. What did you see instead?

No funcs.

aarzilli:
I wouldn't expect to see F, it isn't called and the deadcode pass of the linker will remove it, main should be there. I think there is some problem with mingw 11, it did break our tests. Can you send me the executable?

nd:
https://drive.google.com/file/d/1NIrPxMmIaEIwjtfqDVPou4p6Y3szYFqz/view?usp=sharing

nd:
Also noticed that target version is missing:

```
2021-11-11 09:23:09,734 [ 317970]  DEBUG -           #com.goide.dlv.DlvVm - OUT: {"method":"RPCServer.GetVersion","params":[{}],"id":2} 
2021-11-11 09:23:09,735 [ 317971]  DEBUG -           #com.goide.dlv.DlvVm - IN: {"id":2,"result":{"DelveVersion":"Version: 1.7.2\nBuild: c207db792a0757e8db7420d6729b3da2608c5f31","APIVersion":2,"Backend":"native","TargetGoVersion":"","MinSupportedVersionOfGo":"1.15.0","MaxSupportedVersionOfGo":"1.17.0"},"error":null} 
```

aarzilli:
Blind guess: gcc switched to dwarf5, go still produces dwarf4 and the linker messes everything up. Haven't looked yet.

nd:
It looks like dwarf5 was used, but setting `CGO_CFLAGS=-gdwarf-4` doesn't help.

LomikXP:
[...]

PR #2791, opened 2021-11-15
proc: support DWARF5 on Windows/macOS

aarzilli:
### proc: use debug_line_str section for PE and Macho-O files

Updates #2786

### proc: log errors reading debug_info

Because of an incorrect use of debug/dwarf.Reader errors encountered
while reading debug_info were not reported.

Updates #2786

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -911,10 +911,13 @@ func (image *Image) Close() error {
 	return err2
 }
 
-func (image *Image) setLoadError(fmtstr string, args ...interface{}) {
+func (image *Image) setLoadError(logger *logrus.Entry, fmtstr string, args ...interface{}) {
 	image.loadErrMu.Lock()
 	image.loadErr = fmt.Errorf(fmtstr, args...)
 	image.loadErrMu.Unlock()
+	if logger != nil {
+		logger.Errorf("error loading binary %q: %v", image.Path, image.loadErr)
+	}
 }
 
 // LoadError returns any error incurred while loading this image.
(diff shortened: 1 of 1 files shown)


aarzilli:
Note: this will not do anything until the fixes for https://github.com/golang/go/issues/49590 are merged into Go.

pkg/proc/bininfo.go
```diff
@@ -911,10 +911,11 @@ func (image *Image) Close() error {
 	return err2
 }
 
-func (image *Image) setLoadError(fmtstr string, args ...interface{}) {
+func (image *Image) setLoadError(logger *logrus.Entry, fmtstr string, args ...interface{}) {
 	image.loadErrMu.Lock()
 	image.loadErr = fmt.Errorf(fmtstr, args...)
 	image.loadErrMu.Unlock()
+	logger.Errorf("error loading binary %q: %v", image.Path, image.loadErr)
```
derekparker, line 918: nit: logger could be nil.

pkg/proc/bininfo.go
```diff
@@ -911,10 +911,11 @@ func (image *Image) Close() error {
 	return err2
 }
 
-func (image *Image) setLoadError(fmtstr string, args ...interface{}) {
+func (image *Image) setLoadError(logger *logrus.Entry, fmtstr string, args ...interface{}) {
```
derekparker, line 914: Do we never surface this error? I'm questioning the need to inject the logger at all.

derekparker (changes requested):

pkg/proc/bininfo.go
aarzilli: Fixed.

pkg/proc/bininfo.go
aarzilli: We do when we first create the target, in NewTarget, but not for additional images that appear later (like plugins/dlls). I think this is the right behavior, but we should still log it.
In the connected issue the error was *not* surfaced, but it was because we were dropping it in the for loops I've changed, not because of this.

derekparker (approved):
LGTM

derekparker (approved):
LGTM

Issue #2795, opened 2021-11-16
Delve isn't able to find any of my source files, so it doesn't stop at any breakpoints


davidmichaelkarr:
Please answer the following before submitting your issue:

Note: Please include any substantial examples (debug session output,
stacktraces, etc) as linked gists.

1. What version of Delve are you using (`dlv version`)?
    1.7.1
2. What version of Go are you using? (`go version`)?
    1.17.3 
3. What operating system and processor architecture are you using?
    Linux Ubuntu/amd64
4. What did you do?
    Using vscode with the vscode-go extension, I created a multimodule/multipackage/multifolder project (although still pretty small), where one module uses cgo.  The project has no compile errors, and it starts up successfully, although it has operational issues in testing, which is simply issues with my application code.

I set a breakpoint on the first line of the main function, which is a print statement.  When I start the debugger, I see the print statement in the console log, but it doesn't stop at the breakpoint.  When the application starts, the breakpoint red circle changes to a hollow circle, and when I hover on that, it prints "could not find file /home/dk068x/git/voltagego/main.go". That path is valid, and the file does exist.

The start of the log shows this:
```
Starting: /home/dk068x/go/bin/dlv-dap dap --check-go-version=false --listen=127.0.0.1:41147 --log-dest=3 from /home/dk068x/git/voltagego
DAP server listening at: 127.0.0.1:41147
```

Note that I had worked on another application in vscode-go, and when I switch to that, even without bringing vscode down, and then run that in the debugger, it stops at breakpoints. That application is much simpler, being a single module.

5. What did you expect to see?
For the debugger to stop at the breakpoint.
6. What did you see instead?
It didn't stop at the breakpoint.

aarzilli:
Any symlinks in the path of main.go? cc @polinasok

davidmichaelkarr:
Yes, one directory in the path is a symlink to another directory on the VM, and perhaps unrelated, that path is within a VirtualBox shared folder path, so the actual files reside on the Windows host (I run Ubuntu on a VirtualBox VM on a Windows 10 host).

aarzilli:
It is likely that go is resolving those symlinks while compiling and the executable contains the real path of the files rather than what vscode thinks is the path. There is a substitute path option in launch.json for this.

davidmichaelkarr:
I'd appreciate a little more information.  How exactly do I resolve this?

[...]

PR #2797, opened 2021-11-17
service/dap: expose sources command in evaluate request

suzmue:
This change exposes the list sources feature in the debug console in order to facilitate easier substitute path mappings. When user's have symlinks, are debugging on a remote host, or use trimpath to build, the source paths may not line up with the source files sent from the dap client. By allowing users to see the source names that the program was actually built with, this will allow users to more easily set up a correct configuration of substitutePath to debug their programs.

Updates #2795

### DIFF
--- a/service/dap/command.go
+++ b/service/dap/command.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/go-delve/delve/pkg/config"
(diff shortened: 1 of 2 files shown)


service/dap/command.go
```diff
@@ -113,3 +120,16 @@ func (s *Session) evaluateConfig(_, _ int, expr string) (string, error) {
 		return res, nil
 	}
 }
+
+func (s *Session) sources(_, _ int, args string) (string, error) {
+	sources, err := s.debugger.Sources(args)
```
polinasok, line 125: s/args/filter/?

service/dap/command.go
```diff
@@ -113,3 +120,16 @@ func (s *Session) evaluateConfig(_, _ int, expr string) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	sort.Strings(sources)
```
polinasok, line 129: rpc client doesn't see to do any sorting. Isn't the data coming from debugger sorted?

service/dap/command.go
```diff
@@ -113,3 +120,16 @@ func (s *Session) evaluateConfig(_, _ int, expr string) (string, error) {
...
+	var buf bytes.Buffer
+	for _, d := range sources {
+		fmt.Fprintln(&buf, d)
+	}
+	return buf.String(), nil
```
polinasok, line 134: Why not use strings.Join?

polinasok (changes requested):

service/dap/command.go
suzmue: the terminal client sorts the results before printing the list of source files.

service/dap/command.go
suzmue: done.

service/dap/command.go
suzmue: done.

aarzilli (commented):
See test failures. Also I guess at some point we might want to consider if this and pkg/terminal/command.go might want to share code.

service/dap/command.go
polinasok: I see. Then, yeah, what @aarzilli said: we should consider sharing code with the terminal package at some point.

polinasok (approved):

aarzilli (approved):

derekparker (approved):
LGTM

PR #2798, opened 2021-11-17
_scripts: add option to disable invoking git during build

herbygillot:
This is useful to be able to build using Delve's build scripts when
working from within a directory that has been extracted from a source tarball,
or wherever we wouldn't have git history.

This can be invoked as:

```
  go run _scripts/make.go build --no-git
```

### DIFF
--- a/_scripts/make.go
+++ b/_scripts/make.go
@@ -24,6 +24,7 @@ var TestSet, TestRegex, TestBackend, TestBuildMode string
 var Tags *[]string
 var Architecture string
 var OS string
+var DisableGit bool
 
 func NewMakeCommands() *cobra.Command {
 	RootCommand := &cobra.Command{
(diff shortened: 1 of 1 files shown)


herbygillot:
At first glance, build failures look unrelated to this change...

aarzilli (approved):

aarzilli:
Approved but this is just going the long way around to run `go install github.com/go-delve/delve/cmd/dlv`

herbygillot:
> Approved but this is just going the long way around to run `go install github.com/go-delve/delve/cmd/dlv`

So I'm a "package maintainer", if you will, maintaining the port (the "package") for `delve` in the [MacPorts](https://www.macports.org) project.  You can see more details here: https://ports.macports.org/port/delve/details/

There is value in this build script (`_scripts/make.go`) as it makes certain determinations for how to build and test `delve` by making decisions from a number of environment-specific factors, such as the macOS version, the presence of certain system headers, and the version of Go being used. We can't just unilaterally do a `go install` when a user is installs `delve` through MacPorts because all these details matter (especially the macOS version the user might be on).

We need this change because the MacPorts project in general prefers to use checksummed source tarballs because tarballs can be verified, cached and mirrored.  Since we're building from an extracted tarball, we don't have git history, and so before this change, this script would panic when `git` crashes due to the absence of the `.git` directory.

herbygillot:
> > We can't just unilaterally do a `go install` when a user is installing `delve` through MacPorts because all these details matter (especially the macOS version the user is currently on).
> 
> You can unilaterally do `go install` (see [install instructions](https://github.com/go-delve/delve/tree/master/Documentation/installation)) and all those details don't matter at this point in time.

Alright, well that's good to know in that case. Thanks @aarzilli.

derekparker (approved):
LGTM

PR #2803, opened 2021-11-23
proc: fix signal handling during stepping

### LINKED ISSUE #2792: Debugger stops in single iteration cycle twice
```
Delve Debugger
Version: 1.7.2
Build: $Id: 5b6f24e7fbcad3fe3bc574d3763b2f20afa1d6a1 $
```

```
go version go1.17.1 darwin/arm64
```

BigSur 11.6, apple m1

4. What did you do?

Debugged
```
package main

func main() {
  for i := 0; i < 1; i++ {
    println(i) // break
  }
}
```

Debugger stops at breakpoint once, and exits on continue

6. What did you see instead?

Debugger stops at breakpoint, after continue it stops there again, after one more continue it exits.

Log: https://gist.github.com/nd/7239ae4f4f82f98ecfe04cf54a81f9f3

### LINKED ISSUE #2801: pkg/proc/native: can not resume execution from breakpoint set on instruction that dereferences a nil pointer
Given the following code:

```
TEXT ·asmFunc(SB),0,$0-16
	MOVQ arg+0(FP), AX
	MOVQ (AX), AX                // HERE
	MOVQ AX, ret+8(FP)
	RET
```

If a breakpoint is set on the line marked "HERE" it will not be possible to resume from it on linux/amd64 (possibly other architectures as well).

aarzilli:
Fix signal handling during thread single stepping so that signals that
are generated by executing the current instruction are immediately
propagated to the inferior, while signals other signals sent to the
thread are delayed until the full resume happens.

Fixes a bug where a breakpoint set on an instruction that causes a
SIGSEGV would make Delve hang and a bug where signals received during
single step would make it look like an instruction is executed twice.

Fixes #2801
Fixes #2792

### DIFF
--- a/_fixtures/asmnilptr/main.go
+++ b/_fixtures/asmnilptr/main.go
@@ -1,0 +1,9 @@
+package main
+
+import "fmt"
+
+func asmFunc(*int) int
+
+func main() {
+	fmt.Printf("%d\n", asmFunc(nil))
+}
--- a/_fixtures/asmnilptr/main_386.s
+++ b/_fixtures/asmnilptr/main_386.s
@@ -1,0 +1,7 @@
+#include "textflag.h"
+
+TEXT ·asmFunc(SB),0,$0-16
+	MOVL arg+0(FP), AX
+	MOVL (AX), AX
+	MOVL AX, ret+4(FP)
+	RET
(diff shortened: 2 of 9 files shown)


derekparker (approved):
LGTM

PR #2806, opened 2021-11-24
pkg/terminal: support ctrlz for shell job control

### LINKED ISSUE #2157: Ctrl-Z can't suspend dlv
related to https://github.com/peterh/liner/issues/131

derekparker:
This required forking peterh/liner under the go-delve org and using that
instead since upstream is not open to supporting this feature.

Fixes #2157


pkg/terminal/terminal.go
```diff
@@ -101,6 +101,7 @@ func New(client service.Client, conf *config.Config) *Term {
 		cmds:   cmds,
 		stdout: os.Stdout,
 	}
+	t.line.SetCtrlZStop(true)
```
aarzilli, line 104: given that this fork of liner exists solely for this feature why not just have it always behave as if ctrlZStop == true?

derekparker: 🤔 Good point. I suppose the only argument to keep it this way would be if others rely on this fork, but I don't feel strongly about that. I'll likely change this on Monday.

polinasok:
What will happen on Windows and other systems that don't treat Ctrl-Z as suspend?

pkg/terminal/terminal.go
polinasok: I assumed this was for consistency with `SetCtrlCAborts`. If we follow their convention then the owners of liner might reconsider integrating this into the main repo?

aarzilli: This feature was already rejected upstream.

polinasok: I saw that. But you never what the future holds. Hope dies last, right? :0
I think this is fine either way.

aarzilli: Well, we aren't scuplting this in granite either.

aarzilli:
This will need to take into account the fix in #2874

derekparker:
I've updated the liner dep to v1.2.2 but decided to keep the patch as is in case we ever need the functionality to toggle this at any point in the future. I believe it's better to have that option already instead of adding it back later if we decide we need it.

go.mod
```diff
@@ -18,6 +20,6 @@ require (
 	go.starlark.net v0.0.0-20200821142938-949cc6f4b097
 	golang.org/x/arch v0.0.0-20190927153633-4e8777c89be4
 	golang.org/x/sys v0.0.0-20211117180635-dee7805ff2e1
-	golang.org/x/tools v0.1.8
+	golang.org/x/tools v0.1.8-0.20211028023602-8de2a7fd1736
```
aarzilli, line 23: This actually downgraded x/tools to an older version which makes TestGeneratedDocs fail. I think also upgrading to 0.1.9 would be fine (or possibly pereferable).

derekparker: Done.

aarzilli (approved):
LGTM, not merging due to freeze.

PR #2807, opened 2021-11-26
proc: add dummy files to fix vendoring of Delve

aarzilli:
Add some dummy go files so that 'go mod vendor' works for modules that
require Delve, becuase directories that do not contain any go code will
not be vendored.


derekparker (approved):
LGTM

PR #2810, opened 2021-11-29
service/dap: move presentationHint to frame from source

suzmue:
This change visually indicates which frames are runtime code using frame.presentationHint. This removes "deemphasize" from the source presentation hint. "deemphasize" indicated that the source should be skipped in stepping, causing the frames to be collapsed. This was inconvenient for users if they wanted to inspect the runtime code that they were stepping through.

Updates golang/vscode-go#1916

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -3264,7 +3264,7 @@ func (s *Session) stacktrace(goroutineID int, g *proc.G) (string, error) {
 	}
 
 	var buf bytes.Buffer
-	fmt.Fprintln(&buf, "User Stack:")
+	fmt.Fprintln(&buf, "Stack:")
 	userLoc := g.UserCurrent()
 	userFuncPkg := fnPackageName(&userLoc)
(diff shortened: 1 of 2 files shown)


suzmue:
Before:
Screen Shot 2021-11-29 at 12 00 26 PM
After:
Screen Shot 2021-11-29 at 4 33 10 PM
Screen Shot 2021-11-29 at 4 33 13 PM

service/dap/server.go
```diff
@@ -3264,7 +3264,7 @@ func (s *Session) stacktrace(goroutineID int, g *proc.G) (string, error) {
 	}
 
 	var buf bytes.Buffer
-	fmt.Fprintln(&buf, "Stack:")
+	fmt.Fprintln(&buf, "User Stack:")
```
polinasok, line 3267: That's a bit misleadinig for system goroutines, isn't it?
Maybe prinit "Stack" and append "(excluding runtime frames)" if you exclude them?

suzmue: removed this part of the change for now.

polinasok (approved):

PR #2814, opened 2021-12-02
terminal: add transcript command

aarzilli:
Adds a transcript command that appends all command output to a file.

This command is equivalent to gdb's 'set logging'.

As part of this refactor the pkg/terminal commands to always write to a
io.Writer instead of using os.Stdout directly (through
fmt.Printf/fmt.Println).


derekparker (approved):
LGTM

PR #2815, opened 2021-12-05
proc/gdbserial: allow rewind to work after process exit with rr

aarzilli:
### proc/gdbserial: allow rewind to work after process exit with rr

It is sometimes useful to set breakpoints and rewind a terminated
process when using rr, for example if interested in the last execution
of some function.

RR will not allow a backward continue after the process exit packet has
been  sent, however rr will also generate a synthetic SIGKILL right
before process exit.

Treat this packet as a process exit and change some things so both
continuing backwards and setting breakpoints can be done, on recorded
targets, after process exit has been reported.

### service/debugger: fix bug internal err with Restart on recorded target

If Restart is called after a Continue and Rewind on a recorded target
that has already terminated it will return an internal error.

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -640,7 +640,10 @@ func (t *Target) SetWatchpoint(scope *EvalScope, expr string, wtype WatchType, c
 
 func (t *Target) setBreakpointInternal(addr uint64, kind BreakpointKind, wtype WatchType, cond ast.Expr) (*Breakpoint, error) {
 	if valid, err := t.Valid(); !valid {
-		return nil, err
+		recorded, _ := t.Recorded()
+		if !recorded {
+			return nil, err
+		}
 	}
 	bpmap := t.Breakpoints()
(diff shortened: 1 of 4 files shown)


hyangah:
What's the impact on DAP's rr support?
A dev workflow I expect is to use rr to catch a flaky program failure. This PR helps inspecting the issue after the program and test exits. 

On the other hand, if the failure is flaky so the program exits successfully most of time, I'd just want to restart debugging (F5 in vscode) until I hit the failure case eventually. Will users need to click an extra button to terminate the debug session before restarting? Or just issuing another `restart` request work already? @polinasok

aarzilli:
The Restart API call works as before (it doesn't care about the program having terminated anyway), but from what I can see the DAP restart request isn't implemented anyway.

polinasok:
DAP restart is not implemented. When the program exits, everything shuts down, unless you are in multi-client mode using headless dlv. In that case, we just disconnect from the client. So either way the DAP client won't still have a running session with controls.

derekparker (approved):
LGTM

PR #2816, opened 2021-12-07
cmd/dlv: make some defaults settable at build-time

briandealwis:
This PR make defaults for `--check-go-version` and `--only-same-user` settable at build-time.

Skaffold packages Delve into a container image which is used to enable debugging Go applications when deployed into Kubernetes.  We have to change two of Delve's defaults to make this work.
- `--check-go-version=false`: we slip-stream these images and so cannot break existing users who may be using older versions of the Go toolchain
- `--only-same-user=false`: in this environment, we use Kubernetes' port-forwarding to connect the Delve port to the localhost.  The connection in this case comes from the API Server which runs as root, and so Delve rejects the connection. Kubernetes already validates these connections at a higher level using RBAC.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -46,6 +46,10 @@ var (
 	apiVersion int
 	// acceptMulti allows multiple clients to connect to the same server
 	acceptMulti bool
+	// checkGoVersionDefault sets default for --check-go-version
+	checkGoVersionDefault = "true"
+	// checkLocalConnUserDefault sets default for --only-same-user
+	checkLocalConnUserDefault = "true"
 	// addr is the debugging server listen address.
 	addr string
 	// initFile is the path to initialization file.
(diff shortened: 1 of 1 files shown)


aarzilli:
What if instead we put:

```
var checkGoVersionDefault = true
var checkLocalConnUserDefault = true
```

in a file and then you replace it before building?

briandealwis:
Replacing the file itself is fragile: it will break when delve introduces other default values.

But I could maintain a patch against the file: that’s still better than patching the flag definitions.

The advantage with build-time variables is that they can be passed with `go install`. The disadvantage is that they are strings, and hence the need for `parseBool`.

aarzilli:
> Replacing the file itself is fragile: it will break when delve introduces other default values.

If no one else comes along and wants to soft-fork delve we won't put anything else in there.

> The advantage with build-time variables is that they can be passed with go install. The disadvantage is that they are strings, and hence the need for parseBool.

Yes, that's the disadvantage.

hyangah:
why not package a small shellscript that is named `dlv` but invokes the real `dlv` with the desired flag values?
It looks like dlv accepts flags placed before a command.

briandealwis:
Shell scripts can’t be used in shell-less (distroless-style) images.

hyangah:
Then write a small (go) wrapper program that does so.

briandealwis:
A Go wrapper is no better than the status quo!

I’ll close this PR as there doesn’t seem to be any appetite for this change.

PR #2823, opened 2021-12-10
terminal/command: fix up OoB in `deferred` command

kaddy-tom:
Prior to this, sending a `deferred` command with no arguments would immediately
crash the debug session.


aarzilli (approved):

aarzilli:
Thank you.

PR #2828, opened 2021-12-16
pkg/proc: set stop reason for step instruction

suzmue:
The stop reason is not updated when using step instruction. Update the stop reason to StepNextFinished.


aarzilli (approved):

derekparker (approved):
LGTM

PR #2832, opened 2021-12-20
service/dap: set min/max addresses for invalid instructions

suzmue:
The addresses need to be parseable as an int. The VS Code disassembly view uses a binary search to look for the appropriate address, so we set these to be the min value and the max value, so that the binary search will be effective.

Updates golang/vscode-go#1938

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -18,6 +18,7 @@ import (
 	"go/parser"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM

polinasok (approved):

derekparker (approved):
LGTM

PR #2833, opened 2021-12-21
Automatically disable breakpoints with hitcount conditions that will never be satisfied again

### LINKED ISSUE #2822: Automatically disable breakpoints with hitcount conditions that will never be satisfied again
A breakpoint that has `HitCond = "= 1"` is guaranteed to never be hit again after the first time, however the breakpoint will remain active and, if it happens to be on a codepath that is frequently executed, its existence will slow down the execution of the target program.

For HitCond conditions where we can prove that the breakpoint can no longer be hit (`=`, `<`, `<=`) we should disable and remove it as needed.

pippolo84:
To avoid slowing down the debugged process unnecessarily, we remove breakpoints with a hit condition that can no longer be hit again.

Fixes #2822


pippolo84:
Sorry for the failing test, I've missed it while trying it locally. I am working on it.

pippolo84:
@aarzilli I think I figured out what I missed.

The failed test `TestHitConditionBreakpoints` issues a series of `setBreakpoints` requests to edit the hit condition of the same breakpoint. It relies on the breakpoint to keep its TotalHitCount value between each "edit" operation.
With the changes I introduced, the breakpoint is removed as soon as its hit condition is no more satisfiable, therefore each subsequent `setBreakpoints` creates a new breakpoint with a TotalHitCount equal to 0, ultimately leading to the failure.

If this change in the behaviour of the hit condition breakpoints is not acceptable, a possible solution is to disable the breakpoint without removing it entirely, but TBH I'm unsure about how to implement that. Maybe a new BreakpointState?
Otherwise, I think it is sufficient to update the above test to reflect the new behaviour.

Please let me know what do you think, I would be happy to (try to) contribute either way. Thank you!

aarzilli:
The test failure is a legitimate bug from the point of view of a DAP client.
I think the easiest way to implement this feature is to do it in service/debugger/debugger.go:

* check the breakpoints that have unsatisfiable conditions before Command returns
* disable them using AmendBreakpoint

I think DAP will then call AmendBreakpoint when it changes a hitcount condition, re-enabling the ones that were disabled. There's a bug there, AmendBreakpoint doesn't restore the HitCount map and TotalHitCounts when it re-enables a disabled breakpoint, it should be done after the call to copyBreakpointInfo.

pippolo84:
Thank you for the advice, I'll rework the code as you suggested.

pippolo84:
Hi @aarzilli , I think the PR is ready to be reviewed.

service/test/integration1_test.go
```diff
@@ -371,6 +371,121 @@ func Test1ClientServer_clearBreakpoint(t *testing.T) {
 	})
 }
 
+func Test1ClientServer_disableHitCondLSSBreakpoint(t *testing.T) {
```
aarzilli, line 374: This file (and the withTestClient1 helper) tests the old API. We don't add new tests for the old API anymore, the tests are only here to ensure backwards compatibility with old clients. This test (and the other one should be in integration2_test and should use withTestClient2.
It should be easy to change.

aarzilli (commented):
This is ok, in general. Can you also rebase it on master so we can see if the tests pass?

pippolo84:
@aarzilli I moved the tests and rebased. Unfortunately, a couple of tests are still failing, but I can't figure out how they should be related to my changes. Besides, I see that `TestNoDebug_GoodExitStatus` is failing on Linux/amd64. It's the same environment I use for development, but I can't reproduce the failure.

pippolo84:
I also ran the test using go1.18beta1 in a ubuntu 20.04 vm and it is still passing.

polinasok:
> @aarzilli I moved the tests and rebased. Unfortunately, a couple of tests are still failing, but I can't figure out how they should be related to my changes. Besides, I see that `TestNoDebug_GoodExitStatus` is failing on Linux/amd64. It's the same environment I use for development, but I can't reproduce the failure.

`TestNoDebug_GoodExitStatus` is flaky due to a race and should be fixed by #2766 shortly.

aarzilli (approved):
LGTM, test failures are unrelated. Thank you.

derekparker (approved):
LGTM

PR #2834, opened 2021-12-21
proc: enable core dumping on windows

aarzilli:
Core dumping for Windows was implemented a while ago but never enabled.


derekparker (approved):
LGTM

PR #2842, opened 2021-12-25
Simplify BPF tracing implementation

chenhengqi:
* Remove all `__always_inline` from BPF program, this reduces size of `trace_bpfel_x86.o` file
* Replace `memcpy`(from `<string.h>`) with `__builtin_memcpy`
* Simplify `get_goroutine_id` implementation, we don't need a extra BPF ringbuf


derekparker:
These changes look nice but pulling your branch and running the EBPF test in `cmd/dlv` fails.

derekparker (changes requested):
These changes look nice but pulling your branch and running the EBPF test in cmd/dlv fails.

chenhengqi:
Will investigate.

pkg/proc/internal/ebpf/bpf/trace.bpf.c
```diff
@@ -244,4 +213,4 @@ int uprobe__dlv_trace(struct pt_regs *ctx) {
     return 0;
 }
 
-char _license[] SEC("license") = "GPL";
\ No newline at end of file
+char _license[] SEC("license") = "GPL";
```
aarzilli, line 216: Unrelated to this PR but I didn't notice it before. Should this really say GPL?

chenhengqi: `bpf_probe_read_user` is used in this BPF program, and this helper is GPL only.

https://github.com/torvalds/linux/blob/8bb7eca972ad531c9b149c0a51ab43a417385813/kernel/trace/bpf_trace.c#L169-L176

Please also refer to https://github.com/torvalds/linux/blob/master/Documentation/bpf/bpf_licensing.rst.

aarzilli: That says that to call GPL only functions the probe needs to be GPL compatible, which many licenses are.

chenhengqi: Right, but kernel only accept certain combinations.

https://github.com/torvalds/linux/blob/v5.15/include/linux/license.h#L5-L13

derekparker: We should rewrite it to be `Dual MIT/GPL`.

derekparker:
Closing as there have been no recent updates.

PR #2846, opened 2021-12-28
dap: support 'Env' attribute for launch requests

hyangah:
Env is applied in addition to the delve process environment
variables. The env setting is done by calling `os.Setenv`
as early as possible when a Launch request is received.

Prior discussion is in https://github.com/go-delve/delve/pull/2582

In Visual Studio Code, setting null for an environment variable
in launch.json or tasks.json indicates users want to unset
the environment variable. Support the behavior by accepting
nil value.

Updates golang/vscode-go#1978


aarzilli (approved):

service/dap/types.go
```diff
@@ -129,6 +129,14 @@ type LaunchConfig struct {
 	// directory.
 	DlvCwd string `json:"dlvCwd,omitempty"`
 
+	// Env specifies optional environment variables for Delve server
+	// beyond the environment variables Delve initially started with.
```
polinasok, line 133: s/beyond/in addition/

polinasok (approved):

service/dap/types.go
hyangah: done

derekparker (approved):
LGTM

Issue #2845, opened 2021-12-28
"dlv trace" with "--ebpf" broken?


shadyabhi:
### Version info
```
 ➤ dlv version
Delve Debugger
Version: 1.8.0
Build: $Id: 6a6c9c332d5354ddf1f8a2da3cc477bd18d2be53 $

➤ go version
go version go1.16.2 linux/amd64
```
Picked a simple HTTP server example from https://gobyexample.com/http-servers

```bash
$ go build -o server
$ ./server # starts with PID 15671
```
### dlv trace usage

When I use `dlv trace`, I'm unable to initialize it. 

```
➤ dlv trace -p 15671 --ebpf
could not retrieve CFA for current PC

```

Just a sidenote, even if I don't use `--ebf`, it still fails to trace the program. 

```
 ➤ dlv trace -p 15671
could not find function net.isZeros
```

derekparker:
Hm, not sure exactly what's happening here but I think it's related to misuse of the command. I think the command should print an error if you do not specify anything to trace.

What happens if you actually provide an argument?

derekparker:
I was able to reproduce the `could not retrieve CFA for current PC` issue. I'll dig into it.

PR #2848, opened 2021-12-30
cmd/dlv: require arg for trace subcommand

derekparker:


aarzilli:
IMO we should also change the lines that say `fmt.Fprintln(os.Stderr, err)` after setting breakpoints in `traceCmd`. It should tell the user which breakpoint failed and instead of quitting it should just log the error and continue. Otherwise the problem in #2845 will persist when `dlv trace --ebpf .` is used.

derekparker:
> IMO we should also change the lines that say `fmt.Fprintln(os.Stderr, err)` after setting breakpoints in `traceCmd`. It should tell the user which breakpoint failed and instead of quitting it should just log the error and continue. Otherwise the problem in #2845 will persist when `dlv trace --ebpf .` is used.

Done.

aarzilli (approved):

aarzilli:
Fixes #2845

PR #2852, opened 2021-12-31
proc: better handling of hardcoded breakpoints

aarzilli:
This commit improves the handling of hardcoded breakpoints in Delve.

A hardcoded breakpoint is a breakpoint instruction hardcoded in the
text of the program, for example through runtime.Breakpoint.

1. hardcoded breakpoints are now indicated by setting the breakpoint
field on any thread stopped by a hardcoded breakpoint
2. if multiple hardcoded breakpoints are hit during a single stop all
will be notified to the user.

3. a debugger breakpoint with an unmet condition can't hide a hardcoded
breakpoint anymore.

### DIFF
--- a/pkg/proc/native/threads_freebsd.go
+++ b/pkg/proc/native/threads_freebsd.go
@@ -125,3 +125,8 @@ func (t *nativeThread) ReadMemory(data []byte, addr uint64) (n int, err error) {
 func (t *nativeThread) withDebugRegisters(f func(*amd64util.DebugRegisters) error) error {
 	return proc.ErrHWBreakUnsupported
 }
+
+// SoftExc returns true if this thread received a software exception during the last resume.
+func (t *nativeThread) SoftExc() bool {
+	return false
+}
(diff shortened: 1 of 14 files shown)


pkg/proc/native/threads_freebsd.go
```diff
@@ -125,3 +125,8 @@ func (t *nativeThread) ReadMemory(data []byte, addr uint64) (n int, err error) {
 func (t *nativeThread) withDebugRegisters(f func(*amd64util.DebugRegisters) error) error {
 	return proc.ErrHWBreakUnsupported
 }
+
+// SoftExc returns true if this thread received a software exception during the last resume.
+func (t *nativeThread) SoftExc() bool {
+	return false
```
derekparker, line 131: Why hard code this to false?

aarzilli: The freebsd backend never kept track of which thread got a SIGTRAP, this is one more thing that will have to be done when it gets brought up to par with other backends. Always returning false is the safe thing to do for backends because handleHardcodedBreakpoint will always check trapthread.

derekparker (approved):
LGTM

PR #2853, opened 2022-01-02
proc: change UserCurrent to exclude internal and runtime/internal

aarzilli:
packages

Changes UserCurrent to exclude frames stopped inside the 'internal' and
'runtime/internal' packages of the standard library.

Before this change a goroutine blocked accepting or reading from a
socket would be reported as having a user current frame of:

internal/poll.runtime_pollWait

After this change accepting goroutines will be reported with a user
current frame of:

net.(*netFD).accept

and reading goroutines as:

net.(*netFD).read


derekparker (approved):
LGTM

PR #2856, opened 2022-01-04
Documentation: add DAP API page

polinasok:
I am still figuring out what to do with [api/ClientHowto.md](https://github.com/go-delve/delve/blob/master/Documentation/api/ClientHowto.md). It should either be moved as-is under [api/json-rpc](https://github.com/go-delve/delve/tree/master/Documentation/api/json-rpc) or we can keep the parts common to both JSON-RPC and DAP at the `api/` level and move JSON-RPC specific sections into [api/json-rpc/README.md](https://github.com/go-delve/delve/blob/master/Documentation/api/json-rpc/README.md).

Updates #1515

### DIFF
--- a/Documentation/api/dap/README.md
+++ b/Documentation/api/dap/README.md
@@ -1,0 +1,107 @@
+# DAP Interface
+
+Delve exposes a [DAP](https://microsoft.github.io/debug-adapter-protocol/overview) API interface.
+
+This interface is served over a streaming TCP socket, *not* over HTTP when the `dlv` server is run in one of the two headless modes:
+* `dlv dap` - starts a single-use DAP-only server that waits for a client to specify launch/attach configuration for starting the debug session.
+* `dlv --headless <command> <debugee>` - starts a general server, enters a debug session for the specified debuggee and waits for a [JSON-RPC](../json-rpc/README.md) or a [DAP](https://microsoft.github.io/debug-adapter-protocol/overview) client to begin interactive debugging. Can be used in multi-client mode with the following options:
+   *  `--accept-multiclient` - use to support connections from multiple clients
+   *  `--continue` - use to resume execution of the debuggee on session entry 
+
+The primary user of this mode is [VS Code Go](https://github.com/golang/vscode-go). Please see its 
+detailed [debugging documentation](https://github.com/golang/vscode-go/blob/master/docs/debugging.md) for additional information.
+
+## Debug Adapter Protocol
+
+[DAP](https://microsoft.github.io/debug-adapter-protocol/specification) is a general debugging protocol supported by many [tools](https://microsoft.github.io/debug-adapter-protocol/implementors/tools/) and [programming languages](https://microsoft.github.io/debug-adapter-protocol/implementors/adapters/). We tailored it to Go specifics, such as mapping [threads request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Threads) to communicate goroutines and [exceptionInfo request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_ExceptionInfo) to support panics and fatal errors.
+
+See [dap.Server.handleRequest](https://github.com/go-delve/delve/search?q=handleRequest) and capabilities set in [dap.Server.onInitializeRequest](https://github.com/go-delve/delve/search?q=onInitializeRequest) for an up-to-date list of supported requests and options.
+
+## Launch and Attach Configurations
+
+In addition to the general [DAP spec](https://microsoft.github.io/debug-adapter-protocol/specification), the server supports the following implementation-specific configuration options for starting the debug session:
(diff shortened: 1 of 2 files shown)


polinasok:
cc @suzmue @hyangah

Documentation/api/dap/README.md
```diff
@@ -0,0 +1,107 @@
+# DAP Interface
+
+Delve exposes a [DAP](https://microsoft.github.io/debug-adapter-protocol/overview) API interface.
+
+This interface is served over a streaming TCP socket, *not* over HTTP when the `dlv` server is run in one of the two headless modes:
```
aarzilli, line 5: When is it ever served over HTTP? Is there even a DAP over HTTP?

Documentation/api/dap/README.md
```diff
@@ -0,0 +1,107 @@
...
+* `dlv dap` - starts a single-use DAP-only server that waits for a client to specify launch/attach configuration for starting the debug session.
+* `dlv --headless <command> <debugee>` - starts a general server, enters a debug session for the specified debuggee and waits for a [JSON-RPC](../json-rpc/README.md) or a [DAP](https://microsoft.github.io/debug-adapter-protocol/overview) client to begin interactive debugging. Can be used in multi-client mode with the following options:
+   *  `--accept-multiclient` - use to support connections from multiple clients
+   *  `--continue` - use to resume execution of the debuggee on session entry
```
aarzilli, line 9: s/on session entry/on startup/ maybe? What does "session entry" mean?

Documentation/api/dap/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Single-Client Mode
+
+When used with `dlv dap` or `dlv --headless --accept-multiclient=false` (default), the DAP server will shut itself down at the end of the debug session, when the client sends a [disconnect request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect). If the debuggee was launched, it will be taken down as well. If the debugee was attached to, `terminateDebuggee` option will be respected.
+
+When the program terminates, we send a [terminated event](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Terminated), which is expected to trigger a [disconnect request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect) from the client for a session and a server shutdown. The [restart request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Restart) is not yet supported. 
+
+The server also shuts down in case of a client connection error or SIGTERM signal, but not SIGINT (Ctrl-C) signal, which gets forwarded to the foregrounded debuggee. In either case, the server takes down a launched process, but lets an attached process continue.
```
aarzilli, line 50: This is wrong, signals sent to the delve process are never forwarded to the target program. I think what you mean here is that pressing Ctrl-C on the terminal window will send the signal to the target process, instead of delve.

polinasok (commented):
PTAL

polinasok:
Any thoughts on what other sections this doc needs? Not necessarily for this PR, but a TODO for later?
- no debug mode?
- that most requests are handled synchronously and block?
- many commands not supported when running except for setting breakpoints, etc?

polinasok:
@puremourning Do you have any suggestions on how we can improve the new DAP API documentation for client developers?

puremourning:
the doc looks good.

From a user perspective the most important things are the launch and attach arguments, as that's what they actually have to configure for their projects, so thanks for documenting that so clearly!

The other thing I think that maybe needs to be said is that delve doesn't support "delayed" breakpoints. By which I mean if you set a breakpoint in a file `myplugin.go` and load that plugin built as a `.so`, then the breakpoint won't trigger (which is an improvement on the vscode-go approach of just crashing). The solution is to remove and re-add the breakpoint once the `.so` has been loaded.

Documentation/api/dap/README.md
aarzilli: The JSON-RPC documentation says that because JSON-RPC (version 1) can be served either over TCP or over HTTP.

Documentation/api/dap/README.md
aarzilli: I still think it's wrong, it should say something like: "but pressing Ctrl-C on the terminal where a headless instance of delve is running will send SIGINT to the debuggee. Also the "either case" in the following sentence refers to "client connection error or SIGTERM" but it could be interpreted to mean "when the server shuts down or the debuggee receives SIGINT", which is wrong.

aarzilli:
> Any thoughts on what other sections this doc needs? Not necessarily for this PR, but a TODO for later?
> 
>     * no debug mode?

I think that's documented in the DAP specification.

> 
>     * that most requests are handled synchronously and block?
> 
>     * many commands not supported when running except for setting breakpoints, etc?

I think these are reasonable to mention in the dap readme file.

suzmue:
Looks good to me. I agree with @aarzilli that documenting the request handling would be useful to have for clients, (TODO for later).

Documentation/api/dap/README.md
polinasok: Done

aarzilli (approved):

derekparker (approved):
LGTM

PR #2859, opened 2022-01-04
proc: interpret value of DW_AT_inline correctly

aarzilli:
All values greater than or equal to 1 indicate that the function has
been inlined.


derekparker (approved):
LGTM

PR #2865, opened 2022-01-05
proc,service: pretty print time.Time variables

### LINKED ISSUE #999: Add special formatting for time.Time

Delve Debugger
Version: 1.0.0-rc.2
go version go1.9 windows/amd64
Windows10 64bit
4. What did you do?

``` golang
package main

import (
    "fmt"
    "time"
)
func main() {
    t:=time.Now()
    fmt.Println(t)
}
```

and run 
```bash
dlv debug
b main.main
c
...
print t
```

I want to get string like below formatted time text
`2017-10-23 11:03:00 ...`
6. What did you see instead?

```
time.Time {
        wall: 13723543243807126424,
        ext: 4158612300,
        loc: *time.Location {
                name: "",
                zone: []time.zone len: 0, cap: 0, nil,
                tx: []time.zoneTrans len: 0, cap: 0, nil,
                cacheStart: 0,
                cacheEnd: 0,
                cacheZone: *time.zone nil,},}
```

How to get formatted time text in dlv debug mode?

aarzilli:
Fixes #999

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 	"unsafe"
 
 	"github.com/go-delve/delve/pkg/dwarf/godwarf"
@@ -1304,6 +1305,9 @@ func (v *Variable) loadValueInternal(recurseLevel int, cfg LoadConfig) {
 				v.Children[i].loadValueInternal(recurseLevel+1, cfg)
 			}
 		}
+		if t.Name == "time.Time" {
+			v.formatTime()
+		}
 
 	case reflect.Interface:
 		v.loadInterface(recurseLevel, true, cfg)
@@ -2454,3 +2458,63 @@ type constantValuesByValue []constantValue
 func (v constantValuesByValue) Len() int               { return len(v) }
 func (v constantValuesByValue) Less(i int, j int) bool { return v[i].value < v[j].value }
 func (v constantValuesByValue) Swap(i int, j int)      { v[i], v[j] = v[j], v[i] }
+
+const (
+	timeTimeWallHasMonotonicBit uint64 = (1 << 63) // hasMonotonic bit of time.Time.wall
+
+	//lint:ignore ST1011 addSeconds is the name of the relevant function
+	maxAddSeconds time.Duration = (time.Duration(^uint64(0)>>1) / time.Second) * time.Second // maximum number of seconds that can be added with (time.Time).Add, measured in nanoseconds
+
+	wallNsecShift = 30 // size of the nanoseconds field of time.Time.wall
+
+	unixTimestampOfWallEpoch = -2682288000 // number of seconds between the unix epoch and the epoch for time.Time.wall (1 jan 1885)
+)
+
+// formatTime writes formatted value of a time.Time to v.Value.
+// See $GOROOT/src/time/time.go for a description of time.Time internals.
+func (v *Variable) formatTime() {
+	wallv := v.fieldVariable("wall")
+	extv := v.fieldVariable("ext")
+	if wallv == nil || extv == nil || wallv.Unreadable != nil || extv.Unreadable != nil || wallv.Value == nil || extv.Value == nil {
+		return
+	}
+
+	var loc *time.Location
+
+	locv := v.fieldVariable("loc")
+	if locv != nil && locv.Unreadable == nil {
+		namev := locv.loadFieldNamed("name")
+		if namev != nil && namev.Unreadable == nil {
+			name := constant.StringVal(namev.Value)
+			loc, _ = time.LoadLocation(name)
+		}
+	}
+
+	wall, _ := constant.Uint64Val(wallv.Value)
+	ext, _ := constant.Int64Val(extv.Value)
+
+	hasMonotonic := (wall & timeTimeWallHasMonotonicBit) != 0
+	if hasMonotonic {
+		// the 33-bit field of wall holds a 33-bit unsigned wall
+		// seconds since Jan 1 year 1885, and ext holds a signed 64-bit monotonic
+		// clock reading, nanoseconds since process start
+		sec := int64(wall << 1 >> (wallNsecShift + 1)) // seconds since 1 Jan 1885
+		t := time.Unix(sec+unixTimestampOfWallEpoch, 0).UTC()
+		if loc != nil {
+			t = t.In(loc)
+		}
+		v.Value = constant.MakeString(fmt.Sprintf("%s, %+d", t.Format(time.RFC3339), ext))
+	} else {
+		// the full signed 64-bit wall seconds since Jan 1 year 1 is stored in ext
+		var t time.Time
+		for ext > int64(maxAddSeconds/time.Second) {
+			t = t.Add(maxAddSeconds)
+			ext -= int64(maxAddSeconds / time.Second)
+		}
+		t = t.Add(time.Duration(ext) * time.Second)
+		if loc != nil {
+			t = t.In(loc)
+		}
+		v.Value = constant.MakeString(t.Format(time.RFC3339))
+	}
+}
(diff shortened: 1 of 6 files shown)


pkg/proc/variables.go
```diff
@@ -2454,3 +2458,47 @@ type constantValuesByValue []constantValue
...
+func (v *Variable) formatTime() {
+	wallv := v.fieldVariable("wall")
+	extv := v.fieldVariable("ext")
+	if wallv == nil || extv == nil || wallv.Unreadable != nil || extv.Unreadable != nil || wallv.Value == nil || extv.Value == nil {
+		return
+	}
+
+	wall, _ := constant.Uint64Val(wallv.Value)
+	ext, _ := constant.Int64Val(extv.Value)
+	_ = ext
```
polinasok, line 2484: Why do you need this?

service/api/conversions.go
```diff
@@ -261,7 +261,7 @@ func VariableValueAsString(v *proc.Variable) string {
 		return convertFloatValue(v, 32)
 	case reflect.Float64:
 		return convertFloatValue(v, 64)
-	case reflect.String, reflect.Func:
+	case reflect.String, reflect.Func, reflect.Struct:
```
polinasok, line 264: All structs?

service/test/variables_test.go
```diff
@@ -840,6 +840,9 @@ func TestEvalExpression(t *testing.T) {
 		{"ni8 << 8", false, "0", "0", "int8", nil},
 		{"ni8 >> 1", false, "-3", "-3", "int8", nil},
 		{"bytearray[0] * bytearray[0]", false, "144", "144", "uint8", nil},
+
+		// pretty printing special types
+		{"tim1", false, `time.Time(1977-05-25T18:00:00Z)…`, `time.Time(1977-05-25T18:00:00Z)…`, "time.Time", nil},
```
polinasok, line 845: Is this always GMT+0000 timezone?
I can already imagine users wanting to see their timezone or be able to configure the timezone.

polinasok (commented):
Meta question: this PR manually reconstructs things from the low-level fields. I am assuming somewhere in the standard library this is done as well when time is converted to string. That logic is probably hidden somewhere inside of unexported methods and can't be reused, but but can they at least be referenced beyond only a pointer to time.go? That could be of help in sanity checking the logic.

aarzilli:
> Meta question: this PR manually reconstructs things from the low-level fields. I am assuming somewhere in the standard library this is done as well when time is converted to string. That logic is probably hidden somewhere inside of unexported methods and can't be reused, but but can they at least be referenced beyond only a pointer to time.go? That could be of help in sanity checking the logic.

The work here is done by copying a time.Time object read from the target into a time.Time object allocated by delve, time.go doesn't do this anywhere. The relevant comment is the one inside `type Time struct`.
We could format directly but we'd have to copy considerable more code.

mark-pictor-csec:
@polinasok wrote
> Meta question: this PR manually reconstructs things from the low-level fields. I am assuming somewhere in the standard library this is done as well when time is converted to string. That logic is probably hidden somewhere inside of unexported methods and can't be reused, but but can they at least be referenced beyond only a pointer to time.go? That could be of help in sanity checking the logic.

Kinda tangential: I'd like to see pretty printers as starlark scripts, contributed to (and ultimately maintained in parallel with) the packages they serve. Ideally with tests to ensure they stay in sync with the go code.

This thinking insipired by `gdb` pretty-printer scripts being embedded in the distributed library, for example `libc`. I don't know where those scripts live within the libc source, but for go it'd certainly make sense to bundle them inside the package - perhaps under `time/_dlv/` or `time/_prettyprint/` or so. The underscore is just a hint for humans: `no go code under here` - though if the standard library already has another mechanism to indicate that, we should use it instead. As for actually embedding the script within binaries, that seems less necessary for go than it is for languages like C.

aarzilli:
> I'd like to see pretty printers as starlark scripts, contributed to (and ultimately maintained in parallel with) the packages they serve

At the moment all starlark scripts are executed client side, meaning that if we did this all the pretty printing would only work on the command line client, which isn't that different from defining a `print-time` command and using it.

> I don't know where those scripts live within the libc source, but for go it'd certainly make sense to bundle them inside the package - perhaps under time/_dlv/ or time/_prettyprint/ or so

If we moved starlark evaluation to the backend we wouldn't necessarily be able to access those directories as delve's backend only has access to the executable it's debugging. Also directories that only contain non-go files are stripped from module archives, which means that you wouldn't have access to those files for any package except the packages of the main module and the standard library.

derekparker:
I'm catching up on the discussion still a bit here however I do like the approach of using the scripting interface for the pretty printing... I know it's a lot of work and I'm not suggesting it right now but I really think it's worth a conversation and consideration to take more advantage of the scripting support.

aarzilli:
Moving starlark evaluation to the backend is possible, there's going to be some pain points with making `dlv_command`, `read_file/write_file` and command registration work. Having starlark running in the server could also be used to create more complex breakpoint conditions.
Loading prettyprinters automatically from modules is something I'm not convinced is useful enough given the complications (also we could think about this later).
Before I port starlark to run server-side however I'd like to know three things:
1. what happens to the time.Time pretty printer? Does it become a hardcoded script?
2. are we going to add some starlark init file that gets loaded at startup by headless servers (at the moment the init file is *not* starlark and is only loaded by the frontend)
3. how does this get integrated with DAP?

hyangah:
Let's discuss the general scripting interface in a separate issue.

As @aarzilli pointed out, the implication with DAP is unclear. Starlark scripting is too powerful (It's delve RPC-based so you can do more than just pretty printing) for DAP to embrace it without careful design IMO. It's unclear to me, who's responsible for `time.Time` pretty printing support, and how trivial it is to write in a script (unexperienced users would attempt to utilize dlv `call` command-like features which may not be desirable at this moment).

For `gdb` support, go embeds the location of a very very basic script (from GOROOT) to the `.debug_gdb_scripts` section in the binary, but I am not sure if that approach can generally work for other packages without Go tool's support.

aarzilli:
> unexperienced users would attempt to utilize dlv call command-like features which may not be desirable at this moment

I wouldn't allow call and other commands that resume execution in this context.

derekparker:
> Let's discuss the general scripting interface in a separate issue.

Agreed, this will be a large undertaking if we happen to go for it and would need upfront design, discussion and consideration.

derekparker:
@aarzilli would you mind creating the issue for the scripting changes?

aarzilli:
> There's still a few outstanding questions from @polinasok

Oops I didn't see them.

pkg/proc/variables.go
aarzilli: Leftover, removed.

service/api/conversions.go
aarzilli: I don't think there's a reason to be more specific.

service/test/variables_test.go
aarzilli: I'm not sure we can do any better. We could look up the time.Time.Location.name field, and try to load it with time.LoadLocation, maybe that's good enough but if the program created the timezone using time.FixedZone the name could be anything.

aarzilli:
> @aarzilli would you mind creating the issue for the scripting changes?

Done: #2887

service/test/variables_test.go
derekparker: Possibly in the future we could let users specify this?

aarzilli: How would they specify it?

derekparker: I was thinking more along the lines of a custom format print command or something. Sorry, kinda thinking out loud without concrete ideas.

aarzilli: I've changed it to load the location, if there is one and print it. If we want to go with the scripting thing then picking the location becomes part of that.

derekparker:
Needs rebase. Also @polinasok can you do another review pass?

aarzilli:
Rebased.

polinasok (approved):
Thank you for adding the location logic.

derekparker (approved):
LGTM

PR #2875, opened 2022-01-17
proc: always load eh_frame section from executable

aarzilli:
Always load eh_frame section from the executable file instead of the
split debug info file. The eh_frame section is meant to be loaded in
memory along with the executable file so it will usually not be present
in a split debug info file (or, if it is present, it will be the wrong
one).


derekparker (approved):
LGTM

Issue #2072, opened 2020-06-03
Delve slow startup (60+ seconds in some cases) causes debugging to timeout when running in IntelliJ Idea.


ranganath42:
**Deleve version**
Version: 1.4.0
Build: 8bb93e9ae19589b7ee665225c2206ef118928ae9

**Go version**
go1.14.2 windows/amd64

**OS**
Edition: Windows 10 Pro
Version: 1909
Build: 18363.836

**IntelliJ IDEA**
2020.2 Ultimate EAP
2020.1.2 Ultimate 

**What did you do?**
As a domain user, run dlv.exe debugger from IntelliJ IDEA on a domain-joined computer that is currently not connected to the domain.

**What did you expect to see?**
Deleve debug session successfully starts.

**What did you see instead?**
Debug session gets timed out after 60 seconds waiting for Deleve to start up.

**Root cause**

- At startup, as a part of loading/saving the configuration, Deleve attempts to find out the home directory of the current user.
- This is done using `user.Current()` function in `os/user` package, which takes more than 60 seconds to return.
- This delay causes Deleve to get killed by the IntelliJ IDEA debugger connection timeout.

**Detailed description of the root cause**
- The `user.Current()` function in `os/user` package tries to obtain information about the current user in addition to the home directory.
- One of them is the fully qualified user name.
- Since this is a domain-joined computer, to obtain the full name, the function tries to do a name translation from SAM compatible name to display name using the `TranslateNameW` Windows system call.
- Since the computer is currently not connected to the domain, this syscall fails with 'domain could not be contacted'.
- When this attempt fails, the function then tries to query the full user name using the `NetUserGetInfo` syscall. This syscall is the culprit, taking 60 seconds upwards to return.

**Fix**
Since Deleve only needs to know about the home directory of the current user, the easiest way to fix is to use the `GetUserProfileDirectoryW` syscall. Like so.
```
func homeDir() (string, error) {
	t, err := syscall.OpenCurrentProcessToken()
	if err != nil {
		return "", err
	}
	defer t.Close()

	dir, err := t.GetUserProfileDirectory()
	if err != nil {
		return "", err
	}

	return dir, err
}
```
This function can them be used in place of the `user.Current()`. If this sounds reasonable, I will submit a pull request.

aarzilli:
I think you should report this problem to the go project, this seems like a very general problem that would affect much more than delve and it doesn't seem like a good idea to have a workaround in every program that wants the path to the home directory.

[...]

PR #2879, opened 2022-01-19
dlv can get homepath from env var to avoid slow startup

mainpart:
Delve is very slow on startup with windows active directory user. This PR addresses this issue
User able to specify homedir (that not to be calculated from os.User ```user.Current()``` ) directly and avoid long-time process of ```user.Current()``` discovery

Fixes:
https://github.com/go-delve/delve/issues/2072

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -307,9 +307,12 @@ func hasOldConfig() (bool, error) {
 	if os.Getenv("XDG_CONFIG_HOME") == "" && runtime.GOOS != "linux" {
 		return false, nil
 	}
-
-	userHomeDir := getUserHomeDir()
-
+	var userHomeDir string
+	if os.Getenv("HOMEDIR") != "" {
+		userHomeDir = os.Getenv("HOMEDIR")
+	} else {
+		userHomeDir = getUserHomeDir()
+	}
 	o := path.Join(userHomeDir, configDirHidden, configFile)
 	_, err := os.Stat(o)
 	if err != nil {


aarzilli:
As far as I can tell HOMEDIR is not standard, just something you came up with. I think it should be clear that this is a configuration variable for delve, and maybe just specify the path to the configuration directory of delve.

aarzilli:
PS. also it needs to be documented somewhere.

mainpart:
As far as i know delve does not have env variable specifying path to look to for ```.dlv\config.yml```. 
Instead of that it uses ```user.Current()``` to fetch user homedir and search there and that call under some circumstances under windows leads to timeouts and failures to use debugger (documented in issue)

So i have two options - to specify path where .dlv/config.yml should be searched or specify config file itself. 

Currently config file coded in const, and i thought i specifying a home dir for further config search could be easier. Do you think  that env var for config file will be more correct?
Also i could document if i know where to put docs.

aarzilli:
> Currently config file coded in const, and i thought i specifying a home dir for further config search could be easier. Do you think that env var for config file will be more correct?

It would be an environment variable configuring the directory where the config file is to be found.

>  Also i could document if i know where to put docs.

Documentation/cli/README.md where it describes how the configuration file is searched.

derekparker:
Hmm, in my opinion I think we should instead introduce a new env var that is documented as @aarzilli said named something like `DELVE_CONFIG_DIR` which points directly to where the Delve config lives, short circuiting the entire path construction code below.

mainpart:
There is one actualy. It's called ```XDG_CONFIG_HOME``` and it defines dir where ```/dlv/config.yml``` should be looked. I can't point it to ```$HOME``` since delve stores configs in ```$HOME/.dlv/config.yml```, not ```$HOME/dlv/config.yml``` but this solution works fine when config is in ```dlv``` dir. 

It also solves timeout issue with ```user.Current()``` (even pointing to non-exist dir) by escaping ```getUserHomeDir()``` call.

PR #2880, opened 2022-01-20
pkg/proc: handle double inlined calls

derekparker:
It's possible that an inlined function call also contains an inlined
subroutine. In this case we should also parse the children of
inlined calls to ensure we don't lose this information.


aarzilli (approved):
LGTM

PR #2881, opened 2022-01-25
Documentation: fix typo

chainhelen:
The ext should be '.star' instead of '.start' when starlark script.

Fix #1466


derekparker (approved):
LGTM

PR #2882, opened 2022-01-26
pkg/terminal: add '-clear' option to 'condition' command

chainhelen:
This change adds the 'clearcond' which will clear condtion on a breakpoint.


aarzilli:
I don't think this should be a separate command, either a flag to `condition`, like `condition -clear <breakpoint-name-or-id>` or just `condition` without the second argument.

chainhelen:
done

aarzilli (approved):
LGTM

aarzilli:
I must have missed this

PR #2885, opened 2022-01-29
Implement source listing from debuginfo

Foxboron:
This pull-request does  a few things.

We implement a BuilID call in the RPC so the client can figure out which binary it's working on.

We implement support for fetching file listing through `debuginfod-find` if they are not present locally.

Then lastly we try to remove a bit of the duplicated code around dealing with `debuginfod-find`.

I'm not sure how much testing you want around this, or if new package belongs under `pkg/proc/debuginfod`. Feel free to nitpick and I'll try fix up all the issues around this patch :)

### DIFF
--- a/Documentation/cli/starlark.md
+++ b/Documentation/cli/starlark.md
@@ -20,7 +20,7 @@ Function | API Call
 amend_breakpoint(Breakpoint) | Equivalent to API call [AmendBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AmendBreakpoint)
 ancestors(GoroutineID, NumAncestors, Depth) | Equivalent to API call [Ancestors](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Ancestors)
 attached_to_existing_process() | Equivalent to API call [AttachedToExistingProcess](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AttachedToExistingProcess)
-build_i_d() | Equivalent to API call [BuildID](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.BuildID)
+build_id() | Equivalent to API call [BuildID](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.BuildID)
 cancel_next() | Equivalent to API call [CancelNext](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.CancelNext)
 checkpoint(Where) | Equivalent to API call [Checkpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Checkpoint)
 clear_breakpoint(Id, Name) | Equivalent to API call [ClearBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.ClearBreakpoint)
--- a/Documentation/cli/starlark.md
+++ b/Documentation/cli/starlark.md
@@ -20,6 +20,7 @@ Function | API Call
 amend_breakpoint(Breakpoint) | Equivalent to API call [AmendBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AmendBreakpoint)
 ancestors(GoroutineID, NumAncestors, Depth) | Equivalent to API call [Ancestors](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Ancestors)
 attached_to_existing_process() | Equivalent to API call [AttachedToExistingProcess](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AttachedToExistingProcess)
+build_i_d() | Equivalent to API call [BuildID](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.BuildID)
 cancel_next() | Equivalent to API call [CancelNext](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.CancelNext)
 checkpoint(Where) | Equivalent to API call [Checkpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Checkpoint)
 clear_breakpoint(Id, Name) | Equivalent to API call [ClearBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.ClearBreakpoint)
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -52,6 +52,9 @@ type BinaryInfo struct {
 
 	debugInfoDirectories []string
 
+	// BuildID of this binary.
+	BuildID string
+
 	// Functions is a list of all DW_TAG_subprogram entries in debug_info, sorted by entry point
 	Functions []Function
 	// Sources is a list of all source files found in debug_line.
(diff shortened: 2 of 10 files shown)


pkg/terminal/command.go
```diff
@@ -2676,7 +2677,15 @@ func printfile(t *Term, filename string, line int, showArrow bool) error {
 		arrowLine = line
 	}
 
-	file, err := os.Open(t.substitutePath(filename))
+	var file *os.File
+	path := t.substitutePath(filename)
+	if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
```
aarzilli, line 2682: There's people around using old versions of go, let's just use `os.IsNotExist` here.

pkg/terminal/command.go
```diff
@@ -2676,7 +2677,15 @@ func printfile(t *Term, filename string, line int, showArrow bool) error {
...
+		path, err = debuginfod.GetSource(t.client.BuildID(), filename)
```
aarzilli, line 2683: Only overwrite path if there is no error, if there is an error let the code continue as normal. Should this use a cache or is debuginfod-find supposed to be fast?

Documentation/cli/starlark.md
```diff
@@ -20,6 +20,7 @@ Function | API Call
 amend_breakpoint(Breakpoint) | Equivalent to API call [AmendBreakpoint](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AmendBreakpoint)
 ancestors(GoroutineID, NumAncestors, Depth) | Equivalent to API call [Ancestors](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.Ancestors)
 attached_to_existing_process() | Equivalent to API call [AttachedToExistingProcess](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.AttachedToExistingProcess)
+build_i_d() | Equivalent to API call [BuildID](https://godoc.org/github.com/go-delve/delve/service/rpc2#RPCServer.BuildID)
```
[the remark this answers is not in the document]
Foxboron: Should I try improve it in this PR or?

pkg/terminal/command.go
Foxboron: go.mod explicitly says `go 1.16` so people on older versions can't use this regardless. I think it should be fine?

https://github.com/go-delve/delve/blob/master/go.mod#L3

Documentation/cli/starlark.md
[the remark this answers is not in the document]
Foxboron: I could also do `BuildId` and omit the requires patch? It doesn't have to be `BuildID` even though it feels a bit more natural.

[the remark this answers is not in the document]
aarzilli: BuildId wouldn't follow our naming convention, doing the fix in this PR is ok, whatever you prefer.

pkg/terminal/command.go
Foxboron: We overwrite path if the path we get isn't available. `debuginfod-find` is fast enough and has it's own cache.

pkg/terminal/command.go
aarzilli: The go version in go.mod is not the minimum supported version of Go, delve does build on versions older than 1.16. It's true that we haven't bothered to keep track of how low we can get but errors.Is vs os.IsNotExist is too small of a difference to compromise.

Foxboron: Ack, I'll change it :)

pkg/terminal/command.go
aarzilli: What I mean is, something like this:

```
path2, err := debuginfod.GetSource...
if err == nil {
    path = path2
}
```

so that the error message doesn't talk about debuginfod just because the source file can't be found and debuginfod-find can't be used.

Foxboron: Ah yes. We can do something like that so the errors are consistent.

Documentation/cli/starlark.md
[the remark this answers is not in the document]
Foxboron: I looked at it and deemed it a fairly non-trivial change frankly. It seems like the current code doesn't deal with `CreateEBPFTracepoint` and has workaround for this. I applied the same workaround for `BuildID` and hope that is good enough.

https://github.com/go-delve/delve/blob/master/_scripts/gen-starlark-bindings.go#L114

pkg/proc/bininfo.go
```diff
@@ -1393,8 +1355,37 @@ func (bi *BinaryInfo) loadSymbolName(image *Image, file *elf.File, wg *sync.Wait
...
+		bi.logger.Error("invalid build-id signature")
+		return
+	}
+
+	descBinary := make([]byte, bh.Descsz)
+	if err := binary.Read(br, binary.LittleEndian, descBinary); err != nil {
+		bi.logger.Errorf("can't read build-id desc: %w", err)
+		return
+	}
+	desc := hex.EncodeToString(descBinary)
```
aarzilli, line 1387: bi.BuildID = hex.EncodeToString(descBinary)

aarzilli (commented):
See build failure (bi.logger.Errorf doesn't support the %w verb)

Foxboron:
Woops, sorry. I should probably not be lazy with the tests :)

Amended the commit with the fixes.

aarzilli (approved):
LGTM

service/client.go
```diff
@@ -12,6 +12,9 @@ type Client interface {
 	// Returns the pid of the process we are debugging.
 	ProcessPid() int
 
+	// Returns the BuildID of the process' exectuable we are debugging.
```
stefanvanburen, line 15:
```suggestion
	// Returns the BuildID of the process' executable we are debugging.
```

pkg/proc/bininfo.go
```diff
@@ -52,6 +52,9 @@ type BinaryInfo struct {
 
 	debugInfoDirectories []string
 
+	// BuilID of this binary.
```
stefanvanburen, line 55:
```suggestion
	// BuildID of this binary.
```

😄 sorry for the drive-bys!

Foxboron: Thanks :)!

derekparker (approved):
LGTM

TobixDS:
Hi,
is it correct that the required ELF section is 
".note.gnu.build-id"?

If inspect any go executable (e.g. go itself - version go1.16.13 linux/amd64) I find only a section without a dash:

```
readelf -n /usr/local/go/bin/go

Displaying notes found in: .note.go.buildid
  Owner                Data size        Description
  Go                   0x00000053       GO BUILDID

```

Therefor delve 1.8.1 complains about the missing build id

```
dlv version
Delve Debugger
Version: 1.8.1
Build: $Id: d85f1f6b736db99a1c239c34fd4a081dcff08a3c $

dlv exec /usr/local/go/bin/go
2022-02-09T11:11:03Z error layer=debugger can't find build-id note on binary

dlv exec /go/bin/dlv         
2022-02-09T11:17:03Z error layer=debugger can't find build-id note on binary
```

Foxboron:
It is correct. Debuggers don't understand `.note.go.buildid` and what you are looking at is the fact that not all Go binaries has a build-id embedded. I'm not sure if there is any good fixes for this.

Foxboron:
Right, it's a feature of cgo and Go itself doesn't support it.

https://github.com/golang/go/issues/41004

https://github.com/kata-containers/kata-containers/issues/2587#issuecomment-915705151

PR #2893, opened 2022-02-05
proc: downgrade loadBuildID errors to warnings

aarzilli:
Most binaries do not have build-ids and it's fine, there's no reason to
report it as an error.


derekparker (approved):
LGTM

PR #2895, opened 2022-02-06
terminal/colorize: avoid emitting `Package` as illegal colorTok

hitzhangjie:
we have emitted token.Package, so  we may  avoid  emtting `Package` as illegal colorTok again.

Before there's a typo error here:
"Package" is typed as "Pacakge", if we change it to "Package" and run `dlv debug main.go` and `list main.go:0`, dlv will output "packagepackage main" instead of "package main".

So I think we should remove the `Package` from here.

### DIFF
--- a/pkg/terminal/colorize/colorize.go
+++ b/pkg/terminal/colorize/colorize.go
@@ -126,7 +126,7 @@ func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arr
 			emit(tokval.Interface().(token.Token), tokposval.Interface().(token.Pos), token.NoPos)
 		}
 
-		for _, kwname := range []string{"Case", "Begin", "Defer", "Pacakge", "For", "Func", "Go", "Interface", "Map", "Return", "Select", "Struct", "Switch"} {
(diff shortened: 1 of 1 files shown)


hitzhangjie:
Oh, this PR doesn't solve the problem. Here's an testcase:

```go
package main

// Vehical defines the vehical behavior
type Vehical interface{
	// Run vehical can run in a speed
	Run()
}

// BMWS1000RR defines the motocycle bmw s1000rr
type BMWS1000RR struct {
}

// Run bwm s1000rr run
func (a *BMWS1000RR) Run() {
	println("I can run at 300km/h")
}

func main() {
	var vehical = &BMWS1000RR{}
	vehical.Run()
}
```

run `dlv debug main.go`:

```bash
$ dlv debug main.go
(dlv) l main.go:0
Showing /Users/zhangjie/gotest/main.go:0 (PC: 0x0)
     1:	packagepackage main
     2:	
     3:	// Vehical defines the vehical behavior
     4:	type Vehical interface {
     5:		// Run vehical can run in a speed
(dlv) 
```

PR #2898, opened 2022-02-06
terminal/colorize: list command avoid printing multiple dup `package` string

### LINKED ISSUE #2896: `list` command prints multiple `package` string by multiple colorToks with type token.Illegal

Here is an test code (main.go):

```
package main

// Vehical defines the vehical behavior
type Vehical interface {
	// Run vehical can run in a speed
	Run()
}

// BMWS1000RR defines the motocycle bmw s1000rr
type BMWS1000RR struct {
}

// Run bwm s1000rr run
func (a *BMWS1000RR) Run() {
	println("I can run at 300km/h")
}

func main() {
	var vehical = &BMWS1000RR{}
	vehical.Run()
}
```

Then run `dlv debug main.go` and `list main.go:0`:

```
$ dlv debug main.go 
(dlv) l main.go:0
Showing /Users/zhangjie/gotest/main.go:0 (PC: 0x0)
     1:	packagepackage main
     2:	
     3:	// Vehical defines the vehical behavior
     4:	type Vehical interface {
     5:		// Run vehical can run in a speed
(dlv) exit
```

   Delve Debugger
   Version: 1.8.0
   Build: $Id: 6a6c9c332d5354ddf1f8a2da3cc477bd18d2be53 $
   
   I build from master branch, commit: e3f4c40b
    go version go1.16.3 darwin/amd64
    darwin and amd64
4. What did you do?
    I wanted to test the terminal/colorize.Print function to highlight the source code, but accidently found that the `list` command has problems. It should output `package` rather than `packagepackage`.
   `list main.go:0` outputs `package main` instead of `packagepackage main`.
6. What did you see instead?
   `list main.go:0` outputs `packagepackage main`.

---

Here  there's an typo error  in ast.Inspect(fset, func(*ast.Node)) to generate colorToks, "Pacakge" should be "Package". After I fixed this typo error, `list` will outputs `packagepackagepackage main`.

After checking the code of ast.Node and its implentions (especially ast.File and Ast.FuncType), I find that there's some occassions token's start position maybe token.NoPos. For example, the interface method `Vehical.Run()` method here. Dlv will generate a colorToken(token.Illegal, 0, 7) for this method. While linewriter outputs, it'll output a `package` string for this. ast.File will cause the same problem which outputs another `package` string.

If we check ast.File to emit the `package` colorTok and check ast.FuncType to differentiate the interface method and other functions. This problem may be solved.

hitzhangjie:
Explicitly check ast.File to emit the package colorTok, and check ast.FuncType to find the function name position accurately. By this way, dlv can avoid emitting the colorToks which of them may contains startPos=0 and endPos=7 which lets linewriter outputs dup `package` string in list command.

fix #2896


aarzilli (commented):
There is a problem with this patch, if you print your colorization example in full you'll see that the `func` keyword is never highlighted. I've taken a look at it seems to me that the problem is actually caused by some ast.FuncType nodes that have the Func pos set to token.NoPos, I think emit shouldn't be called when kwpos == token.NoPos.
Aside from removing the mispelled "Pacakge" from the list (not replacing it, just removing it), it should be the only thing that's needed to fix this problem.

Also your test is failing on Windows.

pkg/terminal/colorize/colorize_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	// ensures the AST analysis behaves as expected,
+	// please update `printed` if `terminal/colorize.Print` changes.
+	buf := &bytes.Buffer{}
+	colorize.Print(buf, "main.go", bytes.NewBuffer(dat), 1, 30, 10, colors)
+
+	colorize.Print(os.Stdout, "main.go", bytes.NewBuffer(dat), 1, 30, 10, colors)
+
+	if !reflect.DeepEqual(buf.Bytes(), printed) {
```
aarzilli, line 89: bytes.Equal or just `buf.String() == string(printed)`

pkg/terminal/colorize/colorize_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		t.Fatalf("read test fixture error: %v", err)
+	}
+
+	// ensures the AST analysis behaves as expected,
+	// please update `printed` if `terminal/colorize.Print` changes.
+	buf := &bytes.Buffer{}
+	colorize.Print(buf, "main.go", bytes.NewBuffer(dat), 1, 30, 10, colors)
+
+	colorize.Print(os.Stdout, "main.go", bytes.NewBuffer(dat), 1, 30, 10, colors)
```
aarzilli, line 87: Either remove this line or guard it with a check:

```
const printToStdout = false
if printToStdout {
    colorize.Print(os.Stdout, ...)
}
```

hitzhangjie:
> There is a problem with this patch, if you print your colorization example in full you'll see that the `func` keyword is never highlighted. I've taken a look at it seems to me that the problem is actually caused by some ast.FuncType nodes that have the Func pos set to token.NoPos, I think emit shouldn't be called when kwpos == token.NoPos. Aside from removing the mispelled "Pacakge" from the list (not replacing it, just removing it), it should be the only thing that's needed to fix this problem.
> 
> Also your test is failing on Windows.

@aarzilli Thanks very much.

I misunderstood part of the code. The code wants to highlight the keyword, like 'func'. Now my commited patch highlight the function name, though my patch's purpose is to remove the dup `package` string. 

I have tested just removing  the "Package" and that's not enough. It will generate a `colorTok{token.Illegal, 0, 7}` which will be output as `package` by linewriter. Because of this, I add the `switch case *ast.FuncType` branch. 

I'll modify it and test it later.

pkg/terminal/colorize/colorize.go
```diff
@@ -126,7 +141,7 @@ func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arr
 			emit(tokval.Interface().(token.Token), tokposval.Interface().(token.Pos), token.NoPos)
 		}
 
-		for _, kwname := range []string{"Case", "Begin", "Defer", "Pacakge", "For", "Func", "Go", "Interface", "Map", "Return", "Select", "Struct", "Switch"} {
+		for _, kwname := range []string{"Case", "Begin", "Defer", "Package", "For", "Func", "Go", "Interface", "Map", "Return", "Select", "Struct", "Switch"} {
```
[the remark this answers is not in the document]
hitzhangjie: Remove the  "Package".

done

pkg/terminal/colorize/colorize.go
```diff
@@ -78,20 +79,34 @@ func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arr
...
 			return true
 		}
 
-		switch n := n.(type) {
+		switch n := node.(type) {
+		case *ast.File:
+			emit(token.PACKAGE, f.Package, token.NoPos)
+			return true
+		case *ast.FuncType:
+			// this implies `node` is a method of interface
```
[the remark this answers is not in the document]
hitzhangjie: I may misunderstand `ast.FuncType.Func==token.NoPos`, before I thought only a method of interface could satisfy this condition.

Here we should highlight the keyword `func` ranther than the function name. So when `n.Pos() == token.NoPos` we just return, otherwise `emits(token.Func, n.Pos, token.NoPos)` to hightlight the keyword `func`.

pkg/terminal/colorize/colorize_test.go
hitzhangjie: done

pkg/terminal/colorize/colorize.go
```diff
@@ -78,20 +79,34 @@ func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arr
...
 	for _, cgrp := range f.Comments {
 		for _, cmnt := range cgrp.List {
 			emit(token.COMMENT, cmnt.Pos(), cmnt.End())
 		}
 	}
 
-	ast.Inspect(f, func(n ast.Node) bool {
-		if n == nil {
+	ast.Inspect(f, func(node ast.Node) bool {
+		if node == nil {
```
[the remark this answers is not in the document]
hitzhangjie: done

pkg/terminal/colorize/colorize_test.go
hitzhangjie: done

hitzhangjie:
The Aggregator (Delve) -- TeamCity build failed for:
- linux/amd64/go1.17
- windows/amd64/go1.18
- windows/amd64/tip

I'll check go1.17/go1.18. Please tell me what does the "tip" means?

aarzilli:
> The Aggregator (Delve) -- TeamCity build failed for:
> 
>     * linux/amd64/go1.17
> 
>     * windows/amd64/go1.18
> 
>     * windows/amd64/tip
> 
> 
> I'll check go1.17/go1.18. Please tell me what does the "tip" mean?

The failure on linux is probably unrelated, tip just means go built at master.

pkg/terminal/colorize/colorize.go
```diff
@@ -92,6 +90,15 @@ func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arr
 		}
 
 		switch n := n.(type) {
+		case *ast.File:
+			emit(token.PACKAGE, f.Package, token.NoPos)
+			return true
+		case *ast.FuncType:
```
aarzilli, line 96: Remove this case and change the code:

```
				kwpos, ok := kwposval.Interface().(token.Pos)
				if ok {
					emit(token.ILLEGAL, kwpos, token.NoPos)
}
```

to say `if ok && kwpos != token.NoPos`

hitzhangjie: case *ast.FuncType is removed, and the code mentioned above is changed.

done

aarzilli:
Note that the test still fails on Windows.

aarzilli:
> Note that the test still fails on Windows.

My guess is that you are reading the input from a file on disk and git will convert line endings to crlf on windows, which will make the output not match the recorded output. 

Also "vehical" is spelled "vehicle".

hitzhangjie:
> > Note that the test still fails on Windows.
> 
> My guess is that you are reading the input from a file on disk and git will convert line endings to crlf on windows, which will make the output not match the recorded output.
> 
> Also "vehical" is spelled "vehicle".



> > Note that the test still fails on Windows.
> 
> My guess is that you are reading the input from a file on disk and git will convert line endings to crlf on windows, which will make the output not match the recorded output.
> 
> Also "vehical" is spelled "vehicle".

I just installed a windows 10 in my parallels desktop. I'll test this problem.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

Issue #2900, opened 2022-02-08
Delve panic when starting


vatzmehta:
1. What version of Delve are you using (`dlv version`)?
Version: 1.8.1
2. What version of Go are you using? (`go version`)?
go version go1.13.11 linux/amd64
3. What did you do?
Ran `dlv attach 3054 --listen=:8080 --headless --api-version=2 --accept-multiclient --check-go-version=false`
5. What did you expect to see?
Debugger to start
4. What did you see instead?
```API server listening at: [::]:3000
2022-02-08T17:13:36+05:30 warning layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
2022-02-08T17:13:36+05:30 error layer=debugger can't find build-id note on binary
2022-02-08T17:13:37+05:30 error layer=debugger can't find build-id note on binary
panic: runtime error: slice bounds out of range [:2] with length 0

goroutine 1 [running]:
github.com/go-delve/delve/pkg/proc.(*BinaryInfo).openSeparateDebugInfo(0xc0000c2c80, 0xc00518e000, 0xc005166d20, 0xc000162c80, 0x1, 0x1, 0x0, 0x203001, 0x8, 0x1007fe9897a4be0)
	/go/pkg/mod/github.com/go-delve/delve@v1.8.1/pkg/proc/bininfo.go:1213 +0xa5e
github.com/go-delve/delve/pkg/proc.loadBinaryInfoElf(0xc0000c2c80, 0xc00518e000, 0xc005d1f280, 0x18, 0x7fd7e7537000, 0xc005ca1290, 0x445b1e, 0xc00517eb20)
	/go/pkg/mod/github.com/go-delve/delve@v1.8.1/pkg/proc/bininfo.go:1297 +0x84a
github.com/go-delve/delve/pkg/proc.loadBinaryInfo(0xc0000c2c80, 0xc00518e000, 0xc005d1f280, 0x18, 0x7fd7e7537000, 0x0, 0x0)
	/go/pkg/mod/github.com/go-delve/delve@v1.8.1/pkg/proc/bininfo.go:671 +0x106
github.com/go-delve/delve/pkg/proc.(*BinaryInfo).AddImage(0xc0000c2c80, 0xc005d1f280, 0x18, 0x7fd7e7537000, 0x0, 0x0)
	/go/pkg/mod/github.com/go-delve/delve@v1.8.1/pkg/proc/bininfo.go:831 +0x2c4
[...]

PR #2901, opened 2022-02-08
proc: fix crash trying to open separate debug info

aarzilli:
If debugInfoDirectories is set, the executable does not have debug info
and build-id is not set we should not crash.


aarzilli:
Fixes #2900

derekparker (approved):
LGTM

Issue #2904, opened 2022-02-10
go install dlv@master is broken due to the use of 'replace'


hyangah:
We observed breakage in dlv installation (https://github.com/microsoft/vscode-dev-containers/issues/1300)
caused by the use of 'replace' directive.

I understand why this replacement is necessary, however, that
is not compatible with `go install` command. So, this need to 
be addressed before the next release.

```
% go install github.com/go-delve/delve/cmd/dlv@master  
go: downloading github.com/go-delve/delve v1.8.2-0.20220210175055-dee4437bd088
go: github.com/go-delve/delve/cmd/dlv@master (in github.com/go-delve/delve@v1.8.2-0.20220210175055-dee4437bd088):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

hyangah:
Since `replace` is not an option, I think the fix is like

1. make `github.com/go-delve/liner` to be the real fork (i.e. change module name in `go.mod` to be `github.com/go-delve/liner` and release it.*
2. replace all references to github.com/peterh/liner with github.com/go-delve/liner
3. remove replace statement in go.mod here and 
4. `go mod tidy` and `go mod vendor` 

* I tried to skip (1) and went to (2), but `go mod tidy` keeps insisting to put back `github.com/peterh/liner`

:-(

PR #2905, opened 2022-02-13
go.mod: require go-delve/liner instead of peterh/liner

hyangah:
Use of `replace` in go.mod breaks delve installation using
`go install`. (see https://github.com/golang/go/issues/40276)
Workaround this limitation by explicitly require the fork
github.com/go-delve/liner.

go-delve/liner@v1.2.2-1 already has go.mod module name fixed
to be github.com/go-delve/liner. go-delve/liner@v1.2.2-1 also
has complete go.mod that made github.com/mattn/go-runewidth
declared as an implicit dependency and get pulled into the vendor
directory correctly.
 
Fixes go-delve/delve#2904


aarzilli (approved):

hyangah:
cc @derekparker

aarzilli:
Is vscode-go still installing `@master`?

hyangah:
Not any more, but some systems depending on github/vscode dev containers still install dlv from master. Fixes in github&vscode were committed, but not released. So some users in downstream are stuck. :-(

hyangah:
Thanks a lot!

chattarajoy:
please make changes to the name of module in the fork, this is causing issues when downloading the latest version:

image

Issue #1653, opened 2019-08-08
Breakpoints in plugins are not working if set before plugin is loaded


akamensky:
Please answer the following before submitting your issue:

1. `Version: 1.2.0 Build: dd4fd5dc9cf5d45092f97f2ba376f3e69d7a5a8e`
2. What version of Go are you using? `go version go1.12.7 linux/amd64`
3. What operating system and processor architecture are you using? `linux x86_64`
4. What did you do? 

Attempt to debug plugin from Goland. Using i.e. https://github.com/vladimirvivien/go-plugin-example.git (compile all plugins with `-gcflags='all=-N -l'`). Place breakpoints inside functions of plugin and inside main application functions.

5. What did you expect to see?

Breakpoints of main application work. Breakpoints inside plugins work.

6. What did you see instead?

Breakpoints of main application work. Plugin breakpoints are ignored.

derekparker:
Is there an error or any odd output when you put a breakpoint in the plugin code?

akamensky:
> Is there an error or any odd output when you put a breakpoint in the plugin code?

No errors, the breakpoints are just skipped as if non-existent

![Screenshot from 2019-08-13 12-25-22](https://user-images.githubusercontent.com/985805/62915050-8c34fa80-bdc5-11e9-9ba6-fe8ba0cec8ff.png)

aarzilli:
I'd like to see output of `--log --log-output=rpc`. I'm guessing you are setting all breakpoints before starting the program, the ones on plugins fail (because the plugins aren't loaded when the program starts) and goland isn't telling you. We should have delayed breakpoints, but we don't.

akamensky:
@aarzilli you are right I am setting breakpoints before  starting the application.

Are the arguments you mention for dlv?

aarzilli:
Yes. IIRC there's an option to get those logs from GoLand, I think maybe somewhere under the help menu?

akamensky:
I will get the log later (did not find it this option anywhere in GoLand btw). Meanwhile, it appears that even after the module loaded I am unable to set breakpoints via IDE, I've set a breakpoint right before call to `Greet()` and when I hit it, navigate to the source file and set breakpoint there, this normally works with non-plugin code, but here it appears as disabled/inactive breakpoint:

[...]

PR #2913, opened 2022-02-24
proc,debugger: move breakpoint ID counter to service/debugger

aarzilli:
Moves breakpoindIDCounter out of BreakpointsMap and into
service/debugger.Debuggger to simplify proc.Target's API and aid with
implementing fork follow mode, where we'll have to debug multiple
processes simultaneously.


derekparker (approved):
LGTM

aarzilli:
This is the first in a (currently incomplete) series of patches aimed at supporting #1653 and #2551. An unfortunate amount of refactoring is necessary to implement those because they violate a number of assumptions in our architecture, particularly around breakpoints:

1. it is no longer the case that it is possible to set a breakpoint by simply specifying a memory address
2. we can't discard the "command" that created a breakpoint immediately after using it, because it might be that the breakpoint will need to be enabled on another process later
3. the debugger object needs to hold a list of target processes instead of one and they need to be stopped and resumed simultaneously, this group of processes needs to be plumbed all the way down to the backend beacuse they are all controlled by the same ptrace thread.

aarzilli:
The darwin/amd64 builder sometimes fails with "An error occurred when generating and installing a new certificate". I wonder why. Especially, I wonder why it doesn't happen sistematically or on arm64.

PR #2921, opened 2022-03-10
proc: move resume notify and manual stop handling to Target

aarzilli:
Moves handling of ResumeNotify and manualStopRequested to Target instead of the backends

Updates #2551


derekparker (approved):
LGTM

PR #2940, opened 2022-03-22
service/dap: support terminateDebuggee option for attach only

polinasok:
I realized that we end up with Stop and Disconnect options both for Attach (correct) and Launch (incorrect). So change the code to request this capability in Attach cases only. This also sets us up for when suspendDebuggee option is added, which should only apply to remote attach (microsoft/vscode#134412). Turns out this currently makes no difference in the UI - filed an issue (microsoft/vscode#145650).


aarzilli (approved):
LGTM

PR #2946, opened 2022-03-26
proc,service: represent logical breakpoints explicitly

aarzilli:
Adds a LogicalBreakpoint type to represent logical breakpoints
explicitly. Until now logical breakpoints were constructed implicitly
by grouping physical breakpoints together by their LogicalID.

Having logical breakpoints represented explicitly allows for a simpler
implementation of disabled breakpoints, as well as allowing a simple
implementation of delayed breakpoints (#1653, #2551) and in general of
breakpoints spanning multiple processes if we implement debugging
process trees (#2551).

Updates #1653
Updates #2551

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -47,7 +47,6 @@ type Breakpoint struct {
 
 	Addr         uint64 // Address breakpoint is set for.
 	OriginalData []byte // If software breakpoint, the data we replace with breakpoint instruction.
-	Name         string // User defined name of the breakpoint
 
 	WatchExpr     string
 	WatchType     WatchType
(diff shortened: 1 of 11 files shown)


derekparker:
Please rebase.

pkg/proc/breakpoints.go
```diff
@@ -925,3 +941,36 @@ func returnInfoError(descr string, err error, mem MemoryReadWriter) []*Variable
 	v.Name = "return value read error"
 	return []*Variable{v}
 }
+
+// LogicalBreakpoint represents a breakpoint set by a user.
```
derekparker, line 945: Can you add some more documentation here around what a logical breakpoint represents (in terms of multiple physical, etc...)

derekparker (approved):
LGTM

PR #2959, opened 2022-04-06
proc: skip autogenerated functions correctly

### LINKED ISSUE #2948: Debugger steps into generated code

```
Delve Debugger
Version: 1.8.2
Build: fbfed8f8d90a58a6108daefd0f545f5f7f3beadf
```

```
go version go1.18 windows/amd64
```

windows/amd64

4. What did you do?

debugged the program, once the breakpoint is reached ran the step command:

```
package main

type T struct {
}

func main() {
	t := T{}
	f1 := t.m1
	println(f1(1)) //break
}

func (t T) m1(x int) int { //expected step into position
	return x + 1
}
```

Step in succeeds, debugger is in the expected position 

6. What did you see instead?

Step in succeeds, debugger is in autogenerated sources, `bt` shows

```
0  0x000000000008f006 in main.T.m1-fm
   at <autogenerated>:1
1  0x000000000008ef88 in main.main
   at w:/tmp/dbg.go:9
2  0x0000000000065088 in runtime.main
   at c:/users/nd/go/go1.18/src/runtime/proc.go:250
3  0x00000000000898c1 in runtime.goexit
   at c:/users/nd/go/go1.18/src/runtime/asm_amd64.s:1571
```

It looks like fix implemented in 431dea7ee6593377b75dfba09951b922848b7812 doesn't work or is not applied to method call on non-interface variable.

aarzilli:
Adjust heuristic used to skip autogenerated functions so that it works
correctly with the new naming scheme in go1.18.

Fixes #2948


derekparker (approved):
LGTM

PR #2960, opened 2022-04-06
gobuild: drop support for building on Go < 1.10

aarzilli:
Remove code to build on old versions of Go that didn't support the all=
query for -gcflags.


derekparker (approved):
LGTM

PR #2961, opened 2022-04-06
proc: temporarily disable TestCgoStacktrace on darwin/arm64/go1.19

aarzilli:
Something changed in go1.19 that makes TestCgoStacktrace fail on
darwin/arm64, disable the test for now.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3374,6 +3374,9 @@ func TestCgoStacktrace(t *testing.T) {
 		if ver.Major > 0 && !ver.AfterOrEqual(goversion.GoVersion{Major: 1, Minor: 8, Rev: -1}) {
 			t.Skip("disabled on macOS with go before version 1.8")
 		}
+		if runtime.GOARCH == "arm64" && (ver.Major < 0 || ver.AfterOrEqual(goversion.GoVersion{Major: 1, Minor: 19, Rev: -1})) {
+			t.Skip("disable on macOS/arm64 with Go after version 1.19")
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


aarzilli:
We'll see about making this work when 1.19 is closer.

aarzilli:
I don't know how to do it.

derekparker:
I think this line is supposed to do it but clearly isn't working: https://github.com/go-delve/delve/blob/master/.teamcity/settings.kts#L156.

@artspb thoughts?

artspb:
At the first glance, builds are failing due to a non-zero exit code returned by `make test`. We could ignore it, too, but I'm not sure that it's the right approach as _tip_ will almost always be green after that. Perhaps, we could just exclude _tip_ from the aggregator. @nd Thoughts?

derekparker:
@artspb Is there any way to still run the test and get a report of pass / fail but not fail the overall build?

nd:
Sorry for late reply. I think failing build in a case of failing tests + if number of tests is less than 1 should catch most of the problems, so it seems fine to ignore the exit code as it duplicates failed tests info.

One can mute a test with either a manual resolve (unmute) or an automatic resolve once the test is successful. The latter seems to be more appropriate in case of TestCgoStacktrace and go1.19. This is how to find in UI:

image

Let me know if I should make a pull request disabling failure on non-zero exit code.

aarzilli:
I guess we also need to add an 'exit 0' at the end of test_mac.sh?

aarzilli:
IIRC there used to be a probem on windows where checking the exit result was necessary.

PR #2964, opened 2022-04-07
debugger: better error message for GOOS/GOARCH mismatch

polinasok:
Fixes https://github.com/golang/vscode-go/issues/114

### DIFF
--- a/service/debugger/debugger_unix.go
+++ b/service/debugger/debugger_unix.go
@@ -37,7 +37,7 @@ func verifyBinaryFormat(exePath string) error {
 		panic("attempting to open file Delve cannot parse")
 	}
 	if err != nil {
-		return api.ErrNotExecutable
+		return api.ErrNotExecutableSysMismatch
 	}
 	return nil
 }
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -14,6 +14,9 @@ import (
 // ErrNotExecutable is an error returned when trying
 // to debug a non-executable file.
 var ErrNotExecutable = errors.New("not an executable file")
+// ErrNotExecutableSysMismatch is an error returned when trying
+// to debug a file with GOOS/GOARCH settings incompaatible with the host system.
+var ErrNotExecutableSysMismatch = errors.New("not an executable file - incompatible GOOS/GOARCH")
 
 // DebuggerState represents the current context of the debugger.
 type DebuggerState struct {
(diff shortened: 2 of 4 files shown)


service/debugger/debugger_unix.go
```diff
@@ -37,7 +37,7 @@ func verifyBinaryFormat(exePath string) error {
 		panic("attempting to open file Delve cannot parse")
 	}
 	if err != nil {
-		return api.ErrNotExecutable
+		return api.ErrNotExecutableSysMismatch
```
aarzilli, line 40: This is wrong, the code could get here from `dlv exec exe` and the error wouldn't have anything to do with GOOS/GOARCH, or it could come here from `dlv debug` run inside a library and the error would also have nothing to do with GOOS/GOARCH.

aarzilli: I think at that point we are just saying the same thing twice. In the case of packages that don't have a main it isn't even right, it doesn't have anything to do with the host system, it just isn't an executable. I think you should:
* check if ExecuteKind is ExecutingGeneratedFile or ExecutingGeneratedTest 
* check that it isn't a go object archive file (or whatever it's called the output of `go build` when there's no main)
* check that the env variables GOOS/GOARCH do not match runtime.GOOS/runtime.GOARCH
if all checks pass then return an error that talks about GOOS/GOARCH.

IMO it isn't worth the effort but if you want to do it, sure.

aarzilli: PS.
1. it is probably better to check if something is an executable (PE/ELF/Mach-O) rather than checking for the archive format of go, since the latter probably has changed multiple times and isn't very well documented
2. I just noticed that we are leaking a file descriptor in verifyBinaryFormat, it gets closed by the finalizer but I wonder if that's what was causing the flakyness with deleting generated executable files on windows

polinasok: Eh, this is more intricate than I have time for right now. [firelizzard18](https://github.com/firelizzard18), do you want to give this a go?

derekparker:
Should this be closed?

polinasok:
Yes, I will close. I was thinking of just adding a settings info dump as part of failing, so users can notice things that are off on their own. But that deserves a separate PR. Do you have a preference whether this is done only with dlv-dap or with any mode?

PR #2972, opened 2022-04-18
proc: unlock OS thread on exit from handlePtraceFuncs

4a6f656c:
On FreeBSD and OpenBSD, the use of runtime.LockOSThread is resulting in segfaults
within the Go runtime (see https://github.com/golang/go/issues/52394) - while it
should not be necessary, calling runtime.UnlockOSThread upon exit from
handlePtraceFuncs avoids this issue and allows the tests to run correctly.

### DIFF
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -271,6 +271,13 @@ func (dbp *nativeProcess) handlePtraceFuncs() {
 	// all commands after PTRACE_ATTACH to come from the same thread.
 	runtime.LockOSThread()
 
+	// Leaving the OS thread locked currently leads to segfaults in the
+	// Go runtime while running on FreeBSD and OpenBSD:
+	//   https://github.com/golang/go/issues/52394
+	if runtime.GOOS == "freebsd" || runtime.GOOS == "openbsd" {
+		defer runtime.UnlockOSThread()
+	}
+
 	for fn := range dbp.ptraceChan {
 		fn()
 		dbp.ptraceDoneChan <- nil


4a6f656c:
I also do not believe it is necessary to call `ptrace(2)` from the same OS thread on both of these platforms - another option is to avoid using this code entirely and call ptrace functions directly instead.

pkg/proc/native/proc.go
```diff
@@ -270,6 +270,7 @@ func (dbp *nativeProcess) handlePtraceFuncs() {
 	// while invoking the ptrace(2) syscall. This is due to the fact that ptrace(2) expects
 	// all commands after PTRACE_ATTACH to come from the same thread.
 	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
aarzilli, line 273: I think I like that the thread is not reused after we're done with it, I would prefer this to be:

```
if runtime.GOOS == "openbsd" { 
     // workaround for https://github.com/golang/go/issues/52394
    defer runtime.UnlockOSThread()
}
```

4a6f656c: Sure, done. Alternatively, as mentioned above we could also avoid calling `runtime.LockOSThread` in the first place on these platforms.

aarzilli (approved):
LGTM, if you think not calling LockOSThread is better on those architectures that's also fine (actually in that case it would be even better to have execPtraceFunc simply call fn, you'd get a substantial performance boost out of it).

4a6f656c:
> LGTM, if you think not calling LockOSThread is better on those architectures that's also fine (actually in that case it would be even better to have execPtraceFunc simply call fn, you'd get a substantial performance boost out of it).

Let's land this as is - I'll retest and send a follow up (this way if an issue is found later, the conversion to calling ptrace functions directly can be reverted and this fix will remain).

derekparker (approved):
LGTM

PR #2973, opened 2022-04-18
_scripts/make: do not exit if git SHA ID can not be retrieved

aarzilli:
Something changed in TeamCity's infrastructure that makes 'git
rev-parse HEAD' fail systematically on most configurations.

Since the SHA ID isn't necessary anyway, print an error and continue.


derekparker (approved):
LGTM

PR #2974, opened 2022-04-19
Fix example block in cli README

deathiop:
This missing blank line breaks the example block, every lines are inlined in a single one.


aarzilli:
This documentation file is autogenerated from the contents of pkg/terminal/command.go, that's where the change should be made and then the README.md should be regenerated by running _scripts/gen-cli.docs.go

aarzilli (approved):
LGTM

deathiop:
Not sure why the `Aggregator (Delve) - TeamCity build` fails, seems unrelated to this tiny commit.

derekparker (approved):
LGTM

PR #2975, opened 2022-04-20
proc: fix autogenerated helper skip on linux/386 PIE builds

aarzilli:
The fix in #2959 was incomplete and the skip was still broken on
linux/386 for PIE builds due to calls to the get_pc_thunk helper.


derekparker (approved):
LGTM

PR #2980, opened 2022-04-21
_scripts: always return exit code 0 when testing on tip

aarzilli:
So that we can mute failing tests in TeamCity.

### DIFF
--- a/_scripts/test_mac.sh
+++ b/_scripts/test_mac.sh
@@ -42,4 +42,11 @@ export GOARCH="$ARCH"
 export PATH="$GOROOT/bin:$PATH"
 go version
 
+set +e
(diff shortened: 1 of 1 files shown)


derekparker:
I suppose this will mute them, yes, but it won't give us any visibility that the test is actually failing, right? In other words, I still want to see that the test has failed, I just don't want it to fail the entire build for us. Does that make sense?

aarzilli:
AFAICT what will happen is that the exit status won't make it fail but the failed tests will still make it fail unless they are muted. In theory it should be fine to do it always except that, IIRC, the tests timing out on windows weren't being detected as failed without checking the exit status.

PR #2982, opened 2022-04-21
service/debugger: make TestDebugger_LaunchWithTTY pass on FreeBSD

4a6f656c:
FreeBSD does not have lsof, rather it has fstat. Additionally, the TTY name
does not match up with the current code.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #2997, opened 2022-05-02
_scripts/test_linux.sh: always return exit code 0 when testing on tip

aarzilli:
Same as what we do for test_mac.sh


derekparker (approved):
LGTM

PR #2998, opened 2022-05-02
Documentation: document substitute-path more

aarzilli:
Add FAQ entry about substitute-path since it comes up often and link it
in the issue template.


Documentation/faq.md
```diff
@@ -76,3 +76,34 @@ ssh -NL 4040:localhost:4040 user@remote.ip
...
+
+This problem manifests when one or more of these things happen:
+
+* Can not see source code when the program stops at a breakpoint
+* Setting a breakpoint using full path, or through a IDE, does not work
+
+while doing one of the following things
+
+* **The program is built and run inside a container** and Delve (or an IDE) is remotely connecting to it
+* More in general, every time the build environment (VM, container, computer...) differs from the environment where Delve's front-end (dlv or a IDE) runs
```
derekparker, line 90: s/More in general/Generally/

derekparker (changes requested):
Just a few nits.

Documentation/faq.md
```diff
@@ -76,3 +76,34 @@ ssh -NL 4040:localhost:4040 user@remote.ip
...
+(dlv)
+```
+
+This is not a bug. The Go compiler embeds the paths of source files into the executable so that debuggers, including Delve, can use them. Doing any of the things listed above will prevent this feature from working seamlessly.
+
+The substitute-path feature can be used to solve this problem, see `config help substitute-path` or the `substitutePath` option in launch.json.
+
+The `source` command could also be useful in troubleshooting this problem, it shows the list of file paths that has been embedded by the compiler into the executable.
+
+If you still think this is a bug in Delve and not a configuration problem open an [issue](https://github.com/go-delve/delve/issues), filling the issue template and including the logs produced by delve with the options `--log --log-output=rpc,dap`.
```
polinasok, line 109: comma before "open an"

polinasok (changes requested):
Always happy to see more docs!

polinasok (approved):

derekparker (approved):
LGTM

PR #2999, opened 2022-05-04
Upgrade FreeBSD version

aarzilli:


aarzilli:
It's failing because floatsum doesn't work on freebsd: #3001

PR #3000, opened 2022-05-04
proc: fixes concerning call injection on freebsd and rr

aarzilli:
On FreeBSD it seems we have problems restoring and setting floating
point registers, since at least restoring is necessary for call
injection to function properly fully disable call injection on FreeBSD.

On rr the same problem exists, however due to the fact that we are
acting on a recording and ending a diversion will restore register
values anyway simply disable the floatsum test.

See also: https://github.com/rr-debugger/rr/issues/3208

Updates #3001

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,7 @@ env:
   GOFLAGS: -mod=vendor
 
 freebsd_instance:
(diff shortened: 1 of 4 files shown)


derekparker (approved):
LGTM

PR #3002, opened 2022-05-05
proc: fix RFLAGS corruption after call injection on amd64

aarzilli:
debugCallV2 for amd64 has a bug where it corrupts the flags registers
every time it is called, this commit works around that problem by
restoring flags one extra time to its original value after stepping out
of debugCallV2.


derekparker (approved):
LGTM

Issue #2772, opened 2021-11-02
service/dap: always continue on disconnect from multi-client server?


polinasok:
Terminal client offers a special flag `exit -c` to quit the client and resume execution. Vscode UI doesn't offer nuanced disconnect options (https://github.com/microsoft/vscode/issues/134412). Instead a user must first click Continue and then Disconnect to continue on disconnect. If the process is already running when disconnect is requested, it will keep running. If the server is halted, it will stay halted.

A while back the vscode-go legacy adapter [was updated ](https://github.com/golang/vscode-go/commit/12e064c8af574bc0d80accbbbc72482b78c84e51)to [always ask dlv to continue](https://github.com/golang/vscode-go/blob/44e2d4d52130286a1f7e2fba33f0fcc1cbe2d941/src/debugAdapter/goDebug.ts#L998-L1005) when a remote client disconnected.

Should we do the same in the dap server/adapter? This is likely the desired shortcut for many remote users using multi-client mode, but it would make it impossible to halt on disconnect.

Another option is to control this with a boolean attach request flag - .e.g. `continueOnDisconnect`.

polinasok:
cc @quoctruong
cc @briandealwis

quoctruong:
Can we have this option on a flag? From past feedback from users, they tend to want their app to be running even after the disconnect. However, this is not a blocker for us.

polinasok:
Sounds reasonable. Thanks.

polinasok:
Spoke with @quoctruong offline. It appears that this feature is becoming less critical for CloudCode if necessary at all. CloudCode is starting to provide container setup for the user, so the sessions now consist of launching everything from scratch (as opposed to attaching to an externally created container with dlv running). Thus when users disconnect, everything will be stopped altogether. CloudCode is still collecting user feedback on these changes. And we are yet to receive feedback on remote via DAP after enabling it in vscode-go recently. I am going to keep this open for now. But if we don't hear any explicit user requests in the foreseeable future, I suggest we don't do this. Adding a launch option is easy. Deprecating it is hard. Let's only add what we definitely know our users need, especially when there is a workaround available (explicitly clicking Continue before disconnecting).

polinasok:
@briandealwis

briandealwis:
[...]

PR #3003, opened 2022-05-05
dap: support suspendDebuggee in multi-client mode

polinasok:
Upstream dependency at https://github.com/microsoft/vscode/issues/134412 is in, so we can now provide this in our DA to give people more disconnect options. 

```
- disconnect            = leaves debugger and target running (default)
- disconnect + suspend  = leaves debugger and target suspended
- stop                  = kills  debugger and target
```

Updates #2772


polinasok:
Actually give me a moment to revise something before reviewing.

service/dap/server.go
```diff
@@ -1136,25 +1137,34 @@ func (s *Session) onDisconnectRequest(request *dap.DisconnectRequest) {
 		// This is a multi-use server/debugger, so a disconnect request that doesn't
 		// terminate the debuggee should clean up only the client connection and pointer to debugger,
 		// but not the entire server.
-		status := "halted"
-		if s.isRunningCmd() {
+		status := "running"
+		if request.Arguments.SuspendDebuggee {
+			status = "suspended"
+			s.changeStateMu.Lock()
+			defer s.changeStateMu.Unlock()
```
aarzilli, line 1144: This defer gets executed when the function exits. Is this really the intended scope? There's a lot of code after this if/else that will get executed either with or without holding the mutex.

derekparker:
Going to close for now due to inactivity. Please reopen if this is still necessary / planned to be worked on.

PR #3004, opened 2022-05-08
Add auto-complete for local variables

pippolo84:
Add auto-complete for local variables when using `print` or `whatis` commands.

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -246,6 +246,22 @@ func (t *Term) Run() (int, error) {
 			commands := cmds.FuzzySearch(strings.ToLower(line))
 			c = append(c, commands...)
 		case "print", "whatis":
+			if locs == nil {
+				localVars, err := t.client.ListLocalVariables(
+					api.EvalScope{GoroutineID: -1, Frame: t.cmds.frame, DeferredCall: 0},
+					api.LoadConfig{},
+				)
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "Unable to get local variables: %s\n", err)
+					break
+				}
+
+				locs = trie.New()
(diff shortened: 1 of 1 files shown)


pkg/terminal/terminal.go
```diff
@@ -243,6 +243,28 @@ func (t *Term) Run() (int, error) {
 		case "nullcmd", "nocmd":
 			commands := cmds.FuzzySearch(strings.ToLower(line))
 			c = append(c, commands...)
+		case "print", "whatis":
+			localVars, err := t.client.ListLocalVariables(
+				api.EvalScope{GoroutineID: -1, Frame: t.cmds.frame, DeferredCall: 0},
+				t.loadConfig(),
```
aarzilli, line 249: use `api.LoadConfig{}` so we don't load as much, also the trie should be cached at least until a command is executed (i.e. cleared before the promptForInput call).

pkg/terminal/terminal.go
```diff
@@ -279,6 +289,19 @@ func (t *Term) Run() (int, error) {
 	_, _ = t.client.GetState()
 
 	for {
+		localVars, err := t.client.ListLocalVariables(
```
aarzilli, line 292: What I was thinking was only clearing the cache here (`locs = nil`) but filling it in the completer function (`if locs == nil { call ListLocalVariables, etc... }`)

pippolo84: Sorry, I misunderstood your comment above. You meant a lazy loading of the cache. Seems reasonable, I'm gonna fix it. Thanks.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3007, opened 2022-05-13
proc/gdbserver: fix function call injection on rr

aarzilli:
Issue https://github.com/rr-debugger/rr/issues/3208 was fixed so finish
fixing call injection on rr and remove the test exception.


derekparker (approved):
LGTM

Do we need a version of RR built from source for this to work properly?

Issue #2993, opened 2022-05-01
Failing tests on go1.19


aarzilli:
- [ ] TestCgoStacktrace fails on darwin/arm64 with go1.19
- [x] tests that rely on detecting a fatal throws on amd64 (any OS) fail on go1.19: 
    - [x] pkg/proc: TestThreadFrameEvaluation
    - [x] pkg/proc: TestDeadlockBreakpoint
    - [x] service/dap: TestFatalThrowBreakpoint

aarzilli:
I'm going to keep track here of all go1.19 test failures.

polinasok:
Tests got disabled here: https://github.com/go-delve/delve/pull/2961

polinasok:
What does this apply to?
https://github.com/go-delve/delve/blob/1418cfd3852e2e2ed21e3fe6c74d567eb2e495bb/Documentation/backend_test_health.md?plain=1#L10-L11
https://github.com/go-delve/delve/blob/1418cfd3852e2e2ed21e3fe6c74d567eb2e495bb/Documentation/backend_test_health.md?plain=1#L19-L20

aarzilli:
Those are computed from skipOn calls in proc_test.go

aarzilli:
@nd it's complicated to investigate this (the cgo stacktrace problem on darwin/arm64) without access to hardware, is there any chance I could get the shell access to the M1 builder back? It isn't working at the moment (it rejects my public key).

nd:
@aarzilli please try to log in. Your key is in authorized_keys. I've restarted the machine, maybe it could help.

aarzilli:
Still getting `Permission denied (publickey)`

nd:
I've enabled debug logging, please try again.

nd:
The user should be builduser

aarzilli:
I can in fact connect as builduser, I think it used to be my github username (it still is on the x86 builder).

aarzilli:
PS thank you

nd:
> it still is on the x86 builder

strange, there there is no MacOS user with such a name there. Could you please log in and out to that machine as well?

aarzilli:
Bisects to 0ec08283c8e3a673d9916c186f8d2f5428846862

PR #3009, opened 2022-05-13
proc: fix fatal throw breakpoint on go1.19

aarzilli:
Also set breakpoint on runtime.fatal, which didn't exist before go1.19.

Updates #2993


derekparker (approved):
LGTM

derekparker (changes requested):
Looking at CI I think this will be green after a rebase.

derekparker (approved):
LGTM

PR #3019, opened 2022-05-29
pkg/proc: fix and improve freebsd register handling

### LINKED ISSUE #3001: Can not inject function calls that use floating point registers on FreeBSD
It looks like there is some problem with the xsave area not being retrieved on FreeBSD, see the floatsum injection test in TestCallFunction.
Maybe we should disable function call injection on FreeBSD entirely.

4a6f656c:
proc: deduplicate register setting code on freebsd

proc: re-enable function call injection on freebsd
    
Floating point registers can now be set and restored correctly.
    
This is a partial revert of 51090f003bace1f8cc37b8480ffdb6f6cc91fa5a.
    
Fixes #3001

proc: fix floating point register setting on freebsd
    
The original code could never work - PT_SETREGS on freebsd does not
take an iovec, nor does it set FP registers. Furthermore, the xsave
bytes were not stored in the amd64util.AMD64Xstate struct.
    
Updates #3001

proc: return *amd64util.AMD64Xstate from freebsd ptraceGetRegset
    
Return a pointer to a struct, rather than a struct - this simplifies the
code in both the caller and the ptraceGetRegset function, while also avoiding
struct copying.

proc: check return values of ptrace calls on freebsd
    
The return values of the PT_GETNUMLWPS and PT_GETLWPLIST ptrace calls were
previously unchecked. While these should not fail, panic instead of using
-1 with slice allocation/handling.

proc: convert freebsd ptrace code to cgo
    
There is little point in having cgo call a custom C function, when the same
can be done directly from cgo (with less code and effort). Split the amd64
specific code into ptrace_freebsd_amd64.go. Also avoid mixing C.ptrace()
with syscall.SYS_PTRACE.
    
This will make further changes easier - no functional change intended.


aarzilli (approved):
LGTM, thank you!

derekparker (approved):
LGTM

PR #3023, opened 2022-05-31
proc/native: fix Windows build

aarzilli:
Commit b53fcbe broke the build on Windows, we didn't notice because 2d9a9d broke the test script.


derekparker (approved):
LGTM

PR #3024, opened 2022-06-02
service/test,proc: move variable evaluation test to pkg/proc

aarzilli:
Moves variable evaluation tests to pkg/proc which is where most of the
code they test exists, this also allows the removal of a bunch of
duplicate support code.


derekparker (approved):
LGTM

PR #3030, opened 2022-06-12
proc,service/debugger: introduce TargetGroup abstraction

aarzilli:
Introduces a new TargetGroup abstraction that can be used to manage
multiple related targets.

No actual management of child processes is implemented here, this is
just a refactoring to make it possible to do that in the future.

Updates #2551

### DIFF
--- a/pkg/proc/target_group.go
+++ b/pkg/proc/target_group.go
@@ -1,0 +1,123 @@
+package proc
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TargetGroup reperesents a group of target processes being debugged that
+// will be resumed and stopped simultaneously.
+// New targets are automatically added to the group if exec catching is
+// enabled and the backend supports it, otherwise the group will always
+// contain a single target process.
+type TargetGroup struct {
+	targets  []*Target
+	Selected *Target
+
+	RecordingManipulation
+	recman RecordingManipulationInternal
+
+	// KeepSteppingBreakpoints determines whether certain stop reasons (e.g. manual halts)
+	// will keep the stepping breakpoints instead of clearing them.
+	KeepSteppingBreakpoints KeepSteppingBreakpoints
+
+	LogicalBreakpoints map[int]*LogicalBreakpoint
+
+	continueOnce ContinueOnceFunc
+	cctx         *ContinueOnceContext
+}
+
+// NewGroup creates a TargetGroup containing the specified Target.
+func NewGroup(t *Target) *TargetGroup {
+	if t.partOfGroup {
+		panic("internal error: target is already part of a group")
+	}
+	t.partOfGroup = true
+	return &TargetGroup{
+		RecordingManipulation: t.recman,
+		targets:               []*Target{t},
+		Selected:              t,
+		cctx:                  &ContinueOnceContext{},
+		recman:                t.recman,
+		LogicalBreakpoints:    t.Breakpoints().Logical,
+		continueOnce:          t.continueOnce,
+	}
+}
+
+// Targets returns a slice of targets in the group.
+func (grp *TargetGroup) Targets() []*Target {
+	return grp.targets
+}
+
+// Valid returns true if any target in the target group is valid.
+func (grp *TargetGroup) Valid() (bool, error) {
+	var err0 error
+	for _, t := range grp.targets {
+		ok, err := t.Valid()
+		if ok {
+			return true, nil
+		}
+		err0 = err
+	}
+	return false, err0
+}
+
+// Detach detaches all targets in the group.
+func (grp *TargetGroup) Detach(kill bool) error {
+	var errs []string
+	for _, t := range grp.targets {
+		isvalid, _ := t.Valid()
+		if !isvalid {
+			continue
+		}
+		err := t.detach(kill)
+		if err != nil {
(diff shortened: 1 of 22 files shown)


pkg/proc/target_group.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// TargetGroup reperesents a group of target processes being debugged that
+// will be resumed and stopped simultaneously.
+// New targets are automatically added to the group if exec catching is
+// enabled and the backend supports it, otherwise the group will always
+// contain a single target process.
+type TargetGroup struct {
+	targets []*Target
+	procs   []ProcessInternal
```
derekparker, line 15: The `Target` type will already have it's associated `ProcessInternal` why store them separately?

pkg/proc/target_group.go
```diff
@@ -0,0 +1,125 @@
...
+	Sel     *Target
```
derekparker, line 16: nit: lets replace `Sel` with something more descriptive like `Main` or `Root` (as in root of the process tree) or just don't abbreviate and use `Selected`.

pkg/proc/target_group.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+// Targets returns a slice of targets in the group.
+func (grp *TargetGroup) Targets() []*Target {
+	return grp.targets
+}
+
+// Valid returns true if any target in the target group is valid.
+func (grp *TargetGroup) Valid() (bool, error) {
```
derekparker, line 55: Is this the correct behavior? Since the `TargetGroup` operates on every `Target` within it, wouldn't we want `Valid` to be true if *all* targets are valid?

pkg/proc/target_group.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, t := range grp.targets {
+		if t.Breakpoints().HasSteppingBreakpoints() {
+			return t.ClearSteppingBreakpoints()
+		}
+	}
+	return nil
+}
+
+// ThreadList returns a list of all threads in all target processes.
+func (grp *TargetGroup) ThreadList() []Thread {
```
derekparker, line 107: Should this be a slice of slices where threads belonging to a process are grouped?

service/debugger/debugger.go
```diff
@@ -1400,11 +1467,15 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
 	}
 
 	funcs := []string{}
-	for _, f := range d.target.BinInfo().Functions {
-		if regex.MatchString(f.Name) {
-			funcs = append(funcs, f.Name)
+	for _, t := range d.target.Targets() {
```
derekparker, line 1470: Same comment about grouping per process.

derekparker (changes requested):
First review pass done.

pkg/proc/target_group.go
aarzilli: Trying to not reallocate it every time, but it's an optimization and I didn't really measure it so I'm changing it.

pkg/proc/target_group.go
aarzilli: Changed to Selected (it's meant to be the one on which debugger operates currently).

pkg/proc/target_group.go
aarzilli: As long as one of the processes is valid (= not quit) the group is valid. Child processes can die and leave the group.

pkg/proc/target_group.go
aarzilli: The idea is to use this where we don't care about which process a thread belongs to (there are a couple of cases where this happens)

service/debugger/debugger.go
```diff
@@ -1381,14 +1429,33 @@ func (d *Debugger) Sources(filter string) ([]string, error) {
 	}
 
 	files := []string{}
-	for _, f := range d.target.BinInfo().Sources {
-		if regex.Match([]byte(f)) {
-			files = append(files, f)
+	for _, t := range d.target.Targets() {
```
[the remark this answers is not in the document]
aarzilli: IMO returning a unified list for all processes provides the smoothest backwards compatibility for existing frontends. For example:
	
- getting a list of files and functions used by the program that the user can use to navigate the source code
- autocompletion for breakpoints
	
These are better served by the unified list rather than per-process lists. Dividing them per process would also need augmenting the API so if someone needs it we can add it later anyway.

service/debugger/debugger.go
```diff
@@ -1418,17 +1489,22 @@ func (d *Debugger) Types(filter string) ([]string, error) {
...
-	if err != nil {
-		return nil, err
-	}
+	r := []string{}
 
-	r := make([]string, 0, len(types))
-	for _, typ := range types {
-		if regex.Match([]byte(typ)) {
-			r = append(r, typ)
+	for _, t := range d.target.Targets() {
```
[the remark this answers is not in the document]
derekparker: Ack.

service/debugger/debugger.go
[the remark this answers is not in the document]
derekparker: SGTM

derekparker (approved):

derekparker (approved):
LGTM

PR #3031, opened 2022-06-12
*: run go1.19 'go fmt' on everything and fix problems

aarzilli:
Go 1.19 also formats doc comments according to the new godoc syntax.

Some of our comments, especially unexported symbols did not conform to
the godoc syntax and therefore are mangled by 'go fmt'.

This PR runs 'go fmt' from go1.19 on everything and manually fixes the
problems.

See also:
https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md


derekparker (approved):
LGTM

PR #3032, opened 2022-06-14
dwarf/frame: discard indirect flag reading personality func in eh_frame

### LINKED ISSUE #3015: Could not parse .eh_frame section: pointer encoding not supported

```
Delve Debugger
Version: 1.8.2
Build: 65e03d673d31f9a6c9c2bd187a4ff2219444b16f
```

```
go1.18.2 linux/amd64
```

Ubuntu 20.04.4 LTS, Ubuntu 22.04 LTS

4. What did you do?

Debugged a program and got the error: `debugger error loading binary "/lib/x86_64-linux-gnu/libc.so.6": could not parse .eh_frame section: pointer encoding not supported 0x9b at 0x59f4`.

Is there a ticket for such a support? Is it possible to work around the error?

aarzilli:
We don't do anything with the personality function so there is no point
in complaining that we don't fully support the pointer encoding flags
used to describe it.

This matches the current level of support of pointer encodings in gdb
(they are discarded when reading the personality function and not
supported for FDEs because gcc doesn't generate them).

Fixes #3015

### DIFF
--- a/pkg/dwarf/frame/entries.go
+++ b/pkg/dwarf/frame/entries.go
@@ -108,8 +108,8 @@ func (fdes FrameDescriptionEntries) Append(otherFDEs FrameDescriptionEntries) Fr
 // ptrEnc represents a pointer encoding value, used during eh_frame decoding
 // to determine how pointers were encoded.
 // Least significant 4 (0xf) bytes encode the size  as well as its
-// signed-ness,  most significant 4 bytes (0xf0) are flags describing how
-// the value should be interpreted (absolute, relative...)
+// signed-ness,  most significant 4 bytes (0xf0 == ptrEncFlagsMask) are flags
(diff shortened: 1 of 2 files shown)


pkg/dwarf/frame/entries.go
```diff
@@ -142,7 +144,7 @@ func (ptrEnc ptrEnc) Supported() bool {
 			// These values aren't defined at the moment
 			return false
 		}
-		if ptrEnc&0xf0 != ptrEncPCRel {
+		if (ptrEnc&0xf0)&^ptrEncSupportedFlags != 0 {
```
derekparker, line 147: While we're touching this code, can we assign the integer immediate values to constants and explain their usage.

aarzilli: Done.

PR #3036, opened 2022-06-19
proc: map access with string literal key should always succeed

### LINKED ISSUE #3034: Can't extract values out of a map with string keys - "String too long for comparison" error

```
Delve Debugger                                                                                           │
Version: 1.8.3                                                                                           │
Build: $Id: f92bb46b82b3b92d79ce59c4b55eeefbdd8d040c $
```

3. What version of Go are you using? (`go version`)?
```
go version go1.18.1 linux/amd64
```
5. What operating system and processor architecture are you using?
```
Linux, Fedora 35
```

7. What did you do?
I was trying to debug kubelet from https://github.com/kubernetes/kubernetes/ and I compiled kubelet with debug enabled and then connected using `dlv attach`.  I created a breakpoint and tried to do following:

```
(dlv) print len(rc.actualStateOfWorld.attachedVolumes["kubernetes.io/secret/3e329114-e019-4e15-ac1b-6de60b96cde4-sleep-volume"])
Command failed: string too long for comparison                                                                                                                                                                    
```

9. What did you expect to see?

I expected it to print the entry in the map.

11. What did you see instead?

It threw error instead.

aarzilli:
When doing a map lookup with a string literal we should load as much of
the keys to at least match the length of the string literal, so that
the lookup doesn't fail with the "string too long" error.

Fixes #3034

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -2038,10 +2038,18 @@ func (v *Variable) mapAccess(idx *Variable) (*Variable, error) {
 		return nil, fmt.Errorf("can not access unreadable map: %v", v.Unreadable)
 	}
 
+	lcfg := loadFullValue
+	if idx.Kind == reflect.String && int64(len(constant.StringVal(idx.Value))) == idx.Len && idx.Len > int64(lcfg.MaxStringLen) {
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

PR #3038, opened 2022-06-21
proc: fix arm64SwitchStack for go1.19

### LINKED ISSUE #2993: Failing tests on go1.19

aarzilli:
In commit eee6f9f82 Go changed the order that crosscall2 uses to save
its registers, update arm64SwitchStack to use the new order when the
binary was compiled by go 1.19.

Fixes #2993


derekparker (approved):
LGTM

PR #3039, opened 2022-06-21
proc/native: fix Ctrl-C handling on non-headless instances on windows

aarzilli:
In non-headless instances Ctrl-C should pause the process, not
terminate it. To make this work on Windows we have to pass the
syscall.CREATE_NEW_PROCESS_GROUP flag to os.StartProcess.


derekparker (approved):
LGTM

PR #3046, opened 2022-06-28
_scripts: un-pin mingw version

aarzilli:
The version of mingw was pinned to 10.2.0 because mingw 11 started
using DWARFv5, which wasn't supported by Go 1.17. We are now testing
only 1.18 and 1.19 and both support DWARFv5 correctly.


derekparker (approved):
LGTM

PR #3052, opened 2022-07-01
dwarf/op,proc: output register name when printing location exprs

aarzilli:
Output the architecture specific register name when printing location
expression after DW_OP_regN and DW_OP_bregN operators.


derekparker:
Re-running CI.

derekparker:
TestGeneratedDoc is failing on tip do we need to rebase this or is this a fresh failure?

aarzilli:
It's just the go/packages incompatibility we merged a while back.

PR #3053, opened 2022-07-02
vendor: update version of go/packages to 0.1.11

aarzilli:
The current version we use now crashes with go1.19:

https://github.com/golang/go/issues/45584


derekparker (approved):
LGTM

PR #3061, opened 2022-07-13
Remove ignored flags from dlv_dap.md

timtraversy:
According to https://github.com/go-delve/delve/blob/master/cmd/dlv/cmds/commands.go#L434, many of the flags listed in the docs for `dlv dap` are actually ignored.

I suppose they are technically "inherited from parent commands", but I think it's confusing to list them if they will have no effect.

### DIFF
--- a/Documentation/usage/dlv_dap.md
+++ b/Documentation/usage/dlv_dap.md
@@ -40,22 +40,16 @@ dlv dap [flags]
 ### Options inherited from parent commands
 
 ```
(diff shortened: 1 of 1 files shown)


aarzilli:
The documentation is automatically generated by _scripts/gen-usage-docs.go, hence the test failures.

timtraversy:
Ah should have guessed - ok thanks.

PR #3063, opened 2022-07-13
Add support for windows/arm64

### LINKED ISSUE #2129: Support for arm64 windows machines

go1.14.7
Windows 64-bit ARM
4. What did you do?
expected dlv to install
6. What did you see instead?
$ go get github.com/go-delve/delve/cmd/dlv
service\debugger\debugger.go:26:2: found packages native (proc.go) and your_operating_system_and_architecture_combination_is_not_supported_by_delve (support_sentinel.go) in C:\Users\viggn\Desktop\go\src\github.com\go-delve\delve\pkg\proc\native

qmuntal:
This PR adds support for debugging on `windows/arm64`. There are some tests failing, but with these changes I can use `dlv` to debug big codebases without problems.

I could reuse all work done to support `arm64` on linux. The only difference is that windows use `BRK 0xF000` as breakpoint instruction instead of just `BRK`.

On the windows syscall side, I've redefined `CONTEXT` so it matches Windows ARM64 expectations.

There are a couple of things that are still not supported:
- Core dumps
- Hardware breakpoints
- Some tests using `runtime.Breakpoint()` fail due to https://github.com/golang/go/issues/53837

Fixes #2129


pkg/proc/native/syscall_windows_arm64.go
```diff
@@ -0,0 +1,35 @@
...
+	_CONTEXT_EXCEPTION_REQUEST   = 0x40000000
+	_CONTEXT_EXCEPTION_REPORTING = 0x80000000
+)
+
+// zsyscall_windows.go, an autogenerated file, wants to refer to the context
+// structure as _CONTEXT, but we need to have it in pkg/proc/winutil.CONTEXT
+// because it's also used on non-windows operating systems.
+type _CONTEXT = winutil.ARM64CONTEXT
+
+func GetThreadContext(thread syscall.Handle, context winutil.CONTEXT) (err error) {
```
aarzilli, line 29: We don't want to export this function. But also I'm not a fan of having winutil.CONTEXT be an interface, at least in pkg/proc/native. Because it doesn't actually have a dynamic type, inside pkg/proc/native it will always have the type specific to the architecture we are running on. The code in pkg/proc/native should use the concrete type, code that isn't architecture specific should still manipulate the concrete type using helper methods.
For example `setContextFlags(...)` which is implemented as `setContextFlags(*winutil.AMD64CONTEXT,uint32)` on amd64, and as `setContextFlags(*winutil.ARM64CONTEXT,uint32)` on arm64.

I see you are having problems with minidumps, I think the code there would be clearer and easier to write with concrete types also (but I don't know what the problem is so I could be wrong).

pkg/proc/native/threads_windows.go
```diff
@@ -158,18 +157,18 @@ func (t *nativeThread) ReadMemory(buf []byte, addr uint64) (int, error) {
 }
 
 func (t *nativeThread) restoreRegisters(savedRegs proc.Registers) error {
-	return _SetThreadContext(t.os.hThread, savedRegs.(*winutil.AMD64Registers).Context)
+	return SetThreadContext(t.os.hThread, savedRegs.(winutil.Registers).Ctx())
```
aarzilli, line 160: I'd rather have two implementations of this, one for each architecture, than have an implementation of Ctx for each architecture and the winutil.Registers interface.

aarzilli (commented):
Aside from the minidump problem and the comment on the CONTEXT interface I think there are three questions two answer:

1. I think we should wait to merge this until we have a version of Go that we can test it on. It doesn't have to be a released version, it could be go built at tip.
2. You say that some tests don't pass but made no changes to _test.go files, except core_test.go. Tests that don't pass should be skipped.
3. We would need to have some way to run CI on windows/arm64 otherwise it is going to break very soon

Regarding function call injection, I would expect it to work effortlessly, I suspect the problem you are encountering is that the runtime uses BRK extensively to communicate with delve for this, all of those BRK instruction will have the same problem runtime.Breakpoint() has.

pkg/proc/native/threads_windows.go
qmuntal: Done! It does make it clear.

pkg/proc/native/syscall_windows_arm64.go
qmuntal: Agree and done! Thanks for the examples, they helped me with the refactor.

qmuntal:
> I think we should wait to merge this until we have a version of Go that we can test it on. It doesn't have to be a released version, it could be go built at tip.

Go provides official `windows/arm64` since go1.17, so this should be a blocker.

> You say that some tests don't pass but made no changes to _test.go files, except core_test.go. Tests that don't pass should be skipped.

Will do that soon.

> We would need to have some way to run CI on windows/arm64 otherwise it is going to break very soon

Agree. Could we at least merge the part of this PR that refactors `windows` to be more friendly to non-amd64 architectures? This will pave the path for when we have `windows/arm64` CI and it will allow me, and other people wanting to debug arm64 using this branch, to patch arm64 support with minimal effort.

pkg/proc/native/proc_windows.go
```diff
@@ -466,7 +465,7 @@ func (dbp *nativeProcess) stop(cctx *proc.ContinueOnceContext, trapthread *nativ
 			_, thread.os.delayErr = _SuspendThread(thread.os.hThread)
 			if thread.os.delayErr == nil {
 				// This call will block until the thread has stopped.
-				_ = _GetThreadContext(thread.os.hThread, context)
+				_ = thread.setContext(context)
```
aarzilli, line 468: getContext

aarzilli:
> > I think we should wait to merge this until we have a version of Go that we can test it on. It doesn't have to be a released version, it could be go built at tip.
> 
> Go provides official `windows/arm64` since go1.17, so this should be a blocker.

I don't understand what you mean by this.
> > We would need to have some way to run CI on windows/arm64 otherwise it is going to break very soon
> 
> Agree. Could we at least merge the part of this PR that refactors `windows` to be more friendly to non-amd64 architectures? This will pave the path for when we have `windows/arm64` CI and it will allow me, and other people wanting to debug arm64 using this branch, to patch arm64 support with minimal effort.

I think that's fine.

pkg/proc/native/proc_windows.go
qmuntal: Good catch.

qmuntal:
>> I think we should wait to merge this until we have a version of Go that we can test it on. It doesn't have to be a released version, it could be go built at tip.

>> Go provides official windows/arm64 since go1.17, so this should be a blocker.

> I don't understand what you mean by this.

Then I don't understand your initial concern. Go already supports windows/arm64, so there is a version of Go that we can test it on. What's missing other than the CI?

aarzilli:
> Then I don't understand your initial concern. Go already supports windows/arm64, so there is a version of Go that we can test it on. What's missing other than the CI?

I was thinking about the problem with runtime.Breakpoint, I think we should wait until there's a version of go with a fixed runtime.Breakpoint, a large number of tests depends on runtime.Breakpoint being handled correctly.

qmuntal:
@aarzilli I've rebased this PR and gated windows/arm64 behind the `exp.winarm64` tag. What do you think?

I'm also struggling to run the test suite locally either on amd64 and arm64 because `proc.FindFileLocation` always returns `ErrCouldNotFindLine`. There seems to be a mishandling of the file name casing on windows.

derekparker:
> @aarzilli I've rebased this PR and gated windows/arm64 behind the `exp.winarm64` tag. What do you think?
> 
> I'm also struggling to run the test suite locally either on amd64 and arm64 because `proc.FindFileLocation` always returns `ErrCouldNotFindLine`. There seems to be a mishandling of the file name casing on windows.

That's an odd one, our Windows builders seem to be running fine. Is there anything different about your test / development systems that could cause such an error? I don't have a Windows machine handy to test on right now.

qmuntal:
> That's an odd one, our Windows builders seem to be running fine. Is there anything different about your test / development systems that could cause such an error? I don't have a Windows machine handy to test on right now.

Found the culprit. I'm running the tests inside vscode, and for historical reasons they normalize the drive letter to lower-case. This shouldn't be a problem, because Windows file paths are case-insensitive, but `proc.FindFileLocation` is case sensitive, so it can't find the file even if it is there.

.teamcity/settings.kts
```diff
@@ -201,7 +204,11 @@ class TestBuild(val os: String, val arch: String, version: String, buildId: Abso
                     scriptMode = file {
                         path = "_scripts/test_windows.ps1"
                     }
-                    param("jetbrains_powershell_scriptArguments", "-version ${"go$version"} -arch $arch")
+                    if (arch == "arm64") {
+                        param("jetbrains_powershell_scriptArguments", "-version ${"go$version"} -arch amd64 -winarm64")
```
aarzilli, line 208: I don't understand how this change works. You are making test_windows.ps1 download the amd64 version of the toolchain and then run the tests by setting GOARCH manually to arm64? Why?

.teamcity/settings.kts
```diff
@@ -48,6 +48,9 @@ val targets = arrayOf(
         "windows/amd64/1.18",
         "windows/amd64/tip",
 
+        "windows/arm64/1.18",
```
aarzilli, line 51: IMO we should add these until there are build agents. I don't know if teamcity is going to complain but even if it doesn't there's no point on having them here.

pkg/proc/bininfo.go
```diff
@@ -1656,14 +1655,17 @@ func loadBinaryInfoPE(bi *BinaryInfo, image *Image, path string, entryPoint uint
 		return err
 	}
 
-	//TODO(aarzilli): actually test this when Go supports PIE buildmode on Windows.
-	opth := peFile.OptionalHeader.(*pe.OptionalHeader64)
-	if entryPoint != 0 {
-		image.StaticBase = entryPoint - opth.ImageBase
-	} else {
-		if opth.DllCharacteristics&_IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE != 0 {
```
aarzilli, line 1664: Why did you remove this check?

qmuntal: windows/arm64 binaries are always PIE, even if built with `-buildmode=exe`, so we need to support `entryPoint==0`. I don't know why this check was added, but as per the TODO comment I supposed it was just that it still wasn't tested with a real program, which I'm doing now.

.teamcity/settings.kts
qmuntal: We still don't have a windows arm64 builder to run tests, but at least we can cross-compile from the windows amd64 builder to check for compilation errors.

.teamcity/settings.kts
qmuntal: Related to https://github.com/go-delve/delve/pull/3063#discussion_r933114531.

This target is really running on amd64 and cross-compiling to arm64.

.teamcity/settings.kts
aarzilli: But is this going to work given that there are no arm64 agents? And when there will be (if there will be any) it definitely won't work because they will try to execute the amd64 version of the go toolchain. I'd just add a build command at the end of test_windows.ps1.

pkg/proc/bininfo.go
aarzilli: The TODO command was obsolete, we did end up checking that this worked on windows. AFAIK PIE binaries always have non-zero entryPoint, the check is to make sure that we don't proceed with a PIE binary for which we weren't able to retrieve the entryPoint.

.teamcity/settings.kts
qmuntal: I'll add the build command at the end of the ps1 script.

derekparker:
> > That's an odd one, our Windows builders seem to be running fine. Is there anything different about your test / development systems that could cause such an error? I don't have a Windows machine handy to test on right now.
> 
> Found the culprit. I'm running the tests inside vscode, and for historical reasons they normalize the drive letter to lower-case. This shouldn't be a problem, because Windows file paths are case-insensitive, but `proc.FindFileLocation` is case sensitive, so it can't find the file even if it is there.

Ack, thanks for looking into that.

pkg/proc/arm64_arch.go
```diff
@@ -13,16 +13,23 @@ import (
 )
 
 var arm64BreakInstruction = []byte{0x0, 0x0, 0x20, 0xd4}
+var arm64WindowsBreakInstruction = []byte{0x0, 0x0, 0x3e, 0xd4}
```
derekparker, line 16: Can you add some documentation context for why Windows uses a different breakpoint instruction, please.

pkg/proc/arm64_arch.go
```diff
@@ -102,12 +109,7 @@ func arm64FixFrameUnwindContext(fctxt *frame.FrameContext, pc uint64, bi *Binary
 	if a.crosscall2fn != nil && pc >= a.crosscall2fn.Entry && pc < a.crosscall2fn.End {
 		rule := fctxt.CFA
 		if rule.Offset == crosscall2SPOffsetBad {
-			switch bi.GOOS {
-			case "windows":
-				rule.Offset += crosscall2SPOffsetWindows
-			default:
-				rule.Offset += crosscall2SPOffsetNonWindows
-			}
+			rule.Offset += crosscall2SPOffsetNonWindows
```
derekparker, line 112: Do we need such specific constants now? E.g. can we get rid of Windows vs NonWindows specifiers?

derekparker (commented):
First pass, just a couple things.

pkg/proc/bininfo.go
qmuntal: There are some tests calling `LoadBinaryInfo` with 0 as hardcoded entry point, namely `TestDwarfVersion` and `TestRegabiFlagSentinel`. These tests fail on windows/arm64 because the binary is PIE and the entry point is 0.

Anyway, I can revert this change and fix the tests by using a fake entry point, as they don't really need this information.

pkg/proc/arm64_arch.go
qmuntal: Done!

pkg/proc/arm64_arch.go
qmuntal: We still need these constants, but I've updated their name to be more specific.

pkg/proc/bininfo.go
aarzilli: Yes, it's better to change the entry point used by the tests.

qmuntal:
@aarzilli @derekparker I'm mostly done with this PR and all tests in `pkg/proc` are passing. Could you do one final review?

About the Windows ARM64 builder, anyone can register for the Azure Arm-based VMs public preview at https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRyMSy8VejZVEo6yZykiPSHpURFRIQVY1QTcyWTlJNURUS1pNTktOUTUxVi4u.
See the announcement here: https://azure.microsoft.com/en-us/blog/now-in-preview-azure-virtual-machines-with-ampere-altra-armbased-processors/

It is not free of cost, I can't do anything here, but I can help setting up the testing environment once the VM is up and running.

pkg/proc/variables.go
```diff
@@ -438,8 +438,15 @@ func getGVariable(thread Thread) (*Variable, error) {
 
 	gaddr, hasgaddr := regs.GAddr()
 	if !hasgaddr {
-		var err error
-		gaddr, err = readUintRaw(thread.ProcessMemory(), regs.TLS()+thread.BinInfo().GStructOffset(), int64(thread.BinInfo().Arch.PtrSize()))
+		offset := thread.BinInfo().GStructOffset()
```
aarzilli, line 441: move this logic inside (*BinaryInfo).GStructOffset. Also instead of having a derefGStructOffset just check Arch.Name and BinaryInfo.GOOS.

aarzilli (commented):
> It is not free of cost, I can't do anything here

Steal an arm64 laptop and hide it in a closet.

pkg/proc/variables.go
qmuntal: > move this logic inside (*BinaryInfo).GStructOffset. 

If I understood you correctly, to do this I would have to update `GStructOffset` to accept a `MemoryReader` and return an error. Am I right?

aarzilli: Yes, that would be fine.

qmuntal: Done!

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

nathanhammond:
Thank you all for your work in getting this change landed! This will make a tremendous difference for our team and we're very appreciative. 🎉

PR #3069, opened 2022-07-21
cmd/dlv,proc: misc improvements to trace subcommand

aarzilli:
1. return an error when SetUProbe fails, while creating ebpf tracepoints
2. if no tracepoint can be set for 'dlv trace' exit early
3. never leave the traced program running
4. fix typo in description of --ebpf flag


derekparker (approved):
LGTM

PR #3072, opened 2022-07-22
proc/internal/ebpf: drop dependency on cgo

### LINKED ISSUE #2827: Can't install Delve with certain ld versions
`
[root@dino-ma soft]# go install github.com/go-delve/delve/cmd/dlv@master
go: downloading github.com/go-delve/delve v1.7.4-0.20211214142608-b81f618fb3aa
# github.com/go-delve/delve/cmd/dlv
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-432756995/000008.o(.text+0x74): unresolvable Hø@>HFHøHÁïHÁè@~FÃHø@~HÁè8HøHÁè0FHøHÁè(FHøHÁè FHøHÁèFHøHÁèFHøHÁFÃfD relocation against symbol `stderr@@GLIBC_2.2.5'
/usr/bin/ld: BFD version 2.20.51.0.2-5.48.el6_10.1 20100205 internal error, aborting at reloc.c line 443 in bfd_get_reloc_size

/usr/bin/ld: Please report this bug.

collect2: ld returned 1 exit status

[root@dino-ma soft]# go version
go version go1.16.7 linux/amd64`

aarzilli:
The ebpf implementations uses cgo, but only to access some C struct
definitions. Instead of using cgo simply duplicate the defintion of
those two structs in Go and add a test to check that the duplicate
definitions remain synchronized.

Fixes #2827

### DIFF
--- a/_scripts/test_linux.sh
+++ b/_scripts/test_linux.sh
@@ -41,7 +41,7 @@ GOPATH=$(pwd)/go
 export GOPATH
 export PATH=$PATH:$GOROOT/bin:$GOPATH/bin
 go version
-go install honnef.co/go/tools/cmd/staticcheck@2021.1.1 || true
+go install honnef.co/go/tools/cmd/staticcheck@2022.1.2 || true
 
 uname -a
 echo "$PATH"
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -1182,10 +1182,6 @@ func TestVersion(t *testing.T) {
 }
 
 func TestStaticcheck(t *testing.T) {
-	if goversion.VersionAfterOrEqual(runtime.Version(), 1, 18) {
-		//TODO(aarzilli): remove this before version 1.8.0 is released
-		t.Skip("staticcheck does not currently support Go 1.18")
-	}
 	_, err := exec.LookPath("staticcheck")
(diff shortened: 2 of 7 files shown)


aarzilli:
IMO a couple of duplicate declarations are preferrable to giving people the impression that a C compiler is needed to build delve. Especially since our use of the C compiler is almost non-existent. Also I think this conversion highlighted a problem previously concealed by the C type conversion: shouldn't fn_addr be unsigned long?

derekparker:
> IMO a couple of duplicate declarations are preferrable to giving people the impression that a C compiler is needed to build delve. Especially since our use of the C compiler is almost non-existent. Also I think this conversion highlighted a problem previously concealed by the C type conversion: shouldn't fn_addr be unsigned long?

Thanks for this, and yes I agree about the duplication being preferable. Also you're correct about the incorrect type, we should be using an unsigned long there, that's my mistake.

derekparker:
Want to tackle that change in the PR or follow up?

derekparker (approved):
LGTM

aarzilli:
> Want to tackle that change in the PR or follow up?

I'm not set up to compile bpf programs with clang.

PR #3073, opened 2022-07-23
proc: support gnu_debuglink section

### LINKED ISSUE #2150: Match symbols using .gnu_debuglink section
1. What version of Delve are you using (`dlv version`)? 1.5.0
2. What version of Go are you using? (`go version`)? 1.15
3. What operating system and processor architecture are you using? linux/amd64

### What did you do?
Delve currently matches ELF symbols by either `.note.gnu.build-id` or by just using the base file name. Using the file name is not what GDB does AFAIK/ gdb uses a `.gnu_debuglink` section instead, which also includes a CRC to make sure that the found debug file is the right one, just using the base file name can match a symbols file for a different version of the binary.

See https://github.com/go-delve/delve/blob/328cf87808822693dc611591519689dcd42696a3/pkg/proc/bininfo.go#L1041
And [Separate Debug Files (Debugging with GDB)](https://sourceware.org/gdb/current/onlinedocs/gdb/Separate-Debug-Files.html)

P.S. It doesn't seem like that is a convenient way to set build-id automatically to a CRC like when using gcc/binutils in Go.

### What did you expect to see?
I expect symbol lookup to match with gdb.

### What did you see instead?
It looks like delve doesn't implement symbol lookup the same as gdb.

aarzilli:
Load separate debug info using gnu_debuglink.

Fixes #2150

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -13,6 +13,7 @@ import (
 	"go/ast"
 	"go/token"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

gnurizen:
FWIW, TIL gnuDebugLinkCRC and the go stdlib crc32.ChecksumIEEE produce the same result.

PR #3075, opened 2022-07-23
pkg/locspec: support exact match in SubstitutePath

### LINKED ISSUE #3074: substitutePath accepts only directories as arguments
We've been looking at improving [Encore](https://github.com/encoredev/encore)'s debugging support. It's a bit tricky since it heavily uses code generation along with overlay files to reduce boilerplate.

The way it works is that Encore injects a file into a user-built package (say located at `/myapp/mypkg`) by writing a file in a temporary directory (say `/tmp/foo/generated.go`) and then invokes `go build -overlay`, specifying `{"Replace": {"/myapp/mypkg/generated.go": "/tmp/foo/generated.go"}}`.

The `go` command then reads the overlay and builds the package as if the file at `/tmp/foo/generated.go` actually existed in `/myapp/mypkg`. This works great, except that when debugging this file does not actually exist on disk in that location, which causes editors to get very confused and naturally displays a "file not found" error when stepping into the generated code.

I tried to work around this by passing in `"substitutePath": [{"from": "/myapp/mypkg/generated.go": "/tmp/foo/generated.go"}]`. Unfortunately this does not work, as `substitutePath` only accepts directories as entries. And because encore only generates the single file, there's no single folder that contains all of the files, so we're kind of stuck.

If `substitutePath` could be changed to also accept exact matches in the form of file paths, this works great. I've implemented this change and it's very minor, provided you'd be open to it. (See PR #3075)

eandre:
In order to support substituting a single file within
a larger package, and not substituting a directory wholesale,
this commit introduces the ability for SubstitutePath
to replace individual file paths if there is an exact match.

This behavior is backwards-compatible with the previous
configuration.

Fixes #3074

### DIFF
--- a/pkg/locspec/locations.go
+++ b/pkg/locspec/locations.go
@@ -492,6 +492,12 @@ func SubstitutePath(path string, rules [][2]string) string {
 		from := crossPlatformPath(r[0])
 		to := r[1]
 
+		// If we have an exact match, use it directly.
+		if path == from {
+			return to
+		}
+
+		// Otherwise check if it's a directory prefix.
 		if !strings.HasSuffix(from, separator) {
 			from = from + separator
 		}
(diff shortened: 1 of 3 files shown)


pkg/locspec/locations.go
```diff
@@ -488,10 +488,20 @@ func SubstitutePath(path string, rules [][2]string) string {
 	if strings.Contains(path, "\\") { //dependent on the path
 		separator = "\\"
 	}
+
+	// First try with exact match
```
aarzilli, line 492: I think this check should be part of the normal loop, because we don't do best match.

eandre: Sounds good, I've updated the PR to reflect this behavior

eandre:
> BTW the package name is locspec, not logspec.

Whoops, fixed, thanks.

aarzilli (approved):
LGTM, thank you

derekparker (approved):
LGTM, thanks!

PR #3079, opened 2022-07-27
Refactor windows framework

qmuntal:
This PR refactors how windows support is implemented in `pkg/proc/native` so it is easier to support new architectures in the future, like for example ARM64 (see #3063).

There are two major changes:
- Split architecure-agnostic code from amd64-specific code. In practice all windows architectures share the same APIs and workflows. The only thing that changes are data structures, such as `_Context`.
-  Define helper methods so arch-agnostic code can manipulate arch-specific, e.g. `nativeThread.setPC` and `registers`.

@aarzilli


derekparker (approved):
LGTM

aarzilli (approved):

aarzilli:
@derekparker what if we merged the windows/arm64 support patch, everything except the change to support_sentinel.go, until we have a builder? And we make no promises that we won't break the build on windows/arm64 when we merge changes.

derekparker:
@aarzilli I'm ok with that, I think we should explicitly call that out in docs somewhere until we have a builder. Can we add remote builders via TeamCity? Essentially what I'm thinking is... could Microsoft donate a Windows/arm64 machine to run CI on? (cc @qmuntal )

qmuntal:
> @derekparker what if we merged the windows/arm64 support patch, everything except the change to support_sentinel.go, until we have a builder? And we make no promises that we won't break the build on windows/arm64 when we merge changes.

We can go a little bit further, and gate windows/arm64 behind a build tag so people can start trying it and reporting bugs without forking. We could also add a CI job that cross-compiles from windows/amd64 to windows/arm64 without running tests. at least we will catch compilation errors.

qmuntal:
> @aarzilli I'm ok with that, I think we should explicitly call that out in docs somewhere until we have a builder. Can we add remote builders via TeamCity? Essentially what I'm thinking is... could Microsoft donate a Windows/arm64 machine to run CI on? (cc @qmuntal )

I'll discuss this with my team and get back to you!

aarzilli:
I already asked nd, previously (about some other port), about adding remote agents and he said it should work on anything that has a jvm port: https://www.jetbrains.com/help/teamcity/install-teamcity-agent.html

PR #3080, opened 2022-07-27
pkg/proc/internal/ebpf: Fix size of ebpf type for fn_addr

derekparker:
Must have been an issue overlooked in the initial implementation but
we should be using a 64-bit wide type to store the function address.


derekparker:
I know CI doesn't run tests against this but I've tested locally and everything works.

Also @aarzilli you mentioned not having a setup to build the ebpf object, but in the Makefile there's a command to run the build in a container which should have all deps, are you able to run that?

aarzilli:
I forgot about that so I didn't try. But it turns out that it doesn't work. ubuntu 21.04 is already 404 and bumping to 22.04:

```
Reading state information...
E: Unable to locate package linux-headers-5.11.0-40
E: Couldn't find any package by glob 'linux-headers-5.11.0-40'
E: Couldn't find any package by regex 'linux-headers-5.11.0-40'
```

bumping down to the lts (20.04):

```
Reading state information...
E: Unable to locate package libbpf-dev
```

aarzilli (approved):

PR #3081, opened 2022-07-27
pkg/proc/internal/ebpf: Fix handling of entry / return

derekparker:
This patch removes the old error-prone way of tracking
whether the tracepoint is for a function entry or
return. Instead of trying to guess, let the data structure
simply tell us directly.


aarzilli (approved):
LGTM but merge conflict

derekparker:
Rebased.

derekparker:
There's still a bug lurking here where we now don't correlate function entry & exit with the same goroutine, but I will fix that in a follow up patch. There's still work to do on the ebpf backend and with the trace subcommand in general.

Issue #3082, opened 2022-07-28
service/dap: adding empty string support for substitutePath when dealing with relative paths


RuijieC-dev:
Please answer the following before submitting your issue:
1. What version of Delve are you using (`dlv version`)?
Version: 1.8.2
Build: $Id: dbb493ec14d1e7753504d016b1e1ef1665b75b16 $
2. What version of Go are you using? (`go version`)?
go version go1.18.1 linux/amd64
3. What operating system and processor architecture are you using?
x86_64, centos7

I'm trying to use substitutePath field in the launch config to convert relative paths to absolute paths but substitutePath doesn't support '.' to replace relative paths.
The expectation is having relative paths like: 
- `my/path`
- `another/my/path`
- `/this/is/abs`

Using a mapping like: from:`/my/abs/path` to: `.` paths expect be the following:
-`my/path` => `/my/abs/path/my/path`
- `another/my/path` => `/my/abs/path/another/my/path`
- `/this/is/abs` => `/this/is/abs`

PR #3088, opened 2022-08-01
service/dap: Add support for empty string in substitutePath

RuijieC-dev:
This changes adds the support to replace relative paths sources with "" as representing current directory.
For example:
Rule like '{from: "", to: "/my/project"}' will map path "src/my/code.go" into "/my/project/src/my/code.go"
Rule like '{from: "/my/project", to: ""}' will map path "/my/project/src/my/code.go" into "src/my/code.go"

Solves #3082

### DIFF
--- a/pkg/config/split.go
+++ b/pkg/config/split.go
@@ -41,6 +41,7 @@ func SplitQuotedFields(in string, quote rune) []string {
 			} else if unicode.IsSpace(ch) {
 				r = append(r, buf.String())
 				buf.Reset()
+				state = inSpace
 			} else {
 				buf.WriteRune(ch)
 			}
--- a/pkg/config/split.go
+++ b/pkg/config/split.go
@@ -60,7 +60,7 @@ func SplitQuotedFields(in string, quote rune) []string {
 		}
 	}
 
-	if buf.Len() != 0 {
+	if state == inField || buf.Len() != 0 {
 		r = append(r, buf.String())
 	}
 
--- a/pkg/locspec/locations.go
+++ b/pkg/locspec/locations.go
@@ -498,13 +498,19 @@ func SubstitutePath(path string, rules [][2]string) string {
 		}
 
 		// Otherwise check if it's a directory prefix.
-		if !strings.HasSuffix(from, separator) {
+		if from != "" && !strings.HasSuffix(from, separator) {
 			from = from + separator
 		}
-		if !strings.HasSuffix(to, separator) {
+		if to != "" && !strings.HasSuffix(to, separator) {
 			to = to + separator
 		}
-		if strings.HasPrefix(path, from) {
+
+		//  If 'from' is a empty string and path is not absolute, prepende 'to'.
+		if from == "" && !filepath.IsAbs(path) {
+			return strings.Replace(path, from, to, 1)
+		}
+
+		if from != "" && strings.HasPrefix(path, from) {
 			return strings.Replace(path, from, to, 1)
 		}
 	}
--- a/pkg/locspec/locations.go
+++ b/pkg/locspec/locations.go
@@ -505,7 +505,7 @@ func SubstitutePath(path string, rules [][2]string) string {
 			to = to + separator
 		}
 
-		//  If 'from' is a empty string and path is not absolute, prepende 'to'.
+		// Expand relative paths with the specified prefix
 		if from == "" && !filepath.IsAbs(path) {
 			return strings.Replace(path, from, to, 1)
 		}
--- a/service/dap/types.go
+++ b/service/dap/types.go
@@ -189,7 +189,8 @@ type LaunchAttachCommonConfig struct {
 }
 
 // SubstitutePath defines a mapping from a local path to the remote path.
-// Both 'from' and 'to' must be specified and non-null.
+// Both 'from' and 'to' must be specified and non-null but it can be empty string.
+// Mapping with 'to' as empty string can be used to expand relative paths.
 type SubstitutePath struct {
 	// The local path to be replaced when passing paths to the debugger.
 	From string `json:"from,omitempty"`
--- a/service/dap/types.go
+++ b/service/dap/types.go
@@ -189,8 +189,9 @@ type LaunchAttachCommonConfig struct {
 }
 
 // SubstitutePath defines a mapping from a local path to the remote path.
-// Both 'from' and 'to' must be specified and non-null but it can be empty string.
-// Mapping with 'to' as empty string can be used to expand relative paths.
+// Both 'from' and 'to' must be specified and non-null.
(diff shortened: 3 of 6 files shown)


aarzilli (commented):
Doesn't work on windows. Also I think it we should do this with the empty string rather than `.`.

RuijieC-dev:
Fixing for windows mapping as well. Changes coming soon.
The intention is to use `.` so it's clearer we are dealing with relative paths. Can do it with empty string as well but that means empty string check needs to be removed. https://github.com/go-delve/delve/blob/master/service/dap/types.go#L211

pkg/locspec/locations_unix_test.go
```diff
@@ -0,0 +1,22 @@
+package locspec
```
aarzilli, line 1: You can't use build guards like this and the 'unix' build tag doesn't exist on most versions of go. You should check runtime.GOOS and skip as appropriate.

aarzilli:
Besides the test failures, in my opinion this should support the empty string rather than using `"."`.

pkg/locspec/locations.go
```diff
@@ -504,6 +504,17 @@ func SubstitutePath(path string, rules [][2]string) string {
...
 			to = to + separator
 		}
+
+		// If 'from' is '.' and path is relative, prepende 'to'.
+		if from == "."+separator && !filepath.IsAbs(path) {
+			return strings.Replace(path, "", to, 1)
+		}
+
+		// If 'to' is '.' and path is absolute. Delete matching `from` if any.
+		if to == "."+separator && filepath.IsAbs(path) {
```
aarzilli, line 514: I don't understand why this should be the case. Why shouldn't the mapping `non/absolute/path -> .` work?

pkg/locspec/locations_unix_test.go
RuijieC-dev: Sounds good. Changing the test accordingly. Thanks for the suggestion.

pkg/locspec/locations.go
RuijieC-dev: I didn't thought about relative paths mapping use cases. Thanks pointing it out. Changes coming to support it.

RuijieC-dev:
Changed logic to support empty string instead of "."

aarzilli (approved):
LGTM cc @polinasok

RuijieC-dev:
Cannot reproduce TestAttachStopOnEntry test failure in Mac machine. Looking at the logs looks like a flaky case. Rebase and re-triggering the tests.

aarzilli:
Don't worry about that.

pkg/locspec/locations.go
```diff
@@ -498,13 +498,19 @@ func SubstitutePath(path string, rules [][2]string) string {
...
+		if from != "" && !strings.HasSuffix(from, separator) {
 			from = from + separator
 		}
-		if !strings.HasSuffix(to, separator) {
+		if to != "" && !strings.HasSuffix(to, separator) {
 			to = to + separator
 		}
-		if strings.HasPrefix(path, from) {
+
+		//  If 'from' is a empty string and path is not absolute, prepende 'to'.
```
polinasok, line 508: "prepend"

Although the comment just says the exact same thing as the code. How about "Expand relative paths with the specified prefix"?

service/dap/types.go
```diff
@@ -189,7 +189,7 @@ type LaunchAttachCommonConfig struct {
 }
 
 // SubstitutePath defines a mapping from a local path to the remote path.
-// Both 'from' and 'to' must be specified and non-empty.
+// Both 'from' and 'to' must be specified and non-null.
```
polinasok, line 192: Please add a comment what empty now means here

polinasok (commented):
This is great and very timely! Just had another user request this, so you saved me the trouble :) Thank you!

service/dap/types.go
```diff
@@ -208,10 +211,10 @@ func (m *SubstitutePath) UnmarshalJSON(data []byte) error {
 		}
 		return err
 	}
-	if tmp.From == "" || tmp.To == "" {
+	if tmp.From == nil || tmp.To == nil {
```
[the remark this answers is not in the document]
aarzilli: This isn't very useful for the cli, it can already be run from a different directory and will use relative paths.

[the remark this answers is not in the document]
RuijieC-dev: Yes, the main use case is to achieve relative path support for delve dap using `launch.json`. 

rules specified via the DEBUG CONSOLE should in theory work in the same way but I don't know how empty string can be passed without delimiters.
```
dlv config substitutePath /prefix/  ""     // This will delete the `/prefix/` mapping only.
```

As @aarzilli mentioned, in CLI this isn't a problem considering the binary can be already run in a different directory using `--wd`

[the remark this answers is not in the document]
RuijieC-dev: Added new commit to support for `dlv config substitutePath`.
```
dlv config substitutePath /a/path ""  // Used to be equal doing dlv config substitutePath /a/path. Now it's updating rule with empty string
dlv config substitutePath "" /a/path // Was already working.
dlv config substitutePath "" // Used be error always. Now it's an error when there's no mapping to empty string. Otherwise deletes empty string mapping
```

pkg/config/split.go
```diff
@@ -60,7 +60,7 @@ func SplitQuotedFields(in string, quote rune) []string {
 		}
 	}
 
-	if buf.Len() != 0 {
+	if state == inField || buf.Len() != 0 {
```
aarzilli, line 63: you also need to add a `state = inSpace` after the `buf.Reset()` in `case inField:`

RuijieC-dev: Addressed it and added more test cases to cover it.

aarzilli (approved):
LGTM

service/dap/types.go
```diff
@@ -189,7 +189,8 @@ type LaunchAttachCommonConfig struct {
 }
 
 // SubstitutePath defines a mapping from a local path to the remote path.
-// Both 'from' and 'to' must be specified and non-null.
+// Both 'from' and 'to' must be specified and non-null but it can be empty string.
+// Mapping with 'to' as empty string can be used to expand relative paths.
```
polinasok, line 193: Let's say

```
// Both 'from' and 'to' must be specified and non-null. 
// Empty values can be used to add or remove absolute path prefixes when mapping.
// For example, mapping with empy 'to' can be used to work with binaries with trimmed paths.
```

polinasok (changes requested):
Thank you for adding all the tests!
Just one minor comment and this will be good to go.

polinasok (approved):
TestStepInstruction failure is due to #2757 and not relatedd.

PR #3090, opened 2022-08-03
pkg/dwarf/line: Fix parsing file table for DWARFv5

derekparker:
As we parse this informatin in the loop we must take care to assemble
things correctly. In this situation when we encounter a file name,
the dir index is -1, then subsequently we get the correct dir index
for that file and can put them together. Previously we were adding the
file and then the directory location to the file list instead of
correctly concatenating them, resulting in an incorrect file list making
indexing into the list return incorrect results later on.

### DIFF
--- a/pkg/dwarf/line/line_parser.go
+++ b/pkg/dwarf/line/line_parser.go
@@ -278,6 +278,7 @@ func parseFileEntries5(info *DebugLineInfo, buf *bytes.Buffer) bool {
 	}
 	fileCount, _ := util.DecodeULEB128(buf)
 	info.FileNames = make([]*FileEntry, 0, fileCount)
+	var lastp string
 	for i := 0; i < int(fileCount); i++ {
 		fileEntryFormReader.reset()
 		for fileEntryFormReader.next(buf) {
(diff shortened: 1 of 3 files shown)


pkg/dwarf/line/line_parser.go
```diff
@@ -307,16 +308,20 @@ func parseFileEntries5(info *DebugLineInfo, buf *bytes.Buffer) bool {
 				// not implemented
 			}
 
-			if info.normalizeBackslash {
-				p = strings.ReplaceAll(p, "\\", "/")
+			if p == "" {
```
aarzilli, line 311: I'm thinking that this is still wrong. Compilers are free to order the contents as:

    DW_LNCT_path DW_LNCT_directory_index

and
    
    DW_LNCT_directory_index DW_LNCT_path

This works in the first case because p will be "" and diridx will be correct, when this loop is examining the DW_LNCT_directory_index entry. It won't work in the second case because we don't save the directory index anywhere.

I think that the code that was at the end of the `for fileEntryFormReader.next(buf)`, starting with the `if info.normalizeBackslash` statement, should be moved outside of the loop, which will guarantee that we add exactly `fileCount` entries to the table (consequently the variables `p` and `diridx` also need to be oustide of that loop).

derekparker: Ack, good call. Done.

pkg/dwarf/line/state_machine.go
```diff
@@ -231,14 +231,15 @@ func (lineInfo *DebugLineInfo) stateMachineFor(basePC, pc uint64) *StateMachine
...
-		// there isn't one or it's already past pc try to clone the cached state
+		// there isn't one, or it's already past pc try to clone the cached state
 		// machine stopped at the entry point of the function.
 		// As a last resort start from the start of the debug_line section.
 		sm = lineInfo.lastMachineCache[basePC]
 		if sm == nil || sm.lastAddress >= pc {
 			sm = lineInfo.stateMachineForEntry(basePC)
 			lineInfo.lastMachineCache[basePC] = sm
 		}
+		sm = sm.copy()
```
aarzilli, line 242: This is wrong, the intent is to reuse the last machine and make it advance so that when we call PCToLine many times to get the line numbers for a disassembly it isn't quadratic.

aarzilli: Sorry, I didn't notice it yesterday.

derekparker: Ack, yeah I forgot to revert this change. I was wondering for a moment if this cache was causing issues but it was a red herring. I’ll revert tomorrow.

derekparker: Done.

aarzilli (approved):
LGTM

PR #3091, opened 2022-08-03
:* Improve trace subcommand output

derekparker:
This patch improves the output of the trace subcommand by
adding better line breaks, adding goroutine info to the
return statement, and removing unnecessary output.


aarzilli (approved):
LGTM

PR #3100, opened 2022-08-09
_scripts: fix CI on go1.19/linux/386

aarzilli:
Go 1.19 is broken on linux/386 with some C compilers, this is a
workaround for our build script. See:

https://github.com/golang/go/issues/52919

Also fix TestBuild if the first message reported by delve is not the
message that indicates the server is listening.


derekparker (approved):
LGTM

PR #3105, opened 2022-08-13
proc: use stdlib to compute CRC for gnu_debuglink section

aarzilli:
The standard library package has a the CRC algorithm used for
gnu_debuglink, use that instead of our own implementation.


derekparker (approved):
LGTM

PR #3107, opened 2022-08-15
documentation: add missing links for expr.md

katcipis:
Noticed some links missing on the docs for Documentation/cli/expr.md.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -526,9 +526,7 @@ Print contents of CPU registers.
 
 	regs [-a]
 
-Argument -a shows more registers. Individual registers can also be displayed by 'print' and 'display'.
-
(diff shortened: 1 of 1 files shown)


aarzilli:
That documentation page is automatically generated from pkg/terminal/command.go by WriteMarkdown in pkg/terminal/docgen.go and uses the same source as the `help` command.
We don't want the `help` command to print markdown on the console so if you want to do this you have to change WriteMarkdown to recognize things that can be documentation links and rewrite them as markdown links (or change `help` to strip markdown links from its input, I think changing WriteMarkdown is easier, also because we can statically check that it's working correctly).

aarzilli:
PS. I just noticed that something like this already exists in WriteMarkdown, replaceDocPaths, it just doesn't work with those because they don't start with $GOPATH. Really we should remove $GOPATH everywhere and recognize things that start with `Documentation/`.

Issue #3094, opened 2022-08-08
Send DAP OutputEvents on stdout/stderr of debugged process when console is `internalConsole`


gfszr:
**Go Version**: 1.18
**Delve Version**: 1.8.2

#### Current status
DAP allows sending `stdout`/`stderr` outputs back to the DAP client, by which it is shown on the debug console of the IDE. Delve spawns the debuggees with its own stdout/stderr, so the output is missing from the normal debug console. This is the desired behavior in case the terminal being used is either `integratedTerminal` or `externalTerminal`, but in the case of `internalConsole`, in order to allow showing the output to the user, clients required to somehow still redirect output (For example, [nvim-dap-go](https://github.com/leoluz/nvim-dap-go/blob/main/lua/dap-go.lua#L57)).

#### Desired status
Delve would send [OutputEvent](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output)s of data from the debuggee's `stdout`/`stderr` if the console given is `internalConsole`, which would allow the DAP client to easily present it.

aarzilli:
The problem with this is that some program will behave differently if their stdout/stderr is not a tty. It is unfortunate that DAP was designed like this.

gfszr:
@aarzilli Yet, other debug adapters do the same, even if running with a tty - debugpy for instance.

gfszr:
The specific case in [debugpy](https://github.com/microsoft/debugpy) can be seen [here](https://github.com/microsoft/debugpy/blob/main/src/debugpy/adapter/clients.py#L221):
```python
 redirecting = arguments.get("console") == "internalConsole"
```
Redirection is decided in case the user specifies `internalConsole` as its terminal - by which this means that the user requested output to be redirected in the console. I agree this shouldn't be default behavior unless the user requested `internalConsole`, so perhaps updating this issue with "redirect output when console defined as internalConsole` suffices? @aarzilli

gfszr:
I revised the issue to only send such `OutputEvent`'s when launch configuration is `internalConsole`.

aarzilli:
This was implemented by #3253

PR #3108, opened 2022-08-15
Redirect if internal console

gfszr:
**This PR is currently as draft to better improve implementation and pass tests.**
This PR partially implements https://github.com/go-delve/delve/issues/3094, by implementing redirection of `stdout` and `stderr` to the debug console via DAP's `OutputEvent`'s on Linux. 

It partially does so as current debug backend API support receiving a path to a redirected file only, and not a `io.Writer` as an example. It also supports only the `native` backend and not other backends. 

Achieving full support would require adding a generalized API so that we can pass an `io.Writer` to the native backend nicely, but this can be done later on and be discussed more, while Linux support can be easily implemented.

### DIFF
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -326,7 +326,7 @@ func openRedirects(redirects [3]string, foreground bool) (stdin, stdout, stderr 
 			return dflt
 		}
 		var f *os.File
-		f, err = os.Create(path)
+		f, err = os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0o666)
 		if f != nil {
 			toclose = append(toclose, f)
 		}
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -273,11 +273,9 @@ const (
 	maxStringLenInCallRetVars = 1 << 10 // 1024
 )
 
-var (
-	// Max number of goroutines that we will return.
-	// This is a var for testing
-	maxGoroutines = 1 << 10
-)
+// Max number of goroutines that we will return.
+// This is a var for testing
+var maxGoroutines = 1 << 10
 
 // NewServer creates a new DAP Server. It takes an opened Listener
 // via config and assumes its ownership. config.DisconnectChan has to be set;
@@ -805,7 +803,24 @@ func (s *Session) logToConsole(msg string) {
 		Body: dap.OutputEventBody{
 			Output:   msg + "\n",
 			Category: "console",
-		}})
+		},
+	})
+}
+
+type debugConsoleLogger struct {
+	session  *Session
+	category string
+}
+
+func (l *debugConsoleLogger) Write(p []byte) (int, error) {
+	l.session.send(&dap.OutputEvent{
+		Event: *newEvent("output"),
+		Body: dap.OutputEventBody{
+			Output:   string(p),
+			Category: l.category,
+		},
+	})
+	return len(p), nil
 }
 
 func (s *Session) onInitializeRequest(request *dap.InitializeRequest) {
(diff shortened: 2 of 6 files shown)


pkg/proc/native/proc.go
```diff
@@ -326,7 +326,7 @@ func openRedirects(redirects [3]string, foreground bool) (stdin, stdout, stderr
 			return dflt
 		}
 		var f *os.File
-		f, err = os.Create(path)
+		f, err = os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0o666)
```
aarzilli, line 329: This is a backwards incompatible change because you are not passsing O_TRUNC.

service/dap/server.go
```diff
@@ -273,11 +273,9 @@ const (
 	maxStringLenInCallRetVars = 1 << 10 // 1024
 )
 
-var (
-	// Max number of goroutines that we will return.
-	// This is a var for testing
-	maxGoroutines = 1 << 10
-)
+// Max number of goroutines that we will return.
```
aarzilli, line 276: Avoid unrelated reformats, they pollute blames and make merge conflicts unnecessarily harder.

service/dap/server.go
```diff
@@ -1049,6 +1071,32 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 		return
 	}
 
+	func() {
```
aarzilli, line 1074: Why?

service/dap/stdio_pipe.go
```diff
@@ -0,0 +1,13 @@
...
+// +build !linux
+
+package dap
+
+import (
+	"errors"
+)
+
+func generateStdioTempPipes() (res [2]string, err error) {
+	err = errors.New("Unimplemented")
```
aarzilli, line 11: Merging this with just a linux implementation would be weird. This is enough of a prominent feature that having it implemented for one operating system implies that we have to implement it for at least the more important ones. I think it needs to support Windows and macOS.

service/dap/server.go
```diff
@@ -1049,6 +1071,32 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+		if args.Console == "internalConsole" && runtime.GOOS == "linux" && (args.Backend == "default" || args.Backend == "native") {
```
aarzilli, line 1075: It's weird that using internalConsole on an unsupported configuration fails silently. It's also weird that all values of console that aren't internalConsole are accepted as valid.

service/dap/server.go
gfszr: My editor auto-formats using `gofumpt`, so it formatted some extra editted files. Will revert such changes.

pkg/proc/native/proc.go
gfszr: Will add it - I still need to check whether `O_TRUNC` has any bad effect when opening a fifo.

service/dap/stdio_pipe.go
gfszr: Implementing it to all platforms should be relatively easier if we can change the `Redirects` to not only be paths to `stdout`/`stderr`, but `io.Writer`s. I'll implement complete support for `native`, which should be relatively easy to do soin that backend.

service/dap/server.go
gfszr: The current behavior is that `console` is not checked, so it can also be completely invalid and `dlv dap` would accept it, so I'm not sure we necessarily have to validate the `console` argument.

gfszr:
Thank you @aarzilli for the quick reply! 🙂 
Added a few comments, will address your comments.

service/dap/server.go
aarzilli: If you refer to this file, that pattern is to execute a defer, not to group anything.

service/dap/server.go
aarzilli: Suppose someone misspells internlaConsole, how do they notice?

gfszr: Usually, such syntax can be checked in VSCode itself - it can present an error such as: `"internalConsole" is not a valid "console" value".

aarzilli: Not everyone uses vscode with dap. I thought that was the whole point of this.

gfszr: Hm, true. I am not familiar with how every DAP client handles this, but adding an allowed list of declared DAP consoles (I think it is standardized) should be easy enough.

aarzilli: it should just be the consoles that we support.

gfszr:
Yes - currently having some difficulties on developing and testing on Windows and macOS, as requested in review.

derekparker:
Closing due to age.

PR #3110, opened 2022-08-16
*: switch to int64 for goroutine IDs

aarzilli:
Go 1.20 switched to uint64 to represent goroutine IDs, we can't
actually follow suit because we have allowed clients to use -1 to refer
to the currently selected goroutine, however we should at least switch
to int64 and also update the rtype check to accept the 1.20 type.

### DIFF
--- a/_scripts/rtype-out.txt
+++ b/_scripts/rtype-out.txt
@@ -21,7 +21,7 @@ type eface struct {
 
 type g struct {
 	sched gobuf
-	goid int64
+	goid int64|uint64
 	gopc uintptr
 	startpc uintptr
 	waitsince int64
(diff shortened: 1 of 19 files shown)


aarzilli:
Speaking of go1.20, what do we call the next release of delve? Is 1.10 going to be confusing, should we skip to 1.20?

derekparker (approved):
LGTM

derekparker:
> Speaking of go1.20, what do we call the next release of delve? Is 1.10 going to be confusing, should we skip to 1.20?

Hm, that's interesting to consider. It might be worth just skipping ahead to align with the Go version. Perhaps we should open this to a wider discussion? It would be nice to have the versions be 1:1.

PR #3127, opened 2022-08-28
Add Debugger to UIs list in EditorIntegration

emad-elsaid:
Added Debugger to the list of UIs as suggested [here](https://github.com/go-delve/delve/discussions/3126#discussioncomment-3491801)


aarzilli (approved):
LGTM

PR #3134, opened 2022-09-10
go.mod,proc: upgrade golang.org/x/sys/unix

aarzilli:
Upgrade golang.org/x/sys/unix so that it contains go.dev/cl/419915, a
fix for FreeBSD's ptrace functions that used *byte to represent
addresses on the tracee address space.


derekparker (approved):
LGTM

Issue #3137, opened 2022-09-14
goroutine wait time is incorrect


qingyunha:
https://github.com/go-delve/delve/blob/9bcccf81ed7c93a6f1ea4bfc0d905178e6db8e81/pkg/terminal/command.go#L1079

`g.WaitSince` is a CLOCK_MONOTONIC  nanotime，and it can't used in`time.Unix`.

There is a  `runtime.runtimeInitTime` , we may use it as a base time.

qingyunha:
Can we use `g.WatiSince - runtime.runtimeInitTime` , Then user can use the process start time to calculate the wait time ?

derekparker:
This should be closed via #3139 if not please reopen.

PR #3139, opened 2022-09-19
terminal: do not use WaitSince as unix time

aarzilli:
g.waitsince is the output of runtime.nanotime and represents a
monotonic clock which can not be converted directly into unix time. A
better fix would be to convert it to a time.Duration by reading the
current value of runtime.nanotime. This is complicated, however,
because on some systems (for example macOS) the current value of
runtime.nanotime can only be read by making a system call.

Updates #3137


derekparker (approved):
LGTM

PR #3141, opened 2022-09-20
service/dap: show full value when evaluating log messages

suzmue:
When evaluating variables in a log message, do not truncate the string representation.

Updates golang/vscode-go#2447


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3142, opened 2022-09-21
service/debugger: merge attachErrorMessage, verifyBinaryFormat

aarzilli:
These functions don't need to be conditionally compiled and they share
substanatial amounts of code in all their OS-specific versions.


derekparker (approved):
LGTM

PR #3144, opened 2022-09-21
cmd/dlv: disable TestTraceMultipleGoroutines on FreeBSD

aarzilli:
This test seems to fail frequently on FreeBSD.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -1021,6 +1021,11 @@ func TestTrace(t *testing.T) {
 }
 
 func TestTraceMultipleGoroutines(t *testing.T) {
+	if runtime.GOOS == "freebsd" {
+		//TODO(aarzilli): investigate further when the FreeBSD backend is more stable.
+		t.Skip("temporarily disabled due to issues with FreeBSD in Delve and Go")
+	}
+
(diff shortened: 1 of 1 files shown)


derekparker:
Do we know any kind of root cause?

aarzilli:
I took a look at freebsd backend back in late july (hence go.dev/cl/419915) and there are multiple problems with the way trapWait and singleStep are implemented. It could be that, or it could be other problems in other parts of the backend. I intend to get back to it and fix it but I haven't had time to do it yet.

cmd/dlv/dlv_test.go
```diff
@@ -1020,6 +1020,10 @@ func TestTrace(t *testing.T) {
 }
 
 func TestTraceMultipleGoroutines(t *testing.T) {
+	if runtime.GOOS == "freebsd" {
+		t.Skip("broken")
```
derekparker, line 1024: Let's change this to something more like "temporarily disabled due to issues with FreeBSD in Delve and Go".

derekparker: Let's also add a TODO to investigate further.

aarzilli: Done.

derekparker (approved):
LGTM

PR #3145, opened 2022-09-22
proc: fix openSeparateDebugInfo backwards compatibility path

aarzilli:
Fix openSeparateDebugInfo backwards compatibility path to actually be
backwards compatible.


derekparker (approved):
LGTM

PR #3146, opened 2022-09-22
proc: fix cast to empty interface type

aarzilli:
Fix cast to empty interface type. This is a workaround for a
discrepancy between how type names are serialized by the compiler vs
how type expressions are formatted by go/printer.


derekparker (approved):
LGTM

PR #3147, opened 2022-09-23
proc/native: fix support_sentinel_darwin.go name

aarzilli:
The file incorrectly contains a space, which means it does nothing
besides breaking 'make build'.


derekparker (approved):
LGTM

PR #3150, opened 2022-09-29
proc: remove old method to resolve the type of an interface to a DIE

aarzilli:
Before 1.11 we used to read a bunch of runtime structures to determine
the runtime type of an interface variable. This had significant
dependencies on private structs of the Go runtime and the code is
broken for versions of Go after 1.17.

Remove all this code, since it is no longer used and doesn't work with
newer versions of Go anyway.


derekparker (approved):
LGTM

PR #3165, opened 2022-10-10
proc,proc/native: adds ability to automatically debug child processes

aarzilli:
Adds the ability to automatically debug child processes executed by the
target to the linux native backend.

This commit does not contain user interface or API to access this
functionality.

Updates #2551


pkg/proc/native/proc.go
```diff
@@ -52,12 +54,30 @@ func newProcess(pid int) *nativeProcess {
...
+// newChildProcess is like newProcess but uses the same ptrace thread as dbp.
+func newChildProcess(dbp *nativeProcess, pid int) *nativeProcess {
+	(*dbp.ptraceRefCnt)++
+	return &nativeProcess{
+		pid:            pid,
+		threads:        make(map[int]*nativeThread),
+		breakpoints:    proc.NewBreakpointMap(),
+		firstStart:     true,
+		os:             new(osProcessDetails),
+		ptraceChan:     dbp.ptraceChan,
```
derekparker, line 74: Perhaps we should just store a reference to the root process instead of copying over individual fields? We could hide the reference behind an interface to limit what can be accessed.

pkg/proc/native/proc_linux.go
```diff
@@ -410,6 +447,34 @@ func (dbp *nativeProcess) trapWaitInternal(pid int, options trapWaitOptions) (*n
...
+			tgt, err := dbp.initialize(findExecutable("", dbp.pid), cctx.DebugInfoDirs)
+			if err != nil {
+				_ = dbp.Detach(false)
+				return nil, err
+			}
+			cctx.AddTarget(tgt)
+			if halt {
+				return nil, nil
+			}
+			//TODO(aarzilli): if we want to give users the ability to stop the target
```
derekparker, line 470: nit: Add a space after `//`

pkg/proc/interface.go
```diff
@@ -134,3 +139,18 @@ func (cctx *ContinueOnceContext) GetManualStopRequested() bool {
...
 }
+
+// NumProcs returns the number of processes in the target group being resumed.
+func (cctx *ContinueOnceContext) NumProcs() int {
+	return len(cctx.grp.targets)
+}
+
+// Proc returns the i-th process in the group.
+func (cctx *ContinueOnceContext) Proc(i int) ProcessInternal {
+	return cctx.grp.targets[i].proc
```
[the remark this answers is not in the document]
derekparker: Also, while I'm thinking of it, is it worth returning the `ProcessInternal` interface since we typecast pretty much every time we call this?

derekparker (changes requested):
Quick initial pass done, sorry for the delay.

pkg/proc/interface.go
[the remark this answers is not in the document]
aarzilli: It is always downcast, but to different things. Native downcasts to nativeProcess, gdbserial to gdbProcess. I don't know how to make it better. I also think that returning (ProcessInternal, bool) will just make things more verbose, but I'll change it if you want.

pkg/proc/native/proc.go
aarzilli: If we add a `parent *nativeProcess` field then every time we need to reference these three fields we have to do `if dbp.parent != nil { dbp.parent.ptraceChan... } else { dbp.ptraceChan... }` which is weird.

We could add a type like this:

```
type ptraceThread struct {
   ptraceChan chan func()
   ptraceDoneChan chan interface{}
   ptraceRefCnt int
}
```

and then keep a pointer to it in nativeProcess, which parents and children would share.

Or were you thinking something else?

pkg/proc/interface.go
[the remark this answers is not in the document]
derekparker: Ack, agreed. Let's leave things as is.

pkg/proc/native/proc.go
derekparker: Yes, that solution is along the lines of what I had in mind.

aarzilli: Done.

pkg/proc/target_exec.go
```diff
@@ -191,17 +223,56 @@ func (grp *TargetGroup) Continue() error {
...
+		if grp.KeepSteppingBreakpoints&HaltKeepsSteppingBreakpoints == 0 {
+			dbp.ClearSteppingBreakpoints()
+		}
+	}
+}
+
+// fixCurrentThreads switches traptgt to trapthread, then for each target in
+// the group if its current thread exists it refreshes the current
+// goroutine, otherwise it switches it to a randomly selected thread.
+func (grp *TargetGroup) fixCurrentThreads(traptgt *Target, trapthread Thread) error {
```
derekparker, line 254: nit: maybe `setCurrentThreads` instead. The `fix` prefix implies something was done incorrectly.

pkg/proc/target_exec.go
```diff
@@ -211,39 +282,103 @@ func isTraceOrTraceReturn(bp *Breakpoint) bool {
...
+	return nil
+}
+
+// pickCurrentTarget picks a new current target, with the following property:
+//
+//   - a target with an active stepping breakpoint
+//   - a target with StopReason == StopCallReturned
+//   - a target with an active breakpoint, prioritizing traptgt
+//   - traptgt
+func pickCurrentTarget(grp *TargetGroup, traptgt *Target) *Target {
```
derekparker, line 347: This feels like it should be a method instead of a function as it's use is to set `TargetGroup.Selected` and isn't really used outside of that, so it might as well set that directly and then callers can use `TargetGroup.Selected` directly after that. Or perhaps we also return the selected target for convenience but also set `TargetGroup.Selected` within the method.

pkg/proc/target_group.go
```diff
@@ -12,12 +15,18 @@ import (
...
-	targets  []*Target
-	Selected *Target
+	targets           []*Target
+	Selected          *Target
+	followExecEnabled bool
 
 	RecordingManipulation
 	recman RecordingManipulationInternal
 
+	// StopReason describes the reason why the target process is stopped.
```
derekparker, line 25: This is specific to the selected target, so it should be stated explicitly here since this is `TargetGroup`.

pkg/proc/target_group.go
```diff
@@ -46,6 +55,9 @@ func NewGroup(t *Target) *TargetGroup {
 		LogicalBreakpoints:    t.Breakpoints().Logical,
 		continueOnce:          t.continueOnce,
 	}
+	grp.cctx.grp = grp
+	grp.cctx.DebugInfoDirs = t.debugInfoDirs
```
derekparker, line 59: Should `debugInfoDirs` just live on `TargetGroup` and then let `Target` have a reference back up to the group?

pkg/proc/interface.go
```diff
@@ -134,3 +139,18 @@ func (cctx *ContinueOnceContext) GetManualStopRequested() bool {
 	defer cctx.StopMu.Unlock()
 	return cctx.manualStopRequested
 }
+
+// NumProcs returns the number of processes in the target group being resumed.
+func (cctx *ContinueOnceContext) NumProcs() int {
```
derekparker, line 144: This feels like it should be on `TargetGroup`.

pkg/proc/interface.go
```diff
@@ -134,3 +139,18 @@ func (cctx *ContinueOnceContext) GetManualStopRequested() bool {
...
+	return len(cctx.grp.targets)
+}
+
+// Proc returns the i-th process in the group.
+func (cctx *ContinueOnceContext) Proc(i int) ProcessInternal {
```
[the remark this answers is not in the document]
derekparker: This also feels like it should be on `TargetGroup`.

pkg/proc/interface.go
```diff
@@ -134,3 +139,18 @@ func (cctx *ContinueOnceContext) GetManualStopRequested() bool {
...
+	return cctx.grp.targets[i].proc
+}
+
+// AddTarget adds a target to the group being resumed.
+func (cctx *ContinueOnceContext) AddTarget(tgt *Target) {
```
derekparker, line 154: This also feels like we should just be calling this directly on `TargetGroup`.

pkg/proc/interface.go
derekparker: If we want to prevent backends from messing with the `TargetGroup` too much, we could still add an exported field in `ContinueOnceContext` but limit it via an interface since we're already storing the target group there, just not exported for the backends, and it really is only there to implement these methods.

```
type TargetGroupProcs interface {
  NumProcs()
  Proc(i int)
  AddTarget(t)
}

type ContinueOnceContext struct {
  ...
  TargetGroup TargetGroupProcs
}
```

pkg/proc/native/proc.go
```diff
@@ -351,3 +378,55 @@ func openRedirects(redirects [3]string, foreground bool) (stdin, stdout, stderr
...
+	r := 0
+	for i := 0; i < cctx.NumProcs(); i++ {
+		if !cctx.Proc(i).(*nativeProcess).exited {
+			r++
+		}
+	}
+	return r
+}
+
+func procForThread(cctx *proc.ContinueOnceContext, tid int) *nativeProcess {
```
derekparker, line 392: This also feels like it should be a method on `TargetGroup` and possibly included in the interface I sketched out in another comment.

pkg/proc/native/proc.go
```diff
@@ -351,3 +378,55 @@ func openRedirects(redirects [3]string, foreground bool) (stdin, stdout, stderr
 
 	return stdin, stdout, stderr, closefn, nil
 }
+
+func numValid(cctx *proc.ContinueOnceContext) int {
```
derekparker, line 382: This similarly feels like it should be a method. Anything where we're "asking" something of the specific type I think makes more sense as a method.

pkg/proc/native/proc_linux.go
```diff
@@ -330,14 +340,19 @@ const (
 	trapWaitDontCallExitGuard
 )
 
-func (dbp *nativeProcess) trapWaitInternal(pid int, options trapWaitOptions) (*nativeThread, error) {
+func trapWaitInternal(cctx *proc.ContinueOnceContext, pid int, options trapWaitOptions) (*nativeThread, error) {
+	var waitdbp *nativeProcess = nil
+	if cctx.NumProcs() == 1 {
```
derekparker, line 345: I think we should add a comment here because on the surface this looks like a bug, however once you dig deeper you find that `waitdbp.wait()` will behave accordingly if it's receiver is nil.

derekparker (changes requested):

pkg/proc/interface.go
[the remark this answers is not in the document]
aarzilli: > This might actually be a good use case for generics

I don't want to raise the minimum version of Go this much just for this. 

Also I'm not sure how to do it in a way that doesn't suck more than this. If we add a type parameter to ContinueOnceContext then Target and TargetGroup also need to have it. This already sucks because Target and TargetGroup don't actually care about the specific type of the processes they contain besides that they satisfy an interface. On top of that TargetGroup sits inside Debugger and the specific type of the process it contains is determined at runtime, so we woud have to hide the type parameter through an interface to store the TargetGroup inside Debugger.

A better way to avoid the typecast is to keep two parallel slices, one of Targets and one of Processes. TargetGroup would contain a ProcessGroup interface:

```
type ProcessGroup interface {
    NumTargets() int
    Target(i int) *Target
}
```

which would be implemented like this:

```
type nativeProcessGroup struct {
    targets []*proc.Target
    processes []*nativeProcess
}
```

and the targets and processes have to be kept in sync at all times.

pkg/proc/target_exec.go
```diff
@@ -98,31 +105,56 @@ func (grp *TargetGroup) Continue() error {
...
+				it.Target.ClearSteppingBreakpoints()
+			}
 		}
 
-		callInjectionDone, callErr := callInjectionProtocol(dbp, threads)
-		hcbpErr := dbp.handleHardcodedBreakpoints(trapthread, threads)
+		var callInjectionDone bool
+		var callErr error
+		var hcbpErr error
+		it = ValidTargets{Group: grp}
```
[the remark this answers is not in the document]
aarzilli: I don't like using closures for looping because break and continue don't work with them and return doesn't do what you want it to do.

It also doesn't simplify the code much, for example this:

```
it := ValidTargets{Group: grp}
for it.Next() {
	for _, bp := range it.Breakpoints().M {
		if bp.LogicalID() == lbp.LogicalID {
			n++
			err := it.ClearBreakpoint(bp.Addr)
			if err != nil {
				errs = append(errs, err)
			}
		}
	}
}
```

becomes

```
grp.ForEachValidTarget(func (t *proc.Target) error {
	for _, bp := range t.Breakpoints().M {
		if bp.LogicalID() == lbp.LogicalID {
			n++
			err := t.ClearBreakpoint(bp.Addr)
			if err != nil {
				errs = append(errs, err)
			}
		}
	}
	return nil
})
```

which is the exact same number of lines.

pkg/proc/interface.go
aarzilli: > If we want to prevent backends from messing with the TargetGroup too much

Rather than backends messing with the TargetGroup my concern is exposing extra things as the public interface of TargetGroup. Users of TargetGroup don't need to call `Proc(i)` and get a ProcessInternal (and nothing good can come off it).

pkg/proc/interface.go
aarzilli: Same as above. It would be really bad if someone saw the AddGroup method on TargetGroup and though they could call NewTarget and pass it to (*TargetGroup).AddTarget.

pkg/proc/native/proc.go
aarzilli: I'm going to wait before making changes around this until the general discussion about TargetGroup / ContinueOnceContext is resolved.

pkg/proc/native/proc.go
```diff
@@ -351,3 +378,55 @@ func openRedirects(redirects [3]string, foreground bool) (stdin, stdout, stderr
...
+		ptraceChan:     make(chan func()),
+		ptraceDoneChan: make(chan interface{}),
+		ptraceRefCnt:   1,
+	}
+	go pt.handlePtraceFuncs()
+	return pt
+}
+
+func (pt *ptraceThread) acquire() *ptraceThread {
+	pt.ptraceRefCnt++
```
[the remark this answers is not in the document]
derekparker: Ack, thanks.

pkg/proc/interface.go
derekparker: Ack, this makes sense.

pkg/proc/interface.go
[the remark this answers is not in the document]
derekparker: You're right, introducing generics would be a pretty big rewrite of the types and plumbing it through would be cumbersome. Even if we did want to do that eventually, it wouldn't make sense to add that much noise to the PR. Let's leave the type casting for now until we can figure out a more elegant solution (I don't like the double-list approach).

pkg/proc/target_exec.go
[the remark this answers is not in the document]
derekparker: Valid points.

pkg/proc/native/proc.go
aarzilli: Done.

pkg/proc/target_exec.go
```diff
@@ -98,31 +105,56 @@ func (grp *TargetGroup) Continue() error {
...
 			if pe, ok := contOnceErr.(ErrProcessExited); ok {
-				dbp.exitStatus = pe.Status
+				traptgt.exitStatus = pe.Status
 			}
 			return contOnceErr
 		}
-		if dbp.StopReason == StopLaunched {
-			dbp.ClearSteppingBreakpoints()
+		if stopReason == StopLaunched {
+			it := ValidTargets{Group: grp}
```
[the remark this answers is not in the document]
aarzilli: Done.

pkg/proc/target_exec.go
```diff
@@ -48,48 +48,55 @@ func (grp *TargetGroup) Next() (err error) {
...
+		} else {
+			traptgt = grp.targets[0]
+		}
+		traptgt.StopReason = stopReason
 
-		threads := dbp.ThreadList()
-		for _, thread := range threads {
-			if thread.Breakpoint().Breakpoint != nil {
-				thread.Breakpoint().Breakpoint.checkCondition(dbp, thread, thread.Breakpoint())
+		it := ValidTargets{Group: grp}
```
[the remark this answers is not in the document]
aarzilli: Replaced with Reset.

pkg/proc/native/proc.go
aarzilli: Done.

pkg/proc/target_group.go
aarzilli: The way initialization is done right now, we make a target with NewTarget, which creates a BinaryInfo (and therefore needs debugInfoDirs) and then we create a group with NewTargetGroup(target). So the target group doesn't exist when we need debugInfoDirs (at least, the first time). To do this we would have to make Launch, Attach etc return a TargetGroup.

pkg/proc/target_group.go
```diff
@@ -46,6 +55,9 @@ func NewGroup(t *Target) *TargetGroup {
 		LogicalBreakpoints:    t.Breakpoints().Logical,
 		continueOnce:          t.continueOnce,
 	}
+	grp.cctx.grp = grp
```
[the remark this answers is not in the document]
aarzilli: It's not enough, when we add a new process we have to create the target object to apply the breakpoints to it.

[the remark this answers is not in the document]
derekparker: > It's not enough, when we add a new process we have to create the target object to apply the breakpoints to it.

We can add a new `StopReason` like `ProcessCreated` or something and let `TargetGroup.Continue` handle it accordingly. That would actually prevent the need for maintaining a list of new processes on the context object at all. We still have the problem of keeping `cctx.procs` in sync, but it's relatively easy to just assign it at the top of the loop right before we call `grp.continueOnce`.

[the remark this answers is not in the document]
aarzilli: We'd have to stop every other thread of every process to do that, it's pretty wasteful to exit the ContinueOnce loop if we just inted to re-enter it anyway.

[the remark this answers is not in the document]
derekparker: That's a good point. Let me think on this a bit more.

pkg/proc/target_group.go
derekparker: I'm not against making Launch, Attach, etc return a TargetGroup since that's the level of abstraction we deal with now. Thoughts?

pkg/proc/target_group.go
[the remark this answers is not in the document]
derekparker: What if we got rid of `ContinueOnceContext` and moved the members directly into `TargetGroup`? I think this assignment and circular reference shows that `ContinueOnceContext` is redundant. It's a long lived type that exists as long as the `TargetGroup` and is not independent in any way, so why separate it? Then, so we don't have to expose all of TargetGroup into the backends we have them accept an interface exposing only the methods they need instead of taking a `ContinueOnceContext` object.

aarzilli:
I've made some changes that hopefully makes it better regarding debugInforDirs, ContinueOnceContext and all the type assertions. I've kept them separate to a different commit so it's easier to roll them back and evaluate them separately.
It's a bit noisy, especially with the changes in proc_test.go, the basic gist is this:

- Launch/Attach/etc all return TargetGroup instead of Target
- To get the TargetGroup they call proc.NewGroup, passing to it some configuration and an object implementing the ProcessGroup interface
- The ProcessGroup interface is a simple interface that only has a ContinueOnce method
- From proc.NewGroup they get an empty TargetGroup as well as a addTarget method which they have to call once to add the new target to the group, but that they can also keep a reference of to add more targets if follow exec is enabled.
- The addTarget function has basically the same signature that NewTarget had, and NewTarget no longer exists

aarzilli:
Done.

derekparker (approved):
LGTM

PR #3167, opened 2022-10-19
service/debugger: Assume current dir for exec

derekparker:
This patch modifies the behavior of the exec subcommand such that you don't necessarily have to write the "./" prefix when trying to debug a precompiled binary in your working directory.

For example (given foo.test in working dir), before this change:

dlv exec foo.test

Would result in an error, forcing the user to type:

dlv exec ./foo.test

This just makes things a bit more convenient.

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -11,6 +11,7 @@ import (
 	"go/token"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -2273,11 +2274,9 @@ func verifyBinaryFormat(exePath string) (string, error) {
 	}
 	defer f.Close()
 
-	fi, err := f.Stat()
+	// Ensure the file is executable
+	_, err = exec.LookPath(fullpath)
 	if err != nil {
-		return "", err
-	}
-	if (fi.Mode() & 0111) == 0 {
 		return "", api.ErrNotExecutable
 	}
 
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -9,8 +9,8 @@ import (
 	"fmt"
 	"go/parser"
 	"go/token"
+	"io"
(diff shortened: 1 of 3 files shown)


aarzilli (commented):
See failing tests.
I think calling filepath.Abs instead of filepath.Join would fix it, it would use os.Getwd instead of the wd argument always, but I'm not sure why we should use the wd argument for this anyway.

Also, we should probably clean up the call to filepath.Abs in native.Launch for Windows as well as the call to exec.LookPath in verifyBinaryFormat (and do the verifyBinaryFormat call after this chunk of code).

derekparker:
> See failing tests. I think calling filepath.Abs instead of filepath.Join would fix it, it would use os.Getwd instead of the wd argument always, but I'm not sure why we should use the wd argument for this anyway.
> 
> Also, we should probably clean up the call to filepath.Abs in native.Launch for Windows as well as the call to exec.LookPath in verifyBinaryFormat (and do the verifyBinaryFormat call after this chunk of code).

Done.

service/debugger/debugger.go
```diff
@@ -2259,34 +2262,31 @@ func noDebugErrorWarning(err error) error {
...
-				return err
-			}
-		}
-	default:
-		fi, err := f.Stat()
+	// Skip this check on Windows.
+	// TODO(derekparker) exec.LookPath looks for valid Windows extensions.
+	// We don't create our binaries with valid extensions, even though we should.
+	// Skip this check for now.
+	if runtime.GOOS != "windows" {
```
derekparker, line 2281: Skipping this on Windows because the previous check never actually got executed by our tests. I'll likely submit a follow up PR to make this work on windows, but it involves making sure binaries are generated with the correct extension on Windows (e.g. `.exe`).

derekparker:
@aarzilli PTAL.

aarzilli (approved):

PR #3169, opened 2022-10-20
pkg/terminal: Improve error when unable to read history file

### LINKED ISSUE #3168: Unable to read history error, no clear way to resolve.
I am using delve v1.9.1
go version go1.19.2 linux/amd64
In the course of using delve to debug my program I have at some point generated a line, no idea how, that is too long for the history file.
I now receive this error whenever I start delve:

`Unable to read history file: line 158197 is too longType 'help' for list of commands.`

... and apparently in consequence the history, which is very useful when recalling similar break point placements, is no longer being recorded.

I have no idea where the cache of the history to remove the offending line, and no option to clear the cache or indication of where this cache is in either the error message or the program documentation.

I would appreciate some help in clearing this up.

derekparker:
This patch includes the path to the history file itself so that users can better fix the issue when an error arises.

Fixes #3168

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -292,7 +292,7 @@ func (t *Term) Run() (int, error) {
 		fmt.Printf("Unable to open history file: %v. History will not be saved for this session.", err)
 	}
 	if _, err := t.line.ReadHistory(t.historyFile); err != nil {
-		fmt.Printf("Unable to read history file: %v", err)
+		fmt.Printf("Unable to read history file %s: %v\n", t.historyFile, err)
 	}
 
 	fmt.Println("Type 'help' for list of commands.")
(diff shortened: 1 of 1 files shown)


pkg/terminal/terminal.go
```diff
@@ -292,7 +292,7 @@ func (t *Term) Run() (int, error) {
 		fmt.Printf("Unable to open history file: %v. History will not be saved for this session.", err)
 	}
 	if _, err := t.line.ReadHistory(t.historyFile); err != nil {
-		fmt.Printf("Unable to read history file: %v", err)
+		fmt.Printf("Unable to read history file %s: %v\n", t.historyFile, err)
```
aarzilli, line 295: go vet fails on this line:

```
pkg/terminal/terminal.go:295:3: fmt.Printf format %s has arg t.historyFile of wrong type *os.File
```

derekparker: Fixed.

aarzilli (approved):

Issue #3170, opened 2022-10-21
DAP Evaluate and SetVariables response not containing `type` with `supportsVariableType`. 


gfszr:
When issuing a `SetVariable` request or an `Evaluate` request, the response contains the type of the new value if `supportsVariableType` is negotiated upon DAP initialization. However, it is currently not returned:

https://github.com/go-delve/delve/blob/6440b3ba9216cb936f43318b96df540cf56b216f/service/dap/server.go#L2642-L2644

From a client's perspective the type is very important as SetVariables responses might return a different type than the original variable, let alone Evaluate which can return an arbitrary type. 

**Dlv Version**: `1.9.1`
**Go Version**: `1.16`
**Expected Result**: EvaluateResponse/SetVariablesResponse contains the type of the variable.
**Actual Result**: Type is not returned at all.

aarzilli:
Closing since a fix was merged.

gfszr:
@aarzilli was a fix merged for the missing type in `SetVariableResponse` as well?

aarzilli:
No.

PR #3172, opened 2022-10-22
Add type to simple expression evaluation in DAP's `evaluateRequest`

gfszr:
Solve https://github.com/go-delve/delve/issues/3170 for `evaluateExpression` so that the type of the evaluation is returned upon evaluating an expression. Implemented so when there's an actual evaluated variable with a type with a variable reference, which is the either a `call` expression or a plain expression.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2694,7 +2694,7 @@ func (s *Session) onEvaluateRequest(request *dap.EvaluateRequest) {
 			opts |= showFullValue
 		}
 		exprVal, exprRef := s.convertVariableWithOpts(exprVar, fmt.Sprintf("(%s)", request.Arguments.Expression), opts)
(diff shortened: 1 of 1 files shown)


aarzilli (commented):
What about the EvaluateResponseBody in the first if branch of onEvaluateRequest? Also shouldn't all of those be calls to getTypeIfSupported, since the client might not support variable types?

gfszr:
@aarzilli Thanks for the quick reply!

1. Added the use of the session's `s.getTypeIfSupported)`.
2. Regarding the `call` and `dlv` expression evaluation, type is mostly needed for a watches/evaluate pane rather than a debug console evaluate (which is mostly just printed to the screen), so I think fixing it and then progressing to fix `dlv` and `call` with proper typing the way to go. (I agree `invalid` is not a proper type, I wasn't aware that this was the returned type in `call` scenarios).

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3180, opened 2022-11-02
*: early fixes for go1.20

aarzilli:
* updated go/packages to support new export format
* rewrite testinline.go fixture because the compiler got too smart with
constant folding
* temporarily disable staticcheck on go1.20 because it doesn't support
the new export format.

* workaround for go.dev/cl/429601


aarzilli:
Added a fix for the windows/amd64 change in where the address of the g struct is stored.

PR #3181, opened 2022-11-06
proc/gdbserial/gdbserver: Don't pass DYLD/LD to debugserver

aviramha:
This change fixes recent situations where you can't really load to debugserver/lldb since of new protections
but you still want to load to the debugged process delve already supports moving the parameters to the child using -e, this just supresses issues being loaded to the debugger itself

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -547,6 +547,17 @@ func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs [
 
 	if runtime.GOOS == "darwin" {
 		process.Env = proc.DisableAsyncPreemptEnv()
+		// Filter out DYLD_INSERT_LIBRARIES on Darwin.
+		// This is needed since macOS Ventura, loading custom dylib into debugserver
+		// using DYLD_INSERT_LIBRARIES leads to a crash.
+		// This is unlike other protected processes, where they just strip it out.
+		env := make([]string, 0, len(process.Env))
+		for _, v := range process.Env {
+			if !(v == "DYLD_INSERT_LIBRARIES") {
+				env = append(env, v)
+			}
+		}
+		process.Env = env
 	}
 
 	if err = process.Start(); err != nil {


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -547,6 +547,14 @@ func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs [
 
 	if runtime.GOOS == "darwin" {
 		process.Env = proc.DisableAsyncPreemptEnv()
+		// filter out DYLD_* environment variables
```
aarzilli, line 550: Explain why here.

aarzilli (commented):
Can you be less vague about what problem this fixes? You say "recent situations" and "new protections" but what are those?

pkg/proc/gdbserial/gdbserver.go
aviramha: To have all the darwin env changes in the same spot - or you want me to add a comment saying that?

aviramha:
> Can you be less vague about what problem this fixes? You say "recent situations" and "new protections" but what are those?

Yes, sorry. The changes aren't documented and I am not sure what they changed exactly but I'll try to do my best to explain here.
Until macOS Ventura, you could use `DYLD_INSERT_LIBRARIES` to load a custom dylib into `debugserver`. It would require you to build it with arm64e architecture (or fat binary that contains it as well).
Since macOS Ventura release, `debugserver` would crash saying no matching architecture is found, while all relevant architectures exist.
This fix let's us "skip" loading to `debugserver` when we actually want to load to the debugged process (that we were okay with loading to, because in that case we just did nothing)
There was already a feature to add `-e` to `debugserver` arguments of `DYLD_INSERT_LIBRARIES`/`LD_PRELOAD` so I assume there are more users that use this capability to load into child processes. The only issue is that environment variable being "leaked" to debugserver, which doesn't work anymore in macOS Ventura.
I'm currently chasing Apple engineers to understand if that's a desired change from their side (other binaries that they don't want custom dylibs loaded into have different approach - they just ignore the env var. In case of this binary it just crashes.)
Relevant issue in our project repository:
https://github.com/metalbear-co/mirrord/issues/677

aarzilli:
> Until macOS Ventura, you could use `DYLD_INSERT_LIBRARIES` to load a custom dylib into `debugserver`

As far as I know this is not true, see:

https://stackoverflow.com/questions/33587130/why-doesnt-lldb-forward-my-environment-variable-anymore/33589760#33589760

macOS used to strip all DYLD_ environment variables before running protected processes. This is why getLdEnvVars exists.

pkg/proc/gdbserial/gdbserver.go
aarzilli: I think this comment should explain why we are filtering out DYLD_ variables.
Basically it should repeat what the commit message says and both should be more precise about what changed and when it did than what the current commit message is, if possible.

aviramha:
> 

I was well aware of that - but specifically `debugserver` and other binaries weren't protected like other binaries - i.e you could use `DYLD_INSERT_LIBRARIES`. Post Ventura if you try to use it it crashes instead of stripping it like other protected binaries. That's why this patch exists.

pkg/proc/gdbserial/gdbserver.go
aviramha: Understood - I'll amend in few minutes.

aviramha: I hope it is more clear now.

aarzilli (approved):
LGTM

aviramha:
Thank you. Anything that I need to do?

aarzilli:
No.

derekparker (approved):
LGTM

PR #3184, opened 2022-11-07
proc: fix index access to already-loaded string values

### LINKED ISSUE #3176: Indexing strings crashes Delve
```
(dlv) call len(ref.String())
Values returned:
        len(ref.String()): 56
(dlv) call ref.String()[0]
panic: runtime error: index out of range [0] with length 0
```

aarzilli:
Fixes #3176


derekparker (approved):
LGTM

PR #3185, opened 2022-11-07
service/debugger: allow setting a breakpoint with a specific ID

aarzilli:
Before 4c5b111a it was possible for a client to create a breakpoint
with a specific logical ID, restore this ability for backwards
compatibility.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -2961,3 +2961,20 @@ func TestPluginSuspendedBreakpoint(t *testing.T) {
 		cont("Continue 4", "plugin2.go", 9)
 	})
 }
+
+func TestClientServer_createBreakpointWithID(t *testing.T) {
+	protest.AllowRecording(t)
+	withTestClient2("continuetestprog", t, func(c service.Client) {
+		bp, err := c.CreateBreakpoint(&api.Breakpoint{ID: 2, FunctionName: "main.main", Line: 1})
+		assertNoError(err, t, "CreateBreakpoint()")
+		if bp.ID != 2 {
+			t.Errorf("wrong ID for breakpoint %d", bp.ID)
+		}
+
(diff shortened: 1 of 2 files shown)


service/test/integration2_test.go
```diff
@@ -2961,3 +2961,11 @@ func TestPluginSuspendedBreakpoint(t *testing.T) {
 		cont("Continue 4", "plugin2.go", 9)
 	})
 }
+
+func TestRestart_createBreakpointWithID(t *testing.T) {
+	protest.AllowRecording(t)
+	withTestClient2("continuetestprog", t, func(c service.Client) {
+		_, err := c.CreateBreakpoint(&api.Breakpoint{ID: 2, FunctionName: "main.main", Line: 1, Name: "firstbreakpoint", Tracepoint: true})
```
derekparker, line 2968: We should also validate the breakpoint gets the ID requested here, for completeness.

service/test/integration2_test.go
```diff
@@ -2961,3 +2961,11 @@ func TestPluginSuspendedBreakpoint(t *testing.T) {
 		cont("Continue 4", "plugin2.go", 9)
 	})
 }
+
+func TestRestart_createBreakpointWithID(t *testing.T) {
```
derekparker, line 2965: Why `TestRestart_`?

derekparker (changes requested):

service/test/integration2_test.go
aarzilli: Done.

service/test/integration2_test.go
aarzilli: Bad copy paste.

aarzilli:
There was also a bug with breakpointIDCounter not being updated when setting a breakpoint by ID, so that we don't reuse breakpoint IDs.

derekparker (approved):

PR #3186, opened 2022-11-08
Two fixes to type cast evaluation

aarzilli:
### proc: fix type casts to ptr-to-ptr types

Fix type casts to **type.

### proc: allow casts form unsafe.Pointer to any pointer and vice versa

We've allowed doing this with uintptr but we should allow
unsafe.Pointer to be used like Go uses it.


derekparker (approved):
LGTM

PR #3193, opened 2022-11-15
proc: prefer addresses when setting a breakpoint

aarzilli:
If a breakpoint has both a list of addresses and an expression prefer
the list of addresses, otherwise it is impossible to set breakpoint
with expressions that depend on the current scope, like 'break +0'
which sets a breakpoint on the current line.


derekparker (approved):
LGTM

Issue #3197, opened 2022-11-16
On-prem windows/arm64 builder 


qmuntal:
> Steal an arm64 laptop and hide it in a closet. @aarzilli [source](https://github.com/go-delve/delve/pull/3063#pullrequestreview-1079246540)

I finally got a Windows ARM64 PC without breaking any law nor any Microsoft policy, just had to apply to the [Windows Dev Kit 2023](https://blogs.windows.com/windowsdeveloper/2022/10/24/available-today-windows-dev-kit-2023-aka-project-volterra/) project 😄.

Delve windows/arm64 is still experimental, gated behind the `exp.winarm64` build tag waiting for a builder. Well, I can provide this builder and time to debug issues. Just a couple of conditions:
- I plan to have it up and running all the time, but I can't commit to any SLA nor support it outside office hours.
- I will probably have the hardware for at least one year, can commit to any further timeframe. Hopefully windows/arm64 hardware availability will increase during this time period so my on-prem machine won't be necessary in 1 year.

I already have the TimeCity On-prem agent installer and running named `windows/arm64-qmuntal-2`. Just missing the approval to run builds.

@aarzilli @derekparker

aarzilli:
I have no problems with those conditions. Pretty cool.

derekparker:
This is great! Thank you for setting this all up.

So if I'm not mistaken the agent is [windows/arm64-qmuntal-2](https://delve.teamcity.com/agent/265) however it appears to be offline according to the TeamCity UI. I've authorized the agent so once it's online it should be good to go.

derekparker:
Nevermind, it seems to be connected again now. Thanks for this!

PR #3198, opened 2022-11-16
Run CI tests on windows/arm64 agents

qmuntal:
Now that we have windows/arm64 agents (see #3197), let's try to exercise them.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -48,6 +48,9 @@ val targets = arrayOf(
         "windows/amd64/1.19",
         "windows/amd64/tip",
 
+        "windows/arm64/1.19",
+        "windows/arm64/tip",
+
         "mac/amd64/1.19",
         "mac/amd64/tip",
 
--- a/_scripts/test_windows.ps1
+++ b/_scripts/test_windows.ps1
@@ -77,12 +77,3 @@ $x = $LastExitCode
 if ($version -ne "gotip") {
 	Exit $x
 }
-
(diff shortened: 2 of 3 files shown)


qmuntal:
@derekparker is there any setup missing? Can't see tjhe windows/arm64 subproject in teamcity:

image

derekparker:
@qmuntal looking at the TeamCity agent admin page it seems to think the windows/arm64 agent is disconnected right now.

qmuntal:
> @qmuntal looking at the TeamCity agent admin page it seems to think the windows/arm64 agent is disconnected right now.

I think I have something wrong in the agent configuration. It unregisters every now and them and registers again after a while.

But I think this problem is separated from what I was saying about the subproject. Have you tried creating it a windows/arm64 subproject from the dashboard?

aarzilli:
AFAIK changes to the configuration of teamcity from the dashboards are not allowed: all the targets are created from the settings.kts file that is committed in the repository. The windows/arm64 target will only appear after a change like the one in this PR is merged.

qmuntal:
> AFAIK changes to the configuration of teamcity from the dashboards are not allowed: all the targets are created from the settings.kts file that is committed in the repository. The windows/arm64 target will only appear after a change like the one in this PR is merged.

Makes sense. Then I'm limiting the scope of this PR to just set up windows/arm64 subproject and run a stub command instead of the build + test steps. I'll fix the build scripts and failing unit tests in a follow up PR.

aarzilli (approved):

PR #3200, opened 2022-11-17
Test windows/arm64 pipeline

qmuntal:
This PR does all the wiring so `windows/arm64` gets tested on the TeamCity pipeline using my on-prem builder.

This is the list of changes:

- Update `test_windows.ps1` so it only writes files (namely the Go binary and source code, gcc toolchain and procdump) inside the TeamCity agent home. This is necessary because the on-prem agent is running as a low-privileged service which only have read/write access to its home directory.
- Mingw-w64 toolchain still not supports windows/arm64. I'm using a [LLVM based mingw-64 toolchain](https://github.com/mstorsjo/llvm-mingw) instead, which is inline with what's using Go to test on this platform.
- Drop the `exp.winarm64` build tag.
- Update `TestLaunchRequestWithRelativeExecPath` so it does not fail if symlinks can't be created, but just skip the test. My on-prem agent does not have permissions to create symlinks (on purpose), so that test can't be executed.
- Skip `TestCgoStacktrace`. CGO stack traces are broken on `windows/arm64`. It can be investigated in a follow-up issue.
- Function calls are flaky, so I've marked them as not implemented.

@aarzilli @derekparker

### DIFF
--- a/pkg/proc/native/support_sentinel_windows.go
+++ b/pkg/proc/native/support_sentinel_windows.go
@@ -1,6 +1,8 @@
 // This file is used to detect build on unsupported GOOS/GOARCH combinations.
 
-//go:build windows && !amd64 && !arm64
-// +build windows,!amd64,!arm64
+//go:build windows && !amd64 && !(arm64 && exp.winarm64)
+// +build windows
+// +build !amd64
+// +build !arm64 !exp.winarm64
 
 package your_windows_architecture_is_not_supported_by_delve
--- a/pkg/proc/native/support_sentinel_windows.go
+++ b/pkg/proc/native/support_sentinel_windows.go
@@ -1,8 +1,6 @@
 // This file is used to detect build on unsupported GOOS/GOARCH combinations.
 
-//go:build windows && !amd64 && !(arm64 && exp.winarm64)
-// +build windows
-// +build !amd64
-// +build !arm64 !exp.winarm64
+//go:build windows && !amd64 && !arm64
+// +build windows,!amd64,!arm64
 
 package your_windows_architecture_is_not_supported_by_delve
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -5473,7 +5473,11 @@ func TestLaunchRequestWithRelativeExecPath(t *testing.T) {
 		symlink := "./__thisexe"
 		err := os.Symlink(fixture.Path, symlink)
 		if err != nil {
-			t.Skip("this test requires symlinks to be enabled and allowed")
+			if runtime.GOOS == "windows" {
+				t.Skip("this test requires symlinks to be enabled and allowed")
+			} else {
+				t.Fatal("unable to create relative symlink:", err)
+			}
 		}
 		defer os.Remove(symlink)
 		runDebugSession(t, client, "launch", func() {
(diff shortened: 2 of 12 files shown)


qmuntal:
Yep, I'm fighting with those now trying to find where to set them...

aarzilli:
Have you seen this: https://jetbrains.com.xy2401.com/help/teamcity/build-agent-configuration.html

qmuntal:
> Have you seen this: https://www.jetbrains.com/help/teamcity/configure-agent-installation.html#Optional+Properties

It worked 😄 

image

qmuntal:
@aarzilli @derekparker this is ready for review!

_fixtures/testinline.go
```diff
@@ -7,6 +7,7 @@ func inlineThis(a int) int {
 	return z + a
 }
 
+//go:noinline
```
aarzilli, line 10: Can you undo this and the changes related to this? I have a different fix for this in 1.20 that I prefer, it doesn't matter if tests don't pass on 1.20.

service/dap/server_test.go
```diff
@@ -5472,10 +5472,10 @@ func TestLaunchRequestWithRelativeExecPath(t *testing.T) {
 	runTest(t, "increment", func(client *daptest.Client, fixture protest.Fixture) {
 		symlink := "./__thisexe"
 		err := os.Symlink(fixture.Path, symlink)
-		defer os.Remove(symlink)
 		if err != nil {
-			t.Fatal("unable to create relative symlink:", err)
+			t.Skip("this test requires symlinks to be enabled and allowed")
```
aarzilli, line 5476: Only skip this on windows.

_fixtures/testinline.go
qmuntal: Sure, done.

service/dap/server_test.go
qmuntal: Done

pkg/proc/native/support_sentinel_windows.go
```diff
@@ -1,8 +1,6 @@
 // This file is used to detect build on unsupported GOOS/GOARCH combinations.
 
-//go:build windows && !amd64 && !(arm64 && exp.winarm64)
-// +build windows
-// +build !amd64
-// +build !arm64 !exp.winarm64
+//go:build windows && !amd64 && !arm64
```
aarzilli, line 3: I ran the tests 10 times, of these 3/10 got stuck on TestStepConcurrentDirect. Therefore I think the build guard should stay and TestStepConcurrentDirect should have a skipOn line.

qmuntal: Haven't found this failure mode in my runs. Where are you running the tests? I don't see the failures in the teamcity dashboard (I'm interested in the logs).

aarzilli: Teamcity, you don't see them on the dashboard because they are personal builds:

https://delve.teamcity.com/buildConfiguration/Delve_windows_arm64_1_19/41584
https://delve.teamcity.com/buildConfiguration/Delve_windows_arm64_1_19/41587
https://delve.teamcity.com/buildConfiguration/Delve_windows_arm64_1_19/41588

I think you can see them if you go on the build history of your builder (unless teamcity is only showing them to me because they are mine).

qmuntal: Thanks, I can see them. Will take a look before skipping the test.

qmuntal: Ok, I have no clue why TestStepConcurrentDirect is failing and won't have time during the following weeks to look into it, so I'll skip it for now and keep the `exp.winarm64` gate.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3203, opened 2022-11-20
Bump x/tools to v0.1.12

qmuntal:
This PR upgrades`golang.org/x`/tools dependencies to v0.1.12.

This newer version contains a fix for a test that is failing on the `windows/arm64` port: https://delve.teamcity.com/buildConfiguration/Delve_windows_arm64_tip/41140?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A41140%29%2Cid%3A276

@aarzilli @derekparker


aarzilli:
What test is that and can it be fixed with a workaround? Unfortunately x/sys was update to use unsafe.Slice which is a bit to new. We could update the version in go.mod since 1.16 is technically unsupported, but it would inconvenience a number of people.

qmuntal:
`TestGeneratedDoc` fails to validate that the starlark documentation is up to date:

```cmd
=== RUN   TestGeneratedDoc
    dlv_test.go:455: could not run script [_scripts/gen-starlark-bindings.go go -]: exit status 1 (output: "2022/11/18 17:55:10 internal error: package \"fmt\" without types was imported from \"github.com/go-delve/delve/service/rpc2\"\nexit status 1\n")
```

But in fact the real problem is that `gen-starlark-bindings.go` can't generate the bindings because `golang.org/x/tools/go/packages` refuses to load `github.com/go-delve/delve/service/rpc2`:

MicrosoftTeams-image

qmuntal:
@aarzilli it seems that `x/tools@v0.1.12` also fixes the issue and it does not require bumping the go directive, so I'm reducing the scope of this PR to just upgrading the `x/tools` patch version.

qmuntal:
Now staticcheck is complaining that it can't import a couple of packages, grr...

aarzilli:
If you want to have the tests pass on 1.20 you should build on top of #3180. The version of staticcheck is specified in test_linux.sh but afaik there is no version that supports 1.20 yet (staticcheck is disabled for 1.20 in #3180).

qmuntal:
I'll probably abandon this PR and just skip the failing bits parts of `TestGeneratedDoc` until staticcheck supports 1.20.

PR #3204, opened 2022-11-24
proc: extend macOS workaround to amd64

### LINKED ISSUE #3194: Wrong position after step in on mac
```
Delve Debugger
Version: 1.9.1
Build: $Id: d81b9fd12bfa603f3cf7a4bc842398bd61c42940 $
```
```
go version 1.19.3 darwin/amd64
```

MacOS 12.5.1

4. What did you do?

Debugged the program

```
package main

import "fmt"
import _ "k8s.io/cli-runtime/pkg/resource"

func main() {
	f() // break here
}

func f() {
	fmt.Println("ok")
}
```

Once breakpoint is reached stepped twice.

6. What did you expect to see?

Debugger stops somewhere inside the fmt package.

7. What did you see instead?

Debugger stops in runtime (proc.go). Log: https://gist.github.com/nd/0445e4ed453155388cca9594ab37bd79.

Got wrong position in go1.19.2+ and go1.18.7+, works as expected in go1.19.1 and go1.18.6.

Also happens if k8s import is replaced with cgo:

```
package main

/*
#cgo LDFLAGS: -framework CoreFoundation
#cgo LDFLAGS: -framework CFNetwork
#include <CFNetwork/CFProxySupport.h>
*/
import "C"
import "fmt"

func main() {
	f() // break here
}

func f() {
	fmt.Println("ok")
}
```

Works as expected if the cgo import is removed.

aarzilli:
Go change 064f34f (which exists in Go 1.19.2 and following) removed the
pagezero_size option from linker calls (because it is deprecated). This
expanded the problem that exists on darwin/arm64 as well as PIE builds
on darwin/amd64 to all darwin/amd64 builds.

This problem is described on: https://github.com/golang/go/issues/25841.

This commit extends the darwin/arm64 workaround to darwin/amd64.

Fixes #3194


derekparker (approved):
LGTM

PR #3205, opened 2022-11-25
proc/gdbserver: set child process pgrp as foreground group

aarzilli:
Newer versions of debugserver (which contain [1]) will spawn the target
process on a new process group, when we detect that this happen, and we
are a headless instance and stdin is a tty, make the child process'
process group the controlling group for the terminal.

[1] https://reviews.llvm.org/rG33ac4fddc7906ba712c50cd3a9b02ae041d751ab


derekparker (approved):
LGTM

Issue #3173, opened 2022-10-23
Improve debug console output when building the debuggee


gfszr:
**Delve version:** 1.9.1

Right now, when using a simple launch configuration such as:
```json
{
            "name": "debug main",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "${workspaceFolder}/main.go",
```

gfszr:
@aarzilli I might be able to help with that one, if you have any guidelines before working on it 👍

PR #3207, opened 2022-11-29
Print build command line before executing build

gfszr:
This PR addresses https://github.com/go-delve/delve/issues/3173 by adding an output event before executing build command. In order to not break compatibility, added some functions to `pkg/gobuild`.

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -46,6 +46,16 @@ func GoBuildCombinedOutput(debugname string, pkgs []string, buildflags string) (
 	return gocommandCombinedOutput("build", args...)
 }
 
+func MakeGoBuildCmd(debugname string, pkgs []string, buildflags string) (string, *exec.Cmd) {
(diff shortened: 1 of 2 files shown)


aarzilli:
A bunch of dap tests fail with this.

gfszr:
Ah, I see there are a lot of tests which test output, I'll make sure to update them as well.

gfszr:
Hm, seems like there's an issue with how DAP tests expect packet responses: a lot of tests use `client.ExpectVisibleErrorResponse` after a launch request - problem is, instead of waiting for a proper response for the request sequence, it just reads the next packet in the connection, which is the newly added `OutputEvent` of the building phase. Perhaps this function should also actually wait for the response for the appropriate request that is asserted to fail.

aarzilli:
I think it could be acceptable to change it to discard output events. cc @polinasok

derekparker:
This hasn't been updated in some time, @gfszr are you still working on this change?

derekparker:
Closing due to lack of updates. Please reopen if you still intend to work on this.

PR #3212, opened 2022-12-12
Documentation: add -test.run to test example

aarzilli:
Add -test.run to the example of how 'dlv test' can be used.


derekparker (approved):
LGTM

PR #3222, opened 2022-12-15
Documentation/faq: Correct commands in substitute-path section

phil-blain:
The text of the substitute-path section in the FAQ ("Can not set breakpoints or see source listing in a complicated debugging environment") mentions the command 'config help substitute-path', but 'help' is not a valid argument to the 'config' command.

Use 'help config' instead, which does mention 'substitute-path' and how to set it.

Also, the command to list source files is 'sources', not 'source'. Fix that.


aarzilli (approved):
LGTM

aarzilli:
Thank you.

PR #3223, opened 2022-12-15
TeamCity: fix windows/amd64 builds

aarzilli:
Windows/amd64 builds are not runnning because the current settings.kts
has an implicity requirement to property teamcity.build.systemDir which
does not exist on windows/amd64 agents.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -204,7 +204,7 @@ class TestBuild(val os: String, val arch: String, version: String, buildId: Abso
                     scriptMode = file {
                         path = "_scripts/test_windows.ps1"
                     }
-                    param("jetbrains_powershell_scriptArguments", "-version ${"go$version"} -arch $arch -binDir %teamcity.build.systemDir%")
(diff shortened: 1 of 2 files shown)


aarzilli:
Unfortunately we have to merge this to see if it works.
cc @qmuntal

qmuntal:
Hmm, I'm wondering if the windows/amd64 agent has this property. I couldn't test it when working on the original PR because the settings.kts needs to reach master before being applied. What I did was to to add a sane fallback inside `test_windows.ps1` in case `teamcity.build.systemDir` wasn't there, which does work well.

I think it's better to just rely on the fallback `systemDir` and remove the `teamcity.build.systemDir` for now.

aarzilli:
> I think it's better to just rely on the fallback systemDir and remove the teamcity.build.systemDir for now.

Like this?

qmuntal:
> > I think it's better to just rely on the fallback systemDir and remove the teamcity.build.systemDir for now.
> 
> Like this?

Yep.

derekparker (approved):
LGTM

PR #3224, opened 2022-12-19
proc/native: fix FreeBSD backend

aarzilli:
* use PT_SUSPEND/PT_RESUME to control running threads in
resume/stop/singleStep
* change manual stop signal from SIGTRAP to SIGSTOP to make manual stop
handling simpler
* change (*nativeProcess).trapWaitInternal to suspend newly created
threads when we are stepping a thread
* change (*nativeProcess).trapWaitInternal to handle some unhandled
stop events
* remove misleading (*nativeProcess).waitFast which does not do
anything different from the normal wait variant
* rewrite (*nativeProcess).stop to only set breakpoints for threads of
which we have received SIGTRAP
* rewrite (*nativeThread).singleStep to actually execute a single
instruction and to properly route signals

### DIFF
--- a/pkg/proc/native/proc_freebsd.go
+++ b/pkg/proc/native/proc_freebsd.go
@@ -42,6 +42,10 @@ const (
 type osProcessDetails struct {
 	comm string
 	tid  int
+
+	delayedSignal  syscall.Signal
+	trapThreads    []int
+	selectedThread *nativeThread
 }
 
 func (os *osProcessDetails) Close() {}
@@ -83,7 +87,6 @@ func Launch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs []str
 		process.Stdout = stdout
 		process.Stderr = stderr
 		process.SysProcAttr = &syscall.SysProcAttr{Ptrace: true, Setpgid: true, Foreground: foreground}
-		process.Env = proc.DisableAsyncPreemptEnv()
 		if foreground {
 			signal.Ignore(syscall.SIGTTOU, syscall.SIGTTIN)
 		}
@@ -152,7 +155,12 @@ func (dbp *nativeProcess) kill() (err error) {
 	if dbp.exited {
 		return nil
 	}
-	dbp.execPtraceFunc(func() { err = ptraceCont(dbp.pid, int(sys.SIGKILL)) })
+	dbp.execPtraceFunc(func() {
+		for _, th := range dbp.threads {
+			ptraceResume(th.ID)
+		}
+		err = ptraceCont(dbp.pid, int(sys.SIGKILL))
+	})
 	if err != nil {
 		return err
 	}
@@ -165,7 +173,7 @@ func (dbp *nativeProcess) kill() (err error) {
 
 // Used by RequestManualStop
 func (dbp *nativeProcess) requestManualStop() (err error) {
-	return sys.Kill(dbp.pid, sys.SIGTRAP)
+	return sys.Kill(dbp.pid, sys.SIGSTOP)
 }
 
 // Attach to a newly created thread, and store that thread in our list of
@@ -178,7 +186,7 @@ func (dbp *nativeProcess) addThread(tid int, attach bool) (*nativeThread, error)
 	var err error
 	dbp.execPtraceFunc(func() { err = sys.PtraceLwpEvents(dbp.pid, 1) })
 	if err == syscall.ESRCH {
-		if _, _, err = dbp.waitFast(dbp.pid); err != nil {
+		if _, _, err = dbp.wait(dbp.pid, 0); err != nil {
 			return nil, fmt.Errorf("error while waiting after adding process: %d %s", dbp.pid, err)
 		}
 	}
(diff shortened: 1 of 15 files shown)


pkg/proc/native/proc_freebsd.go
```diff
@@ -269,10 +298,10 @@ func (dbp *nativeProcess) trapWaitInternal(pid int, halt bool) (*nativeThread, e
 					}
 					return nil, err
 				}
-				if halt {
-					return nil, nil
+				if mode == trapWaitStepping {
+					dbp.execPtraceFunc(func() { ptraceSuspend(tid) })
```
derekparker, line 302: Is there any reason not to do this unconditionally? By that I mean just automatically suspending the thread in `dbp.addThread` for FreeBSD.

pkg/proc/native/proc_freebsd.go
```diff
@@ -348,10 +373,34 @@ func (dbp *nativeProcess) resume() error {
 			thread.CurrentBreakpoint.Clear()
 		}
 	}
+	if len(dbp.os.trapThreads) > 0 {
+		// On these threads we have already received a SIGTRAP while stepping on a different thread,
+		// do not resume the process, instead set a breakpoint on them.
```
derekparker, line 378: This comment reads a bit confusing as we're not actually setting a breakpoint here but acknowledging breakpoints that have been hit. I know `stop2` calls `SetCurrentBreakpoint` but it's confusing in this context.

pkg/proc/native/proc_freebsd.go
```diff
@@ -348,10 +373,34 @@ func (dbp *nativeProcess) resume() error {
...
+	})
+	if err != nil {
+		return err
+	}
+	sig := int(dbp.os.delayedSignal)
+	dbp.os.delayedSignal = 0
+	for _, thread := range dbp.threads {
+		thread.Status = nil
+	}
+	return dbp.ptraceCont(sig)
```
derekparker, line 403: Perhaps we should also batch this into the `execPtraceFunc` above to minimize overhead.

pkg/proc/native/proc_freebsd.go
```diff
@@ -360,19 +409,70 @@ func (dbp *nativeProcess) stop(cctx *proc.ContinueOnceContext, trapthread *nativ
...
+	}
+
+	if trapthread.Status == nil || (*sys.WaitStatus)(trapthread.Status).StopSignal() != sys.SIGTRAP {
+		return trapthread, nil
+	}
+
+	return dbp.stop2(trapthread.ID)
+}
+
+func (dbp *nativeProcess) stop2(tids ...int) (*nativeThread, error) {
```
derekparker, line 433: nit: I think we can improve this name a bit. Perhaps something like `postStop` or similar? We don't really stop anything here, we only really acknowledge breakpoints that have been hit.

pkg/proc/native/threads.go
```diff
@@ -22,26 +22,6 @@ type nativeThread struct {
 	common         proc.CommonThread
 }
 
-// Continue the execution of this thread.
-//
-// If we are currently at a breakpoint, we'll clear it
-// first and then resume execution. Thread will continue until
-// it hits a breakpoint or is signaled.
-func (t *nativeThread) Continue() error {
```
derekparker, line 30: Why remove the generic version of this?

derekparker (changes requested):

pkg/proc/native/proc_freebsd.go
aarzilli: If this happens when we are stepping (mode == trapWaitStepping) we want to suspend new threads, like every other thread is suspended, and continue until the thread we are stepping finishes stepping.
If this happens when we are running the process (mode == trapWaitNormal) we want the new thread to keep running, like every other thread.

pkg/proc/native/threads.go
aarzilli: IMHO it's misleading to have this function defined for freebsd and windows, since in either case it is not possible to resume the execution of individual threads. Also, given how it's used on linux, we should probably replace it with a direct call to (*nativeThread).resume but I didn't want to do it in the middle of a bunch of changes to the freebsd backend (maybe the same is also true for the macos backend, but I haven't looked).

pkg/proc/native/proc_freebsd.go
```diff
@@ -152,7 +155,12 @@ func (dbp *nativeProcess) kill() (err error) {
 	if dbp.exited {
 		return nil
 	}
-	dbp.execPtraceFunc(func() { err = ptraceCont(dbp.pid, int(sys.SIGKILL)) })
+	dbp.execPtraceFunc(func() {
+		for _, th := range dbp.threads {
+			ptraceResume(th.ID)
```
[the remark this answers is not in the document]
aarzilli: Seems so, if I remove it it will get stuck.

pkg/proc/native/threads.go
derekparker: Ack, ok that makes sense.

derekparker (approved):
LGTM

PR #3229, opened 2022-12-21
proc: populate pointer values

andreimatei:
This patch changes how Value and Unreadable are populated for pointer
Variables. Before this patch, variables of kind reflect.Ptr did not have
their Value field populated. This patch populates it in
Variable.loadValue(), which seems natural and consistent with other
types of variables. The Value is the address that the pointer points to.
The Unreadable field was populated inconsistently for pointer variables:
it was never populated for an outer pointer, but it could be populated
for an inner pointer in pointer-to-pointer types. Before this patch,
in pointer whose value could not be read was not easily distinguishable
from a pointer with a value that could be read, but that cannot be
dereferenced (i.e. a dangling pointer): neither of these would be marked
as Unreadable, and both would have a child marked as Unreadable. This
patch makes it so that a pointer variable whose pointer value cannot be
read is marked as Unreadable.

Using this new distinction, this patch fixes a bug around dereferencing
dangling pointers: before, attempting such a dereference produced a
"nil pointer dereference" error. This was bogus, since the pointer was
not nil. Now, we are more discerning and generate a different error.

### DIFF
--- a/_fixtures/testunsafepointers.go
+++ b/_fixtures/testunsafepointers.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"runtime"
+	"unsafe"
+)
+
+func main() {
+	// We're going to produce a pointer with a bad address.
+	badAddr := uintptr(0x42)
+	unsafeDanglingPtrPtr := unsafe.Pointer(badAddr)
+	// We produce a **int, instead of more simply a *int, in order for the test
+	// program to test more complex Delve behavior.
+	danglingPtrPtr := (**int)(unsafeDanglingPtrPtr)
+	_ = danglingPtrPtr
+	runtime.Breakpoint()
+}
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -124,6 +124,9 @@ type Variable struct {
 	// number of elements to skip when loading a map
 	mapSkip int
 
+	// Children lists the variables sub-variables. What constitutes a child
+	// depends on the variable's type. For pointers, there's one child
+	// representing the pointed-to variable.
 	Children []Variable
 
 	loaded     bool
@@ -1244,6 +1247,40 @@ func (v *Variable) maybeDereference() *Variable {
 	}
 }
 
+// loadPtr assumes that v is a pointer and loads its value. v also gets a child
+// variable, representing the pointed-to value. If v is already loaded,
+// loadPtr() is a no-op.
+func (v *Variable) loadPtr() {
+	if len(v.Children) > 0 {
+		// We've already loaded this variable.
+		return
+	}
+
+	t := v.RealType.(*godwarf.PtrType)
+	v.Len = 1
+
+	var child *Variable
+	if v.Unreadable == nil {
+		ptrval, err := readUintRaw(v.mem, v.Addr, t.ByteSize)
+		if err == nil {
+			child = v.newVariable("", ptrval, t.Type, DereferenceMemory(v.mem))
+		} else {
+			// We failed to read the pointer value; mark v as unreadable.
+			v.Unreadable = err
+		}
+	}
+
+	if v.Unreadable != nil {
+		// Pointers get a child even if their value can't be read, to
+		// maintain backwards compatibility.
+		child = v.newVariable("", 0 /* addr */, t.Type, DereferenceMemory(v.mem))
+		child.Unreadable = fmt.Errorf("parent pointer unreadable: %w", v.Unreadable)
+	}
+
+	v.Children = []Variable{*child}
+	v.Value = constant.MakeUint64(v.Children[0].Addr)
+}
+
 func loadValues(vars []*Variable, cfg LoadConfig) {
 	for i := range vars {
 		vars[i].loadValueInternal(0, cfg)
@@ -1263,8 +1300,7 @@ func (v *Variable) loadValueInternal(recurseLevel int, cfg LoadConfig) {
 	v.loaded = true
 	switch v.Kind {
 	case reflect.Ptr, reflect.UnsafePointer:
-		v.Len = 1
-		v.Children = []Variable{*v.maybeDereference()}
+		v.loadPtr()
 		if cfg.FollowPointers {
 			// Don't increase the recursion level when dereferencing pointers
 			// unless this is a pointer to interface (which could cause an infinite loop)
--- a/pkg/proc/variables_test.go
+++ b/pkg/proc/variables_test.go
@@ -1610,3 +1610,57 @@ func TestEvalExpressionGenerics(t *testing.T) {
 		}
 	})
 }
+
+// Test the behavior when reading dangling pointers produced by unsafe code.
+func TestBadUnsafePtr(t *testing.T) {
+	withTestProcess("testunsafepointers", t, func(p *proc.Target, fixture protest.Fixture) {
+		assertNoError(p.Continue(), t, "Continue()")
+
+		// danglingPtrPtr is a pointer with value 0x42, which is an unreadable
+		// address.
+		danglingPtrPtr, err := evalVariableWithCfg(p, "danglingPtrPtr", pnormalLoadConfig)
+		assertNoError(err, t, "eval returned an error")
+		t.Logf("danglingPtrPtr (%s): unreadable: %v. addr: 0x%x, value: %v",
+			danglingPtrPtr.TypeString(), danglingPtrPtr.Unreadable, danglingPtrPtr.Addr, danglingPtrPtr.Value)
+		assertNoError(danglingPtrPtr.Unreadable, t, "danglingPtrPtr is unreadable")
+		if val := danglingPtrPtr.Value; val != nil {
+			t.Fatalf("Value unexpectedly set for danglingPtrPtr: %v", val)
+		}
+		if len(danglingPtrPtr.Children) != 1 {
+			t.Fatalf("expected 1 child, got: %d", len(danglingPtrPtr.Children))
+		}
+
+		badPtr, err := evalVariableWithCfg(p, "*danglingPtrPtr", pnormalLoadConfig)
+		assertNoError(err, t, "error evaluating *danglingPtrPtr")
+		t.Logf("badPtr: (%s): unreadable: %v. addr: 0x%x, value: %v",
+			badPtr.TypeString(), badPtr.Unreadable, badPtr.Addr, badPtr.Value)
+		// *danglingPtrPtr is dereferencing the unreadable address, the
+		// expression can't be properly evaluated, but we don't mark pointer
+		// variables as Unreadable.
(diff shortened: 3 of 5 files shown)


aarzilli:
> the API does not permit asking for a pointer's value

It's v.Children[0].Addr.

pkg/proc/variables.go
```diff
@@ -1233,12 +1233,11 @@ func (v *Variable) maybeDereference() *Variable {
 			return &v.Children[0]
 		}
 		ptrval, err := readUintRaw(v.mem, v.Addr, t.ByteSize)
-		r := v.newVariable("", ptrval, t.Type, DereferenceMemory(v.mem))
 		if err != nil {
-			r.Unreadable = err
+			v.Unreadable = err
```
aarzilli, line 1237: The change to this function should be reverted, it should still return a new, dereferenced, variable, even if there is a read error. Otherwise pointers that contain a bad address will end up with themselves as the first child.

pkg/proc/variables.go
```diff
@@ -1264,6 +1263,15 @@ func (v *Variable) loadValueInternal(recurseLevel int, cfg LoadConfig) {
 	switch v.Kind {
 	case reflect.Ptr, reflect.UnsafePointer:
 		v.Len = 1
+
```
aarzilli, line 1266: Rather than reading the value twice use the value of v.Children[0].Addr.

aarzilli:
I'm not sure that we should do this, given that there is already a way to retrieve the address contained in the pointer.

andreimatei (commented):
> > the API does not permit asking for a pointer's value

> It's v.Children[0].Addr.

Ack.
But something is bothering me about the relationship between a pointer and it's child. Isn't the scheme ambiguous with respect to the `Unreadable` field? Consider two cases:
1. A pointer whose value cannot be read (`v.Addr` cannot be read):
```
var v = &Variable{
	Addr: 0xbadc0ffee,
	DwarfType: &godwarf.PtrType{
		Type: &godwarf.IntType{},
	},
}
```
When calling `v.loadValue()`, `v` will *not* have its `Unreadable` flag set; `v` will get a child marked `Unreadable`.

2. A pointer whose value can be read, but is pointing to an unreadable address.
```
var v = &Variable{
	Addr: 0xg00dc0ffee,
	DwarfType: &godwarf.PtrType{
		Type: &godwarf.IntType{},
	},
}
*0xg00dc0ffee == 0xbadc0ffee
```

Like before, when calling `v.loadValue()`, `v` will not have its `Unreadable` flag set; `v` will get a child. If `LoadConfig.FollowPointers` is set, `v.Children[0].LoadValue()` is called, which set `v.Children[0].Unreadable`.
The two cases are indistinguishable, aren't they? Isn't this unfortunate?

So, wouldn't it be better if, in the first case, `v.Unread` would be set (and probably there should be no child)? If you agree with this, would you consider this an extra argument for also setting `v.Value` for pointers?

---

> I'm not sure that we should do this, given that there is already a way to retrieve the address contained in the pointer.

Feel free to close this PR if there's nothing good here.
Without knowing what I'm talking about, I expected a pointer to have it's `Value` set, and didn't think of going to the children for it. On a related note, it also seemed to me that, if `Load.Config.FollowPointers == false`, then I'd expect a pointer `Variable` to not even get a child.

While I still think populating the `Value` makes sense, I now realize that variables that have `Children` currently never seem to have a `Value`, and maybe there's value in this rule.
Again, if you don't buy any of this, please feel free to close and sorry for wasting your time.

pkg/proc/variables.go
aarzilli: > Done. But this is incorrect if the maybeDereference() function took the early return, returning itself

the early return that says `if v.Unreadable != nil`? We return early from loadValueInternal anyway if unreadable is set.

aarzilli (commented):
> So, wouldn't it be better if, in the first case, `v.Unread` would be set (and probably there should be no child)?

I think you make a good argument for setting the v.Unreadable, but removing the child is too much since this is an api that is consumed externally.

I think it would be fine if the reflect.Ptr branch of loadValueInternal copied the Unreadable field from the variable returned by maybeDereference into v.

> While I still think populating the `Value` makes sense, I now realize that variables that have `Children` currently never seem to have a `Value`, and maybe there's value in this rule.

This is not true, complex numbers are an exception.

andreimatei:
Please see now; I've massaged the code more.
Now this patch fixes what I believe to be a real bug, in addition to the rest of the purported rationalizations: a bogus "nil pointer dereference" error that you would get when attempting to dereference a dangling (but not nil) pointer.
As before, pointers get their `Value` set. And now pointers also have their `Unreadable` field set in what I believe to be a consistent manner: if `v.Addr` cannot be read, `v.Unreadable` is set. If the variable is a child of an unreadable pointer (so `v.Addr` is not set at all), then `Unreadable` is again set.
There's now a test to highlight the effects of the patch.

Please take a look and, as always, feel free to close and tell me to go away if this is not improving things.
Happy holidays!

pkg/proc/variables_test.go
```diff
@@ -1610,3 +1610,62 @@ func TestEvalExpressionGenerics(t *testing.T) {
...
+			t.Fatalf("expected x to be unreadable, but got value: %v", badPtrChild.Value)
+		}
+
+		// Evaluating **danglingPtrPtr fails.
+		_, err = evalVariableWithCfg(p, "**danglingPtrPtr", pnormalLoadConfig)
+		if err == nil {
+			t.Fatalf("expected error doing **danglingPtrPtr")
+		}
+		expErr := "couldn't read pointer: input/output error"
+		if !strings.Contains(err.Error(), expErr) {
```
aarzilli, line 1667: The specific error is different on different platforms, this test fails on macOS, windows and freebsd. You should just check that the error is not nil.

pkg/proc/variables.go
```diff
@@ -1220,28 +1223,51 @@ func extractVarInfoFromEntry(tgt *Target, bi *BinaryInfo, image *Image, regs op.
 	return v, nil
 }
 
-// If v is a pointer a new variable is returned containing the value pointed by v.
+// If v is a pointer, its child value is loaded and returned as a new variable.
```
aarzilli, line 1226: maybeDereference gets called in a lot of places, I'm not comfortable with changing its semantics so much like this, especially without chaning its name and signature.

_fixtures/testunsafepointers.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 The Cockroach Authors.
```
aarzilli, line 1: Does this need to be here?

pkg/proc/variables.go
andreimatei: By "changing its semantics" you mean the fact that `maybeDereference` now mutates `v`?
Or the fact that this part is gone?
```
if v.Unreadable != nil {
   return v
}
```

I've reverted this function to the how it was before, but I've had to change `evalPointerDeref()` to not use it any more, in order to avoid the bogus "nil pointer dereference" in the unreadable case. Please consider whether you're indeed happy with `maybeDereference()`, or whether some different semantics would actually be better for all callers.

_fixtures/testunsafepointers.go
andreimatei: no, sorry; this was just my editor's template

aarzilli (approved):
LGTM

andreimatei:
Thanks @aarzilli !
Should I do something here to get this merged? I see some test failures, but I don't know if they're mine :)

aarzilli:
You don't have to do anything, Derek will review/merge this when he has time.

pkg/proc/variables.go
```diff
@@ -1244,6 +1247,41 @@ func (v *Variable) maybeDereference() *Variable {
...
+// loadPtr assumes that v is a pointer and loads its value. v also gets a child
+// variable, representing the pointed-to value. If v is already loaded,
+// loadPtr() is a no-op.
+func (v *Variable) loadPtr() {
+	if v.loaded {
+		return
+	}
+	v.loaded = true
+
+	t := v.RealType.(*godwarf.PtrType)
```
derekparker, line 1259: This could panic, but I think that's ok as it would signify a programmer error, we shouldn't be calling this on non-pointer types.

pkg/proc/variables.go
```diff
@@ -1244,6 +1247,41 @@ func (v *Variable) maybeDereference() *Variable {
...
+			child = v.newVariable("", ptrval, t.Type, DereferenceMemory(v.mem))
+		} else {
+			// We failed to read the pointer value; mark v as unreadable.
+			v.Unreadable = err
+			// Pointers get a child even if their value can't be read, to
+			// maintain backwards compatibility.
+			child = v.newVariable("", 0 /* addr */, t.Type, DereferenceMemory(v.mem))
+			child.Unreadable = fmt.Errorf("parent pointer unreadable: %w", err)
+		}
+	} else {
```
derekparker, line 1275: Both this and the case above produce identical code, perhaps we could dry this up a bit. Maybe we should remove the `else` clause and then trigger the error condition on `v.Unreadable != nil` since that should get hit if it was already set or if the `readUintRaw` failed.

pkg/proc/variables.go
```diff
@@ -1244,6 +1247,41 @@ func (v *Variable) maybeDereference() *Variable {
 	}
 }
 
+// loadPtr assumes that v is a pointer and loads its value. v also gets a child
+// variable, representing the pointed-to value. If v is already loaded,
+// loadPtr() is a no-op.
+func (v *Variable) loadPtr() {
+	if v.loaded {
```
derekparker, line 1254: I think checking `v.loaded` might be the wrong thing to check here, since it complicates the code in `loadValueInternal` a bit (adding the defer and special casing the `time.Time` case). Perhaps we just check if this variable already has children since no matter what after this method is called v will have `len(v.Children) > 0`.

derekparker (changes requested):
Just a few minor changes, thanks for the patch! Sorry for the review delay, holidays and such.

pkg/proc/variables.go
andreimatei: Done.

pkg/proc/variables.go
andreimatei: right; this is called when `v.Kind` is `reflect.Ptr` or `reflect.UnsafePointer`. I expect `v.RealType` to be `*godwarf.PtrType`. That's sane, right?
In my opinion we shouldn't be defensive here, but if you feel otherwise, let me know.

pkg/proc/variables.go
andreimatei: done

andreimatei:
> Just a few minor changes, thanks for the patch! Sorry for the review delay, holidays and such.

Thanks for the review! No rush on my side.

pkg/proc/variables.go
derekparker: Agreed.

derekparker (approved):
LGTM

PR #3232, opened 2022-12-29
cmd/dlv: match go test behavior when dlv test gets a list of .go files

### LINKED ISSUE #3230: BUG: Delve runs tests using wrong working directory

1. What version of Delve are you using (`dlv version`)?   1.9.1
2. What version of Go are you using? (`go version`)? 1.19.4 darwin/arm64
3. What operating system and processor architecture are you using? MacOS Ventura arm64
4. What did you do? dlv est reader/*.go -test.v, test case prints result of os.Getwd()
5. What did you expect to see? package directory
6. What did you see instead? CWD were delve was executed

When _go test_ runs a test case, it performs a `chdir` into the package directory before executing the test. The _dlv test_ command does not perform the same `chdir`, instead executing from the current directory from which _dlv test_ was executed.

This makes it difficult for test cases to open input files for use in testing, because there is no path to the input file that works in both cases.

There are two workarounds:

1. CWD to the package directory before running _dlv_
2. Use a clever hack exploiting runtime.Caller(0) described <a href="https://intellij-support.jetbrains.com/hc/en-us/community/posts/360009685279-Go-test-working-directory-keeps-changing-to-dir-of-the-test-file-instead-of-value-in-template">here</a>. Apparently the issue came up in Intellij as well.

The hack approach relies on knowing the relative path relationship between a special package and the project root, and uses this to `chdir` to the project root in both _dlv test_ and _go test_. This isn't a great resolution, but at least it offers consistency.

My general sense is that _dlv test_ should be replicating the environment provided by _go test_ as exactly as (reasonably) possible.

aarzilli:
When 'dlv test' is called with a list of files it should match the
behavior of 'go test' and chdir to the directory where the go files are
(if they are all in the same directory).

Fixes #3230

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -754,11 +754,7 @@ func testCmd(cmd *cobra.Command, args []string) {
 		processArgs := append([]string{debugname}, targetArgs...)
 
 		if workingDir == "" {
-			if len(dlvArgs) == 1 {
-				workingDir = getPackageDir(dlvArgs[0])
-			} else {
-				workingDir = "."
-			}
+			workingDir = getPackageDir(dlvArgs)
 		}
 
 		return execute(0, processArgs, conf, "", debugger.ExecutingGeneratedTest, dlvArgs, buildFlags)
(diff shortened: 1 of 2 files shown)


derekparker (approved):
LGTM

jsshapiro:
Uh, I'm not convinced that the description of chdir behavior is right. When go test is presented with

    go test ./...

I believe that it executes a chdir independently for each test so that the execution happens in the directory containing the test. It is not a requirement that all of the tests be in a common directory. In fact, that would defeat the purpose of the chdir.

Easy to confirm, if you would like me to do so.

Issue #3238, opened 2023-01-03
can't figure out source-list-* configs


thockin:
1. What version of Delve are you using (`dlv version`)?

```
$ dlv version
Delve Debugger
Version: 1.20.1
Build: $Id: 96e65b6c615845d42e0e31d903f6475b0e4ece6e $
```

2. What version of Go are you using? (`go version`)?

```
$ go version
go version go1.19.1 linux/amd64
```

3. What operating system and processor architecture are you using?

Linux AMD64

Trying to use the color configs, and having 2 different problems.

First, I started by uncommenting the examples I found somewhere:

```
source-list-keyword-color: "\x1b[0m"
source-list-string-color: "\x1b[92m"
source-list-number-color: "\x1b[0m"
source-list-comment-color: "\x1b[95m"
source-list-arrow-color: "\x1b[93m"
```

This does some coloring but not readable for me.  I want to change it:

```
(dlv) config -list
...
source-list-line-color    <nil>
source-list-arrow-color   "\x1b[93m"
source-list-keyword-color "\x1b[0m"
source-list-string-color  "\x1b[92m"
source-list-number-color  "\x1b[0m"
source-list-comment-color "\x1b[95m"
...

(dlv) config source-list-string-color "\x1b[96m"

(dlv) config -list
...
source-list-arrow-color   "\x1b[93m"
source-list-keyword-color "\x1b[0m"
source-list-string-color  "\"\\x1b[96m\""
source-list-number-color  "\x1b[0m"
source-list-comment-color "\x1b[95m"
...
```

Not the extra quotes.  Weird that I can't cut-paste the output to the input.  OK:

```
(dlv) config source-list-string-color \x1b[96m

(dlv) config -list
...
source-list-line-color    <nil>
source-list-arrow-color   "\x1b[93m"
source-list-keyword-color "\x1b[0m"
source-list-string-color  "\\x1b[96m"
source-list-number-color  "\x1b[0m"
source-list-comment-color "\x1b[95m"
...
```

OK, so how am I supposed to do this?

Second problem: `source-list-line-color` looks interesting

```
(dlv) config source-list-line-color 35
Command failed: unsupported type for configuration key "source-list-line-color"

(dlv) config source-list-line-color "\x1b[35m"
Command failed: unsupported type for configuration key "source-list-line-color"

(dlv) config source-list-line-color \x1b[35m
Command failed: unsupported type for configuration key "source-list-line-color"
```

I have no clue what to put here.

[...]

PR #3239, opened 2023-01-03
Handle config cmd values with double-quotes

thockin:
Before this it would retain quotes.

xref #3238


thockin:
Closing in favor of #3240 which includes a similar fix

PR #3240, opened 2023-01-03
pkg/terminal: fix handling list colors via config

derekparker:
Parsing the source list color when set during a Delve debug session resulted in unexpected errors. Additionally the changes were not reflected in the current session, forcing the user to save the config and start a new session. This patch fixes those issues.

Fixes https://github.com/go-delve/delve/issues/3238

### DIFF
--- a/pkg/config/split.go
+++ b/pkg/config/split.go
@@ -144,24 +144,20 @@ func writeField(w io.Writer, field reflect.Value, fieldName string) {
 	case reflect.Interface:
 		switch field := field.Interface().(type) {
 		case string:
-			fmt.Fprintf(w, "%s\t= %q\n", fieldName, field)
+			fmt.Fprintf(w, "%s\t%q\n", fieldName, field)
 		default:
-			fmt.Fprintf(w, "%s\t= %v\n", fieldName, field)
+			fmt.Fprintf(w, "%s\t%v\n", fieldName, field)
 		}
 	case reflect.Ptr:
 		if !field.IsNil() {
-			fmt.Fprintf(w, "%s\t= %v\n", fieldName, field.Elem())
+			fmt.Fprintf(w, "%s\t%v\n", fieldName, field.Elem())
 		} else {
-			fmt.Fprintf(w, "%s\t= <not defined>\n", fieldName)
+			fmt.Fprintf(w, "%s\t<not defined>\n", fieldName)
 		}
 	case reflect.String:
-		if field.IsZero() {
-			fmt.Fprintf(w, "%s\t= <not defined>\n", fieldName)
-		} else {
-			fmt.Fprintf(w, "%s\t= %v\n", fieldName, field)
-		}
+		fmt.Fprintf(w, "%s\t%q\n", fieldName, field)
 	default:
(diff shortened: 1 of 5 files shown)


derekparker:
@thockin can you test this out?

pkg/config/split.go
```diff
@@ -84,7 +84,29 @@ func ConfigureSetSimple(rest string, cfgname string, field reflect.Value) error
 			v := rest == "true"
 			return reflect.ValueOf(&v), nil
 		case reflect.String:
+			var err error
+			rest, err = strconv.Unquote(rest)
```
thockin: Thios will fail if it is not quoted, which will break anyone who relies on YAML sloppy strings or even contextual interp of numbers as strings.  Maybe not important for these fields, but it IS a breaking change.

see https://github.com/go-delve/delve/pull/3239/files

pkg/config/split.go
```diff
@@ -84,7 +84,29 @@ func ConfigureSetSimple(rest string, cfgname string, field reflect.Value) error
...
+			// We special case this particular configuration key because historically we accept both a numerical value and a string value for it.
+			if cfgname == "source-list-line-color" {
+				n, err := strconv.Atoi(rest)
+				if err == nil {
+					if n < 0 {
+						return reflect.ValueOf(nil), fmt.Errorf("argument to %q must be a number greater than zero", cfgname)
+					}
+					return reflect.ValueOf(&n), nil
+				}
+				rest, err = strconv.Unquote(rest)
```
thockin: same

pkg/terminal/terminal.go
```diff
@@ -141,6 +141,34 @@ func New(client service.Client, conf *config.Config) *Term {
 	return t
 }
 
+func (t *Term) updateColorScheme() {
```
thockin, line 144: remove the dup block in `New()` and call this?

thockin (commented):
It works for me

pkg/config/split.go
derekparker: Absolutely, you're right. Was working on some follow-up fixes / refactoring after submitting for testing. They are included in this PR now.

pkg/config/split.go
```diff
@@ -84,7 +84,28 @@ func ConfigureSetSimple(rest string, cfgname string, field reflect.Value) error
 			v := rest == "true"
 			return reflect.ValueOf(&v), nil
 		case reflect.String:
+			unquoted, err := strconv.Unquote(rest)
```
thockin, line 87: How do you want to handle cases like `"forgot the last quote` ?

My PR handles it as an error (clearly they meant to specify a quoted string).  This PR will treat it as a literal (which is more back-compatible, but less correct.

You decide - the rest LGTM :)

derekparker:
Fixed a few failing tests due to changes in how config output is presented, but that's really a minor change inconsequential to the overall change. With that in mind, I may still back it out in case there are clients that depend on the current format. @suzmue @polinasok @hyangah there was a DAP test which failed, does the VSCode DAP client depend on parsing the config format in a specific format at all? If so I can revert that change.

aarzilli (approved):
LGTM

PR #3243, opened 2023-01-03
Add config options for tab printing

thockin:
New config "tab" allows user to set a string instead of literal tab (e.g. "... ").

New config "source-list-tab-color" allows that string to be colored.

This builds on top of the `updateColorScheme` change  #3240

### DIFF
--- a/pkg/terminal/colorize/colorize.go
+++ b/pkg/terminal/colorize/colorize.go
@@ -23,17 +23,28 @@ const (
 	CommentStyle
 	LineNoStyle
 	ArrowStyle
+	TabStyle
 )
 
 // Print prints to out a syntax highlighted version of the text read from
 // reader, between lines startLine and endLine.
-func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arrowLine int, colorEscapes map[Style]string) error {
+func Print(out io.Writer, path string, reader io.Reader, startLine, endLine, arrowLine int, colorEscapes map[Style]string, altTabStr string) error {
 	buf, err := ioutil.ReadAll(reader)
 	if err != nil {
 		return err
 	}
 
-	w := &lineWriter{w: out, lineRange: [2]int{startLine, endLine}, arrowLine: arrowLine, colorEscapes: colorEscapes}
+	w := &lineWriter{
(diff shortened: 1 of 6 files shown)


thockin:
I also wanted to say that I find delve easy to hack on - good job!

pkg/terminal/colorize/colorize.go
```diff
@@ -284,6 +302,11 @@ func (w *lineWriter) Write(style Style, data []byte, last bool) {
 			}
 			cur = i + 1
 		}
+		if data[i] == '\t' {
```
aarzilli, line 305: Turn this and the if above into a single switch on `data[i]`. Also since tab is only ever called once you could inline it here.

pkg/terminal/config.go
```diff
@@ -25,6 +25,8 @@ func configureCmd(t *Term, ctx callContext, args string) error {
 		if t.client != nil { // only happens in tests
 			lcfg := t.loadConfig()
 			t.client.SetReturnValuesLoadConfig(&lcfg)
+			t.updateColorScheme()
```
aarzilli, line 28: It seems to me that these two are always called together and calling only one of them would probably be an error. So they should probably be a single function, which should be called `updateColorizer` or something like that.

aarzilli (commented):
This conflicts with the just-merged change for the colors, so it needs to be rebased.

pkg/terminal/colorize/colorize.go
thockin: I followed the `nl()` convention, but sure.

pkg/terminal/config.go
thockin: ACK.  I chose`updateConfig` is that OK?

thockin:
re-pushed

thockin:
PTAL - retooled to "tabStr" is really an "alternative" and can be "" all the way until the conversion to bytes.  This means cases that create a Term do not need to init that field (O, what I would give for proper constructors or flexible default values for fields!)

derekparker (approved):
LGTM

aarzilli (approved):

PR #3245, opened 2023-01-04
fix stripping DYLD_INSERT_LIBRARIES on macOS

aviramha:
I just noticed that my fix actually didn't work since I had ill assumption (didn't notice) that the code isn't iterating keys, but `key=value` format, hence this never did anything (I tested the previous iteration where it was HasPrefix and probably didn't test the new one, sorry for that)

https://github.com/go-delve/delve/pull/3181


derekparker (approved):
LGTM

Should have caught this in the review. Thanks for the follow up patch!

PR #3246, opened 2023-01-05
debugger,terminal: Allow setting and clearing breakpoints after process exit

### LINKED ISSUE #3242: set breakpoint after exit?

```
$ dlv version
Delve Debugger
Version: 1.20.1
Build: $Id: 96e65b6c615845d42e0e31d903f6475b0e4ece6e $
```

```
$ go version
go version go1.19.1 linux/amd64
```

Linux AMD64

I feel like this SHOULD work?

```
(dlv) c
Process 2330606 has exited with status 0

(dlv) bp
Breakpoint runtime-fatal-throw (enabled) at 0x0 for runtime.fatal() /usr/local/go/src/runtime/panic.go:1059 (0)
Breakpoint unrecovered-panic (enabled) at 0x0 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1143 (0)
	print runtime.curg._panic.arg

(dlv) b getPackages
Process 2330606 has exited with status 0

(dlv) bp
Breakpoint runtime-fatal-throw (enabled) at 0x0 for runtime.fatal() /usr/local/go/src/runtime/panic.go:1059 (0)
Breakpoint unrecovered-panic (enabled) at 0x0 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1143 (0)
	print runtime.curg._panic.arg
```

derekparker:
Fixes #3242

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -734,6 +734,13 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint, locExpr string,
 			return locs[0].PCs
 		}
 	}
+	// If the process has already exited, set a suspended breakpoint so that
+	// it can be enabled later when the process is started again.
+	if ok, err := d.target.Valid(); !ok {
+		if _, ok := err.(proc.ErrProcessExited); ok {
+			suspended = true
+		}
+	}
 	createdBp, err := createLogicalBreakpoint(d, requestedBp, &setbp, suspended)
 
 	if err != nil {
@@ -1893,9 +1900,9 @@ func (d *Debugger) FindLocation(goid int64, frame, deferredCall int, locStr stri
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
 
-	if _, err := d.target.Valid(); err != nil {
-		return nil, err
-	}
+	// if _, err := d.target.Valid(); err != nil {
+	// 	return nil, err
+	// }
 
 	loc, err := locspec.Parse(locStr)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


service/debugger/debugger.go
```diff
@@ -1922,11 +1953,10 @@ func (d *Debugger) FindLocationSpec(goid int64, frame, deferredCall int, locStr
 
 func (d *Debugger) findLocation(goid int64, frame, deferredCall int, locStr string, locSpec locspec.LocationSpec, includeNonExecutableLines bool, substitutePathRules [][2]string) ([]api.Location, error) {
 	locations := []api.Location{}
-	t := proc.ValidTargets{Group: d.target}
-	for t.Next() {
+	for _, t := range d.target.Targets() {
 		pid := t.Pid()
-		s, _ := proc.ConvertEvalScope(t.Target, goid, frame, deferredCall)
-		locs, err := locSpec.Find(t.Target, d.processArgs, s, locStr, includeNonExecutableLines, substitutePathRules)
+		s, _ := proc.ConvertEvalScope(t, goid, frame, deferredCall)
```
aarzilli, line 1958: This is going to fail if the target isn't valid, it will return a nil eval scope which will get passed down to locSpec.Find, depending on what locSpec is this will make it return an error, which will end up returning an error from the whole function.

This is bad because when we add the exec follow mode some of the children will run for a while and then become invalid processes, when this happens it will be impossible to set any new breakpoints.

aarzilli (commented):
IMO it's better to leave the FindLocation stuff as it is and change shouldAskToSuspendBreakpoint in command.go to return true if the process exited. I think that should be all. If the suspended breakpoint looks bad when printed we can do something to make suspended breakpoints look better in general (although they shouldn't necessarily have an address associated).

derekparker:
Done.

aarzilli (approved):
Approving, however I think there's a conflict with the yesno change that will break compilation, so I'm not going to merge it.

derekparker:
@aarzilli rebased.

PR #3247, opened 2023-01-05
godwarf: Attempt to load `DW_AT_specification` if present

brancz:
I noticed this while processing some non-Go debuginfos. The relevant information regarding this type of entry from the DWARF spec:

> A debugging information entry that represents a declaration that completes another (earlier) non-defining declaration may have a `DW_AT_specification` attribute whose value is a reference to the debugging information entry representing the non-defining declaration. A debugging information entry with a `DW_AT_specification` attribute does not need to duplicate information provided by the debugging information entry referenced by that specification attribute.

### DIFF
--- a/pkg/dwarf/godwarf/tree.go
+++ b/pkg/dwarf/godwarf/tree.go
@@ -24,11 +24,13 @@ func (ce compositeEntry) Val(attr dwarf.Attr) interface{} {
 	return nil
 }
 
-// LoadAbstractOrigin loads the entry corresponding to the
-// DW_AT_abstract_origin of entry and returns a combination of entry and its
-// abstract origin.
-func LoadAbstractOrigin(entry *dwarf.Entry, aordr *dwarf.Reader) (Entry, dwarf.Offset) {
-	ao, ok := entry.Val(dwarf.AttrAbstractOrigin).(dwarf.Offset)
+// LoadAbstractOriginAndSpecification loads the entry corresponding to the
+// DW_AT_abstract_origin and/or DW_AT_specification of entry and returns a
+// combination of entry and its abstract origin. If a DIE has both a
+// specification and an abstract origin the specification will be ignored, the
+// DWARF standard is unclear on how this should be handled
+func LoadAbstractOriginAndSpecification(entry *dwarf.Entry, aordr *dwarf.Reader) (Entry, dwarf.Offset) {
+	ao, ok := getAbstractOriginOrSpecification(entry)
 	if !ok {
 		return entry, entry.Offset
 	}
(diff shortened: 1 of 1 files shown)


brancz:
I can't really see an error other than the CI run timing out. Could someone give me a pointer at how I would troubleshoot this? Unit tests seem to pass just fine locally.

aarzilli (commented):
Can a single DIE have both an AttrAbstractOrigin and an AttrSpecification? What would take precedence? If it can't I would just fold LoadSpecification into LoadAbstractOrigin (just check for both attributes at each step and continue with the one you find).
If both can be present still fold everything into LoadAbstractOrigin, and have compositeEntry be a slice representing the visit of the abstract origin/specification tree that the DWARF standard requires.

Either way compositeEntry doesn't need to be exported and it should remain a slice of *dwarf.Entry.

aarzilli:
> I can't really see an error other than the CI run timing out

I think the time out is just a teamcity problem, it doesn't have anything to do with us.

brancz:
I am seeing in this binary:

* An entry having an abstract origin, the abstract origin has a specification
* An entry has a specification

So just loading the spec when there is an abstract origin doesn't work.

brancz:
If I understood you correctly I believe I changed it to the way you requested. Let me know you meant something different.

Thank you so much for the super fast review!

aarzilli:
As long as there isn't both an abstract origin and a specification on the same DIE, it's fine.

```
func LoadAbstractOriginAndSpecification(entry *dwarf.Entry, aordr *dwarf.Reader) (Entry, dwarf.Offset) {
	ao, ok := getAbstractOriginOrSpecification(entry)
	if !ok {
		return entry, entry.Offset
	}

	r := []*dwarf.Entry{entry}

	for {
		aordr.Seek(ao)
		e, _ := aordr.Next()
		if e == nil {
			break
		}
		r = append(r, e)

		ao, ok = getAbstractOriginOrSpecification(e)
		if !ok {
			break
		}
	}

	return compositeEntry(r), entry.Offset
}

func getAbstractOriginOrSpecification(e *dwarf.Entry) (*dwarf.Entry, bool) {
	ao, ok := e.Val(dwarf.AttrAbstractOrigin).(dwarf.Offset)
	if ok {
		return ao, true
	}
	sp, ok := e.Val(dwarf.AttrSpecification).(dwarf.Offset)
	if ok {
		return sp, true
	}
	return nil, false
}
```

brancz:
Yep, that does the trick as well. Adapted the patch to that. Let me know if there is anything else!

pkg/dwarf/godwarf/tree.go
```diff
@@ -24,11 +24,11 @@ func (ce compositeEntry) Val(attr dwarf.Attr) interface{} {
...
 }
 
-// LoadAbstractOrigin loads the entry corresponding to the
-// DW_AT_abstract_origin of entry and returns a combination of entry and its
-// abstract origin.
-func LoadAbstractOrigin(entry *dwarf.Entry, aordr *dwarf.Reader) (Entry, dwarf.Offset) {
-	ao, ok := entry.Val(dwarf.AttrAbstractOrigin).(dwarf.Offset)
+// LoadAbstractOriginAndSpecification loads the entry corresponding to the
+// DW_AT_abstract_origin and/or DW_AT_specification of entry and returns a
+// combination of entry and its abstract origin.
```
aarzilli, line 29: Add a note here: "if a DIE has both a specification and an abstract origin the specification will be ignored, the DWARF standard is unclear on how this should be handled".

brancz: Done!

aarzilli (approved):
LGTM (I don't get why the DWARF standard needs both a DW_AT_specification and a DW_AT_abstract_origin, but whatever, time will tell).

derekparker (approved):
LGTM

PR #3248, opened 2023-01-05
pkg/terminal: use default for Y/n empty response

derekparker:
The typical convention with `[Y/n]` type command line questions is that the capitalized letter represents the default if the user simply hits "enter" instead of typing an actual response. This patch fixes our implementation to use the implied default response.


aarzilli (approved):
LGTM

PR #3253, opened 2023-01-19
Supports sending output to clients when running programs remotely

tttoad:
#3094 
I saw that #3108 has not been updated for a long time, so I committed this PR.


tttoad:
I am currently reading the code related to debug mode, and will support debug mode output write-back in this PR.

service/dap/server.go
```diff
@@ -1082,15 +1140,26 @@ func (s *Session) getPackageDir(pkg string) string {
...
-func (s *Session) newNoDebugProcess(program string, targetArgs []string, wd string) (*exec.Cmd, error) {
+func (s *Session) newNoDebugProcess(program string, targetArgs []string, wd string) (cmd *exec.Cmd, err error) {
 	if s.noDebugProcess != nil {
 		return nil, fmt.Errorf("another launch request is in progress")
 	}
-	cmd := exec.Command(program, targetArgs...)
-	cmd.Stdout, cmd.Stderr, cmd.Stdin, cmd.Dir = os.Stdout, os.Stderr, os.Stdin, wd
-	if err := cmd.Start(); err != nil {
+
+	cmd = exec.Command(program, targetArgs...)
```
aarzilli, line 1148: It's wrong to unconditionally redirect stdout and stderr because programs can behave differently when their output is not a tty. It should only be done if outputMode is "remote".

aarzilli (commented):
This needs to be supported on debugged processes as well, it would be weird to support this only when nodebug (I imagine this is just because it's a WIP). Also it needs a test. The tests are in service/dap/server_test.go.

service/dap/server.go
```diff
@@ -1039,9 +1062,44 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 
 		// Start the program on a different goroutine, so we can listen for disconnect request.
 		go func() {
+			// TODO: Support remote stop program, if the program can not be stopped
+			wg := &sync.WaitGroup{}
+			readerFunc := func(reader io.Reader, category string, wg *sync.WaitGroup) {
+				// Display one line back after outputting one line
+				var scanner = bufio.NewScanner(reader)
```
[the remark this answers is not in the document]
tttoad: The output of the program may take a long time... I've thought about using a timed refresh and a fixed buffer size to achieve this, or maybe that's a better idea.

service/dap/server.go
```diff
@@ -1039,9 +1062,44 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+						} else {
+							fmt.Fprintln(os.Stderr, out)
+						}
+					}
+
+					if s.outputModel&0x0010 != 0 {
+						s.send(&dap.OutputEvent{
+							Event: *newEvent("output"),
+							Body: dap.OutputEventBody{
+								Output:   fmt.Sprintln(out),
```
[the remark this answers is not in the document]
tttoad: `scanner.Text()` will drop n, so I need to add it back.

service/dap/server.go
tttoad: I don't understand... 
When outputMode is not "remote", I will rewrite the output to os.stderr and os.stdout.
Can you give me an error case?

service/dap/server.go
```diff
@@ -159,8 +159,26 @@ type Session struct {
...
+
+	// stderrReader the program's stderr.
+	stderrReader io.ReadCloser
 }
 
+type outputModel int8
+
+const (
+	// osStdMask os.Stdin and os.Stdout
+	osStdMask outputModel = 0x0001
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
```diff
@@ -1025,6 +1043,11 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 	argsToLog.Cwd, _ = filepath.Abs(args.Cwd)
 	s.config.log.Debugf("launching binary '%s' with config: %s", debugbinary, prettyPrint(argsToLog))
 
+	s.outputModel |= osStdMask
+	if args.OutputModel == "remote" {
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
```diff
@@ -1039,9 +1062,44 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
 		// Start the program on a different goroutine, so we can listen for disconnect request.
 		go func() {
+			// TODO: Support remote stop program, if the program can not be stopped
+			wg := &sync.WaitGroup{}
+			readerFunc := func(reader io.Reader, category string, wg *sync.WaitGroup) {
+				// Display one line back after outputting one line
+				var scanner = bufio.NewScanner(reader)
+				for scanner.Scan() {
+					out := scanner.Text()
+					if s.outputModel&0x0001 != 0 {
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
```diff
@@ -1039,9 +1062,44 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+						if category == "stdout" {
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
aarzilli: https://github.com/mattn/go-isatty or https://github.com/pkg/term/

tttoad: I still can't understand...
Will there be a difference between "cmd.stdout->os.stdout" and "cmd.stdout->os.pipe->os.stdout" on non-tty terminal?
Is it possible to call os.Pipe() on a non-tty terminal and get an error?
I'm sorry, I don't know much about non-tty terminal.

aarzilli: >  Will there be a difference between "cmd.stdout->os.stdout" and "cmd.stdout->os.pipe->os.stdout" on non-tty terminal?

All terminals are tty terminals.

>  Is it possible to call os.Pipe() on a non-tty terminal and get an error?

No.

>  I'm sorry, I don't know much about non-tty terminal.

The problem is with tty terminals.

tttoad: Okay, I got it.
done.

tttoad:
@aarzilli @derekparker  PTAL.
I've tested it in the following environments.
(client: macOS arm64. server: ubuntu arm64 in docker)
(client: macOS arm64. server: macOS arm64)
(client: macOS arm64. server: windows amd64)
It can work.
I should be adding test cases next week when you think it's okay.

pkg/util/redirect.go
```diff
@@ -0,0 +1,36 @@
+package util
+
+import (
+	"os"
+	"sync"
+)
+
+var (
+	redirectMap = &RedirectStroe{rs: sync.Map{}}
```
aarzilli, line 9: Sorry but passing around pipes by stuffing them in a global variable is not acceptable.

tttoad: Only in windows system need to use the pipeline...
Using `debugger.Config` to pass the pipeline will change the definition of the `native.Launch` function. 
Is this acceptable to you?

aarzilli: Yes. Also this will eventually need tests, both the nodebug path and the normal path. And Debugger.Restart needs to return an error in case pipe redirects are used (it would be even better if it worked correctly instead of erroring but that's hard and dap doesn't call Restart anyway).

pkg/util/redirect.go
```diff
@@ -0,0 +1,36 @@
+package util
```
aarzilli, line 1: This file shouldn't be used anymore and should be removed.

pkg/proc/redirect/redirect.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"errors"
+	"os"
+)
+
+var ErrorNotImplemented = errors.New("not implemented")
+
+// Redirect
+type Redirect interface {
```
aarzilli, line 11: Make this a concrete type, put it in pkg/proc, not in its own package, make it describe a single redirect rather than three redirects in one. Do not repeat the type name in the name of its methods. For example, it should be (*Redirect).Path not (*Redirect).RedirectPath.

Restart is a word, it shouldn't be spelled `ReStart`.

service/dap/server.go
```diff
@@ -786,6 +807,13 @@ func (s *Session) handleRequest(request dap.Message) {
 }
 
 func (s *Session) send(message dap.Message) {
+	if event, ok := message.(*dap.OutputEvent); ok {
```
aarzilli, line 810: You can't hijack the send method to do other things, it should be done by the caller of this method, or maybe in a method that both waits for the readers to finish and then sends the terminated event.

service/dap/server.go
```diff
@@ -1025,9 +1053,61 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 	argsToLog.Cwd, _ = filepath.Abs(args.Cwd)
 	s.config.log.Debugf("launching binary '%s' with config: %s", debugbinary, prettyPrint(argsToLog))
 
+	var redirected = false
+	s.outputMode |= outputToStd
+	switch args.OutputMode {
```
aarzilli, line 1058: Just have two modes, one that sends stuff through DAP and one that writes to stdout/stderr.

aarzilli (commented):
Still doesn't have tests, also the (*Debugger).Restart function needs to return an error if it is called with pipe redirects.

service/dap/server.go
tttoad: I hope there is a third mode that prints both.

pkg/util/redirect.go
tttoad: OK.

service/dap/server.go
aarzilli: I don't see how it would be useful.

pkg/proc/redirect/redirect.go
tttoad: done.

service/dap/server.go
tttoad: This can be used to compare the difference between the output of the client and the server.
Sometimes when a program is debugged remotely, some components need to ensure that the output is printed to stdout/stderr.
As far as I know, there are programs that pass information to other services via stdout/stderr.
In this case, it is necessary to print both.

aarzilli: This is only going to work if the remote instance of delve is started with the 'dap' command, which we do not recommend doing. Additionally it can create discrepancies because the program is not connected to the real pipe: you are not propagating a close from the debuggeee to delve's stdout/stderr and you aren't propagating the SIGPIPE in the other direction either.

tttoad:
@aarzilli PTAL.

service/dap/server.go
aarzilli: Still TBD.

pkg/proc/redirect/redirect.go
aarzilli: > make it describe a single redirect rather than three redirects in one

Still TBD.

aarzilli (commented):
`debugger.(*Debugger).Restart` needs to return an error if redirects it can't handle are set in the configuration.

service/dap/server.go
```diff
@@ -1135,9 +1272,11 @@ func (s *Session) onDisconnectRequest(request *dap.DisconnectRequest) {
 		status := "halted"
 		if s.isRunningCmd() {
 			status = "running"
-		} else if s, err := s.debugger.State(false); processExited(s, err) {
```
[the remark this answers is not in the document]
tttoad: > Undo unrelated changes.

This change is related. 
It needs to call s.wg.Wait(),the 's' cannot be overwritten.

pkg/proc/redirect/redirect.go
tttoad: Do you mean `WriterFiles [3]*os.File -> StdinWriter *os.File, StdoutWriter *os.File, StderrWriter *os.File`？

aarzilli: I think this should just be a string/io.Writer pair:

```
type OutputRedirect struct {
     Path string
     Writer io.Writer
}
```

tttoad: How about this?
```
type OutputRedirect struct {
	Path string
	File *os.File
}

Launch(.....,redirect [3]OutputRedirect) 
```

I can't use `io.Writer` or `io.WriteColser`, because in `gdbserial/rr.go` it needs `*os.File`.
In `service/service.go`, the data will be read from this structure.

pkg/proc/proc_notwindows_test.go
```diff
@@ -0,0 +1,80 @@
+//go:build !windows
```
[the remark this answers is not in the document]
tttoad: The reason for using build.tags is that the syscall.Mkfifo() does not compile on window systems.

pkg/proc/proc_test.go
```diff
@@ -5935,6 +5935,86 @@ func TestStacktraceExtlinkMac(t *testing.T) {
...
+		errChan                        = make(chan error, 2)
+		canceFunc        func()
+		needCheck        = false
+		stdoutExpectFile = "out_redirect-stdout.txt"
+		stderrExpectFile = "out_redirect-stderr.txt"
+	)
+	switch testBackend {
+	case "native":
+		if runtime.GOOS == "linux" {
+			redirect, canceFunc, err = testGenRedireByPath(t, fixture, stdoutExpectFile, stderrExpectFile, errChan)
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
```diff
@@ -157,8 +157,29 @@ type Session struct {
...
+	outputMode outputMode
+
+	// stdoutReader the programs's stdout.
+	stdoutReader io.ReadCloser
+
+	// stderrReader the program's stderr.
+	stderrReader io.ReadCloser
+
+	// wg the WaitGroup that needs to wait before sending a terminated event.
+	wg sync.WaitGroup
```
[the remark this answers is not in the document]
tttoad: done.

service/dap/server.go
tttoad: done.

pkg/proc/redirect/redirect.go
aarzilli: Ok.

pkg/proc/proc_notwindows_test.go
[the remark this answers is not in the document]
aarzilli: I see. I still think this is wrong. Either this test shouldn't be here or the code that creates the redirect should also be in pkg/proc. I'd lean towards deleting this test.

service/dap/server.go
```diff
@@ -1023,9 +1044,58 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+			if err != nil {
+				if errors.Is(io.EOF, err) {
+					return
+				}
+				s.config.log.Errorf("failed read by %s - %v ", category, err)
+				return
+			}
+			outs := string(out[:n])
+
+			if s.outputMode&outputToStd != 0 {
```
aarzilli, line 1081: This can never happen now.

service/dap/server.go
```diff
@@ -1037,16 +1107,65 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+	if redirected {
+		redirects, err := NewRedirector()
+		if err != nil {
+			s.sendShowUserErrorResponse(request.Request, InternalError, "Internal Error",
+				fmt.Sprintf("failed to generate stdio pipes - %v", err))
+			return
+		}
+
+		s.config.Debugger.Redirect = redirects
+		s.preTerminatedWG.Add(1)
```
aarzilli, line 1144: Don't you think it's weird that the nodebug code and the debug code here do more or less the same thing, but it's not actually the same code, it's a slightly different copy instead?

service/dap/server_test.go
```diff
@@ -7367,3 +7368,101 @@ func TestDisassembleCgo(t *testing.T) {
...
+				switch m.Body.Category {
+				case "stdout":
+					stdout.WriteString(m.Body.Output)
+				case "stderr":
+					stderr.WriteString(m.Body.Output)
+				default:
+					t.Errorf("\ngot %#v\nwant Category='stdout' or 'stderr'", m)
+				}
+			case *dap.TerminatedEvent:
+				hasNext = false
```
aarzilli, line 7423: Use break with a label on the for loop.

service/dap/server.go
tttoad: yes.I will remove the part related to `outputMode` that has `redirected` enough.

service/dap/server_test.go
tttoad: It is inside the `switch` block, and using `break` will only break out of `switch`.

pkg/proc/redirect/redirect.go
tttoad: @aarzilli  Need a reply.

service/dap/server_test.go
aarzilli: https://go.dev/play/p/pmToDaR4fq0

pkg/proc/redirect/redirect.go
aarzilli: What you suggested is ok.

tttoad:
@aarzilli  PTAL.

service/dap/server_test.go
tttoad: done.

service/dap/server.go
tttoad: done.

pkg/proc/proc_notwindows_test.go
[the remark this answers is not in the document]
tttoad: done.

pkg/proc/gdbserial/gdbserver.go
```diff
@@ -450,7 +450,7 @@ func getLdEnvVars() []string {
 // LLDBLaunch starts an instance of lldb-server and connects to it, asking
 // it to launch the specified target program with the specified arguments
 // (cmd) on the specified directory wd.
-func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs []string, tty string, redirects [3]string) (*proc.TargetGroup, error) {
+func LLDBLaunch(cmd []string, wd string, flags proc.LaunchFlags, debugInfoDirs []string, tty string, redirects [3]proc.OutputRedirect) (*proc.TargetGroup, error) {
```
aarzilli, line 453: Can this ever be anything but three strings? I'm guessing no, since this does not do anything if the File field is set. I think the type here should stay `[3]string` and the debugger layer should handle the conversion.

pkg/proc/gdbserial/rr.go
```diff
@@ -20,7 +20,7 @@ import (
 // program. Returns a run function which will actually record the program, a
 // stop function which will prematurely terminate the recording of the
 // program.
-func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]string) (run func() (string, error), stop func() error, err error) {
+func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]proc.OutputRedirect) (run func() (string, error), stop func() error, err error) {
```
aarzilli, line 23: It's weird that the stdin redirect is of type OutputRedirect, and that it can never be anything but a path. I'd say this should be changed to `stdinRedirect string, stdoutRedirect, stderrRedirect proc.OutputRedirect`. The stdin redirect isn't treated like the other two at all, at this point.

pkg/proc/gdbserial/rr.go
```diff
@@ -63,32 +63,42 @@ func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]string) (run
...
-	if redirects[0] != "" {
-		stdin, err = os.Open(redirects[0])
+	if redirects[0].Path != "" {
+		stdin, err = os.Open(redirects[0].Path)
 		if err != nil {
 			return nil, nil, nil, nil, err
 		}
 		toclose = append(toclose, stdin)
-	} else {
+	} else if quiet {
```
aarzilli, line 77: Can you explain this change?

pkg/proc/native/proc.go
```diff
@@ -370,29 +370,38 @@ func (dbp *nativeProcess) writeSoftwareBreakpoint(thread *nativeThread, addr uin
...
-	if redirects[0] != "" {
-		stdin, err = os.Open(redirects[0])
+	if redirects[0].Path != "" {
+		stdin, err = os.Open(redirects[0].Path)
 		if err != nil {
 			return nil, nil, nil, nil, err
 		}
 		toclose = append(toclose, stdin)
-	} else if foreground {
+	} else {
```
aarzilli, line 384: Same as the other, can you explain this change? This in particular just looks wrong. We only want to pass stdin to the target process if we are foregrounding it.

pkg/proc/native/proc.go
```diff
@@ -370,29 +370,38 @@ func (dbp *nativeProcess) writeSoftwareBreakpoint(thread *nativeThread, addr uin
...
+			}
+
+			return f
 		}
-		var f *os.File
-		f, err = os.Create(path)
-		if f != nil {
-			toclose = append(toclose, f)
+
+		if redirect.File != nil {
```
aarzilli, line 398: This should be `else if`.

pkg/proc/redirect.go
```diff
@@ -0,0 +1,27 @@
...
+import "os"
+
+type Mode string
+
+type OutputRedirect struct {
+	Path string
+	File *os.File
+}
+
+const (
```
aarzilli, line 12: These constants aren't used anywhere.

aarzilli (commented):
See also test failures on linux.

service/dap/server_test.go
```diff
@@ -7367,3 +7368,104 @@ func TestDisassembleCgo(t *testing.T) {
...
+				}
+			case *dap.TerminatedEvent:
+				break TerminnatedPoint
+			default:
+				t.Errorf("\n got %#v, want *dap.OutputEvent or *dap.TerminateResponse", m)
+			}
+		}
+
+		var (
+			stdoutFilePath = filepath.Join(fixture.BuildDir, "out_redirect-stdout.txt")
```
aarzilli, line 7432: Instead of reading files from disk just make them literal strings.

pkg/proc/stdio_pipe.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+type stdioRedirector struct {
+	Paths [3]string
+}
+
+func NewRedirector() (redirect *stdioRedirector, err error) {
```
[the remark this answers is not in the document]
tttoad: `gdbserver.go` needs a file path instead of `os.File`.

pkg/proc/gdbserial/rr.go
tttoad: Sorry for the wrong change here. I always thought I had fixed this logic....

pkg/proc/gdbserial/rr.go
tttoad: Maybe the struct name should be changed back to `Redirect`.I want to keep `stdin` passed in as `*os.File`.

pkg/proc/stdio_pipe.go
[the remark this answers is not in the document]
aarzilli: And os.Pipe doesn't create named pipes? Ok. I'd say we should make a function like this:

```
func NamedPipe() (proc.OutputRedirect, err error)`
```

call the files namedpipe_windows.go and namedpipe_other.go

pkg/proc/gdbserial/rr.go
aarzilli: Why? Nothing is using the *os.File version of stdin. Looking at the DAP specification I don't see anything that could use it, but maybe I'm missing it.

pkg/proc/stdio_pipe.go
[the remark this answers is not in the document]
tttoad: `os.Pipe()` should create named pipes， but its return value is `*os.File`.It's harder to get the file path from `*os.File`.Different systems may be placed in different locations.
I don't think it's a good idea to return the `proc.OutputRedirect`.
I want to implement how to get `reader` and `writer` in a method of a structure.
Maybe we just need to change NewRedirector()->NamedPipes(), filename: stdio_pipe->namedpipe_other.go,stdio_pipe_windows.go->namedpipe_windows.go.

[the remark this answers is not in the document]
aarzilli: `func NamedPipe() (io.Reader, proc.OutputRedirect, error)`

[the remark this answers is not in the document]
aarzilli: I don't want a method that does the same thing twice, like NewRedirector does.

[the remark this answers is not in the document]
tttoad: The Get `io.Reader` method needs to be executed asynchronously. 
`proc.OutputRedirect` needs to be passed synchronously to `debugger.config`.
`os.OpenFile(s.Paths[1], os.O_RDONLY, os.ModeNamedPipe)` will be blocked until the file is opened as written.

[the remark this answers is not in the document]
tttoad: `The Get io.Reader method needs to be executed asynchronously.`
This has a bug.  If the `debugger` does not start successfully, the `goroutine` will keep blocking on `Reader()`.I need some time to think about how to solve it gracefully.

tttoad:
@aarzilli PTAL. 
The "TeamCity build failed" error doesn't seem to be caused by my changes?
image

https://github.com/go-delve/delve/compare/31dba49..7d441c2

Here it just changes the function name, but it goes from pass to fail.

aarzilli:
The windows/arm64 build is flaky, it doesn't have anything to do with you. The rest passed. If you click on "details" you'll see what failed.

service/dap/server_test.go
tttoad: done.

pkg/proc/redirect.go
tttoad: done.

pkg/proc/native/proc.go
tttoad: done.

pkg/proc/native/proc.go
tttoad: done.

pkg/proc/gdbserial/rr.go
tttoad: done.

pkg/proc/gdbserial/rr.go
tttoad: I went through the `Dap` specification and didn't find definition of remote input event. Maybe there will be a specification to support remote input events.

pkg/proc/gdbserial/rr.go
aarzilli: You changed this line from `} else {` to `} else if quiet{`, why?

pkg/proc/gdbserial/rr.go
```diff
@@ -63,32 +63,40 @@ func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]string) (run
...
+	create := func(redirect proc.OutputRedirect, dflt *os.File) (f *os.File) {
+		if redirect.Path != "" {
+			f, err = os.Create(redirect.Path)
+			if f != nil {
+				toclose = append(toclose, f)
 			}
-			return dflt
+
+			return f
+		}else if redirect.File != nil {
```
aarzilli, line 89: Run `go fmt` when you are done.

tttoad: > Run `go fmt` when you are done.

I'm so sorry.
This file was modified locally and then submitted without saving...
There is something wrong with the autosave of the IDE.

tttoad: done.

tttoad:
@aarzilli PTAL.

pkg/proc/stdio_pipe.go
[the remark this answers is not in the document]
aarzilli: I still believe what I said in this thread.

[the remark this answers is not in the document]
tttoad: You mean change it to this : `func NamedPipe() (io.Reader, proc.OutputRedirect, error)`?
To do this, I need to start goroutine and open the pipe in ` os.O_WDONLY` mode inside method `NamedPipe`. (Normally the `gdb` will open the `pipe`. 
If you agree with this way, I can modify it.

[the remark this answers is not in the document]
aarzilli: Why do you need to start a goroutine in NamedPipe?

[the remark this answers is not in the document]
tttoad: The `os.OpenFile(s.Paths[1], os.O_RDONLY, os.ModeNamedPipe)` block the current `goroutine`. So, I need to use `goroutine` to open this file with `os.O_WDONLY`.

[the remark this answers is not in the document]
aarzilli: Return a thing that opens on first read:

```
type openOnRead struct {
	p string
	rd io.Reader
}

func (oor *openOnRead) Read(p []byte) (n int, err error) {
	if oor.rd != nil {
		return oor.rd.Read(p)
	}
	fh, err := os.OpenFile(oor.p, ...)
	if err != nil {
		return 0, err
	}
	oor.rd = fh
	return oor.rd.Read(p)
}
```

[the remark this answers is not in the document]
tttoad: done.

pkg/proc/namedpipe_other.go
```diff
@@ -0,0 +1,84 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return oor.rd.Close()
+}
+
+func NamedPipe() (reader [2]io.ReadCloser, output [3]OutputRedirect, err error) {
```
aarzilli, line 59: It should be `NamedPipe() (io.ReadClsoer, OutputRedirect, error)`. It's called `NamedPipe` not `TwoNamedPipes`

pkg/proc/proc_test.go
```diff
@@ -5926,6 +5927,120 @@ func TestStacktraceExtlinkMac(t *testing.T) {
 	})
 }
 
+func testGenRedirect(t *testing.T, fixture protest.Fixture, expectStdout string, expectStderr string, errChan chan error) (redirect [3]proc.OutputRedirect, cancelFunc func(), err error) {
```
aarzilli, line 5930: IMO this functionality is adequately tested throught the dap layer and there is no need for this test here.

service/dap/server_test.go
```diff
@@ -7367,3 +7368,94 @@ func TestDisassembleCgo(t *testing.T) {
...
+			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=5", cdResp)
+		}
+
+		// 6 << output, << terminated
+		var (
+			stdout = bytes.NewBufferString("")
+			stderr = bytes.NewBufferString("")
+		)
+
+	TerminnatedPoint:
```
aarzilli, line 7411: terminatedPoint

pkg/proc/namedpipe_other.go
```diff
@@ -0,0 +1,84 @@
...
+
+	oor.rd = fh
+	return oor.rd.Read(p)
+}
+
+func (oor *openOnRead) Close() error {
+	defer os.Remove(oor.p)
+
+	// try from "init" to "close".
+	if !atomic.CompareAndSwapInt32(&oor.status, 0, 2) {
```
aarzilli, line 45: This is awful, there has to be a better way. At minimum it should use a mutex instead.

aarzilli: I think it should be enough to start the readers only after the target process has started.

aarzilli: This can not work because rr.go (at minimum) would deadlock. However all we need to know is that the fifo has been opened for writing at least once, which we can do by unconditionally opening the fifo with `os.O_WRONLY|syscall.O_NONBLOCK`. This should work.

tttoad: > 
`os.O_WRONLY|syscall.O_NONBLOCK`  can solve the blocking problem. But, until the pipe is opened in write mode, the read pipe will return null.

https://go.dev/play/p/hKYuUNNg868

aarzilli: I meant this:

```
func (oor *openOnRead) Close() error {
	defer os.Remove(oor.p)
	fh, _ := os.OpenFile(oor.p, os.O_WRONLY|syscall.O_NONBLOCK, 0)
	if fh != nil {
		fh.Close()
	}
	return oor.rd.Close()
}
```

tttoad: done.

service/dap/server_test.go
tttoad: done.

pkg/proc/proc_test.go
tttoad: done.

pkg/proc/namedpipe_other.go
tttoad: done.

service/dap/server_test.go
```diff
@@ -7367,3 +7368,94 @@ func TestDisassembleCgo(t *testing.T) {
...
+			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=5", cdResp)
+		}
+
+		// 6 << output, << terminated
+		var (
+			stdout = bytes.NewBufferString("")
+			stderr = bytes.NewBufferString("")
+		)
+
+	terminnatedPoint:
```
aarzilli, line 7411: terminatedPoint

pkg/proc/gdbserial/rr.go
```diff
@@ -63,11 +63,13 @@ func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]string) (run
 	return run, stop, nil
 }
 
-func openRedirects(redirects [3]string, quiet bool) (stdin, stdout, stderr *os.File, closefn func(), err error) {
-	toclose := []*os.File{}
+func openRedirects(stdinPath string, stdoutOR proc.OutputRedirect, stderrOR proc.OutputRedirect, quiet bool) (stdin, stdout, stderr *os.File, closefn func(), err error) {
```
aarzilli, line 66: just use `stdin, stdout, stderr`

aarzilli (commented):
Just a couple of minor changes, otherwise it's OK.

Also, in `(*Debugger).Restart` after the canRestart check but before the call to `detach` add this:

```
if !resetArgs && (d.config.Stdout.File != nil || d.config.Stderr.File != nil) {
	return nil, ErrCanNotRestart
	
}
```

pkg/proc/gdbserial/rr.go
tttoad: `stdin, stdout, stderr` is in conflict with the returned parameter name.

service/dap/server_test.go
tttoad: done.

aarzilli (approved):
LGTM. Thanks.

pkg/proc/namedpipe_other.go
```diff
@@ -0,0 +1,59 @@
...
+
+	fh, _ := os.OpenFile(oor.path, os.O_WRONLY|syscall.O_NONBLOCK, 0)
+	if fh != nil {
+		fh.Close()
+	}
+
+	return oor.rd.Close()
+}
+
+func NamedPipe() (reader io.ReadCloser, output OutputRedirect, err error) {
```
derekparker, line 45: What's the purpose of this file and the Windows variant, why not simply use `os.Pipe` across the board?

derekparker: Also, calling this a "named pipe" seems disingenuous as the caller cannot provide a name, so it's essentially the same as calling `os.Pipe` but with wrapped return values.

pkg/proc/redirect.go
```diff
@@ -0,0 +1,8 @@
+package proc
+
+import "os"
+
+type OutputRedirect struct {
```
derekparker, line 5: Needs doc comment.

derekparker (changes requested):
Sorry for the review delay, just got back from some time off.

pkg/proc/namedpipe_other.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	oor.rd = fh
+	return oor.rd.Read(p)
+}
+
+func (oor *openOnRead) Close() error {
+	defer os.Remove(oor.path)
+
+	fh, _ := os.OpenFile(oor.path, os.O_WRONLY|syscall.O_NONBLOCK, 0)
```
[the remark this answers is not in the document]
tttoad: [service/dap/server.go#L1138](https://github.com/go-delve/delve/blob/a4f0491d4b2570b2e0f87b0472e86498e3c1f3ef/service/dap/server.go#L1138)
When the `debugger.New()` function is wrong, `oor.path` may never be opened.The `os.OpenFile(oor.path, os.O_RDONLY, os.ModeNamedPipe)`  will be blocked.

pkg/proc/namedpipe_other.go
tttoad: [pkg/proc/gdbserial/gdbserver.go](https://github.com/go-delve/delve/blob/a4f0491d4b2570b2e0f87b0472e86498e3c1f3ef/pkg/proc/gdbserial/gdbserver.go#L490)
The file path needs to be used in the `gdbserver.LLDBLaunch()`. I can't get the file path from `os.Pipe`. 
It is called NamedPipe because it uses the Named Pipe mode. But this seems disingenuous in the `namedpipe_other` file. Maybe we can call it `Redirector`?

pkg/proc/namedpipe_other.go
```diff
@@ -0,0 +1,59 @@
+//go:build !windows
```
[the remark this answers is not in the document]
aarzilli: I think it only works for build tags, not as file suffixes: https://github.com/golang/go/issues/51572
Also go1.19 is too soon for us.

tttoad:
@derekparker PTAL.

pkg/proc/namedpipe_other.go
derekparker: Ah, ok that makes sense. Yes, I like the name `Redirector` better, I think.

tttoad: done.

pkg/proc/gdbserial/rr.go
```diff
@@ -63,11 +63,13 @@ func RecordAsync(cmd []string, wd string, quiet bool, redirects [3]string) (run
...
+	var (
```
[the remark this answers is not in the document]
tttoad: done.

pkg/proc/namedpipe_other.go
tttoad: @derekparker PTAL.

mcdoker18:
@tttoad Thanks for the implementing this feature!

@aarzilli @derekparker Can you please take a look again?

derekparker (approved):
LGTM

service/dap/server.go
```diff
@@ -1080,15 +1168,30 @@ func (s *Session) getPackageDir(pkg string) string {
...
+	if redirected {
+		if s.stderrReader, err = cmd.StderrPipe(); err != nil {
+			return nil, err
+		}
+
+		if s.stdoutReader, err = cmd.StdoutPipe(); err != nil {
+			return nil, err
+		}
+	} else {
+		cmd.Stdout, cmd.Stderr = os.Stdin, os.Stderr
```
hyangah, line 1188: should be
`cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr`

service/dap/server.go
```diff
@@ -1020,9 +1029,51 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
...
+		runReadFunc := func(reader io.ReadCloser, category string) {
+			defer s.preTerminatedWG.Done()
+			defer reader.Close()
+			// Read output from `reader` and send to client
+			var out [1024]byte
+			for {
+				n, err := reader.Read(out[:])
+				if err != nil {
+					if errors.Is(io.EOF, err) {
+						return
```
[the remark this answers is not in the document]
tttoad: You're right. Thanks for your guidance.

PR #3257, opened 2023-01-25
terminal/logflags: Added `SetLoggerFactory(LoggerFactory)`

blaubaer:
This change will enable people who want to embed Delve into their applications to adjust the logging better to their needs.

### DIFF
--- a/pkg/logflags/logflags.go
+++ b/pkg/logflags/logflags.go
@@ -29,23 +29,19 @@ var minidump = false
 
 var logOut io.WriteCloser
 
-func makeFlaggableLogger(flag bool, fields Fields) Logger {
-	if flag {
-		return makeLogger(logrus.DebugLevel, fields)
-	}
-	return makeLogger(logrus.ErrorLevel, fields)
-}
-
-func makeLogger(level logrus.Level, fields Fields) Logger {
+func makeLogger(flag bool, fields Fields) Logger {
 	if lf := loggerFactory; lf != nil {
-		return lf(level, fields, logOut)
+		return lf(flag, fields, logOut)
 	}
 	logger := logrus.New().WithFields(logrus.Fields(fields))
 	logger.Logger.Formatter = DefaultFormatter()
 	if logOut != nil {
 		logger.Logger.Out = logOut
 	}
-	logger.Logger.Level = level
+	logger.Logger.Level = logrus.ErrorLevel
+	if flag {
+		logger.Logger.Level = logrus.DebugLevel
+	}
 	return &logrusLogger{logger}
 }
 
@@ -62,7 +58,7 @@ func GdbWire() bool {
 
 // GdbWireLogger returns a configured logger for the gdbserial wire protocol.
 func GdbWireLogger() Logger {
-	return makeFlaggableLogger(gdbWire, Fields{"layer": "gdbconn"})
+	return makeLogger(gdbWire, Fields{"layer": "gdbconn"})
 }
 
 // Debugger returns true if the debugger package should log.
@@ -72,7 +68,7 @@ func Debugger() bool {
 
 // DebuggerLogger returns a logger for the debugger package.
 func DebuggerLogger() Logger {
-	return makeFlaggableLogger(debugger, Fields{"layer": "debugger"})
+	return makeLogger(debugger, Fields{"layer": "debugger"})
 }
 
 // LLDBServerOutput returns true if the output of the LLDB server should be
@@ -89,7 +85,7 @@ func DebugLineErrors() bool {
 
 // DebugLineLogger returns a logger for the dwarf/line package.
 func DebugLineLogger() Logger {
-	return makeFlaggableLogger(debugLineErrors, Fields{"layer": "dwarf-line"})
+	return makeLogger(debugLineErrors, Fields{"layer": "dwarf-line"})
 }
 
 // RPC returns true if RPC messages should be logged.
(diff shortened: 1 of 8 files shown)


blaubaer:
Does someone know why the test `TestClientServer_FindLocationsAddr` (in this run) is failing? I cannot see a reason why it should fail (caused by my changes) nor I can bring it to fail locally. Is this a temporary issue?

aarzilli:
> Does someone know why the test `TestClientServer_FindLocationsAddr` (in this run) is failing? I cannot see a reason why it should fail (caused by my changes) nor I can bring it to fail locally. Is this a temporary issue?

The windows/arm64 port has some problems.

pkg/logflags/logflags.go
```diff
@@ -217,12 +236,17 @@ func Close() {
 	}
 }
 
-// textFormatter is a simplified version of logrus.TextFormatter that
+// DefaultFormatter provides a simplified version of logrus.TextFormatter that
 // doesn't make logs unreadable when they are output to a text file or to a
 // terminal that doesn't support colors.
-type textFormatter struct {
+func DefaultFormatter() logrus.Formatter {
```
aarzilli, line 242: Why do we want to export this?

pkg/logflags/logflags.go
```diff
@@ -29,17 +29,24 @@ var minidump = false
 
 var logOut io.WriteCloser
 
-func makeLogger(flag bool, fields logrus.Fields) *logrus.Entry {
-	logger := logrus.New().WithFields(fields)
-	logger.Logger.Formatter = &textFormatter{}
+func makeFlaggableLogger(flag bool, fields Fields) Logger {
```
aarzilli, line 32: No reason to split this function in to two functions when you only call makeFlaggableLogger anyway.

pkg/logflags/logflags.go
```diff
@@ -80,14 +87,27 @@ func DebugLineErrors() bool {
...
-	return makeLogger(rpc, logrus.Fields{"layer": "rpc"})
+func RPCLogger() Logger {
+	if rpc {
+		return RPCLoggerWithLevel(logrus.DebugLevel)
+	}
+	return RPCLoggerWithLevel(logrus.ErrorLevel)
+}
+
+// RPCLoggerWithLevel returns a logger for RPC messages set to a specific minimal log level.
+func RPCLoggerWithLevel(level logrus.Level) Logger {
```
aarzilli, line 109: No reason to export this.

pkg/logflags/logflags.go
```diff
@@ -139,16 +159,15 @@ func writeListeningMessage(server string, addr net.Addr) {
...
-	logger := RPCLogger()
-	logger.Logger.Level = logrus.WarnLevel
+	logger := RPCLoggerWithLevel(logrus.WarnLevel)
 	logger.Warnln("Listening for remote connections (connections are not authenticated nor encrypted)")
 }
 
 func WriteError(msg string) {
 	if logOut != nil {
-		fmt.Fprintln(logOut, msg)
+		_, _ = fmt.Fprintln(logOut, msg)
```
aarzilli, line 168: We don't do this `_, _ = fmt.Fprintln` thing anywhere.

pkg/logflags/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	Errorln(args ...interface{})
+	Fatalln(args ...interface{})
+	Panicln(args ...interface{})
+}
+
+// LoggerFactory is used to create new logrus.FieldLogger instances.
+// SetLoggerFactory can be used to configure it.
+//
+// The given parameters fields and out can be both be nil.
+type LoggerFactory func(level logrus.Level, fields Fields, out io.Writer) Logger
```
aarzilli, line 50: This function shouldn't take a logrus.Level, it should take the bool flag that says if logging is enabled or not.

pkg/logflags/logflags.go
blaubaer: Ok.

pkg/logflags/logflags.go
blaubaer: Ok.

pkg/logflags/logflags.go
blaubaer: This is caused by [`pkg/logflags/logflags.go:163`](https://github.com/go-delve/delve/pull/3257/files/ccda6c241cf732d8ce033eb443f31beead8da3d7#diff-f89e32a213ab96ab1585632446e21aef2976688c4ac06c461d6ceb95bcc2903cR163):
```golang
logger := RPCLoggerWithLevel(logrus.WarnLevel)
logger.Warnln("Listening for remote connections (connections are not authenticated nor encrypted)")
```

In this case the initial code forced that this Warning should always be displayed, although if tagged or not. With this I tried to preserve this behavior.

aarzilli: We can just pass true as the flag.

blaubaer: Ok.

pkg/logflags/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	Infoln(args ...interface{})
+	Println(args ...interface{})
+	Warnln(args ...interface{})
+	Warningln(args ...interface{})
+	Errorln(args ...interface{})
+	Fatalln(args ...interface{})
+	Panicln(args ...interface{})
+}
+
+// LoggerFactory is used to create new logrus.FieldLogger instances.
```
aarzilli, line 46: s/new logrus.FieldLogger instances/new Logger instances/

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3258, opened 2023-01-27
godwarf: Allow extracting a DWARF entry's field

brancz:
Previously it was only possible to extract a value of type `any` using an attribute name. This poses challenges when fields are allowed to have different classes, and it is ambiguous how to handle them.

I for example am seeing a case where a subroutine has a `DW_AT_name` that is not of class string, but string alt, therefore requiring a lookup in the `.debug_str` table. (https://dwarfstd.org/ShowIssue.php?issue=120604.1)

Being able to extract the whole field is better as we can decide what to do based on the class of the field and not guess based on the data type in Go.

### DIFF
--- a/pkg/dwarf/godwarf/tree.go
+++ b/pkg/dwarf/godwarf/tree.go
@@ -11,14 +11,22 @@ import (
 // entry specified by DW_AT_abstract_origin will be searched recursively.
 type Entry interface {
 	Val(dwarf.Attr) interface{}
+	AttrField(dwarf.Attr) *dwarf.Field
 }
 
 type compositeEntry []*dwarf.Entry
 
 func (ce compositeEntry) Val(attr dwarf.Attr) interface{} {
+	if f := ce.AttrField(attr); f != nil {
+		return f.Val
+	}
+	return nil
+}
+
+func (ce compositeEntry) AttrField(a dwarf.Attr) *dwarf.Field {
 	for _, e := range ce {
-		if r := e.Val(attr); r != nil {
-			return r
+		if f := e.AttrField(a); f != nil {
+			return f
 		}
 	}
 	return nil
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -1248,9 +1248,17 @@ func debugCallProtocolReg(archName string, version int) (uint64, bool) {
 	}
 }
 
-type fakeEntry map[dwarf.Attr]interface{}
+type fakeEntry map[dwarf.Attr]*dwarf.Field
 
 func (e fakeEntry) Val(attr dwarf.Attr) interface{} {
+	if e[attr] == nil {
+		return nil
+	}
+
+	return e[attr].Val
+}
+
+func (e fakeEntry) AttrField(attr dwarf.Attr) *dwarf.Field {
 	return e[attr]
 }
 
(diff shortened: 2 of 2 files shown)


aarzilli:
> I for example am seeing a case where a subroutine has a DW_AT_name that is not of class string, but string alt, therefore requiring a lookup in the .debug_str table. (https://dwarfstd.org/ShowIssue.php?issue=120604.1)

In theory debug/dwarf should automatically extract strings from debug_str, see `debug/dwarf.(*buf).entry` the switch cases for formStrp, formStrx, &c.
I'm not saying we shouldn't do this, but if there is a problem with debug/dwarf it should be fixed there.

Also this breaks compilation of delve itself because of the fakeEntry type in pkg/proc/fncall.go.

brancz:
I did some more research and I now fully understand everything all the pieces. tl;dr we do need this, and here follows a detailed explanation why.

It turns out returning the offset is actually the best thing we can do in this case, as it is a reference to a string in a different file, which we cannot guarantee to be present. Just for the sake of explaining everything end to end: when there is an attribute of form `DW_FORM_GNU_strp_alt ` it means we need to read the `.gnu_debugaltlink` section, which is generated by the `dwz` tool (which is used on all Fedora debuginfos). In short, what `dwz` does is deduplicate as many things as it can to reduce debuginfo size, including in this case, the string of this particular `DW_AT_name`.

An example for postgres, which has the build ID `fc8e3f5bdf24747246833b88be016056e1c742c4`:

```
curl https://debuginfod.fedoraproject.org/buildid/fc8e3f5bdf24747246833b88be016056e1c742c4/debuginfo > postgres-debuginfo
```

We read the `.gnu_debugaltlink` section, which contains a possible file location (this file is meant for the case when the debuginfo is obtained via a debug package, and the last 20 bytes are an identifier):

```
$ readelf --hex-dump=.gnu_debugaltlink postgres-debuginfo

Hex dump of section '.gnu_debugaltlink':
  0x00000000 2e2e2f2e 2e2f2e64 777a2f70 6f737467 ../../.dwz/postg
  0x00000010 72657371 6c2d3134 2e332d32 2e666333 resql-14.3-2.fc3
  0x00000020 362e7838 365f3634 00eecc2e 3b97bfde 6.x86_64....;...
  0x00000030 0244ea93 c67cc39f 324069f1 7f       .D...|..2@i..
```

And here's the last part: The identifier is also a Build ID which in turn we can again use to fetch from debuginfod:

```
curl https://debuginfod.fedoraproject.org/buildid/eecc2e3b97bfde0244ea93c67cc39f324069f17f/debuginfo > postgres-supplemental
```

Now the next part is unfortunately a little trickier to do, because as the Fedora community states, unfortunately `dwarfdump` does not support ``, which means we need to write some Go code to do the next thing:

```

package main

import (
	"debug/dwarf"
	"debug/elf"
	"fmt"
	"log"
	"os"
)

func main() {
	f, err := elf.Open(os.Args[1])
	if err != nil {
		log.Fatal(err)
	}

	d, err := f.DWARF()
	if err != nil {
		log.Fatal(err)
	}

	r := d.Reader()
	r.Seek(0x00a90dca)
	dwarfEntry, err := r.Next()
	if err != nil {
		log.Fatal(err)
	}

	dwarfField := dwarfEntry.AttrField(dwarf.AttrName)
	fmt.Printf("%#+v\n", dwarfEntry)
	fmt.Printf("%#+v\n", uint64(dwarfField.Val.(int64)))
}
```

Running this against the postgres debuginfos returns the offset `0x467f8`. Now we can finally take that offset and read the string from the supplemental file:

```
$ readelf --string-dump=.debug_str postgres-supplemental | grep 467f8
  [ 467f8]  brin_new_memtuple
```

So yeah, since we can't rely on the supplemental file, the best we can do is return the offset.

brancz:
One more example, since what I brought up so far would already result in `string` and `int64`, there is possible ambiguity when it's an `int64`, as an attribute of form `DW_FORM_ref_sup` can also be an `int64`, but it needs to be handled differently to a `DW_FORM_GNU_strp_alt`.

tl;dr we need to be able to know what an attribute's class is to appropriately handle it as there are ambiguous cases that have the same (Go) datatype.

derekparker:
Overall adding `AttrField` to `Entry` is fine with me, but as @aarzilli mentioned this breaks the build and `fakeEntry` must be updated as well.

The `fakeEntry` type is defined [here](https://github.com/go-delve/delve/blob/master/pkg/proc/fncall.go#L1251). You'll have to change the definition to `map[dwarf.Attr]*dwarf.Field` and then upon creation, create `dwarf.Field` structs where `Val` contains what would have been the value for that key in the map. The `Attr` and `Class` fields should be included as well for consistency.

brancz:
Thank you for bringing that to my attention again, I totally missed that. Should be fixed now (CI is also green now). Let me know if there is anything else!

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3271, opened 2023-02-12
*: remove some code for Go 1.12 or earlier

aarzilli:
Delve no longer compiles on Go1.12 and earlier, we don't test it on
these versions and they are 4 years old and unsupported. Remove some
code related to Go 1.12 and earlier, mostly from tests.


derekparker (approved):
LGTM

PR #3273, opened 2023-02-14
proc/native: fix syscalls to SYS_PROCESS_VM_READV (and WRITEV)

aarzilli:
Per https://pkg.go.dev/unsafe#Pointer conversions from unsafe.Pointer
to uintptr are only safe in limited circumstances. In particular only
conversions made in the syscall call are pinned.

Additionally add a call to runtime.KeepAlive to mitigate the bug
described in: https://github.com/golang/go/issues/58351


derekparker (approved):
LGTM

PR #3277, opened 2023-02-15
dwarf/line: handle end_seq properly

derekparker:
This patch changes how we handle end_seq in the debug_line state machine program. Instead of always considering the state machine invalid at the end_seq instead simply consider the *current* address invalid. This solves a number of issues such as incorrect disassemble output for the last few instructions in certain functions, and creating an eval scope at an address within the last few instructions of certain functions. It also handles the case where the end_seq address is the same as the start address of the next adjacent function, which would previously confuse Delve which is why we initially marked end_seq as invalid for the entire state machine. This approach is more nuanced and still solves that initial problem while fixing some problems introduced by that patch.


derekparker:
I haven't added a new test for this but instead no longer pass `localsNoDeclLineCheck` which was added as a workaround previously. The test added during that fix is sufficient to cover this with that change removed.

aarzilli (approved):

PR #3281, opened 2023-02-18
Add flag to replay for rr onprocess pid

jcpowermac:
Add `-p` or `rr-onprocess-pid` to replay
command to pass `-p` or `--onprocess` pid
to `rr`.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -91,6 +91,8 @@ var (
 
 	conf        *config.Config
 	loadConfErr error
+
+	rrOnProcessPid int
 )
 
 const dlvCommandLongDesc = `Delve is a source level debugger for Go programs.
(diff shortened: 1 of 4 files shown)


jcpowermac:
I was unsure about submitting this PR. If there is an alternative way to do this please let me know.

When using rr with openshift installer I couldn't find a way to debug the terraform providers which is forked from terraform and terraform is forked from the installer.

cmd/dlv/cmds/commands.go
```diff
@@ -368,6 +370,10 @@ https://github.com/mozilla/rr
 				os.Exit(execute(0, []string{}, conf, args[0], debugger.ExecutingOther, args, buildFlags))
 			},
 		}
+
+		replayCommand.Flags().IntVarP(&rrOnProcessPid, "rr-onprocess-pid", "p", 0,
```
aarzilli, line 374: I think we should just rename the long version of this to `--onprocess`.

aarzilli (commented):
The documentation needs to be regenerated by executing: `go run _scripts/gen-usage-docs.go`

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3284, opened 2023-02-21
pkg/proc: be resilient to the upcoming sys.PtraceLwpInfo changes

dmgk:
As part of golang/go#58387 work (CL [469835](https://go-review.googlesource.com/c/sys/+/469835)), the signature of sys.PtraceLwpInfo will change to accept *sys.PtraceLwpInfoStruct instead of an uintptr. As a compatibility measure, update ptraceGetLwpInfo to work with both old and new sys.PtraceLwpInfo signatures.


aarzilli:
I don't think this is necessary. Delve links to a specific version of x/sys and it only needs to support that: when we update x/sys we will change that function. I don't want to update it now because it will bump up the minimum version of Go that can build Delve by *a lot* of versions.

dmgk:
That makes sense, please feel free to close this PR.

aarzilli:
Closing. Thank you anyways.

PR #3285, opened 2023-02-22
Default to stop test_windows.ps1 when there is an error

qmuntal:
There are several bootstrapping situations that can leave Windows builders in a bad state.

For example, the windows/arm64 builder has been failing since last week because it failed to download the new go1.20.1 version and left a directory which marks it as installed, making following runs to fail because `go` could not be found.

This PR updates the Windows testing script to fix that.

### DIFF
--- a/_scripts/test_windows.ps1
+++ b/_scripts/test_windows.ps1
@@ -9,7 +9,7 @@ if ($binDir -eq "") {
 }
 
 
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3287, opened 2023-02-23
proc: step breakpoints shouldn't hide normal breakpoints

aarzilli:
When using Step on a function that has a dynamic CALL instruction we
set a Step breakpoint on the call.

When it is hit we determine the destination of the CALL by looking at
registers, set a breakpoint there and continue.

If the Step breakpoint is hit simultaneously with a normal breakpoint
our Step logic will take precedence and the normal breakpoint hit will
be hidden from the user.

Move the Step logic to a breaklet callback so that it does not
interfere with the decision to stop.

### DIFF
--- a/pkg/proc/target_exec.go
+++ b/pkg/proc/target_exec.go
@@ -173,22 +173,7 @@ func (grp *TargetGroup) Continue() error {
 				if err := conditionErrors(grp); err != nil {
 					return err
 				}
-				if grp.GetDirection() == Forward {
-					text, err := disassembleCurrentInstruction(dbp, curthread, 0)
-					if err != nil {
-						return err
-					}
-					var fn *Function
-					if loc, _ := curthread.Location(); loc != nil {
-						fn = loc.Fn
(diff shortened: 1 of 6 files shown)


pkg/proc/target_exec.go
```diff
@@ -797,14 +782,47 @@ func setStepIntoBreakpoints(dbp *Target, curfn *Function, text []AsmInstruction,
...
 		}
 	}
 	return nil
 }
 
+// stepIntoCallback is a callback called when a StepBreakpoint is hit, it
+// disassembles the current instruction to figure out its destination and
+// sets a breakpoint on it.
+func stepIntoCallback(curthread Thread, grp *TargetGroup, p *Target) (bool, error) {
+	if grp.GetDirection() != Forward {
```
derekparker, line 800: Couldn't we use `p.recman.GetDirection()` here instead of passing the group in?

derekparker (approved):
LGTM

PR #3293, opened 2023-03-01
proc: fuzzing expression evaluator and variable loader

aarzilli:
Add code to fuzz the expression evaluator and variable loader, see
comment to FuzzEvalExpression for how to use this.


derekparker (approved):
LGTM

PR #3294, opened 2023-03-01
proc: limit maximum time.Time we try to format

aarzilli:
The loop adding maxAddSeconds to format a time.Time can take multiple
seconds to complete if the time is very far into the future. To avoid
this loop slowing down debugging too much limit it to an arbitrary
maximum.

The chosen maximum is 1000 times the maximum expressible time.Duration,
which is 262 years. This means that we will not format dates beyond
year 262000 AD.

too much time
base address is 0
'fuzzingexpr' on '0a7b051f'.


derekparker (approved):
LGTM

PR #3300, opened 2023-03-07
proc: fix automatic breakpoints visibility

aarzilli:
Unrecovered-panic and fatal-throw were no longer part of the breakpoint
list because starting in 37e44bf they were created before the logical
breakpoints map was switched to the logical breakpoints map of the
target group.


derekparker (approved):
LGTM

PR #3303, opened 2023-03-15
Documentation: expand project layout to match usage

crisman:
The tree view of the project layout seemed to show that "github.com", "cmd", and "pkg" were all in the same directory.  The usage later in the document (and normal go layout) would have them as subdirectories.

The new tree output was generated with the below:
(cd `mktemp -d`
    mkdir -p github.com/me/foo/{cmd/foo,pkg/baz}
    touch github.com/me/foo/{cmd/foo/main,pkg/baz/bar{_test,}}.go
    tree -n --noreport github.com
)

### DIFF
--- a/Documentation/cli/getting_started.md
+++ b/Documentation/cli/getting_started.md
@@ -14,15 +14,14 @@ otherwise it optionally accepts a package path.
 For example given this project layout:
 
 ```
-.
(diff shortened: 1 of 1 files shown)


crisman:
Missed a typo in commit, closing.

Issue #3310, opened 2023-03-20
function call injection has problems when the function call involves passing variables stored in registers


Codelax:
1. What version of Delve are you using (`dlv version`)?
`dlv 1.20.1`
2. What version of Go are you using? (`go version`)?
`go 1.20.1`
3. What operating system and processor architecture are you using?
`Linux fedora 6.1.18-200.fc37.x86_64`
4. What did you do?

I debugged my program and when trying a `call` to get some info, delve took all my ram and my computer crashed.
When limiting with `ulimit` I get `fatal error: out of memory` ([full stacktrace](https://gist.github.com/Codelax/7bbdc1aba1164ccca4722e487016d5b4))

I reproduced the issue with a minimal go program
```go
func reflectFunc(value reflect.Value) {
	fmt.Printf("%s\n", value.Type().Name())
}

func main() {
	i := 2
	val := reflect.ValueOf(i)
	reflectFunc(val)
}
```

I just put a breakpoint on `main.reflectFunc`, did a single step.

```
     4:		"fmt"
     5:		"reflect"
     6:	)
     7:	
     8:	func reflectFunc(value reflect.Value) {
=>   9:		fmt.Printf("%s\n", value.Type().Name())
    10:	}
    11:	
    12:	func main() {
    13:		i := 2
    14:		val := reflect.ValueOf(i)
```

Then I used `call value.Type()`

5. What did you expect to see?
I expected an error or my `call` to work

6. What did you see instead?

My computer crashed :'(

PR #3311, opened 2023-03-24
proc: fix escapeCheck infinite recursion if something can not be

aarzilli:
deref'd

Fix infinite recursion if escapeCheck, at some point during its
recursion, creates an unreadable variable.

The deeper reason for this is that we evaluate function calls in a very
weird order so that we can always have stack space to store
intermediate evaluation results.

The variable 'value' happens to be stored in a register when we try to
make the call and because of our weird evaluation strategy registers
are no longer available to us when we evaluate 'value'.

This is not a complete fix for the issue, the real fix would be to
evaluate everything in its natural order, storing intermediate values
in Delve's memory instead of the target's stack. To do this we need a
mechanism to pin heap allocated objects, which at the moment does not
exist.

Updates #3310

### DIFF
--- a/_fixtures/reflecttypefncall.go
+++ b/_fixtures/reflecttypefncall.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+	"reflect"
+)
+
+func reflectFunc(value reflect.Value) {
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM

PR #3329, opened 2023-04-08
debugger: do not panic if defer location can not be found

aarzilli:
convertDefers should not try to dereference nil if the deferred
location does not belong to a function.


aarzilli:
Saw this while debugging but I don't have a self contained way to reproduce it.

derekparker (approved):
LGTM

Issue #3332, opened 2023-04-13
document that "config set substitute-path" supports empty string substitution


oakad:
Follow up to  #3088

The only way to supply per-project settings to `dlv` appears to be `--init` option. This option takes a file containing cli commands.

In some cases, it is not possible to run the `dlv` from the project working directory, despite what  some people may think ([#3088#discussion_r938455671](https://github.com/go-delve/delve/pull/3088#discussion_r938455671)).

Therefor there either must be a way to set  an "empty" path substitution via the interactive command (init file, actually) or an additional command shall be provided to change the working directory from inside `dlv`.

The primary use case for this is using `dlv` with `bazel`, where path substitution map must be generated by a helper rule and provided to `dlv` invocation.

aarzilli:
> In some cases, it is not possible to run the dlv from the project working directory

why not?

oakad:
Because of the mess bazel rules are. :-)

A good feature for dlv will be to support an explicit `directory` statement, akin to how gdb supports it.

aarzilli:
Empty strings can already be passed to `config substitute-path` as follows:

```
(dlv) config substitute-path "" nothing-into-something
(dlv) config substitute-path something-into-nothing ""
```

I'm going to keep this open to track better documentation for this.

aarzilli:
Fixed by 13ad7dc1d581b86d7000f6fc4bd37309c30ec480

Issue #3179, opened 2022-10-30
`config debug-info-directories` terminal command fails with `unsupported type for configuration key "debug-info-directories"`


dlipovetsky:
1. What version of Delve are you using (`dlv version`)?
```
Delve Debugger
Version: 1.9.1
Build: $Id: d81b9fd12bfa603f3cf7a4bc842398bd61c42940 $
```
2. What version of Go are you using? (`go version`)?
```
go version go1.19.2 linux/amd64
```
3. What operating system and processor architecture are you using?
```
GOOS="linux"
GOARCH="amd64"
```
4. What did you do?
I tried to set the `debug-info-directories` configuration parameter in the terminal.
5. What did you expect to see?
I expected to be able to set the parameter.
6. What did you see instead?
```
(dlv) config debug-info-directories
Command failed: unsupported type for configuration key "debug-info-directories"
(dlv) config debug-info-directories foo 
Command failed: unsupported type for configuration key "debug-info-directories"
(dlv) config debug-info-directories foo bar
Command failed: unsupported type for configuration key "debug-info-directories"
(dlv) config debug-info-directories [foo]
Command failed: unsupported type for configuration key "debug-info-directories"
```

dlipovetsky:
No test exercises setting this parameter. I've added a (failing) test here: https://github.com/dlipovetsky/delve/commit/23fa8a631180cb75fa58ddc58188d8c4a0aa8461. Stepping through (with delve, of course :laughing:), I found that the value is a Slice (as expected), but that the [configureSet](https://github.com/go-delve/delve/blob/922c4cebd4884bb0faca183361f9566da2868984/pkg/terminal/config.go#L40-L63) function only handles a Slice value for the `substitute-path` parameter. Otherwise, it passes the value to the [ConfigureSetSimple](https://github.com/go-delve/delve/blob/20df19e33b2e81f22d5a33d1ec780eb0610dc55f/pkg/config/split.go#L71) function, which returns an error if the value is a Slice.

dlipovetsky:
I'm happy to work on a fix, but before I begin, I would appreciate some feedback. Thanks!

aarzilli:
Delve won't start unless it can find the debug info for the executable, because of this changing the value of debug-info-directories is not very useful. It would matter only for `-save` and for plugins, but it's worth fixing for that.
[...]

Issue #2890, opened 2022-02-03
service/dap: `dlv config <param>` should print, not reset value without an explicit parameter


polinasok:
A very intuitive way to look up a config value is to run `dlv config <param>`. This worked as expected with  dlv@v1.7.3.
![image](https://user-images.githubusercontent.com/51177946/152398688-0d9349ef-a905-4136-870d-b163dfb7abad.png)
But with dlv@v1.8.0 and dlv@latest the same command tries to reset the value, which works only in some cases
![image](https://user-images.githubusercontent.com/51177946/152399014-11460f5b-0061-4485-9b1c-14440f5f10f8.png)
I think we should go back to previous consistent and intuitive behavior. Resetting values should be an explicit action. We don't want users to accidentally reset anything when they try to simply look up what they are or remind themselves what the syntax is. And there should be an easy way to look up an existing setting without having to dig through the full list of configuration values with `dlv config -list`.

@suzmue

polinasok:
I just remember that this now works:
```
dlv config -list showGlobalVariables
showGlobalVariables	false
```
It's not ideal because `-list` appears before the parameter, so switching back and force between setting and looking up values requires changes in the middle of the command, which can get quite tedious.

aarzilli:
Fixed by 13ad7dc1d581b86d7000f6fc4bd37309c30ec480

Issue #2889, opened 2022-02-03
service/dap: `dlv config` should print existing configuration, not panic


polinasok:
I keep thinking that `dlv config`, not `dlv config -list` is the command to list all current configuration values.
With dlv@v1.7.3:
![image](https://user-images.githubusercontent.com/51177946/152393348-891c1f94-a75a-42b4-99af-b5a1e1ea8d79.png)

With dlv@v1.8.0 and dlv@latest:
![image](https://user-images.githubusercontent.com/51177946/152393607-e011e649-44b0-4317-bb0b-479caf80b10a.png)

@suzmue

aarzilli:
Fixed by 13ad7dc1d581b86d7000f6fc4bd37309c30ec480

nyngwang:
@aarzilli I still have this problem.

PR #3335, opened 2023-04-17
*: misc improvements to config command and substitute-path rules

### LINKED ISSUE #2891: service/dap: `dlv config <param>` prints errors with confusing parameter names
This looks like an artifact of reusing cmd dlv config logic:
image
Since `substitute-path` has no meaning with dap, we should be catching these and rewriting them for the user.
For example, a more appropriate message would be:
```
Unable to run dlv command: too many arguments to "config substitutePath"

Usage:
dlv config substitutePath <from> <to> # Add or modify a rule
dlv config substitutePath <from> # Clear a rule
```
@suzmue

aarzilli:
A series of interconnected changes to both the terminal command
'config', DAP command 'dlv config', quality of life improvements to how
substitute-path works, and better documentation.

* Let 'config substitute-path' show the current substitute path rules
* Add a -clear command to 'config substitute-path'
* Support 'config-debug-info-directories'
* rewrite SubstitutePath to be platform independent (see below)
* document path substitution more

Regarding the rewrite of SubstitutePath: the previous version used
runtime.GOOS and filepath.IsAbs to determine which filepath separator to use
and if matching should be case insensitive. This is wrong in all situations
where the client and server run on different OSes, when examining core files
and when cross-compilation is involved.

The new version of SubstitutePath checks the rules and the input path to
determine if Windows is involved in the process, if it looks like it is it
switches to case-insensitive matching. It uses a lax version of
filepath.IsAbs to determine if a path is absolute and tries to avoid having
to select a path separator as much as possible

Fixes #2891, #2890, #2889, #3179, #3332


pkg/locspec/locations.go
```diff
@@ -479,47 +479,97 @@ func (loc *NormalLocationSpec) findFuncCandidates(bi *proc.BinaryInfo, limit int
...
+				}
+				return rest
+			}
 
-		if from != "" && strings.HasPrefix(path, from) {
-			return strings.Replace(path, from, to, 1)
+			toEndsWithSlash := hasPathSeparatorSuffix(to)
+			restStartsWithSlash := hasPathSeparatorPrefix(rest)
+
+			switch {
```
derekparker, line 563: I didn't see a test exercising this logic, did I miss it?

pkg/locspec/locations.go
```diff
@@ -479,47 +479,97 @@ func (loc *NormalLocationSpec) findFuncCandidates(bi *proc.BinaryInfo, limit int
...
-	path = crossPlatformPath(path)
-	// On windows paths returned from headless server are as c:/dir/dir
-	// though os.PathSeparator is '\\'
-
-	separator := "/"                  // make it default
-	if strings.Contains(path, "\\") { // dependent on the path
-		separator = "\\"
+	// Look for evidence that we are dealing with windows somewhere, if we are use case-insensitive matching
+	caseInsensitive := windowsAbsPath(path)
+	for i := range rules {
```
derekparker, line 515: If `caseInsensitive` is true when initialized can we skip this loop?

derekparker (changes requested):
A few questions / comments.

pkg/locspec/locations.go
```diff
@@ -479,47 +479,97 @@ func (loc *NormalLocationSpec) findFuncCandidates(bi *proc.BinaryInfo, limit int
...
+			}
 		}
 
-		// Expand relative paths with the specified prefix
-		if from == "" && !filepath.IsAbs(path) {
-			return strings.Replace(path, from, to, 1)
-		}
+		if match {
+			if to == "" {
+				// make sure we return a relative path, regardless of whether from consumed a final / or not
```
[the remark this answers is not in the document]
aarzilli: It's because `from`  here is a noun referring to the variable named `from`. I think you are reading it as a preposition. I put it in quotes, is it clearer?

pkg/locspec/locations.go
aarzilli: TestSubstitutePath in pkg/terminal/terminal_test.go should be testing everything, I believe it's there because this code once lived in pkg/terminal. Should we move it?

pkg/locspec/locations.go
aarzilli: Yes. Done.

Documentation/cli/substitutepath.md
```diff
@@ -0,0 +1,66 @@
+## Path substitution configuration
+
+Normally Delve finds the path to the source code that was used to produce an executable by looking at the debug symbols of the executable.
+However, under [some circumstances](../faq.md#substpath), the paths that end inside the executable will be different from the paths to the source code on the machine that is running the debugger. If that is the case Delve will need extra configuration to convert the paths stored inside the executable to paths in your local filesystem.
```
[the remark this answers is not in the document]
aarzilli: Done

pkg/locspec/locations.go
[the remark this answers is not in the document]
derekparker: Yes, thank you!

pkg/locspec/locations.go
derekparker: Yes, lets move it closer to the implementation.

aarzilli: Done.

derekparker (approved):
LGTM

PR #3336, opened 2023-04-18
Removed extra docs for Go 1.16.

guettli:
Removed extra docs for Go 1.16.


aarzilli:
Those install instructions are still valid, in fact we should probably move them above the clone-and-build method.

derekparker:
Cl

Issue #3342, opened 2023-04-25
Time string not always displayed


sylr:
1. What version of Delve are you using (`dlv version`)?
```
dlv version
Delve Debugger
Version: 1.20.2
Build: $Id: e0c278ad8e0126a312b553b8e171e81bcbd37f60 $
```
3. What version of Go are you using? (`go version`)?
```
go version go1.20.3 darwin/arm64
```
5. What operating system and processor architecture are you using? 
```Darwin```
7. What did you do?
```go
package main

import (
	"fmt"
	"time"
)

type Interval struct {
	Start time.Time
	End   time.Time
}

func main() {
	t := time.Now().Truncate(time.Second)
	i := Interval{Start: t, End: t.Add(time.Hour * 24)}

	ts := make([]Interval, 0, 1)
	ts = append(ts, i)

	fmt.Printf("ts: %v\n", ts)
}
```
9. What did you expect to see?

I'd like to see the Time string in ts[0].Start and ts[0].End.

11. What did you see instead?

<img width="734" alt="image" src="https://user-images.githubusercontent.com/153052/234259669-2b90d25a-3422-463b-8793-e1aa473f6c2e.png">

alexsaezm:
For what it's worth, this also happens on linux/amd64. I'm wondering what should be the expected output, exactly like `i.Start`?

sylr:
> For what it's worth, this also happens on linux/amd64. I'm wondering what should be the expected output, exactly like `i.Start`?

That would be my expectation yes.

alexsaezm:
For the record, looks like dap can be configured so it might interested to add a configuration option like "depth" and a sane default.

If the solution sounds ok, I'll work on a PR.

derekparker:
I'm not sure if the DAP folks actually intend for it to be configurable given https://github.com/go-delve/delve/blob/master/service/dap/server.go#L256. We could experiment with increasing the depth to 2, or maybe this is something better solved on the VSCode side?

cc @hyangah

suzmue:
Sent a PR for fixing the bug where the time.Time arguments have the values.

![Screenshot 2023-04-28 at 3 19 16 PM](https://user-images.githubusercontent.com/6634754/235259321-75eb4963-721c-4c6d-8ebf-b7d49b4f2910.png)

We also need to address the inconsistency with Interval sometimes having the children loaded and displaying the formatted string in its value and sometimes not.  (`i` displays time.Time but `ts[0]` doesn't).

[...]

PR #3349, opened 2023-04-28
service/dap: use reloaded value

suzmue:
The only part of the reloaded value that was used was the Children. This caused a bug where the Time format string was reloaded, but it was not being displayed.

Fixes go-delve/delve#3342

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -2485,6 +2485,7 @@ func (s *Session) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr 
 			value += fmt.Sprintf(" - FAILED TO LOAD: %s", err)
 		} else {
 			v.Children = vLoaded.Children
+			v.Value = vLoaded.Value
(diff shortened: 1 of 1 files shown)


suzmue:
This is a super minimal change to fix the issue. I think we should probably prioritize reloading in the way described in the TODO above the section. I am planning to take a look at that this afternoon.

suzmue:
Screenshot 2023-04-28 at 3 19 16 PM

sylr:
Is this normal on `ts[0]` line `Start` is shown as `*time.Time` ?

suzmue:
> Is this normal on `ts[0]` line `Start` is shown as `*time.Time` ?

Fixing the variable at that level would definitely require additional loading logic. We'd have to check if we need to reload a level earlier to make this work I think so deciding when we need to reload would be the biggest obstacle I believe and the potential performance costs of reloading more aggressively.

aarzilli (approved):

PR #3358, opened 2023-05-05
trace: add timestamp to the output

### LINKED ISSUE #3356: Feature request: adding timestamp in the output of dlv trace

1. What version of Delve are you using (`dlv version`)? 1.20.2
2. What version of Go are you using? (`go version`)? go version go1.20.3 darwin/arm64
3. What operating system and processor architecture are you using? ARM64
4. What did you do? Use `go trace`
5. What did you expect to see? output with timestamp
6. What did you see instead?

It would be great to add timestamp in the trace output so that the users can sync the function call with the application log or other external source.

For example, the current output is:
```
> goroutine(1967): ...
```

We can add a timestamp at the beginning of each line like:
```
2023-05-05 12:00:00.123 > goroutine(1967): ...
```

Some trace tools like `strace` have several options to enable timestamp, which is quite helpful.

spacewander:
Fix #3356

### DIFF
--- a/Documentation/usage/dlv_trace.md
+++ b/Documentation/usage/dlv_trace.md
@@ -28,6 +28,7 @@ dlv trace [package] regexp [flags]
   -p, --pid int         Pid to attach to.
   -s, --stack int       Show stack trace with given depth. (Ignored with --ebpf)
   -t, --test            Trace a test binary.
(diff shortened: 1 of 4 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3359, opened 2023-05-05
chore: don't ignore directory cmd/dlv

spacewander:
Currently, cmd/dlv is ignored, breaks tools which read the gitignore file. For example, rg can't search files under cmd/dlv.


derekparker (approved):
LGTM

PR #3365, opened 2023-05-11
pkg/proc: pad variable mem in extractVarInfoFromEntry

### LINKED ISSUE #3364: pkg/proc: failed to call method on struct type: write out of bounds

```
Delve Debugger
Version: 1.20.2
Build: $Id: e0c278ad8e0126a312b553b8e171e81bcbd37f60 $
```

```
go version go1.20.4 linux/amd64
```

```
linux/amd64
```
5. What did you do?

Take the following program:

```go
package main

import (
	"fmt"
	"runtime"
)

type T struct {
	a int
	b uint32
}

func (t T) String() string {
	return fmt.Sprintf("%d %d", t.a, t.b)
}

func main() {
	t := T{a: 1, b: 2}
	runtime.Breakpoint()
	fmt.Println(t.String())
}
```

and ran:

```
dlv debug
(dlv) c
(dlv) call t.String()
```

6. What did you expect to see?

```
(dlv) call t.String()
> main.main() ./main.go:20 (PC: 0x49ecbf)
Values returned:
	~r0: "1 2"
```

7. What did you see instead?

```
(dlv) call t.String()
> main.main() ./main.go:20 (PC: 0x49ecbf)
Command failed: write out of bounds
```

**PS**:

This bug is based on a real use case to print the value of  `github.com/shopspring/decimal.Decimal`. This is [the definition of `decimal.Decimal`](https://github.com/shopspring/decimal/blob/f55dd564545cec84cf84f7a53fb3025cdbec1c4f/decimal.go#L75-L84):

```go
type Decimal struct {
	value *big.Int
	exp int32
}
```

See also this question https://stackoverflow.com/questions/76222724/is-it-possible-to-preview-github-com-shopspring-decimal-values-in-vscode-during.

ZekeLu:
On 64 bit system, the byte size of the following struct is 16:

```go
type myStruct struct {
    a int
    b uint32
}
```
But `extractVarInfoFromEntry` only allocates a mem of 12 bytes for it. When calling method of this struct with the `call` command, it will result in this error:

````
write out of bounds
````

This patch extends the mem by adding padding bytes to the end of the mem.

Fixes #3364.

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -1207,6 +1207,11 @@ func extractVarInfoFromEntry(tgt *Target, bi *BinaryInfo, image *Image, regs op.
 			}
 		}
 		if cmem != nil {
+			paddingBytes := int(t.Common().ByteSize) - len(cmem.data)
+			if paddingBytes > 0 && paddingBytes < bi.Arch.ptrSize {
+				padding := make([]byte, paddingBytes)
+				cmem.data = append(cmem.data, padding...)
+			}
 			mem = cmem
 		}
 	}
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -1198,20 +1198,15 @@ func extractVarInfoFromEntry(tgt *Target, bi *BinaryInfo, image *Image, regs op.
 	if pieces != nil {
 		var cmem *compositeMemory
 		if tgt != nil {
-			addr, cmem, err = tgt.newCompositeMemory(mem, regs, pieces, descr)
+			addr, cmem, err = tgt.newCompositeMemory(mem, regs, pieces, descr, t.Common().ByteSize)
 		} else {
-			cmem, err = newCompositeMemory(mem, bi.Arch, regs, pieces)
+			cmem, err = newCompositeMemory(mem, bi.Arch, regs, pieces, t.Common().ByteSize)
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pkg/proc/variables.go [hunk 1]
```diff
@@ -1207,6 +1207,11 @@ func extractVarInfoFromEntry(tgt *Target, bi *BinaryInfo, image *Image, regs op.
 			}
 		}
 		if cmem != nil {
+			paddingBytes := int(t.Common().ByteSize) - len(cmem.data)
```
aarzilli, line 1210: I think this should be moved inside newCompositeMemory in pkg/proc/mem.go (add an extra `size int` parameter to it).

ZekeLu: Thank you Alessandro! I have moved the padding logic to `newCompositeMemory`. Since the type of `CommonType.ByteSize` is `int64`, I add the extra parameter as `size int64`. Please take another look.

aarzilli:
LGTM, thank you!

aarzilli (approved):

ZekeLu:
Thank you! Should I do a force push to squash the two commits into one?

aarzilli:
No need.

derekparker (approved):
LGTM

PR #3370, opened 2023-05-14
proc: fix runtime type handling for Go 1.21

aarzilli:
Go 1.21 renamed runtime._type to internal/abi.Type and changed the name
of its fields. Update Delve so that it uses the new names for loading
interfaces and generic type parameters.


aarzilli:
This fixes the big increase in test failures on 1.21 builds we've had during the past week.

derekparker (approved):
LGTM

PR #3373, opened 2023-05-17
proc: remove addrret field from Stackframe struct

aarzilli:
This field was part of the original stack tracing algorithm, we haven't
used this in years.


derekparker (approved):
LGTM

PR #3380, opened 2023-05-23
Bugfix: buildFlags="-C directory" is invalid

saffraan:
When set buildFlags="-C dir", go build failure. Because the "-C direcotry" option is filled after "go build -o xxx", for example:

go build -o main -C /root/main

but the "-C dir" option should be filled after "go build":

go build -C /root/main -o main

in go version 1.20.

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -60,6 +60,24 @@ func GoTestBuildCombinedOutput(debugname string, pkgs []string, buildflags strin
 	return gocommandCombinedOutput("test", args...)
 }
 
+func buildArgs(args []string) []string {
+	flags := make([]string, 0, len(args))
+	for i := 0; i < len(args); {
+		if args[i] == "-C" { // -C /path
+			if i+1 < len(args) && !strings.HasPrefix(args[i+1], "-") {
+				flags = append([]string{args[i], args[i+1]}, flags...)
+				i++
+			}
+		} else if strings.HasPrefix(args[i], "-C=") { // -C=/path
+			flags = append([]string{args[i]}, flags...)
+		} else {
+			flags = append(flags, args[i])
+		}
+		i++
+	}
+	return flags
+}
+
 func goBuildArgs(debugname string, pkgs []string, buildflags string, isTest bool) []string {
 	args := []string{"-o", debugname}
 	if isTest {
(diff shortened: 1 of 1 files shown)


aarzilli:
I've tried it both ways (`go build -o something -C somethingelse` and `go build -C somethingelse -o something`) and they seem to both do the exact same thing. What is the build failure exactly?

saffraan:
> 

(base) ➜  /tmp cd go-example
(base) ➜  go-example ls
go.mod  main.go
(base) ➜  go-example cat main.go
package main

import "fmt"

func main(){
	fmt.Println("Hello word!")
}
(base) ➜  go-example cat go.mod
module go-example

go 1.20
(base) ➜  go-example cd ..
(base) ➜  /tmp go build -gcflags all="-N -l" -C /tmp/go-example -o main
go: go.mod file not found in current directory or any parent directory; see 'go help modules'
(base) ➜  /tmp go build -C /tmp/go-example -gcflags all="-N -l" -o main
(base) ➜  /tmp ls /tmp/go-example/
go.mod  main    main.go
(base) ➜  /tmp

saffraan:
> I've tried it both ways (`go build -o something -C somethingelse` and `go build -C somethingelse -o something`) and they seem to both do the exact same thing. What is the build failure exactly?

(base) ➜ /tmp cd go-example
(base) ➜ go-example ls
go.mod main.go
(base) ➜ go-example cat main.go
package main

import "fmt"

func main(){
fmt.Println("Hello word!")
}
(base) ➜ go-example cat go.mod
module go-example

go 1.20
(base) ➜ go-example cd ..
(base) ➜ /tmp go build -gcflags all="-N -l" -C /tmp/go-example -o main
go: go.mod file not found in current directory or any parent directory; see 'go help modules'
(base) ➜ /tmp go build -C /tmp/go-example -gcflags all="-N -l" -o main
(base) ➜ /tmp ls /tmp/go-example/
go.mod main main.go
(base) ➜ /tmp

saffraan:
> I've tried it both ways (`go build -o something -C somethingelse` and `go build -C somethingelse -o something`) and they seem to both do the exact same thing. What is the build failure exactly?

sorry, I found the bug in debug mode, I haven't try to run `go build` just with `-C` optional.

aarzilli:
Looks like a bug in cmd/go, let's see.

aarzilli:
According to the [release notes for 1.21](https://tip.golang.org/doc/go1.21) it looks like `-C` needs to be the first flag passed:

```
 The -C dir flag must now be the first flag on the command-line when used. 
```

so we should do something about this. However I don't like the approach in this PR, I think it's too complicated and we should just change goBuildArgs to start the args slice with the buildflags string:

```
func goBuildArgs(debugname string, pkgs []string, buildflags string, isTest bool) []string {
	args := []string{}
	if buildflags != "" {
		args = append(args, config.SplitQuotedFields(buildflags, '\'')...)
	}
	args = append(args, "-o", debugname)
	...
```

PR #3381, opened 2023-05-23
debuginfod-find stderr may contain diagnostics; look at stdout only

fche:
Test by setting $DEBUGINFOD_PROGRESS=1 or $DEBUGINFOD_VERBOSE=1
before invoking dlv.  A "no such file or directory" dlv error results the first time,
even though debuginfod-find succeeded.  If rerun (and thus the file is found in
the debuginfod client cache), diagnostics may not be printed, so dlv may work.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3386, opened 2023-05-24
starbind: fix use of ptr variables in starlark 

andreimatei:
This patch fixes a panic in ptrVariableAsStarlarkValue.AttrNames(), which
would panic was invoked on a pointer var that wasn't loaded (i.e. it had
no children). This patch fixes it by declaring that such a variable has
no fields; this is consistent with the handling of such a variable
elsewhere.
Engineering situations where AttrNames is called on a not-loaded ptr
variable is not easy because as soon as you call .Value() on a variable,
you get into auto-loading magic behavior. But it is possible, as the
tests that I'm adding demonstrate. In particular, I've run into this
when trying to json-encode a whole variable (i.e. a
structAsStarlarkVariable<api.Variable>).


pkg/terminal/starlark_test.go
```diff
@@ -158,3 +158,70 @@ func TestStarlarkVariable(t *testing.T) {
...
+	).Variable;
+v.Children[0].Children[0].Value.XXX
+`,
+				expErr: "*int has no .XXX field or method",
+			},
+		} {
+			t.Run(tc.name, func(t *testing.T) {
+				_ /* out */, err := term.ExecStarlark(tc.script)
+				if err == nil {
+					t.Fatalf("expected error \"%s\", got success", tc.expErr)
```
aarzilli, line 219: %q instead of `\"%s\"`

pkg/terminal/starlark_test.go
```diff
@@ -158,3 +158,70 @@ func TestStarlarkVariable(t *testing.T) {
...
+v = eval(
+		None, "ba", {"MaxVariableRecurse": 0, "MaxStructFields": 10}
+	).Variable;
+v.Children[0].Children[0].Value.XXX
+`,
+				expErr: "*int has no .XXX field or method",
+			},
+		} {
+			t.Run(tc.name, func(t *testing.T) {
+				_ /* out */, err := term.ExecStarlark(tc.script)
```
aarzilli, line 217: Remove the commented out `out`.

pkg/terminal/starlark_test.go
```diff
@@ -158,3 +158,70 @@ func TestStarlarkVariable(t *testing.T) {
...
+			},
+			{
+				// In this test, MaxStructFields = 10 and MaxVariableRecurse = 0
+				// will cause us to load v, and v.Children[0] (in the sense that
+				// v.Children[0] has a child), but _not_ load
+				// v.Children[0].Children[0] (because the recursion limit is
+				// exhausted by the descent into the top-level struct, so we
+				// don't load what hides under the interface).
+				//
+				// v 							   -> structAsStarlarkValue<api.Variable<StructWithInterface>>
```
aarzilli, line 202: If you want to do fancy ascii alignment here you have to do it with spaces, not tabs.

aarzilli (commented):
Just a couple of minor things.

pkg/terminal/starlark_test.go
andreimatei: ugh, I thought I had.
Done.

pkg/terminal/starlark_test.go
andreimatei: done

pkg/terminal/starlark_test.go
andreimatei: done

aarzilli (approved):
LGTM, thank you.

derekparker (approved):
LGTM

PR #3387, opened 2023-05-25
terminal: expand ~ in paths passed to 'source'

aarzilli:
If the path passed to source starts with a ~ expand it into the user's
home directory, like the shell would.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/terminal/command.go
```diff
@@ -2451,15 +2452,26 @@ func (c *Commands) sourceCommand(t *Term, ctx callContext, args string) error {
...
 	}
 
+	if args == "-" {
+		return t.starlarkEnv.REPL()
+	}
+
+	if runtime.GOOS != "windows" && strings.HasPrefix(args, "~") {
+		home, err := os.UserHomeDir()
+		if err == nil {
+			if args == "~" {
```
derekparker, line 2462: Could we simplify this with `strings.Replace(args, "~", home, 1)`.

aarzilli: It should only be expanded if it's at the start, for example `blah~` shouldn't be expanded. Also it shouldn't be expanded if it isn't a directory component, like `~blah/blah.star`.

derekparker (approved):
LGTM

PR #3388, opened 2023-05-25
terminal/starbind: add online help for starlark

aarzilli:
Adds a new starlark builtin 'help' that prints the list of available
builtins when called without arguments and help for the specified
builtin when passed an argument.

The help is autogenerated from godoc comments so it isn't always
exactly accurate for starlark (in particular we sometimes refer to the
In structs), but it's better than nothing.


derekparker (approved):
LGTM

PR #3389, opened 2023-05-25
debugger: inline clearBreakpoint, createLogicalBreakpoint

aarzilli:
These two functions are only called once and can be inlined easily.


derekparker (approved):
LGTM

PR #3390, opened 2023-05-26
DAP sets suspended breakpoints.

jknightdoeswork:
In VSCode, when you set a breakpoint in code that is in a plugin, it will not work. You have to re-activate the breakpoint manually after plugin.Open is hit. This (untested) commit, should fix that.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1379,7 +1379,7 @@ func (s *Session) setBreakpoints(prefix string, totalBps int, metadataFunc func(
 				err = setLogMessage(bp, want.logMessage)
 				if err == nil {
 					// Create new breakpoints.
-					got, err = s.debugger.CreateBreakpoint(bp, "", nil, false)
+					got, err = s.debugger.CreateBreakpoint(bp, "", nil, true)
 				}
 			}
 		}


aarzilli:
See failed tests. Also DAP has this thing about verified/unverified breakpoints that we should be doing for suspended breakpoints, which means sending breakpoint events (which we never do currently). Just doing this means users will be able to set breakpoints on everything and anything without any feedback on whether the breakpoint makes sense.

jknightdoeswork:
> See failed tests.

Once we find common ground on where the finish line is for this PR, I will repair the test suite.

> Just doing this means users will be able to set breakpoints on everything and anything without any feedback on whether the breakpoint makes sense.

I don't think anything has changed here. Currently, users can 'set a breakpoint' in a plugin source in VSCode, and there is 0 user feedback that anything has failed. Even after starting the program, VSCode provides 0 user feedback that this breakpoint has failed to set. I'm not aware of any actual user workflows that would experience a degradation in UX after this change. Can you provide a description of a  user workflow that would be negatively affected by this change? Can you elaborate on what you mean by feedback? What is a breakpoint that 'does not make sense', how can a user set one, and how should the user receive feedback that it has failed? When we say 'user' are we referring to a client of the DAP server or a human being using an editor? Is that editor basically always VSCode?

My current workflow is to set a breakpoint in VSCode shortly after plugin.Open() is called, and then disable and re-enable all my breakpoints in VSCode. This is a poor user experience that we should strive to repair. It requires several complex manual steps to start the program, and discovering these manual steps requires hidden tacit knowledge many users would not have (ie: breakpoints are based on symbols that are not loaded until plugin.Open is called). I'm of the opinion that the only reason this isnt already fixed in delv is because not many go developers use plugins.

aarzilli:
> Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it

This might depend on the theme you are using:

2023-05-27-110432_2182x1651_scrot

In this example the breakpoint on line 38 is unverified and the one on line 44 is verified. If you hover the mouse over the breakpoint on line 38 it will tell you why.

> My current workflow is to set a breakpoint in VSCode shortly after plugin.Open() is called, and then disable and re-enable all my breakpoints in VSCode. This is a poor user experience that we should strive to repair. It requires several complex manual steps to start the program, and discovering these manual steps requires hidden tacit knowledge many users would not have

Agreed, but this PR is insufficient.

derekparker:
No updates on the PR in some time, closing. Please reopen if you still plan to work on this.

PR #3391, opened 2023-05-26
dwarf/op: use readMemory function

aarzilli:
dwarf/op gained the ability to execute DW_OP_deref opcodes a while ago
but because we didn't save the readMemory function in the context
structure it never worked.


derekparker (approved):
LGTM

PR #3392, opened 2023-05-28
TeamCity: actually run linux/386 tests on linux/386

aarzilli:
It seems that our linux/386 builders no longer (or never did?) use
linux/386, the i386/ubuntu:20.04 docker image has become (or always
was?) multiarchitecture and docker starterd picking (always picked?)
the amd64 version. This change should fix the problem (or break
everything).


aarzilli:
Changes to settings.kts aren't used until merged so we have to merge and see what happens. If all linux builds break we revert.

derekparker (approved):
LGTM

PR #3394, opened 2023-05-28
proc/test: reenable cgo testing on FreeBSD

aarzilli:
Cgo testing on FreeBSD was accidentally disabled by 4e24209, FreeBSD
does not have gcc by default anymore.

### DIFF
--- a/pkg/proc/test/support.go
+++ b/pkg/proc/test/support.go
@@ -364,8 +364,9 @@ var hasCgo = func() bool {
 	if strings.TrimSpace(string(out)) != "1" {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


pkg/proc/test/support.go
```diff
@@ -364,6 +364,9 @@ var hasCgo = func() bool {
 	if strings.TrimSpace(string(out)) != "1" {
 		return false
 	}
+	if runtime.GOOS == "freebsd" {
```
derekparker, line 367: Should we instead check for the presence of `clang`?

derekparker (approved):
LGTM

PR #3397, opened 2023-05-30
service/rpccommon: correct wrong comments

### LINKED ISSUE #3315: Wrong documentation for RPC server.Run()?
Latest: 1.20.1

1.20

MacOS

4. What did you do?
Im starting delve RPC server manually, without dlv app.

Right documentation for method.

6. What did you see instead?
Wrong documentation? 

See: https://github.com/go-delve/delve/blob/master/service/rpccommon/server.go

```
// Run starts a debugger and exposes it with an HTTP server. The debugger
// itself can be stopped with the `detach` API. Run blocks until the HTTP
// server stops.
func (s *ServerImpl) Run() error {
```

But by the code, this method starts background goroutine and doesn't blocks invocation or `Run()`.

aarzilli:
This is not an HTTP server.

Fixes #3315

### DIFF
--- a/service/rpccommon/server.go
+++ b/service/rpccommon/server.go
@@ -32,7 +32,7 @@ import (
 type ServerImpl struct {
 	// config is all the information necessary to start the debugger and server.
 	config *service.Config
-	// listener is used to serve HTTP.
+	// listener is used to serve JSON-RPC.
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

PR #3398, opened 2023-05-30
terminal: ask for confirmation when using 'quit -c' with breakpoints

### LINKED ISSUE #1961: Feature request: clear all client breakpoints on exit
Delve is currently keeping all breakpoints from a client after they quit in a headless multiclient context. The problem is if the headless server instance no longer has clients connected, it will continue to break at every breakpoint left without a way to continue execution, unless another client connects to it in order to `continue` and hence "save" its execution. It looks like a kind of leak.

What if delve headless server automatically clears all breakpoints from a client when this client exit/disconnect?

aarzilli:
If there are some breakpoints and the user executes 'quit -c' (quit and
continue) ask for confirmation since the behavior in that case may be
unexpected.

Fixes #1961

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -2876,6 +2876,20 @@ func exitCommand(t *Term, ctx callContext, args string) error {
 		if !t.client.IsMulticlient() {
 			return errors.New("not connected to an --accept-multiclient server")
 		}
+		bps, _ := t.client.ListBreakpoints(false)
+		hasUserBreakpoints := false
+		for _, bp := range bps {
+			if bp.ID >= 0 {
+				hasUserBreakpoints = true
+				break
+			}
+		}
+		if hasUserBreakpoints {
+			yes, _ := yesno(t.line, "There are breakpoints set, do you wish to quit and continue without clearing breakpoints? [Y/n] ", "yes")
+			if !yes {
+				return nil
+			}
+		}
 		t.quitContinue = true
 	}
 	return ExitRequestError{}


pkg/terminal/command.go
```diff
@@ -2876,6 +2876,19 @@ func exitCommand(t *Term, ctx callContext, args string) error {
 		if !t.client.IsMulticlient() {
 			return errors.New("not connected to an --accept-multiclient server")
 		}
+		bps, _ := t.client.ListBreakpoints(false)
+		numbp := 0
+		for _, bp := range bps {
+			if bp.ID >= 0 {
+				numbp++
```
derekparker, line 2883: We can just short circuit here, we don't care how many breakpoints are set, just that they exist.

derekparker (changes requested):
Just a few small changes (posted my approval accidentally without this feedback).

pkg/terminal/command.go
```diff
@@ -2876,6 +2876,19 @@ func exitCommand(t *Term, ctx callContext, args string) error {
 		if !t.client.IsMulticlient() {
 			return errors.New("not connected to an --accept-multiclient server")
 		}
+		bps, _ := t.client.ListBreakpoints(false)
+		numbp := 0
```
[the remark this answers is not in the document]
aarzilli: Done.

pkg/terminal/command.go
aarzilli: Done.

derekparker (approved):
LGTM

PR #3399, opened 2023-05-30
docs: fix link

andreimatei:
A trailing dot was breaking a docs link.


aarzilli:
This documentation is autogenerated from the contents of pkg/terminal/command.go. In this case the problem is with the conversion script (_script/gen-cli-docs.go which calls WriteMarkdown in pkg/terminal/docgen.go).

andreimatei:
Heh, I had noticed that the text in question was present in two places, but I assumed it had just been copy-pasted.

I've now done something to the generation script; PTAL.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3408, opened 2023-06-08
pkg/proc: enable basic debug functionality for stripped ELF binaries

derekparker:
We used to parse the .gopclntab section but removed support in favor of
simply using DWARF debug information, due to lack of C symbols among
other reasons. This makes it impossible to debug stripped binaries,
which some distrubutions ship by default.

Add back in basic support for .gopclntab which survives if the binary
is stripped, allowing for rudimentary debugging such as basic
program navigation, tracing, etc...


pkg/proc/threads.go
```diff
@@ -71,6 +71,24 @@ func topframe(g *G, thread Thread) (Stackframe, Stackframe, error) {
 	var frames []Stackframe
 	var err error
 
+	var pc uint64
```
aarzilli, line 74: What's the problem here with just letting ThreadStacktrace/(*G).Stacktrace get called? I think it should at least be able to return a single frame.

pkg/proc/bininfo.go
```diff
@@ -332,7 +333,7 @@ func FindFunctionLocation(p Process, funcName string, lineOffset int) ([]uint64,
 
 	if lineOffset > 0 {
 		fn := origfns[0]
-		filename, lineno := fn.cu.lineInfo.PCToLine(fn.Entry, fn.Entry)
+		filename, lineno, _ := bi.PCToLine(fn.Entry)
```
aarzilli, line 336: What about adding a `func (*Function) pcToLine` method that does the right thing?

aarzilli (commented):
Just a couple of minor things

pkg/proc/proc_test.go
```diff
@@ -3221,11 +3221,37 @@ func TestAttachStripped(t *testing.T) {
...
 	t.Logf("error is %v", err)
 
-	if err == nil {
-		p.Detach(true)
+	if err != nil {
+		cmd.Process.Kill()
 		t.Fatalf("expected error after attach, got nothing")
 	} else {
-		cmd.Process.Kill()
+		go func() {
```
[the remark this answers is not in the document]
derekparker: Ack, agreed, updated.

pkg/proc/bininfo.go
aarzilli: Just not re-doing the lookup of fn. More of an aesthetic preference than anything. If you don't like it it's also fine.

derekparker: Ack, refactored a bit to remove the double lookup.

pkg/proc/threads.go
derekparker: Done. Might still panic due to instances of `lineInfo` being used still in the stacktrace code but I'm happy to fix those as they come up in subsequent PRs. I want to let this marinate for a bit anyways prior to cutting a release.

aarzilli (approved):
LGTM

kralicky:
@derekparker I'm (erroneously?) getting the "no debug info found" warning for these system libraries:
- `/usr/lib/libnss_mymachines.so.2`, `/usr/lib/libnss_resolve.so.2` (systemd-libs)
- `/usr/lib/libnss_mdns_minimal.so.2` (nss-mdns)
The warning message doesn't specify which binary is missing debug symbols, so this was a bit confusing to run into.

PR #3410, opened 2023-06-10
*: fix arm64 docker builds

derekparker:
There is an error in the TeamCity test logs:

```
docker: Error response from daemon: image with reference
arm64v8/ubuntu:20.04 was found but does not match the specified
platform: wanted linux/arm64v8, actual: linux/arm64/v8.
```


aarzilli:
I think that line is actually correct and the problem is that I forgot to change the rest of the script to actually use `dockerPlatformArch`, the line that says:

```
                        --platform linux/$dockerArch
```

should have been changed to:

```
                        --platform linux/$dockerPlatformArch
```

PR #3415, opened 2023-06-13
terminal,service: better printing of suspended breakpoints

aarzilli:
Show the location expression that will be used to set a suspended
breakpoint in the breakpoints list.

Also change 'target' called without arguments to print a better error
message and 'target follow-exec' without the last argument to print the
state of follow-exec.


derekparker (approved):
LGTM

Issue #3227, opened 2022-12-19
ebpf probes don't work after time.Sleep()


andreimatei:
Introducing a `time.Sleep()` seems to break the ebpf uprobes.
Consider `trace --ebpf` on the program below:
```
sudo ./dlv trace /home/andrei/src/github.com/andreimatei/play/example-go/main.go tracedFunction --ebpf --log --log-output=debugger
```

```
package main

import (
	"fmt"
	"time"
)

func main() {
	i := int64(0)
	for i = 0; i < 5; i++ {
		tracedFunction(i)
	}
	for i = 0; i < 5; i++ {
		tracedFunction(i)
		time.Sleep(time.Second)
	}
}

//go:noinline
func tracedFunction(x int64) {
	fmt.Println(x)
}
```

The output I get looks like this:

```
0
1
2
3
4
0
> (1) main.tracedFunction(0)
> (1) main.tracedFunction(1)
> (1) main.tracedFunction(2)
> (1) main.tracedFunction(3)
> (1) main.tracedFunction(4)
> (1) main.tracedFunction(0)
1
2
3
4
```

So, the probe works fine for the first round of invocations of `tracedFunction` but they don't seem to work after the sleeps start - although the program is still running.
What's mesmerizing is that the duration of the sleep seems to matter. On my machine, durations shorter than 90ms work fine, and then somewhere in between 90ms and 100ms things get weird; sometimes the probe hits for the 7th call, other times is doesn't.

Without `--ebpf`, tracing works as expected.

I'm using Delve from today's `master` at `72c77b1c30fccead4c8d03e9bb9fbe071c156ace`.

mayanksingh2298:
Hi.. Has there been any update on this?

PR #3417, opened 2023-06-14
pkg/proc: fix ebpf probe tracing backen uprobe handling

derekparker:
Uprobes get automatically cleaned and removed when the reference to
the Link object is lost. Hold a reference to any active Uprobe Link
for duration of Delve execution and ensure they are cleaned up
at exit.

Fixes ebpf probes don't work after time.Sleep() #3227


aarzilli (approved):
LGTM

PR #3419, opened 2023-06-15
service,terminal: when libraries don't have debug_info print reason

aarzilli:
Print the reason why libraries don't have debug info in response to the
'libraries' command.


derekparker (approved):
LGTM

PR #3420, opened 2023-06-15
proc: only print warning when gopclntab can not be read for first image

aarzilli:
Only print the warning that gopclntab can not be read for the first
image (i.e. the executable file), also change the returned when neither
DWARF nor gopclntab are found to preserve the DWARF error.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -3169,6 +3169,7 @@ func TestDebugStripped(t *testing.T) {
 	// Currently only implemented for Linux ELF executables.
 	// TODO(derekparker): Add support for Mach-O and PE.
 	skipUnlessOn(t, "linux only", "linux")
+	skipOn(t, "does not work with PIE", "pie")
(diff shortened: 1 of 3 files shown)


aarzilli:
PS. this also fixes this functionality for go1.21

pkg/proc/proc_test.go
```diff
@@ -3169,6 +3169,7 @@ func TestDebugStripped(t *testing.T) {
 	// Currently only implemented for Linux ELF executables.
 	// TODO(derekparker): Add support for Mach-O and PE.
 	skipUnlessOn(t, "linux only", "linux")
+	skipOn(t, "does not work with PIE", "pie")
```
derekparker, line 3172: This test should work with PIE executables via the magic number lookup. I think for 1.21 the magic number lookup function just needs to be updated. I think instead of using `strings.Compare` we should use our own `AfterOrEqual`.

derekparker (approved):
LGTM

Issue #3423, opened 2023-06-17
Some breakpoints get set to address 0 after restart


awmorgan:
Delve version
```
Delve Debugger
Version: 1.20.2
Build: $Id: e0c278ad8e0126a312b553b8e171e81bcbd37f60 $
```
Go version
```
go version go1.20.5 darwin/amd64
```
OS / CPU architecture
```
Darwin Kernel Version 20.6.0: Thu Mar  9 20:39:26 PST 2023; root:xnu-7195.141.49.700.6~1/RELEASE_X86_64 x86_64
```
What did you do?
Set a breakpoint at a line number with `b 4` then restart with `r`

What did you expect to see?
Valid breakpoint after the restart.

What did you see instead?
Invalid breakpoint with address 0.

how do I prevent losing breakpoints after a restart? here you can see that after a restart bp 2 was changed to address 0 and not hit after the restart. bp 1 was retained after the restart.
```
/Volumes/git/x $ cat main.go
package main

func main() {
        println("Hello, world!")
}
/Volumes/git/x $ dlv debug .
Type 'help' for list of commands.
(dlv) b main.main
Breakpoint 1 set at 0x105fb66 for main.main() ./main.go:3
(dlv) c
> main.main() ./main.go:3 (hits goroutine(1):1 total:1) (PC: 0x105fb66)
     1: package main
     2:
=>   3: func main() {
     4:         println("Hello, world!")
     5: }
(dlv) b 4
Breakpoint 2 set at 0x105fb74 for main.main() ./main.go:4
(dlv) c
> main.main() ./main.go:4 (hits goroutine(1):1 total:1) (PC: 0x105fb74)
     1: package main
     2:
     3: func main() {
=>   4:         println("Hello, world!")
     5: }
(dlv) bp
Breakpoint runtime-fatal-throw (enabled) at 0x1031aa0,0x1031ba0 for (multiple functions)() <multiple locations>:0 (0)
Breakpoint unrecovered-panic (enabled) at 0x1031f40 for runtime.fatalpanic() /usr/local/go/src/runtime/panic.go:1145 (0)
        print runtime.curg._panic.arg
Breakpoint 1 (enabled) at 0x105fb66 for main.main() ./main.go:3 (1)
Breakpoint 2 (enabled) at 0x105fb74 for main.main() ./main.go:4 (1)
(dlv) r
Process restarted with PID 24326
(dlv) c
[...]

PR #3424, opened 2023-06-18
debugger: Fix issue with breakpoint reset after restart (#3423)

awmorgan:
This commit addresses an issue where breakpoints would get set to address 0 after a restart. The issue was rooted in the lack of preservation of certain breakpoint information during the process of creating a breakpoint in the Debugger's CreateBreakpoint method.


aarzilli:
I think this is the wrong way to fix the bug, that field is supposed to represent how the breakpoint was set.

aarzilli:
I'll look into it tomorrow.

awmorgan:
OK. Thanks.

aarzilli:
Closing in favor of #3425

PR #3427, opened 2023-06-20
proc: skip TestDebugStripped on linux/386/pie

aarzilli:
PIEs do not have a .data.rel.ro.gopclntab section on linux/386. Disable
the test.


derekparker (approved):
LGTM

PR #3436, opened 2023-07-07
pkg/proc: simplify code with trivial changes

gocurr:


pkg/proc/core/delve_core.go
```diff
@@ -101,9 +101,6 @@ func threadsFromDelveNotes(p *process, notes []*note) (proc.Thread, error) {
 				return nil
 			}
 			buf := make([]byte, len)
-			if readerr != nil {
```
aarzilli, line 104: readerr can be set by the call to read.

pkg/proc/disasm.go
```diff
@@ -92,7 +92,7 @@ func firstPCAfterPrologueDisassembly(p Process, fn *Function, sameline bool) (ui
 			continue
 		}
 		if checkPrologue(text, prologue) {
-			r := &text[len(prologue)]
+			r := text[len(prologue)]
```
aarzilli, line 95: I don't see why I should want to make a copy of the instruction.

gocurr: Thanks. After disassembly,
it is indeed found that some assembly codes will be added if symbol "&" is not used.

aarzilli (approved):
LGTM

PR #3439, opened 2023-07-12
all: use "len == 0" rather than "len <= 0" when checking empty slice/…

gocurr:
…string

This PR simplifies the empty slice/string check since we know that the length of the slice/string must be non-negative.
People may think that this PR is meaningless, and I can totally understand it.


xhd2015:
Looks reasonable, I actually hate that people write 
```go
if 0 >= len(list) {
...
```

derekparker (approved):
LGTM

PR #3441, opened 2023-07-14
proc: refactorings to implement follow-exec mode on Windows

aarzilli:
Miscellaneous non-functional changes to prepare for adding support for
follow-exec mode on Windows:

* removed (*nativeProcess).wait function from Windows backend (unused).
* move close of ptraceDoneChan from release to handlePtraceFuncs, this
makes postExit callable by a function executed by execPtraceFunc.

* change addTarget to detach before creating the target object if we
don't actually want to attach to the child process, also moved the
detach call to (*processGroup).add instead of having one in addTarget
and one in the code that calls (*processGroup).add.

* changed Detach to be a method of TargetGroup/ProcessGroup, the
Windows backend will need access to the process group to call
WaitForDebugEvent.

* moved resume method to processGroup. First all threads stopped at a
breakpoint need to be stepped, then all other threads can be resumed. This is true also for linux even though it didn't cause the current
tests to fail.

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -748,7 +748,7 @@ func (p *gdbProcess) initialize(path, cmdline string, debugInfoDirs []string, st
 	})
 	_, err = addTarget(p, p.conn.pid, p.currentThread, path, stopReason, cmdline)
 	if err != nil {
-		p.Detach(true)
+		p.Detach(p.conn.pid, true)
 		return nil, err
 	}
 	return grp, nil
@@ -1058,7 +1058,9 @@ func (p *gdbProcess) getCtrlC(cctx *proc.ContinueOnceContext) bool {
 
 // Detach will detach from the target process,
 // if 'kill' is true it will also kill the process.
-func (p *gdbProcess) Detach(kill bool) error {
+// The _pid argument is unused as follow exec
+// mode is not implemented with this backend.
+func (p *gdbProcess) Detach(_pid int, kill bool) error {
 	if kill && !p.exited {
 		err := p.conn.kill()
 		if err != nil {
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -96,17 +96,24 @@ func (dbp *nativeProcess) BinInfo() *proc.BinaryInfo {
 // StartCallInjection notifies the backend that we are about to inject a function call.
 func (dbp *nativeProcess) StartCallInjection() (func(), error) { return func() {}, nil }
 
+// detachWithoutGroup is a helper function to detach from a process which we
+// haven't added to a process group yet.
+func detachWithoutGroup(dbp *nativeProcess, kill bool) error {
+	grp := &processGroup{procs: []*nativeProcess{dbp}}
+	return grp.Detach(dbp.pid, kill)
+}
+
 // Detach from the process being debugged, optionally killing it.
-func (dbp *nativeProcess) Detach(kill bool) (err error) {
+func (procgrp *processGroup) Detach(pid int, kill bool) (err error) {
+	dbp := procgrp.procForPid(pid)
 	if dbp.exited {
 		return nil
 	}
 	if kill && dbp.childProcess {
-		err := dbp.kill()
+		err := procgrp.kill(dbp)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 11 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -1042,7 +1042,7 @@ func (p *gdbProcess) getCtrlC(cctx *proc.ContinueOnceContext) bool {
 
 // Detach will detach from the target process,
 // if 'kill' is true it will also kill the process.
-func (p *gdbProcess) Detach(kill bool) error {
+func (p *gdbProcess) Detach(pid int, kill bool) error {
```
derekparker, line 1045: Can you add a comment to the doc string about why `pid` is unused (because process groups aren't implemented with this backend), and maybe prefix with `_` to make that clear in the body.

derekparker (changes requested):

pkg/proc/target.go
```diff
@@ -323,28 +323,6 @@ func (p *Target) SwitchThread(tid int) error {
 	return fmt.Errorf("thread %d does not exist", tid)
 }
 
-// detach will detach the target from the underlying process.
-// This means the debugger will no longer receive events from the process
-// we were previously debugging.
-// If kill is true then the process will be killed when we detach.
-func (t *Target) detach(kill bool) error {
```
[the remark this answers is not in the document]
aarzilli: I'm not sure I understand. The process group doesn't have access to the target object and `(*processGroup).Detach` is in a different package and can't call private methods of proc.Target.

pkg/proc/native/proc_linux.go
```diff
@@ -592,20 +591,29 @@ func exitGuard(dbp *nativeProcess, procgrp *processGroup, err error) error {
...
-			thread.CurrentBreakpoint.Clear()
 		}
 	}
 	// everything is resumed
-	for _, thread := range dbp.threads {
-		if err := thread.resume(); err != nil && err != sys.ESRCH {
-			return err
+	for _, dbp := range procgrp.procs {
+		if valid, _ := dbp.Valid(); valid {
+
```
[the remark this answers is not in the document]
aarzilli: Done

pkg/proc/gdbserial/gdbserver.go
aarzilli: Done

pkg/proc/native/proc.go
```diff
@@ -96,17 +96,24 @@ func (dbp *nativeProcess) BinInfo() *proc.BinaryInfo {
...
+// detachWithoutGroup is a helper function to detach from a process which we
+// haven't added to a process group yet.
+func detachWithoutGroup(dbp *nativeProcess, kill bool) error {
+	grp := &processGroup{procs: []*nativeProcess{dbp}}
+	return grp.Detach(dbp.pid, kill)
+}
+
 // Detach from the process being debugged, optionally killing it.
-func (dbp *nativeProcess) Detach(kill bool) (err error) {
+func (procgrp *processGroup) Detach(pid int, kill bool) (err error) {
```
derekparker, line 107: > I'm not sure I understand. The process group doesn't have access to the target object and `(*processGroup).Detach` is in a different package and can't call private methods of proc.Target.

Quoting the reply from the other review comment. I should have put the comment against this change instead. Why not leave the `(nativeProcess) Detach)` method and add the `(processGroup)` one which simply calls `nativeProcess.Detach`, so the code paths which now use `detachWithoutGroup` can remain unchanged without having to jump through this hoop to detach.

aarzilli: I have to move the kill method to processGroup because on Windows it needs it to call waitForDebugEvent:

```
func (procgrp *processGroup) kill(dbp *nativeProcess) error {
	if dbp.exited {
		return nil
	}

	p, err := os.FindProcess(dbp.pid)
	if err != nil {
		return err
	}
	defer p.Release()

	// TODO: Should not have to ignore failures here,
	// but some tests appear to Kill twice causing
	// this to fail on second attempt.
	_ = syscall.TerminateProcess(dbp.os.hProcess, 1)

	dbp.execPtraceFunc(func() {
		procgrp.waitForDebugEvent(waitBlocking | waitDontHandleExceptions)
	})

	p.Wait()

	dbp.postExit()
	return nil
}
```

derekparker (approved):
lgtm

PR #3445, opened 2023-07-19
proc: add waitfor option to attach

aarzilli:
Adds a waitfor option to 'dlv attach' that waits for a process with a
name starting with a given prefix to appear before attaching to it.

Debugserver on macOS does not support follow-fork mode, but has this
feature instead which is not the same thing but still helps with
multiprocess debugging somewhat.


derekparker (approved):
LGTM

Issue #2887, opened 2022-02-02
allow pretty printing of variables using starlark scripts


aarzilli:
Split from discussion in PR #2865

It could be useful to let user pretty print variables using starlark scripts. There are however some obstacles to doing this:

* starlark evaluation would have to be moved server side
* how are pretty printing script going to be loaded automatically?
* what happens to the pretty printer for time.Time, which is much easier to write in go rather than starlark
* for the JSON-RPC API we can save the pretty printed value in the Value field of api.Variable, which is unused for things that users would want to prettyprint (structs, pointers to structs, interfaces -- but what happens if a pretty printer is defined for an integer type?) but how does this interact with DAP?

mark-pictor-csec:
> what happens to the pretty printer for time.Time, which is much easier to write in go rather than starlark

I had a thought related to this, but it's flawed. I'll throw it out there in case someone sees a workaround for the flaw...

Would it be possible to have a generic pretty printer that'd work for any type implementing `fmt.Stringer`, `fmt.GoStringer`, and/or `fmt.Formatter`, by invoking that interface method? Such a pretty printer could be written in go, should be straightforward, and would cover many types.

The flaw is, there's no guarantee that the method is side effect free.

PR #3454, opened 2023-08-05
starbind: fix Starlark slice unmarshaling

andreimatei:
The unmarshaling code for slices wasn't actually setting the destination. This patch fixes it.

### DIFF
--- a/pkg/terminal/starbind/conv.go
+++ b/pkg/terminal/starbind/conv.go
@@ -664,8 +664,9 @@ func unmarshalStarlarkValueIntl(val starlark.Value, dst reflect.Value, path stri
 		if dst.Kind() != reflect.Slice {
 			return converr()
 		}
+		dst.Set(reflect.MakeSlice(dst.Type(), val.Len(), val.Len()))
 		for i := 0; i < val.Len(); i++ {
-			cur := reflect.New(dst.Type().Elem())
+			cur := dst.Index(i).Addr()
 			err := unmarshalStarlarkValueIntl(val.Index(i), cur, path)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


aarzilli (approved):

aarzilli:
LGTM thank you

aarzilli:
Are you using the scripting engine for something?

If your using it I'm curious what you think about this: I've been thinking about moving starlark evaluation server side for a while (#2887) this would make it faster and allow for it to be used for more things (pretty printing, breakpoint conditions), OTOH it wouldn't be possible to invoke command line commands (i.e. `dlv_command`).

andreimatei:
Hi Alessandro!
I've been using the scripting engine for the past few weeks to prototype a debugging product. Delve is probably not what I'll ultimately use, but rather some dynamic instrumentation techniques like BPF (for the sake of speed of the instrumented binary). Still, prototyping with Delve is immensely useful.
I've hacked Delve to allow running scripts when breakpoints are hit, and I've also added my own RPC that stops the target and runs a custom script. For example, I have a script that walks all the stacks and recognizes certain frames. I could clean up and upstream this support for running scripts in more contexts, if you're interested. Depending on what exactly you mean by moving script evaluation "server side", I may have already hacked it.

#2887 talks specifically about pretty-printing, which is something that I've also been thinking a lot about. Besides Starlark, I'm wondering if you've considered ways to directly invoke `String()` methods on the respective type (for example through general the support Delve already has for invoking function from the target).

derekparker:
> Hi Alessandro! I've been using the scripting engine for the past few weeks to prototype a debugging product. Delve is probably not what I'll ultimately use, but rather some dynamic instrumentation techniques like BPF (for the sake of speed of the instrumented binary).

Delve does also have the ability to set eBPF probes, any help with regard to support there would be great if you're interested!

derekparker (approved):
LGTM

PR #3455, opened 2023-08-07
[WIP] fixes for function call injection in 1.22

aarzilli:


aarzilli:
The call injection fixes actually also affect problems with versions prior to go1.22.

aarzilli:
cc @hyangah TestFatalThrowBreakpoint of server/dap is currently failing with go1.22 because optimizations make the variable holding the error reason unreadable at entry. What do you think we should do about it?

thanm:
I commented on the issue -- https://github.com/golang/go/issues/62523#issuecomment-1711756719

I think there may need to be a Delve change here if possible, unless we want to roll back or change CL http://go.dev/cl/c/go/+/390420 in some way.

derekparker (approved):
lgtm

PR #3456, opened 2023-08-08
gobuild: pass user specified arguments of go build first

aarzilli:
The -C argument must come first on the command line of 'go build' so
pass the flags specified by build-flags first.

Replaces #3380

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -61,13 +61,24 @@ func GoTestBuildCombinedOutput(debugname string, pkgs []string, buildflags strin
 }
 
 func goBuildArgs(debugname string, pkgs []string, buildflags string, isTest bool) []string {
-	args := []string{"-o", debugname}
+	var args []string
+
+	bfv := config.SplitQuotedFields(buildflags, '\'')
+	if len(bfv) >= 2 && bfv[0] == "-C" {
+		args = append(args, bfv[:2]...)
+		bfv = bfv[2:]
+	} else if len(bfv) >= 1 && strings.HasPrefix(bfv[0], "-C=") {
+		args = append(args, bfv[0])
+		bfv = bfv[1:]
+	}
+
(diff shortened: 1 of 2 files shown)


service/debugger/debugtty
derekparker, line 1: Leftover binary file?

pkg/gobuild/gobuild.go
```diff
@@ -61,14 +61,15 @@ func GoTestBuildCombinedOutput(debugname string, pkgs []string, buildflags strin
...
+	args := []string{}
+	if buildflags != "" {
+		args = append(args, config.SplitQuotedFields(buildflags, '\'')...)
+	}
+	args = append(args, "-o", debugname)
 	if isTest {
 		args = append([]string{"-c"}, args...)
 	}
 	args = append(args, "-gcflags", "all=-N -l")
-	if buildflags != "" {
```
derekparker, line 69: This actually changes behavior. Previously, these user-provided build flags could override our defaults since they came later in the argument list, now there's no way to override any of our defaults. I know it's kinda less than ideal but perhaps we just special case `-C`?

aarzilli: Special casing just `-C` is hard because to do it we have to actually parse the arguments of `go build`, for example `go build -o -C test.go` outputs an executable called `-C`. The naive thing of just looking for `-C` and moving it first is wrong.

The other thing we could do is nothing. It's unlikely that passing `-C` in buildflags will work anyway, we pass our own `-o` as a relative path, unless the user overrides that with an absolute path we won't be able to find the output executable.

aarzilli: On second thought we can assume that if it's there it's the first thing of buildflags which makes it easy.

aarzilli: Changed to only move -C

derekparker (approved):
lgtm

PR #3458, opened 2023-08-08
proc: fix stacktraces on freebsd/amd64/go1.20

aarzilli:
TestStacktraceGoroutine failed intermittently on freebsd/amd64/go1.20
due to stacktraces being broken when doing a runtime.nanotime system
call.

### DIFF
--- a/pkg/proc/amd64_arch.go
+++ b/pkg/proc/amd64_arch.go
@@ -200,8 +200,16 @@ func amd64SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
 
 		return true
 
-	case "runtime.goexit", "runtime.rt0_go", "runtime.mcall":
+	case "runtime.goexit", "runtime.rt0_go":
 		// Look for "top of stack" functions.
+		it.atend = true
+		return true
+
+	case "runtime.mcall":
+		if it.systemstack && it.g != nil {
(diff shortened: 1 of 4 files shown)


pkg/proc/amd64_arch.go
```diff
@@ -200,11 +200,19 @@ func amd64SwitchStack(it *stackIterator, _ *op.DwarfRegisters) bool {
...
 		return true
 
-	case "runtime.goexit", "runtime.rt0_go", "runtime.mcall":
+	case "runtime.goexit", "runtime.rt0_go":
 		// Look for "top of stack" functions.
 		it.atend = true
 		return true
 
+	case "runtime.mcall":
+		if it.systemstack && it.g != nil && it.bi.GOOS == "freebsd" {
```
derekparker, line 209: This only happens on freebsd?

aarzilli: I haven't seen anywhere else, the same tests don't fail on other OSes in CI. System calls are done differently on different OSes, I'm not surprised by this.

derekparker: Just curious if we've just gotten lucky on other OSes or if there's cause for implementing this elsewhere. If not, happy to merge this for now.

derekparker (approved):
lgtm

PR #3460, opened 2023-08-10
*: Fix ppc64le test runs

derekparker:
We must pass the build tag through during the 'vet' check, and additionally there was some extra commands at the end of test_linux.sh that were not necessary.


aarzilli (approved):

aarzilli:
This is fine and we can merge it as it is but the ppc64le agent is still failing.

PR #3463, opened 2023-08-11
goversion,teamcity: update test matrix and fix build scripts

aarzilli:
Update test matrix to remove 1.18 and add 1.21, fix build scripts to
deal with the new format returned by:
https://golang.org/VERSION?m=text
which now has an extra second line with the time.


derekparker (approved):
lgtm

PR #3464, opened 2023-08-12
service/dap: use == to test io.EOF

gocurr:
The documentation of io.EOF:
Read must return EOF itself, not an error wrapping EOF, because callers will test for EOF using ==.

This is a trivial change; people may think it's normal use of "errors.Is", even it's OK, it could be
replaced with "errors.Is(err, io.EOF)" in idiomatic way.


aarzilli (approved):
LGTM

PR #3465, opened 2023-08-12
teamcity: hold back linux/386 builds to 1.20

aarzilli:
Programs built with `-gcflags='all=-N -l'` do not work on
linux/386/1.21 due to https://github.com/golang/go/issues/61975


derekparker (approved):
lgtm

PR #3467, opened 2023-08-14
proc: fix PIE support on macOS

aarzilli:
Go1.22 switched to emitting PIE by default and also changed some
details of the PIE implementation. This breaks delve entirely on macOS.

Fix how relocations are handled on macOS.


derekparker (approved):
lgtm

PR #3470, opened 2023-08-16
goversion: parse new version format and simplify version representation

aarzilli:
Add parsing for the new version format with toolchain (X.Y.Z-something)
and simplify internal representation of versions so that revision, beta
version and rc version are all represented on a single field with rc
and beta versions being negative numbers (this limits rc versions to a
maximum of 1000 which will not be a problem in practice).


derekparker (approved):
lgtm

PR #3476, opened 2023-08-22
teamcity: misc changes to CI

aarzilli:
* add architecture rule for ppc64le so that incompatible agents don't
pick up the build
* disable PIE tests on linux/ppc64le (the tests claim it doesn't work)
* enable PIE tests on darwin/amd64 now that the entry point calculation
has been fixed
* remove dependency on wget and curl in the test script for linux to
reduce test time
* only install git in the linux test script when we need it
* remove staticcheck from linux/ppc64le builds (it takes almost 5
minutes between installation and execution and makes the test timeout
sometimes)
* drop windows/arm64/tip build, the windows/arm64 build is broken
anyway and since there is only one agent it makes CI runs slow
* drop linux/ppc64le/tip build, there is only one agent, it's slow and
it will always timeout. CI runs in excess of 1h are too long.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -46,13 +46,11 @@ val targets = arrayOf(
         "linux/arm64/tip",
 
         "linux/ppc64le/1.21",
-        "linux/ppc64le/tip",
 
         "windows/amd64/1.21",
         "windows/amd64/tip",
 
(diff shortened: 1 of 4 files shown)


derekparker (approved):
lgtm

Issue #3206, opened 2022-11-28
DWARF unwinding (CFI) fixes


javierhonduco:
Hi!! First of all, thanks so much for Delve, been using it since I've started using Go and works incredibly well and fast :tada:.

Long story short, we've used some of Delve's code, in particular, the [DWARF CFI parser ](https://github.com/go-delve/delve/blob/d481eabab941a7baafbcd583be72fbc9d60442ed/pkg/dwarf/frame/table.go)which has been working great for us. We've modified this code to suit our needs better, and while working on this we noticed that some of the DWARF opcodes might not be handled correctly, in particular:
- **save/restore opcodes**, which don't restore the CFA appropriately;
- **DWARF expressions**, some of the constants are not correct;

I was wondering why this wasn't noticed before. As far as I understand Delve implements this to debug cgo code exclusively, is this correct? 

I am more than happy to submit PRs for both of these issues in the next few weeks, but first I wanted to bring this up to see if this would be something you would be interested in.

Thanks!!

cc/ @aarzilli @kakkoyun

javierhonduco:
Sounds good! Will aim to send the PRs in a few weeks :). 

For context, I work on [Parca Agent](https://github.com/parca-dev/parca-agent/) project, which is a continuous profiler written in Go which uses BPF to collect stack traces. As we speak I am working on a blog post that will be published tomorrow with all the details, but the gist of it is that currently there's no easy way to profile applications that were compiled without frame pointers. And as you are probably aware, this is a very common toggle that gets disabled by many distros and packagers.

To allow profiling without frame pointers we've developed a BPF-based unwinder that uses information extracted from dwarf unwinding tables. We use quite a bit of Delve's code to do this. Unfortunately, we had to fork it for a variety of reasons, but we plan to upstream all the improvements that we might make, such as the correctness patches I mentioned 😄 

> It's also used for Go but the subset used by Go is small and cgo testing is lacking.

Out of curiosity, do you happen to know when this is the case? I was under the impression that Go unwinding/stack walking could be done with `gopclntab`. Why is DWARF unwinding needed in non-cgo cases?

aarzilli:
> Out of curiosity, do you happen to know when this is the case? I was under the impression that Go unwinding/stack walking could be done with gopclntab. Why is DWARF unwinding needed in non-cgo cases?

[...]

PR #3480, opened 2023-08-23
DWARF CFI: use a stack for restore/remember opcodes

javierhonduco:
Implementing the `DW_CFA_remember_state` and `DW_CFA_restore_state` according to the DWARF specification requires us to create a stack that can store an arbitrary number of elements, that is, there could be multiple "pushes" before "popping" them.

From the 5th revision of the spec [0]:

> 6.4.2.4 Row State Instructions
> DW_CFA_remember_state
> 	The DW_CFA_remember_state instruction takes no operands. The required
>	action is to push the set of rules for every register onto an implicit stack.
> DW_CFA_restore_state
>	The DW_CFA_restore_state instruction takes no operands. The required action
>	is to pop the set of rules off the implicit stack and place them in the
>	current row.

0: https://dwarfstd.org/doc/DWARF5.pdf

Fixes: https://github.com/go-delve/delve/issues/3206

### DIFF
--- a/pkg/dwarf/frame/table.go
+++ b/pkg/dwarf/frame/table.go
@@ -18,18 +18,46 @@ type DWRule struct {
 
 // FrameContext wrapper of FDE context
 type FrameContext struct {
-	loc           uint64
-	order         binary.ByteOrder
-	address       uint64
-	CFA           DWRule
-	Regs          map[uint64]DWRule
-	initialRegs   map[uint64]DWRule
-	prevRegs      map[uint64]DWRule
-	buf           *bytes.Buffer
-	cie           *CommonInformationEntry
-	RetAddrReg    uint64
-	codeAlignment uint64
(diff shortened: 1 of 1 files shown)


pkg/dwarf/frame/table.go
```diff
@@ -18,18 +18,46 @@ type DWRule struct {
...
+	initialRegs     map[uint64]DWRule
+	buf             *bytes.Buffer
+	cie             *CommonInformationEntry
+	RetAddrReg      uint64
+	codeAlignment   uint64
+	dataAlignment   int64
+	rememberedState *stateStack
+}
+
+type RowState struct {
```
aarzilli, line 35: This should also not be exported.

aarzilli (approved):
LGTM

derekparker (approved):
lgtm

PR #3482, opened 2023-08-23
proc: disable "wait-for" tests on freebsd

aarzilli:
They are flaky in CI, I can not reproduce the problem locally.


derekparker (approved):
lgtm

PR #3483, opened 2023-08-23
*: reduce sleep timeout in test program

derekparker:
This larger timeout is not essential for the test and can be reduced to make full test runs quicker.


aarzilli:
Did you close this intentionally?

PR #3487, opened 2023-08-29
api,dap: fix hexadecimal printing of vars with symbolic const values

### LINKED ISSUE #3485: Inconsistent hex representation of typed constants with unsigned int type
1. What version of Delve are you using (`dlv version`)? 1.21.0
2. What version of Go are you using? (`go version`)? 1.21
3. What operating system and processor architecture are you using? Linux
4. What did you do? Debug the following program in VS Code, using Delve's DAP adapter:

```go
package main

import "fmt"

type A uint32

const (
	zero = A(iota)
	one
	two
)

func main() {
	a := zero
	b := one
	c := two
	d := A(3)
	e := A(4)
	fmt.Println(a, b, c, d, e)  // halt here
}
```

In particular, the program defines a new type with an underlying `unsigned int` type and a few constants of that type.

8. What did you see instead?

The hex representation of values of type `A` is inconsistent with the numeric value. This only happens for unsigned int types, and as long as the value can be represented as a binary OR of predefined constants of that type:
image

The expected representation should be:
```text
a: zero (0) = 0x0
b: one (1) = 0x1
c: two (2) = 0x2
d: one|two (3) = 0x3
e: 4 = 0x4
```

aarzilli:
Fix hexadecimal printing of variables that can be represented using
symbolic const values in DAP as well as the command line interface.

Fixes #3485

### DIFF
--- a/pkg/proc/variables_test.go
+++ b/pkg/proc/variables_test.go
@@ -1104,16 +1104,16 @@ func TestPackageRenames(t *testing.T) {
 
 func TestConstants(t *testing.T) {
 	testcases := []varTest{
-		{"a", true, "constTwo (2)", "", "main.ConstType", nil},
-		{"b", true, "constThree (3)", "", "main.ConstType", nil},
-		{"c", true, "bitZero|bitOne (3)", "", "main.BitFieldType", nil},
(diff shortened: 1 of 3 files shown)


derekparker (approved):
lgtm

PR #3492, opened 2023-08-31
service/test: disable TestClientServer_chanGoroutines with rr backend

aarzilli:
It doesn't work with rr because it uses runtime.Breakpoint to stop.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -3063,7 +3063,6 @@ var waitReasonStrings = [...]string{
 }
 
 func TestClientServer_chanGoroutines(t *testing.T) {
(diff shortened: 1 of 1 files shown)


derekparker:
Why does `runtime.Breakpoint` cause rr to fail?

aarzilli:
It has the same problem that the tests in pkg/proc/variables_test.go have. Under rr it stops inside runtime.Breakpoint because that's where the recording ends. We would need to do here what we do with `findFirstNonRuntimeFrame` in those tests, but I don't think it would add anything, this feature is entirely independent from the backend code.

derekparker (approved):
lgtm

PR #3496, opened 2023-09-07
service/dap: accept a string list as launch request's buildFlags

### LINKED ISSUE #2718: dap: consider []string for Launch BuildFlags
DAP [`LaunchConfig.BuildFlags`](https://github.com/go-delve/delve/blob/c7e2a9b6fb902d685e77b28b001c0c95e5eb95b3/service/dap/types.go#L101) is defined as `string` type.
It is `string` because VS Code Go defined the `buildFlags` in that way in its legacy debug adapter that invokes `dlv debug` or `dlv test`, and `dlv` command line takes the build flags (--build-flags) as a string.

DAP can take Launch configuration in JSON, which can support more rich types than a simple command line flag.
Historically, the subtle difference between how dlv is parsing the build flags and how go command is parsing the flags was annoying and sometimes frustrating (e.g. https://github.com/golang/vscode-go/issues/1027).

Please consider to take `[]string` and bypass delve's flag parsing.

hyangah:
This change accepts both string type and []string type (actually, []interface{} type due to Go's json decoding behavior).

Fixes #2718
For golang/vscode-go#1831, golang/vscode-go#1027

cc @suzmue

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -41,8 +41,11 @@ func GoBuild(debugname string, pkgs []string, buildflags string) error {
 
 // GoBuildCombinedOutput builds non-test files in 'pkgs' with the specified 'buildflags'
 // and writes the output at 'debugname'.
-func GoBuildCombinedOutput(debugname string, pkgs []string, buildflags string) (string, []byte, error) {
-	args := goBuildArgs(debugname, pkgs, buildflags, false)
+func GoBuildCombinedOutput(debugname string, pkgs []string, buildflags interface{}) (string, []byte, error) {
+	args, err := goBuildArgs2(debugname, pkgs, buildflags, false)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


aarzilli (approved):
LGTM

aarzilli:
Actually, it's making TestBadLaunchRequest fail.

hyangah:
Changed to detect the error from parsing time.
BTW I noticed TestBadAccess fails on my mac (intel, go1.21.1) - looks like the stopped reason description/format was changed, and sometimes it fails to get exception info. Hard to believe this is related to this PR though. Will see if the CI catches it.

aarzilli:
> looks like the stopped reason description/format was changed, and sometimes it fails to get exception info

Only sometimes, that's weird. Are you sure it's 1.21 and not 1.22?

aarzilli (approved):
LGTM

PR #3497, opened 2023-09-10
service/dap: add hidePackagePaths option

### LINKED ISSUE #3323: VSCode: Function Names not Visible in the Call Stack View
Originally reported as https://github.com/microsoft/vscode/issues/178964 but redirected here.

1. What version of Delve are you using (`dlv version`)? `1.20.1`
2. What version of Go are you using? (`go version`)? `go1.20.2`
3. What operating system and processor architecture are you using? **macOS Big Sur 11.7.3, amd64**
4. What did you do?
    1. Loaded [the `gopls` project](https://github.com/golang/tools/blob/master/gopls/) into VSCode
    2. Set a breakpoint
    3. Launched the `TestLSP` test
    4. Waited until the execution stopped at my breakpoint
6. What did you expect to see? **A stack trace with function names in it**, see video below
7. What did you see instead? **A stack trace with all the function names out of sight**, see video below

This video shows how wide I have to make the Call Stack View to see the relevant parts of the stack trace rows.

https://user-images.githubusercontent.com/158201/229415502-941c4bee-cc7f-4190-8599-58c9f538e08a.mov

stefanhaller:
This solves the problem that function names with long package paths are hard to read when the callstack window is narrow, because all you see is the beginning of the package path.

By default we change the display so that the function name is at the beginning of the line, and the package path appended in parenthesis. For example, instead of

```
  github.com/some/long/package/path/pkg.(*SomeType).AnotherMethod
```

we now display

```
  pkg.(*SomeType).AnotherMethod (in github.com/some/long/package/path/pkg)
```

On top of that, we provide an option hidePackagePaths, which, when true, causes the path in parenthesis to be omitted entirely, making it just

```
  pkg.(*SomeType).AnotherMethod
```

Fixes #3323.


aarzilli:
cc @suzmue @hyangah

service/dap/server.go
```diff
@@ -1680,11 +1685,19 @@ func (s *Session) onContinueRequest(request *dap.ContinueRequest, allowNextState
...
 	if loc.Fn == nil {
 		return "???"
 	}
-	return loc.Fn.Name
+	name := loc.Fn.Name
+	if idx := strings.LastIndex(name, "/"); idx != -1 {
+		if hidePackagePaths {
+			return name[idx+1:]
+		} else {
+			return fmt.Sprintf("%s (in %s)", name[idx+1:], name[:idx])
```
suzmue, line 1697: Could we make this be the whole package path (for `(in package/path)`)? so people don't have to refer back to the start of the string for the full path.

service/dap/types.go
```diff
@@ -167,6 +167,10 @@ type LaunchAttachCommonConfig struct {
 	// Default is 50.
 	StackTraceDepth int `json:"stackTraceDepth,omitempty"`
 
+	// Boolean value to indicate whether package paths should be dropped
+	// from function names in the call stack view.
```
suzmue, line 171: Replace "function names" with "stack frame names".

I think this is a little less confusing since we already moved the package path away from the stack frame.

suzmue (changes requested):
Thanks for the PR @stefanhaller! This is a definitely a step forward for viewing call stacks while debugging.

service/dap/server.go
```diff
@@ -1680,11 +1685,19 @@ func (s *Session) onContinueRequest(request *dap.ContinueRequest, allowNextState
...
 }
 
-func fnName(loc *proc.Location) string {
+func fnName(loc *proc.Location, hidePackagePaths bool) string {
 	if loc.Fn == nil {
 		return "???"
 	}
-	return loc.Fn.Name
+	name := loc.Fn.Name
+	if idx := strings.LastIndex(name, "/"); idx != -1 {
```
aarzilli, line 1693: This doesn't always work with functions that have type parameters (the last slash could be inside the type parameter list), the methods of proc.(*Functions) should be used instead.

service/dap/server.go
stefanhaller: Yep, fixed in ca8e87194a50. (Let me know if you prefer amending and force-pushing over fixup commits, not sure what the conventions around here are.)

service/dap/types.go
stefanhaller: Sure, see ddbf55d9b3c4.

service/dap/server.go
stefanhaller: Should be fixed in ca8e87194a50, please have a look. Do you think it's worth adding a fixture that allows us to test this? (I wouldn't mind if the answer is "no" 😄)

aarzilli: https://go.dev/play/p/-hZGE17iZxT

The symbol name for the instance of typeparametric function that is called is: `main.typeparametric[go.shape.struct { example.com/blah/reader.x int }]`

aarzilli:
> Here's a beginner's question about actually using the new config in VS Code. I confirmed that it works when I type `dlv config hidePackagePaths true` into the debug console in VS Code, but I can't find a way to enable it permanently. 

It should be read from launch.json.

service/dap/server.go
stefanhaller: I see, thanks. OK, I thought it was worth adding tests for this, so I added simple tests for just the `fnName` function: 99483f9ff687

stefanhaller:
> It should be read from launch.json.

Ah, nice. I thought this would only work after the golang extension is updated to support it (given that it shows a warning squiggle "Property hidePackagePaths is not allowed." But it does work already.

stefanhaller:
Seems that we prefer to strip the paths unconditionally without a config option, so I'm closing this and open #3500 instead.

PR #3498, opened 2023-09-12
Fix PIE tests on ppc64le port

archanaravindar:
PIE tests (tests on binaries built with PIE mode) on ppc64le port were having issues such as not matching expected patterns as in non PIE mode while stepping and not stepping into functions as expected creating errors due to which these tests were skipped on PIE mode 
These problems were occurring due to the special prologue emitted on ppc64le to deal with the TOC register or R2 
In some cases we had to modify the expected line numbers while stepping 
and in some other cases where the breakpoint was set in the prologue instruction even before function entry as a result of which the breakpoint was deemed void and the program control was not entering inside the function we had to modify delve to recognize these patterns to appropriately skip the special prologue and move the breakpoint a few instructions later so that it was set at the right location and the breakpoint was actually hit. 
Very soon we will add a change to the Go runtime as well to emit DWARF attributes to recognize such situations so that Delve can recognize this pattern with ease. 
For now the current change in Delve fixes the issue and the tests pass.

### DIFF
--- a/pkg/proc/ppc64le_disasm.go
+++ b/pkg/proc/ppc64le_disasm.go
@@ -22,7 +22,7 @@ func init() {
 	var unixGetG = opcodeSeq{uint64(ppc64asm.LD)}
 	var prologue opcodeSeq
 	prologuesPPC64LE = make([]opcodeSeq, 0, 3)
-	prologue = make(opcodeSeq,0,1)
+	prologue = make(opcodeSeq, 0, 1)
 	for _, getG := range []opcodeSeq{unixGetG} {
 		for _, stacksplit := range []opcodeSeq{tinyStacksplit, smallStacksplit, bigStacksplit} {
 			prologue = append(prologue, getG...)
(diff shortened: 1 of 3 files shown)


pkg/proc/ppc64le_disasm.go
```diff
@@ -17,17 +17,25 @@ func init() {
...
-			prologue := make(opcodeSeq, 0, len(getG)+len(stacksplit))
 			prologue = append(prologue, getG...)
 			prologue = append(prologue, stacksplit...)
 			prologuesPPC64LE = append(prologuesPPC64LE, prologue)
 		}
 	}
+	// On PIE mode special prologue is generated two instructions before the function entry point that correlates to call target 
+	// address, Teach delve to recognize this sequence to appropriately adjust PC address so that the breakpoint is actually hit
+	// while doing step
+			TOCprologue := make(opcodeSeq, 0, len(adjustTOCPrologueOnPIE))
```
derekparker, line 36: nit: please format with `gofmt`

aarzilli (commented):
The go compiler should emit an end of prologue flag in debug_line, this shouldn't matter at all.

derekparker:
> The go compiler should emit an end of prologue flag in debug_line, this shouldn't matter at all.

You're correct, it *should* but it isn't for this particular sequence. Needs to be fixed upstream.

aarzilli (approved):
Ok.

derekparker:
The ppc64le machine keeps timing out, but the `pkg/proc` tests passed according to the logs, so I say we merge.

derekparker (approved):
lgtm

aarzilli:
> The ppc64le machine keeps timing out, but the `pkg/proc` tests passed according to the logs, so I say we merge.

If you look at the logs you'll see that a lot of the time is actually spent downloading and installing updates as well as programs we need for the script (jq, gcc, etc). Maybe we should make a docker image that is already up to date and has the things we need installed.

PR #3499, opened 2023-09-15
pkg/terminal: add 'packages' command

hyangah:
This command lists the packages included in the debugee.
The implementation utilizes "ListPackagesBuildInfo" RPC.

In order to support server-side filtering like `sources` and other commands,
expands the ListPackagesBuildInfo RPC to take an optional filter field.


pkg/terminal/command.go
```diff
@@ -356,6 +356,11 @@ If regex is specified only the functions matching it will be returned.`},
 	types [<regex>]
 
 If regex is specified only the types matching it will be returned.`},
+		{aliases: []string{"packages"}, cmdFn: packages, helpMsg: `Print list of packages.
```
aarzilli, line 359: _scripts/gen-cli-docs.go needs to be rerun.

hyangah: Done.

aarzilli (approved):
LGTM

PR #3502, opened 2023-09-16
pkg/proc/internal/ebpf: remove redundant nil check

Juneezee:
From the Go specification:

> "1. For a nil slice, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary.


aarzilli (approved):
LGTM

derekparker (approved):
lgtm

PR #3510, opened 2023-09-23
service/dap: add the concrete type to interface type

suzmue:
As part of improvements to the variables view, add the concrete type information for interfaces to the Type field. This is the only information that would be lost by removing the types from the variable value, the rest is already duplicated in the Type field.

For golang/vscode-go#2569


suzmue:
cc @hyangah

hyangah (approved):

PR #3520, opened 2023-10-04
Support frame conds

### LINKED ISSUE #3515: breakpoint condition on other frames
It would be useful to have breakpoints that can join conditions from multiple frames, e.g.

```
break <locspec>
cond 1 -frame 2 id == 1
cond 1 -and -frame 1 subId == 0
```

where `-and` appends the new condition to the previous condition instead of overwriting it, and `-frame` indicates the frame in which the expression is evaluated.

right now the alternative would be to use two breakpoints and operate manually:

```
break 1 <locspec>
frame 2
break 2 0
cond 2 id == 1
cond 1 false
continue
clear 2
cond -clear 1
continue
```

derekparker:
This patch allows a user to set a breakpoint condition targeting a specific frame.

This is more of a partial implementation of the referenced issue because it does
not allow for setting multiple conditions. That, in my opinion, is a separate feature
request and can be implemented independently.

Fixes #3515


pkg/proc/breakpoints.go
```diff
@@ -447,6 +449,10 @@ func evalBreakpointCondition(tgt *Target, thread Thread, cond ast.Expr) (bool, e
 			return true, err
 		}
 	}
+	scope, err = ConvertEvalScope(tgt, scope.g.ID, int(frameCond), 0)
```
aarzilli, line 452: This is going to break if the thread isn't running a goroutine. This could be the reason for all the new test failures.

aarzilli (commented):
IMO the simplest way to do this would be to integrate it into the expression evaluator, something like `runtime.frame(N).a` gets you variable `a` from the N-th frame.

derekparker:
> IMO the simplest way to do this would be to integrate it into the expression evaluator, something like `runtime.frame(N).a` gets you variable `a` from the N-th frame.

That's interesting, I like that approach. That would mean landing this after your PR #3508. I can close this out for now and we can move forward with that approach.

PR #3522, opened 2023-10-06
service/dap: fix bugs in stdout/stderr handling

hyangah:
Fixes bugs introduced in v1.21.1

* Avoid dropping the last bytes from stderr/stdout when Read returns an error. (Read returns n>0). And skip sending Output event if Read returns n==0.

* Fix the bug that drops all stdout in the existing noDebug mode.

For #3094 #3253 

With this, the vscode-go test passes.
I wish we had testing for non-redirect mode, but I want to get this in soon since these bugs affect vscode go users.

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1050,6 +1050,15 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 			var out [1024]byte
 			for {
 				n, err := reader.Read(out[:])
+				if n > 0 {
+					outs := string(out[:n])
+					s.send(&dap.OutputEvent{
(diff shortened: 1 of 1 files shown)


hyangah:
Looks like teamcity builders had issues for some time. Is this a known problem?

aarzilli (approved):
LGTM

aarzilli:
> Looks like teamcity builders had issues for some time. Is this a known problem?

The builders for ppc and windows/arm64 are offline, the builders for mac/amd64 sometimes die and have to be restarted.

derekparker (approved):
lgtm

PR #3527, opened 2023-10-12
Use a valid timezone in tested binary

upils:
With recent changes to debian/ubuntu tzdata package, the Mexico/BajaSur is not valid anymore (not present by default). It is now present in the tzdata-legacy package, which is not a essential package. Without this timezone, the call to time.LoadLocation() will send a nil location, which will provoke a panic in time.ParseInLocation(). This change uses the underlying associated timezone America/Mazatlan.

Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052030


aarzilli (approved):
LGTM. Sad too see such a large landmass go.

PR #3528, opened 2023-10-15
proc: use DW_AT_trampoline to detect auto-generated code

aarzilli:
Use the trampoline attribute to detect auto-generated code. This fixes
a bug where stepping into a method of a generic type called through an
interface will take the debugger into an auto-generated wrapper that
does not have a dictionary and using next will step out of the wrapper.

Fixes a bug reported on the #delve channel of the gophers slack server.


derekparker (approved):
LGTM

PR #3532, opened 2023-10-20
proc: remove expr evaluation goroutine from EvalExpressionWithCalls

aarzilli:
We have used a goroutine to keep track of some of the expression
evaluation status across target resumes during call injections.

Now that the expression interpreter has been rewritten to use a stack
machine we can move what little state is left into the stack machine
and get rid of the goroutine-and-channel mechanism.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -9,7 +9,6 @@ import (
 	"go/constant"
 	"go/parser"
 	"go/printer"
-	"go/scanner"
 	"go/token"
 	"reflect"
 	"runtime/debug"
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM but let's merge #3534 first since I think it'll be easier to rebase this PR than that one.

aarzilli:
Ok (however I suspect they are fully independent and can be merged in any order)

derekparker:
> Ok (however I suspect they are fully independent and can be merged in any order)

You're right, went ahead and merged and there's no conflicts in the other PR.

PR #3534, opened 2023-10-20
proc: allow evaluator to reference previous frames

### LINKED ISSUE #3515: breakpoint condition on other frames

derekparker:
Fixes #3515

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -976,6 +976,10 @@ func (scope *EvalScope) executeOp(stack *evalStack, ops []evalop.Op, curthread T
 		if op.Frame != 0 {
 			var frameScope *EvalScope
 			frameScope, err = ConvertEvalScope(scope.target, scope.g.ID, int(op.Frame), 0)
+			if err != nil {
+				stack.err = err
+				return
+			}
 			vars, err = frameScope.Locals(0)
 		} else {
 			vars, err = scope.Locals(0)
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -971,6 +971,7 @@ func (scope *EvalScope) executeOp(stack *evalStack, ops []evalop.Op, curthread T
 		newscope, err := ConvertEvalScope(scope.target, scope.g.ID, int(op.Frame), 0)
 		if err != nil {
 			stack.err = err
+			return
 		}
 		*scope = *newscope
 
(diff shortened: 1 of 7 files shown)


pkg/proc/evalop/evalcompile.go
```diff
@@ -169,6 +169,31 @@ func (ctx *compileCtx) compileAST(t ast.Expr) error {
...
+				f, ok := ident.X.(*ast.Ident)
+				if ok && f.Name == "runtime" && ident.Sel.Name == "frame" {
+					switch arg := x.Args[0].(type) {
+					case *ast.BasicLit:
+						fr, err := strconv.ParseInt(arg.Value, 10, 8)
+						if err != nil {
+							return err
+						}
+						// Update scope to requested frame.
+						ctx.pushOp(&UpdateScopeFrame{Frame: fr})
```
aarzilli, line 184: I think it's better to add a `Frame` parameter to `PushLocal` and then do `if op.Frame != 0` in `case evalop.PushLocal`. To do a proper UpdateScopeFrame op it would also have to update spoff/bpoff/fboff (with the way it's used here it doesn't matter, because it's rolled back immediately, but I think we'll forget that it only works in specific circumstances).

derekparker: Good point, and a lot cleaner. Done.

pkg/proc/eval.go
```diff
@@ -971,7 +971,15 @@ func (scope *EvalScope) executeOp(stack *evalStack, ops []evalop.Op, curthread T
 		stack.push(newConstant(op.Value, scope.Mem))
 
 	case *evalop.PushLocal:
-		vars, err := scope.Locals(0)
+		var vars []*Variable
+		var err error
+		if op.Frame != 0 {
+			var frameScope *EvalScope
+			frameScope, err = ConvertEvalScope(scope.target, scope.g.ID, int(op.Frame), 0)
```
aarzilli, line 978: This err here is being ignored and frameScope on the next line could be nil.

derekparker: Fixed.

aarzilli (approved):
LGTM

PR #3535, opened 2023-10-20
Shorten variable types

### LINKED ISSUE #3516: dap: strip package paths from variable types
Now that we strip paths from function names in the stacktrace view (since #3500), I want the same for the types of variables in the variables view.

Here's a simple example. With a reasonably sized side bar, I can't see any variable names or values, only the beginnings of the paths:
image

This is how ridiculously wide I would have to make the side bar so that I can see variable names and values (impractical on a laptop screen):
image

This is what I would like to see instead:
image

I'd be interested in working on this, but it looks a lot less straighforward than #3500 was, so I'd need some guidance.

stefanhaller:
The variables view in VS Code is a lot easier to read if long type names are
shortened in much the same way as we shorten them for functions in the call
stack view.

We only shorten them in the value strings; the Type field of dap.Variable is
kept as is. Since this only appears in a tooltip, it isn't a problem to have the
full type visible there.

Fixes #3516.

### DIFF
--- a/service/api/prettyprint.go
+++ b/service/api/prettyprint.go
@@ -24,11 +24,13 @@ const (
 	prettyTop prettyFlags = 1 << iota
 	prettyNewlines
 	prettyIncludeType
+	prettyShortenType
 )
 
 func (flags prettyFlags) top() bool         { return flags&prettyTop != 0 }
 func (flags prettyFlags) includeType() bool { return flags&prettyIncludeType != 0 }
 func (flags prettyFlags) newlines() bool    { return flags&prettyNewlines != 0 }
+func (flags prettyFlags) shortenType() bool { return flags&prettyShortenType != 0 }
 
 func (flags prettyFlags) set(flag prettyFlags, v bool) prettyFlags {
 	if v {
@@ -42,6 +44,13 @@ func (flags prettyFlags) set(flag prettyFlags, v bool) prettyFlags {
 func (v *Variable) SinglelineString() string {
 	var buf bytes.Buffer
 	v.writeTo(&buf, prettyTop|prettyIncludeType, "", "")
+	return buf.String()
+}
+
+// SinglelineStringWithShortTypes returns a representation of v on a single line, with types shortened.
+func (v *Variable) SinglelineStringWithShortTypes() string {
+	var buf bytes.Buffer
+	v.writeTo(&buf, prettyTop|prettyIncludeType|prettyShortenType, "", "")
 	return buf.String()
 }
 
@@ -59,6 +68,14 @@ func (v *Variable) MultilineString(indent, fmtstr string) string {
 	return buf.String()
 }
 
+func (v *Variable) typeStr(flags prettyFlags) string {
+	if flags.shortenType() {
+		return ShortenType(v.Type)
+	}
+
+	return v.Type
+}
+
 func (v *Variable) writeTo(buf io.Writer, flags prettyFlags, indent, fmtstr string) {
 	if v.Unreadable != "" {
 		fmt.Fprintf(buf, "(unreadable %s)", v.Unreadable)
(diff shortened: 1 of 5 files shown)


service/api/prettyprint.go
```diff
@@ -21,33 +21,48 @@ const (
 // SinglelineString returns a representation of v on a single line.
 func (v *Variable) SinglelineString() string {
 	var buf bytes.Buffer
-	v.writeTo(&buf, true, false, true, "", "")
+	v.writeTo(&buf, true, false, true, false, "", "")
+	return buf.String()
+}
+
+// SinglelineStringWithShortTypes returns a representation of v on a single line, with types shortened.
+func (v *Variable) SinglelineStringWithShortTypes() string {
```
stefanhaller, line 29: I decided to add a new function instead of adding a flag to `SinglelineString`, as that function has tons of call sites that I would have had to adapt.

service/api/prettyprint.go
```diff
@@ -21,33 +21,48 @@ const (
...
-func (v *Variable) writeTo(buf io.Writer, top, newlines, includeType bool, indent, fmtstr string) {
+func (v *Variable) typeStr(shortenType bool) string {
+	if shortenType {
+		return ShortenType(v.Type)
+	}
+
+	return v.Type
+}
+
+func (v *Variable) writeTo(buf io.Writer, top, newlines, includeType, shortenType bool, indent, fmtstr string) {
```
stefanhaller, line 57: I also considered turning the `includeType` bool into an enum with three values (off, short, full). That didn't work well though, because of the way includeType is negated sometimes.

stefanhaller (commented):
I haven't had a lot of time yet to test this thoroughly, but a cursory test makes this look very promising.

Putting it out there so that others can play with it. I'll be on vacation for two weeks now.

service/api/prettyprint.go
aarzilli: SinglelineStringFormatted as very few calls.

aarzilli (commented):
There isn't going to be any kind of configuration to turn this off?

service/api/prettyprint.go
aarzilli: I think having four booleans on a row crosses a threshold. If @derekparker agrees I would rebase this on something like this https://github.com/go-delve/delve/pull/3536

stefanhaller:
> There isn't going to be any kind of configuration to turn this off?

I don't think it needs to be configurable, I really can't imagine why anybody would want to see the long types. We didn't make #3500 configurable either, and I think this is very similar. Also, you can see the full type in a tooltip by hovering over a variable name.

service/api/prettyprint.go
stefanhaller: That would be fine with me. I suppose I'd wait until that PR is merged and then rebase on master? There's no rush, I'm now on vacation for two weeks, so can only continue with this after I'm back anyway.

aarzilli:
> why anybody would want to see the long types

Because short ones could be ambiguous. But if they are visible on hover it's probably ok.

service/api/prettyprint.go
stefanhaller: Do you mean I should use `SinglelineStringFormatted` instead of adding `SinglelineStringWithShortTypes`? If so, should I change its behavior to always shorten types, or add a bool parameter to it to make it configurable? Personally I'd find it preferable to add the new function, as I don't need the ability to pass a format string.

stefanhaller:
Sorry for the delay, I was on vacation.

I rebased this onto master, I'm quite happy about how much cleaner the change is now. Can you please have another look?

aarzilli (approved):
LGTM

aarzilli:
cc @hyangah @suzmue you wanted to review this I think.

suzmue (approved):
LGTM. Thanks @stefanhaller!!

PR #3536, opened 2023-10-22
service/api: use bitfield for prettyprint flags

aarzilli:
Use a bitfield for the internal flags used by the prettyprinting code.


derekparker (approved):
LGTM

PR #3540, opened 2023-10-26
*: Add explicit code of conduct

derekparker:
Adds a callout to the standard Go code of conduct in the CONTRIBUTING.md doc.


aarzilli:
I would have a few reservations about this but given that this is unlikely to ever matter in practice I'll just merge it.

aarzilli (approved):

PR #3541, opened 2023-10-28
proc,service/dap,proc/gdbserial: fixes for debugserver --unmask-signals

aarzilli:
* fix TestRefreshCurThreadSelGAfterContinueOnceError and TestBadAccess
to work when debugserver has --unmask-signals
* when a fatal signal is received while singlestepping delay its
delivery until the subsequent continue, otherwise debugserver will get
stuck completely (fixes TestNilPtrDerefInBreakInstr)


derekparker (approved):
lgtm

PR #3547, opened 2023-10-31
proc/native: wherever we check for exited we should also check detached

aarzilli:
Only Linux and Windows are affected but it's better to do this for all
platforms for consistency.


derekparker (approved):
lgtm

PR #3549, opened 2023-11-01
pkg/proc: add inline function support for stripped binaries

derekparker:
This patch adds support for listing and setting breakpoints on inlined functions within stripped binaries. It uses a forked version of `debug/gosym` copied from golang.org/x/vuln/internal/vulncheck/internal/gosym which adds support for parsing the inline tree of the pclntab section. Parsing this section requires knowing the offset of the "go:func.*" symbol, which is not present in stripped binaries via the `.symtab` section so instead, we search the `.noptrdata` section which contains `runtime.moduledatap` which contains the value of that missing symbol, which we then can use to find the inline tree for a given function.

Given all this we parse the inline tree for each function we find, and then add that information the the appropriate `Function` contained in `bi.Functions`, using a relatively empty `Function` struct as what would be the abstract origin.

The forked version of `debug/gosym` is in an internal package because it is expected that this functionality will be exposed in the future and this package will be removed. Making it internal prevents anyone from relying on it outside this project, allowing us to remove it at will.

Currently this only supports Go 1.20+ 64-bit targets, however follow up commits can add module data support for older Go versions and 32-bit support. The parsing of the moduledata was taken and adapted from https://github.com/goretk/gore. It is my intention to contribute the necessary changes Delve requires and then simply use that as a dependency, so it is also internal.

Finally, thanks to my colleague @archanaravindar whom I worked together with on this feature.

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -12,17 +12,17 @@ Tests skipped by each supported backend:
 	* 2 broken - cgo stacktraces
 * darwin/lldb skipped = 1
 	* 1 upstream issue
-* freebsd skipped = 9
+* freebsd skipped = 10
 	* 2 flaky
 	* 2 follow exec not implemented on freebsd
 	* 4 not implemented
-	* 1 not working on freebsd
-* linux/386/pie skipped = 2
+	* 2 not working on freebsd
+* linux/386/pie skipped = 3
 	* 1 broken
-	* 1 not working on linux/386 with PIE
-* linux/ppc64le skipped = 2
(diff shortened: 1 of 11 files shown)


aarzilli:
We are still testing 1.19 so the tests need an explicit Skip for that, also it seems to have problems with 386.

aarzilli (approved):
I wonder if it will be feasible to keep this in sync.

derekparker:
> I wonder if it will be feasible to keep this in sync.

The plan is to have this information exposed in a more supported way in stripped binaries in the future.

PR #3550, opened 2023-11-02
cmd: fix a bunch of linter warnings from GoLand

derekparker:
Also fixes a potential nil pointer deref error in a test.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -180,7 +180,7 @@ func testOutput(t *testing.T, dlvbin, output string, delveCmds []string) (stdout
 
 	// ignore "dlv debug" command error, it returns
 	// errors even after successful debug session.
-	_ = cmd.Wait()
+	cmd.Wait()
 	stdout, stderr = stdoutBuf.Bytes(), stderrBuf.Bytes()
 
 	_, err = os.Stat(debugbin)
@@ -308,7 +308,7 @@ func TestRedirect(t *testing.T) {
 
 	// and detach from and kill the headless instance
 	client := rpc2.NewClient(listenAddr)
-	_ = client.Detach(true)
+	client.Detach(true)
 	cmd.Wait()
 }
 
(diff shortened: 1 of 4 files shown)


cmd/dlv/dlv_test.go
```diff
@@ -180,17 +180,17 @@ func testOutput(t *testing.T, dlvbin, output string, delveCmds []string) (stdout
 
 	// ignore "dlv debug" command error, it returns
 	// errors even after successful debug session.
-	cmd.Wait()
+	_ = cmd.Wait()
```
aarzilli, line 183: Let's not do this, it's everywhere and it looks ugly.

cmd/dlv/dlv_test.go
```diff
@@ -180,17 +180,17 @@ func testOutput(t *testing.T, dlvbin, output string, delveCmds []string) (stdout
...
 	_, err = os.Stat(debugbin)
 	if err == nil {
 		// Sometimes delve on Windows can't remove the built binary before
 		// exiting and gets an "Access is denied" error when trying.
-		// See: https://travis-ci.com/go-delve/delve/jobs/296325131)
+		// See: https://travis-ci.com/go-delve/delve/jobs/296325131.
 		// We have added a delay to gobuild.Remove, but to avoid any test
 		// flakiness, we guard against this failure here as well.
-		if runtime.GOOS != "windows" || !strings.Contains(err.Error(), "Access is denied") {
+		if runtime.GOOS != "windows" {
```
aarzilli, line 193: But we do want to complain for errors other than access is denied.

cmd/dlv/main.go
```diff
@@ -15,11 +15,16 @@ func main() {
 	if Build != "" {
 		version.DelveVersion.Build = Build
 	}
+
 	const cgoCflagsEnv = "CGO_CFLAGS"
 	if os.Getenv(cgoCflagsEnv) == "" {
-		os.Setenv(cgoCflagsEnv, "-O0 -g")
+		err := os.Setenv(cgoCflagsEnv, "-O0 -g")
+		if err != nil {
```
aarzilli, line 22: os.Setenv can't fail except for a malformed key.

cmd/dlv/dlv_test.go
derekparker: Hah, agreed, automated change. Removed.

cmd/dlv/dlv_test.go
derekparker: The enclosing branch checks that `err == nil`.

aarzilli: Ah ok, you are right.

cmd/dlv/main.go
derekparker: True, reverted.

aarzilli (approved):

PR #3551, opened 2023-11-02
service: fix a bunch of linter warnings from GoLand

derekparker:


service/dap/server.go
```diff
@@ -516,28 +516,35 @@ func (s *Session) address() string {
 func (s *Session) ServeDAPCodec() {
 	// Close conn, but not the debugger in case we are in AcceptMulti mode.
 	// If not, debugger will be shut down in Stop().
+	var (
+		err     error
+		request dap.Message
+	)
 	defer s.conn.Close()
+	defer func() {
```
aarzilli, line 524: I don't think this refactoring is actually a refactoring. The original code only does this after checking `s.config.StopTriggered`.

service/dap/server.go
```diff
@@ -606,10 +613,10 @@ func (s *Session) handleRequest(request dap.Message) {
 	case *dap.PauseRequest: // Required
 		s.onPauseRequest(request)
 		return
-	case *dap.TerminateRequest: // Optional (capability ‘supportsTerminateRequest‘)
```
aarzilli, line 609: I guess GoLand doesn't like unicode in comments? I'm ok with this, I just think it's a weird thing to complain about for a linter.

service/dap/server.go
```diff
@@ -379,7 +379,7 @@ func (s *Server) Stop() {
 
 	if s.listener != nil {
 		// If run goroutine is blocked on accept, this will unblock it.
-		_ = s.listener.Close()
```
aarzilli, line 382: Same as the other PR.

service/debugger/debugger.go
```diff
@@ -546,7 +546,7 @@ func (d *Debugger) Restart(rerecord bool, pos string, resetArgs bool, newArgs []
 		return nil, fmt.Errorf("could not launch process: %s", err)
 	}
 
-	discarded := []api.DiscardedBreakpoint{}
```
aarzilli, line 549: I don't like this level of style enforcement. This seems too arbitrary to me.

service/dap/server.go
derekparker: This is actually removing the `_ = ` not adding it.

service/dap/server.go
derekparker: You're right, we should add a sentinel value to determine if we should call this, good catch. I just wanted to avoid a `defer` call within the loop, especially because that loop has a `continue` which could cause a leak.

service/dap/server.go
derekparker: This was me actually, I think it's just cleaner to have uniform quotes.

service/debugger/debugger.go
derekparker: That's fine, we don't have to enforce it going forward to new PRs or anything, it's just a bit of cleanup. Changes things somewhat semantically, but nothing that has any real effect here.

service/dap/server.go
derekparker: Done.

aarzilli (approved):

PR #3552, opened 2023-11-02
*: update dependencies

derekparker:


aarzilli:
We are still supporting 1.19 which this makes fail.

derekparker:
> We are still supporting 1.19 which this makes fail.

Reverted the `cilium/ebpf` update which was causing 1.19 to fail.

derekparker:
> Is this going to increase the effective minimum version of go we support? It's 1.17 now, is it going to make it be 1.18? I don't think there's sufficient upside here to override that downside.

No, the project still builds with 1.17 after this update.

However, I do think we should update our Go version to be equal to the lowest Go version we support in CI. I don't believe that arbitrarily limiting ourselves to 1.17 feature set is the right thing to do.

aarzilli:
We used to get complaints when we stopped building on older versions of go. We haven't recently and maybe it won't be a problem at all in the future with the new backwards/forward compatibility stuff in go. I'm also not saying we shouldn't move on from 1.17 for no reason, rather that we should have a reason for doing so.

aarzilli (approved):

PR #3553, opened 2023-11-03
proc: add regression test for issue #3548

### LINKED ISSUE #3548: E08 during memory read for packet

If this is about source listing not showing up (or breakpoints not being
accepted) while running in a container please read our
[FAQ](https://github.com/go-delve/delve/blob/master/Documentation/faq.md#substpath)
first.

```
❯ dlv version
Delve Debugger
Version: 1.21.2
```
```
❯ go version
go version go1.21.1 darwin/amd64
```
```
macOS Sonoma 14.1 amd64
```
4. What did you do?
When trying to debug a test that contains a struct with a non-pointer object that has a function with a pointer receiver defined, interacting with that object causes all locals to return `unreadable could not read string pointer protocol error E08 during memory read for packet`.

Steps:
```
b TestHi:13
c
p c
```
I expected to able to print values for locals.
6. What did you see instead?
I see all values being returned as error E08.

Log file and files to reproduce error are [here](https://gist.github.com/dhoizner/d728ca57bfd3a4cf1ebc287bc96ce8f7).

Changing `Thing` to be `*Thing`  on [line 11](https://gist.github.com/dhoizner/d728ca57bfd3a4cf1ebc287bc96ce8f7#file-foo_test-go-L11) resolves the error.

aarzilli:
Issue #3548 describes a bug in the compiler which was fixed by commit
505e50b. But this case wasn't covered by our current tests (obviously)
and the fix in the compiler looks accidental so it's worth adding a
test for it.

Fixes #3548

### DIFF
--- a/pkg/proc/variables_test.go
+++ b/pkg/proc/variables_test.go
@@ -1722,3 +1722,23 @@ func TestBadUnsafePtr(t *testing.T) {
 		}
 	})
 }
+
+func TestCapturedVariable(t *testing.T) {
+	// Checks that variables captured by a closure (that are not pointers) are
+	// readable. See issue #3548.
+	// This was broken in Go 1.21 due to a compiler bug.
+	if goversion.VersionAfterOrEqual(runtime.Version(), 1, 21) && !goversion.VersionAfterOrEqual(runtime.Version(), 1, 22) {
+		t.Skip("broken")
+	}
+	withTestProcess("issue3548", t, func(p *proc.Target, grp *proc.TargetGroup, fixture protest.Fixture) {
+		setFileBreakpoint(p, t, fixture.Source, 32)
+		assertNoError(grp.Continue(), t, "Continue()")
+		v := evalVariable(p, t, "c")
+		assertVariable(t, v, varTest{
+			name:         "c",
+			preserveName: true,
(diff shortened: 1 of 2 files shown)


pkg/proc/variables_test.go
```diff
@@ -1722,3 +1722,20 @@ func TestBadUnsafePtr(t *testing.T) {
 		}
 	})
 }
+
+func TestCapturedVariable(t *testing.T) {
+	if !goversion.VersionAfterOrEqual(runtime.Version(), 1, 22) {
+		t.Skip("broken")
```
derekparker, line 1728: Can you add more context here, like `broken in versions prior to 1.22 due to a compiler bug`. Also, does this affect all versions < 1.22 or was the bug introduced in 1.21 and we didn't catch it? If the latter, we should be more specific in the skip.

Finally, can you add a link to the Github issue in a comment somewhere for more context.

aarzilli: Done. I hadn't checked before but it really is a regression in 1.21.

derekparker (approved):
LGTM

balleriniGsped:
sill facing this bug with GO 1.25.1 on darwin/amd64, se issue #3548

PR #3560, opened 2023-11-09
TeamCity: reupgrade linux/386 builder to Go 1.21

aarzilli:
The bug that prevented updating to linux/386 has been fixed upstream,
update linux/386 to 1.21 in the test matrix and fix some broken bugs.


derekparker (approved):
lgtm

PR #3568, opened 2023-11-16
*: remove checks for TRAVIS env variable

aarzilli:
We had a few checks left over for the TRAVIS variable that detected if
we were running under the TravisCI build system.

This variable hasn't been set since 2020.


derekparker (approved):
lgtm

PR #3570, opened 2023-11-17
proc/gdbserver: clean up rr directory on detach

aarzilli:
We used to autoremove the trace recorded by rr but as a result of
various refactorings done to implement follow exec mode this broke.

Restore the functionality.

Also remove the _fixtures/testfnpos.go file which is autogenerated
during testing.


derekparker:
Reflecting on this, it seems like maybe the best user experience would be to add a prompt on detach on whether or not to clean up the trace dir (defaulting to true). It could be to the users benefit to keep that recording around for future debugging.

aarzilli:
> Reflecting on this, it seems like maybe the best user experience would be to add a prompt on detach on whether or not to clean up the trace dir (defaulting to true). It could be to the users benefit to keep that recording around for future debugging.

If we do that we also have to save the executable since the recording is useless without it.

derekparker:
> > Reflecting on this, it seems like maybe the best user experience would be to add a prompt on detach on whether or not to clean up the trace dir (defaulting to true). It could be to the users benefit to keep that recording around for future debugging.
> 
> If we do that we also have to save the executable since the recording is useless without it.

That's true. What are your thoughts on conditionally keeping both around?

aarzilli:
My opinion is that it isn't worth it, there's other ways to permanently make a recording and in an emergency you could always make a copy of both manually if you really needed to. But I don't feel strongly about it.

derekparker (approved):
LGTM

PR #3571, opened 2023-11-17
proc: simplify and generalize runtime.mallocgc workaround

aarzilli:
Instead of having a different version for each architecture have a
single version that uses an architecture specific list of registers.

Also generalize it so that, if we want, we can extend the workaround to
other runtime functions we might want to call (for example the channel
send/receive functions).


derekparker (approved):
lgtm

PR #3572, opened 2023-11-17
TeamCity: remove windows/arm64 builders from chain

aarzilli:
We haven't had a builder for 2 months so it can be declared legally
dead.


derekparker (approved):
lgtm

PR #3581, opened 2023-11-23
update TeamCity DSL version to 2023.05 and remove tip configurations from Aggregator

artspb:
It's not strictly necessary but doesn't seem to hurt, either. TeamCity suggests doing it for now, but in the future, it might require it.


aarzilli (approved):
LGTM

aarzilli:
Am I mistaken or is the aggregator always going to fail like this because the tip builds are disabled?

artspb:
Hm, I was under the impression that I removed tip builds from the aggregator. I'll take a look.

PR #3583, opened 2023-11-24
tests: fix tests in go1.22

aarzilli:
Go1.22 has changed some line number assignments. The new line number
assignments are still valid however some tests in dap relied on them
being different and broke as a result. This commit fixes those tests
and makes them less brittle.

Also disables TestDebugStripped and TestDebugStripped2 temporarily on
1.22.


derekparker (approved):
lgtm

PR #3584, opened 2023-11-25
service/dap: support waitfor option for 'dap attach' only

muggle-nil:
add a waitfor option for 'dap attach' that waits for a process with a given name to appear before attaching to it.

### DIFF
--- a/pkg/proc/native/proc_freebsd.go
+++ b/pkg/proc/native/proc_freebsd.go
@@ -181,7 +181,7 @@ func waitForSearchProcess(pfx string, seen map[int]struct{}) (int, error) {
 		seen[int(proc.ki_pid)] = struct{}{}
 
 		argv := strings.Join(getCmdLineInternal(ps, proc), " ")
-		log.Debugf("waitfor: new process %q", argv)
+		log.Debugf("waitfor: new process:\n\tpid=%d\n\tcmdline=%q", int(proc.ki_pid), argv)
 		if strings.HasPrefix(argv, pfx) {
 			return int(proc.ki_pid), nil
 		}
--- a/pkg/proc/native/proc_linux.go
+++ b/pkg/proc/native/proc_linux.go
@@ -216,7 +216,7 @@ func waitForSearchProcess(pfx string, seen map[int]struct{}) (int, error) {
 				buf[i] = ' '
 			}
 		}
-		log.Debugf("waitfor: new process %q", string(buf))
+		log.Debugf("waitfor: new process:\n\tpid=%d\n\tcmdline=%q", pid, string(buf))
 		if strings.HasPrefix(string(buf), pfx) {
 			return pid, nil
 		}
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -5891,6 +5891,48 @@ func TestAttachRequest(t *testing.T) {
 	})
 }
 
+func TestAttachWaitForRequest(t *testing.T) {
+	if runtime.GOOS == "freebsd" {
+		// The value of /proc/[pid]/cmdline might be wrong on FreeBSD: zero or the same as the parent process.
+		t.Skip("test skipped on freebsd")
+	}
+	runTest(t, "loopprog", func(client *daptest.Client, fixture protest.Fixture) {
+		var cmd *exec.Cmd
+		defer func() {
+			if cmd != nil {
+				// prevent the process from becoming a zombie after the waitfor fails.
+				_ = cmd.Process.Kill()
+			}
+		}()
+		runDebugSessionWithBPs(t, client, "attach",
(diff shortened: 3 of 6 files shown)


aarzilli:
We are not going to add this amount of dependencies for this feature.

pkg/proc/native/proc_linux.go
```diff
@@ -177,48 +177,29 @@ func Attach(pid int, waitFor *proc.WaitFor, debugInfoDirs []string) (*proc.Targe
...
-			return false
-		}
-	}
-	return true
-}
-
 func waitForSearchProcess(pfx string, seen map[int]struct{}) (int, error) {
 	log := logflags.DebuggerLogger()
-	des, err := os.ReadDir("/proc")
+	pids, err := process.Pids()
```
aarzilli, line 182: Why do you think this function needs to be reimplemented? Note that it is already tested (TestWaitFor in pkg/proc/proc_test.go) and works.

muggle-nil: In some cases, cmdline may be empty, such as a zombie process, so I want to use this library to help us handle various exceptional cases.
After reading the relevant code, I found that it doesn't need to be so complicated, I have already reverted it.

muggle-nil: image

aarzilli (approved):
LGTM
cc @suzmue @hyangah

service/dap/types.go
```diff
@@ -242,6 +242,15 @@ type AttachConfig struct {
 	// The numeric ID of the process to be debugged. Required and must not be 0.
 	ProcessID int `json:"processId,omitempty"`
 
+	// Wait for a process with a name beginning with this prefix.
+	AttachWaitFor string `json:"waitFor,omitempty"`
+
+	// Interval between checks of the process list, in millisecond.
```
hyangah, line 248: 1. Is it possible to come up with good default values for AttachWaitForInterval and AttachWaitForDuration, and not exposing them in the args yet? I am not sure about the usefulness of `AttachWaitForDuration` param in IDE UX yet, and I don't know when and why users are supposed to tweak AttachWaitForInterval. How about waiting until we find they are needed?

2. If we need them, can we use time.Duration instead of float64? And also comment what are the default values.
(I see these are from waitfor-interval and waitfor-duration flags, but wonder why they weren't using flag.Duration and got converted internally)

pkg/proc/native/proc_freebsd.go
```diff
@@ -181,7 +181,7 @@ func waitForSearchProcess(pfx string, seen map[int]struct{}) (int, error) {
 		seen[int(proc.ki_pid)] = struct{}{}
 
 		argv := strings.Join(getCmdLineInternal(ps, proc), " ")
-		log.Debugf("waitfor: new process %q", argv)
+		log.Debugf("waitfor: new process:\n\tpid=%s\n\tcmdline=%q", proc.ki_pid, argv)
```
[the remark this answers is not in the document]
muggle-nil: no

service/dap/types.go
muggle-nil: I agree with your point of view, not to expose these attributes

service/dap/types.go
```diff
@@ -242,6 +242,15 @@ type AttachConfig struct {
 	// The numeric ID of the process to be debugged. Required and must not be 0.
```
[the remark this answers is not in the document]
muggle-nil: I will add a comment and exception prompts.

hyangah (approved):

aarzilli:
The failure has nothing to do with git. The tests for service/dap are timing out, likely because of the changes in this PR.

service/dap/server_test.go
```diff
@@ -5813,6 +5813,44 @@ func TestAttachRequest(t *testing.T) {
 	})
 }
 
+func TestAttachWaitForRequest(t *testing.T) {
```
aarzilli, line 5816: This test should be skipped on freebsd.

service/dap/server_test.go
```diff
@@ -5891,6 +5891,48 @@ func TestAttachRequest(t *testing.T) {
...
+		runDebugSessionWithBPs(t, client, "attach",
+			// Attach
+			func() {
+				client.AttachRequest(map[string]interface{}{
+					"mode":        "local",
+					"waitFor":     fixture.Path,
+					"stopOnEntry": false,
+				})
+				// Start the program to attach to
+				cmd = execFixture(t, fixture)
```
aarzilli, line 5916: This is a race condition, unfortunately. Debugserver's waitfor feature ignores already existing processes, so if the AttachRequest arrives after loopprog is executed it will skip it. That is why sometimes the tests are timing out on macos.

PR #3585, opened 2023-11-27
proc: fix TestIssue1101 flake

aarzilli:
If the target process exits and receives a signal at the same time and
we receive the signal first we should call waitpid again to read the
target's exit status.

This also fixes a nil pointer dereference when trapWaitInternal returns
an error, this fix is probably incomplete but I wasn't able to
reproduce its circumstances after 30000 runs of TestIssue1101 to
properly address it.


derekparker (approved):
lgtm

PR #3589, opened 2023-11-28
all: run go mod tidy

### LINKED ISSUE #3588: Minimum Go version has changed to 1.21?
As noted [on Mastodon](https://mastodon.social/@cks/111484989725312650) when running with Go 1.21, Go will not allow building with `-mod=readonly`, and will insist you use `go mod tidy`. 

As per a [follow up comment](https://www.jvt.me/mf2/2023/11/1asep/) this appears to be due to one of our dependencies, `github.com/goretk/gore`, expecting a minimum of Go 1.21.

prattmic:
Fixes #3588.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/go-delve/delve
 
-go 1.17
+go 1.21
 
 require (
 	github.com/cilium/ebpf v0.11.0


derekparker (approved):
LGTM

This is fine for now since it won't actually prevent older toolchains from building Delve since it's only been since Go 1.21+ that the `go` directive became a mandatory requirement as opposed to a suggestion. However, Delve supports the latest version of Go as well as the last 2 versions (e.g. currently we support back to 1.19). Going forward we will have to be more mindful of dependency `go` directives and ensure we don't pull anything that breaks a supported build (CI will catch that) and potentially fork any dependency which sets a higher `go` version.

PR #3594, opened 2023-12-07
Support to add a new suboption --follow-calls to trace subcommand 

archanaravindar:
Support for an option to the trace subcommand, "--follow-calls <n>" which traverses the children of the functions being traced upto the desired depth
specified by the user.
n=0 is the same as the plain trace subcommand 
n=1 is print the trace of functions at level=1 which is again equivalent to the plain trace subcommand
n=2 is print the trace of functions at level=2 which prints the immediate child of the function being traced along with the function 
n=3 is print the trace of functions at level=3 which prints the function being traced along with two levels of children down the call tree 
and so on.. 

This option also indents the output according to depth for the calls occurring in the program
Calls such as runtime.morestack_noctxt are treated specially and do not adhere to the definition of depth like the other functions 
Though we can also trace morestack_noctxt calls if desired 

For example, if we have a simple .go program such as the following
```
package main
import "fmt"
func D(i int) int {
        return i*i*i
}
func C(i int) int {

        return i+20
}
func B(i int) int {
        d:=C(i)+40
        return d+D(i)
}
func A(i int) int {
        return 10+B(i)
}
func main() {
        j := 0
        j += A(2)
        fmt.Println(j)
}
```
building leaf4 and running delve as follows would
give the following trace output
```
go build -gcflags '-N -l' ./leaf4.go 


../dlv trace main.A --follow-calls 3 -e=./leaf4^M
> goroutine(1): runtime.morestack_noctxt()^M
> goroutine(1): main.A(2)^M
 > goroutine(1): main.B(2)^M
  > goroutine(1): main.C(2)^M
  >> goroutine(1): => (22)^M
  > goroutine(1): main.D(2)^M
  >> goroutine(1): => (8)^M
 >> goroutine(1): => (70)^M
>> goroutine(1): => (80)^M
> goroutine(1): runtime.morestack_noctxt()^M
80^M
```


cmd/dlv/cmds/commands.go
```diff
@@ -644,6 +646,9 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
 
 			processArgs = append([]string{debugname}, targetArgs...)
 		}
+		if dlvArgsLen >= 2 && traceFollowCalls <= 1 {
+			fmt.Fprintln(os.Stderr, "Need to specify a trace depth of atleast 1")
```
aarzilli, line 650: Shouldn't we call os.Exit(1) here?

cmd/dlv/dlv_test.go
```diff
@@ -941,6 +941,175 @@ func TestTrace2(t *testing.T) {
 	assertNoError(cmd.Wait(), t, "cmd.Wait()")
 }
 
+func TestTraceDepth(t *testing.T) {
```
aarzilli, line 944: I think it would be better to just have one of these end to end tests and then test the rest of the detailis in service/test/integration2_test.go by just calling ListFunctions directly.

pkg/terminal/command.go
```diff
@@ -2845,12 +2845,25 @@ func printBreakpointInfo(t *Term, th *api.Thread, tracepointOnNewline bool) {
...
+	depth = make(map[int64]int)
+)
+
 func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, args string, hasReturnValue bool) {
 	if t.conf.TraceShowTimestamp {
 		fmt.Fprintf(t.stdout, "%s ", time.Now().Format(time.RFC3339Nano))
 	}
 
 	if th.Breakpoint.Tracepoint {
+		idx := strings.Index(fn.Name(), "runtime.morestack_noctxt")
```
aarzilli, line 2858: This code isn't going to work correctly if someone sets a tracepoint manually (depth will never be decremented). It will also not work if one of the traced functions doesn't return but panics instead, and the panic is caught by a recover higher on the stack.

service/debugger/debugger.go
```diff
@@ -1467,6 +1467,65 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
 	return funcs, nil
 }
 
+func traverse(t proc.ValidTargets, f *proc.Function, depth int, FollowCalls int) ([]string, error) {
```
aarzilli, line 1470: You should keep track of which functions have been visited already and avoid revisiting them.

service/debugger/debugger.go
```diff
@@ -1467,6 +1467,65 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+		return nil, nil
+	}
+	funcs := []string{}
+	funcs = append(funcs, f.Name)
+	text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+	if err != nil {
+		return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
+	}
+	depth = depth + 1
+	for _, instr := range text {
```
aarzilli, line 1481: if depth > FollowCalls this loop doesn't do anything (in fact the prior call to Disassemble doesn't do anything either).

service/debugger/debugger.go
```diff
@@ -1467,6 +1467,65 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+				}
+			}
+		}
+	}
+	return funcs, nil
+
+}
+
+// FunctionsDeep returns a list of functions and it's children upto required depth indicated by FollowCalls in the target process.
+func (d *Debugger) FunctionsDeep(filter string, FollowCalls int) ([]string, error) {
```
aarzilli, line 1501: This should probably be a change to Debugger.Functions rather than a second copy with extra stuff.

service/debugger/debugger.go
archanaravindar: FollowCalls is the depth till which the user wants the trace to be printed out 
if we are reaching a depth greater than that we simply return hence the code has been written in this way

pkg/terminal/command.go
archanaravindar: Hi Alessandro i tried the following tests around panic, in both cases it appears that it works as expected. 

1. Panic does not recover
Code: 
```
package main

import "fmt"

func D(i int) int {
        panic("boom!")
        return i * i * i
}
func C(i int) int {

        return i + 20
}
func B(i int) int {
        d := C(i) + 40
        return d + D(i)
}
func A(i int) int {
        return 10 + B(i)
}
func main() {
        j := 0
        j += A(2)
        fmt.Println(j)
}

```
Trace output: 
```
 [archanaravindar@fedora _fixtures]$ ../dlv trace main.A --follow-calls 4 -e=./leafpanic
> goroutine(1): main.A(2)
 > goroutine(1): main.B(2)
  > goroutine(1): main.C(2)
  >> goroutine(1): => (22)
  > goroutine(1): main.D(2)
   > goroutine(1): runtime.gopanic(interface {}(string) "boom!")
> [unrecovered-panic] runtime.fatalpanic() /home/archanaravindar/Go/Go_master/go/src/runtime/panic.go:1175 (hits goroutine(1):1 total:1) (PC: 0x431024)
	runtime.curg._panic.arg: interface {}(string) "boom!"
  1170:	// fatalpanic implements an unrecoverable panic. It is like fatalthrow, except
  1171:	// that if msgs != nil, fatalpanic also prints panic messages and decrements
  1172:	// runningPanicDefers once main is blocked from exiting.
  1173:	//
  1174:	//go:nosplit
=>1175:	func fatalpanic(msgs *_panic) {
  1176:		pc := getcallerpc()
  1177:		sp := getcallersp()
  1178:		gp := getg()
  1179:		var docrash bool
  1180:		// Switch to the system stack to avoid any stack growth, which

```
2. Panic recovers
Code: 
```
package main

import "fmt"

func D(i int) int {
        defer func() {
                if r:=recover() ; r==nil {
                        fmt.Errorf("function should panic")
                } else {
                        fmt.Printf("successful recovery from panic\n")
                }
        }()
        panic("boom!")
        return i * i * i
}
func C(i int) int {

        return i + 20
}
func B(i int) int {
        d := C(i) + 40
        return d + D(i)
}
func A(i int) int {
        return 10 + B(i)
}
func main() {
        j := 0
        j += A(2)
        fmt.Println(j)
}

```
Trace output: 
```
[archanaravindar@fedora _fixtures]$ ../dlv trace main.A --follow-calls 3 -e=./leafpanic
> goroutine(1): main.A(2)
 > goroutine(1): main.B(2)
  > goroutine(1): main.C(2)
  >> goroutine(1): => (22)
  > goroutine(1): main.D(2)
  > goroutine(1): runtime.morestack_noctxt()
successful recovery from panic
  >> goroutine(1): => (0)
 >> goroutine(1): => (0)
>> goroutine(1): => (62)
>> goroutine(1): => (72)
72
Process 2135496 has exited with status 0

```

archanaravindar: > This code isn't going to work correctly if someone sets a tracepoint manually (depth will never be decremented).

You are right, as of now, we have implemented this option only for running with dlv at the command line prompt before entering into the dlv shell command environment 
we can still set a tracepoint manually using the trace option from within dlv and it will print a trace for that particular function but we dont support follow-calls option inside dlv yet 
```
[archanaravindar@fedora _fixtures]$ ../dlv exec ./leaf4
(dlv) trace main.C
Tracepoint 1 set at 0x47b064 for main.C() ./leaf4.go:8
(dlv) continue
> goroutine(1): main.C(2)
>> goroutine(1): => (22)
80
Process 2120528 has exited with status 0
(dlv) q
[archanaravindar@fedora _fixtures]$ ../dlv exec ./leaf4
(dlv) trace main.B
Tracepoint 1 set at 0x47b0aa for main.B() ./leaf4.go:12
(dlv) continue
> goroutine(1): main.B(2)
>> goroutine(1): => (70)
80
Process 2120556 has exited with status 0
```

service/debugger/debugger.go
aarzilli: The point I was trying to make is that depth > FollowCalls you can return early, before this loop and the previous Disassemble call.

service/debugger/debugger.go
aarzilli: Sure, but you can't set more than one breakpoint on D.

pkg/terminal/command.go
aarzilli: The problem is that a tracepoint set the the `trace` command doesn't necessarily create the corresponding TraceReturn tracepoints so depth isn't guaranteed to be decremented. Same goes for recovered panics.

pkg/terminal/command.go
```diff
@@ -2845,12 +2845,25 @@ func printBreakpointInfo(t *Term, th *api.Thread, tracepointOnNewline bool) {
...
+		// Treat morestack_noctxt as a special case
```
derekparker, line 2859: Please run `gofmt` on the project sources.

pkg/terminal/command.go
derekparker: The case of the panic + recover is interesting however. @archanaravindar as you can see from the example you provided in that case, after the panic is recovered the depth in the output is still the same. We could try and detect a recover, but then how do we know how much to decrement depth by?

service/debugger/debugger.go
archanaravindar: After some discussion with @derekparker we realized that we may end up going further down the call graph of the function depending on how we got to it and how far down in depth we are... so there can be a situation we may miss on a child if we short circuit the traversal  
alternatively looking at it, we can come to a function from two different parents and we would then need to traverse such a function twice anyway

pkg/terminal/command.go
aarzilli: As far as I can tell this affects all tracepoints, including the ones set with the trace command. Even so it's wrong because depth is not decremented with recovered panics. I think it would be better to ask for a stacktrace with the tracepoint (there's an option to do that) and try to determine the depth from that.

cmd/dlv/cmds/commands.go
```diff
@@ -697,13 +701,27 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
...
+					_, err = client.CreateBreakpoint(&api.Breakpoint{
+						FunctionName:     funcs[i],
+						Tracepoint:       true,
+						Line:             -1,
+						Stacktrace:       traceStackDepth,
+						LoadArgs:         &terminal.ShortLoadConfig,
+						TraceFollowCalls: traceFollowCalls,
+						RootFuncName:     regexp,
+					})
+				}
```
aarzilli, line 724: Just change the value of `traceStackDepth` instead of duplicating the call.

cmd/dlv/cmds/commands.go
```diff
@@ -716,13 +734,27 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
...
+						_, err = client.CreateBreakpoint(&api.Breakpoint{
+							Addr:             addrs[i],
+							TraceReturn:      true,
+							Stacktrace:       traceStackDepth,
+							Line:             -1,
+							LoadArgs:         &terminal.ShortLoadConfig,
+							TraceFollowCalls: traceFollowCalls,
+							RootFuncName:     regexp,
+						})
+					}
```
aarzilli, line 757: Same as previous comment, change the value of `traceStackDepth` instead of duplicating the call.

pkg/terminal/command.go
```diff
@@ -2849,17 +2848,65 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+
+		if fn.Name() == stack[curi].Function.Name() {
+			//                        fmt.Printf(" i %d equal %s %s\n",curi,fn.Name(), stack[i].Function.Name())
+			wantindex = curi
+		}
+	}
+	//sdepth:=mainindex-wantindex
+	sdepth := rootindex - wantindex + 1
+	//fmt.Printf("follow calls depth %d\n",th.Breakpoint.TraceFollowCalls)
+	//fmt.Printf("Root func name %s\n",th.Breakpoint.RootFuncName)
```
aarzilli, line 2876: I don't understand what this is doing. Shouldn't you be counting all appearances of traced calls in the stack?

service/debugger/debugger.go
aarzilli: > After some discussion with @derekparker we realized that we may end up going further down the call graph of the function depending on how we got to it and how far down in depth we are

True, let's make this a breadth first visit instead.

service/debugger/debugger.go
```diff
@@ -1467,6 +1467,65 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+	if depth > FollowCalls {
+		return nil, nil
+	}
+	funcs := []string{}
+	funcs = append(funcs, f.Name)
+	text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+	if err != nil {
+		return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
+	}
+	depth = depth + 1
```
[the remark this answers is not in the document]
archanaravindar: done

service/debugger/debugger.go
```diff
@@ -1467,6 +1467,65 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
 	return funcs, nil
 }
 
+func traverse(t proc.ValidTargets, f *proc.Function, depth int, FollowCalls int) ([]string, error) {
+	if depth > FollowCalls {
```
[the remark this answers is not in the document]
archanaravindar: When i make this change the traversal seems to be done for depth-1 when a trace is asked for depth

pkg/terminal/command.go
archanaravindar: Implemented computation of depth based on stacktrace because of this case

pkg/terminal/command.go
```diff
@@ -2840,29 +2840,90 @@ func printBreakpointInfo(t *Term, th *api.Thread, tracepointOnNewline bool) {
...
+	}
+
+	// Trace Follow Calls; stack is required to calculate depth of functions
+	wantindex := -1
+	rootindex := -1
+	if th.BreakpointInfo == nil || th.BreakpointInfo.Stacktrace == nil {
+		return
+	}
+
+	stack := th.BreakpointInfo.Stacktrace
```
aarzilli, line 2885: Can you explain what the logic behind this calculation is? Why is `wantindex` even needed? Why are you calculating curi as `len(stack) - 1 - i` when you later use it in a way that will always cancel the `len(stack)` term?

service/debugger/debugger.go
```diff
@@ -1464,6 +1467,82 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+				}
+			}
+		}
+	}
+	for _, fbinary := range t.BinInfo().Functions {
+		if strings.HasPrefix(fbinary.Name, f.Name) && fbinary.Name != f.Name {
+			children, err := traverse(t, &fbinary, depth, FollowCalls)
+			funcs = append(funcs, children...)
+			if err != nil {
+				return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
```
derekparker, line 1510: s/disassemble/traverse/ ?

service/debugger/debugger.go
```diff
@@ -1464,6 +1467,82 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+			if depth <= FollowCalls {
+				children, err := traverse(t, cf, depth, FollowCalls)
+				funcs = append(funcs, children...)
+				if err != nil {
+					return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
+				}
+			}
+		}
+	}
+	for _, fbinary := range t.BinInfo().Functions {
```
derekparker, line 1505: Add a comment around this (this is to support deferred functions with `main.foo.func1` type naming, correct?)

derekparker, line 1505: Can we optimize this by building an index as we go instead of looping through all program functions again? This doesn't necessarily have to happen in this PR, but perhaps it would be better to represent all program functions using a tree data structure as opposed to a slice. Let's do this later actually, but something to keep in mind.

service/debugger/debugger.go
```diff
@@ -1464,6 +1467,82 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+	depth++
+	for _, instr := range text {
+		if instr.IsCall() && instr.DestLoc != nil && instr.DestLoc.Fn != nil {
+			cf := instr.DestLoc.Fn
+			if cf.Name == f.Name {
+				continue
+			}
+			if depth <= FollowCalls {
+				children, err := traverse(t, cf, depth, FollowCalls)
+				funcs = append(funcs, children...)
```
derekparker, line 1498: Is it worth appending in the case of an error?

service/debugger/debugger.go
```diff
@@ -1464,6 +1467,82 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+				if err != nil {
+					return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
```
derekparker, line 1500: s/disassemble/traverse/ ?

derekparker: Also, wrap with `%w`.

service/debugger/debugger.go
```diff
@@ -1464,6 +1467,82 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+
+	if depth > FollowCalls {
+		// If we already reached desired depth, return
+		return nil, nil
+	}
+	funcs := []string{}
+	funcs = append(funcs, f.Name)
+	text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+	if err != nil {
+		return nil, fmt.Errorf("disassemble failed with error %s", err.Error())
```
derekparker, line 1487: Can wrap with `%w`.

derekparker (changes requested):

pkg/terminal/command.go
archanaravindar: There were many reasons for this- we needed rootindex to mark the depth of the starting func name as without that, it would report depth relative to main and not the function which was being given as a parameter, calculating it as len(stack)-1-i helps us to traverse from bottom up so that we can set rootindex properly at the first match in case the same function gets recursively invoked later (to avoid overwriting it with a higher depth etc), wantindex was required to mark depth of the function corresponding to the tracepoint.

aarzilli: Right, but why is wantindex the deepest call of the current tracepoint function and rootindex the least deep call? What if the current tracepoint function is recursively calling itself? Should that not increment the depth? What if the deepest call of the current tracepoint function is deeper than the least deep call of the root function?

service/debugger/debugger.go
aarzilli: Just to stress why this is important: tracing a function that calls fmt.Printf with `--follow-calls=10` on my system takes 10 seconds, just to set up the breakpoints. The number of breakpoints it ends up setting is relatively small (222) but to do that it ends up disassembly the same functions thousands of times. For example `fmt.(*buffer).writeString` is visited by `traverse` 4727 times (and redisassembled each time), `runtime.panicIndex` is visited 7900 times.

This implementation is extremely inefficient.

pkg/terminal/command.go
aarzilli: To be honest I undestand what this code is trying to do so little that I can't really tell what's intended behavior and what's a bug.

pkg/terminal/command.go
```diff
@@ -2873,16 +2875,72 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+		// Trace Follow Calls; stack is required to calculate depth of functions
+		wantindex = -1
+		rootindex = -1
+		if th.BreakpointInfo == nil || th.BreakpointInfo.Stacktrace == nil {
+			return
+		}
+
+		stack := th.BreakpointInfo.Stacktrace
+		for i := range stack {
+			curi := len(stack) - 1 - i
```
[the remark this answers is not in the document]
derekparker: I agree this is a bit confusing to read through.

So, `rootindex` is the closest frame of the root function (e.g. `foo` in the case of `dlv trace foo --follow-calls`) from the current function, and `wantindex` is the frame of a child call of that function, correct?

E.g. given:

```go
func f1() {
        f2()
}

func f2() {
        f3()
}

func f3() {}
```

Running `dlv trace f1 --follow-calls=3` once we hit `f3` the stack would look something like like:

```
[0] main.main
[1] f1 <-- rootindex
[2] f2
[3] f3 <-- wantindex
```

I think `wantindex` would always equal `len(stack)` right? Because we want to calculate depth from where we are at currently. In that case I don't think we even need to track it with a variable.

In which case couldn't we refactor the current algorithm like so:

```go
for i := len(stack)-1; i > 0; i-- {
        if stack[i].Function.Name() == th.Breakpoint.RootFuncName {
		rootindex = i
                break // no need to continue looping at this point
	}
}
/*
depth we want == 2
len(stack) == 4
rootindex == 1
(len(stack) - 1) - rootindex == 2
*/
sdepth = len(stack) - rootindex - 1
```

However this wouldn't handle recursive function call depth correctly, I think you would want to find `rootindex` closest to the top of the stack, right? Kind of like what you're doing with `wantindex` currently.

As it stands right now I think there might be a bug in your calculation, if I'm understanding correctly. We only assign `rootindex` once (guarded by the `if rootindex == -1` branch) however we assign `wantindex` everytime we see it, because we don't short circuit the loop. So given the same situation above, but if the stack somehow looked like this (this is a contrived example, you can't get into this situation with the above code):

```
[0] main.main
[1] f3 <-- wantindex
[2] f1
[3] f3
[4] f1 <-- rootindex
[5] f3
```
Then in the current algorithm would place `wantindex` at `1` and the calculation would be:

```
wantindex == 1
rootindex == 4
4 - 1 + 1 == 4
```

Which would be incorrect. Again, please correct me if I am misunderstanding the current algorithm.

I think it may be worth extracting this code into it's own function and unit testing it to ensure correctness in different scenarios.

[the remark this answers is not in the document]
aarzilli: I've been thinking about this a bit more and there are two possible interpretation to this feature:

1. the "static" interpretation: tracepoints are set on all functions reachable from the root function up to the specified depth. These tracepoints then behave like normal tracepoints, so, for example, if we set a tracepoint on `f` with `--follow-calls=1` and `f` recursively calls itself to a depth of 10 we print all ten recursive calls
2. the "dynamic" interpretation: all calls made from the root function up to the specified depth are traced. If we set a tracepoint on `f` with `--follow-calls=1` and `f` recursively calls itself to a depth of 10 we only print the first call.

I thought we were going for the "static" interpretation (that's what it did originally and the "dynamic" interpretation is pretty hard to do), but the current implementation is a mix of both. It shouldn't be.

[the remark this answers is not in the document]
derekparker: I updated my comment above because I had a few things reversed which may have made my thought a bit more difficult to understand. Also, the issue with recursion isn't an issue at all as I thought about it more.

The more I think about this, I believe this would always be correct:

```go
/*
Given the following stack:

[5] runtime.goexit
[4] runtime.main
[3] main.main
[2] f1 <-- rootindex
[1] f2
[0] f3 <-- wantindex
*/

// We do not need to visit stack[0] (current frame) because that will always be `wantindex`, the child function
// which triggered the breakpoint.
for i := len(stack)-1; i > 0; i-- {
        if stack[i].Function.Name() == th.Breakpoint.RootFuncName {
		sdepth = i
                break // no need to continue looping at this point
	}
}
```

In other words, we don't need to do any arithmetic at all. Since we're going from the top of the stack down towards the bottom (current frame), the iterator value of the rootindex will always be equal to the depth.

We could rework the algorithm to go from the bottom of the stack (current frame) up, and use the iterator as the depth:

```go
for i := range stack {
    if stack[i].Function.Name() == th.Breakpoint.RootFuncName {
        sdepth = i // always overwrite sdepth whenever we see the root function
    }
}
```

However it is less efficient because we would have to traverse the entire stack every time, whereas with the current approach of going top down we can short circuit and break as soon as we see the root function appear.

archanaravindar:
Hi @aarzilli thanks for your suggestion to improve the performance of the traversal algorithm, we are working on this now, can you please give the same example as mentioned in https://github.com/go-delve/delve/pull/3594#discussion_r1521272063
you had seen which was very inefficient so that we could try it with the new implementation and see if we got improvement, thanks!

aarzilli:
>can you please give the same example as mentioned in [#3594 (comment)](https://github.com/go-delve/delve/pull/3594#discussion_r1521272063) 

Just:

```
package main

import "fmt"

func f() {
    fmt.Printf("hello world\n")
}

func main() {
    f()
}
```

and trace main.f with `--follow-calls=10`, if that's fast for you try `--follow-calls=10000`.

archanaravindar:
Optimized traversal and tested with example provided in https://github.com/go-delve/delve/pull/3594#issuecomment-2049857470 
Testing with depth of 10000 the execution time remains well below 1 second
image

aarzilli:
> Optimized traversal

I think that making it a BFS is still preferable to optimizing a DFS with caching:

```
d := make(map[string]int)
fringe := []string
add := func(fname, dist int) {
	if curdist, ok := d[fname]; ok {
		if curdist > dist {
			d[fname] = dist
		}
		return
	}
	d[fname] = dist
	fringe = append(fringe, fname)
}

add(f.Name)

for len(fringe) > 0 {
	cur = fringe[0]
	fringe = fringe[1:] // leaking slightly but it should be fine
	if d[cur] >= depth {
		continue
	}
	dcur := d[cur]
	for _, next := range functionsCalledFromDisassembly(cur) {
		add(next, dcur+1)
	}
}

r := []string{}
for fname, dist := range d {
	if dist < depth {
		r = append(r, fname)
	}
}
return r
```

archanaravindar:
> > Optimized traversal
> 
> I think that making it a BFS is still preferable to optimizing a DFS with caching:
> 
> ```
> d := make(map[string]int)
> fringe := []string
> add := func(fname string, dist int) {
> 	if curdist, ok := d[fname]; ok {
> 		if curdist > dist {
> 			d[fname] = dist
> 		}
> 		return
> 	}
> 	d[fname] = dist
> 	fringe = append(fringe, fname)
> }
> 
> add(f.Name)
> 
> for len(fringe) > 0 {
> 	cur = fringe[0]
> 	fringe = fringe[1:] // leaking slightly but it should be fine
> 	if d[cur] >= depth {
> 		continue
> 	}
> 	dcur := d[cur]
> 	for _, next := range functionsCalledFromDisassembly(cur) {
> 		add(next, dcur+1)
> 	}
> }
> 
> r := []string{}
> for fname, dist := range d {
> 	if dist < depth {
> 		r = append(r, fname)
> 	}
> }
> return r
> ```
> 
> (not tested, didn't even compile it)

Hi @aarzilli are you saying that the remaining cost now to optimize is that of recursion? if yes we could try the same implementation with BFS instead of rewriting it all over again with a new algorithm. 
if not, can you highlight which part of the code would be suboptimal or give an example like you had clearly mentioned in the previous repeated disassembly case that would help in this case, thanks!

aarzilli:
The problem is not the performance, is that all these optimizations are unreadable and using the correct algorithm would render them unnecessary. Also, there are other comments on this code that have not been addressed in addition to traverse using the wrong algorithm, before we should merge this.

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+	data := element.Value
+	F, ok := data.(TraceFuncptr)
+	if !ok {
+		fmt.Printf("Type error\n")
+		os.Exit(1)
+	}
+	return F
+}
+
+func traverse(t proc.ValidTargets, f *proc.Function, depth int, FollowCalls int) ([]string, error) {
```
derekparker, line 1532: s/FollowCalls/followCalls/

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+		}
+		for _, cf := range parent.children {
+			if cf.Name == parent.Func.Name {
+				//recursive functions
+				continue
+			}
+			childnode := TraceMap[cf.Name]
+			if childnode == nil {
+				// childnodes should have already been computed in the above loop
+				fmt.Printf("unexpected error child nil \n")
```
derekparker, line 1607: Since this would be a programmer error I think it's ok to panic here instead.

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+				os.Exit(1)
+			}
+			if childnode.Depth < parent.Depth+1 {
+				// This path has already been explored no need to revisit
+				continue
+			}
+			if childnode.Depth <= FollowCalls {
```
derekparker, line 1614: Won't this always be true due to the above line:

```
if parent.Depth+1 > FollowCalls {
	// Skip disassembling and computing children if we are going to cross FollowCalls
	continue
}
```

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+	queue = list.New()
+	queue.Init()
+
+	funcs := []string{}
+
+	rootnode := &TraceFunc{Func: new(proc.Function), Depth: depth, children: []*proc.Function{}, visited: false, dis: false}
+	rootnode.Func = f
+
+	TraceMap[f.Name] = rootnode
+	queue.PushBack(TraceMap[f.Name])
```
derekparker, line 1544: If we go with a slice instead, you could do:

```
queue := make([]traceFuncPtr, 0, 20)
queue[0] = rootnode
```

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+					if !filter(cf.Name) {
+						continue
+					}
+					childnode := TraceMap[cf.Name]
+					if childnode == nil {
+						childnode = &TraceFunc{Func: new(proc.Function), Depth: parent.Depth + 1, children: []*proc.Function{}, visited: false, dis: false}
+						childnode.Func = cf
+						TraceMap[cf.Name] = childnode
+						parent.children = append(parent.children, cf)
+					} else {
```
derekparker, line 1579: I think we can drop this else clause right? Since we're doing a BFS now, if we've seen the function before then we won't find anything new if we see it again at a larger depth. I think you can drop the loop below too (`for _, cf := range parent.children {`), we can just directly enqueue any new function right here, right?

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+							deferdepth = childnode.Depth
+							deferparent = parent
+						}
+					}
+
+				}
+			}
+			parent.dis = true
+		}
+		for _, cf := range parent.children {
```
derekparker, line 1599: (as noted in an earlier comment) I think this loop is unnecessary and we can directly enqueue in the above loop.

derekparker (changes requested):
Looks like, I like the BFS approach, but I think due to using BFS we can simplify and clean this up even more.

pkg/terminal/command.go
```diff
@@ -2921,16 +2923,62 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+	if th.Breakpoint.TraceFollowCalls > 0 {
+		// Trace Follow Calls; stack is required to calculate depth of functions
+		rootindex = -1
+		if th.BreakpointInfo == nil || th.BreakpointInfo.Stacktrace == nil {
+			return
+		}
+		stack := th.BreakpointInfo.Stacktrace
+		for i := len(stack) - 1; i >= 0; i-- {
+			if stack[i].Function.Name() == th.Breakpoint.RootFuncName {
+				if rootindex == -1 {
```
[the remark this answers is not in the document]
derekparker: Talked through offline, this is ok.

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+				// This path has already been explored no need to revisit
+				continue
+			}
+			if childnode.Depth <= FollowCalls {
+				queue.PushBack(TraceMap[cf.Name])
+			}
+		}
+
+		if deferdepth > 0 {
+			for _, fbinary := range t.BinInfo().Functions {
```
derekparker, line 1620: We can actually get rid of this function name lookup heuristic. This will only work when deferring anonymous functions such as:

```
defer func() {}()
```

What we should actually be doing is looking for an assembly sequence such as:

```
lea rcx, ptr [rip+0x24ab3]
mov qword ptr [rsp+0x20], rcx
```

Which is loading into a register (in this case `$rcx`) the address of the deferred function, which it is referencing as an offset from the current instruction pointer. It then will eventually load this into a stack slot. You can calculate what the instruction pointer value will be with some arithmetic. Essentially, it would be `image.StaticBase + (function entry address) + (offset of this instruction, the number of bytes decoded so far)`.

It's necessary to take the `image.StaticBase` into account for handling PIE binaries. On non-PIE binaries that value will be 0 anyways.

derekparker: Note that also this address may point to a defer wrapper function on certain Go versions, but that's ok for now, we can skip past the wrapper function later, or not increment depth if we hit the wrapper function so that we effectively skip it.

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+var TraceMap = make(map[string]TraceFuncptr)
+
+func filter(fname string) bool {
+	idx := strings.Index(fname, "runtime.")
+	idx2 := strings.Index(fname, "runtime.defer")
+	idx3 := strings.Index(fname, "runtime.gorecover")
+	idx4 := strings.Index(fname, "runtime.gopanic") 
+	// Avoid runtime/internal functions to figure in the trace function list as they dont have a stack associated
+	idx5 := strings.Index(fname, "runtime/internal")
+	if (idx != -1 || idx5 != -1) && idx2 == -1 && idx3 == -1 && idx4 == -1 {
```
aarzilli, line 1513: `return !((strings.HasPrefix(fname, "runtime.") || strings.HasPrefix(fname, "runtime/internal.")) && fname != "runtime.defer" && fname != "runtime.gorecover" && fname != "runtime.gopanic")`

service/debugger/debugger.go
```diff
@@ -1919,6 +2083,7 @@ func (d *Debugger) convertDefers(defers []*proc.Defer) []api.Defer {
 				SP: defers[i].SP,
 			}
 		}
+
```
aarzilli, line 2086: Remove empty line.

service/rpc2/server.go
```diff
@@ -587,7 +588,9 @@ type ListFunctionsOut struct {
 
 // ListFunctions lists all functions in the process matching filter.
 func (s *RPCServer) ListFunctions(arg ListFunctionsIn, out *ListFunctionsOut) error {
-	fns, err := s.debugger.Functions(arg.Filter)
+	var fns []string
```
aarzilli, line 591: Why?

pkg/terminal/command.go
```diff
@@ -2921,16 +2923,62 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+		}
+		// compute relative depth from root function passed as regexp in follow-calls in the stack
+		sdepth = rootindex + 1
+	}
+
 	if th.Breakpoint.Tracepoint {
-		fmt.Fprintf(t.stdout, "> goroutine(%d): %s%s(%s)\n", th.GoroutineID, bpname, fn.Name(), args)
+		if th.Breakpoint.TraceFollowCalls > 0 {
+			if rootindex >= 0 {
+				for i := 1; i < sdepth; i++ {
```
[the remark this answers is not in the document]
aarzilli: I'd add that this block of code is repeated almost verbatim below, the calculation of depthPrefix as well as deciding whether to print `frame(%d)` after `goroutine(%d)` should be extracted from here.

service/rpc2/server.go
archanaravindar: thanks for catching this, corrected

service/debugger/debugger.go
archanaravindar: done.

service/debugger/debugger.go
archanaravindar: done

service/debugger/debugger.go
archanaravindar: done

service/debugger/debugger.go
archanaravindar: we need this check to avoid queueing functions even if we have crossed depth

service/debugger/debugger.go
archanaravindar: as discussed the support for defer calls in the follow funcs feature will be done in a follow up PR

pkg/terminal/command.go
```diff
@@ -2921,16 +2923,62 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+	depthPrefix = ""
 	if th.Breakpoint.TraceReturn {
 		retVals := make([]string, 0, len(th.ReturnValues))
 		for _, v := range th.ReturnValues {
 			retVals = append(retVals, v.SinglelineString())
 		}
-		fmt.Fprintf(t.stdout, ">> goroutine(%d): %s => (%s)\n", th.GoroutineID, fn.Name(), strings.Join(retVals, ","))
+		if th.Breakpoint.TraceFollowCalls > 0 {
+			if rootindex >= 0 {
+				for i := 1; i < sdepth; i++ {
```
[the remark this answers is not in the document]
archanaravindar: done

pkg/terminal/command.go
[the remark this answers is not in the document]
archanaravindar: done

pkg/proc/breakpoints.go
```diff
@@ -63,6 +63,11 @@ type Breakpoint struct {
 	// ReturnInfo describes how to collect return variables when this
 	// breakpoint is hit as a return breakpoint.
 	returnInfo *returnBreakpointInfo
+
+	// Name of root function from where tracing needs to be done
+	RootFuncName string
+	// depth of tracing
```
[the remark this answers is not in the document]
archanaravindar: done

pkg/proc/breakpoints.go
```diff
@@ -63,6 +63,11 @@ type Breakpoint struct {
 	// ReturnInfo describes how to collect return variables when this
 	// breakpoint is hit as a return breakpoint.
 	returnInfo *returnBreakpointInfo
+
+	// Name of root function from where tracing needs to be done
```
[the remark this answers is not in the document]
archanaravindar: done

pkg/terminal/command.go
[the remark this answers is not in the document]
archanaravindar: the new version implements the short circuit idea

cmd/dlv/cmds/commands.go
```diff
@@ -716,12 +730,20 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
 					continue
 				}
 				for i := range addrs {
+					var stackdepth int
+					if traceFollowCalls > 0 {
+						stackdepth = 20
```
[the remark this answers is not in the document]
archanaravindar: since stacktrace and follow calls are mutually exclusive we make use of -s to indicate stack depth the default being 20

pkg/terminal/command.go
```diff
@@ -2840,29 +2840,90 @@ func printBreakpointInfo(t *Term, th *api.Thread, tracepointOnNewline bool) {
...
+			printStack(t, t.stdout, bpi.Stacktrace, "\t\t", false)
+		}
 	}
 }
 
 func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, args string, hasReturnValue bool) {
 	if t.conf.TraceShowTimestamp {
 		fmt.Fprintf(t.stdout, "%s ", time.Now().Format(time.RFC3339Nano))
 	}
+	// Simple Trace; no stack required to calculate depth of functions
```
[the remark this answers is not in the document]
archanaravindar: done

service/debugger/debugger.go
```diff
@@ -1475,6 +1486,73 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+		}
+		f := parent.Func
+		text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+		if err != nil {
+			return nil, fmt.Errorf("disassemble failed with error %w", err)
+		}
+		for _, instr := range text {
+			if instr.IsCall() && instr.DestLoc != nil && instr.DestLoc.Fn != nil {
+				cf := instr.DestLoc.Fn
+				if !filterRuntimeFuncs(cf.Name) {
```
aarzilli, line 1539: Inline filterRuntimeFuncs here.

service/debugger/debugger.go
```diff
@@ -1475,6 +1486,73 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+					continue
+				}
+				childnode := TraceMap[cf.Name]
+				if childnode == nil {
+					childnode = &TraceFunc{Func: new(proc.Function), Depth: parent.Depth + 1, visited: false}
```
aarzilli, line 1544: It doesn't make sense to initialize Func with `new(proc.Function)` only to replace it with `cf` in the very next line.

service/debugger/debugger.go
archanaravindar: Done

service/debugger/debugger.go
```diff
@@ -1475,6 +1487,158 @@ func (d *Debugger) Functions(filter string) ([]string, error) {
...
+	if !ok {
+		fmt.Printf("Type error\n")
+		os.Exit(1)
+	}
+	return F
+}
+
+func traverse(t proc.ValidTargets, f *proc.Function, depth int, FollowCalls int) ([]string, error) {
+
+	var queue *list.List
```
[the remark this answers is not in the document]
archanaravindar: done

derekparker (approved):
LGTM

Issue #3539, opened 2023-10-26
`p len(string(data))` != `p len(data)`


extemporalgenome:
Please answer the following before submitting your issue:

1. What version of Delve are you using (`dlv version`)?
```
Delve Debugger
Version: 1.21.0
Build: $Id: fec0d226b2c2cce1567d5f59169660cf61dc1efe $
```

3. What version of Go are you using? (`go version`)?
`go version go1.21.0 darwin/amd64`

4. What operating system and processor architecture are you using?

darwin/amd64

5. What did you do?
Print the backing array contents of a large-ish []byte as a string.

6. What did you expect to see?

```
(dlv) p len(data)
1202
(dlv) p len(string(data))
1202
```

7. What did you see instead?

```
(dlv) p len(data)
1202
(dlv) p len(string(data))
64
```

extemporalgenome:
It appears that when given a []byte, `max-array-values` must be at least as large as `max-string-len` for it to be respected when performing this command: `p string(theByteSlice)`.

My expectation would be that `max-array-values` is only utilized at the time of presentation, and that `p string(theByteSlice)` would ignore the value of `max-array-values` entirely, convert (or interpret) the entire slice as a string, and then use the value of `max-string-len` to configure the presentation of the result.

aarzilli:
This is essentially by design. The various limits (max-array-values, max-string-length, &c) don't apply to presentation but to how much memory is read. This is so that we don't try to read infinite amounts of memory even in presence of corrupted values.

Converting a byte array to a string involves reading the byte array and reinterpreting its contents, thus the limits on how many array elements we read are applied. We could fix this particular instance of the problem but it would just push the problem one level up: `p len([]rune(string(byteslice)))`

derekparker:
I wonder if there could be a way to optimize for this in our evaluator, because technically in this case we don't need to load the value of the string, we really just read the length field from the structs and compare those.

aarzilli:
Closing, fixed by 57dad9342a9443e298d43fcc653475408402131e

extemporalgenome:
That's fantastic! Thanks!

PR #3596, opened 2023-12-11
proc: make some type casts less counterintuitive

### LINKED ISSUE #3595: Bug in display: Certain slices not working within string()
Delve Version: 1.21.2
Go Version: 1.21.5 , but this has been an issue for many many versions
MacOS/ARM

Issue:

Input is often handled as []byte, making it important to be able to display such values as strings during debugging. Given a byte slice of the form `bytes[e:]` (where e is an expression)

    dlv -a string(bytes[5:]) // Prints expected byte slice as a string
    dlv -a string(bytes[pos:])  // Always shows the empty string

That is: the expression evaluation fails if the first argument of the slice operation is non-constant.

Further, I have just noticed that `string(bytes[0:40])` doesn't work either, no matter what constants are used - it shows the empty string.

Using the `%s` format in the display addition has no impact on this outcome. In fact, it doesn't print the string representation at al. Even more fun, `display -a %s string(bytes[5:]` prints no value at all, even though it works fine without the format specifier.

I haven't dug in to the code, but I'm guessing this is a deficiency or a bug within the expression evaluator.

aarzilli:
The interaction of type casts with load configuration is sometimes
counterintuitive. This commit changes the way it is performed so that
when converting slices to strings and vice versa the maximum size
corresponding to the target type is used (i.e. MaxStringLen when
converting slices to strings and MaxArrayValues when converting slices
to strings).

This doesn't fully solve the problem (conversions to []rune are
problematic and multiple chained type casts will still be confusing)
but removes the problem for the majority of practical uses.

Fixes #3595, #3539

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -344,6 +344,7 @@ func main() {
 	runearray := [4]rune{116, 232, 115, 116}
 
 	longstr := "very long string 0123456789a0123456789b0123456789c0123456789d0123456789e0123456789f0123456789g012345678h90123456789i0123456789j0123456789"
+	longbyteslice := []byte(longstr)
 	m5 := map[C]int{{longstr}: 1}
 	m6 := map[string]int{longstr: 123}
 	m7 := map[C]C{{longstr}: {"hello"}}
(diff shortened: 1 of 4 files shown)


derekparker (approved):
LGTM

PR #3597, opened 2023-12-11
*: Use forked goretk/gore module

derekparker:
This allows us to update this dependency on our schedule which is important because the module relies on manually updating the known list of Go versions to function correctly. Forking allows us to keep this up to date ourselves and possibly create a new system to prevent having to perform this manual step in the future.


aarzilli (approved):
LGTM

PR #3601, opened 2023-12-20
teamcity,version: add 1.22 to supported versions and CI matrix

aarzilli:


aarzilli:
In the past we used to do a release and then add the new go version to CI after the stable version was released, however a couple of times this has caused us to release delve without full support for the new version of go (both times because of bugs in the go compiler/runtime on linux/386).

So I think we should first add 1.22 to supported and test matrix and then make a release of delve.

derekparker (approved):
LGTM

PR #3607, opened 2023-12-23
goversion: include pre-releases in VersionAfterOrEqual check

aarzilli:
Change VersionAfterOrEqual(x, A, B) to return true for pre-releases of
version A.B. This is what it did before, it was broken when goversion
was changed to support the new version format.


derekparker (approved):
lgtm

PR #3622, opened 2024-01-03
TeamCity: speed up ppc64le CI

aarzilli:
The builder is currently spending 15 to 20 minutes installing gcc and
upgrading packages every time we run the tests.

Because of this the build fails sometimes by running out of time.

This change reduces that to 5 minutes by:

* switching from curl to wget (which seems to have fewer dependencies)
* not installing gcc on ppc64le
* skipping tests that depend on gcc or other binutils

### DIFF
--- a/_scripts/test_linux.sh
+++ b/_scripts/test_linux.sh
@@ -2,11 +2,15 @@
 set -e
 set -x
 
-apt-get -qq update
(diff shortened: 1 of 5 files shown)


derekparker (approved):
lgtm

PR #3623, opened 2024-01-03
proc: if a reslice operator is used load the whole slice

### LINKED ISSUE #3600: Printing a subscriptable type should respect the subscript size, even if larger than max-array-values

Delve Debugger
Version: 1.21.1
Build: $Id: a358c02f24aa7047ecc562b0587dc2d08330b2cf $

go version go1.21.3 linux/amd64

4. What operating system and processor architecture are you using?

linux amd64

6. What did you do?

Print a slice variable with an explicit range.  I have my `max-array-values` set to 4.  In this case I wanted to see more, so I said  `p var[0:27]` and I got...4 values.  It seems like the explicit indexing should be respected here, within reason?  I guess if it was 25000 entries that would be bad, but in this case it was totally reasoable.

Same with strings!

7. What did you expect to see?

More information

9. What did you see instead?

Not much information

aarzilli:
When the users uses a reslice operation load the whole resliced
variable, ignoring the MaxArrayValues setting.

Only apply this when the 'high' part is specified and a literal and the
'low' part is either unspecified or a literal.

Fixes #3600

### DIFF
--- a/pkg/proc/evalop/evalcompile.go
+++ b/pkg/proc/evalop/evalcompile.go
@@ -483,6 +483,7 @@ func (ctx *compileCtx) compileReslice(node *ast.SliceExpr) error {
 		return err
 	}
 
+	trustLen := true
 	hasHigh := false
 	if node.High != nil {
 		hasHigh = true
@@ -490,6 +491,10 @@ func (ctx *compileCtx) compileReslice(node *ast.SliceExpr) error {
 		if err != nil {
 			return err
 		}
+		_, isbasiclit := node.High.(*ast.BasicLit)
+		trustLen = trustLen && isbasiclit
+	} else {
+		trustLen = false
 	}
 
 	if node.Low != nil {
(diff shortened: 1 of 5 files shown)


pkg/proc/evalop/evalcompile.go
```diff
@@ -483,25 +483,32 @@ func (ctx *compileCtx) compileReslice(node *ast.SliceExpr) error {
...
+	} else {
+		trustLen = false
 	}
 
 	if node.Low != nil {
 		err = ctx.compileAST(node.Low)
 		if err != nil {
 			return err
 		}
+		_, isbasiclit := node.Low.(*ast.BasicLit)
```
derekparker, line 505: Why perform this check on the low side, what would cause us to not trust the length depending on the low side of the slice operation?

aarzilli: My thinking is that we only do this when the expression has a fixed length and something like `s[a:1000000]` doesn't.

derekparker (approved):
LGTM

thockin:
Yay, thanks!

PR #3632, opened 2024-01-15
cmd/dlv: print out message with stack trace when breakpoint is hit but has no waiting client

### LINKED ISSUE #3469: Feature request: headless startup without break-on-panic

```
Delve Debugger
Version: 1.21.0
Build: $Id: fec0d226b2c2cce1567d5f59169660cf61dc1efe $
```

```
go version go1.20.5 darwin/amd64
```

```
darwin/amd64
```

4. What did you do?

We make local development convenient by starting the program under development with both a restart-on-change and debug-server setup by default. For Go, this looks like wrapping the process in `gow` and `dlv debug`, e.g. 

```dockerfile
CMD ["gow", "-g=dlv", "-s", "-r=false", "-w=protos", "-w=server", "debug", "--headless", "--listen=0.0.0.0:9002", "--accept-multiclient", "--continue", "./server"]
```

This has been working fine for a while but we recently realised it has an unfortunate side effect. Its standard practice to have package-level initialisers panic if they fail for some reason, e.g.

```go
var somePattern = regexp.MustCompile(`(??)`)
```

In the above example, the regular expression is invalid, and will cause a panic. In this case, with the setup we have, the break-on-panic kicks in and the process is stopped, without any indication as to what is happening.

I've also tried adding `--init <(echo clear)` but `--init` is not allowed in headless mode.

Related (I think): #3028 

We expect, or rather would like, a way to have uncaught panics not trigger a breakpoint when dlv is started in headless mode, until the developer connects and alters the behaviour.

7. What did you see instead?

Timeouts or even just no output instead of the errors the process should be showing due to the panic.

fatanugraha:
Print out a message and stack trace to the stderr when we hit a breakpoint and the client that sends the continue command has disconnected.

Fixes #3469

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -251,6 +251,36 @@ func TestOutput(t *testing.T) {
 	}
 }
 
+// TestUnattendedBreakpoint tests whether dlv will print a message to stderr when the client that sends continue is disconnected
+// or not.
+func TestUnattendedBreakpoint(t *testing.T) {
+	const listenAddr = "127.0.0.1:40573"
+
+	fixturePath := filepath.Join(protest.FindFixturesDir(), "panic.go")
+	cmd := exec.Command(getDlvBin(t), "debug", "--continue", "--headless", "--accept-multiclient", "--listen", listenAddr, fixturePath)
+	stderr, err := cmd.StderrPipe()
+	assertNoError(err, t, "stdout pipe")
+	defer stderr.Close()
+
+	assertNoError(cmd.Start(), t, "start headless instance")
+
+	scan := bufio.NewScanner(stderr)
+	for scan.Scan() {
+		fmt.Println(scan.Text())
+		t.Log(scan.Text())
+		if strings.Contains(scan.Text(), "execution is paused because your program is panicking") {
+			break
+		}
+	}
+
+	// and detach from and kill the headless instance
+	client := rpc2.NewClient(listenAddr)
+	if err := client.Detach(true); err != nil {
+		t.Fatalf("error detaching from headless instance: %v", err)
+	}
+	cmd.Wait()
+}
+
 // TestContinue verifies that the debugged executable starts immediately with --continue
 func TestContinue(t *testing.T) {
 	const listenAddr = "127.0.0.1:40573"
@@ -1422,35 +1452,5 @@ func TestUnixDomainSocket(t *testing.T) {
 	}
 
 	client.Detach(true)
-	cmd.Wait()
-}
-
-// TestUnattendedBreakpoint tests whether dlv will print a message to stderr when the client that sends continue is disconnected
-// or not.
-func TestUnattendedBreakpoint(t *testing.T) {
-	const listenAddr = "127.0.0.1:40573"
-
-	fixturePath := filepath.Join(protest.FindFixturesDir(), "panic.go")
-	cmd := exec.Command(getDlvBin(t), "debug", "--continue", "--headless", "--accept-multiclient", "--listen", listenAddr, fixturePath)
-	stderr, err := cmd.StderrPipe()
-	assertNoError(err, t, "stdout pipe")
-	defer stderr.Close()
-
-	assertNoError(cmd.Start(), t, "start headless instance")
-
-	scan := bufio.NewScanner(stderr)
-	for scan.Scan() {
-		fmt.Println(scan.Text())
-		t.Log(scan.Text())
-		if scan.Text() == "execution is paused because your program is panicking" {
-			break
-		}
-	}
-
-	// and detach from and kill the headless instance
-	client := rpc2.NewClient(listenAddr)
-	if err := client.Detach(true); err != nil {
-		t.Fatalf("error detaching from headless instance: %v", err)
-	}
 	cmd.Wait()
 }
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -1424,3 +1424,33 @@ func TestUnixDomainSocket(t *testing.T) {
 	client.Detach(true)
 	cmd.Wait()
 }
+
+// TestUnattendedBreakpoint tests whether dlv will print a message to stderr when the client that sends continue is disconnected
+// or not.
+func TestUnattendedBreakpoint(t *testing.T) {
+	const listenAddr = "127.0.0.1:40573"
+
+	fixturePath := filepath.Join(protest.FindFixturesDir(), "panic.go")
+	cmd := exec.Command(getDlvBin(t), "debug", "--continue", "--headless", "--accept-multiclient", "--listen", listenAddr, fixturePath)
+	stderr, err := cmd.StderrPipe()
+	assertNoError(err, t, "stdout pipe")
+	defer stderr.Close()
+
+	assertNoError(cmd.Start(), t, "start headless instance")
+
+	scan := bufio.NewScanner(stderr)
+	for scan.Scan() {
+		fmt.Println(scan.Text())
+		t.Log(scan.Text())
+		if scan.Text() == "execution is paused because your program is panicking" {
+			break
+		}
+	}
+
+	// and detach from and kill the headless instance
+	client := rpc2.NewClient(listenAddr)
+	if err := client.Detach(true); err != nil {
+		t.Fatalf("error detaching from headless instance: %v", err)
+	}
+	cmd.Wait()
+}
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -266,7 +266,6 @@ func TestUnattendedBreakpoint(t *testing.T) {
 
 	scan := bufio.NewScanner(stderr)
 	for scan.Scan() {
-		fmt.Println(scan.Text())
 		t.Log(scan.Text())
 		if strings.Contains(scan.Text(), "execution is paused because your program is panicking") {
 			break
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -533,7 +533,6 @@ func (s *Session) ServeDAPCodec() {
 			s.config.triggerServerStop()
 		}
 	}()
-
 	reader := bufio.NewReader(s.conn)
 	for {
 		request, err := dap.ReadProtocolMessage(reader)
@@ -2756,7 +2755,7 @@ func (s *Session) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr 
 				variablesReference = maybeCreateVariableHandle(v)
 			}
 		}
-	case reflect.Struct:
+	case reflect.Struct, reflect.Func:
 		if v.Len > int64(len(v.Children)) { // Not fully loaded
 			if len(v.Children) == 0 { // Fully missing
 				value = reloadVariable(v, qualifiedNameOrExpr)
@@ -2781,7 +2780,7 @@ func (s *Session) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr 
 			v.Children[1].Kind = reflect.Float64
 		}
 		fallthrough
-	default: // Complex, Scalar, Chan, Func
+	default: // Complex, Scalar, Chan
 		if len(v.Children) > 0 {
 			variablesReference = maybeCreateVariableHandle(v)
 		}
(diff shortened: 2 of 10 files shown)


aarzilli (commented):
I still think that needing this option is a strong symptom that there is something very wrong with your development environment. In the context of running something locally during development you would want the program to stop at a panic to examine why the panic happened.
In the context of running the program in production having a debugger attached to it constantly is a security risk.

derekparker:
I'm also inclined to agree with @aarzilli, however I am familiar with the workflow described in the associated issue, and in that context I could see the benefit to the actual crash being apparent and not silenced by Delve by virtue of waiting on user input to produce the panic.

My initial feedback is I am against wholesale deletion of these breakpoints, I think it would be better to actually disable them, so they can be re-enabled later without having to know the specifics of the runtime.

aarzilli:
We could also print something if we stop at a breakpoint while there isn't a connected client.

derekparker:
> We could also print something if we stop at a breakpoint while there isn't a connected client.

That's good too. Might even be nice to print the traceback and panic message automatically when hitting the panic breakpoints anyways, to be honest.

fatanugraha:
> I still think that needing this option is a strong symptom that there is something very wrong with your development environment.

I agree with you. This option is needed in our orgs because not everyone are familiar with the debugger and we want to have a setup that is practical and convenient for everyone.

> My initial feedback is I am against wholesale deletion of these breakpoints, I think it would be better to actually disable them, so they can be re-enabled later without having to know the specifics of the runtime.

I tried this one actually, instead of deleting them I tried to toggling off those breakpoints, but it's not possible to do that through the API because we will reject any request that has `-` in it. If we're okay with changing the API behaviour I can adjust the MR to do this instead.

> That's good too. Might even be nice to print the traceback and panic message automatically when hitting the panic breakpoints anyways, to be honest.

Ideally yeah this is what I'm looking for, but it seems to be a bit tricky to pull off. Let me see if i can find a way to do this.

aarzilli:
> Ideally yeah this is what I'm looking for, but it seems to be a bit tricky to pull off. Let me see if i can find a way to do this.

One way to do it would be like this:

* There's a function in service/rpccommon/server.go called serveJSONCodec, that's the function that serves requests from a client. You can have a boolean there that you set to false when that function exits and check it in `(*RPCCallback).Return`.

* This boolean needs to be protected with a mutex so you could copy the `syncflag` type in service/dap/server.go. 

* Have `(*RPCCallback).Return` return false if the client has quit and then print something in `(*RPCServer).Command` in service/rpc2/server.go

fatanugraha:
> One way to do it would be like this:

oh ya i have implemented this in this PR but in a different fashion. 

> Might even be nice to print the traceback and panic message automatically when hitting the panic breakpoints anyways, to be honest.

Initially im thinking to move the unhandled panic breakpoint to somewhere after we print out the message but it's going to be a bit tricky to make it sure it works for older (and future) go versions so I'm going with implementing this on the server side. Hope that's acceptable.

aarzilli (commented):
(this was a leftover comment)

derekparker:
Hey @fatanugraha there has been no update on this PR in a month, are you still planning to work on this?

fatanugraha:
Hi @derekparker I have plans to pick this up again in 1-2 weeks from now, but if anyone else wants to work on this issue I'm also okay with that. Sorry for the delay :(

derekparker:
> Hi @derekparker I have plans to pick this up again in 1-2 weeks from now, but if anyone else wants to work on this issue I'm also okay with that. Sorry for the delay :(

No worries, and no rush! Just wanted to check in.

service/rpc2/server.go
```diff
@@ -132,11 +135,48 @@ func (s *RPCServer) Command(command api.DebuggerCommand, cb service.RPCCallback)
 		cb.Return(nil, err)
 		return
 	}
+
+	// don't show the error if we're not on accept-multi mode as we can assume that
+	// the user will have their client connected at all times.
+	if s.config.AcceptMulti {
```
TomK, line 141: do we need to assume the user has their client connected, when we now have `activeClientCounter` ?

TomK:
@fatanugraha would you like some additional hands on this?
I'm happy to make the changes suggested by @aarzilli

fatanugraha:
Hey folks, sorry for the bad delays. Here's how it looks right now if we hit a breakpoint but the client that sends the `continue` gone already

Screenshot 2024-04-07 at 15 49 09

service/rpc2/server.go
```diff
@@ -19,10 +20,12 @@ type RPCServer struct {
 	config *service.Config
 	// debugger is a debugger service.
 	debugger *debugger.Debugger
+	// activeClientCounter tracks how many active clients we have right now.
+	activeClientCounter *atomic.Int32
```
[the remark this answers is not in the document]
fatanugraha: Sorry, I have the wrong assumption when writing this.

service/rpc2/server.go
```diff
@@ -132,11 +135,48 @@ func (s *RPCServer) Command(command api.DebuggerCommand, cb service.RPCCallback)
...
+			fmt.Println("execution is paused, connect your client to resume execution.")
+		}
+	}
+
 	var out CommandOut
 	out.State = *st
 	cb.Return(out, nil)
 }
 
+func maybeHandlePanicThrowBreakpoint(curThread *api.Thread, debugger *debugger.Debugger) {
```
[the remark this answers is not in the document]
fatanugraha: Done

service/rpc2/server.go
```diff
@@ -132,11 +135,48 @@ func (s *RPCServer) Command(command api.DebuggerCommand, cb service.RPCCallback)
...
+		return
+	}
+
+	switch curbp.Name {
+	case proc.FatalThrow:
+		fmt.Println("fatal error occured in your program, execution is paused.")
+	case proc.UnrecoveredPanic:
+		var stacktrace string
+		frames, err := debugger.Stacktrace(curThread.GoroutineID, 3, api.StacktraceSimple)
+		if err == nil {
```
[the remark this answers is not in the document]
fatanugraha: Done, thanks!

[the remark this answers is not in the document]
fatanugraha: Done, thanks!

service/rpc2/server.go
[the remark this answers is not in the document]
fatanugraha: Done

service/rpc2/server.go
fatanugraha: yup, adressed on the new change. Thanks!

cmd/dlv/dlv_test.go
```diff
@@ -1424,3 +1424,33 @@ func TestUnixDomainSocket(t *testing.T) {
 	client.Detach(true)
 	cmd.Wait()
 }
+
+// TestUnattendedBreakpoint tests whether dlv will print a message to stderr when the client that sends continue is disconnected
+// or not.
+func TestUnattendedBreakpoint(t *testing.T) {
```
aarzilli, line 1430: This fails by timing out every time.

service/rpc2/server.go
```diff
@@ -132,9 +133,49 @@ func (s *RPCServer) Command(command api.DebuggerCommand, cb service.RPCCallback)
 		cb.Return(nil, err)
 		return
 	}
+
 	var out CommandOut
 	out.State = *st
-	cb.Return(out, nil)
+	if clientIsConnected := cb.Return(out, nil); s.config.Headless && !clientIsConnected {
```
aarzilli, line 139: We can't do this in this layer. It will not work if delve is started with `dlv dap` or if someone connects to it using VSCode (or any other program that uses the DAP protocol) and then that client quits.

fatanugraha: Right, I missed that.  I'm leaning towards doing this in both `rpc2` and `dap` or do you suggest that we should do it in the `debugger` layer instead?

cmd/dlv/dlv_test.go
fatanugraha: My bad, done.

service/rpc2/server.go
aarzilli: I lean slightly towards doing it in debugger but both would be ok.

fatanugraha: Hi @aarzilli , correct me if i'm wrong, after checking dap's code I'm not sure if we need to implement this for dap. reasons being:
1. `dap` doesn't support multiclient
2. `dap` will quit when the client disconnected (so printing the log here won't really make sense)

ref:
```
The --client-addr flag is a special flag that makes the server initiate a debug session
by dialing in to the host:port where a DAP client is waiting. This server process
will exit when the debug session ends.
```

aarzilli: You might be right about `dlv dap` but DAP connections are also accepted when delve is started with `dlv --headless debug`. See serveConnectionDemux in service/rpccommon/server.go as well as onDisconnectRequest in service/dap/server.go.

fatanugraha: Ah yeah you're right, thanks for the clarification.

service/dap/server.go
```diff
@@ -524,6 +527,12 @@ func (s *Session) ServeDAPCodec() {
 			s.config.triggerServerStop()
 		}
 	}()
+
+	s.disconnectCh = make(chan struct{})
+	defer func() {
```
derekparker, line 532: Can just `defer close(s.disconnectCh)` without the anonymous function.

derekparker: Actually, maybe the channel should live on the `s.conn` connection and be closed via the above `defer s.conn.Close()`?

derekparker (changes requested):

service/debugger/debugger.go
```diff
@@ -1190,6 +1188,24 @@ func (d *Debugger) Command(command *api.DebuggerCommand, resumeNotify chan struc
 
 	if command.Name != api.SwitchGoroutine && command.Name != api.SwitchThread && command.Name != api.Halt {
 		d.target.ResumeNotify(resumeNotify)
+
+		if clientStatusCh != nil {
+			defer func() {
```
aarzilli, line 1193: I think it's better to move this just before the `return state, err` at the end of the function. I think it's clearer to do it directly instead of using a defer because there is only one path that returns a non nil state and we don't need to actually receive from clientStatusCh in cases where we return an error.

service/rpccommon/server.go
```diff
@@ -412,9 +415,28 @@ func (cb *RPCCallback) Return(out interface{}, err error) {
 		outbytes, _ := json.Marshal(out)
 		cb.s.log.Debugf("(async %d) -> %T%s error: %q", cb.req.Seq, out, outbytes, errmsg)
 	}
+
+	if hasDisconnected := cb.hasDisconnected(); hasDisconnected {
```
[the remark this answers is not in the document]
fatanugraha: Whoops, done!

service/dap/server.go
fatanugraha: Agree, didn't realise s.conn is a wrapper of `net.Conn`

service/debugger/debugger.go
fatanugraha: Done!

fatanugraha:
Sorry folks for the long delay, life got in the way 😅

aarzilli (commented):
Overall this is good, except for a couple of changes that may have been caused by a merge conflict. Also could you remove all the extra newlines that you have added at various points.

service/debugger/debugger.go
```diff
@@ -1321,7 +1320,8 @@ func (d *Debugger) Command(command *api.DebuggerCommand, resumeNotify chan struc
 		}
 		return nil, err
 	}
-	state, stateErr := d.state(api.LoadConfigToProc(command.ReturnInfoLoadConfig), withBreakpointInfo)
+	var stateErr error
```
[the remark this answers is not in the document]
fatanugraha: Ah sorry, i need to revert this since we're not using defer anymore.

service/dap/server.go
```diff
@@ -2746,7 +2756,7 @@ func (s *Session) convertVariableWithOpts(v *proc.Variable, qualifiedNameOrExpr
 				variablesReference = maybeCreateVariableHandle(v)
 			}
 		}
-	case reflect.Struct, reflect.Func:
+	case reflect.Struct:
```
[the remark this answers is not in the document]
fatanugraha: I'm not sure why this got added, removed.

aarzilli (approved):
LGTM, thank you.

TomK:
Great work @fatanugraha 🙏

aarzilli:
Review by Derek.

service/debugger/debugger.go
```diff
@@ -2426,3 +2429,51 @@ var attachErrorMessage = attachErrorMessageDefault
...
+
+	apiFrames, err := d.convertStacktrace(frames, nil)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "err", err)
+		return
+	}
+
+	bp := currentThread.Breakpoint
+	if bp == nil {
+		fmt.Fprintln(os.Stderr, "bp", bp)
```
TomK, line 2457: i don't think this log line is necessary. When attaching a debugger to a long running process we get `bp <nil>` logged out. It'll never be more useful than that.

TomK: @fatanugraha

PR #3637, opened 2024-01-17
proc/native: remove dead code in nonative_darwin.go

aarzilli:
Removes unnecessary code from nonative_darwin.go


derekparker (approved):
lgtm

PR #3642, opened 2024-01-23
proc: skip TestDebugStripped on development versions of Go

aarzilli:
The test can never work on development versions of Go.


derekparker (approved):
LGTM

PR #3655, opened 2024-02-03
cmd/dlv: support unix domain sockets

### LINKED ISSUE #3654: Feature request: Make it possible to listen on a Unix Domain Socket
At the moment, it is only possible to listen to a TCP port.
However, it would be very handy if a domain socket could be used instead. 
Especially in multi-service situations, it would make many things easier. 

For example, if you run multiple services in the headless mode, you must know many ports without any semantics to it. 
If it would be possible to give a Unix Domain Socket, instead, you could have the socket named by the service name. Something like `/tmp/service_x_debug.sock` or so.

Unix Domain Sockets are easy to use in Go.  In the end, you must change:
```
listener, err = net.Listen("tcp", addr)
```
to
```
listener, err = net.Listen("unix", path)
```

I would propose:
1.  a new command line option like `--unix` could be used as an alternative to  `--listen` 
or
2.  the possibility of giving `--listen` a connection string-like format, something like `unix:path`.

aarzilli:
Add support for listening on a unix domain socket in headless mode and
in the 'connect' command.

Fixes #3654


derekparker (approved):
LGTM

PR #3656, opened 2024-02-06
service/dap: support waitfor option for 'dap attach' only

aarzilli:
Add a waitfor option for 'dap attach' that waits for a process with a
given name to appear before attaching to it.

This recovers PR #3584, originally by @muggle-nil, which was fine
except for a broken test.


derekparker (approved):
LGTM

PR #3658, opened 2024-02-11
cmd/dlv: fix --continue with unix domain sockets

### LINKED ISSUE #3657: unix domain sockets only work in headless mode without --continue flag
In #3655 the option that  unix domain sockets  are can be used as listen would be added.
By testing this I have notice that it only work in headless mode without --continue flag.

The reason is that `listener.Addr().String()` is used as input to `rpc2.NewClient` and `connect` in the `execute` command. However,  the String method on a unix address (`net.UnixAddr`) only returns the path without the network ("unix:") as prefix.

Further the `rpc2.NewClient` does connection again directly with a "tcp" network without checking for the prefix before.

aarzilli:
Fixes #3657

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -1076,7 +1076,11 @@ func execute(attachPid int, processArgs []string, conf *config.Config, coreFile 
 	var status int
 	if headless {
 		if continueOnStart {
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

PR #3660, opened 2024-02-13
Don't use a fixed width for s when parsing /proc/net/tcp lines

michaelmerg:
If /proc/net/tcp contains a large number of entries (>9999), parsing lines with len(sl) > 4 fails with the fixed width pattern. In that case the lines are silently ignored.


aarzilli (approved):

derekparker (approved):
LGTM

PR #3663, opened 2024-02-14
*: misc fixes for go1.23

aarzilli:
* skip staticcheck on go1.23 for now
* fix reading interface values in go1.23
* sync list of waitreasons from go1.23


derekparker (approved):
LGTM

PR #3664, opened 2024-02-15
pkg/proc: remove usage of gore dependency

### LINKED ISSUE #3661: Highlight (new) AGPL (Affero General Public License dependency go-delve/gore
Since 2023-11-03 (6c77c3558606bdc8735a815af7520e44962cf719), Delve has an AGPL dependency in the form of:

 1. First pkg/internal/moduledata/moduledata.go (mostly taken from goretk/gore), the separate license is not highlighted here but can be inferred by the attribution to the source project.
 2. Later: Direct dependency on goretk/gore (6e8e1cee9b4b068f0720de2fc0d39d908e00ce15).
 3. Now: A fork of the dependency (go-delve/gore) (dda8f693e6f2f8ad80f615d37ed5fea5e6709999)

The authors can (of course) use any compatible licensed library that they wish. But, the AGPL is often prohibited for use/import in larger companies due to the stringency of the requirements. This means that Delve won't be usable in those places (without patching at least). There may also be other users who'd like to know about the new restrictions.

My request isn't to stop using the library or anything like that (that's up to the authors), but to highlight the new AGPL nature of Delve more prominently somewhere. Thanks!

derekparker:
Implement the specific behavior we used from the gotetk/gore library without copying any code to avoid licensing issues.

Fixes #3661


derekparker:
Currently marking as draft, this PR still needs work but I wanted to push for awareness on the upcoming change.

Still to do:

* Document offsets used when digging through moduledata
* Support 32-bit
* Support Darwin

aarzilli (approved):
LGTM

aktau:
Thanks a lot!

PR #3666, opened 2024-02-17
elfwriter: add WriteSectionHeaders

aarzilli:
Add ability to write out section headers to elfwriter.

### DIFF
--- a/pkg/elfwriter/writer.go
+++ b/pkg/elfwriter/writer.go
@@ -2,7 +2,6 @@
 // contents in memory at any one time.
 // This package is incomplete, only features needed to write core files are
 // implemented, notably missing:
-// - section headers
 // - program headers at the beginning of the file
 
 package elfwriter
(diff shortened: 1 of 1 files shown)


pkg/elfwriter/writer.go
```diff
@@ -135,6 +143,58 @@ func (w *Writer) WriteProgramHeaders() {
...
+		Offset:    uint64(w.Here()),
+		Addralign: 1,
+	}
+	w.Sections = append(w.Sections, strtab)
+
+	strtabStart := w.Here()
+
+	strToIndex := map[string]int64{}
+
+	for _, sect := range w.Sections {
```
derekparker, line 162: Typically the first byte in the `.strtab` section is the null character, is that not necessary here?

aarzilli: You are right, I missed it.

PR #3671, opened 2024-02-23
pkg/terminal,pkg/proc: Implement next-instruction

derekparker:
The next-instruction (nexti) command behaves like
step-instruction (stepi) however, similar to the
`next` command it will step over function calls.


pkg/proc/target_exec.go
```diff
@@ -586,10 +586,28 @@ func (grp *TargetGroup) StepInstruction() (err error) {
...
 		return err
 	}
+	isCall = len(instr) > 0 && instr[0].IsCall()
+	if skipCalls && isCall {
+		// Step into call.
+		err = grp.procgrp.StepInstruction(thread.ThreadID())
+		if err != nil {
+			return err
+		}
+		err = grp.StepOut()
```
aarzilli, line 601: I think the cleanup that's done at the end of the function should be before the call to StepOut (it's fixing the state of the current thread) and isn't needed afterwards (so this could just be `return grp.StepOut()`).

service/api/types.go
```diff
@@ -405,6 +405,8 @@ type DebuggerCommand struct {
 	ReturnInfoLoadConfig *LoadConfig
 	// Expr is the expression argument for a Call command
 	Expr string `json:"expr,omitempty"`
+	// SkipCalls is true if 'step-instruction' commands should step over function calls
```
aarzilli, line 408: I think maybe this should just be a different value of `Name`. If we decide to leave it this way however `SkipCalls` must be added after `UnsafeCalls` for backwards compatibility reasons.

pkg/proc/target_exec.go
```diff
@@ -586,6 +586,20 @@ func (grp *TargetGroup) StepInstruction() (err error) {
...
 	}
+	var isCall bool
+	instr, err := disassembleCurrentInstruction(dbp, thread, 0)
+	if err != nil {
+		return err
+	}
+	isCall = len(instr) > 0 && instr[0].IsCall()
+	if skipCalls && isCall {
+		// Step into call.
+		err = grp.procgrp.StepInstruction(thread.ThreadID())
```
aarzilli, line 597: Suppose the current thread hits a breakpoint and then the user calls `StepInstruction(skipCalls = true)`, this is what happens:

1. The StepInstruction call here is executed, it eventually calls `stepInstruction` in proc/native/threads.go which removes the breakpoint and calls `ptraceSingleStep`, one instruction has been executed
2. Then `StepOut` is called, which calls `Continue`, which calls `ContinueOnce`, which calls `(*processGroup).resume` in proc/native/proc_linux.go
3. The first thing resume does is loop through all the threads and check if their Breakpoint field is set
4. It will find that for the current thread Breakpoint is still set, even though we stepped one instruction, because we didn't do the cleanup, therefore the current thread will be stepped by a second instruction

The cleanup at the end of this function needs to always happen after the `grp.procgrp.StepInstruction` call, I think. I would leave the `isCall` detection as it is, remove this if block and add this at the end of the function:

```
if skipCalls && isCall {
     return grp.StepOut()
}
return nil
```

derekparker: Ack, great point, you're right thanks. Patch updated.

aarzilli (approved):
LGTM

PR #3679, opened 2024-03-04
Cirrus-CI: update to FreeBSD 13.3

aarzilli:


aarzilli:
It looks like the 13.2 image we were using is no longer available.

derekparker (approved):
LGTM

PR #3680, opened 2024-03-04
proc,go.mod: update x/sys remove KeepAlive calls

aarzilli:
### proc,go.mod: update x/sys remove KeepAlive calls

Update version of golang.org/x/sys to the latest version and remove
calls to runtime.KeepAlive calls that were added to the FreeBSD backend
to work around an issue in the version of x/sys that we were using.


derekparker (approved):
LGTM

PR #3682, opened 2024-03-07
proc: support reading captured variables of closures

### LINKED ISSUE #3612: closure captures
It would be useful to get the captured variables when we get a closure.

Right now the only way to guess what is behind a closure is to `x -fmt hex -count 2 -size 8 -x &closure`, copy the second number, then `x -count {n} -size 1 {paste}` and guess the structure of the closure. I don't know if delve can infer the structure behind a closure, but I would prefer to at least make this process a bit simpler and more convenient.

aarzilli:
Supports showing captured variables for function closures on versions
of Go that export informations about the closure struct (Go >= 1.23)

Fixes #3612

### DIFF
--- a/_fixtures/closurecontents.go
+++ b/_fixtures/closurecontents.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func makeAcc(scale int) func(x int) int {
+	a := 0
+	return func(x int) int {
+		a += x * scale
+		return a
+	}
+}
+
+func main() {
+	acc := makeAcc(3)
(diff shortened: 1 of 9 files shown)


derekparker (approved):
LGTM

SOF3:
Does this only show the captured values when printing the closure, or is there a way to actually get them and call methods on the captured values etc?

aarzilli:
> Does this only show the captured values when printing the closure, or is there a way to actually get them and call methods on the captured values etc?

No, but it's a good idea: https://github.com/go-delve/delve/pull/3866

PR #3686, opened 2024-03-20
proc: change 'step' command so that it steps through go statements

aarzilli:
Change 'step' command so that when stepping into a 'go statement' the
debugger will stop on the newly created goroutine, instead of just
stepping over the go statement.


pkg/proc/target_exec.go
```diff
@@ -999,6 +1010,107 @@ func setStepIntoBreakpoint(dbp *Target, curfn *Function, text []AsmInstruction,
...
+			callfile = instr.Loc.File
+			callline = instr.Loc.Line
+			break
+		}
+	}
+	if callfile == "" {
+		logflags.DebuggerLogger().Error("could not find " + runtimeRunqput + " call in " + runtimeNewprocFunc1)
+		return
+	}
+	var pcs []uint64
```
derekparker, line 1045: We only ever use `pcs[0]` after this, so why not just a `var pc uint64` and assign the first address where `pcstmt.Stmt == true`?

derekparker (approved):
LGTM

derekparker:
@aarzilli needs rebase.

aarzilli:
Done

PR #3691, opened 2024-04-02
pkg/proc: implement composite literals

### LINKED ISSUE #1465: Support map evaluation by composite literal keys

```
$ dlv version
Delve Debugger
Version: 1.1.0
```

```
$ go version
go version go1.11.2 linux/amd64
```

Linux/amd64

4. What did you do?

Compile the following program, debug it with delve, try to print some map values:

```go
package main

import "fmt"

type Pair struct {
        A, B int
}

func main() {
        m := map[Pair]string{}
        m[Pair{A: 1, B: 1}] = "x"
        m[Pair{A: 1, B: 2}] = "y"
        fmt.Println(m)
}
```

```
$ dlv debug test.go
(dlv) b main.main:4
...
(dlv) c
...
(dlv) p m[main.Pair{A: 1, B: 1}]
Command failed: expression *ast.CompositeLit not implemented
(dlv) p m[Pair{A: 1, B: 1}]
Command failed: expression *ast.CompositeLit not implemented
```

Print the corresponding map values.

6. What did you see instead?

"Command failed: expression *ast.CompositeLit not implemented"

Looks like CompositeLit evaluation wasn't implemented: https://github.com/go-delve/delve/blob/4c9a72e486f1f0d0c90ecede8415a871dced8117/pkg/proc/eval.go#L258

firelizzard18:
This change implements composite literals for struct, slice, array, and map types.

Fixes #1465

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -252,6 +252,7 @@ func main() {
 	m2 := map[int]*astruct{1: &astruct{10, 11}}
 	m3 := map[astruct]int{{1, 1}: 42, {2, 2}: 43, {1, 0}: 44, {0, 1}: 45, {}: 46}
 	m4 := map[astruct]astruct{{1, 1}: {11, 11}, {2, 2}: {22, 22}}
+	var mint map[int]int
 	upnil := unsafe.Pointer(nil)
 	up1 := unsafe.Pointer(&i1)
 	i4 := 800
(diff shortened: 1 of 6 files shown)


firelizzard18:
My use case is the original ask in #1465 - accessing map elements (e.g. a watch expression in vscode). I'm not sure how much value there is in map and slice literals (since those can't me map keys). The map literals my code makes are at least somewhat broken - they can't be iterated since `Variable.Addr` is zero. I'm not sure how much else is broken.

aarzilli (commented):
This is the wrong approach. It works if all you want to do with the literal is either nothing or a map lookup but taking its address, using it in assignments or passing them to function calls will be broken, either in obvious or subtle ways (for example if the return value of a function call is assigned to the field of a struct we need to make sure that any pointers it might contain are visible to the garbage collector of the target program).

With the way call injection currently works this is approximately impossible, which is why that issue hasn't been closed in such a long time. I've [been working](https://go-review.googlesource.com/c/go/+/558276) on making this possible however.

aarzilli:
Basically, you are creating a Variable that can only be used in very specific expressions. As I said, there isn't a really way to make this work properly, at the moment. Sorry.

firelizzard18:
> As I said, there isn't a really way to make this work properly, at the moment.

Sure, I get that and I'm not surprised my 'solution' doesn't work. I'm not clear on what you'd prefer to do with this PR: close it or leave it open/as a draft until it can be done properly. It doesn't make much of a difference to me either way.

aarzilli:
You should close it, I already have a version privately that does this with the allocation and it's very different from this.

AlexanderMakarov:
> You should close it, I already have a version privately that does this with the allocation and it's very different from this.

@aarzilli are you planning to release your version? Is it possible to download it somewhere?

PR #3693, opened 2024-04-04
pkg/terminal: allow postfix if for breakpoint conds 

derekparker:
Allows for a user to specify the breakpoint condition directly
when creating the breakpoint. The new syntax looks like the
following:

```
break <name> <locspec> [if <expression>]
```

Also updates docs to include more examples and locspec description
instead of directing users to the online / source documentation.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -122,9 +122,32 @@ func DebugCommands(client service.Client) *Commands {
 Type "help" followed by the name of a command for more information about it.`},
 		{aliases: []string{"break", "b"}, group: breakCmds, cmdFn: breakpoint, helpMsg: `Sets a breakpoint.
 
-	break [name] [locspec]
+	break [name] [locspec] [if <condition>]
 
-See Documentation/cli/locspec.md for the syntax of locspec. If locspec is omitted a breakpoint will be set on the current line.
+Locspec is a location specifier in the form of:
+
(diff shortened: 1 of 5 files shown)


derekparker:
Apologies for the formatting noise, editor did that automatically and I noticed after. Tried to keep all that in a separate commit to make it easier to review the core change.

pkg/terminal/command.go
```diff
@@ -1796,11 +1825,23 @@ func formatBreakpointAttrs(prefix string, bp *api.Breakpoint, includeTrace bool)
...
 
 func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) ([]*api.Breakpoint, error) {
+	var (
+		cond string
+		spec string
+
+		requestedBp = &api.Breakpoint{}
+	)
+
+	r := regexp.MustCompile(`^if | if `)
```
aarzilli, line 1835: We accept arbitrary expressions here so this would fail for something like `break *m[" if "]`. I think the only way to do this is to do the same thing we do for `call` and assignments: try to parse the expression as a normal location expression, and if there is a syntax error look up where the error is in the string if it is an `if`. See pkg/proc/evalop/evalcompile.go:59.

aarzilli: PS. another situation where this fails is `break /a/path/that/happens/to/have an if inside/it.go:20`.

derekparker:
Rebased and removed formatting changes not related to this patch.

aarzilli (approved):
LGTM

PR #3696, opened 2024-04-08
cmd/dlv: add shell (non-)completions to flags taking args

scop:
Mostly to avoid the default filename completions to flags that don't operate on filenames. Then again, mark ones that do explicitly as such, and add more specific completions for a number of other cases, too.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


cmd/dlv/cmds/commands.go
```diff
@@ -138,21 +138,48 @@ func New(docCall bool) *cobra.Command {
 	}
 
 	rootCommand.PersistentFlags().StringVarP(&addr, "listen", "l", "127.0.0.1:0", "Debugging server listen address. Prefix with 'unix:' to use a unix domain socket.")
+	if err := rootCommand.RegisterFlagCompletionFunc("listen", cobra.NoFileCompletions); err != nil {
```
aarzilli, line 141: add 

```
func must(err error) {
    if err != nil {
        log.Fatal(err)
    }
}
```

and wrap all of this calls with it (or we could also not check the error).

scop: Done. Would rather check the error as the failure cases are delve side programming errors.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3697, opened 2024-04-09
tests: fix tests on Go 1.23

aarzilli:
* Adjust rtype.go script to handle constants moved to internal/abi from
runtime
* Remove tests in service/dap/server_test that relied on knowledge of
the internal layout of channels.


derekparker (approved):
LGTM

PR #3699, opened 2024-04-10
cmd/dlv: improve positional argument completion

scop:
Avoid default filename completions when not applicable.

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -183,6 +183,12 @@ option to let the process continue or kill it.
 			return nil
 		},
 		Run: attachCmd,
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

aarzilli (commented):
Actually, no. Debug and test can take an arbitrary number of arguments. Also how are the arguments after `--` counted?

aarzilli:
See the test failures, for example.

scop:
Does not seem cobra handles `--` for us re max number of args. Narrowed the PR down to just completion improvements (some of which aren't quite perfect regarding also `--`, but should be improvements over current state anyway).

aarzilli (approved):
I've never liked smart autocompletions and always disable them, so I'll trust you that this is an improvement.

derekparker (approved):
LGTM

Issue #3704, opened 2024-04-15
`rr` parsing broken by recent rr changes


howardjohn:
https://github.com/rr-debugger/rr/commit/07d2aeb7e7a7e9b632681107f4f75602a5362060 changed the output of the `rr replay` command.

Now it looks like
```
Launch debugger with
  'gdb' '-l' '10000' '-ex' 'set sysroot /' '-ex' 'target extended-remote 127.0.0.1:12007' ...
```

I think before it looked like

```
Launch debugger with
  gdb '-l' '10000' '-ex' 'set sysroot /' '-ex' 'target extended-remote 127.0.0.1:12007' ...
```

(notice `gdb` is unquoted)

PR #3705, opened 2024-04-15
rr: fix gdb parsing

howardjohn:
Attempt to fix https://github.com/go-delve/delve/issues/3704

I was having issues testing this locally and wasn't able to understand why. Unfortunately, all machines i have access to require the new `rr`, so its hard for me to test old vs new; I have low confidence this works

### DIFF
--- a/pkg/proc/gdbserial/rr.go
+++ b/pkg/proc/gdbserial/rr.go
@@ -216,7 +216,8 @@ type rrInit struct {
 }
 
 const (
-	rrGdbCommandPrefix = "  gdb "
+	rrGdbCommandLegacyPrefix = "  gdb "
+	rrGdbCommandPrefix = "  'gdb' "
 	rrGdbLaunchPrefix  = "Launch gdb with"
 	targetCmd          = "target extended-remote "
 )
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

pkg/proc/gdbserial/rr.go
```diff
@@ -238,6 +239,11 @@ func rrStderrParser(stderr io.ReadCloser, initch chan<- rrInit, quiet bool) {
 			close(initch)
 			break
 		}
+		if strings.HasPrefix(line, rrGdbCommandLegacyPrefix) {
```
derekparker, line 242: Can you fold this into the above branch since the branch contents are essentially the same aside from the prefix used to reslice.

derekparker: For example:

```go
	prompt := "  'gdb' '-l' '10000' '-ex' 'set sysroot /' '-ex' 'target extended-remote 127.0.0.1:12007'"
	prefix1 := "  gdb "
	prefix2 := "  'gdb' "
	flags := ""
	foundPrefix := false
	if flags, foundPrefix = strings.CutPrefix(prompt, prefix1); !foundPrefix {
		flags, foundPrefix = strings.CutPrefix(prompt, prefix2)
	}
	if foundPrefix {
		initch <- rrParseGdbCommand(flags)
		close(initch)
		break
	}
```

howardjohn: Thanks! Done and pushed up a commit

derekparker (approved):
LGTM

PR #3712, opened 2024-04-23
Add function name even in return trace

archanaravindar:
Currently the trace option prints function names only for call instructions and does not print it for return instructions
Add function name even on return, this will help in pairing call and returns when we eventually add a depth parameter to trace children as well


derekparker (approved):
LGTM

PR #3718, opened 2024-05-01
gdbserial: fixes for rr 5.7.0

aarzilli:
The following fixes have been applied to make delve work with rr 5.7.0

* added a new launch prefix to exclude from stderr output
* allow the thread selection packet to be sent for 'c' commands even
when the stub supports thread suffixes, because the specification is
unclear over what should be done with bc and bs packets with thread
suffixes.

* changed the way qRRCmd are escaped and added a thread selector to
them to match changes to rr codebase


derekparker (approved):
LGTM

PR #3721, opened 2024-05-08
terminal,service: add raw examinemem dump

### LINKED ISSUE #3706: Feature request: dump byte slice memory contents into file
[Using dlv 1.22.1]

Recently, I had to debug a number of failing Go batch pipelines. Because of how these pipelines are structured, it is prohibitively complicated to run them locally, so all I had was a core dump.

With a running program, I would expect `call os.WriteFile("/tmp/b.bin", b, 0600)` to write the contents of the byte slice `b` to `/tmp/b.bin`. Unfortunately this didn’t actually work in my test, but that’s probably worth a separate issue:

```
(dlv) call os.WriteFile("/tmp/b.bin", b, 0644)
> [runtime-fatal-throw] runtime.fatalsignal() third_party/go/gc/src/runtime/signal_unix.go:795 (hits goroutine(1):1 total:1) (PC: 0x558a764f39ae)
Command failed: could not find symbol value for os
```

Anyway, because I only have a core dump, I needed to find another way to dump the contents.

Luckily, I found https://www.youtube.com/watch?v=sV5f1dF8ZU0, which explains how to display the Go slice header:

```
(dlv) x -size 8 -count 3 -x &b
0xc004406e30:   0x000000c005c80000   0x0000000000001d6c   0x000000000002ca2b
```

With the memory address and element count, we can try to display the slice contents:

```
(dlv) x -size 1 -count 0x1d6c 0x000000c005c80000
Command failed: count/len must be a positive integer
(dlv) x -size 1 -count 7532 0x000000c005c80000
Command failed: read memory range (count*size) must be less than or equal to 1000 bytes
```

I removed the 1000 bytes check from the code:
```patch
diff --git i/pkg/terminal/command.go w/pkg/terminal/command.go
index dc629437..4c0c19d6 100644
--- i/pkg/terminal/command.go
+++ w/pkg/terminal/command.go
@@ -2131,9 +2131,9 @@ loop:
 	}
 
 	// TODO, maybe configured by user.
-	if count*size > 1000 {
-		return fmt.Errorf("read memory range (count*size) must be less than or equal to 1000 bytes")
-	}
+	// if count*size > 1000 {
+	// 	return fmt.Errorf("read memory range (count*size) must be less than or equal to 1000 bytes")
+	// }
 
 	if len(args) == 0 {
 		return fmt.Errorf("no address specified")

diff --git i/service/rpc2/server.go w/service/rpc2/server.go
index f806873b..cbcd4d18 100644
--- i/service/rpc2/server.go
+++ w/service/rpc2/server.go
@@ -977,9 +977,9 @@ type ExaminedMemoryOut struct {
 }
 
 func (s *RPCServer) ExamineMemory(arg ExamineMemoryIn, out *ExaminedMemoryOut) error {
-	if arg.Length > 1000 {
-		return fmt.Errorf("len must be less than or equal to 1000")
-	}
+	// if arg.Length > 1000 {
+	// 	return fmt.Errorf("len must be less than or equal to 1000")
+	// }
 	Mem, err := s.debugger.ExamineMemory(arg.Address, arg.Length)
 	if err != nil {
 		return err
```

Now, dlv lets me print the memory contents, which I can redirect into a file using the `transcript` command:

```
(dlv) transcript /tmp/b.hex
(dlv) x -size 1 -count 7532 0xc005c80000
0xc005d80000:   0x68  0x65   0x6c   0x6c   0x6f   0x20   0x63   0x75   
[…]
(dlv) transcript -off
```

However, dlv’s current format is not compatible with `xxd -r`, so would require further work to turn from hex back to binary.

I found it easier to modify dlv’s printing:
```patch
diff --git i/service/api/prettyprint.go w/service/api/prettyprint.go
index 001ab17f..8d98d7e9 100644
--- i/service/api/prettyprint.go
+++ w/service/api/prettyprint.go
@@ -511,12 +511,12 @@ func PrettyExamineMemory(address uintptr, memArea []byte, isLittleEndian bool, f
 		cols = 8
 		colFormat = fmt.Sprintf("%%0%dd", colBytes*3)
 	case 'x':
-		cols = 8
-		colFormat = fmt.Sprintf("0x%%0%dx", colBytes*2) // Always keep one leading '0x' for hex.
+		cols = 16
+		colFormat = "%02x"
 	default:
 		return fmt.Sprintf("not supported format %q\n", string(format))
 	}
-	colFormat += "\t"
+	colFormat += " "
 
 	l := len(memArea)
 	rows := l / (cols * colBytes)
@@ -528,13 +528,14 @@ func PrettyExamineMemory(address uintptr, memArea []byte, isLittleEndian bool, f
 	if l != 0 {
 		addrLen = len(fmt.Sprintf("%x", uint64(address)+uint64(l)))
 	}
-	addrFmt = "0x%0" + strconv.Itoa(addrLen) + "x:\t"
+	addrFmt = "%0" + strconv.Itoa(addrLen) + "x: "
 
 	var b strings.Builder
 	w := tabwriter.NewWriter(&b, 0, 0, 3, ' ', 0)
 
+	startAddress := address
 	for i := 0; i < rows; i++ {
-		fmt.Fprintf(w, addrFmt, address)
+		fmt.Fprintf(w, addrFmt, address-startAddress)
 
 		for j := 0; j < cols; j++ {
 			offset := i*(cols*colBytes) + j*colBytes
```

With this modification, dlv will print like so:

```
(dlv) transcript /tmp/b.hex
(dlv) x -size 1 -count 7532 0xc005c80000
0000000000: 68 65 6c 6c 6f 20 63 75  72 69 6f 75 73 20 67 6f
[…]
(dlv) transcript -off
```

…and then I can use `xxd -r /tmp/b.hex > /tmp/b.bin` (after deleting the command itself from /tmp/b.hex).

---

My feature request is for a command built into delve which can dump memory into a file. It would be tremendously useful and is rather hard to accomplish today (see above). 

Thanks

aarzilli:
Change the examinemem command to have a new format 'raw' that just
prints the raw memory bytes.

Change the transcript command to add a new flag that disables prompt
echo to the output file.

Fixes #3706

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -354,7 +354,7 @@ Examine memory:
 	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <address>
 	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] -x <expression>
 
-Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal).
+Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal) and raw.
 Length is the number of bytes (default 1) and must be less than or equal to 1000.
 Address is the memory location of the target to examine. Please note '-len' is deprecated by '-count and -size'.
 Expression can be an integer expression or pointer value of the memory location to examine.
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -586,7 +586,7 @@ Examine memory:
 	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] <address>
 	examinemem [-fmt <format>] [-count|-len <count>] [-size <size>] -x <expression>
 
-Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal).
+Format represents the data format and the value is one of this list (default hex): bin(binary), oct(octal), dec(decimal), hex(hexadecimal) and raw.
 Length is the number of bytes (default 1) and must be less than or equal to 1000.
 Address is the memory location of the target to examine. Please note '-len' is deprecated by '-count and -size'.
 Expression can be an integer expression or pointer value of the memory location to examine.
@@ -2057,9 +2057,10 @@ func examineMemoryCmd(t *Term, ctx callContext, argstr string) error {
 
 	// Default value
 	priFmt := byte('x')
-	count := 1
-	size := 1
+	count := int64(1)
+	size := int64(1)
 	isExpr := false
+	rawout := false
 
 	// nextArg returns the next argument that is not an empty string, if any, and
 	// advances the args slice to the position after that.
@@ -2085,19 +2086,23 @@ loop:
 			if arg == "" {
 				return errors.New("expected argument after -fmt")
 			}
-			fmtMapToPriFmt := map[string]byte{
-				"oct":         'o',
-				"octal":       'o',
-				"hex":         'x',
-				"hexadecimal": 'x',
-				"dec":         'd',
-				"decimal":     'd',
-				"bin":         'b',
-				"binary":      'b',
-			}
-			priFmt, ok = fmtMapToPriFmt[arg]
-			if !ok {
-				return fmt.Errorf("%q is not a valid format", arg)
+			if arg == "raw" {
+				rawout = true
+			} else {
+				fmtMapToPriFmt := map[string]byte{
+					"oct":         'o',
+					"octal":       'o',
+					"hex":         'x',
+					"hexadecimal": 'x',
+					"dec":         'd',
+					"decimal":     'd',
+					"bin":         'b',
+					"binary":      'b',
+				}
+				priFmt, ok = fmtMapToPriFmt[arg]
+				if !ok {
+					return fmt.Errorf("%q is not a valid format", arg)
+				}
 			}
 		case "-count", "-len":
 			arg := nextArg()
@@ -2105,7 +2110,7 @@ loop:
 				return errors.New("expected argument after -count/-len")
 			}
 			var err error
-			count, err = strconv.Atoi(arg)
+			count, err = strconv.ParseInt(arg, 0, 64)
 			if err != nil || count <= 0 {
 				return errors.New("count/len must be a positive integer")
 			}
@@ -2115,7 +2120,7 @@ loop:
 				return errors.New("expected argument after -size")
 			}
 			var err error
-			size, err = strconv.Atoi(arg)
+			size, err = strconv.ParseInt(arg, 0, 64)
 			if err != nil || size <= 0 || size > 8 {
 				return errors.New("size must be a positive integer (<=8)")
 			}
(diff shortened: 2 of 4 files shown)


aarzilli:
@stapelberg thoughts on this as an implementation for your feature request?

stapelberg:
> @stapelberg thoughts on this as an implementation for your feature request?

Hey, thank you so much for sending this PR!

I just tried it out like so:

```
(dlv) x -size 8 -count 3 -x &b
0xc0045b7db0:   0x000000c005a38000   0x000000000002ca2b   0x000000000002ca2c   
(dlv) transcript -t -x -n /tmp/test.binary
(dlv) x -fmt raw -count 182827 -size 1 0x000000c005a38000
(dlv) transcript -off
```

This seems to work: the byte slice contents match exactly.

It would be convenient if examinemem also accepted hexadecimal `-count` values, so that one can just copy&paste instead of having to do the hex→dec conversion:

```
(dlv) x -fmt raw -count 0x000000000002ca2b -size 1 0x000000c005a38000
Command failed: count/len must be a positive integer
```

Or, even better, if the whole sequence of commands (examine slice header, construct transcript and examinemem commands) could somehow be encapsulated into a “dumpslice” command or similar.

Thanks again

aarzilli:
> It would be convenient if examinemem also accepted hexadecimal -count values

Done

> Or, even better, if the whole sequence of commands (examine slice header, construct transcript and examinemem commands) could somehow be encapsulated into a “dumpslice” command or similar.

That seems a bit too specific to be a command, and it could be added anyway with a user script.

stapelberg:
Thanks for the example! I ended up with this version:

```python
# Syntax: dumpbytearr <var> <outputfile>
def command_dumpbytearr(args):
	var_name, filename = args.split(" ")
	s = eval(None, var_name).Variable
	dlv_command("transcript -t -x -n %s" % filename)
	dlv_command("x -fmt raw -count %d -size 1 %d" % (s.Len, s.Base))
	dlv_command("transcript -off")
```

From my perspective, the PR is ready to be merged — it’s helpful and seems to work in all of my tests :)

derekparker:
Overall the implementation seems fine for the `examinemem` command changes, but I'm unsure about the `transcript` changes. Would that change ever be useful outside of this specific use case? It seems unlikely that you wouldn't want the command echo in any other case.

Perhaps it would be better to omit the change to `transcript` in favor of the `examinemem` starlark builtin accepting the format argument and then combine that with `write_file` in a starlark command?

aarzilli:
If we change the way we convert `[]byte` to a string, like this https://github.com/go-delve/delve/pull/3721/commits/55905239afb70a9b906ed4fda18818bbf621ea83, we don't have to do the change to transcript. The `examinemem` builtin is already returning the "raw" version always.

stapelberg:
Any update on this PR? I still find myself frequently reaching for this functionality, so it would be great to have it included in Delve properly. Thank you!

derekparker:
> If we change the way we convert `[]byte` to a string, like this [5590523](https://github.com/go-delve/delve/commit/55905239afb70a9b906ed4fda18818bbf621ea83), we don't have to do the change to transcript. The `examinemem` builtin is already returning the "raw" version always.

I think this is ok. @stapelberg would this work for you? You'd still have to write a custom command to dump the bytes to a file or something however.

stapelberg:
> I think this is ok. @stapelberg would this work for you? You'd still have to write a custom command to dump the bytes to a file or something however.

Just had a chance to try this out. I used the following `dumpbytearr.star`:

```python
# Syntax: dumpbytearr <var> <outputfile>
def command_dumpbytearr(args):
	var_name, filename = args.split(" ")
	s = eval(None, var_name).Variable
        mem = examine_memory(s.Base, s.Len).Mem
        write_file(filename, mem)
```

And it seems to work fine. So, yes, this works.

Thank you!

aarzilli:
Rebased.

derekparker (approved):
LGTM

Issue #2655, opened 2021-08-11
Support parsing and analyzing heap dumps


derekparker:
Go can produce heap dumps on demand. I think Delve should be able to read and analyze them and produce useful information for the user.

https://pkg.go.dev/runtime/debug#WriteHeapDump
https://github.com/golang/go/wiki/heapdump15-through-heapdump17

hyangah:
cc @randall77

jayantxie:
I have tried viewcore, but considering Delve as the de facto standard Go debugger, I plan to integrate heap analysis functionality into the Delve tool to provide a more stable and user-friendly memory analysis tool. cc @aarzilli 

Is there any update on the support for this proposal? Actually, I am currently working on the development of this.

aarzilli:
Viewing heap memory is a fundamentally different thing than debugging. That is not to say that we couldn't do it (or shouldn't do it) but we would have to integrate the code of viewcore.

The problem with viewcore is that it deeply depends on go runtime internals and often doesn't get updated when they change. Moving it into delve wouldn't change this problem at all.

jayantxie:
Is it possible to leverage the active community of Delve to maintain this functionality in the long term, thereby addressing the stability issues that have arisen due to the historically inactive maintenance of the viewcore tool? 

Or is the current plan of our community to rely on the APIs which are planned to be provided by viewcore for heap analysis integration? https://github.com/golang/go/issues/57447

I would prefer Delve to support viewing heap memory, as it would make problem troubleshooting more convenient compared to using viewcore separately.

aarzilli:
Using dwarf for heap analysis is likely the wrong approach because it won't be able to analyze things kept alive through unsafe.Pointer like the GC does. Before 1.23 it won't be able to analyze closures correctly either. There's probably other edge cases it doesn't cover that I haven't thought about.
A naive implementation using dwarf is also probably excruciatingly slow.
This thing needs to be done the way viewcore does it, which means it needs to be similar to the runtime's GC. And the problem is keeping it in sync with the runtime through the years.

jayantxie:
[...]

PR #3724, opened 2024-05-13
[WIP]feat: add objref subcommand to support outputting object references

jayantxie:
#2655

### DIFF
--- a/pkg/proc/address.go
+++ b/pkg/proc/address.go
@@ -1,0 +1,36 @@
+package proc
+
+// An Address is a location in the inferior's address space.
+type Address uint64
+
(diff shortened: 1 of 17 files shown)


jayantxie:
Complete basic functionality development and debugging, instructions for use are as follows:

1. dlv core ${elf_file} ${core_file}  / dlv attach ${pid}
2. objref
3. go tool pprof -http=:8080 ./objref.out

aarzilli:
As I've stated on the issue I'm opposed to merging something like this since this is the same approach that viewcore took and has been proven to be too hard to maintain in practice.

jayantxie:
@aarzilli I see, I'll move it to a standalone project later.

PR #3726, opened 2024-05-14
proc/gdbserial: add environment variables to configure rr invocation

### LINKED ISSUE #3670: RR chaos mode
Please provide a way to pass the `--chaos` flag to the `rr` backend. Either by a dedicated flag such as `--rr-chaos` or as a general `--rr-flags`, `--backend-flags`, `--backend-opts`, etc.

```
$ rr record -h
 rr record [OPTION]... <exe> [exe-args]...
  [...]
  -h, --chaos                randomize scheduling decisions to try to 
                             reproduce bugs
```

Chaos mode is an extremely useful tool for reproducing weird bugs.

aarzilli:
Adds two environment variables to configure rr invocations.

Fixes #3670

### DIFF
--- a/pkg/proc/gdbserial/rr.go
+++ b/pkg/proc/gdbserial/rr.go
@@ -15,6 +15,11 @@ import (
 	"github.com/go-delve/delve/pkg/proc"
 )
 
+const (
+	delveRecordFlagsEnvVar = "DELVE_RR_RECORD_FLAGS"
+	delveReplayFlagsEnvVar = "DELVE_RR_REPLAY_FLAGS"
+)
+
 // RecordAsync configures rr to record the execution of the specified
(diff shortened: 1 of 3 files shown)


pkg/proc/gdbserial/rr.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/go-delve/delve/pkg/proc"
 )
 
+const (
+	delveRecordFlagsEnvVar = "DELVE_RECORD_FLAGS"
```
derekparker, line 19: I think we should name these `DELVE_RR_*` just in case we ever happen to switch to another record & replay backend which doesn't accept the given flags, and just to make things more explicit.

aarzilli: Done.

Issue #3733, opened 2024-05-24
polish go1.23 range-over-func UX 


hyangah:
Tested with the current master:

![Screenshot 2024-05-24 at 10 27 25 AM](https://github.com/go-delve/delve/assets/4999471/d05dd642-efc2-4187-afaa-96cea96bf0e3)

UX improvement candidates:

* Collapse/hide the range body closure (or give an option like DAP implementation's `HideSystemGoroutines` or config) - Related golang/go#66468

* Hide range closure related variables from the local variable set (e.g. `.closureptr`)

PR #3738, opened 2024-05-30
proc: refactor identifier evaluation for range-over-func support

aarzilli:
Because of how range-over-func is implemented it is difficult to
determine the set of visible local variables during expression
compilation (i.e. it is difficulto to keep the HasLocal function
correct).

This commit moves that logic from expression compilation to expression
evaluation.

Updates #3733

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -203,81 +203,6 @@ func (s scopeToEvalLookup) FindTypeExpr(expr ast.Expr) (godwarf.Type, error) {
 	return s.BinInfo.findTypeExpr(expr)
 }
 
-func (scope scopeToEvalLookup) HasLocal(name string) bool {
-	if scope.Fn == nil {
-		return false
-	}
-
-	flags := reader.VariablesOnlyVisible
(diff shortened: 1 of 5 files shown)


pkg/proc/eval.go
```diff
@@ -680,7 +605,20 @@ func (scope *EvalScope) findGlobal(pkgName, varName string) (*Variable, error) {
...
+
+type errCouldNotFindSymbol struct {
+	name string
+}
+
+func (e *errCouldNotFindSymbol) Error() string {
+	return fmt.Sprintf("could not find symbol %s", e.name)
+}
+
+func isSymbolNotFound(e error) bool {
```
derekparker, line 619: Couldn't this just be `errors.As()`?

aarzilli: No, but I can implement it with `errors.As`, which I have done (although after seeing the implementation of errors.As I'm really tempted to revert it).

derekparker: Yeah, it's not pretty, but if we ever wrap errors further this will be more correct.

derekparker (approved):
LGTM

PR #3740, opened 2024-06-05
proc: fix TestRangeOverFuncNext

aarzilli:
Fix TestRangeOverFuncNext on non-linux operating systems, on arm64 and
when the build mode is PIE.


_scripts/test_linux.sh
```diff
@@ -80,6 +80,7 @@ fi
 
 set +e
 go run _scripts/make.go test
+#go test -run 'TestRangeOverFuncNext' -v github.com/go-delve/delve/pkg/proc
```
derekparker, line 83: Leftover?

aarzilli: Yes, same for the other. Fixed.

derekparker (approved):
LGTM

PR #3741, opened 2024-06-06
refactor: move loadModuleData from runtimeTypeToDIE and expose the apis

jayantxie:


jayantxie:
@aarzilli  Could you help me take a look at these changes to see if they are appropriate? If they are okay, I will add some necessary comments.

aarzilli (commented):
It seems premature to do this for something that isn't released.

jayantxie:
I will later publish  [this repository](https://github.com/jayantxie/goref,) under the organization [cloudwego](https://github.com/cloudwego). However, currently, I can only specify it to use my personal forked repository using the `go mod replace` command. This prevents me from installing it using the "go install" command. Alternatively, could the go-delve/delve repository create a branch for me so that I can directly reference it?

pkg/proc/variables.go
```diff
@@ -102,7 +102,7 @@ type Variable struct {
 	DwarfType godwarf.Type
 	RealType  godwarf.Type
 	Kind      reflect.Kind
-	mem       MemoryReadWriter
+	Mem       MemoryReadWriter
```
aarzilli, line 105: I'd rather not export this at this point, I would suggest that you fish this out with unsafe and write a test for correctness.

pkg/proc/types.go
```diff
@@ -98,19 +98,14 @@ func (ctxt *loadDebugInfoMapsContext) lookupAbstractOrigin(bi *BinaryInfo, off d
...
 
 	// go 1.11 implementation: use extended attribute in debug_info
 
-	mds, err := loadModuleData(bi, _type.mem)
-	if err != nil {
-		return nil, 0, fmt.Errorf("error loading module data: %v", err)
-	}
-
-	md := findModuleDataForType(bi, mds, _type.Addr, _type.mem)
+	md := findModuleDataForType(bi, mds, _type.Addr, _type.Mem)
```
aarzilli, line 108: This change is because you want to cache the module data list right?

pkg/proc/mem.go
```diff
@@ -62,7 +62,7 @@ func CreateLoadedCachedMemory(data []byte) MemoryReadWriter {
 	return &memCache{loaded: true, cacheAddr: fakeAddressUnresolv, cache: data, mem: nil}
 }
 
-func cacheMemory(mem MemoryReadWriter, addr uint64, size int) MemoryReadWriter {
+func CacheMemory(mem MemoryReadWriter, addr uint64, size int) MemoryReadWriter {
```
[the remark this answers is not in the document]
jayantxie: You are right, I made a mistake before by caching local variables, which caused Dereference error. Now, I only cache heap objects, and there is indeed no need to reference them anymore.

pkg/proc/variables.go
jayantxie: Using unsafe to access fields in the short term is also an option. I'll handle it by this way for now.

aarzilli: I'm actually thinking that you shouldn't need this at all. The only time this is meaningfully different from the object you get from calling `proc.(*Target).Memory` is for stack variables, but really you shouldn't rely on those to find heap references: for one thing they might not be alive anymore and for another there could be (large) chunks of memory kept alive from references that are not reachable from any local variable that we have access to. Think for example the return value of a function that just returned.

pkg/proc/bininfo.go
```diff
@@ -505,20 +505,20 @@ type Function struct {
...
 	// extraCache contains informations about this function that is only needed for
 	// some operations and is expensive to compute or store for every function.
-	extraCache *functionExtra
+	extraCache *FunctionExtra
 }
 
-type functionExtra struct {
-	// closureStructType is the cached struct type for closures for this function
-	closureStructType *godwarf.StructType
+type FunctionExtra struct {
```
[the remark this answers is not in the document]
jayantxie: Because parsing the closure type separately from the dwarf requires accessing private functions or variables of the `Image` type, such as `getDwarfTree`, `dwarf`, `index`, etc., it is not easy to obtain them.

pkg/proc/bininfo.go
```diff
@@ -2273,7 +2273,7 @@ func (bi *BinaryInfo) findType(name string) (godwarf.Type, error) {
 	return godwarf.ReadType(image.dwarf, ref.imageIndex, ref.offset, image.typeCache)
 }
 
-func (bi *BinaryInfo) findTypeExpr(expr ast.Expr) (godwarf.Type, error) {
+func (bi *BinaryInfo) FindTypeExpr(expr ast.Expr) (godwarf.Type, error) {
```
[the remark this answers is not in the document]
jayantxie: Functions like `findType`, `expandPackagesInType`, and `findArrayType` are all private functions, and it can be quite challenging to implement these functions externally.

pkg/proc/types.go
jayantxie: Yes, if module data is created every time calling findModuleDataForType, the performance will be very bad.

pkg/proc/variables.go
jayantxie: The 'locals' variable is necessary for memory references. If the memory referenced by these variables becomes invalid and the corresponding heap memory cannot be found, my approach is to skip that variable. And when retrieving references, I also consider the 'args' as a 'locals' variable to prevent the second case.

pkg/proc/bininfo.go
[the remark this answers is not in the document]
jayantxie: @aarzilli Any suggestions? Decoding dwarf externally seems quite complex.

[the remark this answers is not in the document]
aarzilli: You can figure out which image it is by looking at Entry, the only other thing you need at that point is `offset`, which we can make public.

pkg/proc/bininfo.go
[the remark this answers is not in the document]
aarzilli: The only call to FindTypeExpr in the repository you linked that I can find is this:

```
				t, err := parser.ParseExpr("runtime.specialfinalizer")
				if err != nil {
					return err
				}
				spty, _ := special.bi.FindTypeExpr(t)
```

You don't need `FindTypeExpr` for this, you can just use `godwarf.ReadType`.

pkg/proc/variables.go
aarzilli: Judging from what you are using I think using `pkg/dwarf/reader.Variables` and then `BinInfo.Location` is more appropriate, if you want to go the route where you are using DWARF (which I think is the wrong approach).

jayantxie: I'm considering using gc bitmap to traverse objects, using field offset combined with dwarf types to locate downstream types in order to avoid the issue of potentially missing some live memory by purely traversing the dwarf.

jayantxie: > You can't call Locals, it's too high level for what you are doing, you need to read the registers directly or you will not find some live memory. Also, to support range-over-func I will be making changes to `Locals` that will break it completely for your use case.

Get~

aarzilli (approved):
LGTM

jayantxie:
> LGTM

Please help to merge it.

pkg/proc/moduledata.go
```diff
@@ -1,13 +1,13 @@
 package proc
 
-// delve counterpart to runtime.moduledata
-type moduleData struct {
+// ModuleData counterpart to runtime.moduleData
+type ModuleData struct {
```
derekparker, line 4: Why export this when all members are unexported without accessor methods?

derekparker (commented):
One question.

I don't have access to the repo you listed in the earlier comment so couldn't investigate myself.

pkg/proc/moduledata.go
aarzilli: It's for storing it somewhere else and then passing it to RuntimeTypeToDIE.

jayantxie:
> One question.
> 
> I don't have access to the repo you listed in the earlier comment so couldn't investigate myself.

I made the temporary repo private, but I will soon open source the code to the github.com/cloudwego/goref repository.

derekparker (approved):
LGTM

PR #3742, opened 2024-06-06
proc: fix bug with stack watchpoints going out of scope

### LINKED ISSUE #3739: Error clearing stack watchpoint
Backend=rr (5.7.0), on the recent commit with fixes for rr 5.7.0 (https://github.com/go-delve/delve/commit/7c7265f4e68a333bb084217e7b7ae065024e428b) 
```
Delve Debugger
Version: 1.22.1
Build: $Id: 0c3470054da6feac5f6dcf3e3e5144a64f7a9a48 $
```
3. What version of Go are you using? (`go version`)?
`go1.20.1 linux/amd64`
5. What operating system and processor architecture are you using?
Ubuntu 20.04 LTS, amd64

I’m noticing stack watchpoints aren’t cleared the way I’d expect. The CLI session and the target file are in this gist: https://gist.github.com/emilykmarx/c9bb39fdefd2ac8cfb843d53e867a12d, but as a summary:

- I set write watchpoints on the first four elements of a stack array
- The watchpoints hit as expected
- Continuing after the last hit, the watchpoints should all go out of scope
- But an error appears when it tries to clear one of them (index 3, address 0xc000051f28): `could not clear out-of-scope watchpoint`
- Also, the `went out of scope and was cleared` message appears twice for index 3, and does not appear for index 1 (0xc000051f18)
- The watchpoint at index 1 seems to stick around: Continuing once more, the target exits and it is still there (but with a nil address). After restart, it hits a few times in runtime functions – continuing past those, it eventually clears.

aarzilli:
When stack watchpoints go out of scope simultaneously they can hide (or
duplicate the effect) of other breakpoints (including other watchpoints
going out of scope) that are placed on the same physical memory
location.

This happens because we delete the watchpoint-out-of-scope breakpoint
while we are evaluating hit breakpoints, mangling the breaklets list.

This commit moves breakpoint deletion out of the
watchpoint-out-of-scope condition, delaying it until all hit
breakpoints have been evaluated.

Also fix a bug where on amd64 if all four watchpoints are in use the last one is not checked.

Fixes #3739

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -1,9 +1,9 @@
 Tests skipped by each supported backend:
 
-* 386 skipped = 7
+* 386 skipped = 8
 	* 1 broken
 	* 3 broken - cgo stacktraces
-	* 3 not implemented
+	* 4 not implemented
 * arm64 skipped = 1
 	* 1 broken - global variable symbolication
 * darwin skipped = 3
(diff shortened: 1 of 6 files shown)


derekparker (approved):
LGTM

PR #3744, opened 2024-06-08
refactor: use builtin max

Zxilly:


Zxilly:
The CI failed on go 1.20 as it didn't have builtin `max`. But https://github.com/go-delve/delve/commit/d8ed28ff35d6b23c222e6e0fc11bb61f153786db has bump the version requirement to go 1.21. The CI machine should be removed.

aarzilli:
We will drop 1.20 when 1.23 is released, the fact that CI passes on 1.20 means it still works and there is no reason to drop it. This can not be merged until 1.20 is out of the matrix and actually we shouldn't merge it even after because it's too small of an improvement to warrant dropping support for versions of go that would otherwise work.
Sorry.

PR #3745, opened 2024-06-08
ci: remove go 1.20

Zxilly:
https://github.com/go-delve/delve/commit/d8ed28ff35d6b23c222e6e0fc11bb61f153786db bumps the version requirement to 1.21


aarzilli:
We're going to do this when 1.23 is released.

PR #3747, opened 2024-06-16
service: print better message for unattended stops

aarzilli:
Print a better message when an unattended stop happens for reasons
other than hitting a breakpoint.


derekparker (approved):
LGTM

PR #3750, opened 2024-06-17
proc: initial support for expressions with range-over-func

aarzilli:
Supports viewing local variables and evaluating expressions correctly
when range-over-func is used.

The same limitations that the previous commit on this line had still
apply (no inlining, wrong way to identify the range parent in some
cases).

Updates #3733


derekparker (approved):
LGTM

PR #3755, opened 2024-06-25
proc: support stepping through range-over-func statements with inlining

aarzilli:
Extends support for stepping through range-over-func statement to
programs compiled with inlining enabled.

Updates #3733


derekparker (approved):
LGTM

Issue #3760, opened 2024-06-27
panic when data length is greater than memCache size


jayantxie:
1. What version of Delve are you using (`dlv version`)?
v1.22.1
2. What version of Go are you using? (`go version`)?
go1.22
3. What operating system and processor architecture are you using?
linux/amd64
4. What did you do?
Use `*memCache.ReadMemory()` to read data with `33554432` length, and then get the error:

```go
panic: runtime error: slice bounds out of range [140022440016544:92856]

goroutine 1 [running]:
github.com/go-delve/delve/pkg/proc.(*memCache).ReadMemory(0xc00045e700, {0xc002b80000, 0x2000000, 0x2000000}, 0x7f59841ab000)
        /Users/bytedance/go/pkg/mod/github.com/go-delve/delve@v1.22.2-0.20240625150328-a4196f35a95c/pkg/proc/mem.go:50 +0x395
github.com/cloudwego/goref/pkg/proc.(*memCache).ReadMemory(0xc0007064c0, {0xc0002ef188, 0x8, 0x8}, 0x7f59841ab000)
        /Users/bytedance/go/src/pkg/proc/mem.go:38 +0x10e
github.com/cloudwego/goref/pkg/proc.readUintRaw({0x1860cc0, 0xc0007064c0}, 0x7f59841ab000, 0x8)
        /Users/bytedance/go/src/pkg/proc/variables.go:454 +0xdd
```

Code line: https://github.com/go-delve/delve/blob/master/pkg/proc/mem.go#L38

It will overflow when m.cache < size.

PR #3761, opened 2024-06-27
fix: mem cache out of range panic caused by overflow

jayantxie:
fix https://github.com/go-delve/delve/issues/3760

### DIFF
--- a/pkg/proc/mem.go
+++ b/pkg/proc/mem.go
@@ -35,7 +35,12 @@ type memCache struct {
 }
 
 func (m *memCache) contains(addr uint64, size int) bool {
-	return addr >= m.cacheAddr && addr <= (m.cacheAddr+uint64(len(m.cache)-size))
+	end := addr + uint64(size)
+	if end < addr {
+		// overflow
+		return false
+	}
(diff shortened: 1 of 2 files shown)


aarzilli:
See failing test (TestIssue554 in proc_general_test.go).

jayantxie:
> See failing test (TestIssue554 in proc_general_test.go).

Fixed.

aarzilli (approved):

pkg/proc/mem.go
```diff
@@ -35,7 +35,12 @@ type memCache struct {
 }
 
 func (m *memCache) contains(addr uint64, size int) bool {
-	return addr >= m.cacheAddr && addr <= (m.cacheAddr+uint64(len(m.cache)-size))
+	end := addr + uint64(size)
+	if end < addr {
```
derekparker, line 39: This branch is never exercised by any test. Can you add a case for it?

derekparker (changes requested):

pkg/proc/mem.go
jayantxie: Done, please take a look again.

derekparker (approved):
LGTM

PR #3762, opened 2024-06-27
pkg/proc,service/debugger: fix debuginfod-find source

derekparker:
Fixes bug where the incorrect Build ID could be used to try and download the source for a binary. This is because the Build ID was stored on the BinInfo object for a target and not the image itself.


aarzilli (approved):
LGTM

PR #3776, opened 2024-07-11
go.mod: update gopkg.in/yaml to v3

aarzilli:
We are already importing v3 indirectly for github.com/spf13/cobra/doc,
update our own dependency to v3 as well so that we can delete one
module from the vendor directory.


derekparker (approved):
LGTM

PR #3778, opened 2024-07-12
proc: fix bug with range-over-func stepping

aarzilli:
Set a breakpoint on the return address of the current function, if it's
a range-over-func body, and clear the stepping breakpoints for the
current function (except the entry one) when its hit.

Without this what can happen is the following:

1. the range-over-func body finishes and returns to the iterator
2. the iterator calls back into the range-over-func body
3. a stepping breakpoint that's inside the prologue gets hit

Updates #3733

### DIFF
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -144,7 +144,10 @@ const (
 	// goroutine.
 	StepIntoNewProcBreakpoint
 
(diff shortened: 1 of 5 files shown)


aarzilli:
In my opinion we should release 1.23.0 after this. There's more that we could do (fixing the stutter with range-over-func bodies, better support for coroutines) but it's more useful to have a release out that supports 1.23 than doing any of that. We can improve those things in 1.23.1.

derekparker (approved):
LGTM

PR #3780, opened 2024-07-15
goversion: add 1.23 to supported versions, update test matrix

aarzilli:
Add 1.23 to supported versions and to test matrix, remove 1.19 and 1.20
from supported versions (1.19 should have been removed long ago, but we
forgot to do it on release).


aarzilli:
Test failure is expected (it's 1.20 because it is no longer supported).

derekparker (approved):
LGTM

PR #3784, opened 2024-07-19
proc: move stepping test to their own file

aarzilli:
Move all tests of the functionality of proc.next,
proc.(*TargetGroup).Next, proc.(*TargetGroup).Step,
proc.(*TargetGroup).StepOut to their own test file.


derekparker (approved):
LGTM

PR #3787, opened 2024-07-30
proc: adds pointer pinning to call injection

aarzilli:
This commit adds a new mode to call injection. If the runtime.debugPinner
function is available in the target executable it obtains a pinner by
calling it and then uses it to pin the pointers in the results of call
injection.

This allows the code for call injection to be refactored to execute the
calls in the normal order, since it doesn't need to be concerned with having
space on the target's memory to store intermediate values.

Updates #3310

### DIFF
--- a/pkg/proc/dwarf_export_test.go
+++ b/pkg/proc/dwarf_export_test.go
@@ -30,3 +30,14 @@ func NewCompositeMemory(p *Target, pieces []op.Piece, base uint64) (*compositeMe
 func IsJNZ(inst archInst) bool {
 	return inst.(*x86Inst).Op == x86asm.JNE
 }
+
+// HasDebugPinner returns true if the target has runtime.debugPinner.
+func (bi *BinaryInfo) HasDebugPinner() bool {
+	return bi.hasDebugPinner()
+}
+
+// DebugPinCount returns the number of addresses pinned during the last
+// function call injection.
+func DebugPinCount() int {
+	return debugPinCount
+}
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -186,9 +186,17 @@ func GoroutineScope(t *Target, thread Thread) (*EvalScope, error) {
 	return FrameToScope(t, thread.ProcessMemory(), g, threadID, locations...), nil
 }
 
+func (scope *EvalScope) evalopFlags() evalop.Flags {
+	flags := evalop.Flags(0)
+	if scope.BinInfo.hasDebugPinner() {
+		flags |= evalop.HasDebugPinner
+	}
+	return flags
+}
+
 // EvalExpression returns the value of the given expression.
 func (scope *EvalScope) EvalExpression(expr string, cfg LoadConfig) (*Variable, error) {
-	ops, err := evalop.Compile(scopeToEvalLookup{scope}, expr, false)
+	ops, err := evalop.Compile(scopeToEvalLookup{scope}, expr, scope.evalopFlags())
 	if err != nil {
 		return nil, err
 	}
@@ -642,7 +650,7 @@ func (scope *EvalScope) setValue(dstv, srcv *Variable, srcExpr string) error {
 
 // SetVariable sets the value of the named variable
 func (scope *EvalScope) SetVariable(name, value string) error {
-	ops, err := evalop.CompileSet(scopeToEvalLookup{scope}, name, value)
+	ops, err := evalop.CompileSet(scopeToEvalLookup{scope}, name, value, scope.evalopFlags())
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


pkg/proc/eval.go
```diff
@@ -182,9 +182,17 @@ func GoroutineScope(t *Target, thread Thread) (*EvalScope, error) {
 	return FrameToScope(t, thread.ProcessMemory(), g, threadID, locations...), nil
 }
 
+func (scope *EvalScope) evalopFlags() evalop.Flags {
+	flags := evalop.Flags(0)
+	if scope.BinInfo.lookupOneFunc(evalop.DebugPinnerFunctionName) != nil {
```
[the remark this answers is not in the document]
derekparker: Also, this is the same as calling `BinInfo.HasDebugPinner()`.

derekparker (changes requested):

pkg/proc/dwarf_export_test.go
```diff
@@ -30,3 +31,14 @@ func NewCompositeMemory(p *Target, pieces []op.Piece, base uint64) (*compositeMe
 func IsJNZ(inst archInst) bool {
 	return inst.(*x86Inst).Op == x86asm.JNE
 }
+
+// HasDebugPinner returns true if the target has runtime.debugPinner.
+func (bi *BinaryInfo) HasDebugPinner() bool {
```
derekparker, line 36: Maybe this should be promoted outside of test cases since it's used elsewhere.

pkg/proc/fncall.go
```diff
@@ -864,29 +905,33 @@ func funcCallStep(callScope *EvalScope, stack *evalStack, thread Thread) bool {
...
-				break
-			}
-			if err = setLR(thread, oldlr); err != nil {
-				fncall.err = fmt.Errorf("could not restore LR: %v", err)
-				break
+		if fncall.doPinning {
+			stack.callInjectionContinue = false
+			for _, v := range fncall.retvars {
+				saveVariable(v)
+				//fmt.Printf("addr %#v %s %s %#v\n", v.Addr, v.Name, v.DwarfType.String(), v.mem)
```
derekparker, line 919: Leftover

pkg/proc/fncall.go
```diff
@@ -864,29 +905,33 @@ func funcCallStep(callScope *EvalScope, stack *evalStack, thread Thread) bool {
...
+				allPointers(v, "", func(addr uint64, _ string) error {
+					if addr != 0 && pointerEscapes(addr, callScope.g.stack, callScope.callCtx.stacks) {
+						fncall.addrsToPin = append(fncall.addrsToPin, addr)
+					}
+					return nil
+				})
 			}
+			sort.Slice(fncall.addrsToPin, func(i, j int) bool { return fncall.addrsToPin[i] < fncall.addrsToPin[j] })
```
derekparker, line 927: Can you use `slices.Sort` here.

pkg/proc/fncall.go
```diff
@@ -864,29 +905,33 @@ func funcCallStep(callScope *EvalScope, stack *evalStack, thread Thread) bool {
...
+			fncall.addrsToPin = uniqUint64(fncall.addrsToPin)
```
derekparker, line 928: Can use `slices.Compact` here.

derekparker (changes requested):

pkg/proc/dwarf_export_test.go
aarzilli: Where?

aarzilli:
There's a rare failure that seems to be happening with this changeset and windows, I want to investigate it before merging.

pkg/proc/dwarf_export_test.go
derekparker: https://github.com/go-delve/delve/pull/3787/files#diff-59fa7be9add4e1bb71e059c33240b9d77b645837db52c32a18810508d47631e9R320

https://github.com/go-delve/delve/pull/3787/files#diff-b6813f76b09053fe4467f8fe8f18410ca57cca127343a973a43eb9a9de2ce590R187

aarzilli:
> There's a rare failure that seems to be happening with this changeset and windows, I want to investigate it before merging.

For the record: this issue was fixed. It was a preexisting issue exposed by this change where if we have to call injections happening simultaneously it's possible that when we resume execution of the target one of them will not make any progress (the other stops the process before the thread executing the first one is ever scheduled).

When that happened we didn't notice and assumed that both had progressed. The solution is to check for thread.SoftExc() so that we only take into consideration threads for which we received a breakpoint exception.

pkg/proc/dwarf_export_test.go
aarzilli: I've added hasDebugPinner to BinaryInfo. IMO this isn't worth exporting publicly.

derekparker: SGTM

derekparker (approved):
LGTM

PR #3788, opened 2024-08-06
proc: fix step stuttering when entering range-over-func bodies

aarzilli:
Fixes a repeated stop on the first line of a range-over-func body when
stepping through a program. The entry point breakpoint is changed to a
special breakpoint that, when hit:

* clears the current temporary breakpoints
* sets breakpoints like 'next' would, and
* resumes execution.

### DIFF
--- a/pkg/proc/target_exec.go
+++ b/pkg/proc/target_exec.go
@@ -202,7 +202,8 @@ func (grp *TargetGroup) Continue() error {
 
 		switch {
 		case curbp.Active && curbp.Stepping:
-			if curbp.SteppingInto {
+			switch {
+			case curbp.SteppingInto:
 				// See description of proc.(*Process).next for the meaning of StepBreakpoints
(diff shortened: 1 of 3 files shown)


pkg/proc/target_exec.go
```diff
@@ -213,7 +214,18 @@ func (grp *TargetGroup) Continue() error {
...
+				if err := conditionErrors(grp); err != nil {
+					return err
+				}
+				if err := dbp.ClearSteppingBreakpoints(); err != nil {
+					return err
+				}
+				if err := next(dbp, false, false); err != nil {
+					return err
+				}
+				//continue
```
derekparker, line 227: Leftover?

derekparker (approved):
LGTM

pkg/proc/target_exec.go
aarzilli: Actually, no. I meant to say that target's execution continues here. I've expanded it.

PR #3791, opened 2024-08-13
proc: step into coroutine

aarzilli:
The step command is changed such that when the function being currently
called is a coroutine switch function it will move to the associated
coroutine.

Functions that switch coroutines are currently the next, stop and yield
closures produced by the iter.Pull function.


derekparker (approved):
LGTM

PR #3794, opened 2024-08-20
proc: fix result type of division of untyped constants

### LINKED ISSUE #3793: Inconsistent go and delve expression results
Delve Debugger
Version: 1.23.0
Build: $Id: e673f2da02185a1c2b6dea52842b0a8c7e7a03b9 $
go version go1.23.0 linux/amd64
Archlinux in a distrobox container, amd64
4. What did you do?
See example code `example.go`:
```go
package main

import "fmt"

func main() {
        var v float64 = 0.4

        if v > 1/2 {
                fmt.Printf("%v is greater than 1/2\n", v)
        } else {
                fmt.Printf("%v is less than 1/2\n", v)
        }
}
```

Run it with `dlv debug example.go` and place a breakpoint at the line 8 (`if v > 1/2`).
In this example, 1/2 evaluates to 0 and execution takes the `true` branch. But when stopping at the breakpoint and executing `p v > 1/2` delve reports `false`.
I expected executing `p v > 1/2` in delve should report `true`, consistent with go's behavior.
6. What did you see instead?
I saw `false`.

aarzilli:
The Go specification says that the result of division of untyped
constants should be an untyped integer constant if both constants are
untyped integers, the go/constant package however does something
different by always producing an untyped floating constant.

Fixes #3793

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -415,6 +415,7 @@ func main() {
 	enum4 := IntConst
 	enum5 := BoolConst
 	enum6 := FloatConst
+	var zeropoint4 float64 = 0.4
 
 	var amb1 = 1
 	runtime.Breakpoint()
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM after rebase.

PR #3795, opened 2024-08-20
proc: fix TestRangeOverFuncNext on linux/386

aarzilli:
TestRangeOverFuncNext and TestRangeOverFuncNextInlined are failing on
linux/386 with go1.23 due to an inconsistency on how append works on
32bit platforms.


Jille:
LGTM

derekparker (approved):
LGTM

PR #3796, opened 2024-08-20
eval: Allow reslicing a slice up to its cap, rather than its length

Jille:
fix a couple of places where we needed to track the cap

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -690,7 +690,7 @@ func newVariable(name string, addr uint64, dwarfType godwarf.Type, bi *BinaryInf
 		v.Kind = reflect.Array
 		v.Base = v.Addr
 		v.Len = t.Count
-		v.Cap = -1
+		v.Cap = t.Count
 		v.fieldType = t.Type
 		v.stride = 0
 
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -2712,7 +2712,11 @@ func (v *Variable) reslice(low int64, high int64, trustLen bool) (*Variable, err
 	wrong := false
 	cptrNeedsFakeSlice := false
 	if v.Flags&VariableCPtr == 0 {
-		wrong = low < 0 || low > v.Len || high < 0 || high > v.Len
+		if v.Kind == reflect.Slice {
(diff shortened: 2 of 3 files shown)


pkg/proc/stepping_test.go
```diff
@@ -956,7 +956,7 @@ func TestRangeOverFuncNext(t *testing.T) {
 					"result", "[]int len: 0, cap: 0, nil"),
 				nx(87), // if w == 2000
 				assertLocals(t, "result", "w"),
-				assertEval(t, "result", "[]int len: 1, cap: 1, [1000]"),
+				assertEval(t, "result", "[]int len: 1, cap: 2, [1000]"),
```
aarzilli, line 959: If you changed this to make the linux/386 tests pass, the fix is different: see https://github.com/go-delve/delve/pull/3795

pkg/proc/variables.go
```diff
@@ -690,7 +690,7 @@ func newVariable(name string, addr uint64, dwarfType godwarf.Type, bi *BinaryInf
 		v.Kind = reflect.Array
 		v.Base = v.Addr
 		v.Len = t.Count
-		v.Cap = -1
```
aarzilli, line 693: Why are we changing this?

Jille: I thought it was more correct to record the actual capacity of an array rather than -1. Go doesn't allow us to call `cap()` on a string, but it does allow it on an array, so I thought it'd be cleaner to have it.

Alternatively we can give reslicing a special case for arrays too, but this felt cleaner in the long run.

Let me know if you'd prefer this to remain -1 and I'll change reslice instead.

aarzilli: Ok, I think it's fine to change it.

aarzilli (approved):
LGTM

Jille:
@SnoozeThis https://github.com/go-delve/delve/pull/3795

SnoozeThis:
(https://snoozeth.is/dZ3T-dmNzAY) I will wait until https://github.com/go-delve/delve/pull/3795 is merged and then add a comment.

SnoozeThis:
Resolved.

Jille:
I've rebased on master.

derekparker (approved):
LGTM

PR #3798, opened 2024-08-27
service/dap: make handlesMap generic

aarzilli:
The oldest version of Go we compile with is 1.20 at this point, we can
thus make use of generic type parameters and the handlesMap support
type in service/dap is improved by them.


derekparker (approved):
This is great!

LGTM

PR #3799, opened 2024-08-27
proc: workaround for macOS section name truncation

### LINKED ISSUE #3797: DW_FORM_strx with no .debug_str_offsets section
Minimal code to reproduce an issue: https://go.dev/play/p/mT1pjsWIx9x. The same issue happens when using this dependency:
```
helm.sh/helm/v3/pkg/cli/values
```

`go build` works normally with no errors.

```sh
Delve Debugger
Version: 1.23.0
Build: $Id: e673f2da02185a1c2b6dea52842b0a8c7e7a03b9 $
```
```
go version go1.23.0 darwin/arm64
```
MacOS on ARM, issue started to be reproduced since MacOS 15 Beta(24A5327a).
4. What did you do?
run dlv debugger via VSCode and manually
Debugger is working for my project. 
6. What did you see instead?
The following error:
```
2024-08-20T21:03:44+02:00 error layer=debugger error loading binary "/Users/mariiak/repos/mycli/__debug_bin2843838000": error reading debug_info: decoding dwarf section info at offset 0x942ab7: DW_FORM_strx with no .debug_str_offsets section
could not launch process: error reading debug_info: decoding dwarf section info at offset 0x942ab7: DW_FORM_strx with no .debug_str_offsets section
```
I wasn't able to dissect what exactly is causing the issue - any directions are appreciated!

I also found this issue with similar problem that is happening on MacOS 15 Beta https://github.com/mattn/go-sqlite3/issues/1269

aarzilli:
Go 1.23 and earlier does not take into account that Mach-O section
names are limited to 16 characters, which causes DWARF sections with
long names to be truncated and become unreadable.

Fixes #3797

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -1971,6 +1971,7 @@ func loadBinaryInfoMacho(bi *BinaryInfo, image *Image, path string, entryPoint u
 		return &ErrUnsupportedArch{os: "darwin", cpuArch: exe.Cpu}
 	}
 	var dwerr error
+	macOSShortSectionNamesWorkaround(exe)
 	image.dwarf, dwerr = exe.DWARF()
 	if dwerr != nil {
 		if len(bi.Images) <= 1 {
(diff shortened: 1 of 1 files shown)


AntonPieper:
Looking at the Wikipedia article of Mach-O, there seems to be [two 16-Bit fields for section names](https://en.m.wikipedia.org/wiki/Mach-O#:~:text=Segment%20section32/64). Could it be, that either go or macOS 15 beta does not correctly read/write into the second section name field?

derekparker (approved):
LGTM

linzhp:
Can you create a new release of Delve to include this?

twz915:
https://github.com/mattn/go-sqlite3/issues/1269#issuecomment-2333030995

PR #3800, opened 2024-08-29
proc: cache module data

aarzilli:
Cache module data so that we don't reload it every time we look up a
variable with a generic type.

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -260,6 +260,7 @@ func (t *Target) SupportsFunctionCalls() bool {
 func (t *Target) ClearCaches() {
 	t.clearFakeMemory()
 	t.gcache.Clear()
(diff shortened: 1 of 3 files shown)


pkg/proc/target.go
```diff
@@ -78,6 +78,8 @@ type Target struct {
 	fakeMemoryRegistry    []*compositeMemory
 	fakeMemoryRegistryMap map[string]*compositeMemory
 
+	moduleDataCache []ModuleData
```
derekparker, line 81: Maybe this type would make more sense on `BinaryInfo` directly so we don't have to add the extra param below.

aarzilli: Done.

derekparker (approved):
LGTM

Issue #3804, opened 2024-08-31
MacOS M3 could not launch process: stub exited while waiting for connection: exit status 0


exosyphon:
I have run through many different steps to getting dlv installed and configured inside Neovim. At this point I cannot run dlv even on the CLI. It is likely user error but I simply don't know what else to do at this point and can use some help since I am new to Golang.

1. What version of Delve are you using (`dlv version`)? 
Delve Debugger
Version: 1.23.0
Build: $Id: e673f2da02185a1c2b6dea52842b0a8c7e7a03b9 $

2. What version of Go are you using? (`go version`)?
go version go1.23.0 darwin/arm64

3. What operating system and processor architecture are you using?
Apple M3 Pro (Apple Silicon)
Sonoma 14.6.1

4. What did you do?
I used asdf to install golang (confirmed that it is arm64 using go version)
installed dlv using these commands after setting these environment variables in my ~/.zshrc:
export GOOS="darwin"
export GOARCH="arm64"
export GOPATH="$HOME/go"
export PATH="$GOPATH/bin:$PATH"

go install github.com/go-delve/delve/cmd/dlv@latest
go install github.com/aarzilli/gdlv@latest

I also already uninstalled and reinstalled xcode and ran through each of the MacOS commands:
xcode-select --install
sudo /usr/sbin/DevToolsSecurity -enable
sudo dscl . append /Groups/_developer GroupMembership $(whoami)

5. What did you expect to see?
When I run dlv debug hello.go to not see an error

6. What did you see instead?
could not launch process: stub exited while waiting for connection: exit status 0

aarzilli:
There is a plethora of similar issues: https://github.com/go-delve/delve/issues?q=is%3Aissue+is%3Aclosed+debugserver++%22stub+exited%22
It's some kind of configuration problem, possibly caused by some amd64 emulation layer running without your knowledge or possibly caused by a broken installation of xcode. The fact that you are explicitly setting GOOS and GOARCH is suspect, these should never be set manually unless you are cross-compiling.

exosyphon:
I was able to figure this out. 

I ran `arch` and the output was `i386` in tmux and Kitty. This should have been `arm64`. I confirmed by running `arch` in Terminal and got `arm64`.

I believe when I initially installed Kitty initially it picked up the wrong architecture from iTerm2.

I re-installed Kitty from inside Terminal:
`curl -L https://sw.kovidgoyal.net/kitty/installer.sh | sh /dev/stdin`

[...]

PR #3810, opened 2024-09-13
proc: improve Rosetta check

aarzilli:
We have a check for Rosetta that should point users towards
misconfiguration, however occasionally we still get new issues about
debugserver crashing and some of those, as it turns out, are still
caused by Rosetta (see for example #3804).

Check the output of 'uname -a' and check that it isn't 'x86_64' if we
are an 'arm64' process: if that happens we are running unemulated but
debugserver will refuse to work.

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -297,6 +297,9 @@ func (p *gdbProcess) Listen(listener net.Listener, path, cmdline string, pid int
 		return p.Connect(conn, path, cmdline, pid, debugInfoDirs, stopReason)
 	case status := <-p.waitChan:
 		listener.Close()
+		if err := checkRosettaExpensive(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/proc/gdbserial/gdbserver.go
```diff
@@ -2228,3 +2234,29 @@ func machTargetExcToError(sig uint8) error {
...
+		return nil
+	}
+	if runtime.GOARCH != "arm64" {
+		return nil
+	}
+
+	// Additionally check the output of 'uname -a' if it's x86_64 it means that
+	// the shell we are running on is being emulated by Rosetta even though our
+	// process isn't. In this condition debugserver will crash.
+	out, err := exec.Command("uname", "-a").Output()
```
derekparker, line 2249: If you do `uname -m` you don't have to do the string munging below.

derekparker (approved):
LGTM

PR #3830, opened 2024-10-11
Update README.md

adriancuadrado:
Fixed all html tags in the markdown

Many things in the README.md don't make sense at all because the original author accidentally wrote html tags that won' render.

For instance:

image

I think we have to review all markdown files in the project for similar issues like this one.

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -112,16 +112,16 @@ Sets a breakpoint.
 
 Locspec is a location specifier in the form of:
 
-  * *<address> Specifies the location of memory address address. address can be specified as a decimal, hexadecimal or octal number
-  * <filename>:<line> Specifies the line in filename. filename can be the partial path to a file or even just the base name as long as the expression remains unambiguous.
-  * <line> Specifies the line in the current file
-  * +<offset> Specifies the line offset lines after the current one
-  * -<offset> Specifies the line offset lines before the current one
-  * <function>[:<line>] Specifies the line inside function.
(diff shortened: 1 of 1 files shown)


aarzilli:
That file is autogenerated, the script needs to be fixed.

adriancch-inditex:
https://github.com/adriancuadrado/delve/blob/04129b9b599e933db2f7d74972f4589acd15d3f4/pkg/terminal/docgen.go#L41-L79

It's just a matter of replacing all `<` with `&lt;` and `>` with `&gt;`.

adriancuadrado:
https://github.com/adriancuadrado/delve/blob/04129b9b599e933db2f7d74972f4589acd15d3f4/pkg/terminal/docgen.go#L41-L79

It's just a matter of replacing all `<` with `&lt;` and `>` with `&gt;`.

adriancuadrado:
Apparently you can write code between 2 pairs of 2 backticks to escape backticks inside, like so:
```
``This is a backtick --> ` ``
```
That would result in...
``This is a backtick --> ` ``

I'm just pointing out this to make sure this particular use case is considered.

More details here: https://stackoverflow.com/a/71150969/5744858

aarzilli:
The source isn't markdown.

adriancuadrado:
@aarzilli
> The source isn't markdown.

I know? The source isn't Markdown but the output of the script is, so you have to modify the script to escape &lt; and &gt; before writing.

Either that, or I'm not understanding what you meant.

PR #3834, opened 2024-10-14
pkg/proc/core: Clean up some repetitive code

aarzilli:
Cleans up repetitive code in pkg/proc/core and also adds coredumpctl
support to our test suite.


derekparker (approved):
LGTM

PR #3836, opened 2024-10-15
Documentation: fix autogenerated markdown documentation

### LINKED ISSUE #3829: Delve's source code quality: There seems to be a pointless check that always returns true
https://github.com/go-delve/delve/blob/abc948b0ae93b74d4aed3cf089cb44b9d7f8c255/pkg/locspec/locations.go#L181-L184

https://pkg.go.dev/strings#Split

image

I was reading the source code to understand the syntax to set breakpoints in functions and I discovered this. I am not an expert in golang, but my understanding is that this code would always return true right?

aarzilli:
Changes < to &lt; in autogenerated documentation for CLI commands so that markdown does not interpret them as HTML tags.

Supersedes #3830
Fixes #3829

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -112,26 +112,26 @@ Sets a breakpoint.
 
 Locspec is a location specifier in the form of:
 
-  * *<address> Specifies the location of memory address address. address can be specified as a decimal, hexadecimal or octal number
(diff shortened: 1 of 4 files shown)


derekparker (approved):
LGTM

PR #3837, opened 2024-10-18
goversion: parse version for development builds

aarzilli:
Once upon a time Go version strings for development builds did not
reference the major/minor version they belonged to, therefore the best
we could do was to always assume that a development build belonged to
the most recently released version of Go.

This commit changes pkg/goversion so that the new style of development
version string is fully parsed.

### DIFF
--- a/pkg/goversion/compat.go
+++ b/pkg/goversion/compat.go
@@ -21,7 +21,7 @@ var (
 // this version of delve.
 func Compatible(producer string, warnonly bool) error {
 	ver := ParseProducer(producer)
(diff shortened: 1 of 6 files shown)


aarzilli:
I want to use this in the (future) PR to support swisstables. I want to add a check to the swisstables test so that it will start failing when 1.25 is released, so that we can remember to change it when swisstables become the default but I can't do that if we can't distinguish 1.24 development builds from 1.26 development builds (which we can not without this change).

derekparker (approved):
LGTM

PR #3840, opened 2024-10-23
pkg/proc/native,pkg/proc/amd64util: xsave decoding cleanup

aarzilli:
* move the cpuid querying code to pkg/proc/native/cpuid since
pkg/proc/native is the only package entitled to calling it
* add a type to describe the xstate_bv bitmap instead of using
hardcoded constants everywhere
* use xcr0 instead of xstate_bv for the offset heuristic like gdb does


derekparker (approved):
LGTM

PR #3844, opened 2024-10-30
TeamCity: add linux/loong64 builders

abner-chenc:


abner-chenc:
The Loong64 proxy is already running, and the configuration is as follows:

```
serverUrl=https://delve.teamcity.com
name=Linux-Fedora38-loong64
authorizationToken=
```

aarzilli:
There is no point in merging this separately from #3685 (which at the moment has merge problems and test failures).

PR #3849, opened 2024-11-04
proc: fix rangeFuncStackTrace with inlined functions

aarzilli:
If an inlined function is encountered we should keep searching for its
rangeParent even if we don't have a closurePtr because it could be that
the function has been inlined into its rangeParent.

This does not need a new test, the existing tests already fail on
go1.24.


derekparker (approved):
LGTM

PR #3870, opened 2024-11-25
feat: print a message when the debugee process exits

jakejx:
This MR makes a simple change to print a message when the debugee process exits. This is useful when debugging a process that is abruptly killed (for e.g. OOM). Currently, it will just look like the process has "hung".


aarzilli (commented):
The client already prints something on exit. Is this because of headless instances?

jakejx:
Hello, yeap the main use case is for headless instances and when `--continue` is used. I did notice that on MacOS, LLDB does print an exit message. But this does not happen when using `dlv` on Linux.

aarzilli:
Then it needs to work like the `maybePrintUnattendedBreakpointWarning` (see the call to it below). In fact, it should probably be renamed `maybePrintUnattendedStopWarning` and be called on both places.

jakejx:
Thanks, let me check those out.

jakejx:
Hi @aarzilli, I have updated the PR with your suggestion. Hopefully I did not misunderstand your suggestion.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3872, opened 2024-12-03
terminal/starbind: allow modification of structs returned by API

aarzilli:
Structs returned to starlark scripts by API calls were immutable, this
made amend_breakpoint nearly impossible to use since its argument must
be a api.Breakpoint struct which the caller has received from
get_breakpoint and modified.


derekparker (approved):
LGTM

PR #3874, opened 2024-12-05
proc: expose breakpoint hitcounts in expressions

aarzilli:
Expose breakpoint hitcounts in the expression language through the
special variable runtime.bphitcount:
runtime.bphitcount[1]
runtime.bphitcount["bpname"]
will evaluate respectively to the hitcount of breakpoint with id == 1
and to the hitcount of the breakpoint named "bpname".

This is intended to be used in breakpoint conditions and allows
breakpoints to be chained such that one breakpoint is only hit after a
different is hit first.

A few optimizations are implemented so that chained breakpoints are
evaluated efficiently.

### DIFF
--- a/Documentation/cli/expr.md
+++ b/Documentation/cli/expr.md
@@ -119,6 +119,7 @@ Delve defines two special variables:
 
 * `runtime.curg` evaluates to the 'g' struct for the current goroutine, in particular `runtime.curg.goid` is the goroutine id of the current goroutine.
 * `runtime.frameoff` is the offset of the frame's base address from the bottom of the stack.
+* `delve.bphitcount[X]` is the total hitcount for breakpoint X, which can be either an ID or the breakpoint name as a string.
 
 ## Access to variables from previous frames
 
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -1253,6 +1253,9 @@ func (stack *evalStack) executeOp() {
 	case *evalop.PushDebugPinner:
 		stack.push(stack.debugPinner)
 
+	case *evalop.PushBreakpointHitCount:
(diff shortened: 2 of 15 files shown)


Documentation/cli/expr.md
```diff
@@ -119,6 +119,7 @@ Delve defines two special variables:
 
 * `runtime.curg` evaluates to the 'g' struct for the current goroutine, in particular `runtime.curg.goid` is the goroutine id of the current goroutine.
 * `runtime.frameoff` is the offset of the frame's base address from the bottom of the stack.
+* `runtime.bphitcount[X]` is the total hitcount for breakpoint X, which can be either an ID or the breakpoint name as a string.
```
derekparker, line 122: This is more for discussion, but I'm not crazy about this going into the `runtime` pseudo package we create for these special variables. It makes sense for the other two, as they are more related to the Go runtime (especially `runtime.curg`) however `bphitcount` is an internal (to Delve) related variable, so it seems odd to lump them together. Then again, is it worth users having to remember another pseudo package prefix?

pkg/proc/eval.go
```diff
@@ -2064,6 +2067,33 @@ func (scope *EvalScope) evalIndex(op *evalop.Index, stack *evalStack) {
 		return
 	}
 
+	if xev.Name == "runtime."+evalop.BreakpointHitCountVarName {
```
derekparker, line 2070: We do this construction of `runtime.bphitcount` repeatedly, perhaps we just have a constant for the fully qualified name.

aarzilli: Done.

Documentation/cli/expr.md
aarzilli: Do you have something in mind? Should we call it `delve.bphitcount`? I don't have a strong opinion about this.

derekparker: I like that, let's go with that.

aarzilli: Done.

derekparker (approved):
LGTM

PR #3879, opened 2024-12-14
proc: fix loading goroutine labels on Go 1.24

aarzilli:
Go 1.24 changed the way goroutine labels are stored.


derekparker (approved):
LGTM

PR #3882, opened 2024-12-17
goversion,proc_test: add go1.24 compatiblity, disable broken test

aarzilli:
Add go1.24 to compatibility set and test matrix.

Remove go1.21 from test matrix.

TestRangeOverFuncNextInlined is disabled because the improved inlining
of go1.24 produces symbol names that are too complicated for us to
correlate, a fix for this will require a change to the compiler that
will probably be too complex to make it into 1.24.


derekparker (approved):
LGTM

PR #3887, opened 2024-12-19
fix: wrong returned errors

ldez:
There are 2 cases where the returned error is not the right one.


aarzilli:
You have to fight with #3886, whoever wins get the PR merged.

PR #3890, opened 2024-12-27
service/test: disable TestGuessSubstitutePath on TeamCity/linux/tip

aarzilli:
It does not work because of how TeamCity creates the build directory.

### DIFF
--- a/service/test/integration2_test.go
+++ b/service/test/integration2_test.go
@@ -3156,6 +3156,14 @@ func TestBreakpointVariablesWithoutG(t *testing.T) {
 
 func TestGuessSubstitutePath(t *testing.T) {
 	t.Setenv("NOCERT", "1")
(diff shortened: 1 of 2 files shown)


service/test/integration2_test.go
```diff
@@ -3156,6 +3156,11 @@ func TestBreakpointVariablesWithoutG(t *testing.T) {
 
 func TestGuessSubstitutePath(t *testing.T) {
 	t.Setenv("NOCERT", "1")
+	ver, _ := goversion.Parse(runtime.Version())
+	if ver.IsDevelBuild() && os.Getenv("CI") != "" && runtime.GOOS == "linux" {
+		t.Skip("does not work in TeamCity + tip + linux")
```
derekparker, line 3161: Can you add a quick comment for reference on the TeamCity issue that's causing this failure.

PR #3891, opened 2024-12-29
proc: remove irrelevant TODO comment

aarzilli:
The arm64 disassembler is in fact already compiled in unconditionally.


derekparker (approved):
LGTM

PR #3894, opened 2025-01-01
pkg/proc/internal/ebpf: Fix uprobe address

3u13r:
The update of gilium/ebpf introduced a beaking change in it's API: The `offset` in the UprobeOptions is now relative to the added option `address`. Since `address` was only default initalized, the library did not use `address` and `offset` as address for the uprobe, but tried to calculate the offset itself based on the given symbol. Since we set the path of the executable as symbol, the library errored when trying to resolve it.

Fixes https://github.com/go-delve/delve/pull/3491

How to test this PR: Run `sudo -E go run _scripts/make.go test -v --test-set cmd/dlv --test-run='^.*EBPF.*$' ` on the main branch, notice how it fails and re-run on this branch.


aarzilli (approved):

derekparker (approved):
LGTM

PR #3899, opened 2025-01-08
pkg/terminal: support zed editor and allow args in DELVE_EDITOR

### LINKED ISSUE #3627: `DELVE_EDITOR` should support extra arguments
As of now, `DELVE_EDITOR` expects just a binary name, but it would be useful if it could also accept extra arguments. For example, with Emacs, I would want the editor to be something like `emacsclient -c` so that it creates a new frame. This is the behavior with `GIT_EDITOR` btw. I think the [logic in git](https://github.com/git/git/blob/a26002b62827b89a19b1084bd75d9371d565d03c/editor.c#L58) is to split the env var by space. If this sounds good, I can probably drop a PR with the change. As a workaround I currently have it call a shell script which in turn calls emacsclient with proper args.

---

1. What version of Delve are you using (`dlv version`)? ... `1.21.0`
2. What version of Go are you using? (`go version`)? ... `go1.21.3`
3. What operating system and processor architecture are you using? ... `linux/amd64`
4. What did you do? ... Run edit at a breakpoint with `DELVE_EDITOR` as `emacsclient -n`
5. What did you expect to see? ... Open Emacs to the location
6. What did you see instead? ... Fails with message that `emacsclient -n` is not a binary

derekparker:
Enhance editor support for the DELVE_EDITOR environment variable

1. Add support for the Zed editor by recognizing the 'zed' command and setting the appropriate line number syntax

2. Allow users to specify editor-specific arguments in DELVE_EDITOR by parsing the env var value into command and args parts

3. Modify runEditor() to pass through any user-specified args before appending the file/line args

This change allows more configuration flexibility through DELVE_EDITOR while maintaining backwards compatibility. The Zed editor is now supported alongside VSCode, Helix and Vim variants.

Fixes #3627

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1960,22 +1960,32 @@ func tracepoint(t *Term, ctx callContext, args string) error {
 	return err
 }
 
-func getEditorName() (string, error) {
+func getEditorName() (string, []string, error) {
 	var editor string
 	if editor = os.Getenv("DELVE_EDITOR"); editor == "" {
 		if editor = os.Getenv("EDITOR"); editor == "" {
-			return "", errors.New("Neither DELVE_EDITOR or EDITOR is set")
+			return "", nil, errors.New("Neither DELVE_EDITOR or EDITOR is set")
 		}
 	}
-	return editor, nil
+
+	editorParts := strings.Fields(editor)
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

PR #3902, opened 2025-01-14
service/dap: provide process events after launch

shaunduncan:
This change adds support to emit the process DAP event after handling launch requests.

Note: I found myself in a situation where I needed this exact event get process information from delve. I discovered this event was not implemented, so I have added it in the place that seems to be the most logical place (just based on my limited exposure to the codebase). Please let me know if this change is unwarranted or if useful, if it needs any additional work.

Thanks!


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

Issue #3906, opened 2025-01-19
Long pause when debugging a program Raylib or SDL


lilliemarck:
When stepping over, or setting a breakpoint after `InitWindow` the debugger pauses for about 15-30 seconds, depending on which computer I use. The dlv process consumes 5 GB of memory during the pause.

Same thing happens with programs written in [SDL](https://github.com/veandco/go-sdl2) or when writing a simple Cgo binding to Raylib myself. Both these libraries and my code do the `runtime.LockOSThread()` thing.

The problem is partially fixed by debugging with Goland but only if I use the Flatpak install. With Goland Flatpak the debugger freezes the first run and then only occasionally. Unfortunately I can only run my SDL programs like this but Raylib, which defaults to using Wayland, doesn't work at all unless I switch to the .tar.gz istall. I had other issues with VS Code too so I am not surprised Flatpak is not suitable for development, except for it doesn't freeze as often...

I am guessing it has something to do with lots of shared libraries being loaded.

```
go version go1.23.4 linux/amd64
 
Delve Debugger
Version: 1.24.0
Build: $Id: 2d55e4d30150e3fc68e64357e21293d7920fc1f1 $
```

main.go
```Go
package main

import rl "github.com/gen2brain/raylib-go/raylib"

func main() {
	println("No problem on this line")

	rl.InitWindow(800, 450, "raylib [core] example - basic window")
	defer rl.CloseWindow()

	rl.SetTargetFPS(60)

	for !rl.WindowShouldClose() {
		rl.BeginDrawing()
		rl.ClearBackground(rl.RayWhite)
		rl.DrawText("Congrats! You created your first window!", 190, 200, 20, rl.LightGray)
		rl.EndDrawing()
	}
}
```

go.mod
```Go
module basic_window

go 1.23.4

require github.com/gen2brain/raylib-go/raylib v0.0.0-20250109172833-6dbba4f81a9b

require (
	github.com/ebitengine/purego v0.7.1 // indirect
	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842 // indirect
	golang.org/x/sys v0.20.0 // indirect
)
```

lilliemarck:
I guess an example with Cgo is better because it doesn't need the binding libraries. Just install the raylib dev package  `dnf install raylib-devel`.

```Go
package main

import (
	"runtime"
	"unsafe"
)

/*
#include <raylib.h>
#include <stdlib.h>
#cgo LDFLAGS: -lraylib
*/
[...]

PR #3907, opened 2025-01-19
proc/debuginfod: add timeouts to debuginfod-find

aarzilli:
Calls to debuginfod-find can take an arbitrarily long time to complete.

Set timeouts for them unless they were explicitly set by user, also
redirect its stderr to our stderr so that, if they produce progress
output, it is seen.

Updates #3906

### DIFF
--- a/pkg/proc/debuginfod/debuginfod.go
+++ b/pkg/proc/debuginfod/debuginfod.go
@@ -1,17 +1,26 @@
 package debuginfod
 
 import (
+	"os"
 	"os/exec"
 	"strings"
 )
 
-const debuginfodFind = "debuginfod-find"
+const (
+	debuginfodFind       = "debuginfod-find"
+	debuginfodMaxtimeEnv = "DEBUGINFOD_MAXTIME"
+	debuginfodTimeoutEnv = "DEBUGINFOD_TIMEOUT"
+)
 
 func execFind(args ...string) (string, error) {
 	if _, err := exec.LookPath(debuginfodFind); err != nil {
 		return "", err
 	}
 	cmd := exec.Command(debuginfodFind, args...)
+	if os.Getenv(debuginfodMaxtimeEnv) == "" || os.Getenv(debuginfodTimeoutEnv) == "" {
+		cmd.Env = append(os.Environ(), debuginfodMaxtimeEnv+"=1", debuginfodTimeoutEnv+"=1")
+	}
+	cmd.Stderr = os.Stderr
 	out, err := cmd.Output() // ignore stderr
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


pkg/proc/debuginfod/debuginfod.go
```diff
@@ -1,17 +1,26 @@
...
+	debuginfodTimeoutEnv = "DEBUGINFOD_TIMEOUT"
+)
 
 func execFind(args ...string) (string, error) {
 	if _, err := exec.LookPath(debuginfodFind); err != nil {
 		return "", err
 	}
 	cmd := exec.Command(debuginfodFind, args...)
+	if os.Getenv(debuginfodMaxtimeEnv) == "" || os.Getenv(debuginfodTimeoutEnv) == "" {
+		cmd.Env = append(os.Environ(), debuginfodMaxtimeEnv+"=1", debuginfodTimeoutEnv+"=1")
```
derekparker, line 21: A 1 second timeout?

aarzilli: Yes, it's the minimum possible timeout.

aarzilli: We  could set instead DEBUGINFOD_PROGRESS. The problem is that this is still going to be invisible with other frontends (vscode-go, goland), I think. We are strictly limited to what our interface can look like by the existence of those other frontends, in this case.
One second is enough to get already downloaded symbols, if users need more they can opt-in by setting the environment variables explicitly (maybe we should add those to the documentation).

derekparker: That's a good thought (re: DEBUGINFOD_PROGRESS). For IDE integrations, is there any prior art for this, or standard within DAP for the debug server to send a message to the client to display? There are certain events mentioned in the standard such as the Progress Start/Update/End events, but that might be tedious to implement, as we currently don't expose any of that information at all, so plumbing it back up might not be worth it.

@hyangah (I know you've moved on, but maybe can ping other folks) any thoughts?

aarzilli: That's not a bad idea. I [tried implementing it](https://github.com/aarzilli/delve/tree/progress-debuginfod) but I couldn't get it to show anything. Either it does nothing (bad) or it does something but it's very subtle (also bad). Or maybe I did something wrong, I don't know.

To try it cd into _fixtures/cgostacktrace, `export PATH=.:$PATH` and `export GODEBUG=execerrdot=0` so that the debuginfod-find script gets run instead of the real debuginfod-find and then run vscode in this environment.

derekparker:
Closing in favor of https://github.com/go-delve/delve/pull/3980.

PR #3908, opened 2025-01-22
proc: allow access to thread registers after a function call

### LINKED ISSUE #3310: function call injection has problems when the function call involves passing variables stored in registers

aarzilli:
After a call injection sequence terminates allow the evaluator to
access thread registers again so that variables stored in registers can
be used as arguments.

Fixes #3310

### DIFF
--- a/_fixtures/issue3310.go
+++ b/_fixtures/issue3310.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"fmt"
+	"reflect"
+)
+
+var i = 2
+var val = reflect.ValueOf(i)
+
+func reflectFunc(value reflect.Value) {
+	fmt.Printf("%s\n", value.Type().Name())
+}
+
+func main() {
+	reflectFunc(val)
+	fmt.Println(&i)
+}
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM

PR #3909, opened 2025-01-22
proc: some refactorings for supporting struct literals

aarzilli:
* deduplicates exprToString, defined in multiple packages, and moves it
to astutil
* moves resolveTypedef into pkg/dwarf/godwarf, this is currently only
used by pkg/proc but we will need to call it from pkg/proc/evalop
* creates a new FakePointerType function in pkg/dwarf/godwarf, this is
also a function that is only used by pkg/proc but pkg/proc/evalop will
also need.

Updates #1465

### DIFF
--- a/pkg/astutil/astutil.go
+++ b/pkg/astutil/astutil.go
@@ -3,7 +3,9 @@
 package astutil
 
 import (
+	"bytes"
 	"go/ast"
+	"go/printer"
 	"go/token"
 	"strconv"
 )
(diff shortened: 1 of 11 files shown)


pkg/astutil/astutil.go
```diff
@@ -43,3 +45,9 @@ func Or(x, y ast.Expr) ast.Expr {
 	}
 	return &ast.BinaryExpr{Op: token.LOR, X: x, Y: y}
 }
+
+func ExprToString(t ast.Expr) string {
```
derekparker, line 49: Can you add a doc comment.

pkg/dwarf/godwarf/type.go
```diff
@@ -1101,3 +1101,16 @@ func zeroArray(t Type) {
 		t = at.Type
 	}
 }
+
+func ResolveTypedef(typ Type) Type {
```
derekparker, line 1105: Can you add a doc comment.

derekparker (commented):
Overall LGTM just needs some doc comments.

pkg/astutil/astutil.go
aarzilli: Done.

pkg/dwarf/godwarf/type.go
aarzilli: Done.

derekparker (approved):
LGTM

PR #3914, opened 2025-02-06
proc: better error message for breakpoints on stripped binaries

### LINKED ISSUE #3913: Unable to set line breakpoints on MacOS
MacOS 15.2
Delve 1.24 (homebrew)
Go 1.23.5 darwin/arm64

I have an existing project using a bunch of substitute-path configs. After recent upgrade, or for some other reason, attempt to set any breakpoints, other than "main.main", results in a cryptic message:

> Command failed: could not find file my/app/file.go

No further debug logging is produced ("--log" is specified on the command line).

`list` command works just fine with both symbol names and files, thus I assume that my substitute-path config is good (and it used to work before). Correct presentation of current file when stepping also works.

I'm invoking dlv directly from the command line with "--init" file supplied, so no VSCode shenanigans and such.

aarzilli:
Produce a better error when trying to set a breakpoint, by line, on
stripped binaries.

Fixes #3913

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -192,7 +192,14 @@ func FindFileLocation(p Process, filename string, lineno int) ([]uint64, error) 
 	}
 
 	if len(pcs) == 0 {
-		return nil, &ErrCouldNotFindLine{fileFound, filename, lineno}
+		stripped := false
+		for _, image := range bi.Images {
+			if image.Stripped() && image.symTable != nil && image.symTable.Files[filename] != nil {
+				stripped = true
+				break
+			}
+		}
+		return nil, &ErrCouldNotFindLine{fileFound, stripped, filename, lineno}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/proc/bininfo.go
```diff
@@ -192,7 +192,13 @@ func FindFileLocation(p Process, filename string, lineno int) ([]uint64, error)
 	}
 
 	if len(pcs) == 0 {
-		return nil, &ErrCouldNotFindLine{fileFound, filename, lineno}
+		stripped := false
+		for _, image := range bi.Images {
+			if image.Stripped() && image.symTable != nil && image.symTable.Files[filename] != nil {
+				stripped = true
```
derekparker, line 198: Not that the amount of images would likely be a performance issue, but we can break after this assignment, right?

aarzilli: Done.

derekparker (approved):
LGTM

PR #3916, opened 2025-02-17
*: remove uses of reflect.MethodByName from all of Delve

aarzilli:
When reflect.MethodByName is used the linker can not fully perform
deadcode elimination. This commit updates cobra and rewrites the
suitableMethods part of service/rpccommon so that reflect.MethodByName
is not used and the linker can fully execute deadcode elimination.

The executable size on go1.24.0 on linux is reduced from 25468606 bytes
to 22453382 bytes or a reduction of approximately 12%.

See also:

https://github.com/spf13/cobra/pull/1956
https://github.com/aarzilli/whydeadcode


derekparker (approved):
LGTM

This is really great, thanks for working on this. Glad the change finally landed in cobra too.

PR #3917, opened 2025-02-17
CirrusCI: update FreeBSD version

aarzilli:


aarzilli:
The version we had there is no longer available.

derekparker (approved):
LGTM

PR #3918, opened 2025-02-17
logflags: replace logrus with log/slog

aarzilli:
We weren't using most features of logrus and doing this allows us to
drop one external dependency.

The package log/slog was added in Go 1.21 which is already the minimum
version we can build with.


aarzilli:
Made it on top of the deadcode elimination change because both touch dependencies and therefore conflict with eachother anyway.

derekparker (approved):
LGTM

PR #3919, opened 2025-02-17
Reforeground debugger process after debugge exits

ElectricPulse:
I am debugging a bubbletea application and a lot of times I would love to mash CTRL-C to exit the application and the server. This very crude commit allows that. I also added a TODO for the windows side, because there the debugger needs to notify (via a channel) the `debug` command to start capturing signals.
One other thing that should probably be implemented is that client should exit on a headless server's exit.

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -18,7 +18,10 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
+
+	"github.com/snabb/tcxpgrp"
 
 	"github.com/go-delve/delve/pkg/dwarf/op"
 	"github.com/go-delve/delve/pkg/gobuild"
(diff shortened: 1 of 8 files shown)


service/debugger/debugger.go
```diff
@@ -1205,6 +1208,16 @@ func (d *Debugger) Command(command *api.DebuggerCommand, resumeNotify chan struc
 
 	if err != nil {
 		var errProcessExited proc.ErrProcessExited
+
+		// Process exited so we should "re-foreground" the main process, so that it can capture interrupts
+		if errors.As(err, &errProcessExited) && !tcxpgrp.IsForeground() {
+			pgid := syscall.Getpgrp()
```
derekparker, line 1214: The body of this branch can be folded into the branch below, I think.

Also, this functionality must be encapsulated somewhere else in a unix specific file, with another implementation available for Windows (can be stubbed initially if you don't plan to fix this for Windows in this PR). Can you place place this code in a function maybe within `pkg/terminal/` somewhere with proper build tags via file name for unix/windows, might even be worth creating another small subpackage of terminal, but perhaps not. Then you can import and call here.

The reason is this layer of the code is *supposed* to be platform agnostic as much as possible.

derekparker (changes requested):
Thanks for the patch! Just needs some more work to be accepted.

aarzilli:
A few things:
* The build is currently failing on macOS and windows.
* service/debugger is probably the wrong place to do this, it should probably be pkg/proc/native.exitGuard
* doing it unconditionally if IsForeground fails is wrong, maybe we never were the foreground process, it should only be done if pkg/proc/native.Launch was passed proc.LaunchForeground as a flag
* I'd rather not drag in an external dependency just for a single line function like tcsetpgrp

derekparker:
Closing due to lack of progress.

PR #3935, opened 2025-03-06
proc: add support for struct literals

aarzilli:
Adds support for evaluating struct literals, and allocating them into
the target's memory.

Updates #1465


derekparker (approved):
LGTM

PR #3937, opened 2025-03-08
service/debugger: fix FindLocation with child processes

### LINKED ISSUE #3933: Misleading hint when setting breakpoints for debugging child process
dlv version: 1.24.0
go version: 1.24.0
OS: Ubuntu 20.04.6 LTS
Arch: x64
What did I do? Use `target follow-exec -on` to debug the child process. Set breakpoints in the source code of the executable of the child process.
What did you expect to see? Successfully set breakpoints.
What did you see instead? `Command failed: location ... not found`

I tried to set breakpoints to debug the child process. But the misleading hint `Command failed: location ... not found` tempted me to choose `n`.

If I didn't try to press `Y` at the end of the following video, the title of the issue should be "unable to set breakpoints when debugging child process".

[asciicast](https://asciinema.org/a/706784)

The reproduction repository is [Lslightly/delve-dbg-child-proc](https://github.com/Lslightly/delve-dbg-child-proc). README.md contains the steps to reproduce the following video.

The expression `when plugin is loaded` doesn't cover the scenario I think. Besides, `Command failed: location ... not found` can be avoided.

aarzilli:
FindLocation should only return a "location not found" error if none of
the attached target processes can find the location.

Fixes #3933

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -1943,7 +1943,7 @@ func (d *Debugger) findLocation(goid int64, frame, deferredCall int, locStr stri
 		if s1 != "" {
 			subst = s1
 		}
-		if err != nil {
+		if err != nil && !errors.As(err, new(*locspec.ErrLocationNotFound)) {
 			return nil, "", err
 		}
 		for i := range locs {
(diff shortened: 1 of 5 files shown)


service/debugger/debugger.go
```diff
@@ -1961,6 +1961,9 @@ func (d *Debugger) findLocation(goid int64, frame, deferredCall int, locStr stri
 		}
 		locations = append(locations, locs...)
 	}
+	if len(locations) == 0 {
```
Lslightly, line 1964: If there are multiple locations found in different targets, it's ok without ambiguity, right?

aarzilli: If you look at the for loop that immediately follows, it decorates each location with the PID of the current target.

derekparker (approved):
LGTM

PR #3938, opened 2025-03-10
terminal/command: bind if condition to location earlier

Lslightly:
There is only one valid [commit](https://github.com/go-delve/delve/pull/3938/commits/16e64547d205dd292505a9f084409cf6f43803c3).

Fix the problem that "if condition" is not bound to breakpoint when the location is not found. See command_test.go for the use case.
The overview of the testcase is to test whether a condition is bound to a breakpoint whose location is not found yet.

> [!important]
> Help needed! [command_test.go](https://github.com/Lslightly/delve/blob/16e64547d205dd292505a9f084409cf6f43803c3/pkg/terminal/command_test.go#L1579) is not yet completed. The remaining part is how to handle with `yesno` decision automatically in [command.go#L1869](https://github.com/Lslightly/delve/blob/16e64547d205dd292505a9f084409cf6f43803c3/pkg/terminal/command.go#L1869). I failed to get the default answer "yes".

Thanks very much!

This PR is also related to #2551, #3937, #3934


Lslightly:
https://github.com/go-delve/delve/blob/fc6a439f2206c796bb20f56f37f8e10eb26cb158/pkg/terminal/command.go#L1837-L1850

Is it possible to first detect whether the condition exists then find the location, which may optimize the performance and make the `spec` to be found more clear?

aarzilli:
> Is it possible to first detect whether the if condition exists and then find the location? It may optimize the performance and make the `spec` to be found more clear.

No, we would have to write our own parser instead of using the one in the standard library.

Also, the test you added is failing.

pkg/terminal/command.go
```diff
@@ -1840,6 +1840,7 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) ([]
 		if match := r.FindStringIndex(argstr); match != nil {
 			requestedBp.Name = ""
 			cond = argstr[match[1]:]
+			requestedBp.Cond = cond
```
aarzilli, line 1843: Assign `argstr[match[1]:]` directly to `requestedBp.Cond` and delete the `cond` variable.

aarzilli:
> The cause is that [`yesno`](https://github.com/go-delve/delve/blob/fc6a439f2206c796bb20f56f37f8e10eb26cb158/pkg/terminal/command.go#L1868) always return false when using `go run _scripts/make.go test -s pkg/terminal -r TestBreakPointFailWithCond`.

You can instrument the `yesno` function by changing it into this:

```
var yesno = func(...) {
    the body of yesno as it is now
}
```

and then writing this helper:

```
func (ft *FakeTerminal) MustExecWithYesNo(cmdstr string, newYesno func(...)) string {
    oldYesno := yesno
    yesno = newYesno
    defer func() { yesno = oldYesno }()
    return ft.MustExec(cmdstr)
}
```

PR #3946, opened 2025-03-12
*: fix various test problems with go1.25

aarzilli:
* disable dwz test if it fails because of dwarf5
* fix TestExamineMemoryCmd, which made unwarranted assumptions about memory


derekparker (approved):
LGTM

PR #3948, opened 2025-03-14
service/dap,service/rpc2,service/debugger,service/rpccommon: enable debugging multiple process by default

### LINKED ISSUE #3947: enable debugging multiple process in DAP
Currently, [vscode-go](https://github.com/golang/vscode-go/tree/eeb3c24fe991e47e130a0ac70a9b214664b4a0ea) has two debug adapter, the legacy one and DAP one.
1. Both do not support setting follow-exec.
    - the legacy one goDebug.ts does call RPCServer.FollowExec https://github.com/go-delve/delve/blob/ab208089dea6b9afd73b2a1ea5a10b08821f65ef/service/rpccommon/suitablemethods.go#L32
2. `Suspended` config is not in interface.
    - The legacy one does not expose `suspended` in interface, and therefore it's always false. https://github.com/go-delve/delve/blob/f6e0c3e59bb0d0852844ea06685c88b968ea4f0b/service/rpc2/server.go#L243-L249 and https://github.com/go-delve/delve/blob/f6e0c3e59bb0d0852844ea06685c88b968ea4f0b/service/rpc2/server.go#L259-L263
    - In the DAP one, `suspended` is hard-coded as false. https://github.com/go-delve/delve/blob/f6e0c3e59bb0d0852844ea06685c88b968ea4f0b/service/dap/server.go#L1523-L1526

So the proposal is:
1. enable follow-exec by default
2. set suspended as true by default
3. return a fake breakpoint to vscode-go/vim-go etc.

related issue: https://github.com/golang/vscode-go/issues/3712, https://github.com/golang/vscode-go/issues/3713

code: https://github.com/Lslightly/delve/tree/fixSuspended

Lslightly:
The motivation of this pull request is to Fix #3947.


Lslightly:
Sorry for creating a pull request without passing test. But it's really an exciting feature and enhancement for me.

aarzilli:
This is the wrong thing to do:

* There are good reasons to have follow-exec disabled and it's more common to not want it enabled (gdb has it off by default and lldb doesn't have it at all).
* The right place to do this is not the service layer anyway
* The legacy adapted could, if it wanted, set suspended to true

The right fix would probably something like:

1. Have "dlv" commands in DAP to control follow-exec status, like we have on the normal command line
2. Emit ProcessEvents when we attach to new processes as described in the protocol specification
3. Return suspended breakpoints as unverified, convert them to verified with a BreakpointEvent when they become unsuspended (it might be fine to do this when the process stops).

derekparker:
@Lslightly thank you for submitting this patch, but I do also agree with @aarzilli's feedback. With that in mind, I am going to close this PR out, but feel free to submit a new patch with the above guidance.

PR #3952, opened 2025-03-18
Documentation: fix substitutePath explanation for DAP server

codesoap:
The DAP server uses "from" and "to" with an opposite meaning to the command line client. This was previously documented incorrectly. Here is what the [official documentation for launch.json](https://github.com/golang/vscode-go/blob/77a4fdb77ded9b3b6a70a70d810eed1cb7e9a631/docs/debugging.md#launchjson-attributes) says:
substitutePath_DAP_doc

Moreover, it was seemingly falsely claimed, that empty strings have special meaning with the DAP server. As far as I can tell, this is not the case. Instead, special placeholders must be used with the DAP server. An example can also be found in [the official documentation](https://github.com/golang/vscode-go/blob/77a4fdb77ded9b3b6a70a70d810eed1cb7e9a631/docs/debugging.md#connect-to-headless-delve-with-target-specified-at-server-start-up):
```
"substitutePath": [
      { "from": "${workspaceFolder}", "to": "/path/to/remote/workspace" },
      ...
]
```

### DIFF
--- a/Documentation/cli/substitutepath.md
+++ b/Documentation/cli/substitutepath.md
@@ -5,6 +5,7 @@ However, under [some circumstances](../faq.md#substpath), the paths that end up 
 
 This configuration is done by specifying a list of path substitution rules.
 
+The path substitution code is SubstitutePath in pkg/locspec/locations.go.
 
 ### Where are path substitution rules specified
 
@@ -22,6 +23,13 @@ substitute-path:
 
 If you are starting a headless instance of Delve and connecting to it through `dlv connect` the configuration file that is used is the one that runs `dlv connect`.
 
+When Delve needs to convert a path P found inside the executable file into a path in the local filesystem it will scan through the list of rules looking for the first one where P starts with the from-path and replace from-path with to-path.
+
+Empty paths in both from-path and to-path are special, they represent relative paths:
+
+- `(from="" to="/home/user/project/src")` converts all relative paths in the executable to absolute paths in `/home/user/project/src`
+- `(from="/build/dir" to="")` converts all paths in the executable that start with `/build/dir` into relative paths.
+
 The rules can also be modified while Delve is running by using the [config substitute-path command](./README.md#config):
 
 ```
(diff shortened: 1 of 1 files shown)


aarzilli:
> Moreover, it was seemingly falsely claimed, that empty strings have special meaning with the DAP server. As far as I can tell, this is not the case.

It's the same code.

codesoap:
> > Moreover, it was seemingly falsely claimed, that empty strings have special meaning with the DAP server. As far as I can tell, this is not the case.
> 
> It's the same code.

Alright. I still couldn't get it to work in my case. I tried to strip a prefix from the paths that are embedded in the executable file and use the remainder as relative paths inside my repository. I can try to reproduce it again tomorrow.

Do you know where delve is executed on the development machine, when it is used through Visual Studio Code? I assumed that it would be run in the root directory of my repository, but this didn't seem to be the case, assuming that the substitution worked.

aarzilli:
> Do you know where delve is executed on the development machine, when it is used through Visual Studio Code?

I don't.

codesoap:
I have tried again today and could only determine that the initial path of the DAP server is neither the root directory of the git repository, nor the root directory of the file system. I think there is no point in documenting relative paths here, as long as we don't know what the path is relative to.

The other way around, where relative paths inside the executable file are mapped using an empty string, might work. I'm not sure how to test it though, because I don't understand how I would compile relative paths into the executable file. If you think this works, I can add some documentation.

aarzilli:
> I think there is no point in documenting relative paths here, as long as we don't know what the path is relative to.

Keep in mind that this is a general documentation for both the command line client as well as the DAP adapter and that the specifics of what the vscode-go extension in particular does are secondary (yes, it's the most popular by far, but there are other clients that use DAP to connect to delve).

codesoap:
I misunderstood the documentation then. It directly references Visual Studio Code's documentation, so I had assumed the documentation was targeted at users of delve that want to integrate it with Visual Studio Code.

I think you should clarify that the DAP documentation is aimed at developers of adapters for different editors. I would consider it helpful, if the documentation pointed out that the most common implementation, `vscode-go`, does not work as intended.

If you think there is no use to this pull request anymore, please close it. If you are interested in only a subset of my suggested changes, please let me know which.

aarzilli:
> I think you should clarify that the DAP documentation is aimed at developers of adapters for different editors.

It's not, it's for end users but end users exist also for other clients.

> I would consider it helpful, if the documentation pointed out that the most common implementation, vscode-go, does not work as intended.

I'm not sure how it doesn't work as intended, btw. I checked and dlv gets run on in the project's directory so it should in fact work.

> If you think there is no use to this pull request anymore, please close it. If you are interested in only a subset of my suggested changes, please let me know which.

I think the PR is mostly fine, I'm mostly objecting/trying to understand the commit message.

Documentation/cli/substitutepath.md
```diff
@@ -35,32 +36,27 @@ Double quotes can be used to specify paths that contain spaces, or to specify em
...
-Regardless of how they are specified the path substitution rules are an ordered list of `(from-path, to-path)` pairs. When Delve needs to convert a path P found inside the executable file into a path in the local filesystem it will scan through the list of rules looking for the first one where P starts with from-path and replace from-path with to-path.
-
-Empty paths in both from-path and to-path are special, they represent relative paths: 
-
-- `(from="" to="/home/user/project/src")` converts all relative paths in the executable to absolute paths in `/home/user/project/src`
-- `(from="/build/dir" to="")` converts all paths in the executable that start with `/build/dir` into relative paths.
-
-The path substitution code is SubstitutePath in pkg/locspec/locations.go.
-
+When Delve needs to convert a path P found in the local filesystem into a path inside the executable file it will scan through the list of rules looking for the first one where P starts with from-path and replace from-path with to-path.
```
aarzilli, line 62: This should be move above, with the rest of the paragraph.

codesoap: I have moved the paragraph up and re-introduced the section about the special meaning of empty paths. I have removed the related section from the commit message.

For consistency, I have also moved the similar paragraph in the cli section up.

codesoap:
> I'm not sure how it doesn't work as intended, btw. I checked and dlv gets run on in the project's directory so it should in fact work.

I forgot to mention one thing: In my tests, I used remote debugging. Could this make a difference? When I did `ps aux | grep dlv` I found nothing, while the debugging session was running in Visual Studio Code.

aarzilli (approved):
LGTM

PR #3955, opened 2025-03-19
service/test: skip substitutepath test when modules disabled

derekparker:
See: https://issues.redhat.com/browse/RHEL-83958


aarzilli (approved):
LGTM

PR #3961, opened 2025-03-27
proc: relax DWARFv5 check

aarzilli:
When using cgo some of the compilation units are produced by the C
compiler and could be in DWARFv5 even if most of the executable is
DWARFv4. In those cases debugging will still work fine outside of C
areas of the executable, it's probably better to proceed.


derekparker (approved):
LGTM

PR #3967, opened 2025-04-02
proc: fix bug telemetry counter increment

aarzilli:
The stack counter for executeOp errors is accidentally incremented too early.


derekparker (approved):
LGTM

Issue #3928, opened 2025-03-04
Debugger hangs when debugging with breakpoints


kushicp:
When a test case with breakpoints in code or test case the itself, is executed, below error is thrown in logs and the test case execution hangs forever.

```
org.jetbrains.concurrency.MessageError: Internal debugger error: <nil> not a String
runtime.gopanic (0x43cdf1)
	/mnt/agent/temp/buildTmp/go/src/runtime/panic.go:770
go/constant.StringVal (0x6173e4)
	/mnt/agent/temp/buildTmp/go/src/go/constant/value.go:500
github.com/go-delve/delve/pkg/proc.(*G).Labels (0x8436f9)
	/mnt/agent/work/4c3e5e3eabfaf2d2/pkg/proc/variables.go:594
github.com/go-delve/delve/service/api.ConvertGoroutine (0x8650d8)
	/mnt/agent/work/4c3e5e3eabfaf2d2/service/api/conversions.go:309
github.com/go-delve/delve/service/debugger.(*Debugger).state (0x8fc0b7)
	/mnt/agent/work/4c3e5e3eabfaf2d2/service/debugger/debugger.go:586
github.com/go-delve/delve/service/debugger.(*Debugger).Command (0x9017cc)
	/mnt/agent/work/4c3e5e3eabfaf2d2/service/debugger/debugger.go:1219
github.com/go-delve/delve/service/rpc2.(*RPCServer).Command (0xa4db1c)
	/mnt/agent/work/4c3e5e3eabfaf2d2/service/rpc2/server.go:130
reflect.Value.call (0x4e3f45)
	/mnt/agent/temp/buildTmp/go/src/reflect/value.go:596
reflect.Value.Call (0x4e3058)
	/mnt/agent/temp/buildTmp/go/src/reflect/value.go:380
github.com/go-delve/delve/service/rpccommon.(*ServerImpl).serveJSONCodec.func3 (0xb3d314)
	/mnt/agent/work/4c3e5e3eabfaf2d2/service/rpccommon/server.go:369
runtime.goexit (0x4736c0)
	/mnt/agent/temp/buildTmp/go/src/runtime/asm_amd64.s:1695
```

The issue can be seen in Intellij IDEA and VSCode, latest versions. Below are the versions used,

GO version - v1.24.0
Delve version - v1.24.0

aarzilli:
Do you have a way that we can reproduce this problem?

kushicp:
Steps to reproduce,

[...]

PR #3968, opened 2025-04-03
proc: limit the max length of goroutine's label map (#3928)

hongyuh-hh:
## Problem

This commit https://github.com/go-delve/delve/pull/3962/commits/a5c88606b8905ce9c61e4dc838c8da0223afe32b will `continue` the loop when it detects invalid goroutine labels to fix the issue https://github.com/go-delve/delve/issues/3928

However, when testing with Delve master https://github.com/go-delve/delve/commit/4dc820ee3c3364152aff34976927f051d800d6ca, corrupted `labelMap.Len` values still cause infinite parsing loops, hanging the debugger:

```
variables.go:588: labelMap len:  32210697599938149
```

## Solution

~~Skip corrupted goroutines entirely instead of partially parsing them. This is should be safe because valid workflows wouldn't produce invalid label structures.~~

Limit the max length of the map

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -43,6 +43,8 @@ const (
 	maxMapBucketsFactor = 100 // Maximum numbers of map buckets to read for every requested map entry when loading variables through (*EvalScope).LocalVariables and (*EvalScope).FunctionArguments.
 
 	maxGoroutineUserCurrentDepth = 30 // Maximum depth used by (*G).UserCurrent to search its location
+
+	maxGoroutinesLabelEntries = 1_000 // Maximum number of label entries for a goroutine's label map
 )
 
 type floatSpecial uint8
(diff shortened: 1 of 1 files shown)


aarzilli:
We should put an upper limit on the loop instead.

hongyuh-hh:
> We should put an upper limit on the loop instead.

Hi @aarzilli, 

I'm ok with limiting the max length of it instead. Now I use `1_000` as the max value in the latest commit. Do let me know if you have any concerns on that, thanks!

pkg/proc/variables.go
```diff
@@ -584,6 +586,11 @@ func (g *G) Labels() map[string]string {
 				case reflect.Struct:
 					labelMap, _ = labelMap.structMember("list")
 					if labelMap != nil {
+						// If the label map length exceeds upper limit, it is likely corrupted or invalid.
+						// In such cases, skip parsing to prevent infinite loops or excessive resource usage.
+						if labelMap.Len > maxGoroutinesLabelEntries {
```
aarzilli, line 591:
```
if labelMap.Len < 0 {
    labelMap.Len = 0
}
if labelMap.Len > maxGoroutinesLabelEntries {
    labelMap.Len = maxGoroutinesLabelEntries
}
```

hongyuh-hh: ok, updated

derekparker (approved):
LGTM

PR #3971, opened 2025-04-04
fix(services/dap) fix race condition when launching with backend=rr

### LINKED ISSUE #3970: [dap/launch] race condition cause nil pointer deref when using `--backend=rr`
(It appears the issue template is broken for me right now and not showing up, I'll do my best to edit it back in)

dlv version
```sh
$ dlv version
Delve Debugger
Version: 1.24.1
Build: $Id: 7c2184d58d69f0717b8620fdebc56d34174f9945 $

$ go version
go version go1.24.1 linux/amd64
```
(dlv, go, and rr have been installed from nixpkgs@26b2bef8b3c73a0931af73d902af2d806588f6bb)

When running dlv as dap server (`dlv dap`) and sending a launch config using `rr` as backend
example config: ```json
{
	"mode": "test",
	"program": "/REDACTED/internal/merge/model",
	"args": [
		"-rapid.checks=5"
	],
	"cwd": "/REDACTED/internal/merge/model",
	"buildFlags": {},
	"output": "//REDACTED/internal/merge/model/__debug_bin1638317219",
	"traceDirPath": "/REDACTED/internal/merge/model/testdata/rr/",
	"dlvCwd": "/REDACTED/internal/merge/model",
	"env": {
		"GOPATH": "/REDACTED/go"
	},
	"backend": "rr",
	"stackTraceDepth": 50
}
```
delve will panic trying to deref debugger.target inside of https://github.com/go-delve/delve/blob/4dc820ee3c3364152aff34976927f051d800d6ca/service/dap/server.go#L1163C23-L1163C42

I have tracked the cause to a simple race condition. for backend RR, [setting the target group is delayed ](https://github.com/go-delve/delve/blob/4dc820ee3c3364152aff34976927f051d800d6ca/service/debugger/debugger.go#L316) to wait for the asynchronous recording to finish. [The  comment above the goroutine](https://github.com/go-delve/delve/blob/4dc820ee3c3364152aff34976927f051d800d6ca/service/debugger/debugger.go#L301) points this out as intentional and safe due to the mutex, however, [`Target()`](https://github.com/go-delve/delve/blob/4dc820ee3c3364152aff34976927f051d800d6ca/service/debugger/debugger.go#L2222) and [`TargetGroup()`](https://github.com/go-delve/delve/blob/4dc820ee3c3364152aff34976927f051d800d6ca/service/debugger/debugger.go#L2226) don't respect the mutex before accessing `d.target`.

I'm unsure whether the public getters are intentionally not blocked by the mutex, however this directly leads to the nil deref, since the recording, especially of more complicated tests (i.e. rapid as in my example above) can take a while, making the recording goroutine lose the race.

I have confirmed that simply adding a Lock and Unlock to the start up of the dap server
```diff
diff --git a/service/dap/server.go b/service/dap/server.go
index 46ce6a0d..365504a1 100644
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1157,6 +1157,8 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
                s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs)
 
                if s.debugger != nil {

+                       s.debugger.LockTarget()
+                       // Unlock immediatey, otherwise ProcessPid() will hang waiting for the mutex
+                       s.debugger.UnlockTarget()
                         s.send(&dap.ProcessEvent{
                                Event: *newEvent("process"),
                                Body: dap.ProcessEventBody{
					SystemProcessId: s.debugger.ProcessPid(),
```
fixes the issue.

I'm not certain if this fix is durable enough, since it implicitly assumes that if d.target is not yet initialized, it will be once the first holder of the TargetMutex releases it and that it will never become invalid again (athough this might be a reasonable internal contract). I will create a basic pull request using the diff above and then test whether adding the mutex to the Target() and TargetGroup() methods also works

BronzeDeer:
fixes #3970

As discussed in #3970 it might be better to put Target() and TargetGroup() und the mutex protection as well for a more comprehensible fix


service/dap/server.go
```diff
@@ -1157,6 +1157,9 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 		s.debugger, err = debugger.New(&s.config.Debugger, s.config.ProcessArgs)
 
 		if s.debugger != nil {
+			s.debugger.LockTarget()
+			// Unlock immediatey, otherwise ProcessPid() will hang waiting for the mutex
```
aarzilli, line 1161: This should be:

```
s.debugger.LockTarget()
cmdLine := s.debugger.Target().CmdLine
s.debugger.UnlockTarget()
```

and then cmdLine is used later instead `s.debugger.Target().CmdLine`.

BronzeDeer: Yeah, you're absolutely right!

I'll update this commit.

I've also gone through the debugger's public methods and there are some more missing targetMutex.Lock(). Details are in [this comment](https://github.com/go-delve/delve/issues/3970#issuecomment-2779013352) on the issue. I'd love a  pointer as to whether it is preferrable for me to just add the missing mutex locks

BronzeDeer:
Updated with changes suggested in https://github.com/go-delve/delve/pull/3971#discussion_r2028942988

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3972, opened 2025-04-05
proc: replace Thread.Location method with a function

aarzilli:
Replaces the Thread.Location method with a ThreadLocation function. All
backends implemented the Location method exactly the same way, this
change deduplicates it.


derekparker (approved):
LGTM

PR #3978, opened 2025-04-09
Add support to trace defer function calls under trace follow option

### LINKED ISSUE #3743: Add defer functions to follow-calls tracing option
Currently there is a PR that adds a new option --follow-calls <depth> to trace the children of the function that matches the regular expression upto the desired depth being reviewed in here- https://github.com/go-delve/delve/pull/3594 
As of now, it supports very limited tracing of defer calls such as deferreturns and if the function in which the defer closure is defined matches the regular expression it traces the deferred functions as well.  
This is a follow up issue to the PR that adds complete support to trace defer calls.

The main challenge in here is that the defer call mechanics does not operate as the regular calls because the interplay between the registers and the stack comes into the picture in this case making it a more involved task to compute the function (address) that is deferred. 
Some of the initial ideas are mentioned in https://github.com/go-delve/delve/pull/3594#discussion_r1619531496
This involves pattern recognition to catch places in the assembly where the deferred function address is being computed and loaded into the stack. And from there we can get the name of the function and add it to the trace list. 
Since the assembly looks different on different architectures we will need to likewise parse the information differently depending on the architecture and glean the function address from that.

archanaravindar:
Is an extension of the basic support added to mention depth upto which functions and their descendants need to be traced https://github.com/go-delve/delve/pull/3594
The same option can now trace defer functions and their descendants 

Fixes #3743


_fixtures/testnextdefer.go
```diff
@@ -3,8 +3,8 @@ package main
 import "fmt"
 
 func main() {
-	defer func() {
-		fmt.Println("hi")
-	}()
-	fmt.Println("bye")
+        defer func() {
```
aarzilli, line 6: Undo indentation changes.

_fixtures/testtracefns.go
```diff
@@ -69,13 +69,31 @@ func F4() {
 	panic("blah")
 }
 
+var intc, intd int
```
aarzilli, line 72: gofmt

_fixtures/testtracefns.go
```diff
@@ -69,13 +69,31 @@ func F4() {
...
+func swap() {
+        defer func() {
+                intc+=100
+        }()
+        temp := intc
+        intc = intd
+        intd = temp
+}
+
+func unnamed_defer() {
```
aarzilli, line 82: It is customary to use camel case with go.

service/debugger/debugger.go
```diff
@@ -407,12 +408,11 @@ func (d *Debugger) LastModified() time.Time {
...
-// for the given function, a list of addresses corresponding
-// to 'ret' or 'call runtime.deferreturn'.
-func (d *Debugger) FunctionReturnLocations(fnName string) ([]uint64, error) {
-	d.targetMutex.Lock()
-	defer d.targetMutex.Unlock()
+
+// functionReturnLocationsInternal is same as FunctionReturnLocations
+// except that it does not have a lock and unlock as its called from
+// within the callback which has already acquired a lock
+func (d *Debugger) functionReturnLocationsInternal(fnName string) ([]uint64, error) {
```
aarzilli, line 416: remove empty line

service/debugger/debugger.go
```diff
@@ -407,12 +408,11 @@ func (d *Debugger) LastModified() time.Time {
...
-// FunctionReturnLocations returns all return locations
-// for the given function, a list of addresses corresponding
-// to 'ret' or 'call runtime.deferreturn'.
-func (d *Debugger) FunctionReturnLocations(fnName string) ([]uint64, error) {
-	d.targetMutex.Lock()
-	defer d.targetMutex.Unlock()
+
+// functionReturnLocationsInternal is same as FunctionReturnLocations
+// except that it does not have a lock and unlock as its called from
+// within the callback which has already acquired a lock
```
aarzilli, line 414: Missing period at the end of the comment

service/debugger/debugger.go
```diff
@@ -677,6 +691,19 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint, locExpr string,
...
+	} else {
+		return bp, nil
+	}
+
+}
+
+// createInternalBreakpoint is the same as CreateBreakpoint except that it is called from only within the debugger process
+// when the lock is already been acquired so the step of acquiring and releasing the lock is skipped
+func (d *Debugger) createBreakpointInternal(requestedBp *api.Breakpoint, locExpr string, substitutePathRules [][2]string, suspended bool) (*api.Breakpoint, error) {
+
```
aarzilli, line 706: Remove empty line.

service/debugger/debugger.go
```diff
@@ -677,6 +691,19 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint, locExpr string,
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
 
+	bp, err := d.createBreakpointInternal(requestedBp, locExpr, substitutePathRules, suspended)
```
aarzilli, line 694: Just `return d.createBreakpointInternal...`

service/debugger/debugger.go
```diff
@@ -452,6 +452,20 @@ func (d *Debugger) FunctionReturnLocations(fnName string) ([]uint64, error) {
 	return addrs, nil
 }
 
+// FunctionReturnLocations returns all return locations
+// for the given function, a list of addresses corresponding
+// to 'ret' or 'call runtime.deferreturn'.
+func (d *Debugger) FunctionReturnLocations(fnName string) ([]uint64, error) {
+	d.targetMutex.Lock()
+	defer d.targetMutex.Unlock()
+	raddrs, err := d.functionReturnLocationsInternal(fnName)
```
aarzilli, line 461: Just `return d.functionReturnLocationsInternal...`

service/debugger/debugger.go
```diff
@@ -1371,12 +1398,39 @@ func (d *Debugger) Functions(filter string, followCalls int) ([]string, error) {
...
+	return res
+
+}
+func (d *Debugger) traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int, rootstr string) ([]string, error) {
+
 	type TraceFunc struct {
-		Func    *proc.Function
-		Depth   int
-		visited bool
+        	Func    *proc.Function
```
aarzilli, line 1429: gofmt

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+					deferbp, err := t.SetBreakpoint(0, instr.Loc.PC, proc.NextBreakpoint, nil)
+					deferbp.RootFuncName = rootstr
+					if err != nil {
+						fmt.Printf("error setting breakpoint inside deferreturn\n")
+					}
+					deferbrklet := deferbp.Breaklets[len(deferbp.Breaklets)-1]
+					deferbrklet.Callback = func(th proc.Thread, _ *proc.Target) (bool, error) {
+						rawlocs, err := proc.ThreadStacktrace(t.Group.Selected, t.CurrentThread(), 20)
+						if err != nil {
+							fmt.Printf("Thread stack trace returned error\n")
```
aarzilli, line 1486: fmt.Printf

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+							name, _, value := t.Group.Selected.BinInfo().Arch.DwarfRegisterToString(i, reg)
+							if isDeferCallReg(t.Group.Selected, name) {
+								addr, err := strconv.ParseUint(value, 0, 64)
+								if err != nil {
+									fmt.Printf("error parsing function address\n")
+								}
+								fn := t.Group.Selected.BinInfo().PCToFunc(addr)
+								_, err = createFnTracepoint(d, fn.Name, rootstr, followCalls)
+								if err != nil {
+									fmt.Printf("1469 error creating tracepoint in function %s\n", fn.Name)
```
aarzilli, line 1532: fmt.Printf

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+								}
+								deferchildren, err := d.traverse(t, fn, sdepth+1, followCalls, rootstr)
+								if err != nil {
+									fmt.Printf("error calling traverse on defer children\n")
```
aarzilli, line 1536: fmt.Printf

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+								}
+								for i := 0; i < len(deferchildren); i++ {
+									_, err := createFnTracepoint(d, deferchildren[i], rootstr, followCalls)
+									if err != nil {
+										fmt.Printf("1480 error creating tracepoint in function %s\n", deferchildren[i])
```
aarzilli, line 1541: fmt.Printf

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+
+						if sdepth+1 > followCalls {
+							return false, nil
+						}
+						regs, err := th.Registers()
+						if err != nil {
+							fmt.Printf("registers inside callback returned err\n")
+
+						}
+						dregs := t.Group.Selected.BinInfo().Arch.RegistersToDwarfRegisters(0, regs)
```
aarzilli, line 1516: You can't use Selected here because it might not be the Target that you want. The same applies to the rest of the function.

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+						dregs.Reg(^uint64(0))
```
aarzilli, line 1517: If you are doing this to make sure loadMore is called, there is no need to do it, the defer register is always loaded within the basic registers.
The approach here however is all wrong, you should just ask for the register you want rather than looping over all registers, formatting each one of them and only then checking if they are the one you want.

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+								continue
+							}
+							name, _, value := t.Group.Selected.BinInfo().Arch.DwarfRegisterToString(i, reg)
+							if isDeferCallReg(t.Group.Selected, name) {
+								addr, err := strconv.ParseUint(value, 0, 64)
+								if err != nil {
+									fmt.Printf("error parsing function address\n")
+								}
+								fn := t.Group.Selected.BinInfo().PCToFunc(addr)
+								_, err = createFnTracepoint(d, fn.Name, rootstr, followCalls)
```
aarzilli, line 1530: You shouldn't create additional breakpoints from within a breaklet callback, it will modify the breaklet array as it is being scanned. I don't know if it actually leads to problem in this particular case, but it's too brittle.

aarzilli (commented):
I don't really get what this PR is trying to achieve. Right now the trace with depth thing is determining statically from the call graph which functions to trace. This PR seems to change that so that deferred functions are added, at runtime, to the traced functions. Why are we doing this and why only deferred functions and not any other kind of dynamic dispatch?

archanaravindar:
Hi @aarzilli Thank you for reviewing the patch in detail and also suggesting improvements to optimize the code
This patch adds support for functions called from within defer {..} that cant be determined statically unlike other functions so that their execution can be traced more accurately. Defers are handled in this patch specifically because they are an explicit feature of Go and one of the most common dynamic dispatch types used within a Go program

service/debugger/debugger.go
```diff
@@ -1410,10 +1464,91 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+						}
+						bpi := &api.BreakpointInfo{}
+						// Since the defer function is known only at runtime, the depth is likewise
+						// calculated by referring to the stack and the mechanism is exactly the same
+						// as that used in pkg/terminal/command.go:printTraceOutput
+						bpi.Stacktrace, err = d.convertStacktrace(rawlocs, nil)
+						if err != nil {
+							fmt.Printf("convert stack trace returned error\n")
+						}
+						stack := bpi.Stacktrace
```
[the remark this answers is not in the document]
archanaravindar: Fixed in new version, thanks for pointing it out!

aarzilli (commented):
All new code should be gofmt'd.

archanaravindar:
> > Defers are handled in this patch specifically because they are an explicit feature of Go and one of the most common dynamic dispatch types used within a Go program
> 
> Defers are an explicit feature of Go just like function variables and interfaces, I'm not even sure that they really are the most common type of dynamic dispatch compared to the others (or by how much if they are), but even if they were I'm not sure it's enough to single them out. Or even how you would explain that to users, i.e. this changes the behavior of `follow-calls` without updating its documentation, but how would you update it "defer calls are followed regardless of what they call but other other types of dynamic calls aren't"?

Thank you for the detailed comment @aarzilli, it appears that with minimal modifications the code can trace functions that are dynamically passed as procedure parameters or returned from functions. Made changes to documentation and added couple of test cases for the same, kindly let us know your thoughts on this, Thanks again for your time on this!

aarzilli:
Sorry to not have reviewed this yet, but I have been busy this past month and I will continue to be busy for a while still.

cmd/dlv/cmds/commands.go
```diff
@@ -364,7 +364,7 @@ only see the output of the trace operations you can redirect stdout.`,
 	must(traceCommand.RegisterFlagCompletionFunc("stack", cobra.NoFileCompletions))
 	traceCommand.Flags().String("output", "", "Output path for the binary.")
 	must(traceCommand.MarkFlagFilename("output"))
-	traceCommand.Flags().IntVarP(&traceFollowCalls, "follow-calls", "", 0, "Trace all children of the function to the required depth")
+	traceCommand.Flags().IntVarP(&traceFollowCalls, "follow-calls", "", 0, "Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.")
```
derekparker, line 372: I would simplify the last bit ("and cases where functions are dynamically returned and passed as parameters.") to "and other dynamically called functions.".

cmd/dlv/dlv_test.go
```diff
@@ -990,6 +990,78 @@ func TestTraceDirRecursion(t *testing.T) {
 	assertNoError(cmd.Wait(), t, "cmd.Wait()")
 }
 
+func TestTraceNestDefer(t *testing.T) {
```
derekparker: All these new tests follow a similar pattern. Would be nice to consolidate the shared code into a function that can be invoked by the actual test.

pkg/proc/breakpoints.go
```diff
@@ -106,7 +106,7 @@ type Breaklet struct {
 	// the return value will determine if the breaklet should be considered
 	// active.
 	// The callback can have side-effects.
-	callback func(th Thread, p *Target) (bool, error)
+	Callback func(th Thread, p *Target) (bool, error)
```
derekparker: I wonder if instead of making this public we instead add a new method such as `SetCallback` on the breakpoint / breaklet. That or expand the API of `Target.SetBreakpoint` to accept a callback, but that will also be a noisy change. Overall I think I'd prefer to keep `Breaklet.callback` private.

service/debugger/debugger.go
```diff
@@ -1392,7 +1408,7 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 		parent := queue[0]
 		queue = queue[1:]
 		if parent == nil {
-			panic("attempting to open file Delve cannot parse")
+			panic("Queue has a nil node, cannot traverse!")
```
derekparker: Please start panic string in lower case.

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 		}
 		f := parent.Func
 		text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+
 		if err != nil {
 			return nil, fmt.Errorf("disassemble failed with error %w", err)
 		}
 		for _, instr := range text {
+			// Defer functions are called in a special way wherein the destination location is nil
```
derekparker: Since this code path is no longer specific to defers please update the comment to be more generic about dynamically dispatched functions and the reason we cannot determine the location statically and must instead find the address at runtime (during program execution).

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+			// Hence its required to put a breakpoint inorder to acquire the address of the function
+			// at runtime and we do this via a call back mechanism
+			if instr.IsCall() && instr.DestLoc == nil {
+				deferbp, err := t.SetBreakpoint(0, instr.Loc.PC, proc.NextBreakpoint, nil)
+				deferbp.RootFuncName = rootstr
```
derekparker: You shouldn't need to set this on the breakpoint as `rootstr` should be available in the callback function (it's closed over).

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+				if err != nil {
+					return nil, fmt.Errorf("error setting breakpoint inside deferreturn")
+				}
+				deferbrklet := deferbp.Breaklets[len(deferbp.Breaklets)-1]
```
derekparker: Is `len(deferbp.Breaklets)` ever greater than 1? If so I wonder if we need to iterate and set the callback on all breaklets. If not, we should add an assertion as a sanity check.

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+				deferbrklet.Callback = func(th proc.Thread, tgt *proc.Target) (bool, error) {
+					rawlocs, err := proc.ThreadStacktrace(tgt, tgt.CurrentThread(), 20)
+					if err != nil {
+						return false, fmt.Errorf("thread stack trace returned error")
+					}
+					// Since the defer function is known only at runtime, the depth is likewise
```
derekparker: s/defer/dynamic/

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+					// calculated by referring to the stack and the mechanism is exactly the same
+					// as that used in pkg/terminal/command.go:printTraceOutput
+					stack, err := d.convertStacktrace(rawlocs, nil)
```
derekparker: This conversion is unnecessary, just use `rawlocs`.

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+				}
+				deferbrklet := deferbp.Breaklets[len(deferbp.Breaklets)-1]
+				deferbrklet.Callback = func(th proc.Thread, tgt *proc.Target) (bool, error) {
+					rawlocs, err := proc.ThreadStacktrace(tgt, tgt.CurrentThread(), 20)
+					if err != nil {
+						return false, fmt.Errorf("thread stack trace returned error")
+					}
+					// Since the defer function is known only at runtime, the depth is likewise
+					// calculated by referring to the stack and the mechanism is exactly the same
+					// as that used in pkg/terminal/command.go:printTraceOutput
```
derekparker: Would be great to share this code somehow, although I appreciate that may be difficult or awkward. Maybe there's something in the algorithm we can generalize and share?

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+						if stack[i].Function.Name() == deferbp.RootFuncName {
+							if rootindex == -1 {
+								rootindex = i
+								break
+							}
+						}
+					}
+					sdepth := rootindex + 1
+
+					if sdepth+1 > followCalls {
```
derekparker, line 1472: I haven't fully thought this through yet, but just a thought:

Do we need to calculate depth like this? For example, instead of hard coding `20` as the number of stack frames we return, limit it to `followCalls` (anything past that we don't care about anyway) and if `rootindex == -1` we return early instead of `if sdepth+1 > followCalls {`. Again, I may be missing something or not thinking through the logic correctly, but this seems like it could optimize this a bit as stack traces are expensive.

service/debugger/debugger.go
```diff
@@ -1410,10 +1426,90 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+					if err != nil {
+						return false, fmt.Errorf("registers inside callback returned err")
+
+					}
+					dregs := tgt.BinInfo().Arch.RegistersToDwarfRegisters(0, regs)
+					archName := tgt.BinInfo().Arch.Name
+					var addr uint64
+					// Get function address from the register used to call defer function from within
+					// runtime.deferreturn
+					if archName == "amd64" {
```
derekparker: This feels like it should be set on `Arch` and exposed via something like `Arch.DynamicCallRegister` so we would then remove this switch and have instead something like `debugger.targetgroup.arch.DynamicCallRegister()`.

service/debugger/debugger.go
archanaravindar: Modified the call to allocate rawlocs to not exceed depth beyond followCalls since we need to check only that many entries in the maximum case

service/debugger/debugger.go
archanaravindar: Can we address this in a new issue of its own ?

service/debugger/debugger.go
archanaravindar: done, thanks for the feedback!

pkg/proc/breakpoints.go
archanaravindar: done, thanks for the suggestion

cmd/dlv/dlv_test.go
archanaravindar: moved over tests to service/test/integration2_test.go

service/debugger/debugger.go
archanaravindar: Done, Thanks for the suggestion!

pkg/proc/breakpoints.go
```diff
@@ -106,17 +108,23 @@ type Breaklet struct {
...
 
 	// For WatchOutOfScopeBreakpoints and StackResizeBreakpoints the watchpoint
 	// field contains the watchpoint related to this out of scope sentinel.
 	watchpoint *Breakpoint
+
 }
 
+// SetCallBack sets the call back field, this was primarily added to prevent exporting callback field
```
aarzilli: Why make the Callback member public and then also add a setter method?

service/debugger/debugger.go
```diff
@@ -1371,12 +1384,14 @@ func (d *Debugger) Functions(filter string, followCalls int) ([]string, error) {
 	return funcs, nil
 }
 
-func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int) ([]string, error) {
+func (d *Debugger) traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int, rootstr string) ([]string, error) {
+
```
aarzilli: go fmt

service/debugger/debugger.go
```diff
@@ -1410,10 +1425,73 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+						_, err := createFnTracepoint(d, dynchildren[i], rootstr, followCalls)
+						if err != nil {
+							return false, fmt.Errorf("error creating tracepoint in function %s", dynchildren[i])
+						}
+					}
+					return false, nil
+				}
+		for _, dynbrklet_i := range dynbp.Breaklets {
+				dynbrklet_i.SetCallBack(dynCallback)
+			}
```
aarzilli: go fmt

service/debugger/debugger.go
```diff
@@ -1432,6 +1510,33 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 	return funcs, nil
 }
 
+// For dynamic functions, since we get to know the functions that are being called from deferreturn quite late in execution
+// we need to create the trace point ourselves so that it can be included in the trace output just in time
+func createFnTracepoint(d *Debugger, fname string, rootstr string, followCalls int) (*api.Breakpoint, error) {
+
```
aarzilli, line 1543: go fmt

service/debugger/debugger.go
```diff
@@ -1432,6 +1510,33 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+	tbp, err1 := d.createBreakpointInternal(&api.Breakpoint{FunctionName: fname, Tracepoint: true, RootFuncName: rootstr, Stacktrace: 20, TraceFollowCalls: followCalls}, "", nil, false)
+	if tbp == nil {
+		if err1 != nil && strings.Contains(err1.Error(), "Breakpoint exists") {
```
aarzilli: See how BreakpointExistsError is used in the rest of the code.

service/debugger/debugger.go
```diff
@@ -1410,10 +1425,73 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+					if sdepth+1 > followCalls {
+						return false, nil
+					}
+					regs, err := th.Registers()
+					if err != nil {
+						return false, fmt.Errorf("registers inside callback returned err")
+
+					}
+					dregs := tgt.BinInfo().Arch.RegistersToDwarfRegisters(0, regs)
+					addr := dregs.Uint64Val(tgt.BinInfo().Arch.DynamicCallReg)
```
aarzilli: AFAICT this is wrong, it assumes that the compiler will only ever emit calls using a single register per architecture. If this was true it would be very brittle, because it would relie on go's register allocator to be bad, but it isn't, if I disassemble some random go program of substance I can find `CALL AX`, `CALL CX`, etc.

archanaravindar: Thanks for the review! have made changes to remove dependency on the actual register, pls let me know if it looks right now

pkg/proc/breakpoints.go
archanaravindar: done!

pkg/proc/breakpoints.go
```diff
@@ -113,6 +115,11 @@ type Breaklet struct {
 	watchpoint *Breakpoint
 }
 
+// SetCallBack sets the call back field, this was primarily added to prevent exporting callback field
+func (B *Breaklet) SetCallBack(callback DynCallFn) {
```
aarzilli: We don't use capital letters for local variable names. Also the name `DynCallFn` is odd for this and personally I'd just write the inlined type here.

service/debugger/debugger.go
```diff
@@ -1448,6 +1539,33 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 	return funcs, nil
 }
 
+// For dynamic functions, since we get to know the functions that are being called from deferreturn quite late in execution
+// we need to create the trace point ourselves so that it can be included in the trace output just in time
+func createFnTracepoint(d *Debugger, fname string, rootstr string, followCalls int) (*api.Breakpoint, error) {
+
+	tbp, err1 := d.createBreakpointInternal(&api.Breakpoint{FunctionName: fname, Tracepoint: true, RootFuncName: rootstr, Stacktrace: 20, TraceFollowCalls: followCalls}, "", nil, false)
```
aarzilli: You are calling a fairly complicated function here to use maybe 5 out of over 100 lines. What you actually need is not createBreakpointInternal, but something like this:

```
func (d *Debugger) createBreakpointForTracepoints(lbp *proc.LogicalBreakpoint) error {
	d.breakpointIDCounter++
	lbp.LogicalID = d.breakpointIDCounter
	lbp.HitCount = make(map[int64]uint64)
	d.target.LogicalBreakpoints[lbp.LogicalID] = lbp
	err := d.target.SetBreakpointEnabled(lbp, true)
	if err != nil {
		delete(d.target.LogicalBreakpoints, lbp.LogicalID)
	}
	return err
}
```

Then this line becomes: `err1 := d.createBreakpointForTracepoints(&proc.LogicalBreakpoint{ Set: proc.SetBreakpoint{ FunctionName: fnname }, Tracepoint: true, RootFuncName: rootstr, Stacktrace: 20, TraceFollowCalls: followCalls })`

Same for the call below. Then you can fold the special handling of proc.BreakpointExistsError into it.

service/debugger/debugger.go
```diff
@@ -1448,6 +1539,33 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 	return funcs, nil
 }
 
+// For dynamic functions, since we get to know the functions that are being called from deferreturn quite late in execution
+// we need to create the trace point ourselves so that it can be included in the trace output just in time
+func createFnTracepoint(d *Debugger, fname string, rootstr string, followCalls int) (*api.Breakpoint, error) {
```
aarzilli: Function comments start with the name of the function.

service/debugger/debugger.go
```diff
@@ -1426,10 +1441,86 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+					}
+					_, err = createFnTracepoint(d, fn.Name, rootstr, followCalls)
+					if err != nil {
+						return false, fmt.Errorf("error creating tracepoint in function %s", fn.Name)
+					}
+					dynchildren, err := d.traverse(t, fn, sdepth+1, followCalls, rootstr)
+					if err != nil {
+						return false, fmt.Errorf("error calling traverse on dynamic children")
+					}
+					for i := 0; i < len(dynchildren); i++ {
```
aarzilli: Use a for range.

service/debugger/debugger.go
```diff
@@ -1426,10 +1441,86 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+						_, err := createFnTracepoint(d, dynchildren[i], rootstr, followCalls)
+						if err != nil {
+							return false, fmt.Errorf("error creating tracepoint in function %s", dynchildren[i])
+						}
+					}
+					return false, nil
+				}
+				for _, dynbrklet_i := range dynbp.Breaklets {
```
aarzilli: In Go snake case isn't used.

service/debugger/debugger.go
aarzilli: Remove the empty line after the function header.

service/debugger/debugger.go
```diff
@@ -1426,10 +1441,86 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 		}
 		f := parent.Func
 		text, err := proc.Disassemble(t.Memory(), nil, t.Breakpoints(), t.BinInfo(), f.Entry, f.End)
+
 		if err != nil {
 			return nil, fmt.Errorf("disassemble failed with error %w", err)
 		}
 		for _, instr := range text {
+			// Dynamic functions are called in a special way wherein the destination location is nil
```
derekparker: I would update this to say that the destination location is not known statically.

service/debugger/debugger.go
```diff
@@ -1426,10 +1441,86 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+			// Hence its required to put a breakpoint inorder to acquire the address of the function
```
derekparker: s/inorder/in order/

service/debugger/debugger.go
```diff
@@ -1426,10 +1441,86 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+			// at runtime and we do this via a call back mechanism
+			if instr.IsCall() && instr.DestLoc == nil {
+				dynbp, err := t.SetBreakpoint(0, instr.Loc.PC, proc.NextBreakpoint, nil)
+				if err != nil {
+					return nil, fmt.Errorf("error setting breakpoint inside deferreturn")
+				}
+				dynCallback := func(th proc.Thread, tgt *proc.Target) (bool, error) {
+					rawlocs, err := proc.ThreadStacktrace(tgt, tgt.CurrentThread(), followCalls+2)
```
derekparker, line 1454: Note: this could probably be optimized by using an iterator so we only materialize up to the point where we have calculated depth (or up to followCalls depth, I suppose). Can you leave a TODO comment for this?

service/test/integration2_test.go
```diff
@@ -826,6 +826,36 @@ func TestTraceFollowCallsCommand(t *testing.T) {
 		assertNoError(err, t, "ListFunctions()")
 		expected = []string{"main.F0", "main.F0.func1", "main.F1", "main.F2", "main.F3", "main.F4", "runtime.deferreturn", "runtime.gopanic", "runtime.gorecover"}
 		matchFunctions(t, functions, expected, depth)
+
```
derekparker: I think this will be a lot cleaner if you turn it into a table driven test.

derekparker (changes requested):

pkg/proc/breakpoints.go
```diff
@@ -113,6 +113,11 @@ type Breaklet struct {
 	watchpoint *Breakpoint
 }
 
+// SetCallBack sets the call back field, this was primarily added to prevent exporting callback field
+func (b *Breaklet) SetCallBack(callback func(th Thread, p *Target) (bool, error)) {
```
aarzilli: Callback is a single [noun](https://www.merriam-webster.com/dictionary/callback) so this should be capitalized `SetCallback`.

service/debugger/debugger.go
```diff
@@ -679,6 +685,13 @@ func (d *Debugger) CreateBreakpoint(requestedBp *api.Breakpoint, locExpr string,
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
 
+	return d.createBreakpointInternal(requestedBp, locExpr, substitutePathRules, suspended)
```
aarzilli: This change is no longer needed.

archanaravindar: Thank you for your patient review!

service/debugger/debugger.go
```diff
@@ -1449,6 +1542,70 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
 	return funcs, nil
 }
 
+// createBreakpointForTracepoints creates a logical breakpoint specifically for tracepoints.
+// It handles the BreakpointExistsError by silently ignoring it, as duplicate tracepoints
+// are expected during dynamic function discovery.
+func (d *Debugger) createBreakpointForTracepoints(lbp *proc.LogicalBreakpoint) error {
```
derekparker: This is only ever called by the function below, so maybe we should just create a function variable in the below function like:

```go
tryEnableBreakpoint := func(lbp *proc.LogicalBreakpoint) error { ... }
```

service/debugger/debugger.go
```diff
@@ -1449,6 +1542,70 @@ func traverse(t proc.ValidTargets, f *proc.Function, depth int, followCalls int)
...
+		}
+		return err
+	}
+	return nil
+}
+
+// createFnTracepoint is a way to create a trace point late in the cycle but just in time that it can be
+// included in the trace output as in the case of dynamic functions, we get to know the functions that are
+// being called from deferreturn quite late in execution
+func createFnTracepoint(d *Debugger, fname string, rootstr string, followCalls int) (*api.Breakpoint, error) {
```
derekparker: Sorry to be pedantic, but technically this creates multiple breakpoints (entry, exit), so maybe we rename this to `createFunctionTracepoints` (note plural).

derekparker: Also, it doesn't look like any of the callers of this function use the returned `*api.Breakpoint` so let's just return an error from this function. If callers did need the created breakpoint, it would be better to return all the breakpoints we created, I think (including return tracepoints).

PR #3980, opened 2025-04-09
proc,service,terminal: add events call use it for dld notifications

aarzilli:
Add a GetEvents method that can be used to retrieve debug events, adds
events for downloading debug info through debuginfod and shows those
events in the command line client as well as console messages in DAP.

In the future this events mechanism can be used to unify EBPF
tracepoints with old style tracepoints.

Update #3906

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -115,6 +115,7 @@ type BinaryInfo struct {
 
 	debugPinnerFn *Function
 	logger        logflags.Logger
+	eventsFn      func(*Event)
 }
 
 var (
@@ -1559,7 +1560,19 @@ func (bi *BinaryInfo) openSeparateDebugInfo(image *Image, exe *elf.File, debugIn
 	// has debuginfod so that we can use that in order to find any relevant debug information.
 	if debugFilePath == "" {
 		var err error
-		debugFilePath, err = debuginfod.GetDebuginfo(image.BuildID)
+		var notify func(string)
+		if bi.eventsFn != nil {
+			notify = func(s string) {
+				bi.eventsFn(&Event{
+					Kind: EventBinaryInfoDownload,
+					BinaryInfoDownloadEventDetails: &BinaryInfoDownloadEventDetails{
+						ImagePath: image.Path,
+						Progress:  s,
+					},
+				})
+			}
+		}
+		debugFilePath, err = debuginfod.GetDebuginfo(notify, image.BuildID)
 		if err != nil {
 			return nil, nil, ErrNoDebugInfoFound
 		}
--- a/pkg/proc/debuginfod/debuginfod.go
+++ b/pkg/proc/debuginfod/debuginfod.go
@@ -1,17 +1,51 @@
 package debuginfod
 
 import (
+	"bufio"
+	"bytes"
+	"os"
 	"os/exec"
 	"strings"
+	"time"
 )
 
 const debuginfodFind = "debuginfod-find"
+const notificationThrottle time.Duration = 1 * time.Second
 
-func execFind(args ...string) (string, error) {
+func execFind(notify func(string), args ...string) (string, error) {
 	if _, err := exec.LookPath(debuginfodFind); err != nil {
 		return "", err
 	}
 	cmd := exec.Command(debuginfodFind, args...)
+	if notify != nil {
+		cmd.Env = append(os.Environ(), "DEBUGINFOD_PROGRESS=yes")
+		stderr, err := cmd.StderrPipe()
+		if err != nil {
+			return "", err
+		}
+		s := bufio.NewScanner(stderr)
+		s.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
+			if atEOF && len(data) == 0 {
+				return 0, nil, nil
+			}
+			if i := bytes.IndexAny(data, "\n\r"); i >= 0 {
+				return i + 1, dropCR(data[0:i]), nil
+			}
+			if atEOF {
+				return len(data), dropCR(data), nil
+			}
+			return 0, nil, nil
+		})
+		go func() {
+			var tlast time.Time
+			for s.Scan() {
+				if time.Since(tlast) > notificationThrottle {
+					tlast = time.Now()
+					notify(string(s.Text()))
+				}
+			}
+		}()
+	}
 	out, err := cmd.Output() // ignore stderr
 	if err != nil {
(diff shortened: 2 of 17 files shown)


pkg/proc/debuginfod/debuginfod.go
```diff
@@ -1,28 +1,69 @@
...
+		}()
+	}
 	out, err := cmd.Output() // ignore stderr
 	if err != nil {
 		return "", err
 	}
 	return strings.TrimSpace(string(out)), err
 }
 
+func dropCR(data []byte) []byte {
```
derekparker, line 56: We could just use https://pkg.go.dev/bytes@go1.24.4#CutSuffix instead of implementing this.

service/api/types.go
```diff
@@ -684,3 +688,22 @@ type GuessSubstitutePathIn struct {
 	ClientGOROOT            string
 	ClientModuleDirectories map[string]string
 }
+
+// Event is an event that happened while the debugged program runs
```
derekparker: Change to: `Event is an event that happened during execution of the debugged program`.

pkg/proc/bininfo.go
```diff
@@ -852,13 +853,13 @@ func (bi *BinaryInfo) LoadBinaryInfo(path string, entryPoint uint64, debugInfoDi
 	return bi.AddImage(path, entryPoint)
 }
 
-func loadBinaryInfo(bi *BinaryInfo, image *Image, path string, entryPoint uint64) error {
+func loadBinaryInfo(bi *BinaryInfo, image *Image, path string, entryPoint uint64, eventsFn func(*Event)) error {
```
derekparker: The event function is already attached to `bi`, so there shouldn't be a need to pass a separate param, I couldn't find anywhere in the change where `bi.eventsFn != eventsFn`.

pkg/proc/target_group.go
```diff
@@ -537,6 +538,13 @@ func (grp *TargetGroup) FollowExecEnabled() bool {
 	return grp.followExecEnabled
 }
 
+// SetEventsFn sets a function that is called to communicate events
+// happening while the target process is running.
+func (grp *TargetGroup) SetEventsFn(eventsFn func(*Event)) {
+	grp.eventsFn = eventsFn
```
derekparker: Why set this twice? Seems like it would be easy to get these out of sync. Also, I didn't see any use case for `TargetGroup.eventsFn`. I understand possibly using this in the future for the ebpf tracing backend, however I don't want to add code that isn't used. I would prefer this only be stored one place.

pkg/proc/target_group.go
```diff
@@ -564,3 +572,22 @@ func (it *ValidTargets) Reset() {
 	it.Target = nil
 	it.start = 0
 }
+
+// Event represents an event that happened during execution.
```
derekparker: Expand a bit more in the comment of the types of use cases for these events (either here or other places such as api/types).

service/debugger/debugger.go
```diff
@@ -1085,8 +1085,20 @@ func (d *Debugger) Command(command *api.DebuggerCommand, resumeNotify chan struc
 	d.setRunning(true)
 	defer d.setRunning(false)
 
+	d.target.SetEventsFn(nil)
 	if command.Name != api.SwitchGoroutine && command.Name != api.SwitchThread && command.Name != api.Halt {
 		d.target.ResumeNotify(resumeNotify)
+
+		if command.WithEvents {
```
derekparker: Do we need to check this boolean or even store it? We can infer this from `eventsFn != nil`. Unless you're worried about catching a programmer error or something, but that should still be caught by a test or something.

service/rpc2/client.go
```diff
@@ -146,45 +149,84 @@ func (c *RPCClient) continueDir(cmd string) <-chan *api.DebuggerState {
...
+			}
+		}
+	}()
+	return done
+}
+
 func (c *RPCClient) Next() (*api.DebuggerState, error) {
 	var out CommandOut
-	err := c.call("Command", api.DebuggerCommand{Name: api.Next, ReturnInfoLoadConfig: c.retValLoadCfg}, &out)
+	done := c.drainEvents()
```
derekparker: I don't mind some duplication, but seems like a wrapper function that does:

```go
done := c.drainEvents()
...
<-done
```

would be useful here.

service/rpc2/client.go
```diff
@@ -146,45 +149,84 @@ func (c *RPCClient) continueDir(cmd string) <-chan *api.DebuggerState {
...
+	err := c.call("Command", api.DebuggerCommand{Name: api.Step, ReturnInfoLoadConfig: c.retValLoadCfg, WithEvents: c.eventsFn != nil}, &out)
+	<-done
 	return &out.State, err
 }
 
 func (c *RPCClient) ReverseStep() (*api.DebuggerState, error) {
 	var out CommandOut
-	err := c.call("Command", api.DebuggerCommand{Name: api.ReverseStep, ReturnInfoLoadConfig: c.retValLoadCfg}, &out)
+	done := c.drainEvents()
+	err := c.call("Command", api.DebuggerCommand{Name: api.ReverseStep, ReturnInfoLoadConfig: c.retValLoadCfg, WithEvents: true}, &out)
```
derekparker: Why is this one explicitly `true` and not inferred via `c.eventsFn != nil`?

pkg/proc/target_group.go
aarzilli: Can you explain what you have in mind here. I thought that what the events are would be obvious by reading the documentation of the embedded fields.

service/debugger/debugger.go
aarzilli: RPCServer.Command always passes a non-nil events function, whether or not it gets used is up to the request made by the client. We could change it so that RPCServer.Command passes nil if WithEvents is not set, I'm fine with it either way.

service/rpc2/client.go
aarzilli: Mistake

service/debugger/debugger.go
derekparker: Yeah I think I prefer the latter approach, less state being passed around and tracked.

pkg/proc/target_group.go
derekparker: Ack, you're right I think that is enough context.

service/debugger/debugger.go
aarzilli: Done.

PR #3985, opened 2025-04-12
proc: remove dieToRuntimeType

aarzilli:
The dwarfToRuntimeType does basically the same thing in another way and
doesn't have a bug with making unsigned comparisons that should be
signed. It also does more things compared to dieToRuntimeType but it
shouldn't matter since function call injection is very slow anyway.

Fixes an intermittent bug on windows:
https://delve.teamcity.com/buildConfiguration/Delve_windows_amd64_1_24/67152


derekparker (approved):
LGTM

PR #3987, opened 2025-04-14
version: use git hash instead of file blob hash for Version.Build

### LINKED ISSUE #3986: Build ID uses version.go blob hash and not built commit hash
When users install Delve using `go install github.com/go-delve/delve/cmd/dlv@latest`, the build ID is always is the blob hash of `pkg/version/version.go` given $Id$ expands to be the blob hash of the file it resides (see https://git-scm.com/docs/gitattributes). Generally build identifiers are the git commit hash that the binary is built/installed with, not the hash of a file's contents. Should the Build id in Version structure not be the git commit hash or should it actually be the blob hash of `pkg/version/version.go`

codeaucafe:
## Description
Fixes #3986

### Problem
Prior to this change, the Delve debugger's version information (via `dlv version` command) displays the blob hash of `pkg/version/version.go` as the a Git ident identifier $Id$. Upon reviewing the code a handful of times I wonder if this is intended, given we search for gitrevision key and try to write its value to Version.Build. From this logic in `buildInfoFixBuild` I would assume we would want the latest commit hash dlv is built with and not the file blob hash of `pkg/version/version.go`. That, and many users (myself included) originally though Build should/would represent the commit hash delve was installed/built with (thus the reason for my going down this rabbit hole surmising why its done this way)

If this should in fact be the built git hash then `buildInfoFixBuild` func in `fixbuild.go` needs to be updated to search for the git hash in `vcs.revision` key's value rather than the current implementation searching for `gitrevision`.

### Solution
Add settings search for `vcs.revision` and if it doesn't find it, we fall back to searching for `gitrevision` to maintain backwards compatibility (I couldn't determine where that was used previously, but maybe a linker flag from awhile ago?)

### Note
If we do want to instead keep the Build Id as the blob hash of `pkg/verison/version.go` then I'm happy to close this PR, but the semantics of Build generally mean the commit hash the binary was installed/built with.

### DIFF
--- a/pkg/version/fixbuild.go
+++ b/pkg/version/fixbuild.go
@@ -2,21 +2,38 @@
 
 package version
 
-import "runtime/debug"
+import (
+	"runtime/debug"
+	"strings"
+)
 
 func init() {
 	fixBuild = buildInfoFixBuild
 }
 
 func buildInfoFixBuild(v *Version) {
+	// Return if v.Build already set, but not if it is Git ident expand file blob hash
+	if !strings.HasPrefix(v.Build, "$Id: ") && !strings.HasSuffix(v.Build, " $") {
+		return
+	}
+
 	info, ok := debug.ReadBuildInfo()
 	if !ok {
 		return
 	}
-	for i := range info.Settings {
-		if info.Settings[i].Key == "gitrevision" {
-			v.Build = info.Settings[i].Value
-			break
+
+	for _, setting := range info.Settings {
+		if setting.Key == "vcs.revision" {
+			v.Build = setting.Value
+			return
+		}
+	}
+
+	// If we didn't find vcs.revision, try the old key for backward compatibility
+	for _, setting := range info.Settings {
+		if setting.Key == "gitrevision" {
+			v.Build = setting.Value
+			return
 		}
 	}
 }


pkg/version/fixbuild.go
```diff
@@ -2,21 +2,48 @@
...
+	info, ok := debug.ReadBuildInfo()
+	if ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" {
+				v.Build = setting.Value
+				return
+			}
 		}
 	}
+
```
aarzilli, line 43: It doesn't make sense to do this, it will only work correctly if delve is run from its own checkout directory.

pkg/version/fixbuild.go
```diff
@@ -2,21 +2,48 @@
...
-	for i := range info.Settings {
-		if info.Settings[i].Key == "gitrevision" {
-			v.Build = info.Settings[i].Value
-			break
+
+	// Try build info
+	info, ok := debug.ReadBuildInfo()
+	if ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" {
```
aarzilli, line 37: At some point the key in settings that we wanted was gitrevision, I don't know when it was changed but I doesn't hurt to still check for it.

pkg/version/version.go
```diff
@@ -38,5 +38,8 @@ func BuildInfo() string {
 }
 
 var fixBuild = func(v *Version) {
-	// does nothing
+	// Default implementation will be replaced in init()
```
aarzilli, line 41: No reason to do this.

pkg/version/fixbuild.go
codeaucafe: Thanks for reviewing quickly. I'll look for it and resolve other comments

pkg/version/fixbuild.go
```diff
@@ -9,14 +9,27 @@ func init() {
 }
 
 func buildInfoFixBuild(v *Version) {
+	if v.Build != "$Id$" {
```
codeaucafe, line 12: catches when not using git hash (i.e. linker field like `go build -ldflags="-X github.com/go-delve/delve/pkg/version.BuildVersion=your-custom-build" -o dlv ./cmd/dlv`); without it'll fall back to git revision hash.

pkg/version/fixbuild.go
```diff
@@ -9,14 +9,27 @@ func init() {
...
-			break
+
+	for _, setting := range info.Settings {
+		if setting.Key == "vcs.revision" {
+			v.Build = setting.Value
+			return
+		}
+	}
+
+	// If we didn't find vcs.revision, try the old key for backward compatibility
```
codeaucafe, line 28: leaving as backwards compatibility fall back

codeaucafe:
@aarzilli I made changes. I realized the issue was the fact that we were trying to set a struct's field with linker flags doesn't work. Linker flags only work with package-level fields https://github.com/golang/go/issues/26042#issuecomment-400056126

pkg/version/fixbuild.go
codeaucafe, line 12: catches when not using git hash (i.e. linker field like `go build -ldflags="-X github.com/go-delve/delve/pkg/version.BuildVersion=your-custom-build" -o dlv ./cmd/dlv`); without it'll fall back to git revision hash.

pkg/version/fixbuild.go
codeaucafe, line 28: leaving as backwards compatibility fall back

aarzilli:
> @aarzilli I made changes. I realized the issue was the fact that we were trying to set a struct's field with linker flags doesn't work. Linker flags only work with package-level fields [golang/go#26042 (comment)](https://github.com/golang/go/issues/26042#issuecomment-400056126)

No we are not setting a struct field with a linker flag. We are setting `main.Build` with a linker flag which then is copied into a struct field in cmd/dlv/main.go.

codeaucafe:
sorry, I completely misread that when I was reading and debugging/testing it, let me remove and properly review, as well as look further into `gitrevision` just in case I missed something reading searching git lots/etc. It may just be adding `vcs.revision` check instead resolves it.

codeaucafe:
@aarzilli should be good to review now. From my further understanding, I believe the fix should simply be the additions I made in `fixbuild.go`'s `buildInfoFixBuild` function.

pkg/version/fixbuild.go
```diff
@@ -2,21 +2,38 @@
...
+	"strings"
+)
 
 func init() {
 	fixBuild = buildInfoFixBuild
 }
 
 func buildInfoFixBuild(v *Version) {
+	// Return if v.Build already set, but not if it is Git ident expand file blob hash
+	if !strings.HasPrefix(v.Build, "$Id: ") && !strings.HasSuffix(v.Build, " $") {
```
codeaucafe, line 16: technically don't need this IF the $Id$ is removed

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #3988, opened 2025-04-14
gdbserial: fix call injection with newer versions of macOS

### LINKED ISSUE #3977: register error when evaluating testify mock expressions
Originally when trying to evaluate a testify mock expression with delve it was causing errors with recent updates. That was fixed with #3966 

However, now I'm getting the following error `Command failed: could not restore registers: protocol error E32 during register write for packet $Pb0=000000000000000...`

Is this expected due to having issues restoring system state after using `call` and how mocks complex runtime mechanisms? Or is this an error that could be fixed? From the little knowledge I have I think maybe this is expected, but I could have sworn that in older versions of delve (before recent macOS Sequoia update caused #3964) I was able to evaluate mock vals like I showed I was trying in #3969 

I'm using the same repo/repro steps as mentioned here https://github.com/go-delve/delve/issues/3969

Any information will help. 

If this is expected, then I believe this can be closed, BUT I'm curious if there is away to get around this to avoid having to use print statements (which isn't not a big issues for the these rare mock cases, but I'd prefer to avoid)

Thank you

aarzilli:
This is the write side of commit f8dda39c.

Fixes #3977

### DIFF
--- a/pkg/proc/gdbserial/gdbserver.go
+++ b/pkg/proc/gdbserial/gdbserver.go
@@ -200,6 +200,7 @@ func (err *ErrBackendUnavailable) Error() string {
 type gdbRegisters struct {
 	regs     map[string]gdbRegister
 	regsInfo []gdbRegisterInfo
+	loaded   []bool
 	tls      uint64
 	gaddr    uint64
 	hasgaddr bool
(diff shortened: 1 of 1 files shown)


codeaucafe:
FYI this fixed my evaluate expression register error issues I was experiencing in #3977. 
Thank you @aarzilli

derekparker (approved):
LGTM

PR #3993, opened 2025-04-17
ci: correct Go package name for FreeBSD

codeaucafe:
The FreeBSD package manager requires using 'lang/go' instead of just 'go' when installing the Go programming language. This change fixes the Cirrus CI build failure on FreeBSD by using the correct package name.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,5 +5,5 @@ freebsd_instance:
   image_family: freebsd-14-2
 
 test_task:
-  install_script: pkg install -y go gcc git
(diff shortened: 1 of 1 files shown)


aarzilli:
As per the comment in the PR this is an upstream bug and for now I wouldn't do anything.

codeaucafe:
Thanks @aarzilli for context the PR description needed to be updated since when was debugging Go just wasn't in this in the packages for FreeBSD used. But I think you are right to close this for now. I made an issue and will discuss further there.

Thanks again for all your fast help!

Cheers
David

PR #3998, opened 2025-04-23
version: remove build constraint

aarzilli:
Remove the go1.18 build constraint, we no longer build on go1.17 or
earlier.


derekparker (approved):
LGTM

Issue #3494, opened 2023-09-05
Random title of exe file causes problems to windows firewall


matwachich:
1. What version of Delve are you using (`dlv version`)? 1.21.0
2. What version of Go are you using? (`go version`)? 1.20.7
3. What operating system and processor architecture are you using? Windows 10 x64
4. What did you do?
5. What did you expect to see?
6. What did you see instead?

In a recent release, delve temporary exe file became random (__debug_binXXXXX.exe where X are random numbers).

This causes problem with windows firewall because each new debug session creates a rule in windows firewall.
![image](https://github.com/go-delve/delve/assets/1791230/c64d9670-f3ef-48a1-afc0-4ec9ae80430a)

A nice behaviour IMO is that delve first checks if there is already a file __debug_bin.exe, if no, then compile normally to __debug_bin.exe, if it exists, add an incremental value to file name.

matwachich:
https://github.com/go-delve/delve/pull/3366

rustyoz:
@matwachich Can you give the latest master version a try?

aarzilli:
Closing, this should be fixed.

PR #4003, opened 2025-05-14
For DAP try using debug filename pattern for debug binary path before using random string. fix for #3494 (new Windows Firewall detections)

rustyoz:
Instead of using the pattern that is always randomized, check if it doesn't exist otherwise use a random filename including the pattern. 

Removes annoying windows firewall issue #3494

### DIFF
--- a/pkg/gobuild/defaultexe.go
+++ b/pkg/gobuild/defaultexe.go
@@ -11,9 +11,21 @@ import (
 // directory named 'name' followed by a random string
 func DefaultDebugBinaryPath(name string) string {
 	pattern := name
+
 	if runtime.GOOS == "windows" {
-		pattern += "*.exe"
+		pattern += ".exe"
+		// Try to create the file atomically with O_EXCL to avoid race conditions
+		f, err := os.OpenFile(pattern, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
+		if err == nil {
+			// Successfully created the file
+			r := f.Name()
+			f.Close()
+			return r
+		}
+		// If file already exists or other error, fall back to CreateTemp
+		// (we don't need to specifically check os.IsExist because CreateTemp will handle any case)
 	}
+
 	f, err := os.CreateTemp(".", pattern)
 	if err != nil {
 		logflags.DebuggerLogger().Errorf("could not create temporary file for build output: %v", err)


derekparker:
@rustyoz are you still planning on working on this?

pkg/gobuild/defaultexe.go
```diff
@@ -11,9 +13,35 @@ import (
...
+		}
+
+		// if there are no files, create a new one using the pattern directly
+		if len(files) == 0 {
+
+			// create a new file with the pattern
+			f, err := os.Create(pattern)
+			if err != nil {
+				logflags.DebuggerLogger().Errorf("could not create temporary file for build output: %v", err)
+				fmt.Println(err)
```
aarzilli: remove fmt.Println.

pkg/gobuild/defaultexe.go
```diff
@@ -11,9 +13,35 @@ import (
...
+				return ""
+			}
+			r := f.Name()
+			f.Close()
+
+			fmt.Println(r)
```
aarzilli: remove fmt.Println

pkg/gobuild/defaultexe.go
```diff
@@ -11,9 +13,35 @@ import (
 // directory named 'name' followed by a random string
 func DefaultDebugBinaryPath(name string) string {
 	pattern := name
+
 	if runtime.GOOS == "windows" {
-		pattern += "*.exe"
+		// check for existing file that matches the pattern
+		globPattern := name + "*.exe"
+		pattern += ".exe"
+		files, err := filepath.Glob(globPattern)
```
aarzilli: Doing it this way is racy. I think it would be better to do `os.OpenFile(pattern+".exe", os.O_CREATE|os.O_EXCL,...)` and if an error is returned that the file exists fallback to `os.CreateTemp`.

pkg/gobuild/defaultexe.go
```diff
@@ -11,9 +11,21 @@ import (
...
+		// Try to create the file atomically with O_EXCL to avoid race conditions
+		f, err := os.OpenFile(pattern, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
+		if err == nil {
+			// Successfully created the file
+			r := f.Name()
+			f.Close()
+			return r
+		}
+		// If file already exists or other error, fall back to CreateTemp
+		// (we don't need to specifically check os.IsExist because CreateTemp will handle any case)
```
aarzilli, line 26: Because you are no longer adding a star to the pattern the following CreateTemp will create a file name of the form `name.exe<randomstring>`, instead of `name<randomstring>.exe`. Is this going to be a problem?

rustyoz: I tested it with windows and it seems to be ok. delve will still execute the binary.

aarzilli (approved):
LGTM

PR #4014, opened 2025-05-29
proc: fix TestRangeOverFunc for arm64 on go1.25

aarzilli:
A couple of lines in rangeoverfunc.go do not produce instructions with
go1.25 on arm64, change the tests to handle this.

Also extend the TestRangeOverFuncNextInlined skip to go1.25 since the
problem with closure names will not be fixed in time.


derekparker (approved):
LGTM

PR #4022, opened 2025-06-02
proc,dwarf: support string constant parse

Zxilly:
This patch provides corresponding parsing support for CL https://go.dev/cl/677915, which is currently under review.

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -767,6 +767,8 @@ func newVariable(name string, addr uint64, dwarfType godwarf.Type, bi *BinaryInf
 		v.Kind = reflect.Invalid
 	case *godwarf.UnspecifiedType:
 		v.Kind = reflect.Invalid
+	case *godwarf.ConstStringType:
+		v.Kind = reflect.String
 	default:
 		v.Unreadable = fmt.Errorf("unknown type: %T", t)
 	}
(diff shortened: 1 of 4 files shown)


pkg/proc/variables.go
```diff
@@ -2307,11 +2309,11 @@ func (cm constantsMap) Get(typ godwarf.Type) *constantType {
 	if !ctyp.initialized {
 		ctyp.initialized = true
 		slices.SortFunc(ctyp.values, func(a, b constantValue) int {
-			return cmp.Compare(a.value, b.value)
+			return cmp.Compare(a.value.ExactString(), b.value.ExactString())
```
aarzilli, line 2312: Numbers shouldn't be sorted as string (nor should we do a conversion to string to sort numbers). I think that storing everything as a constant.Value is probably a mistake, strings and ints should stay in different fields.

pkg/proc/variables.go
```diff
@@ -2321,7 +2323,7 @@ func (cm constantsMap) Get(typ godwarf.Type) *constantType {
 
 func (ctyp *constantType) describe(n int64) string {
```
aarzilli, line 2324: This function should be changed to also handle string constants.

aarzilli (commented):
Please disregards my comments here until the linker change is merged, I don't want you to do work that has to be changed later.

derekparker:
Closing for now, this has been open for some time and the upstream patch has also not been merged.

Issue #3653, opened 2024-02-01
feature request: option to disable unrecovered panic breakpoint


twpayne:
Delve version 1.21.0, Go version 1.21.6, linux/amd64.

I'm using Delve through go-pls with VS Code.

Firstly, thank you for your incredible software! Delve has helped me so much as a Go developer.

Delve [creates a breakpoint for unrecoverable panics](https://github.com/go-delve/delve/blob/efaf5523d8425267f379f4efab8e3144620a0a40/pkg/proc/target.go#L364-L377) that cannot be disabled.

This is often annoying when debugging because it forces the user to interact with the debugger even when they have no interest in the panic itself.

It would be fantastic if there were an option to disable this, and [other obligatory breakpoints](https://github.com/go-delve/delve/blob/efaf5523d8425267f379f4efab8e3144620a0a40/pkg/proc/target.go#L199-L201).

I could not find an existing issue for this, although it [has been requested in JetBrains](https://youtrack.jetbrains.com/issue/GO-3391/Debugger-Configurable-break-on-panic).

I would be happy to contribute a PR to implement this configurability (presumably via a command line flag). Would you accept such a PR?

aarzilli:
The default breakpoints for unrecovered panics and fatal errors aren't special they can be disabled (with toggle/AmendBreakpoint) or removed (clear/RemoveBreakpoint) like any other breakpoint.

It seems a pretty niche feature request, to me, for someone to want to debug a program but also be completely uninterested in actually investigating unrecovered panics (and that these happen often enough that it is inconvenient). 

DAP already has an option (noDebug) to launch the program disabling the debugger.

The fact that JetBrains has a feature request for this but they haven't implemented it is further evidence that this feature is requested only rarely.

Furthermore we are probably not going to add an option to disable the creation of these breakpoints on startups *even on headless mode*.

Because of the above reasons, I don't think we should do this.

TomK:
@aarzilli 
**Summary:** A lack of any output whatsoever that delve has paused on a panic breakpoint, in headless mode, with no active connections. [This](https://github.com/go-delve/delve/issues/317) has caused a headache for our development environment.

**Proposed change:** If delve server is running headless and has no active connections, please do NOT pause execution on the panic.

**Details:**
We run our software suite in tilt to closely represent a production environment, with the addition of compiling the binaries with symbols and attaching a delve headless server onto the process.

Developers (mainly running Jetbrains Goland), can debug their tasks by connecting to delve using the UI.

[...]

PR #4027, opened 2025-06-11
Exclude internal breakpoints from response

ConradIrwin:
While working on debugger support for [Zed](https://zed.dev), I noticed that
delve returns internal breakpoints in the "hit breakpoints" array when handling
a panic.

Although there's nothing really wrong with doing this, it caused an edge-case
in our parsing code (as in practice everyone uses unsigned integers for
breakpoint ids), and there's no particular value in doing so (because the
client has no record of the internal breakpoints).

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -3768,7 +3768,12 @@ func (s *Session) runUntilStopAndNotify(command string, allowNextStateChange *sy
 				if strings.HasPrefix(bp.Name, instructionBpPrefix) {
 					stopped.Body.Reason = "instruction breakpoint"
 				}
-				stopped.Body.HitBreakpointIds = []int{bp.ID}
+				// Filter out internal delve breakpoints (panic, fatal, hardcoded, etc.)
+				if bp.ID > 0 {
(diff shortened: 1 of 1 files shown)


aarzilli:
Those aren't internal breakpoints, they just exist by default. You could delete them if you wanted (which is something people want to do https://github.com/go-delve/delve/issues/3653).

ConradIrwin:
@aarzilli makes sense. Is there a way as a client to know that these exist, or as a delve consumer should I just expect them to be there and expose UI to control them?

I didn't see an immediately obvious "list all breakpoints" command in the DAP, so sorry if this is just a lack of my understanding.

aarzilli:
We'd have to emit [breakpoint events](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Breakpoint) ourselves and clients would have to handle them correctly (if we start doing this it would have to check that at least vscode supports it correctly).

There is no "list all breakpoints" API, you are not missing anything (or we are both missing it).

derekparker (approved):
LGTM

PR #4032, opened 2025-06-16
proc: handle moving of direct interface flag in Go 1.26

randall77:
The "direct interface" flag in internal/abi.Type is moving from the Kind_ field to the TFlag field. See https://go-review.googlesource.com/c/go/+/681936

A few of the other Kind fields dlv doesn't use (and don't exist in Go anymore), so this code can be simplified somewhat.


aarzilli:
Thanks.

derekparker (approved):
LGTM

Issue #4033, opened 2025-06-18
Remember breakpoints


nexovec:
Hi, is there a way to make the breakpoints and watchpoints and tracepoints persist across delve program runs? I am not comfortable with using `runtime.Breakpoint` everywhere, because I need to disable the breakpoints sometimes.

derekparker:
Hey, currently there is no way to checkpoint the state of all breakpoints, etc...

Typically editors handle this pretty well, preserving breakpoints set in the editor across debugger runs, but if you're using Delve directly from the CLI we currently do not have support for this, but I am definitely not opposed to this feature.

nexovec:
You think it would be possible to disable a `runtime.Breakpoint()` breakpoint by replacing it with a noop?

aarzilli:
You could by overwriting the CALL with a series of NOPs.

PR #4045, opened 2025-06-27
Add savestate command

alexsaezm:
This is a very simplistic approach to fixing #4033. Basically, it creates a file with the breakpoint commands. Kind of like the counterpart of the source command.

It's incomplete because it doesn't take in consideration all the possible stuff a breakpoint might have, but I open the PR to see if this approach is something desirable, or it needs a more robust approach.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -152,9 +152,9 @@ A tracepoint is a breakpoint that does not stop the execution of the program, in
 
 See also: "help on", "help cond" and "help clear"`},
 		{aliases: []string{"watch"}, group: breakCmds, cmdFn: watchpoint, helpMsg: `Set watchpoint.
-	
+
 	watch [-r|-w|-rw] <expr>
-	
+
 	-r	stops when the memory location is read
 	-w	stops when the memory location is written
 	-rw	stops when the memory location is read or written
@@ -176,7 +176,7 @@ For recorded targets the command takes the following forms:
 	restart					resets to the start of the recording
 	restart [checkpoint]			resets the recording to the given checkpoint
 	restart -r [newargv...]	[redirects...]	re-records the target process
-	
+
 For live targets the command takes the following forms:
 
 	restart [newargv...] [redirects...]	restarts the process
@@ -213,9 +213,9 @@ Optional [count] argument allows you to skip multiple lines.
 `},
 		{aliases: []string{"stepout", "so"}, group: runCmds, allowedPrefixes: revPrefix, cmdFn: c.stepout, helpMsg: "Step out of the current function."},
 		{aliases: []string{"call"}, group: runCmds, cmdFn: c.call, helpMsg: `Resumes process, injecting a function call (EXPERIMENTAL!!!)
-	
+
 	call [-unsafe] <function call expression>
-	
+
 Current limitations:
 - only pointers to stack-allocated objects can be passed as argument.
 - only some automatic type conversions are supported.
@@ -274,34 +274,34 @@ To only display goroutines where the specified location contains (or does not co
 	curloc: filter by the location of the topmost stackframe (including frames inside private runtime functions)
 	goloc: filter by the location of the go instruction that created the goroutine
 	startloc: filter by the location of the start function
-	
+
 To only display goroutines that have (or do not have) the specified label key and value, use:
 
 	goroutines -with label key=value
 	goroutines -without label key=value
-	
+
 To only display goroutines that have (or do not have) the specified label key, use:
 
 	goroutines -with label key
 	goroutines -without label key
-	
+
 To only display goroutines that are running (or are not running) on a OS thread, use:
 
 
 	goroutines -with running
 	goroutines -without running
-	
+
 To only display user (or runtime) goroutines, use:
 
 	goroutines -with user
 	goroutines -without user
 
 CHANNELS
-	
+
 To only show goroutines waiting to send to or receive from a specific channel use:
 
 	goroutines -chan expr
-	
+
 Note that 'expr' must not contain spaces.
 
 GROUPING
@@ -339,7 +339,7 @@ Called without arguments it will show information about the current goroutine.
 Called with a single argument it will switch to the specified goroutine.
 Called with more arguments it will execute a command on the specified goroutine.`},
 		{aliases: []string{"breakpoints", "bp"}, group: breakCmds, cmdFn: breakpoints, helpMsg: `Print out info for active breakpoints.
-	
+
 	breakpoints [-a]
 
 Specifying -a prints all physical breakpoint, including internal breakpoints.`},
@@ -401,9 +401,9 @@ If regex is specified only package variables with a name matching it will be ret
 
 Argument -a shows more registers. Individual registers can also be displayed by 'print' and 'display'. See Documentation/cli/expr.md.`},
 		{aliases: []string{"exit", "quit", "q"}, cmdFn: exitCommand, helpMsg: `Exit the debugger.
-		
+
 	exit [-c]
-	
+
 When connected to a headless instance started with the --accept-multiclient, pass -c to resume the execution of the target process before disconnecting.`},
 		{aliases: []string{"list", "ls", "l"}, cmdFn: listCommand, helpMsg: `Show source code.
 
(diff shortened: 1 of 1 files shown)


pkg/terminal/command.go
```diff
@@ -473,10 +473,15 @@ Executes the specified command (print, args, locals) in the context of the n-th
 		{aliases: []string{"source"}, cmdFn: c.sourceCommand, helpMsg: `Executes a file containing a list of delve commands
 
 	source <path>
-	
+
 If path ends with the .star extension it will be interpreted as a starlark script. See Documentation/cli/starlark.md for the syntax.
 
 If path is a single '-' character an interactive starlark interpreter will start instead. Type 'exit' to exit.`},
+		{aliases: []string{"savestate"}, cmdFn: c.saveStateCommand, helpMsg: `Saves all current breakpoints to a file
```
derekparker: Put this in `group: breakCmds`.

pkg/terminal/command.go
```diff
@@ -2633,6 +2638,48 @@ func (c *Commands) sourceCommand(t *Term, ctx callContext, args string) error {
...
 
+func (c *Commands) saveStateCommand(t *Term, ctx callContext, args string) error {
+	if len(args) == 0 {
+		return errors.New("wrong number of arguments: source <filename>")
+	}
+	breakPoints, err := t.client.ListBreakpoints(args == "-a")
+	if err != nil {
+		return err
+	}
+	var fileContent string
```
derekparker: Will be more efficient to use a `strings.Builder` here: https://pkg.go.dev/strings#Builder (I know this is a WIP just doing an initial review pass anyways).

pkg/terminal/command.go
```diff
@@ -2633,6 +2638,48 @@ func (c *Commands) sourceCommand(t *Term, ctx callContext, args string) error {
...
+	for _, bp := range breakPoints {
+		// We don't need to store these breakpoints
+		if bp.ID < 0 {
+			continue
+		}
+		fileContent += fmt.Sprintf("break %s:%d\n", bp.File, bp.Line)
```
aarzilli: This needs to be `trace` for tracepoints.

pkg/terminal/command.go
aarzilli: Alternatively `w := bufio.NewWriter(file)` and then write everything to `w`.

aarzilli (commented):
I think that maybe instead of being a separate command this should be a flag of the breakpoints command. The name `savestate` may imply to some that it does much more than just save breakpoints.

pkg/terminal/command.go
alexsaezm: Done.

pkg/terminal/command.go
alexsaezm: I followed @aarzilli suggestion and used `bufio.NewWriter`.

pkg/terminal/command.go
alexsaezm: I added a conditional to track if they are tracepoints.

alexsaezm:
> I think that maybe instead of being a separate command this should be a flag of the breakpoints command. The name `savestate` may imply to some that it does much more than just save breakpoints.

Didn't see this suggestion until now :(. I'm not sure if there are more things that can be saved, but if there is nothing else to save that can be restored, I totally see this.

pkg/terminal/command.go
```diff
@@ -1726,10 +1727,77 @@ func toggle(t *Term, ctx callContext, args string) error {
...
+		for i, arg := range argv {
+			switch arg {
+			case "-a":
+				showAll = true
+			case "-save":
+				if i+1 >= len(argv) {
+					return errors.New("missing filename after -save flag")
+				}
+				saveFile = argv[i+1]
+				break // Exit loop since we found -save and its argument
```
aarzilli: This doesn't exit the loop because it's inside a switch, it doesn't do anything. You need to use a label.

aarzilli (commented):
The documentation needs to be regenerated with `go run _scripts/gen-cli-docs.go`.

alexsaezm:
> The documentation needs to be regenerated with `go run _scripts/gen-cli-docs.go`.

Thanks, I always forget to update it :(

aarzilli (approved):
LGTM

pkg/terminal/command.go
```diff
@@ -1726,10 +1727,78 @@ func toggle(t *Term, ctx callContext, args string) error {
...
+	}
+
+	breakPoints, err := t.client.ListBreakpoints(showAll)
 	if err != nil {
 		return err
 	}
+
+	// If -save flag is provided, save breakpoints to file
+	if saveFile != "" {
+		file, err := os.OpenFile(saveFile, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
```
derekparker: Can just use `os.Create` here for simplicity, only main difference is it uses `0666` as file perms, which should be ok.

pkg/terminal/command.go
```diff
@@ -1726,10 +1727,78 @@ func toggle(t *Term, ctx callContext, args string) error {
...
+		if err != nil {
+			if os.IsExist(err) {
```
derekparker: Is this the behavior we want? I would think allowing the user to overwrite the breakpoint file if they want to is ok, since it's something they are explicitly doing (e.g. we don't implicitly set the output file name or anything, so if the user wanted to preserve the old file they can choose a new name).

Using `os.Create` as mentioned above will lend itself well to this because it truncates the file if it already exists.

pkg/terminal/command.go
```diff
@@ -1726,10 +1727,78 @@ func toggle(t *Term, ctx callContext, args string) error {
...
+		w := bufio.NewWriter(file)
+		defer w.Flush()
+
+		for _, bp := range breakPoints {
+			// We don't need to store these breakpoints
+			if bp.ID < 0 {
+				continue
+			}
+			var err error
+			if bp.Tracepoint {
```
derekparker, line 1743: Watchpoints aren't included in this output, but if we're including tracepoints I think they should be included as well.

Also does this code work for hit conditions? E.g. `api.Breakpoint.{HitCond,HitCondPerG}`.

derekparker (changes requested):

pkg/terminal/command.go
aarzilli: I don't think it's possible to save watchpoints, the address is going to change on restart and the expression used to set them is probably not going to be evaluatable at startup. The exception to this is watchpoints set on global variables, but it's pretty complicated to which ones are those so IMO it isn't worth it.

alexsaezm: I was not properly saving hit conditions. I added them and also a condition to skip Watchpoints.

pkg/terminal/command.go
alexsaezm: While playing with the feature I realized it was a problem nagging constatly about already existing files. I also changed to os.Create as you suggested.

alexsaezm:
I added a bunch of tests to cover the functionality. Also, I realized that it might be better to use aliases instead of storing the IDs (I found this during the tests).

pkg/terminal/command_test.go
```diff
@@ -1626,6 +1626,64 @@ func TestDisplay(t *testing.T) {
...
+				"trace main.main",
+			},
+		},
+		{
+			name:           "empty breakpoint list",
+			breakpointCmds: []string{},
+		},
+	}
+	for _, tc := range tests {
+		withTestTerminal("break", t, func(term *FakeTerminal) {
```
derekparker: The call to `withTestTerminal` can happen within `t.Run`.

pkg/terminal/command_test.go
```diff
@@ -1626,6 +1626,64 @@ func TestDisplay(t *testing.T) {
...
+				f, err := os.CreateTemp("", "test*.txt")
+				if err != nil {
+					t.Fatalf("Failed to create temp file: %v", err)
+				}
+				f.Close()
+				defer os.Remove(f.Name())
+
+				term.MustExec("breakpoints -save " + f.Name())
+				term.MustExec("clearall")
+				term.MustExec("source " + f.Name())
```
derekparker, line 1744: I think we need one more assertion in this test, which is that the debugger state is what we expect it to be after we source the previously saved file. Right now we stop after we source the file without error, but that file could be empty and this test would still pass, because we don't assert that what we sourced resulted in the saved state.

derekparker (changes requested):

pkg/terminal/command_test.go
alexsaezm: Swapped

pkg/terminal/command_test.go
alexsaezm: I tried to do it using the breakpoints output and matching the strings, but it was extremely complex at the end and really error-prone. I hope this new approach is slightly better. It is not as thorough as it should be, but at least it is not a pain in the eyes like regular expressions all over the place.

derekparker (approved):
LGTM

PR #4046, opened 2025-06-29
pkg/goversion: update max supported go minor version

derekparker:
This should have been done in the 1.25 release.

Also update golang.org/x/tools

This fixes the following error when running `make test`:

``` # golang.org/x/tools/internal/tokeninternal
vendor/golang.org/x/tools/internal/tokeninternal/tokeninternal.go:78:9:
invalid array length -delta * delta (constant -256 of type int64) ```

This is the minimum version update which fixes the problem. It bumps our
go.mod up to 1.22, but since we also bumped our max supported version I
think this is ok.

### CODE REVIEW

File: pkg/goversion/compat.go [hunk 1]
```diff
@@ -14,7 +14,7 @@ var (
 	MinSupportedVersionOfGoMajor = 1
 	MinSupportedVersionOfGoMinor = 21
```
aarzilli: While we're here we should also update this one.

derekparker: Done

PR #4049, opened 2025-07-04
pkg/proc: make closure captured vars visible on closure's first line

### LINKED ISSUE #4000: conditional breakpoint evaluation fails: could not find symbol value for test (local variable)
Today, I wanted to debug a specific entry of a table-driven test with many hundreds of table entries.

In my environment, launching delve on a test is a single-click action, but figuring out how to pass the `-test.run=…` flag to delve is more complicated. 

So, I figured I would just use a conditional breakpoint within delve to stop program execution at the test of interest.

Here’s what I did:

```
% go version
go version go1.24.3 linux/amd64

% go install github.com/go-delve/delve/cmd/dlv@latest
go: downloading github.com/go-delve/delve v1.24.2

% git clone https://github.com/protocolbuffers/protobuf-go
# git revision e5d44688be415e6cea6c21ebf5d832ab04378837

% cd protobuf-go/proto

% dlv test -- -test.run=TestEncode
(dlv) b encode_test.go:31
Breakpoint 1 set at 0xa5b06c,0xa5b2d6 for (multiple functions)() ./encode_test.go:31
(dlv) c
> [Breakpoint 1] google.golang.org/protobuf/proto_test.TestEncode() ./encode_test.go:31 (hits goroutine(6):1 total:1) (PC: 0xa5b06c)
    26:	)
    27:	
    28:	func TestEncode(t *testing.T) {
    29:		for _, test := range testValidMessages {
    30:			for _, want := range test.decodeTo {
=>  31:				t.Run(fmt.Sprintf("%s (%T)", test.desc, want), func(t *testing.T) {
    32:					opts := proto.MarshalOptions{
    33:						AllowPartial: test.partial,
    34:					}
    35:					wire, err := opts.Marshal(want)
    36:					if err != nil {
(dlv) p test.desc
"basic scalar types"
(dlv) cond 1 test.desc == "oneof (message)"
(dlv) c
> [Breakpoint 1] google.golang.org/protobuf/proto_test.TestEncode.func1() ./encode_test.go:31 (hits goroutine(7):1 total:2) (PC: 0xa5b2d6)
Command failed: error evaluating expression: could not find symbol value for test

```

I am not sure why this breakpoint condition doesn’t work. As the transcript shows, printing `test.desc` works fine at the location of the breakpoint, but for some reason, the breakpoint evaluation doesn’t see this variable…?!

Thanks in advance for taking a look

aarzilli:
Variables normally become visible on the line after the one they are
defined at, because on their definition line they will not be
initialized. This is a problem for variables captured by closures
because they get the closure function's first line as the declaration
line.

Make variables captured by closures visible on their declaration line
and load their value from the closure struct if they are unreadable.

Fixes #4000


derekparker (approved):
LGTM

stapelberg:
Thanks for the fix!

PR #4052, opened 2025-07-11
proc_test, server_test: workaround for go1.25 line numbering problem

aarzilli:
This commit changes some tests to make them work with a line numbering
regression introduced by go1.25:
See https://github.com/golang/go/issues/74576


aarzilli:
There's a [CL about this](https://go-review.googlesource.com/c/go/+/687815) so let's see how that one shakes out before merging this.

PR #4057, opened 2025-07-16
service/dap: implement restart request

derekparker:


service/dap/server.go
```diff
@@ -3042,10 +3042,76 @@ func (s *Session) onTerminateRequest(request *dap.TerminateRequest) {
...
+		return
+	}
+
+	var newArgs []string
+	var resetArgs bool
+
+	// Update launch/attach arguments if provided
+	if len(request.Arguments) > 0 {
+		// RestartArguments has an "arguments" field that contains launch/attach config
+		var restartArgs struct {
```
aarzilli, line 3073: Can we do this whole thing in a single unmarshal, by unmarshalling into something like this:

```
type restartArgs struct {
    Arguments struct {
        Request string `json:"request"`
        LaunchConfig
    }
}
```

also if the unmarshal fails shouldn't we return an error instead of ignoring the arguments and doing a restart?

service/dap/server.go
```diff
@@ -3042,10 +3042,76 @@ func (s *Session) onTerminateRequest(request *dap.TerminateRequest) {
...
+	// TODO(deparker) handle args from the launch regarding re-recording.
+	discardedBreakpoints, err := s.debugger.Restart(false, "", resetArgs, newArgs, [3]string{}, false)
+	if err != nil {
+		s.sendErrorResponse(request.Request, FailedToLaunch, "Failed to restart", err.Error())
+		return
+	}
+
+	s.config.log.Infof("Restart completed. Discarded breakpoints: %v", discardedBreakpoints)
+
+	// Send success response
```
aarzilli: This comment and the one following are redundant.

aarzilli (approved):
LGTM

PR #4062, opened 2025-07-22
pkg/proc: return better error attempting to call nonexistent function

### LINKED ISSUE #4051: Can we call go internal functions using delve?
Can we call Go internal functions (like from `runtime/coverage` package) using delve? 

Something like this

```
cl.Call(goroutineID, `runtime/coverage.WriteCountersDir("cov")`, false)
```

I was not able to get this to working

derekparker:
Fixes #4051

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -520,6 +520,7 @@ type Function struct {
 	cu         *compileUnit
 
 	trampoline bool // DW_AT_trampoline attribute set to true
+	closure    bool // DW_AT_go_closure attribute set to true on child variable
 
 	// InlinedCalls lists all inlined calls to this function
 	InlinedCalls         []InlinedCall
@@ -2741,6 +2742,11 @@ func (bi *BinaryInfo) loadDebugInfoMapsCompileUnit(ctxt *loadDebugInfoMapsContex
 				inlined = inval >= 1
 			}
 
+			if err := setIsClosure(reader, *entry, ctxt, bi); err != nil {
+				image.setLoadError(bi.logger, "error reading debug_info: %v", err)
+				return
+			}
+
 			if inlined {
 				bi.addAbstractSubprogram(entry, ctxt, reader, image, cu)
 			} else {
@@ -2756,6 +2762,40 @@ func (bi *BinaryInfo) loadDebugInfoMapsCompileUnit(ctxt *loadDebugInfoMapsContex
 			if entry.Children {
 				depth++
 			}
+		}
+	}
+}
+
+// Looks through variable entries to find if the function contains
+// any closure variables.
+// Note: This only works for Go binaries build with 1.23+ when the
+// DW_AT_go_closure_offset attribute is set on the closure variables.
+func setIsClosure(r *reader.Reader, entry dwarf.Entry, ctxt *loadDebugInfoMapsContext, bi *BinaryInfo) error {
+	off := entry.Offset
+	defer func(off dwarf.Offset, reader *reader.Reader) {
+		reader.Seek(off)
+		reader.Next()
+	}(off, r)
+	for {
+		varentry, err := r.Next()
+		if err != nil {
+			return err
+		}
+		if varentry == nil {
+			return nil
+		}
+		switch varentry.Tag {
+		case dwarf.TagVariable:
+			if _, ok := varentry.Val(godwarf.AttrGoClosureOffset).(int64); ok {
+				originIdx := ctxt.lookupAbstractOrigin(bi, off)
+				fn := &bi.Functions[originIdx]
+				fn.closure = true
+				return nil
+			}
+		case dwarf.TagSubprogram:
+			return nil
+		default:
+			continue
 		}
 	}
 }
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -743,6 +743,7 @@ func (fn *Function) extra(bi *BinaryInfo) *functionExtra {
 		for _, v := range vars {
 			off, ok := v.Val(godwarf.AttrGoClosureOffset).(int64)
 			if ok {
+				fn.closure = true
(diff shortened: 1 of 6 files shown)


pkg/proc/eval.go
```diff
@@ -1359,12 +1359,26 @@ func (stack *evalStack) pushLocal(scope *EvalScope, name string, frame int64) (f
 }
 
 func (stack *evalStack) pushIdent(scope *EvalScope, name string) (found bool) {
+	switch name {
```
aarzilli: These are predeclared identifiers, so a user defined variable can override them, they have to be checked last. For example: https://go.dev/play/p/XLCzmQwt9Ta.

aarzilli:
I'm getting the same error with the issue's example (i.e. calling `main.Hello("something")` when `main.Hello` exists) and I think the new test passes whether or not the other changes are merged.
Looking more into it, the weird error in the issue happens because the identifier `main` exists in the program causing delve to interpret `main.Hello` as a selector expression. This used to give the right error before but once we added the selector access to view captured closure variables it tries to treat `main` as a closure and that's what causes the error.

The fix for that issue is to change `(*Variable).structMember`, in the switch case for `reflect.Func` add this check before the `loadFunctionPtr` call:

```
		if v.closureAddr == 0 {
			// Not a closure, normal function
			return nil, fmt.Errorf("%s has no member %s", vname, memberName)
		}
```

derekparker:
That change did solve the issue but broke `TestReadClosure`. I decided to add a new attribute to `Function` which denotes if the function is a closure by scanning child variables and looking for `DW_AT_go_closure_offset`. I know this only exists in 1.23+ so if the test fails in CI for older versions I think it's best to skip them for older toolchains.

aarzilli:
I think you can still make the problem happen by referring to an existing closure: `main.f.func1.somethingthatdoesntexist`, if we can find `main.f.func1` we'll create the variable for it and it will be marked as a closure, but it won't be, because we didn't get it from a variable.

I think the check should be something about the Variable object rather than the function it refers to.

I think it's better to check the DwarfType. We are making those variables for functions at [/home/a/n/delve/pkg/proc/eval.go:827](https://github.com/go-delve/delve/blob/2e540ac0f0a2f0fdb463b422a9a628bad870cc69/pkg/proc/eval.go#L827), we could do this check at that switch case:

```
		if v.DwarfType != nil && v.DwarfType.Common().ByteSize == 0 {
			// Not a closure, normal function
			return nil, fmt.Errorf("%s has no member %s", vname, memberName)
		}
```

If we ever decide to start creating fake types for those variables and start filling the ByteSize field too (which is probably a bad idea), we'll have a failing test from this issue. We could also add something to the Flags field of Variable at the site where those objects are created.

derekparker:
I added the example you mentioned to the test case. We get an error, not technically the correct one, but slightly better than the input/output error. I think the bigger issue (one that probably shouldn't be solved in this PR) is that we need better type checking in the evaluator to determine that `main.f` is a function and shouldn't be used in a selector expression like this.

aarzilli:
> I think the bigger issue (one that probably shouldn't be solved in this PR) is that we need better type checking in the evaluator to determine that main.f is a function and shouldn't be used in a selector expression like this.

That's what adding a flag would be doing.

pkg/proc/bininfo.go
```diff
@@ -2747,6 +2753,40 @@ func (bi *BinaryInfo) loadDebugInfoMapsCompileUnit(ctxt *loadDebugInfoMapsContex
...
+			if _, ok := varentry.Val(godwarf.AttrGoClosureOffset).(int64); ok {
+				originIdx := ctxt.lookupAbstractOrigin(bi, off)
+				fn := &bi.Functions[originIdx]
+				fn.closure = true
+				return nil
+			}
+		case dwarf.TagSubprogram:
+			return nil
+		default:
+			continue
```
aarzilli: If we are going with this solution it needs at least to return early on `case 0:` or in some cases it will read all the way to the end of debug_info.

derekparker:
Went ahead and took a different approach, let me know what you think.

pkg/proc/bininfo.go
```diff
@@ -2747,6 +2747,11 @@ func (bi *BinaryInfo) loadDebugInfoMapsCompileUnit(ctxt *loadDebugInfoMapsContex
 	}
 }
 
+// Looks through variable entries to find if the function contains
```
aarzilli: Leftover comment?

aarzilli:
PS. while the example code is pretty weird consider that the PackagesMap contains every package linked into the program so the probability of name collisions is higher than it looks like.

derekparker:
@aarzilli good point, thanks for pointing that out. Sent another patch based on the older approach. Should've used the `fn.extra()` calculation all along!

pkg/proc/variables_test.go
```diff
@@ -1575,6 +1575,57 @@ func testCallFunctionIntl(t *testing.T, grp *proc.TargetGroup, p *proc.Target, t
 	}
 }
 
+func TestIssue4051(t *testing.T) {
+	protest.AllowRecording(t)
+	withTestProcess("issue4051", t, func(p *proc.Target, grp *proc.TargetGroup, fixture protest.Fixture) {
+		err := grp.Continue()
+		assertNoError(err, t, "initial continue to breakpoint failed")
+
+		err = proc.EvalExpressionWithCalls(grp, p.SelectedGoroutine(), `main.Hello("world")`, pnormalLoadConfig, true)
```
aarzilli, line 1585: This fails on linux/386 because we don't support function calls there and the error is different. I think we can just add a `protest.MustSupportFunctionCalls` at the beginning.

aarzilli (commented):
Besides the test failure, LGTM

pkg/proc/variables_test.go
derekparker: Done.

PR #4070, opened 2025-07-30
proc: early fixes for Go 1.26

aarzilli:
Change 091ff506 fixed loading the direct flag for Go 1.26 when we go
from a runtime type to a dwarf type, we also did the same check from
the dwarf type to a runtime type when converting variables into
'interface {}'.

This commit deletes unifies the two code paths deleting the one that
wasn't updated.

Fixes the tests on tip.

### DIFF
--- a/pkg/proc/stepping_test.go
+++ b/pkg/proc/stepping_test.go
@@ -1227,7 +1227,7 @@ func TestRangeOverFuncNextInlined(t *testing.T) {
 	if !goversion.VersionAfterOrEqual(runtime.Version(), 1, 23) {
 		t.Skip("N/A")
 	}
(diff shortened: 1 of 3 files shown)


derekparker:
Still seeing failures on 1.25 and tip.

aarzilli:
The failures on 1.25 are due to the line numbering bug in go that has been merged but is yet unreleased. For tip I see it passing with this (except for the new test TestConfigurationDoneWithoutDebugSession which is flaky but I think it's just a problem with the test itself).

derekparker:
> The failures on 1.25 are due to the line numbering bug in go that has been merged but is yet unreleased. For tip I see it passing with this (except for the new test TestConfigurationDoneWithoutDebugSession which is flaky but I think it's just a problem with the test itself).

Ack, I think I was looking at an older build with other tip failures.

The `TestConfigurationDoneWithoutDebugSession` flake I will look into.

derekparker (approved):
LGTM

PR #4075, opened 2025-08-02
service/dap: Implement suspended breakpoints

### LINKED ISSUE #4074: service/dap: support suspended breakpoints
I am attempting to add support for suspended breakpoints to `dlv dap`. I am using [a dumb plugin project](https://gitlab.com/ethan.reesor/dumb-go-plugin) to test this. Currently, attempting to set a breakpoint before the plugin is loaded fails. If I hardcode `suspended: true`, then the breakpoint is recorded as suspended, and is activated once the plugin is loaded. However there are two issues:

1. The breakpoint as returned to the DAP client has no filename or line number.
2. **The DAP client needs to be notified that the status of the breakpoint changed.**

(1) can be fixed with a small change to `(*service/dap.Session).setBreakpoints` (though it doesn't feel clean). I believe I know how to fix (2) but it requires sending an event back to the DAP client and AFAIK delve doesn't have any kind of "send an event to the debug client" mechanism.

**How should I do this?** It seems like adding a list of 'notices' to `service/api.DebuggerState` would be the least obtrusive, but `(*pkg/proc.TargetGroup).Continue` doesn't return until the target is paused again, and notices should be sent to the client before then. I thought about modifying the continue command to return early if there's a notice to send to the client, but given that I don't have deep knowledge of delve that seems like a recipe for painful bugs. The two most obvious solutions are a callback function or a channel. A channel seems like the more Go way, though that may have concurrency concerns (with the DAP client, since responses must be serialized).

firelizzard18:
This enables support for 'suspended' breakpoints, for example for a plugin that hasn't been loaded yet.

Fixes #4074

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -585,6 +585,9 @@ func (t *Target) pluginOpenCallback(Thread, *Target) (bool, error) {
 			} else {
 				logger.Debugf("suspended breakpoint %d enabled", lbp.LogicalID)
 			}
+			if lbp.DidUnsuspend != nil {
+				lbp.DidUnsuspend(lbp)
+			}
 		}
 	}
 	return false, nil
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1522,20 +1522,26 @@ func (s *Session) setBreakpoints(prefix string, totalBps int, metadataFunc func(
 				err = errors.New("breakpoint already exists")
 			} else {
 				bp := &api.Breakpoint{
-					Name:    want.name,
-					File:    wantLoc.file,
-					Line:    wantLoc.line,
-					Addr:    wantLoc.addr,
-					Addrs:   wantLoc.addrs,
-					Cond:    want.condition,
-					HitCond: want.hitCondition,
+					Name:         want.name,
+					File:         wantLoc.file,
+					Line:         wantLoc.line,
+					Addr:         wantLoc.addr,
+					Addrs:        wantLoc.addrs,
+					Cond:         want.condition,
+					HitCond:      want.hitCondition,
+					DidUnsuspend: s.didUnsuspendBreakpoint,
 				}
 				err = setLogMessage(bp, want.logMessage)
 				if err == nil {
 					// Create new breakpoints.
-					got, err = s.debugger.CreateBreakpoint(bp, "", nil, false)
+					got, err = s.debugger.CreateBreakpoint(bp, "", nil, true)
 				}
 			}
+		}
+		if len(got.Addrs) == 0 {
+			// Handle suspended breakpoints.
+			got.File = wantLoc.file
+			got.Line = wantLoc.line
 		}
 		createdBps[want.name] = struct{}{}
 		s.updateBreakpointsResponse(breakpoints, i, err, got)
@@ -1556,7 +1562,11 @@ func setLogMessage(bp *api.Breakpoint, msg string) error {
 }
 
 func (s *Session) updateBreakpointsResponse(breakpoints []dap.Breakpoint, i int, err error, got *api.Breakpoint) {
-	breakpoints[i].Verified = err == nil
+	if len(got.Addrs) > 0 {
+		breakpoints[i].Verified = true
+	} else {
+		breakpoints[i].Message = "Unable to set breakpoint"
+	}
 	if err != nil {
 		breakpoints[i].Message = err.Error()
 	} else {
@@ -1565,6 +1575,22 @@ func (s *Session) updateBreakpointsResponse(breakpoints []dap.Breakpoint, i int,
 		breakpoints[i].Line = got.Line
 		breakpoints[i].Source = &dap.Source{Name: filepath.Base(path), Path: path}
 	}
+}
+
+func (s *Session) didUnsuspendBreakpoint(bp *api.Breakpoint) {
+	path := s.toClientPath(bp.File)
+	s.send(&dap.BreakpointEvent{
+		Event: *newEvent("breakpoint"),
+		Body: dap.BreakpointEventBody{
+			Reason: "changed",
+			Breakpoint: dap.Breakpoint{
+				Verified: true,
+				Id:       bp.ID,
+				Line:     bp.Line,
+				Source:   &dap.Source{Name: filepath.Base(path), Path: path},
+			},
+		},
+	})
 }
 
 // functionBpPrefix is the prefix of bp.Name for every breakpoint bp set
--- a/pkg/proc/breakpoints.go
+++ b/pkg/proc/breakpoints.go
@@ -1104,7 +1104,8 @@ type LogicalBreakpoint struct {
 	// condUsesHitCounts is true when 'cond' uses breakpoint hitcounts
 	condUsesHitCounts bool
 
-	UserData interface{} // Any additional information about the breakpoint
+	DidUnsuspend func(*LogicalBreakpoint) // A function that is called if the breakpoint is unsuspended
+	UserData     interface{}              // Any additional information about the breakpoint
 	// Name of root function from where tracing needs to be done
 	RootFuncName string
 	// depth of tracing
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -132,7 +132,8 @@ type Breakpoint struct {
 	// Disabled flag, signifying the state of the breakpoint
 	Disabled bool `json:"disabled"`
 
-	UserData interface{} `json:"-"`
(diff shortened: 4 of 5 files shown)


firelizzard18:
This is a draft because I'm not sure callbacks are the right approach.

service/dap/server.go
```diff
@@ -1522,21 +1522,27 @@ func (s *Session) setBreakpoints(prefix string, totalBps int, metadataFunc func(
...
+					Addrs:        wantLoc.addrs,
+					Cond:         want.condition,
+					HitCond:      want.hitCondition,
+					DidUnsuspend: s.didUnsuspendBreakpoint,
 				}
 				err = setLogMessage(bp, want.logMessage)
 				if err == nil {
 					// Create new breakpoints.
-					got, err = s.debugger.CreateBreakpoint(bp, "", nil, false)
+					got, err = s.debugger.CreateBreakpoint(bp, "", nil, true)
```
aarzilli: I think this should work like the command line client does: only create suspended breakpoints if the plugin package is present or if follow exec is enabled (https://github.com/go-delve/delve/blob/be3019a9eb16255d0e170cdd7bf65e16e49e998c/pkg/terminal/command.go#L1862).

About the notifications I have commented on the issue.

firelizzard18:
I realized that the hack for setting the path of the breakpoint was totally unnecessary. Simply moving the Line and Source assignment behind `if got.File != ""` is enough.

aarzilli:
Some of the tests in `service/dap` are failing.

service/dap/server.go
```diff
@@ -1556,12 +1566,24 @@ func setLogMessage(bp *api.Breakpoint, msg string) error {
 }
 
 func (s *Session) updateBreakpointsResponse(breakpoints []dap.Breakpoint, i int, err error, got *api.Breakpoint) {
+	// Including (breakpoint will be suspended) isn't helpful but some
+	// message to the user is.
+	var errSuspended *debugger.ErrSuspended
+	if errors.As(err, &errSuspended) {
```
aarzilli: Instead of this ErrSuspended used as a sentinel check `got.ExprString != ""` (it's set when there are no physical breakpoints https://github.com/go-delve/delve/blob/master/service/api/conversions.go#L55).

firelizzard18: When I debug, this (below) is what I get for the suspended breakpoint. I removed ErrSuspended and instead used `len(got.Addrs) == 0` on the assumption that a materialized breakpoint will always have an address.

```
api.Breakpoint {
	ID: 3,
	Name: "sourceBp Path=\"/home/firelizzard/src/golang/dumb-plugin/plugin/plugin.go\"  Line=6 Column=0",
	Addr: 0,
	Addrs: []uint64 len: 0,
	cap: 0,
	nil,
	AddrPid: []int len: 0,
	cap: 0,
	nil,
	File: "",
	Line: 0,
	FunctionName: "",
	ExprString: "",
	Cond: "",
	HitCond: "",
	HitCondPerG: false,
	Tracepoint: false,
	TraceReturn: false,
	Goroutine: false,
	Stacktrace: 0,
	Variables: []string len: 0,
	cap: 0,
	nil,
	LoadArgs: *github.com/go-delve/delve/service/api.LoadConfig nil,
	LoadLocals: *github.com/go-delve/delve/service/api.LoadConfig nil,
	WatchExpr: "",
	WatchType: 0,
	VerboseDescr: []string len: 0,
	cap: 0,
	nil,
	HitCount: map[string]uint64 [],
	TotalHitCount: 0,
	Disabled: false,
	UserData: interface {} nil,
	RootFuncName: "",
	TraceFollowCalls: 0}
```

pkg/proc/target.go
```diff
@@ -585,6 +585,14 @@ func (t *Target) pluginOpenCallback(Thread, *Target) (bool, error) {
 			} else {
 				logger.Debugf("suspended breakpoint %d enabled", lbp.LogicalID)
 			}
+			if fn := t.BinInfo().eventsFn; fn != nil {
```
derekparker: Shouldn't this go in the above `else` clause? This will send a materialized breakpoint event even if we got an error trying to set it in the target.

pkg/terminal/terminal.go
```diff
@@ -137,6 +137,8 @@ func New(client service.Client, conf *config.Config) *Term {
 				if !firstEventBinaryInfoDownload {
 					fmt.Fprintf(t.stdout, "\n")
 				}
+			case api.EventBreakpointMaterialized:
+				// Ignore
```
derekparker: I think it would actually be a good idea to print to the terminal that certain breakpoints have been activated now that we have this plumbed through.

service/dap/server.go
```diff
@@ -1508,6 +1508,15 @@ func (s *Session) setBreakpoints(prefix string, totalBps int, metadataFunc func(
 	// Any breakpoint that existed before this request but was not amended must be deleted.
 	s.clearBreakpoints(existingBps, createdBps)
 
+	// Check if the plugin package is present or follow-exec is enabled.
+	// Suspended breakpoints are only relevant when new executable code is
+	// loaded. That can happen when a new process is spawned - which requires
+	// follow-exec - or when new executable code is added to an existing binary.
+	// The only fully supported way of doing the latter is plugins, hence this
+	// check.
+	fns, _ := s.debugger.Functions(`^plugin\.Open$`, 0)
```
derekparker, line 1517: We have similar logic in `pkg/terminal/command.go`, can we centralize this to prevent this code duplication?

firelizzard18: Creating `(*Debugger).PluginsEnabled() bool` is simple enough. However the terminal front end uses `service.Client`/`rpc2.RPCClient` whereas the DAP front end uses the debugger directly. If you judge that adding a new method to the service client/RPC system is worthwhile, I'm happy to do that.

As far as creating a centralized "breakpoints can be suspended/materialized" function that checks both, that's less straightforward because the terminal prompts the user and chooses a different prompt based on the state of (plugins enabled, follow-exec enabled, process is exited). So if that behavior is going to be preserved, the "should I suspend" method would need to return an enum, multiple booleans, or something like that. Given that the behavior in question (different prompts) is specific to the TUI, if I were the maintainer I would likely not add a generalization (the enum) that only applies to one case since other front ends may want a different set of conditions.

derekparker: Ack, good points. Too much work for minimal code duplication.

pkg/proc/target.go
```diff
@@ -585,6 +585,14 @@ func (t *Target) pluginOpenCallback(Thread, *Target) (bool, error) {
...
+				fn(&Event{
```
firelizzard18: @derekparker In `enableBreakpointOnTarget`, if `len(addrs) == 0` does it make sense to say the breakpoint is enabled? When I was testing, I set the breakpoint on line 6 of plugin.go (in my dumb-plugin repo) and that breakpoint was not hit (I assume) because no address was found. In that case, notifying the front end that a breakpoint was materialized seems misleading.

pkg/proc/target.go
firelizzard18: I refactored the nested ifs in a way that makes it easier to follow the control flow IMO. If you prefer, I'll revert that and make the simple change you suggested.

pkg/terminal/terminal.go
firelizzard18: I wasn't sure what to print so I went with "Breakpoint {id} materialized at {file}:{line} ({function})" with the function omitted if FunctionName is blank. As you can see below, the function name is maybe not helpful for plugins. I made the file path relative (if possible) because that seems nicer than printing out an absolute path.

Screenshot_20250902_152937

pkg/proc/target.go
derekparker: Your latest change is great, thanks.

pkg/terminal/terminal.go
```diff
@@ -137,6 +138,27 @@ func New(client service.Client, conf *config.Config) *Term {
 				if !firstEventBinaryInfoDownload {
 					fmt.Fprintf(t.stdout, "\n")
 				}
+			case api.EventBreakpointMaterialized:
+				bp := event.BreakpointMaterializedEventDetails.Breakpoint
+				file := bp.File
+
+				// Make the file path relative if possible.
+				if cwd, err := os.Getwd(); err == nil {
```
derekparker: You should be able to use `t.formatPath` here.

firelizzard18: 👍 Ah, I thought there should be something like that but I looked for fields, not methods.

Issue #2259, opened 2020-12-11
headless attach + connect results in SIGSEGV on detach


polinasok:
1. What version of Delve are you using (`dlv version`)?
Version: 1.5.1
I actually built at the head revision.

2. What version of Go are you using? (`go version`)?
go version go1.14.3 darwin/amd64

3. What operating system and processor architecture are you using?
MacOS Catalina 10.15.7
1.6 GHz Dual-Core Intel Core i5

4. What did you do?
I attached to a running process and then detached without killing it.

5. What did you expect to see?
Clean shutdown.

6. What did you see instead?
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x438623d]

**Details**

I found this issue when using "attach"+"local" mode in vscode-go.
I was able to reproduce with `dlv attach --headless` and `dlv connect` using `_fixtures/loopprog.go`
```
$ pgrep loopprog
23540
polina-macbookair2:delve polina$ ./dlv attach 23540 --log --log-output=debugger,rpc --headless
2020-12-10T22:23:01-08:00 info layer=rpc Using API v1
API server listening at: 127.0.0.1:63029
2020-12-10T22:23:01-08:00 debug layer=rpc API server pid = 23589
2020-12-10T22:23:01-08:00 info layer=debugger attaching to pid 23540
......
2020-12-10T22:23:47-08:00 debug layer=rpc (async 7) <- RPCServer.State(rpc2.StateIn{"NonBlocking":false})
[...]

PR #4078, opened 2025-08-03
service/dap: add target command in console and support follow-exec in DAP interface

### LINKED ISSUE #4077: service/dap: support target command in debug console
Currently, vscode-go does not have user interface for `target follow-exec`. So adding support to `target follow-exec` may help user control the debug process.

The proposed commands are just like terminal:
1. `dlv target list`
2. `dlv target follow-exec [-on [regex]] [-off]`

`dlv switch` is hard since it may cause inconsistency between debugger and frontend.

I have demos in vscode-go. The code is in https://github.com/Lslightly/delve/commits/dap-console-target. Should I create a PR or a drafted one?

Debug Console:

https://github.com/user-attachments/assets/5fa12430-48a6-4d3a-8a82-c0bc836afbd9

dlv follow-exec -off:

https://github.com/user-attachments/assets/76cb5906-8f3d-4ea2-b218-d6bb58aaf0b3

dlv follow-exec -on:

https://github.com/user-attachments/assets/ffd88013-a48b-45b8-9a82-9701b0a2e495

Lslightly:
add target command so that users can control `target follow-exec` and see `target list` from debug console.

Fixes #4077, #3947.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -821,21 +821,29 @@ func TestAttachWithFollowExec(t *testing.T) {
 			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=1", initResp)
 		}
 
-		// 2 >> attach, << error
-		client.AttachRequest(map[string]interface{}{
+		// 2 >> attach, << output, << initialized, << attach
+		client.AttachRequest(map[string]any{
 			"mode":        "local",
 			"processId":   cmd.Process.Pid,
 			"stopOnEntry": stopOnEntry,
 			"followExec":  true,
 			"backend":     "default",
 		})
+		outputEvent := client.ExpectOutputEvent(t)
+		if outputEvent.Seq != 0 ||
+			outputEvent.Body.Output != "Follow exec not supported in attach request yet." ||
+			outputEvent.Body.Category != "important" {
+			t.Errorf("\ngot %#v\nwant Seq=0, Body={Output=\"Follow exec not supported in attach request yet.\", Category=\"important\"}", outputEvent)
+		}
+
 		client.ExpectCapabilitiesEventSupportTerminateDebuggee(t)
-		errResp := client.ExpectErrorResponse(t)
-		if errResp.Seq != 0 ||
-			errResp.RequestSeq != 2 ||
-			errResp.Message != "Failed to attach" ||
-			errResp.Body.Error.Format != "Failed to attach: Follow exec not supported in attach request yet." {
-			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=2 Message=\"Failed to attach\" Format=\"Follow exec not supported in attach request yet.\"", errResp)
+		initEvent := client.ExpectInitializedEvent(t)
+		if initEvent.Seq != 0 {
+			t.Errorf("\ngot %#v\nwant Seq=0", initEvent)
+		}
+		attachResp := client.ExpectAttachResponse(t)
+		if attachResp.Seq != 0 || attachResp.RequestSeq != 2 {
+			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=2", attachResp)
 		}
 
 		// 3 >> disconnect, << disconnect


service/dap/command.go
```diff
@@ -75,14 +77,111 @@ Type "help" followed by the name of a command for more information about it.`
...
 )
 
+type TargetCommand struct {
+	Subcommand  string
+	SwitchPID   int // valid if Subcommand == "switch"
+	FollowMode  FollowExecMode
+	FollowRegex string // only valid if FollowMode == FollowOnRegex
+}
+
+func parseTargetArgs(argstr string) (*TargetCommand, error) {
```
aarzilli: Inline this into `targetCmd` and do the actions directly, like `target` in `pkg/terminal/command.go` works. `FollowExecMode` and `TargetCommand` don't need to exist.

service/dap/command.go
```diff
@@ -5,18 +5,20 @@ import (
...
+	"strconv"
 	"strings"
 
 	"github.com/go-delve/delve/pkg/config"
+	"github.com/go-delve/delve/pkg/proc"
 	"github.com/google/go-dap"
 )
 
 func (s *Session) delveCmd(goid, frame int, cmdstr string) (string, error) {
-	vals := strings.SplitN(strings.TrimSpace(cmdstr), " ", 2)
```
aarzilli, line 16: This was done this way because the arguments of commands could have significant whitespaces. I'm also not sure what is accomplished by splitting everything only to rejoin immediately afterwards.

firelizzard18:
Are you planning on opening a separate PR for https://github.com/Lslightly/delve/commit/d52c2612f58adddf4a13cd4187feca49a2f41297? Since they're both about using follow-exec in DAP-mode and on the smaller side, personally it would make sense to merge them together.

Lslightly:
I can't get what's going wrong from the testing CI.

service/dap/server_test.go
```diff
@@ -4239,12 +4468,65 @@ Type 'dlv help' followed by a command for full documentation.
 						t.Errorf("\ngot: %#v, want sources=\"\"", got)
 					}
 
+					if runtime.GOOS == "freebsd" || runtime.GOOS == "darwin" {
```
aarzilli: Invert the tests and put the follow exec tests inside the if, it's too easy to miss this and add stuff to this test that will be skipped on freebsd and darwin (there's already some of that).

service/dap/server_test.go
```diff
@@ -638,6 +638,232 @@ func TestAttachStopOnEntry(t *testing.T) {
...
+		client.ExpectCapabilitiesEventSupportTerminateDebuggee(t)
+		errResp := client.ExpectErrorResponse(t)
+		if errResp.Seq != 0 ||
+			errResp.RequestSeq != 2 ||
+			errResp.Message != "Failed to attach" ||
+			errResp.Body.Error.Format != "Failed to attach: Follow exec not supported in attach request yet." {
+			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=2 Message=\"Failed to attach\" Format=\"Follow exec not supported in attach request yet.\"", errResp)
+		}
+
+		// TODO(polina): once https://github.com/go-delve/delve/issues/2259 is
```
aarzilli: You are not polina.

firelizzard18: Also #2259 is closed

service/dap/server_test.go
Lslightly: done

service/dap/server_test.go
Lslightly: done

aarzilli (approved):
LGTM

service/dap/server_test.go
```diff
@@ -638,6 +638,229 @@ func TestAttachStopOnEntry(t *testing.T) {
...
+		}
+
+		// 8 >> setBreakpoints, << setBreakpoints
+		client.SetBreakpointsRequest(fixture.Source, []int{54})
+		sbpResp = client.ExpectSetBreakpointsResponse(t)
+		if sbpResp.Seq != 0 || sbpResp.RequestSeq != 8 || len(sbpResp.Body.Breakpoints) != 1 {
+			t.Errorf("\ngot %#v\nwant Seq=0, RequestSeq=8, len(Breakpoints)=1", sbpResp)
+		}
+		if !sbpResp.Body.Breakpoints[0].Verified {
+			t.Errorf("got breakpoints[0].Verified=true, want true")
```
derekparker: This should be `breakpoints[0].Verified=false`, right?

derekparker:
@firelizzard18 your commit for suspended breakpoints is included here. Would you prefer to merge your PR first and have @Lslightly rebase again, or just close your other PR in favor of this one?

firelizzard18:
@derekparker I'm fine with closing mine

service/dap/server_test.go
Lslightly: Done. But [c42f0ff](https://github.com/go-delve/delve/pull/4078/commits/c42f0ff9dfbbad8e29ca039e59b558067596010e) is not shown in https://github.com/go-delve/delve/pull/4078/commits.

service/dap/server_test.go
```diff
@@ -638,6 +638,229 @@ func TestAttachStopOnEntry(t *testing.T) {
...
+		}
+		stopEvent = client.ExpectStoppedEvent(t)
+		if stopEvent.Seq != 0 ||
+			stopEvent.Body.Reason != "breakpoint" ||
+			stopEvent.Body.ThreadId != 1 ||
+			!stopEvent.Body.AllThreadsStopped ||
+			len(stopEvent.Body.HitBreakpointIds) != 1 ||
+			stopEvent.Body.HitBreakpointIds[0] != sbpResp.Body.Breakpoints[0].Id {
+			t.Errorf("\ngot %#v\nwant Seq=0, Body={Reason=\"breakpoint\", ThreadId=1, AllThreadsStopped=true, HitBreakpointIds=[1]}", stopEvent)
+		}
```
Lslightly, line 761: > @firelizzard18 your commit for suspended breakpoints is included here. Would you prefer to merge your PR first and have @Lslightly rebase again, or just close your other PR in favor of this one?

@firelizzard18 @derekparker ~~I think rebasing is better because I find this PR is not completed yet~~ I regret. See the next comment.

When I was learning how `proc.EventBreakpointMaterialized` worked, I found `BreakpointEvent` with `Reason: "changed"` was not sent to dap client when only `followExec` was enabled(no `plugin.Open`).

Only receiving `StoppedEvent` after `Continue` is not enough. Client should also receive `BreakpointEvent` with `Reason: "changed"`. If not, the effect is as follows, not expected by users.

https://github.com/user-attachments/assets/25e0e0bd-c716-409a-8d09-f7584c966ebb

Explanation of the video:

components:
main.go is compiled into the driver binary, which will invoke the binary compiled from build.go through `cmd.CombinedOutput()`.

steps:
1. follow-exec is automatically enabled through launch.json
2. main.go is debugged. It first stops at `drive` function call statement. You can see breakpoints in main.go are all verified, while breakpoints in build.go are all unverified.
3. Then the continue button is clicked, send continue request to delve.
4. Because `suspended` breakpoint is supported now in DAP, the StopEvent in the child process, i.e. build.go, is sent back to vscode through DAP. We can see vscode now stops at `b := 2+a`, the breakpoint in the child process.
5. Then the continue button is clicked again, and the second stopevent in child process is sent back to vscode. vscode stopped at `fmt.Println`

However, in 4th and 5th steps, the left bottom "BREAKPOINTS" bar shows that **both breakpoints, build.go:8, build.go:13 are unverified after switching to child process**. The normal way I think is that these breakpoints would become verified, while breakpoints in main.go will become unverified when switching to child process, i.e. when the debugging cursor jump to build.go.

That's why I say `BreakpointEvent` with `Reason: "changed"` should also be sent back to DAP client when switching the process.

Lslightly: But to support sending `BreakpointEvent` back to client when switching to new target or hitting suspended breakpoints, it seems that there are many things to do. Is it better to create a new issue to resolve the further problem? I don't want to make this PR too big. Just as the PR title says, the PR is only responsible for adding command support and interface support.

firelizzard18: > Is it better to create a new issue to resolve the further problem?

Yes, IMO. My experience as a project manager and maintainer leads me to prefer smaller PRs that solve a single, well-defined issue/requirement/etc. As a contributor, it doesn't matter that much if my PR is merged separately or together with this one, whichever is less hassle/delay, but yes I absolutely think it's better to merge this PR and handle follow-exec breakpoint materialization separately.

pkg/proc/target_exec.go
```diff
@@ -51,6 +51,36 @@ func (grp *TargetGroup) Next() (err error) {
...
+				Kind: EventBreakpointMaterialized,
+				BreakpointMaterializedEventDetails: &BreakpointMaterializedEventDetails{
+					Breakpoint: lbp,
+					Verified:   true,
+				},
+			})
+		}
+	}
+}
+
```
Lslightly: I implemented notifying changes of breakpoints when switching targets to solve the problem mentioned in https://github.com/go-delve/delve/pull/4078#discussion_r2293981698.

The core idea is:

When TargetGroup switches its selected target, i.e. dbp != olddbp, breakpoints in olddbp will become unverified while breakpoints in dbp which is not suspended will become verified.

**But I'm not sure whether there are some cases where one breakpoint is not suspended but should not be verified.**

Lslightly: Now the effect is as follows:

https://github.com/user-attachments/assets/75d38c19-6abd-424c-a468-eca9bbd5f7e2

Explanation of the video:

components:
main.go is compiled into the driver binary, which will invoke the binary compiled from build.go through `cmd.CombinedOutput()`.

steps:
1. follow-exec is automatically enabled through launch.json
2. main.go is debugged. It first stops at `drive` function call statement. You can see breakpoints in main.go are all verified, while breakpoints in build.go are all unverified.
3. Then the `step into` and `step over` button are clicked, and pkg/proc/target_exec.go will execute `Continue` after `cmd.CombinedOutput()` is called. The process of build.go binary is added to `TargetGroup` and is selected in `Continue`.
4. **Because `notifyBreakpointChanges` is implemented, `BreakpointEvent`s are sent back. Breakpoints in `main.go` become unverified, while breakpoints that are not suspended in `build.go` are verified.**
5. Because `suspended` breakpoint is supported now in DAP, the `StopEvent` in the child process, i.e. build.go, is sent back to vscode through DAP. We can see vscode now stops at `b := 2+a`, the breakpoint in the child process.
6. Then the continue button is clicked again. The cursor in the debug session now stops at `fmt.Println` and the "BREAKPOINTS" bar shows breakpoints in main.go verified and breakpoints in build.go unverified.
    1. **Because `notifyBreakpointChanges` is implemented, `BreakpointEvent` will be sent back. Breakpoints in `build.go` become unverified, while breakpoints that are not suspended in `main.go` are verified.**

aarzilli: Why do you think some breakpoints should become unverified after switching targets?

Lslightly: I think users can infer which target process the current cursor is in by the status(verified/unverified) of breakpoints.

Take the above program as an example. Before invoking build.go binary, breakpoints in build.go are not verified. After switching to build.go process, breakpoints in main.go are unverified which indicates the cursor is now in build.go process and which breakpoints in build.go will be hit. After build.go process finishes, breakpoints in main.go are reverified, which indicates the cursor is now in main.go process and which breakpoints will be hit.

Yes, users can also get the selected process through debug console accurately, but to unverify breakpoints of not selected process might give some information to user? I'm not sure. Maybe unverifying is redundant.

There might be a case where a new child process is spawned in `go func()` to interact with the parent process. Then breakpoints in parent process should not be unverified.

So, the better solution is to only verify unsuspended breakpoints in `notifyBreakpointChanges` when switching process, i.e. the number of verified breakpoints will always increase?

firelizzard18: Just FYI, there's a mechanism to inform the IDE (vscode) that there's an additional target that can be attached to. I haven't looked into it that much but from your discussion I'm guessing that is not happening. Which is fine, as I said before my inclination is to make small, incremental PRs, but it's something to keep in mind, as I would like to add that capability in the future.

Lslightly: Ok, I will revert PR back to [968912d](https://github.com/go-delve/delve/pull/4078/commits/968912d906e6319344c76b348ac9ab95c067136d) and discuss about better ways to solve the problem in issue #3947.

service/dap/server.go
```diff
@@ -2048,6 +2066,12 @@ func (s *Session) onAttachRequest(request *dap.AttachRequest) {
 		s.args.substitutePathServerToClient = serverToClient
 	}
 
+	if s.args.followExec {
```
aarzilli: Move this early in the function, before the switch, doing it like this leaves us attached to the target process but with the client believing we didn't attach.

service/dap/server_test.go
```diff
@@ -638,6 +638,229 @@ func TestAttachStopOnEntry(t *testing.T) {
...
+		client.ExpectTerminatedEvent(t)
+
+		// If this call to KeepAlive isn't here there's a chance that stdout will
+		// be garbage collected (since it is no longer alive long before this
+		// point), when that happens, on unix-like OSes, the read end of the pipe
+		// will be closed by the finalizer and the target process will die by
+		// SIGPIPE, which the rest of this test does not expect.
+		runtime.KeepAlive(stdout)
+	})
+
```
aarzilli: Remove newline.

service/dap/server.go
```diff
@@ -1195,6 +1204,15 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 		s.send(&dap.CapabilitiesEvent{Event: *newEvent("capabilities"), Body: dap.CapabilitiesEventBody{Capabilities: dap.Capabilities{SupportsStepBack: true}}})
 	}
 
+	if s.args.followExec {
+		err = s.debugger.FollowExec(s.args.followExec, s.args.followExecRegex)
+		if err != nil {
+			s.sendShowUserErrorResponse(request.Request, FailedToLaunch, "Failed to launch",
+				fmt.Sprintf("Failed to enable follow exec: %v", err))
+			return
```
aarzilli: This return leaves the server in a bad state (the debugger has been started but the client will believe it hasn't). Either we clean up here by calling `s.debugger.Detach()` or we convert the `s.sendShowUserErrorResponse` to `s.send(&dap.OutputEvent{...})` and remove the return here (basically, not being able to enable follow exec mode becomes a warning).

service/dap/server.go
Lslightly: done

service/dap/server_test.go
Lslightly: done

service/dap/server.go
Lslightly: I do what `stopDebugSession` does in its final stage.

https://github.com/go-delve/delve/blob/fd4fc92c74fd00fc44498e5adf43c18dfce80b23/service/dap/server.go#L1382-L1395

aarzilli: In this case detach can never return ErrProcessExited because nothing ever ran.

Lslightly: > because nothing ever ran.

If I understand correctly, `debugger.New` in line 1174 starts the process before calling `s.debugger.FollowExec`

aarzilli: I guess it depends by what you mean by 'starts': `debugger.New` launches the process but it will not run any code in the target process, it will be stopped at the entry point.

firelizzard18: Is it a critical error for follow-exec to fail? Personally I think what I would expect is for delve to print out a warning in the console, possibly to send a DAP notification to the user if that's a thing, then to continue debugging without follow-exec support.

service/dap/server.go
```diff
@@ -1950,6 +1969,12 @@ func (s *Session) onAttachRequest(request *dap.AttachRequest) {
 	}
 	s.config.log.Debug("parsed launch config: ", prettyPrint(args))
 
+	if s.args.followExec {
+		s.sendShowUserErrorResponse(request.Request, FailedToAttach, "Failed to attach",
+			"Follow exec not supported in attach request yet.")
+		return
+	}
```
firelizzard18: Is this a critical error? Is it better UX for the session to fail outright, or for it to continue as best it can? As I said above, my inclination would be to continue.

service/dap/server.go
Lslightly: "important" OutputEvents are sent back to the client now.

Lslightly:
Hi. Why is the PR stuck now? Is there something else I should do, like documentation? Or merge together with #4171?

derekparker (approved):
LGTM

Apologies for the delay in final review here.

PR #4082, opened 2025-08-04
service: workaround for non-unicode strings in Variables

### LINKED ISSUE #4072: Wrong handling of `string([]byte)` ?
I have a variable which is `[]byte`:

```
(dlv) p p.output
[]uint8 len: 26, cap: 16384, [116,121,112,101,...+22 more]
```

I want to see it as a string:

```
(dlv) p string(p.output)
"type T struct {\f\f\t�fldcom�...+-4 more"
```

First - why so short?  I have `max-array-values` = 4 and `max-string-len` = 256.  This is neither of those.

Second: What does `+-4` mean - At first I thought it was just 4, but:

```
(dlv) p string(p.output[8:])
"truct {\f\f\t�fldcom�...+-4 more"
```

???  I can only guess those mystery characters are screwing something up?  Here's raw data:

```
(dlv) p p.output[0:26]
[]uint8 len: 26, cap: 16384, [116,121,112,101,32,84,32,115,116,114,117,99,116,32,123,12,12,9,255,102,108,100,99,111,109,255]
```

My environment:

```
$ dlv version
Delve Debugger
Version: 1.25.0
Build: $Id: e323af07680631b9cbdcc9d02e0a37891d12d972 $
```

```
$ go version
go version go1.24.1 linux/amd64
```

```
$ uname -a
Linux thockin-glaptop4 6.12.27-1rodete1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.27-1rodete1 (2025-05-19) x86_64 GNU/Linux
```

aarzilli:
When string Variables containing non-unicode strings are encoded into
json their value is changed, which leads to subtle bugs such as
described in issue #4072.

This CL addresses the problem by adding a new field to the api.Variable
struct that will be used to transport the variable's value for strings
that can not be unicode-encoded.

This is the best we can do without breaking backwards compatibility.

For our Go code the change is handled transparently by a
MarshalJSON/UnmarshalJSON on the api.Variable type.

Fixes #4072

### DIFF
--- a/service/api/types.go
+++ b/service/api/types.go
@@ -2,11 +2,13 @@ package api
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/go-delve/delve/pkg/proc"
 )
@@ -326,6 +328,14 @@ type Variable struct {
 	// Function variables will store the name of the function in this field
 	Value string `json:"value"`
 
+	// StringValueBytes contains the Value for string variables that can not be
+	// represented in unicode.
+	// Specifically if v.Kind is reflect.String and v.Value is not a valid
+	// unicode string the JSON marshaller will convert it to a valid unicode
+	// string, altering its contents. This fields contains the unaltered string
+	// contents.
+	StringValueBytes []byte
+
 	// Number of elements in an array or a slice, number of keys for a map, number of struct members for a struct, length of strings, number of captured variables for functions
 	Len int64 `json:"len"`
 	// Cap value for slices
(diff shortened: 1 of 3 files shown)


service/api/types.go
```diff
@@ -353,6 +363,36 @@ type Variable struct {
 	DeclLine int64
 }
 
+type variableWithoutMarshaler Variable
+
+func (v *Variable) MarshalJSON() (text []byte, err error) {
+	var v2 *variableWithoutMarshaler = (*variableWithoutMarshaler)(v)
+	if v2.Kind == reflect.String && !utf8.ValidString(v2.Value) {
+		value := v2.Value
+		// Should we blot out the value with something obviously wrong? Like:
```
derekparker, line 372: I don't think this is a bad idea.

aarzilli: The reason I'm esitant about this is that by not doing it we do send a mangled string (if it is not representable in utf8), but only a little bit. So it's more backwards compatible. However it could also be valuable to make the problem more obvious.

You decide.

derekparker (approved):
LGTM - I'm fine as-is.

PR #4083, opened 2025-08-08
dap: implement readMemory for strings and expose memoryReference

MistaTwista:
## Motivation

Large Go strings are still truncated in common debugging workflows (variables view uses ~512 chars; hover / “Copy Value” use ~4096 chars). That’s often not enough for JSON payloads, logs, etc., and users rely on awkward workarounds (manual slicing, multiple reads).

Adding DAP readMemory enables clients to page the underlying bytes on demand and render full values.
Aug-09-2025 00-51-53

Fixes: https://github.com/go-delve/delve/issues/4084

Vscode-go feature: https://github.com/golang/vscode-go/issues/3817

### DIFF
--- a/_fixtures/readmem_json.go
+++ b/_fixtures/readmem_json.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"runtime"
+	"strings"
+	"unsafe"
+)
+
+// Long json for MemoryReference
+func main() {
+	runtime.Breakpoint()
+
+	payload := strings.Repeat("AB", 2500)
+
+	b, _ := json.Marshal(struct {
+		Data string `json:"data"`
+	}{Data: payload})
+
+	bytesString := []byte("this\nis\nit")
+	nonprint := []byte{242, 243, 244, 245}
+	maps := map[string]string{"some": "non"}
+
+	jsonString := string(b)
+
+	hashed := sha256.Sum256(b)
+	jsonHash := hex.EncodeToString(hashed[:]) // used to validate fullness of a string
+
+	ptr := unsafe.StringData(jsonString)
+	jsonAddr := fmt.Sprintf("%p", ptr) // used to validate string address
+
+	runtime.Breakpoint()
+
+	fmt.Println(jsonString)
+	fmt.Println(jsonHash)
+	fmt.Println(jsonAddr)
+	fmt.Println(bytesString)
+	fmt.Println(nonprint)
+	fmt.Println(maps)
+}
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -11,6 +11,7 @@ package dap
 import (
 	"bufio"
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -104,6 +105,71 @@ type Server struct {
 	sessionMu sync.Mutex
 }
 
+// memRef describe address and its size to stream data from
+type memRef struct {
+	addr uint64
+	size int64
+}
+
+type referencesCollection struct {
+	mu   sync.Mutex
+	refs map[string]memRef
+}
+
+func (r *referencesCollection) get(reference string) (memRef, bool) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	ref, ok := r.refs[reference]
+
+	return ref, ok
+}
+
+func isAddressable(v *proc.Variable) bool {
+	if v == nil || v.Unreadable != nil {
+		return false
+	}
+
+	switch v.Kind {
+	case reflect.Slice, reflect.String:
+		return true
+	}
+
+	return false
+}
+
+func (r *referencesCollection) put(v *proc.Variable) string {
+	if !isAddressable(v) {
+		return ""
+	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.refs == nil {
+		r.refs = make(map[string]memRef)
+	}
+
+	addr := v.Addr
+	if v.Base != 0 {
+		addr = v.Base
+	}
+
+	ref := fmt.Sprintf("0x%x", addr)
+	r.refs[ref] = memRef{addr: addr, size: v.Len}
+
+	return ref
+}
+
+func (r *referencesCollection) reset() {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if len(r.refs) > 0 {
+		r.refs = make(map[string]memRef)
+	}
+}
+
 // Session is an abstraction for serving and shutting down
 // a DAP debug session with a pre-connected client.
 // TODO(polina): move this to a different file/package
@@ -119,6 +185,8 @@ type Session struct {
 	// Reset at every stop.
 	// See also comment for convertVariable.
 	variableHandles *handlesMap[*fullyQualifiedVariable]
+	// referencesCollection track references map for DAP client
+	referencesCollection referencesCollection
 	// args tracks special settings for handling debug session requests.
 	args launchAttachArgs
 	// exceptionErr tracks the runtime error that last occurred.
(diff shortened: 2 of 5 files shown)


service/dap/server.go
```diff
@@ -3261,10 +3323,138 @@ func (s *Session) onLoadedSourcesRequest(request *dap.LoadedSourcesRequest) {
...
+		Response: *newResponse(req),
+		Body: dap.ReadMemoryResponseBody{
+			Address:         fmt.Sprintf("%#x", addr),
+			Data:            response,
+			UnreadableBytes: unreadable,
+		},
+	}
+}
+
+type readPlan struct {
```
aarzilli: There is no reason for this type and its associated function to exist, they are only used once each in a single function, they can be inlined.

vendor/github.com/google/go-dap/schematypes.go
```diff
@@ -451,6 +451,7 @@ type InitializeRequestArguments struct {
 	SupportsVariablePaging              bool   `json:"supportsVariablePaging,omitempty"`
 	SupportsRunInTerminalRequest        bool   `json:"supportsRunInTerminalRequest,omitempty"`
 	SupportsMemoryReferences            bool   `json:"supportsMemoryReferences,omitempty"`
+	SupportsReadMemoryRequest           bool   `json:"supportsReadMemoryRequest,omitempty"`
```
aarzilli: You can't change vendored packages like this, you need to make a PR to google/go-dap first.

MistaTwista: yeap, i know, i did https://github.com/google/go-dap/pull/94

service/dap/server.go
MistaTwista: wanted to do it more readable and explicit, but this make sense

vendor/github.com/google/go-dap/schematypes.go
aarzilli: Great!

service/dap/server.go
MistaTwista: done 👍

service/dap/server.go
```diff
@@ -3261,10 +3323,109 @@ func (s *Session) onLoadedSourcesRequest(request *dap.LoadedSourcesRequest) {
...
+	}
+
+	endReq := int64(args.Offset + args.Count)
+
+	startRead := clamp(int64(args.Offset), 0, ref.size)
+	endRead := clamp(endReq, 0, ref.size)
+
+	readCount := endRead - startRead
+	if readCount < 0 {
+		readCount = 0
```
aarzilli:
```
if readCount <= 0 {
s.send(makeReadMemoryResponse(...))
return
}
```

nothing good is happening after this if readCount is equal or less than zero

MistaTwista: done, thanks, I was thinking about it

aarzilli:
LGTM but held because we need the go-dap change.
@h9jiang do you know who's maintaining google/go-dap now?

h9jiang:
Hi @aarzilli 

I think the go tools team is the maintainer of google/go-dap, I will discuss this within the team and review the PR and CL that @MistaTwista drafted.

MistaTwista:
sorry for inconvenience with [go-dap PR](https://github.com/google/go-dap/pull/94). I closed it because it looks like we don't need those changes here 😐

aarzilli:
Many tests fail.

aarzilli:
PS. the TODO on line service/dap/server.go:841 needs to be removed.

MistaTwista:
Hey @aarzilli, I’ve fixed the previously failing tests and also added a new one for my changes to be sure everything works. 

The main idea of this new test is to generate a large JSON payload, retrieve it from the Delve server via the read-memory request page-by-page, compute a hash, and compare it with the original hash to ensure there are no duplicates or missing data.

All tests pass locally, but the CI still fails for my test with:

```bash
=== RUN   TestReadMemory_StringPagination
Error compiling ../../_fixtures/readmem_json.go: wait6: no child processes

FAIL	github.com/go-delve/delve/service/dap	50.196s
?   	github.com/go-delve/delve/service/dap/daptest	[no test files]
?   	github.com/go-delve/delve/service/dap/daptest/gen	[no test files]
```

It seems like an environment difference between my setup and the CI runner, and I can’t reproduce it locally. Do you have any idea what might be wrong?

aarzilli:
I think the reason the tests do not pass is just because your branch is fairly old and needs to be rebased.
However, I also tested this with vscode and I don't get the "open in new document" option that you have.

MistaTwista:
> I think the reason the tests do not pass is just because your branch is fairly old and needs to be rebased.

Ok, I'll rebase it and try again, thanks

> However, I also tested this with vscode and I don't get the "open in new document" option that you have.

I'm not sure I understand you correctly. In the description of the original [issue](https://github.com/golang/vscode-go/issues/3817) I showed what functionality I'm trying to archive in vscode and why.

And this [comment](https://github.com/golang/vscode-go/issues/3817#issuecomment-3170404359) is about why I’m proposing delve changes.

aarzilli:
So, this would also require a change to vscode?

MistaTwista:
not vscode itself, but vscode-go plugin for vscode https://github.com/golang/vscode-go/pull/3818

service/dap/server_test.go
```diff
@@ -7894,6 +7896,101 @@ func TestBreakpointAfterDisconnect(t *testing.T) {
 	cmd.Process.Kill()
 }
 
+func TestReadMemory_StringPagination(t *testing.T) {
+	signal.Reset(syscall.SIGCHLD)
```
aarzilli: Why?

_fixtures/readmem_json.go
```diff
@@ -0,0 +1,30 @@
...
+	}{Data: payload})
+
+	jsonString := string(b)
+
+	hashed := sha256.Sum256(b)
+
+	jsonHash := hex.EncodeToString(hashed[:])
+
+	runtime.Breakpoint()
+	_ = jsonString
```
aarzilli: This is not enough to make sure that these strings are alive. You should pass them to fmt.Println.

service/dap/server.go
```diff
@@ -104,6 +105,61 @@ type Server struct {
...
 }
 
+// memRef describe address and its size to stream data from
+type memRef struct {
+	addr uint64
+	size int64
+}
+
+type referencesCollection struct {
+	mu   sync.Mutex
```
aarzilli, line 115: Looking at how you are using referencesCollection it seems to me that it's only ever used by requests served synchronously. Am I wrong?

service/dap/server.go
```diff
@@ -104,6 +105,61 @@ type Server struct {
...
+	refs map[string]memRef
+	seq  uint64
```
aarzilli: I don't think you need this variable, it will always be equal to `len(refs)`

service/dap/server.go
MistaTwista: I'm not fully confident about synchronicity of a callbacks in Delve that's why I'm using mutex here. Are requests within a session always strictly sequential? If so, then we could probably drop the mutex. But if there's any chance multiple requests could run in parallel (now or in the future), I'd rather keep it to avoid hard‑to‑catch issues or fatals.

service/dap/server_test.go
MistaTwista: I don't understand why the tests are failing in CI but not locally. I thought it might be related to signals and tried fixing it, but that didn't solve the issue. I'll remove that change.

MistaTwista:
> I think it would be prudent to put memory references for disassembly and the ones for the memory of variables in the same address space, in case clients expects them to be

done 👌 test is also updated for the new format, but still there are some issues

aarzilli:
I don't know why your test is failing like that on freebsd, there is no obvious problem that would lead to that. I think it's fine to just add a skip for freebsd (`if runtime.GOOS == "freebsd" { t.Skip(...) }`)

MistaTwista:
Screenshot 2025-10-02 at 3 32 59 PM
Screenshot 2025-10-02 at 3 32 51 PM

I'm thinking about support that kind of strings representation too. Make it possible in this PR or make the new one? Current version gives me error for them like

Screenshot 2025-10-02 at 3 36 08 PM

But it looks obvious for me that this is a good case to support. What do you think?

aarzilli:
The []byte variable, sure, not the `string()` thing we return.

MistaTwista:
Oct-04-2025 22-09-41
I've added slices

Oct-04-2025 22-12-52
and removed max limit to get memory address, because it can be useful to check even small variables in a hex editor

aarzilli (approved):
LGTM

MistaTwista:
> LGTM

Do I need to do something (like rebase to current master) or just wait until merge?

aarzilli:
Just wait.

derekparker:
@MistaTwista this patch does need a rebase now. After the rebase I will merge.

MistaTwista:
Done, would you like me to squash the commits, or should I improve the commit messages for better documentation?

MistaTwista:
fixed conflict with master, squashed commits, made better description of a commit

Issue #4080, opened 2025-08-04
`unreadable could not resolve interface type` value for likely optimized out variables


nickfromXXII:
Delve reports that it can't resolve empty interface typed variable starting with go1.25rc1 if it has been assigned with an empty struct. It seems that the compiler started optimizing them out recently. It has been working fine before, checked with go1.24. It still works as expected if the value's struct type has some fields though.

#### Steps to reproduce:

1. Create a new empty go module with following main.go file:
```go
package main

type T struct{}

func main() {
	var i interface{} = T{}
	println(i) //break
}
```
2. Start debugging: `$ dlv debug .`
2. Set a breakpoint at the line 7 (where the comment is) : `(dlv) break main:go:7`
3. Continue the program: `(dlv) continue`
4. Request list of local variables: `(dlv) locals`

#### Expected result: (reproducible with go1.24)
`i = interface {}(main.T) {}`

#### Actual result: (repoducible with go1.25rc1 and go1.25rc2)
`i = (unreadable could not resolve interface type)`

aarzilli:
CL: https://go-review.googlesource.com/c/go/+/696955

PR #4096, opened 2025-08-18
proc: add test for interfaces with otherwise unreachable runtime types

aarzilli:
### proc: add test for interfaces with otherwise unreachable runtime types

Add a test to read the runtime type of an interface variable that is
not used in any other ways except through interfaces.

This is a regression check for issue #4080 which is a compiler problem.

### proc: early fixes for Go 1.26

Change 091ff506 fixed loading the direct flag for Go 1.26 when we go
from a runtime type to a dwarf type, we also did the same check from
the dwarf type to a runtime type when converting variables into
'interface {}'.

This commit deletes unifies the two code paths deleting the one that
wasn't updated.

Fixes the tests on tip.

### *: fix some failing tests on go1.25

* Skip TestIssue256 on linux/386, it fails because the line it relies
on
(a return) does not have code generated for it.

* Skip running gen-usage-docs in TestGeneratedDocs on linux/ppc64le, I
don't know why this fails, likely just a problem with the build
agent,
but it doesn't matter: the test doesn't need to run on every
architecture.

* Change TestRestartRequestWithNewArgs so that it doesn't depend on the
cap value of a slice (it changed in go1.25/windows/amd64)

### DIFF
--- a/_fixtures/testvariables2.go
+++ b/_fixtures/testvariables2.go
@@ -169,6 +169,10 @@ type ThreeInts struct {
 	a, b, c int
 }
 
+type OnlyUsedInInterface struct {
+	s string
+}
+
 type Message []byte
 
 var _ I = (*W2)(nil)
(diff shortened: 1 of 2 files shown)


aarzilli:
1.25.2 has been releasaed with the fix, we can now merge this as a regression test.

PR #4103, opened 2025-08-22
service/dap: allow clients to specify rebuild on restart

derekparker:
Adds a new param to the restart request which if set will cause the debugger to recompile the program before restarting it.

Note the existing behavior is unchanged, clients *must* specify this new flag to trigger a recompile.

Fixes https://github.com/golang/vscode-go/issues/3835#issuecomment-3204417618

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -1003,8 +1003,12 @@ func (s *Session) onLaunchRequest(request *dap.LaunchRequest) {
 
 		switch args.Mode {
 		case "debug":
+			s.config.Debugger.ExecuteKind = debugger.ExecutingGeneratedFile
+			s.config.Debugger.Packages = []string{args.Program}
 			cmd, out, err = gobuild.GoBuildCombinedOutput(args.Output, []string{args.Program}, args.BuildFlags.value)
 		case "test":
+			s.config.Debugger.ExecuteKind = debugger.ExecutingGeneratedTest
+			s.config.Debugger.Packages = []string{args.Program}
 			cmd, out, err = gobuild.GoTestBuildCombinedOutput(args.Output, []string{args.Program}, args.BuildFlags.value)
 		}
 		args.DlvCwd, _ = filepath.Abs(args.DlvCwd)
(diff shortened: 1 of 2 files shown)


service/dap/server.go
```diff
@@ -3086,13 +3090,15 @@ func (s *Session) onRestartRequest(request *dap.RestartRequest) {
...
 	var resetArgs bool
+	var rebuild bool
 
 	// Update launch/attach arguments if provided
 	if len(request.Arguments) > 0 {
 		var restartArgs struct {
 			Arguments struct {
 				Request string `json:"request"`
 				LaunchConfig
+				Rebuild bool `json:"rebuild,omitempty"`
```
aarzilli: How is this going to be specified by, for example, vscode? Should we make this default to rebuilding since it was the previous behavior?

service/dap/server.go
```diff
@@ -3112,7 +3119,7 @@ func (s *Session) onRestartRequest(request *dap.RestartRequest) {
 	}
 
 	// TODO(deparker) handle args from the launch regarding re-recording.
-	discardedBreakpoints, err := s.debugger.Restart(false, "", resetArgs, newArgs, [3]string{}, false)
+	discardedBreakpoints, err := s.debugger.Restart(false, "", resetArgs, newArgs, [3]string{}, rebuild)
```
aarzilli, line 3124: How can this work? The rebuild flag of restart only works if the configuration of the debugger has `ExecuteKind` set to `ExecutingGeneratedFile` or `ExecutingGeneratedTest` but I don't see the dap server setting this and it does its own compilation.

service/dap/server.go
derekparker: I figured the client could provide the extra parameter as we've made it available. However, I'm fine with preserving the original behavior instead.

service/dap/server.go
derekparker: Did you miss this: https://github.com/go-delve/delve/pull/4103/files#diff-aa525563efc47874aa048791ab73f4bc5ca312fae0d989ba08b809fb1dc8a7a1R1006 ?

aarzilli: Yes.

aarzilli (approved):
LGTM

PR #4106, opened 2025-08-27
cmd/dlv: add capslock check

derekparker:
The capslock program analyzes packages and dependencies for "capabilities" that those packages have. Adding a baseline set of capabilities and a test validating the capabilities will ensure that updates to our dependencies do not introduce new capabilities without further explicit review and approval by a maintainer.

https://github.com/google/capslock


go.mod
```diff
@@ -16,19 +16,22 @@ require (
...
+	golang.org/x/sys v0.35.0
+	golang.org/x/telemetry v0.0.0-20250807160809-1a19826ec488
+	golang.org/x/tools v0.36.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
 require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.6 // indirect
+	github.com/fatih/color v1.18.0 // indirect
+	github.com/google/capslock v0.2.8 // indirect
```
aarzilli: Unless I'm missing something capslock is never imported and the install script uses `@latest` so this shouldn't be needed, just like we don't depend on staticcheck.

derekparker: No you're right, I think this is leftover from when I was trying to "tool" approach, but that isn't supported until 1.24.

aarzilli:
Conflict with the trie update I just merged.

aarzilli:
Also I don't see the go-cmp dependency being imported either and all the updates to the other dependencies shouldn't be needed and are inappropriate if we are going to do the https://github.com/go-delve/build-tools split.

derekparker:
> Also I don't see the go-cmp dependency being imported either and all the updates to the other dependencies shouldn't be needed and are inappropriate if we are going to do the https://github.com/go-delve/build-tools split.

Cleared all that out, the go.mod file got a bit wonky when I tried using tool support. Reverted all that.

_scripts/capslock-out.txt
```diff
@@ -0,0 +1,40 @@
+Capslock is an experimental tool for static analysis of Go packages.
```
aarzilli: I think this file should be deleted now, if I'm not missing anything only the capslock_platform_output.txt files are used.

derekparker: Done.

aarzilli (approved):
LGTM

PR #4107, opened 2025-08-27
*: update github.com/derekparker/trie

derekparker:
Includes improvements including generic support (which we don't leverage) but also a significant amount of performance and memory usage / allocation improvements, and support for iterators.


pkg/terminal/terminal.go
```diff
@@ -281,8 +281,8 @@ func (t *Term) Run() (int, error) {
 	signal.Notify(ch, syscall.SIGINT)
 	go t.sigintGuard(ch, multiClient)
 
-	fns := trie.New()
-	cmds := trie.New()
+	fns := trie.New[any]()
```
aarzilli, line 284: Why do we not make this type specific?

derekparker: The type param is for the value, keys are always strings. In this case, we don't utilize the value ever, so it's fine to be `any`.

aarzilli (approved):
LGTM

PR #4108, opened 2025-08-27
pkg/goversion: update MinSupportedVersionOfGoMinor

derekparker:
The oldest version of Go we run in CI is 1.23,
MinSupportedVersionOfGoMinor should reflect that.


aarzilli (approved):
LGTM

PR #4113, opened 2025-08-29
pkg/proc: optimize range body lookup cost by trie searching and use moduledata cache to reduce LoadModuleData cost

### LINKED ISSUE #4112: Performance: optimize the overhead of loadValue
I have a rather extreme scenario: when retrieving package values, the excessive number of variable/function declarations causes the delve's `vars` command to time out (taking 30 minutes). After observing the cpu profile, I identified the following optimizable performance issues:
- Modify LoadModuleData to an O(1) operation and reuse the moduleDataCache;
- Change the retrieval of functionExtra to use a trie to avoid O(n*m) lookups;
Image

jayantxie:
Fixes #4112

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -24,6 +24,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/derekparker/trie/v3"
+	"github.com/hashicorp/golang-lru/simplelru"
+
 	"github.com/go-delve/delve/pkg/astutil"
 	pdwarf "github.com/go-delve/delve/pkg/dwarf"
 	"github.com/go-delve/delve/pkg/dwarf/frame"
@@ -37,7 +40,6 @@ import (
 	"github.com/go-delve/delve/pkg/logflags"
 	"github.com/go-delve/delve/pkg/proc/debuginfod"
 	"github.com/go-delve/delve/pkg/proc/evalop"
-	"github.com/hashicorp/golang-lru/simplelru"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/proc/bininfo.go
```diff
@@ -2642,6 +2650,16 @@ func (bi *BinaryInfo) LookupFunc() map[string][]*Function {
 	return bi.lookupFunc
 }
 
+func (bi *BinaryInfo) LookupRangeBodyFunc() *trie.Trie {
+	if bi.lookupRangeBodyFunc == nil {
```
aarzilli, line 2647: Needs to be cleared when we load plugins.

derekparker:
Hey @jayantxie thanks for the patch!

As @aarzilli mentioned we just updated `trie` internally to `v3` which adds support for generics amongst some other changes & optimizations. It should simplify your changes once you rebase because you won't need the type cast anymore.

pkg/proc/bininfo.go
jayantxie: I think it will be cleared by `loadDebugInfoMaps` when loading plugins, just like `lookupFunc` and `lookupGenericFunc`.

aarzilli (approved):

pkg/proc/bininfo.go
```diff
@@ -766,10 +770,12 @@ func (fn *Function) extra(bi *BinaryInfo) *functionExtra {
 
 	// Find range-over-func bodies of this function
 	if fn.extraCache.rangeParent == nil {
-		for i := range bi.Functions {
-			fn2 := &bi.Functions[i]
-			if strings.HasPrefix(fn2.Name, fn.Name) && fn2.rangeParentName() == fn.Name {
-				fn.extraCache.rangeBodies = append(fn.extraCache.rangeBodies, fn2)
+		lookupFunc := bi.LookupRangeBodyFunc()
+		for _, key := range lookupFunc.PrefixSearch(fn.Name) {
```
derekparker, line 774: Could probably improve this by introducing a `PrefixSearchIter` func which returns keys & values (this is more of a note for myself to fix and refactor this in the future).

PR #4115, opened 2025-08-29
*: parallelize tests where possible

derekparker:
Run as many tests as we can in parallel. This change knocks off almost 1 minute from the full suite run `make test`. There are several tests which lend themselves nicely to running parallel, namely unit tests which don't do any actually debugging or process manipulation, but also the tests in `cmd/dlv`. Those tests exec a compiled version of delve and so we can safely run multiple of the tests in parallel as the forked delve processes will each do their own signal handling, etc... which is unsafe for us to do in the majority of the `pkg/proc` tests as the test binary ends up acting as the debugger.

This leads me to the though process that most tests in `pkg/proc` should be converted to true integration tests similar to the `cmd/dlv` test suite, allowing up to parallelize those expensive tests as well.


derekparker:
I see TestStaticCheck is failing, I'll look into this.

derekparker:
These tests which are now parallel have all been run with the race detector and do not have any data races found. I would like to run the race detector in CI by default, however we actually do have some existing races (at least one of which is intentional to trigger a bug and prevent a regression) in `service/dap` so these will have to be ironed out first.

aarzilli (approved):
LGTM

Issue #4116, opened 2025-08-30
dap: evaluate command returns a mismatched type


wenxuan70:
dlv version: `1.25.0`

go version: `go version go1.21.13 linux/amd64`

When I mouse is hover on `stmt.Model`, vscode-go initiates the request for evaluation. But dlv-dap returns the method in the `DB` field embedded in the `Statement`, rather than the `Model` field in the `Statement`.

<img width="872" height="350" alt="Image" src="https://github.com/user-attachments/assets/ab2fe578-989e-4dc3-a32e-b906ba936a6d" />

<img width="1138" height="392" alt="Image" src="https://github.com/user-attachments/assets/244c627f-0acb-4a58-9108-968211240652" />

<img width="2146" height="1028" alt="Image" src="https://github.com/user-attachments/assets/87e165db-45aa-4a65-b6c5-1e93f4459d8c" />

wenxuan70:
From the perspective of go syntax, when the method name of the embedded field is the same as the original structure field, the corresponding method should not be obtained.

https://github.com/go-delve/delve/blob/7cb015b4397deace7e81f00737c1722215a56257/pkg/proc/eval.go#L3009

This code can reproduce this scenario:
https://go.dev/play/p/sfXtceVrdNp

PR #4118, opened 2025-08-30
pkg/proc: hierarchical search structMember or method

wenxuan70:
fix go-delve/delve#4116

Let `eval` get the value of `struct.field` more in line with the syntax of go

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -2112,16 +2112,7 @@ func (scope *EvalScope) evalStructSelector(op *evalop.Select, stack *evalStack) 
 		return
 	}
 
-	rv, err := xv.findMethod(op.Name)
-	if err != nil {
-		stack.err = err
-		return
-	}
-	if rv != nil {
-		stack.push(rv)
-		return
-	}
-	stack.pushErr(xv.structMember(op.Name))
+	stack.pushErr(xv.findStructMemberOrMethod(op.Name, true))
 }
 
 // Evaluates expressions <subexpr>.(<type>)
@@ -3005,18 +2996,60 @@ func (v *Variable) reslice(low int64, high int64, trustLen bool) (*Variable, err
 	return r, nil
 }
 
-// findMethod finds method mname in the type of variable v
-func (v *Variable) findMethod(mname string) (*Variable, error) {
-	if _, isiface := v.RealType.(*godwarf.InterfaceType); isiface {
-		v.loadInterface(0, false, loadFullValue)
+// findStructMemberOrMethod finds struct member or method name in the type of variable v
+func (v *Variable) findStructMemberOrMethod(name string, includeStructMember bool) (*Variable, error) {
+	if v.Unreadable != nil {
+		return v.clone(), nil
+	}
+	vname := v.Name
+	if v.Name == "" {
+		vname = v.DwarfType.String()
+	}
+	if v.loaded && (v.Flags&VariableFakeAddress) != 0 {
+		for i := range v.Children {
+			if v.Children[i].Name == name {
+				return &v.Children[i], nil
+			}
+		}
+		return nil, fmt.Errorf("%s has no member %s", vname, name)
+	}
+	closure := false
+	switch v.Kind {
+	case reflect.Chan:
+		v = v.clone()
+		v.RealType = godwarf.ResolveTypedef(&(v.RealType.(*godwarf.ChanType).TypedefType))
+	case reflect.Interface:
+		v.loadInterface(0, false, LoadConfig{})
+		if len(v.Children) > 0 {
+			v = &v.Children[0]
+		}
+	case reflect.Func:
+		fn := v.bi.PCToFunc(v.Base)
+		v.loadFunctionPtr(0, LoadConfig{MaxVariableRecurse: -1})
 		if v.Unreadable != nil {
+			cst := fn.extra(v.bi).closureStructType
+			if cst == nil || cst.ByteSize == 0 {
+				// Not a closure, normal function
+				if _, ok := v.bi.PackageMap[vname]; ok {
+					return nil, fmt.Errorf("package %s has no function %s", vname, name)
+				}
+				return nil, fmt.Errorf("%s has no member %s", vname, name)
+			}
 			return nil, v.Unreadable
 		}
-		return v.Children[0].findMethod(mname)
+		if v.closureAddr != 0 {
+			fn = v.bi.PCToFunc(v.Base)
+			if fn != nil {
+				cst := fn.extra(v.bi).closureStructType
+				v = v.newVariable(v.Name, v.closureAddr, cst, v.mem)
+				closure = true
+			}
+		}
 	}
 
 	queue := []*Variable{v}
 	seen := map[string]struct{}{}
+	first := true
 
 	for len(queue) > 0 {
 		v := queue[0]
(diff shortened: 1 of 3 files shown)


wenxuan70:
You're absolutely correct. After further consideration, the proper approach here should be to implement a hierarchical search for methods and struct members, prioritizing returning the first encountered ones. However, this could involve non-trivial code changes. If possible, could you give me some time to work on fixing this issue? Alternatively, do you have any other better approaches or suggestions?

wenxuan70:
An implementation for reference: https://github.com/golang/go/blob/6a08e80399bd65b95e60e3c74b7e1f86754752a7/src/go/types/lookup.go#L144

aarzilli:
We already have the hierarchical search in findMethod and structMember, they need to be fused.

pkg/proc/eval.go
```diff
@@ -2112,16 +2112,12 @@ func (scope *EvalScope) evalStructSelector(op *evalop.Select, stack *evalStack)
 		return
 	}
 
-	rv, err := xv.findMethod(op.Name)
+	v, err := xv.findStructMemberOrMethod(op.Name)
```
aarzilli: `stack.pushErr(xv.findStructMemberOrMethod(op.Name))` instead of this line and everything that follows.

pkg/proc/eval.go
```diff
@@ -3005,6 +3001,148 @@ func (v *Variable) reslice(low int64, high int64, trustLen bool) (*Variable, err
...
+					return structVar.toField(field)
+				}
+				if len(queue) == 0 && field.Name == "&"+name && closure {
+					f, err := structVar.toField(field)
+					if err != nil {
+						return nil, err
+					}
+					return f.maybeDereference(), nil
+				}
+				isEmbeddedStructMember :=
```
aarzilli: Delete all of this and just use field.Embedded. It's been there since 2017, it's legacy code we don't need anymore.

pkg/proc/eval.go
```diff
@@ -3005,6 +3001,148 @@ func (v *Variable) reslice(low int64, high int64, trustLen bool) (*Variable, err
...
+		case *godwarf.InterfaceType:
+			v.loadInterface(0, false, LoadConfig{})
+			if len(v.Children) > 0 {
+				if rv, _ := v.Children[0].findMethod(name); rv != nil {
+					return rv, nil
+				}
+			}
+		default:
+			if first {
+				return nil, fmt.Errorf("%s (type %s) is not a struct", vname, structVar.TypeString())
```
aarzilli: Return `fmt.Errorf("%s (type %s) has no member %s", vname, structVar.TypeString(), name)` instead.

pkg/proc/eval.go
```diff
@@ -3044,49 +3182,88 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
...
+				}
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+func lookupMethod(v *Variable, isptr bool, pkg, receiver, name string) (*Variable, error) {
+	if isptr {
```
aarzilli: Can you explain this check? IIRC go won't allow the same method name on pointer and non-pointer receiver, it shouldn't matter which order you look them up.

pkg/proc/variables_test.go
```diff
@@ -2053,3 +2053,72 @@ func TestCapturedVarVisibleOnFirstLine(t *testing.T) {
...
+					t.Fatalf("Expected error %s, got no error: %s\n", tc.err.Error(), api.ConvertVar(variable).SinglelineString())
+				}
+				if tc.err.Error() != err.Error() {
+					t.Fatalf("Unexpected error. Expected %s got %s", tc.err.Error(), err.Error())
+				}
+			}
+		}
+
+		for _, tc := range funcTestcases {
+			testCallFunction(t, grp, p, tc)
```
aarzilli: This is too much, just check that the returned variable has the correct name, calling them is slow and complicates things.

pkg/proc/eval.go
```diff
@@ -3005,6 +3001,148 @@ func (v *Variable) reslice(low int64, high int64, trustLen bool) (*Variable, err
...
+			}
+		}
+		first = false
+	}
+
+	return nil, fmt.Errorf("%s has no member %s", vname, name)
+}
+
 // findMethod finds method mname in the type of variable v
 func (v *Variable) findMethod(mname string) (*Variable, error) {
```
aarzilli: This now never gets called except through the code of findStructMemberOrMethod, it should be deleted (it's a lot of duplicated code) and a flag should be added to findStructMemberOrMethod that enables or disables struct member lookup.

pkg/proc/eval.go
wenxuan70: When a struct is embedded in an interface type, different usage methods will be different.

https://go.dev/play/p/5fXZE_HB_cL

image

wenxuan70:
Thanks for your review – I'll address these issues today.

Besides this point, I also realized there's a corner case that hasn't been resolved: specifically, we can't retrieve only the interface methods implemented by a struct.

For example: https://go.dev/play/p/VVD5if7OcDQ

In this example, it's not possible to determine that `main.(*B).V` belongs to `interface A`.

aarzilli:
> Besides this point, I also realized there's a corner case that hasn't been resolved: specifically, we can't retrieve only the interface methods implemented by a struct.

This is correct, that information does not exist in dwarf. I don't think it's worth bothering with it, it's too much of a corner case.

pkg/proc/eval.go
wenxuan70: I removed the `findMethod` method and added a parameter `includeStructMember` to the `findStructMemberOrMethod` method to control whether to query struct members.

pkg/proc/eval.go
```diff
@@ -3032,61 +3065,138 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
...
+
+func lookupMethod(v *Variable, isptr bool, pkg, receiver, name string) (*Variable, error) {
+	if isptr {
+		if fns := v.bi.LookupFunc()[fmt.Sprintf("%s.(*%s).%s", pkg, receiver, name)]; len(fns) == 1 {
 			r, err := functionToVariable(fns[0], v.bi, v.mem)
 			if err != nil {
 				return nil, err
 			}
 			if isptr {
-				r.Children = append(r.Children, *(v.maybeDereference()))
```
aarzilli, line 3053: This still has an `if isptr` within an `if isptr`.

pkg/proc/variables_test.go
```diff
@@ -2053,3 +2053,61 @@ func TestCapturedVarVisibleOnFirstLine(t *testing.T) {
...
+		{"v5.B5.X", false, "main.main.func1", "", "func() string", nil},
+		{"v5.B5.A5.X", false, "main.main.func1", "", "func() string", nil},
+		{"v5.A5.X", false, "main.main.func1", "", "func() string", nil},
+		{"*(ch.buf)", true, "[0]int []", "", "[0]int", nil},
+		{"*(v6.Chan.buf)", true, "[0]struct struct {} []", "", "[0]struct struct {}", nil},
+		{"*(v6.buf)", false, "", "", "", errors.New("v6 has no member buf")},
+	}
+
+	withTestProcess("issue4116", t, func(p *proc.Target, grp *proc.TargetGroup, fixture protest.Fixture) {
+		testCallFunctionSetBreakpoint(t, p, grp, fixture)
```
aarzilli: This call doesn't do anything, the program will stop at `runtime.Breakpoint` anyway.

pkg/proc/eval.go
```diff
@@ -3079,14 +3090,97 @@ func (v *Variable) findMethod(mname string) (*Variable, error) {
...
+				return nil, fmt.Errorf("%s (type %s) has no member %s", vname, structVar.TypeString(), name)
+			}
+		}
+		first = false
+	}
+
+	return nil, fmt.Errorf("%s has no member %s", vname, name)
+}
+
+func lookupMethod(v *Variable, isptr bool, pkg, receiver, name string) (*Variable, error) {
```
aarzilli, line 3148: This method is still very repetitive, I think it should be rewritten like this:

```
func lookupMethod(v *Variable, isptr bool, pkg, receiver, name string) (*Variable, error) {
	checks := []struct {
		fmt     string
		ptrRecv bool
	}{
		{"%s.(*%s).%s", true},
		{"%s.%s.%s", false},
	}
	if !isptr {
		checks[0], checks[1] = checks[1], checks[0]
	}

	for _, check := range checks {
		if fns := v.bi.LookupFunc()[fmt.Sprintf(check.fmt, pkg, receiver, name)]; len(fns) == 1 {
			r, err := functionToVariable(fns[0], v.bi, v.mem)
			if err != nil {
				return nil, err
			}
			switch {
			case isptr == check.ptrRecv:
				r.Children = append(r.Children, *v)
			case isptr && !check.ptrRecv:
				r.Children = append(r.Children, *(v.maybeDereference()))
			case !isptr && check.ptrRecv:
				r.Children = append(r.Children, *(v.pointerToVariable()))
			}
			return r, nil
		}
	}
	
	return nil, nil
}
```

pkg/proc/variables_test.go
```diff
@@ -2053,3 +2053,59 @@ func TestCapturedVarVisibleOnFirstLine(t *testing.T) {
 		}
 	})
 }
+
+func TestIssue4116(t *testing.T) {
```
aarzilli: We decided a while back that we are no longer going to make tests named after issues because it's hard to remember what they are testing so change this `TestEmbeddedStructMethodsAndFieldLookup` and add this comment on the first line:

```
// See issue #4116
```

pkg/proc/variables_test.go
```diff
@@ -2053,3 +2053,59 @@ func TestCapturedVarVisibleOnFirstLine(t *testing.T) {
...
+	protest.MustSupportFunctionCalls(t, testBackend)
```
aarzilli: This is not needed, there are no function calls anymore here.

pkg/proc/variables_test.go
```diff
@@ -2053,3 +2053,59 @@ func TestCapturedVarVisibleOnFirstLine(t *testing.T) {
...
+	protest.AllowRecording(t)
```
aarzilli: This test will not work with a recording, remove this line too.

pkg/proc/eval.go
wenxuan70: This looks much cleaner.

pkg/proc/variables_test.go
wenxuan70: Done.

pkg/proc/variables_test.go
wenxuan70: Done.

pkg/proc/variables_test.go
wenxuan70: Done.

pkg/proc/eval.go
wenxuan70: Done.

aarzilli (approved):
LGTM

PR #4120, opened 2025-08-31
service: remember that we attached in WaitFor attach mode

### LINKED ISSUE #4119: Attach always terminate waitFor processes
### dlv version
Delve Debugger
Version: 1.25.2
Build: 7cb015b4397deace7e81f00737c1722215a56257

### go version
go version go1.25.0 darwin/arm64

### OS
MacOS 26 (public beta) M2 Max

When running dlv in attach mode (both as a CLI debugger or in DAP mode through an editor) and then disconnecting the debugger, the debuggee is always terminated if `--waitfor` was used to attach to the process.

This can be reproduced through the CLI as well. Say I have a Go HTTP server `myserver`. Then if I run 
```fish
dlv attach --waitfor myserver --log
```
and then run `exit` in dlv repl, the process gets terminated without a prompt.

However if I run
```fish
dlv attach (pgrep myserver) --log
```
and then type `exit` in the repl, I get prompted on whether I want to terminate the process.

Here is also a debug log from the editor (Neovim + nvim-dap)
```
[DEBUG] 2025-08-30 19:48:16 dap/session.lua:1855	"request"	{
  arguments = {
    restart = false,
    terminateDebuggee = false
  },
  command = "disconnect",
  seq = 37,
  type = "request"
}
[DEBUG] 2025-08-30 19:48:16 dap/session.lua:1051	1	{
  body = {
    allThreadsStopped = true,
    reason = "pause"
  },
  event = "stopped",
  seq = 0,
  type = "event"
}
[DEBUG] 2025-08-30 19:48:16 dap/session.lua:1051	1	{
  body = {
    category = "console",
    output = "Detaching and terminating target process\n"
  },
  event = "output",
  seq = 0,
  type = "event"
}
```

As you can see `terminateDebuggee` was set to `false`, however the process still got terminated.

According to my investigation the reason is that whenever there's a need to check whether dlv is attached to a process, only AttachPid is checked to be non-zero.
But when --waitfor is passed - AttachPid is still zero. So waitFor process gets terminated as if it was spawned by dlv.

I tried to fix it myself, but I got lost a little since the implementations for DAP and Term's service client differs. I can try again with some guidance though.
Thank you!

aarzilli:
Save the attached to PID to the configuration when WaitFor is used so
that we later know that we attached and behave identically to when the
normal attach command is used, particularly with respect to killing the
process on exit.

Fixes #4119

### DIFF
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -192,6 +192,7 @@ func New(config *Config, processArgs []string) (*Debugger, error) {
 			err = noDebugErrorWarning(err)
 			return nil, attachErrorMessage(d.config.AttachPid, err)
 		}
+		d.config.AttachPid = d.target.Selected.Pid()
 
 	case d.config.CoreFile != "":
 		var err error
--- a/service/rpc2/server.go
+++ b/service/rpc2/server.go
@@ -731,9 +731,7 @@ type AttachedToExistingProcessOut struct {
 
 // AttachedToExistingProcess returns whether we attached to a running process or not
 func (s *RPCServer) AttachedToExistingProcess(arg AttachedToExistingProcessIn, out *AttachedToExistingProcessOut) error {
-	if s.config.Debugger.AttachPid != 0 {
-		out.Answer = true
-	}
+	out.Answer = s.debugger.AttachPid() != 0
 	return nil
 }
 


derekparker (approved):
LGTM

Issue #4117, opened 2025-08-30
dap: race conditions in tests


derekparker:
The output of `go test -race ./service/dap`

<details><summary>Details</summary>
<p>

DAP server listening at: [::]:41907
2025-08-30T11:09:19-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
DAP server listening at: [::]:41247
2025-08-30T11:09:19-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
DAP server listening at: [::]:38739
2025-08-30T11:09:19-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:38739
DAP server listening at: [::]:39559
2025-08-30T11:09:20-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:39559
DAP server listening at: [::]:43137
2025-08-30T11:09:20-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:43137
DAP server listening at: [::]:37531
2025-08-30T11:09:20-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:37531
DAP server listening at: [::]:35993
2025-08-30T11:09:21-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:35993
DAP server listening at: [::]:35647
2025-08-30T11:09:22-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:35647
DAP server listening at: [::]:40779
2025-08-30T11:09:23-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:40779
4
Connecting to server at: [::]:46149
Connecting to server at: [::]:46191
Connecting to server at: [::]:37613
Connecting to server at: [::]:36693
4
DAP server listening at: [::]:38061
2025-08-30T11:09:27-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
Connecting to server at: [::]:38061
DAP server listening at: [::]:42873
2025-08-30T11:09:28-07:00 warn layer=rpc Listening for remote connections (connections are not authenticated nor encrypted)
[...]

PR #4121, opened 2025-08-31
service/dap: fix nativeProcess race condition in tests

derekparker:
Fixes one race in https://github.com/go-delve/delve/issues/4117.


aarzilli:
The freebsd tests failed at first but I don't think it's this change. If it isn't a fluke I bet it's the parallelization in dlv_test. We'll see.

aarzilli (approved):

PR #4123, opened 2025-09-01
proc,service,terminal: allow cancelling of debuginfod downloads

### LINKED ISSUE #3906: Long pause when debugging a program Raylib or SDL

aarzilli:
Adds an RPC call to cancel in-progress debuginfod-find downloads. Wires
this call to ^C in the terminal interface and to receiving a Pause
request while already pausing in DAP.

Adds a new terminal command to download the debug info for libraries
for which it hasn't been downloaded already.

Fixes #3906

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -88,7 +88,7 @@ Command | Description
 [exit](#exit) | Exit the debugger.
 [funcs](#funcs) | Print list of functions.
 [help](#help) | Prints the help message.
-[libraries](#libraries) | List loaded dynamic libraries
+[libraries](#libraries) | List loaded dynamic libraries.
 [list](#list) | Show source code.
 [packages](#packages) | Print list of packages.
 [source](#source) | Executes a file containing a list of delve commands
@@ -511,7 +511,11 @@ Type "help" followed by the name of a command for more information about it.
 Aliases: h
 
 ## libraries
-List loaded dynamic libraries
+List loaded dynamic libraries.
+	
+	libraries [-d N]
+
+If used with the -d option it will re-attempt to download the debug symbols for library N, using debuginfod-find.
 
 
 ## list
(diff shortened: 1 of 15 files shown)


derekparker (approved):
LGTM

PR #4125, opened 2025-09-01
pkg/proc: parallelize more tests

derekparker:
In an effort to speed up the full test suite this PR updates more tests in `pkg/proc/proc_test.go` to run in parallel. It does this by using an external Delve process and using the RPC2 client to connect to it and issue commands, making it more of an integration-style test. This allows us to run these tests in parallel where we previously could not because the test binary was acting as the debugger and receiving signals, etc... This won't hurt our ability to debug these tests as we can still use follow-fork mode to debug Delve. Eventually we can move these integration tests to a different package that doesn't import any internal package making it easier to set delayed breakpoints.

I've started with just a handful of tests as I know it's going to be arduous to review these changes and ensure that no test logic gets changed as this should be a pure refactor.


aarzilli:
This is too anti-ergonomic:

* it's adding too many layers of code into some pretty fundamental tests
* most panics will be in the debugger process rather than the test process meaning they won't even appear in the test logs
* we can use follow-exec to attach to the real tested delve but we can't easily step from the test code into the tested code

It makes the tests too useless.

aarzilli:
You know, we could probably get about the same effect as this, in terms of speed up, by dividing all the proc tests into 8 groups (say by hashing the test name and taking the last 3 bits) and using make.go to run each group into its own process. The problem its going to be keeping the outputs distinct.

derekparker:
I agree about this being less ergonomic. Since this patch set is still in draft mode I'm happy to iterate on that aspect. I do think there are a lot of improvements we can get by running the tests this way. I think that all that you mentioned above can be handled in a more ergonomic way. I'm really hesitant to do anything outside of `go test` for running tests, including creating our own custom test runner.

Also, I know this is touching the most fundamental of our tests, so I want to do it carefully, in batches, so it's easier to review and catch any possible unintended changes in the test behavior. However, I also honestly believe a lot of these tests are better ran at the integration test level. Yes, we introduce a lot of code inbetween that we didn't have before (service layer on up, essentially), but I think all that service code is worthy of being tested as well. You could make the argument our testing isn't as comprehensive as it could be as we're testing at a lower level than how a real user would interact with Delve. I think we should still have lower level tests in pkg/proc that are true unit tests. For everything else we should move tests to a higher level integration test.

Finally, we kind of have similar tests spread out all over the place. We have tests in `service/test` that test similar things in `pkg/proc` (e.g. we have tests which cover different situational uses of `next` in both locations. Why?). I'd argue this hurts test discoverability and doesn't give great intuition for where a particular test should live (low level `pkg/proc`? higher level `service/test`?). The answer to that question should be obvious, but right now I don't think it is.

aarzilli:
> I'm really hesitant to do anything outside of `go test` for running tests, including creating our own custom test runner.

I'm not suggesting using something other than `go test`. We could write something like this:

```
var testGroup int

func TestMain(m *testing.M) {
	flags.IntVar(&testGroup, "testgroup", "which group of tests to run")
	...
}

func testGroupCheck(t *testing.T) {
	if convertNameToGroup(t.Name()) != testGroup {
		t.Skip("WRONG GROUP")
	}
}
```

Then call testGroupCheck(t) on every test. `make.go` would run 8 `go test` in parallel with `-testgroup 0` through 7 and gather the output. If we do this only in `-json` mode it's not too much work, in principle the only thing we would have to do is filter out the "WRONG GROUP" skips, although I would prefer if we grouped outputs from the same test together, so they are easier to read.

> However, I also honestly believe a lot of these tests are better ran at the integration test level

I think we fundamentally disagree on this. Debugging failing tests outside of pkg/proc is always a much bigger bother.

> You could make the argument our testing isn't as comprehensive as it could be as we're testing at a lower level than how a real user would interact with Delve

Code coverage for the service layer is actually really good, you can check it out by running:

	$ go test -cover -coverprofile=cover.out -coverpkg=./... ./...
	$ go tool cover -html=cover.out

The main service layer files are like this:

service/debugger/debugger.go 71% 
service/rpc2/client.go 77.6%
service/rpc2/server.go 72.1%
service/api/conversions.go 89.5%
service/dap/server.go 86.3%

Most of the untested code is related to rr, windows and ebpf (stuff that command wouldn't collect). You can't go much higher than this. Incidentally if we run everything in a subprocess we would also lose all code coverage functionality (dlv/cmds/commands.go only has a 2.7% coverage for example, because it only gets tested through subprocesses).

> Finally, we kind of have similar tests spread out all over the place. We have tests in service/test that test similar things in pkg/proc (e.g. we have tests which cover different situational uses of next in both locations. Why?).  I'd argue this hurts test discoverability and doesn't give great intuition for where a particular test should live (low level pkg/proc? higher level service/test?).

Because the ones in service test the API and the ones in proc test the functionality. For that matter basically all tests in service/dap are duplicates of a test somewhere else, they exist to test the DAP API. I don't think it's hard to decide where a test goes, if it's testing a property of the API it goes in service, otherwise it goes in proc.

derekparker:
> > I'm really hesitant to do anything outside of `go test` for running tests, including creating our own custom test runner.
> 
> I'm not suggesting using something other than `go test`. We could write something like this:
> 
> ```
> var testGroup int
> 
> func TestMain(m *testing.M) {
> 	flags.IntVar(&testGroup, "testgroup", "which group of tests to run")
> 	...
> }
> 
> func testGroupCheck(t *testing.T) {
> 	if convertNameToGroup(t.Name()) != testGroup {
> 		t.Skip("WRONG GROUP")
> 	}
> }
> ```
> 
> Then call testGroupCheck(t) on every test. `make.go` would run 8 `go test` in parallel with `-testgroup 0` through 7 and gather the output. If we do this only in `-json` mode it's not too much work, in principle the only thing we would have to do is filter out the "WRONG GROUP" skips, although I would prefer if we grouped outputs from the same test together, so they are easier to read.
> 
> > However, I also honestly believe a lot of these tests are better ran at the integration test level
> 
> I think we fundamentally disagree on this. Debugging failing tests outside of pkg/proc is always a much bigger bother.

I definitely agree it's a bigger bother, and less ergonomic.

> 
> > You could make the argument our testing isn't as comprehensive as it could be as we're testing at a lower level than how a real user would interact with Delve
> 
> Code coverage for the service layer is actually really good, you can check it out by running:
> 
> ```
> $ go test -cover -coverprofile=cover.out -coverpkg=./... ./...
> $ go tool cover -html=cover.out
> ```
> 
> The main service layer files are like this:
> 
> service/debugger/debugger.go 71% service/rpc2/client.go 77.6% service/rpc2/server.go 72.1% service/api/conversions.go 89.5% service/dap/server.go 86.3%
> 
> Most of the untested code is related to rr, windows and ebpf (stuff that command wouldn't collect). You can't go much higher than this. Incidentally if we run everything in a subprocess we would also lose all code coverage functionality (dlv/cmds/commands.go only has a 2.7% coverage for example, because it only gets tested through subprocesses).
> 

Ok, I somewhat misspoke here, I wasn't suggesting we don't have proper code coverage or overall a totally comprehensive test suite, we've always worked hard to maintain that and require new tests for regressions. I do agree however that the loss of code coverage reports would but a huge net loss with this approach.

> > Finally, we kind of have similar tests spread out all over the place. We have tests in service/test that test similar things in pkg/proc (e.g. we have tests which cover different situational uses of next in both locations. Why?).  I'd argue this hurts test discoverability and doesn't give great intuition for where a particular test should live (low level pkg/proc? higher level service/test?).
> 
> Because the ones in service test the API and the ones in proc test the functionality. For that matter basically all tests in service/dap are duplicates of a test somewhere else, they exist to test the DAP API. I don't think it's hard to decide where a test goes, if it's testing a property of the API it goes in service, otherwise it goes in proc.

Right, I get that, but the ones testing the API are *also* by nature testing the functionality. They are testing the functionality more comprehensively because even if something works in proc but it isn't exposed properly through the higher service layers, that feature is still broken as far as a user is concerned.

Anyways, I think this discussion has generated enough reasons to abandon this particular approach, but I like that it was a jumping off point for a wider discussion and ideas for improving our test execution time (and maybe deduplicating tests, if possible). I'll close this PR, and I would actually be interested in the parallelization approach you mentioned, if you'd like to submit a PR for that.

Issue #4128, opened 2025-09-02
Remove ppc64le clause from support_sentinel_linux.go to avoid conflicting package declaration errors.


tshah14:
While attempting to build Delve on the ppc64le architecture (Debian), the build fails with the following error:

`obj-powerpc64le-linux-gnu/src/[github.com/go-delve/delve/service/debugger/](http://github.com/go-delve/delve/service/debugger/)debugger.go:31:2: found packages native (dump_linux.go) and your_linux_architecture_is_not_supported_by_delve (support_sentinel_linux.go) in /build/reproducible-path/delve-1.24.0/obj-powerpc64le-linux-gnu/src/[github.com/go-delve/delve/pkg/proc/nativedh_golang](http://github.com/go-delve/delve/pkg/proc/nativedh_golang): error: go list -f '{{ range .Deps }}{{.}} {{ end }}' returned exit code 1 make: *** [debian/rules:10: binary] Error 25 dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2`

aarzilli:
Fixed by #4129

PR #4129, opened 2025-09-02
support_sentinel_linux.go: Fixed the ppc64le clause in the file to  prevent build failures

tshah14:
-Delve is now being built for ppc64le, so support_sentinel_linux.go should not be compiled for that architecture.

-This prevents build errors due to conflicting packages when compiling on ppc64le.


tshah14:
This commit fixes https://github.com/go-delve/delve/issues/4128

aarzilli (approved):
LGTM

PR #4130, opened 2025-09-02
terminal: trace /regexp/ should set ret breakpoints correctly

### LINKED ISSUE #4036: Is it a bug here when setting breakpoints at return address of multiple functions
```go
func setBreakpoint(t *Session, ctx callContext, tracepoint bool, argstr string) ([]*api.Breakpoint, error) {
        ...
	switch t := loc.(type) {
	case *locspec.NormalLocationSpec:
		shouldSetReturnBreakpoints = t.LineOffset == -1 && t.FuncBase != nil
	case *locspec.RegexLocationSpec:
		shouldSetReturnBreakpoints = true
	}


	if tracepoint && shouldSetReturnBreakpoints && locs[0].Function != nil {
		for i := range locs {
			if locs[i].Function == nil {
				continue
			}

                        // here, if we use RegexLocationSpec for multiple functions, 
                        // is `locs[0].Function.Name()` a bug here?
			addrs, err := t.client.(*rpc2.RPCClient).FunctionReturnLocations(locs[0].Function.Name())
			if err != nil {
				return nil, err
			}
			for j := range addrs {
				_, err = t.client.CreateBreakpoint(&api.Breakpoint{
					Addr:        addrs[j],
					TraceReturn: true,
					Line:        -1,
					LoadArgs:    &ShortLoadConfig,
				})
				if err != nil {
					return nil, err
				}
			}
		}
	}
	return created, nil
}
```go

aarzilli:
Fix the way the trace command sets return breakpoints when a regexp is passed.

Fixes #4036

### DIFF
--- a/_fixtures/traceret.go
+++ b/_fixtures/traceret.go
@@ -1,0 +1,16 @@
+package main
+
+import "fmt"
+
+func fncall1() int {
+	return 1
+}
+
+func fncall2() int {
+	return 2
+}
+
+func main() {
+	fmt.Println(fncall1())
+	fmt.Println(fncall2())
+}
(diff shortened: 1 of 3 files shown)


aarzilli:
It looks like the darwin/amd64 agent is also failing for no reason now.

PR #4135, opened 2025-09-04
service,proc: fix tests to enable parallel runs

derekparker:
This patch fixes the test suite to enable tests to be ran in parallel test processes via https://github.com/go-delve/paratest.

* Ensure fixtures which bind to a socket use a random port
* Ensure we use `t.Setenv`
* Ensure no tests are modifying fixtures in place that other tests may rely on

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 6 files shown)


pkg/proc/proc_test.go
```diff
@@ -1821,15 +1852,45 @@ func TestCmdLineArgs(t *testing.T) {
 func TestIssue462(t *testing.T) {
 	skipOn(t, "broken", "windows") // Stacktrace of Goroutine 0 fails with an error
 	withTestProcess("testnextnethttp", t, func(p *proc.Target, grp *proc.TargetGroup, fixture protest.Fixture) {
+		pid := p.Pid()
+		portFile := fmt.Sprintf("/tmp/testnextnethttp_port_%d", pid)
+		
+		// Ensure cleanup of port file
+		defer func() {
```
aarzilli, line 1858: The `func() { }()` wrapping doesn't do anything.

pkg/proc/proc_test.go
```diff
@@ -526,21 +527,51 @@ func TestNextConcurrentVariant2(t *testing.T) {
...
 func TestNextNetHTTP(t *testing.T) {
 	testcases := []nextTest{
-		{11, 12},
-		{12, 13},
+		{14, 15},
+		{15, 16},
 	}
 	withTestProcess("testnextnethttp", t, func(p *proc.Target, grp *proc.TargetGroup, fixture protest.Fixture) {
+		pid := p.Pid()
+		portFile := fmt.Sprintf("/tmp/testnextnethttp_port_%d", pid)
```
aarzilli: I think this is a problem for the windows runners since there is not /tmp directory. I think this may be why windows CI is failing with this change.

derekparker:
> The change to `t.Setenv` is just aesthetic/modernization, right? It shouldn't have any effect since the tests are run sequentially in separate threads.

That's right.

derekparker:
I’ll fix the other windows failure that came up as well.

PR #4145, opened 2025-09-10
dlv_test: disable TestUnattendedBreakpoint on FreeBSD

aarzilli:
After the parallellization this test, for some reason, gets stuck
specifically on FreeBSD. Disable it.


derekparker:
Might be better to just remove the call to `t.Parallel` for this test with a TODO to investigate on FreeBSD.

Issue #4146, opened 2025-09-11
Performance: huge memory usage issue caused by trie.Trie.Add()


jayantxie:
<img width="3840" height="1916" alt="Image" src="https://github.com/user-attachments/assets/24a4d44e-06f9-469d-aa39-8eae0858d419" />

I obtained a delve heap flame graph with extremely high memory usage, from which it can be seen that calls to trie.Add() occupy nearly 8GB of memory. I suspect that when the number of function definitions is extremely large, the trie will become a major performance bottleneck.

aarzilli:
I wish we could delete PackageVariables.

jayantxie:
`terminal.(*Term).Run` should also be optimized, but I have no idea for now.

aarzilli:
In what way?

aarzilli:
i.e. what is about Term.Run that is slow?

jayantxie:
There are two feasible approaches: one is to optimize the implementation of trie/v3 by using a radix tree to construct the prefix tree. This avoids creating a tree node for every rune—for example, in the case of `mypackage/internal/private.veryLongFunctionName`, the current implementation would create dozens of nodes. If optimized to a radix tree, nodes would only be split when necessary. However, this optimization requires significant effort.
A simpler workaround is lazy initialization, where the trie tree is only initialized when the fuzzy search functionality for functions is used. The trade-off is that users may experience a brief delay the first time they use code completion, but this is generally acceptable.

jayantxie:
@aarzilli It seems that the memory issue with `func (t *Term) Run()` here still hasn't been resolved. This problem should be quite serious, right? Don't you consider fixing it?

aarzilli:
What is the issue with (*Term).Run?

jayantxie:
@aarzilli As shown in the flame graph, in a large Go program, if the number of functions is large, the cost of `func (t *Term) Run()` to support autocompletion is very high. Therefore, I propose changing it to lazy init, loading the functions trie only when necessary.

derekparker:
@jayantxie can you give me steps to reproduce this? I am working on more patches to optimize trie, and would like to test against this case.

jayantxie:
@derekparker I can't provide you with the binary file reproduced by our internal environment, but I can give you a script that simulates this binary file. Their characteristics are similar, both having a large number of function definitions, which leads to high memory consumption when constructing the trie. 
[...]

PR #4147, opened 2025-09-11
pkg/proc: use map to replace trie for optimizing the large memory usage in rangeBodyFunction lookup

jayantxie:
Fixes https://github.com/go-delve/delve/issues/4146

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -25,7 +25,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/derekparker/trie/v3"
 	"github.com/hashicorp/golang-lru/v2"
 
 	"github.com/go-delve/delve/pkg/astutil"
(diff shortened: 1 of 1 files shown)


aarzilli:
What's the difference in memory usage?

aarzilli:
Also, see TestStaticcheck failure.

jayantxie:
> What's the difference in memory usage?

After changing to map, the overhead of `trie.Add ` can be optimized, and indexing is only built for functions with range bodies

aarzilli:
How about instead we revert everything to the way it was before lookupRangeBodyFunc was introduced and we don't calculate the rangeBodies in `extra()` but instead we add a `rangeBodies()` method to `*Function` that does the calculation, per-function, and caches it.

jayantxie:
If rangeBodies() inadvertently becomes a hot function again in a future update, the same performance degradation issue could recur.

aarzilli:
rangeBodies are only needed when we make stacktraces, presumably they are going to be needed only for a subset of functions (the ones where breakpoints are set).
To be honest I find this whole series of patches odd: the `extra` method only gets called through PackageVariables for functions that are stored into function pointers, so in most cases it shouldn't be a problem either.

jayantxie:
In that case, I agree with your proposal to separate rangeBodies() from extra() then.

derekparker:
> How about instead we revert everything to the way it was before lookupRangeBodyFunc was introduced and we don't calculate the rangeBodies in `extra()` but instead we add a `rangeBodies()` method to `*Function` that does the calculation, per-function, and caches it.

Agreed on this.

jayantxie:
Close this pull request as discussed.

PR #4152, opened 2025-09-12
ci/cd: fix codesigning on mac/arm cloud agent

vietage:
TeamCity build fails on Mac arm cloud agent because codesigning hangs on a password prompt in GUI. This commit changes how we generate a certificate and how we import it into a keychain so that it does not prompt passwords in GUI.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -258,6 +258,7 @@ class TestBuild(val os: String, val arch: String, val version: String, buildId: 
                     name = "Test"
                     path = "_scripts/test_mac.sh"
                     arguments = "${"go$version"} $arch %system.teamcity.build.tempDir%"
+                    param("env.CI", "true")
                 }
             }
         }
--- a/_scripts/gencert.sh
+++ b/_scripts/gencert.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# Check if the certificate is already present in keychains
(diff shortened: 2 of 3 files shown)


aarzilli:
The tests will pass like this because the certificate is only necessary for the macOS native backend, which we don't test (it's been broken for a long time). However if someone ever were to fix it the certificate would need to be specifically installed in the system keychain, so changing  gencert.sh to install the certificate somewhere else is just breaking the backend further.

I think we should just add something like this to checkCert in _scripts/make.go:

```
if os.Getenv("CI") != "" {
        return true
}
```

the rest probably works.

vietage:
Yes, it is way easier to disable codesigning at all if you do not need it.

We still need somehow pass CI=true to the pipeline. I've modified TC config and added CI=true for Mac build, but it is tricky to test because we need these changes in VCS in order for TeamCity to apply them. So I only tested this approach with custom build with manually provided CI=true env variable.

And looks like Linux/Amd64 build is flaky and sometimes fails with"cannot find git" message. I've re-run the build again and it passed.

aarzilli:
> And looks like Linux/Amd64 build is flaky and sometimes fails with"cannot find git" message. I've re-run the build again and it passed.

That's not how it fails. If you download the full log you will see that there is an error from Go: `exec: WaitDelay expired before I/O complete`. The way teamcity organizes the output hides it. I have no idea what the cause is.

aarzilli (approved):
LGTM

PR #4159, opened 2025-09-16
proc: simplify and rename structMember

aarzilli:
With the recent changes in 1e3ccee (#4118) structMember became largely
reduntant with the new findStructMemberOrMethod. This commit simplifies
the code of structMember so that it does not handle embedded structs
and interfaces and changes its name to structField so that it can be
kept as an internal lightweight version of findStructMemberOrMethod.

### DIFF
--- a/pkg/proc/variables.go
+++ b/pkg/proc/variables.go
@@ -281,11 +281,11 @@ func GetG(thread Thread) (*G, error) {
 		// For our purposes it's better if we always return the real goroutine
 		// since the rest of the code assumes the goroutine ID is univocal.
 		// The real 'current goroutine' is stored in g0.m.curg
-		mvar, err := g.variable.structMember("m")
+		mvar, err := g.variable.structField("m")
 		if err != nil {
 			return nil, err
 		}
-		curgvar, err := mvar.structMember("curg")
+		curgvar, err := mvar.structField("curg")
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


pkg/proc/variables.go
```diff
@@ -1168,64 +1168,41 @@ func (v *Variable) structMember(memberName string) (*Variable, error) {
...
-			if first {
-				return nil, fmt.Errorf("%s (type %s) is not a struct", vname, structVar.TypeString())
+		if field.Name == "&"+memberName && closure {
+			f, err := structVar.toField(field)
+			if err != nil {
+				return nil, err
 			}
+			return f.maybeDereference(), nil
+		}
+		if !field.Embedded {
```
derekparker: Shouldn't this be removed as this was supposed to be updated to not support embeds?

pkg/proc/variables.go
```diff
@@ -1168,64 +1168,41 @@ func (v *Variable) structMember(memberName string) (*Variable, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		// Check for embedded field referenced by type name
+		parts := strings.Split(field.Name, ".")
+		if len(parts) > 1 && parts[1] == memberName {
+			return embeddedVar, nil
 		}
-		first = false
+		embeddedVar.Name = structVar.Name
```
derekparker: We drop the reference to this variable after this, so why update `Name`? I think this is leftover from when it was added to `queue`:

```
embeddedVar.Name = structVar.Name
queue = append(queue, embeddedVar)
```

derekparker (changes requested):

pkg/proc/variables.go
aarzilli: Done.

pkg/proc/variables.go
aarzilli: Done.

PR #4161, opened 2025-09-17
proc: produce materialized breakpoint event for follow-exec mode

aarzilli:
When a previously suspended breakpoint is un-suspended we should
generate a BreakpointMaterialized event, we already did this when
breakpoints were materialized by opening plugins but we forgot to do it
when targets are added by follow-exec mode.


derekparker (approved):
LGTM

PR #4163, opened 2025-09-19
tests: add missing test cleanup

aarzilli:
* Changes BuildFixture so that fixtures are not built multiple times
* Add some missing calls to AddPathToRemove


derekparker (approved):
LGTM

PR #4164, opened 2025-09-20
cmd/dlv: prevent trace killing attached process

alex-emery:
When using `dlv trace --pid`, the process should continue running after detach.

This change only kills the process when dlv started it, not when attaching to an existing process.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -17,6 +17,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


derekparker (changes requested):
Sorry, test is failing on windows.

cmd/dlv/dlv_test.go
```diff
@@ -1071,6 +1072,11 @@ func TestTracePid(t *testing.T) {
 	if !bytes.Contains(output, expected) {
 		t.Fatalf("expected:\n%s\ngot:\n%s", string(expected), string(output))
 	}
+
+	err = targetCmd.Process.Signal(syscall.Signal(0))
```
aarzilli: This doesn't work on windows. I'd call targetCmd.Process.Wait on a separate goroutine and close a channel when it returns, here I'd check that the channel doesn't get closed for, say 500ms (a select between reading on the channel and reading on time.After).

Doing the check immediately races with the operating system: the process struct could still be around for a while after the process has died and if we wait a while and do FindProcess again the pid could have been reused for something else.

alex-emery: I tried this within this test, but I ended up with a racy test since the underlying fixture is rather short lived. I've created a new test which attaches to loopprog and immediately exits instead. Having said that, it's still failing in CI but I'm investigating.

aarzilli: AFAICT, the only CI failure is related to the linux/loong64 builder which we are trying to add (so far unsuccessfully).

derekparker (approved):
LGTM

PR #4169, opened 2025-10-01
service/dap: use exception breakpoints for predefined breakpoints

### LINKED ISSUE #4038: dap: Use exception breakpoints to handle panic/fatal error breakpoints
In https://github.com/go-delve/delve/pull/4027 it was mentioned that panic/fatal error breakpoints could be represented by adding them via `breakpoint` events (of type `added`); I'd like to suggest using exception breakpoints for these instead:
- Exception breakpoints can be advertised in adapter's capabilities, which means that the client's ui can display them nicely/distinguish them from other breakpoints. That's what Zed does for example.
- https://github.com/go-delve/delve/pull/4027 could basically be reverted, because Delve could assign valid breakpoint IDs to these breakpoints.
- It is suggested in OP that the client can disable these breakpoints right now (and that they're not anyhow special), but as is, the client has to special-case Delve to do that; otherwise it cannot know that breakpoint ID's -1 and -2 have any meaning (and what meaning they have).
- https://github.com/go-delve/delve/issues/3653 could be implemented with it (assuming that the user is interfacing with Delve via DAP); we could keep these on by default, but if the debug adapter client persisted exception breakpoint options across sessions (which is what Visual Studio Code does), it'd make it possible to disable unwanted breakpoints.

aarzilli:
Use DAP's exception breakpoints to represent the predefined breakpoints
we set for unrecovered panics and fatal throws, so that DAP users can
disable those breakpoints.

Fixes #4038


service/dap/server.go
```diff
@@ -1707,9 +1711,24 @@ func (s *Session) getMatchingBreakpoints(prefix string) map[string]*api.Breakpoi
...
+		enabled[filter] = true
+	}
+	for _, bp := range s.debugger.Breakpoints(false) {
+		if bp.ID < 0 && enabled[bp.Name] != !bp.Disabled {
+			bp.Disabled = !enabled[bp.Name]
+			s.debugger.AmendBreakpoint(bp)
+		}
+	}
+	resp := &dap.SetExceptionBreakpointsResponse{Response: *newResponse(request.Request)}
+	for _, bp := range s.debugger.Breakpoints(false) {
```
derekparker, line 1725: I think this can be folded into the loop above right? Move the creation of `SetExceptionBreakpointsResponse` above the loop, then after amend add to the `resp.Body.Breakpoints` and update the response.

derekparker (approved):
LGTM

service/dap/server.go
aarzilli: It doesn't work (that's how I did it at first) because AmendBreakpoint doesn't update the list of addresses in its argument. I did it this way because to make it work with a single loop we'd have to add a loop over all breakpoints to amendBreakpoint to collect addresses, like convertBreakpoint does, work that in most cases will be discarded (because AmendBreakpoint doesn't send the breakpoint back).

PR #4170, opened 2025-10-01
pkg/terminal: allow custom starlark commands to be used in on prefix

### LINKED ISSUE #4160: Support running starlark commands when breakpoint is hit?
My actual problem is that I want to break on panics, but I want to ignore the panic/recover flow control in the fmt package.

I have a starlark script to create a command cif that continues if we're under the fmt package. Unfortunately, I still have to invoke it manually at every breakpoint. I'd like to be able to invoke this command whenever the breakpoint is hit.

```
def command_continue_if_fmt(args):
    """Continues execution if the current goroutine's backtrace contains the fmt package.
    
    Usage: continue_if_fmt
    
    This command checks if any frame in the current goroutine's stack trace
    is from the fmt package. If it finds one, it continues execution.
    Otherwise, it reports that no fmt package was found in the stack.
    """
    
    # Get the current goroutine's stack trace
    # Using None for the first argument gets the current goroutine
    # The second argument (Depth) of -1 gets the full stack
    stack = stacktrace(-1, 10, False, False, None, None)
    
    if not stack or not stack.Locations:
        print("Could not get stack trace")
        return
    
    # Check each frame in the stack trace
    fmt_found = False
    for i, frame in enumerate(stack.Locations):
        # Check if this frame's function is from the fmt package
        if frame.Function and "src/fmt" in frame.File:
            fmt_found = True
            print("Found fmt package in frame {}: {}".format(i, frame.Function.Name_))
            break
    
    if fmt_found:
        print("Continuing execution...")
        dlv_command("continue")
    else:
        print("No fmt package found in current stack trace")
        dlv_command("bt")


def command_cif(args):
    """Alias for continue_if_fmt"""
    command_continue_if_fmt(args)

def main():
    # Create an alias for the command
    dlv_command("config alias continue_if_fmt cif")
    print("Commands 'continue_if_fmt' and 'cif' are now available")
    print("Use 'continue_if_fmt' to continue execution if fmt package is in the backtrace")
```

derekparker:
Fixes #4160


Documentation/cli/README.md
```diff
@@ -564,11 +564,18 @@ Executes a command when a breakpoint is hit.
 
 	on <breakpoint name or id> <command>
 	on <breakpoint name or id> -edit
-	
 
-Supported commands: print, stack, goroutine, trace and cond. 
+
+Supported commands: print, stack, goroutine, trace and cond.
+
+You may also use custom starlark commands with the 'on' prefix by annotating the command with
```
aarzilli: I find this description confusing, when I first read it I interpreted where it says *"annotating the command"* as referring to the `on` command, because that's what we are describing and the real referent (*"custom starlark commands"*) is plural.
But also I don't think this is necessary? Why not allow all starlark defined commands to appear here?

pkg/terminal/command.go
```diff
@@ -2937,6 +2947,15 @@ func printBreakpointInfo(t *Term, th *api.Thread, tracepointOnNewline bool) {
 			printStack(t, t.stdout, bpi.Stacktrace, "\t\t", false)
 		}
 	}
+
```
aarzilli, line 2933: I don't think this is the right place for this, suppose two breakpoints are hit and the first one has extra commands:

* we print the first breakpoint
* we execute the commands associated with it, which may even restart the program
* when we finish that we print the second breakpoint, using a stale `state` object (and the breakpoint might not even be valid anymore)

Also the `frame` and `list` command end up calling `printcontextThread` so they would end up executing the starlark script sometimes.

I think this needs to be somewhere else, in the functions for the execution-resuming commands. Also with regards to scripts that end up resuming execution I'm not sure what the correct behavior is in general (if we hit two breakpoints with scripts and one wants to `continue` and one wants to `next` what should we do? Should the first one win? Should the second script be executed at all? In what context?)

aarzilli:
PS. if we constrain the things that an `on` command can do to things that don't resume execution (and don't change variables) all these problems go away and we can even leave it in printcontextThread, however the motivation for the original issue was specifically about resuming execution. 
Arguably that maybe should be done with starlark in `cond` rather than in `on` but that would involve running starlark server side, which is a bigger refactor.

aarzilli:
> Arguably that maybe should be done with starlark in `cond` rather than in `on` but that would involve running starlark server side, which is a bigger refactor.

PPS. Thinking more about this having a reduced server-side version of the starlark interpreter that can only call a small subset of our API (eval, stacktraces, goroutine lists...) could be useful.

pkg/terminal/command.go
derekparker: Yeah definitely agreed, this was an awkward place for it. Should've updated before pushing.

Documentation/cli/README.md
derekparker: I'm ok with making all custom commands "on-able".

derekparker:
> Arguably that maybe should be done with starlark in `cond` rather than in `on` but that would involve running starlark server side, which is a bigger refactor.

Yeah, I had this thought originally too. This was simpler, but also I think it's a useful feature outside the original request.

derekparker:
Marking as draft as there is an edge case I thought of that I'd like to fix before another review.

derekparker:
Alright, updated to handle more situations explicitly.

The flow now is:

* Program runs, hits a breakpoint w/ 3 custom commands.
* The 2nd of the 3 commands causes the program to continue execution.
* The 3rd custom command is canceled (does not run)
* Any breakpoint hit during continued execution will have its custom commands (if any) execute, following the same rule as above.

The code handles multiple layers of this invalidation logic by maintaining a stack of booleans on the `Term` struct any time we execute custom commands.

One caveat that still remains is `raw_command` from within a starlark command will not follow these rules. This seems ok to me because `raw_command` comes with its own set of caveats, and if someone is using that chances are they know what they're doing and why they are using that over `dlv_command`. Either way, I'm happy to be more explicit there in the docs or have *any* `raw_command` invalidate the custom commands. We could also throw an error or print a warning if `raw_command` is used in a command with the `on` prefix.

Documentation/cli/README.md
aarzilli: Did you change your mind about this or did you forget to remove the on_prefix?

derekparker: I'm indifferent, I suppose, but my thought process for the sake of discussion:

We already restrict the set of commands that can be used with the `on` prefix. Making commands opt-in seems like we're continuing to be explicit about which commands we want (or the user wants) to be used with that prefix. On the other hand... 🤷‍♂️ I really don't have strong feelings. Custom commands, and adding them to a breakpoint via `on` is also at the users discretion.

If you're ok with making all custom commands on-able then I will update this PR to remove the `@on_prefix` and just add the prefix to all custom commands.

aarzilli: IIRC we restricted the `on` prefix to some commands because most of them couldn't be implemented at all. IMO the `@on_prefix` thing is just making the users check a box twice.

derekparker: Done.

PR #4173, opened 2025-10-04
pkg/proc: split off rangeParent from function extras

### LINKED ISSUE #4146: Performance: huge memory usage issue caused by trie.Trie.Add()

aarzilli:
The rangeParent calculation was changed 0b74953f0 to address slowness
(due to repeated calculations) when calling PackageVariables. That
changed caused to program to use a lot of memory due to having to
calculate the range bodies of every function. This change reverts
0b74953f0 and splits off rangeBodies calculation from the extras method
of Function, so that it isn't called from PackageVariables.

Fixes #4146

### DIFF
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -149,7 +149,7 @@ func dwarfToRuntimeType(bi *BinaryInfo, mem MemoryReadWriter, typ godwarf.Type) 
 		return 0, false, false, nil
 	}
 
-	mds, err := bi.getModuleData(mem)
+	mds, err := LoadModuleData(bi, mem)
 	if err != nil {
 		return 0, false, false, err
 	}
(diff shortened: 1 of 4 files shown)


derekparker (approved):
LGTM

pkg/proc/types.go
```diff
@@ -149,7 +149,7 @@ func dwarfToRuntimeType(bi *BinaryInfo, mem MemoryReadWriter, typ godwarf.Type)
 		return 0, false, false, nil
 	}
 
-	mds, err := bi.getModuleData(mem)
+	mds, err := LoadModuleData(bi, mem)
```
jayantxie, line 152: Why is this change reverted too?

PR #4177, opened 2025-10-09
pkg/terminal: lazy init functions trie to reduce unnecessary memory usage

jayantxie:
Fixes https://github.com/go-delve/delve/issues/4146

### DIFF
--- a/pkg/proc/types.go
+++ b/pkg/proc/types.go
@@ -149,7 +149,7 @@ func dwarfToRuntimeType(bi *BinaryInfo, mem MemoryReadWriter, typ godwarf.Type) 
 		return 0, false, false, nil
 	}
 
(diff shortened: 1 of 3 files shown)


derekparker:
I'm going to merge this now, as I'm not sure how often users take advantage of the tab completion feature. However, I use it nearly every single debug session, and I'm somewhat hesitant to introduce any noticeable lag with the first tab completion, as I do think that could degrade the user experience. I did test against the example generated file you included, and the lag wasn't that bad, but definitely noticeable. If this becomes something I notice regularly in my debug sessions I will move to revert this, focusing instead on what is causing the memory utilization (note that the memory utilization will happen anyways if you even accidentally hit tab during a debug session...).

PR #4181, opened 2025-10-12
proc: make PushPackageVarOrSelect check local variables first

### LINKED ISSUE #4179: Wrong anonymous method shown when local var name matches package name
If a local variable shares the same name as a package, and that variable’s field points to a struct from that package, dlv p ends up showing an incorrect anonymous method generated by the Go compiler.

Looks like the name resolver inside Delve gets confused between the package and the variable scope.

Minimal reproducible example:

`config/test.go`

```go
package config

type Test struct {
	Name interface{}
	Age  int
}
```

`main.go`

```go
package main

import (
	"log"
	"wenxuan70/config"
)

type Config struct {
	*config.Test
}

func f() *config.Test {
	return &config.Test{}
}

func main() {
	config := &Config{}

	config.Test = f()

	log.Printf("%+#v", config)
}

```

Image

The problem comes from this line: 
https://github.com/go-delve/delve/blob/d5d75768ffc63353f1dd5e0d40d4137614832d86/pkg/proc/eval.go#L825

We could probably skip methods prefixed with type:.eq. — they’re compiler-generated and not meant to be visible to users anyway.
Here’s a small patch I tested locally that fixes it.

```patch
diff --git a/pkg/proc/eval.go b/pkg/proc/eval.go
index 609022c0..cb0e91a7 100644
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -822,7 +822,7 @@ func (scope *EvalScope) findGlobalInternal(name string) (*Variable, error) {
 		}
 	}
 	for _, fn := range scope.BinInfo.Functions {
-		if fn.Name == name || strings.HasSuffix(fn.Name, "/"+name) {
+		if fn.Name == name || (strings.HasSuffix(fn.Name, "/"+name) && !strings.HasPrefix(fn.Name, "type:.eq.")) {
 			//TODO(aarzilli): convert function entry into a function type?
 			r := newVariable(fn.Name, fn.Entry, &godwarf.FuncType{}, scope.BinInfo, scope.Mem)
 			r.Value = constant.MakeString(fn.Name)
```

aarzilli:
Fixes #4179

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -2539,7 +2539,7 @@ func (bi *BinaryInfo) findTypeExpr(expr ast.Expr) (godwarf.Type, error) {
 		}
 		return bi.findType(typn)
 	}
-	bi.expandPackagesInType(expr)
+	expr = bi.expandPackagesInType(expr)
 	if snode, ok := expr.(*ast.StarExpr); ok {
 		// Pointer types only appear in the dwarf information when
 		// a pointer to the type is used in the target program, here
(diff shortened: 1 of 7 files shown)


aarzilli:
To be honest I'm not sure it's worth doing this. It matches better the semantics of Go but It brings in some complication and it's mostly about collisions between local variables and global variables (which are rare anyway).

wenxuan70:
This PR seems to address another issue I discovered the other day. When the local structure variable and the package name in dwarf are the same, when calling the method of the local structure variable, it will be displayed that the method cannot be found. 🤣
I'll test it tonight when I have some free time to see if I can solve another problem.

wenxuan70:
I tried debugging the unit tests using the Delve command fixed by this patch and found other issues.
When I try to call `ctx.Value("ok")`, I receive an error message: `could not find symbol`. This seems to be because the local variable name conflicts with another package of the same name.
I took a cursory look at the dlv source code and its associated fncall logs. When compiling to an ast expression, ctx is replaced by another name.

example:
image

fncall logs:
image

aarzilli:
What's code.byted.org/gin/ginex/ctx.Value?

wenxuan70:
the `code.byted.org/gin/ginex/ctx` is the package name. No `code.byted.org/gin/ginex/ctx.Value`.

wenxuan70:
I've determined that this code is causing the problem. But I'm not sure if this scenario makes sense.

https://github.com/go-delve/delve/blob/b271ba362b43fa0cdaff0899cf076393a436c0d9/pkg/proc/bininfo.go#L3002-L3009

aarzilli:
Should be fixed.

wenxuan70:
I used your patch and it worked fine locally.

derekparker:
> To be honest I'm not sure it's worth doing this. It matches better the semantics of Go but It brings in some complication and it's mostly about collisions between local variables and global variables (which are rare anyway).

I've been holding off on review due to this concern. Do you still feel this way?

aarzilli:
> I've been holding off on review due to this concern. Do you still feel this way?

I'm inclined to merge this.

pkg/proc/bininfo.go
```diff
@@ -2976,44 +2976,71 @@ func (bi *BinaryInfo) loadDebugInfoMapsInlinedCalls(ctxt *loadDebugInfoMapsConte
...
+				results.List[i] = &field
+				field.Type = bi.expandPackagesInType(e.Results.List[i].Type)
 			}
 		}
+		return &r
 	case *ast.MapType:
-		bi.expandPackagesInType(e.Key)
-		bi.expandPackagesInType(e.Value)
+		r := *e
+		e.Key = bi.expandPackagesInType(e.Key)
```
derekparker: We're mutating `e` instead of `r` here.

pkg/proc/bininfo.go
```diff
@@ -2976,44 +2976,71 @@ func (bi *BinaryInfo) loadDebugInfoMapsInlinedCalls(ctxt *loadDebugInfoMapsConte
...
+		params := *(r.Params)
+		params.List = make([]*ast.Field, len(params.List))
+		r.Params = &params
 		for i := range e.Params.List {
-			bi.expandPackagesInType(e.Params.List[i].Type)
+			field := *(e.Params.List[i])
+			params.List[i] = &field
+			field.Type = bi.expandPackagesInType(e.Params.List[i].Type)
 		}
+		results := *(r.Results)
```
derekparker: This will panic if `r.Results` is nil we should move the `if e.Results != nil` branch above this.

derekparker (changes requested):

derekparker (approved):
LGTM

PR #4184, opened 2025-10-18
replay subcommand must keep trace directory.

lwintermelon:
If user record the process by `rr record -o trace.dir program` manually then replay it with `dlv replay trace.dir`, we shouldn't remove the record trace directory.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4191, opened 2025-10-29
pkg/proc: add way to disable stop-on-error for breakpoint conditions

aarzilli:
Add a fake delve.catch function that can be used to wrap the expression
of breakpoint conditions so that they do not stop if the expression can
not be evaluated correctly.

The main use case for this is that it's hard to set breakpoint
conditions on interface variables that assume different types but it
can also be useful in other cases (for example checking the value of a
pointer-to-struct that sometimes is nil).

### DIFF
--- a/_fixtures/delvecatch.go
+++ b/_fixtures/delvecatch.go
@@ -1,0 +1,9 @@
+package main
+
+import "fmt"
+
+func main() {
+	for i, iface := range []any{12, "test", nil, 2.2, "hello", 7} {
+		fmt.Println(i, iface)
+	}
+}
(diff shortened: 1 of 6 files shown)


derekparker:
This is an interesting feature and I definitely see the use case. I'm wondering if a pseudo function is the right way to go here. Can we annotate the entire condition as being "no stop on error"? That does give the user less granularity over which parts of the expression they want to ignore errors for though. Also, `catch` implies `try/catch` semantics with control flow, which this doesn't have.

aarzilli:
> This is an interesting feature and I definitely see the use case. I'm wondering if a pseudo function is the right way to go here. Can we annotate the entire condition as being "no stop on error"?

The delve.catch function here is only allowed if it surrounds the entire expression. Were you thinking of something like a flag (that would then translate into a field of `api.Breakpoint`)? The downside of that is that every frontend will have to do work to support it, whereas if it's part of the expression it seamlessly supported by everything.

PR #4195, opened 2025-11-04
proc: fix defer loading in go 1.26

aarzilli:
Go 1.26 removed the 'pc' field from the _defer struct. Use the defer
wrapper function, when available to discover the line where the defer
was created. For every other case leave the field empty without
crashing.

Also disables capslock check on everything except the latest minor
version of Go (because the output can change)

### DIFF
--- a/_scripts/rtype-out.txt
+++ b/_scripts/rtype-out.txt
@@ -4,7 +4,6 @@ var debug anytype
 
 type _defer struct {
 	fn anytype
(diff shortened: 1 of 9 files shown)


derekparker:
That patch got reverted upstream, holding off until it lands again.

aarzilli:
The patch hasn't been unreverted before the release of 1.26.

PR #4196, opened 2025-11-05
Replace github.com/hashicorp/golang-lru with a custom lru cache

qmuntal:
(Feel free to disregard this PR is you don't like what's in in)

Delve uses a LRU cache to speed up some workflows. It just makes a basic use of it: adding and getting entries. Implementing that subset using the Go standard library is straightforward, it can be done in 70 lines of code. IMO dropping an external dependency is worth the small maintenance overhead of implementing a custom basic LRU cache.

### DIFF
--- a/pkg/internal/lru/lru.go
+++ b/pkg/internal/lru/lru.go
@@ -1,26 +1,26 @@
-package proc
+package lru
 
 import (
 	"container/list"
 	"sync"
 )
 
-// lruCache is a simple LRU (Least Recently Used) cache implementation.
-type lruCache[K comparable, V any] struct {
+// Cache is a simple LRU (Least Recently Used) cache implementation.
+type Cache[K comparable, V any] struct {
 	mu        sync.Mutex
 	capacity  int
 	items     map[K]*list.Element
 	evictList *list.List
 }
 
-type lruEntry[K comparable, V any] struct {
+type entry[K comparable, V any] struct {
 	key   K
 	value V
 }
 
-// newLRUCache creates a new LRU cache with the given capacity.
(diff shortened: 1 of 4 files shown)


pkg/proc/lrucache.go
```diff
@@ -0,0 +1,71 @@
+package proc
+
+import (
+	"container/list"
+	"sync"
+)
+
+// lruCache is a simple LRU (Least Recently Used) cache implementation.
+type lruCache[K comparable, V any] struct {
+	mu        sync.RWMutex
```
aarzilli: The one we use is actually simplelru which isn't thread safe.

aarzilli:
If we decide to switch from hashicorp/lru to this (I have no opinion in merit, I leave it to Derek) I think we should have some tests.

pkg/proc/lrucache.go
qmuntal: You mean that we don't need to be thread safe? `github.com/hashicorp/golang-lru.Cache`, the one used before this PR, was thread-safe.

qmuntal:
> If we decide to switch from hashicorp/lru to this (I have no opinion in merit, I leave it to Derek) I think we should have some tests.

Added some basic tests

pkg/proc/lrucache.go
aarzilli: Oh, you're right 8e2d698c544aa44eb17f2b29a8bd11351f6e2dfd was wrong, we used to use simplelru.

derekparker:
@qmuntal hey just wanted to ping if you have time to address the above. Thanks!

qmuntal:
> go run _scripts/gen-capslock-all.go

Wops, forgot about this PR. Addressing your comments right now.

PR #4209, opened 2025-12-01
github: fix issue template

aarzilli:
It looks like github stopped showing our issue template, probably
because we are using the legacy format (this may be a github bug or
not, the documentation isn't clear). Convert it to the new format so it
is shown again.


aarzilli:
To see what it looks like I have pushed it to master on [my private fork](https://github.com/aarzilli/delve/issues), it can be tested by clicking "New Issue" there.

PR #4211, opened 2025-12-06
More fixes for Go 1.26

aarzilli:
### proc: fix loading goroutine labels

The format used to store goroutine labels changed again. Adapt to the change.

### TeamCity: fix download url in test_mac.sh

The direct link to storage.googleapis.com no longer works, use the one
that's on the website.


aarzilli:
Removed the load module data change since it's going to be fixed in Go.

PR #4212, opened 2025-12-06
service/test: disable TestGuessSubstitutePath on riscv64

aarzilli:
Disable this test in CI for linux/riscb64 because it is not supported
by the builder.


derekparker:
What's the error or underlying issue here?

aarzilli:
`error obtaining VCS status`, it's probably something about how the repository is being cloned on the agent or maybe there's no git command. I haven't looked into it, the test isn't architecture specific so we don't need it to run there.

PR #4215, opened 2025-12-07
service/dap: fix restart handling when compilation fails

### LINKED ISSUE #4213: Delve loses it's mind if it fails to recompile when restarting

Version: 1.25.2, Build: $Id: af3df277866d7175e816eab14e56611053d4cdea $

go version go1.25.3 X:nodwarf5 linux/amd64

Intel x86_64, Linux 6.12.41, Gentoo, KDE

### Description

1. Create a go.mod and main.go with a main function.
2. Put something in the main function (e.g. `println("Hello world")`).
3. Put a breakpoint in the main function.
4. Start debugging - I'm using vscode-go's default, I just hit F5 without configuring anything.
5. Once the breakpoint is hit, introduce a compilation error - I removed the trailing `}` of the main function.
6. Restart the session - I did this via vscode's restart debug session button.
7. Delve is now borked.

Image

Image

Those errors are expected, but now none of the actions (step/continue/restart/stop) have an effect. Restart *sometimes* does something. If I restart the session a second time, sometimes it restarts (and fails to start) but I think that's because vscode is falling back to the old restart-by-stopping-and-starting method. But sometimes it stays borked. I think it is perfectly reasonable for the step and continue operations to not work, but the fact that I can't stop the session is problematic. **Delve should either terminate the session, or allow me to manually terminate the session** (via the button/DAP request).

aarzilli:
* always send the restart response message, it's just an acknowledgment
and if it isn't sent VSCode will wedge itself (even if an error is sent
instead).

* if Restart errors after detaching from the target process also send a
terminated event.

Fixes #4213

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -3247,6 +3247,9 @@ func (s *Session) onRestartRequest(request *dap.RestartRequest) {
 	s.changeStateMu.Lock()
 	defer s.changeStateMu.Unlock()
 
+	// The response is just an acknowledgement, if we don't send it VSCode gets wedged
+	s.send(&dap.RestartResponse{Response: *newResponse(request.Request)})
+
 	// Cannot restart in noDebug mode
 	if s.isNoDebug() {
 		s.sendErrorResponse(request.Request, FailedToLaunch, "Cannot restart", "cannot restart in noDebug mode")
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -3307,16 +3310,27 @@ func (s *Session) onRestartRequest(request *dap.RestartRequest) {
 		}
 	}
 
+	validBefore := true
+	if _, err := s.debugger.State(false); err != nil {
+		validBefore = false
+	}
+
+	s.config.log.Debug("valid before (blah): ", validBefore)
```
derekparker: Leftover?

aarzilli: oops

PR #4220, opened 2025-12-17
v1.26.0

derekparker:
Release v1.26.0:

* Update min/max supported version
* Update test matrix
* Update capslock file so that it always runs the latest version of capslock via `go run`


aarzilli (approved):
LGTM

PR #4225, opened 2025-12-28
tests: disable TestDebugStripped on go1.26

aarzilli:
The code can not find the module data.


derekparker:
Not merging for now, sent in actual fix: https://github.com/go-delve/delve/pull/4228.

PR #4226, opened 2025-12-28
lru: remove mutex from lru cache

aarzilli:
We only use the lru cache in contexts that are non-thread reentrant,
regardless. The mutex is superfluous (it was originally introduced by a
dependency update, originally the code used simplelru which did not
have a mutex).


derekparker:
There are tests which are now failing which test concurrent access to the cache. These will need to be removed.

PR #4230, opened 2026-01-12
service/dap: Add hit conditional breakpoint capability

DrSergei:
We already have support for this kind of breakpoints but corresponding capability was missed.

You can find existed tests [here](https://github.com/go-delve/delve/blob/master/service/dap/server_test.go#L3999).


aarzilli (approved):
LGTM

derekparker (approved):
LGTM, thanks!

PR #4232, opened 2026-01-14
*: downgrade riscv64

derekparker:
The riscv64 builder has been broken since it was enabled. While we wait for a stable builder and assurances that riscv64 is passing, disable the CI target and require the `exp.linuxriscv64` tag.


aarzilli (approved):
LGTM

PR #4234, opened 2026-01-15
goversion: Fix ProducerAfterOrEqual comparison for devel builds

typesanitizer:
The VersionAfterOrEqual function uses the versionedDevel constant
whereas ProducerAfterOrEqual used 0. This can cause inconsistency
when comparing the "same" version string in different situations.

One of the tests: TestRangeOverFuncNext/TestGotoA1 passes for
gotip due to two wrongs making a right. In particular,
- The version string for checking the stepping sequence is
  derived from `runtime.Version()` instead of
  `p.BinInfo().Producer()`. However, the debuginfo
  depends on the toolchain used to compile the binary,
  not the toolchain used to compile Delve itself.
- When `p.BinInfo().Producer()` is used, the comparison
  function used for it should be `ProducerAfterOrEqual()`
  and not `VersionAfterOrEqual()`.

So we also fix the check in TestGotoA1 to use p.BinInfo().Producer()
and ProducerAfterOrEqual().

### DIFF
--- a/pkg/proc/stepping_test.go
+++ b/pkg/proc/stepping_test.go
@@ -1122,7 +1122,7 @@ func TestRangeOverFuncNext(t *testing.T) {
 			})
 		})
 
-		if goversion.ProducerAfterOrEqual(p.BinInfo().Producer(), 1, 26) && runtime.GOARCH == "arm64" {
+		if goversion.VersionAfterOrEqual(runtime.Version(), 1, 26) && runtime.GOARCH == "arm64" {
 			t.Run("TestGotoA1", func(t *testing.T) {
 				testseq2intl(t, fixture, grp, p, nil, []seqTest{
 					funcBreak(t, "main.TestGotoA1"),


pkg/proc/stepping_test.go
```diff
@@ -1122,7 +1122,7 @@ func TestRangeOverFuncNext(t *testing.T) {
 			})
 		})
 
-		if goversion.VersionAfterOrEqual(runtime.Version(), 1, 26) && runtime.GOARCH == "arm64" {
+		if goversion.ProducerAfterOrEqual(p.BinInfo().Producer(), 1, 26) && runtime.GOARCH == "arm64" {
```
aarzilli: This makes no difference, the version of the test harness and the version of the executable is always the same. We use `runtime.Version` everywhere for tests.

typesanitizer: OK. Should I close the PR or just keep the change for the ProducerAfterOrEqual function?

aarzilli: The change to ProducerAfterOrEqual looks good to me.

typesanitizer: Undid the change to this line.

https://github.com/go-delve/delve/pull/4234/commits/7072c85cdf48a3fba3aac71da02d281456fcea7f

Issue #3910, opened 2025-01-28
Option to show nested variable binding types without full module path?


anderseknert:
Hi there, and thanks for providing this tool!

Using Delve DAP via VS Code, I'm working on some deeply nested data structures. While the first level of bindings have their type information displayed as `package.Type`, any level of nesting below shows the full module path prepended. This makes debugging this code quite annoying as I constantly have to widen the view and click to expand items that would otherwise have been visible directly.

Is there a setting somewhere to use the `package.Type` format for display of bindings at all levels of nesting? Or is this a feature that could be considered?

<img width="831" alt="Image" src="https://github.com/user-attachments/assets/70642a33-7edc-4da4-a202-a601c04a3ace" />

jurisevo:
I absolutely agree, this is very frustrating experience. This is one of those little things which will improve UX for real life projects.

igadmg:
I second that. Also automatic dereference of pointers would be nice. Such a debug productivity killer. I recently complained about that on vscode-go page https://github.com/golang/vscode-go/issues/3969
I'd prefer to remove type information completely form inspector view, or make it some nested fake property

aarzilli:
Fixed by #4235

PR #4235, opened 2026-01-15
Propagate short type option when printing structs.

igadmg:
Fix for issue https://github.com/go-delve/delve/issues/3910
It will propagate same prettyFlags keeping shortTypes passed to writeToStruct

### DIFF
--- a/service/api/prettyprint.go
+++ b/service/api/prettyprint.go
@@ -313,7 +313,7 @@ func (v *Variable) writeStructTo(buf io.Writer, flags prettyFlags, indent, fmtst
 			fmt.Fprintf(buf, "\n%s%s", indent, indentString)
 		}
 		fmt.Fprintf(buf, "%s: ", v.Children[i].Name)
-		v.Children[i].writeTo(buf, flags.set(prettyNewlines, nl), indent+indentString, fmtstr)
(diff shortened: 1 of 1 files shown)


igadmg:
Also updated pretty print for interfaces, slices and maps.

aarzilli:
See failing tests.

igadmg:
@aarzilli all fixed )

service/api/prettyprint.go
```diff
@@ -357,7 +357,7 @@ func (v *Variable) writeMapTo(buf io.Writer, flags prettyFlags, indent, fmtstr s
 
 		key.writeTo(buf, 0, indent+indentString, fmtstr)
 		fmt.Fprint(buf, ": ")
-		value.writeTo(buf, prettyFlags(0).set(prettyNewlines, nl), indent+indentString, fmtstr)
+		value.writeTo(buf, prettyFlags(0).set(prettyShortenType, flags.shortenType()).set(prettyNewlines, nl), indent+indentString, fmtstr)
```
aarzilli, line 360: Doesn't this mean that types will always be printed shortened, no matter what the original flags said?

igadmg: I believe that way it will copy shortenType value from original flags, so it will depend on caller.

aarzilli (approved):
LGTM

PR #4241, opened 2026-01-20
proc,service: detect and warn about trimpath

aarzilli:
Detect if the binary we are debugging was built using trimpath,
propagate it through the API and warn about it in DAP.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"debug/buildinfo"
 	"debug/dwarf"
 	"debug/elf"
 	"debug/macho"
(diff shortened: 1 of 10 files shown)


pkg/dwarf/line/line_parser.go
```diff
@@ -336,3 +340,15 @@ func parseFileEntries5(info *DebugLineInfo, buf *bytes.Buffer) bool {
 	}
 	return true
 }
+
+func detectTrimpath(dirs []string) bool {
```
derekparker: Can't we check `.go.buildinfo` for the `-trimpath` flag and then fall back to this manual checking?

aarzilli: I didn't know it was there. What if I just delete this code and we just use the flag from .go.buildinfo?

derekparker: I think that should be ok, as long as that information doesn't get removed when stripped.

aarzilli: If it's stripped we won't have debug_line either.

derekparker: Good point.

derekparker (approved):
LGTM

PR #4244, opened 2026-01-23
service/api: Update waitReason switch for Go 1.27

typesanitizer:
There are no new waitReason values in Go 1.27,
so we can reuse the existing slice.

This fixes the test failure for
`TestClientServer_chanGoroutines` when running
against gotip.


derekparker:
Approved, just re-running CI.

PR #4248, opened 2026-01-28
Documentation: document AI policy / add agent files

derekparker:
AI coding agents are here and seemingly here to stay. They have become pretty quickly ingrained in a lot of developer workflows. To that end, I would like to provide some guidance on AI usage within this project, and provide an initial AGENTS.md/CLAUDE.md file so that others using AI agents on this project can benefit from those files and hopefully produce better code more quickly. Committing these agent files to the repo will also let other contributors update this document as needed or as new features are added.

The AI policy has been taken from the ghostty project, with attribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ The GitHub issue tracker is for **bugs** only. Please use the [developer mailing
   - [Plugins and GUIs](Documentation/EditorIntegration.md)
   - [Frequently Asked Questions](Documentation/faq.md)
 - [Contributing](CONTRIBUTING.md)
+  - [AI Usage Policy](Documentation/AI/AI_POLICY.md)
   - [Internal Documentation](Documentation/internal)
   - [API documentation](Documentation/api)
   - [How to write a Delve client](Documentation/api/ClientHowto.md)


derekparker:
@mitchellh just a friendly heads up that we would like to adopt the AI policy present in the ghostty project with attribution of course. If our adoption of this policy is in any way unacceptable please let me know and I would be happy to create something from scratch. Ghostty's is so nicely written, however.

mitchellh:
Happy to see it, I don't need attribution at all, just want everyone to be happier working. Thanks!

Documentation/AI/AI_POLICY.md
```diff
@@ -0,0 +1,74 @@
...
+  test on.
+
+- **Issues and discussions can use AI assistance but must have a full
+  human-in-the-loop.** This means that any content generated with AI
+  must have been reviewed _and edited_ by a human before submission.
+  AI is very good at being overly verbose and including noise that
+  distracts from the main point. Humans must do their research and
+  trim this down.
+
+- **No AI-generated media is allowed (art, images, videos, audio, etc.).**
```
aarzilli: I don't see why we should make a distinction between text and non-text. Not that we're ever going to get artwork contributions (we haven't for 10+ years) but if we did why should we reject AI contributions?

Documentation/AI/AI_POLICY.md
```diff
@@ -0,0 +1,74 @@
...
+  Text and code are the only acceptable AI-generated content, per the
+  other rules in this policy.
+
+- **Bad AI drivers will be banned and ridiculed in public.** You've
```
aarzilli: This is unnecessarily aggressive IMO. And besides, we can ban and ridicule whoever we want anyway.

Documentation/AI/AI_POLICY.md
```diff
@@ -0,0 +1,74 @@
...
+
+In a perfect world, AI would produce high-quality, accurate work
+every time. But today, that reality depends on the driver of the AI.
+And today, most drivers of AI are just not good enough. So, until either
+the people get better, the AI gets better, or both, we have to have
+strict rules to protect maintainers.
+
+## AI is Welcome Here
+
+Delve is written with plenty of AI assistance, and many maintainers embrace
```
aarzilli: As things are right now, this is factually false. We did not use plenty of AI assistence, we should use "some".

CLAUDE.md
```diff
@@ -0,0 +1,321 @@
...
+CLI (cmd/dlv) → Cobra commands
+    ↓
+Service Layer (service/) → RPC2, DAP protocol implementations
+    ↓
+Debugger (service/debugger) → High-level debugging operations
+    ↓
+Process Abstraction (pkg/proc) → TargetGroup, Target, ProcessInternal
+    ↓
+Backend Implementations:
+  - native/        → Direct OS debugging (ptrace, Windows APIs)
```
aarzilli: Shouldn't this (and following) be `pkg/proc/native`? Or maybe the AI is going to understand anyway...

CLAUDE.md
```diff
@@ -0,0 +1,321 @@
...
+- **`pkg/gobuild/`** - Go build system integration
+- **`pkg/proc/BinaryInfo`** - Debug symbols, function entries, line tables
+
+### Platform/Architecture Handling
+
+Uses **filename-based conditional compilation**:
+
+- OS-specific: `*_linux.go`, `*_darwin.go`, `*_windows.go`, `*_freebsd.go`
+- Architecture-specific: `regs_amd64.go`, `regs_arm64.go`, etc.
+- **DO NOT put platform-specific code in generic files**
```
aarzilli, line 194: I think we should add this: "in most cases platform-specific code should only be added to the `pkg/proc/native` and `pkg/proc/gdbserial backends`"

derekparker: SGTM

CLAUDE.md
derekparker: Yeah I don't think there's any harm in being as specific as we can.

Documentation/AI/AI_POLICY.md
derekparker: Very true, good catch. Updated.

Documentation/AI/AI_POLICY.md
derekparker: Agreed. I've removed this. As it stands we don't really have an issue with low-quality obviously AI PRs so, we can cross that bridge if we ever get to it.

Documentation/AI/AI_POLICY.md
derekparker: Agreed, this isn't a problem we currently have so we can discuss if it ever becomes an issue in the future. I don't have strong opinions at the moment other than I think I would prefer the artwork of humans, but this can be discussed if/when it ever becomes worth discussing.

aarzilli:
LGTM

PR #4249, opened 2026-01-28
pkg/proc: fix ebpf uprobe placement to skip function prologue

derekparker:
Use FirstPCAfterPrologue instead of fn.Entry when setting entry uprobes to avoid duplicate events when the runtime grows the stack.

No test is included because this actually fixes an intermittent failure in the existing tests.


aarzilli (approved):
LGTM

PR #4251, opened 2026-01-31
Support printing function parameters with varying verbosity levels [0 .. 4] with the -trace option

### LINKED ISSUE #3586: Can dlv trace print the value of the arguments passed to a function?
```
Delve Debugger
Version: 1.21.2
```

1.21.3

4. What operating system and processor architecture are you using?

Linux x86

6. What did you do?

```
dlv trace . k8s.io --timestamp=true 
```
8. What did you expect to see?

The values of the parameters passed to the function

9. What did you see instead?

```
2023-11-27T14:56:54.338358579-08:00 > goroutine(1): main.(*repoSync).SetupDefaultGitConfigs((*main.repoSync)(0xc000099040), context.Context(*context.timerCtx) 0xb
eef000000000108)
2023-11-27T14:56:54.340307306-08:00 > goroutine(1): main.(*repoSync).Run((*main.repoSync)(0xc000099040), context.Context(*context.timerCtx) 0xbeef000000000108, "", []string len: 4, cap: 4, [...])
2023-11-27T14:56:54.3420434-08:00 > goroutine(1): [k8s.io/git-sync/pkg/cmd.Runner.WithCallDepth(k8s.io/git-sync/pkg/cmd.Runner](http://k8s.io/git-sync/pkg/cmd.Runner.WithCallDepth%28k8s.io/git-sync/pkg/cmd.Runner) {log: [k8s.io/git-sync/pkg/cmd.logintf(*k8s.io/git-sync/pkg/logging.Logger)](http://k8s.io/git-sync/pkg/cmd.logintf%28*k8s.io/git-sync/pkg/logging.Logger%29) ...}, 1)
```

archanaravindar:
Fixes #3586 
The current implementation supports the following verbosity levels 
**Verbosity Levels:**
- **Level 0**: No parameter output (just function name)
- **Level 1**: Type names only (e.g., `<int>, <Person>`)
- **Level 2**: Inline values, limited depth (default)
- **Level 3**: Multi-line, expanded detail
- **Level 4**: Full detail with pointer dereferencing

To give an idea of what the output looks like pls refer to the attached .go test file and the related output seen with -trace and -trace-follow options 
[plainout.txt](https://github.com/user-attachments/files/24977514/plainout.txt)
[testout.txt](https://github.com/user-attachments/files/24977515/testout.txt)
[traceverbcomplex.zip](https://github.com/user-attachments/files/24977531/traceverbcomplex.zip)

### DIFF
--- a/cmd/dlv/cmds/commands.go
+++ b/cmd/dlv/cmds/commands.go
@@ -370,8 +370,7 @@ only see the output of the trace operations you can redirect stdout.`,
 	traceCommand.Flags().String("output", "", "Output path for the binary.")
 	must(traceCommand.MarkFlagFilename("output"))
 	traceCommand.Flags().IntVarP(&traceFollowCalls, "follow-calls", "", 0, "Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.")
-	traceCommand.Flags().IntVarP(&traceVerbose, "trace-verbose", "V", 0, "Parameter verbosity: 0=values, 1=types, 2=inline, 3=expanded, 4=full (default 0)")
-	must(traceCommand.RegisterFlagCompletionFunc("trace-verbose", cobra.NoFileCompletions))
+	traceCommand.Flags().IntVarP(&traceVerbose, "verbose", "v", 0, "Parameter verbosity: 0=values, 1=types, 2=inline, 3=expanded, 4=full (default 0)")
 	rootCommand.AddCommand(traceCommand)
 
 	coreCommand := &cobra.Command{
@@ -842,7 +841,7 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
 		}
 		t := terminal.New(client, cfg)
 		t.SetTraceNonInteractive()
-		t.SetTraceVerbosity(traceVerbose)
+		t.TraceVerbosity = traceVerbose
 		t.RedirectTo(os.Stderr)
 		defer t.Close()
 		if traceUseEBPF {
@@ -903,7 +902,6 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
 									fmt.Fprintf(os.Stderr, "\n")
 								}
 								fmt.Fprintf(os.Stderr, "\n%s)\n", paramStr)
-
 							}
 						}
 					}
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -873,14 +873,14 @@ func TestTraceVerbosityLevel0Default(t *testing.T) {
 >> goroutine(1): main.testPrimitives => (84)
 `)
 
-	// Test with explicit --trace-verbose=0
+	// Test with explicit --verbose=0
 	cmd1 := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug1"),
-		filepath.Join(fixtures, "traceverb.go"), "main.testPrimitives", "--trace-verbose", "0")
+		filepath.Join(fixtures, "traceverb.go"), "main.testPrimitives", "--verbose", "0")
 	rdr1, err := cmd1.StderrPipe()
 	assertNoError(err, t, "stderr pipe")
 	defer rdr1.Close()
 	cmd1.Dir = filepath.Join(fixtures, "buildtest")
-	assertNoError(cmd1.Start(), t, "running trace with --trace-verbose=0")
+	assertNoError(cmd1.Start(), t, "running trace with --verbose=0")
 	output1, _ := io.ReadAll(rdr1)
 	assertNoError(cmd1.Wait(), t, "cmd1.Wait()")
 
@@ -897,7 +897,7 @@ func TestTraceVerbosityLevel0Default(t *testing.T) {
 
 	// Both should contain the same values-only format
 	if !bytes.Contains(output1, expected) {
-		t.Fatalf("--trace-verbose=0 expected:\n%s\ngot:\n%s", expected, output1)
+		t.Fatalf("--verbose=0 expected:\n%s\ngot:\n%s", expected, output1)
 	}
 	if !bytes.Contains(output2, expected) {
 		t.Fatalf("default expected:\n%s\ngot:\n%s", expected, output2)
@@ -915,7 +915,7 @@ func TestTraceVerbosityLevel1(t *testing.T) {
 
 	fixtures := protest.FindFixturesDir()
 	cmd := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug"),
-		filepath.Join(fixtures, "traceverb.go"), "main.testPrimitives", "--trace-verbose", "1")
+		filepath.Join(fixtures, "traceverb.go"), "main.testPrimitives", "--verbose", "1")
 	rdr, err := cmd.StderrPipe()
 	assertNoError(err, t, "stderr pipe")
 	defer rdr.Close()
@@ -942,7 +942,7 @@ func TestTraceVerbosityLevel2(t *testing.T) {
 
 	fixtures := protest.FindFixturesDir()
 	cmd := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug"),
-		filepath.Join(fixtures, "traceverb.go"), "main.testStruct", "--trace-verbose", "2")
+		filepath.Join(fixtures, "traceverb.go"), "main.testStruct", "--verbose", "2")
 	rdr, err := cmd.StderrPipe()
 	assertNoError(err, t, "stderr pipe")
 	defer rdr.Close()
@@ -971,7 +971,7 @@ func TestTraceVerbosityLevel3(t *testing.T) {
 
 	fixtures := protest.FindFixturesDir()
 	cmd := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug"),
-		filepath.Join(fixtures, "traceverb.go"), "main.testStruct", "--trace-verbose", "3")
+		filepath.Join(fixtures, "traceverb.go"), "main.testStruct", "--verbose", "3")
 	rdr, err := cmd.StderrPipe()
 	assertNoError(err, t, "stderr pipe")
 	defer rdr.Close()
@@ -1006,7 +1006,7 @@ func TestTraceVerbosityLevel4(t *testing.T) {
 
 	fixtures := protest.FindFixturesDir()
 	cmd := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug"),
-		filepath.Join(fixtures, "traceverb.go"), "main.testNested", "--trace-verbose", "4")
+		filepath.Join(fixtures, "traceverb.go"), "main.testNested", "--verbose", "4")
 	rdr, err := cmd.StderrPipe()
 	assertNoError(err, t, "stderr pipe")
 	defer rdr.Close()
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -1945,7 +1945,7 @@ func setBreakpoint(t *Term, ctx callContext, tracepoint bool, argstr string) ([]
 		requestedBp.AddrPid = loc.PCPids
 		if tracepoint {
 			requestedBp.LoadArgs = &ShortLoadConfig
-			t.traceVerbosity = 0 // Default verbosity for terminal traces
+			t.TraceVerbosity = 0 // Default verbosity for terminal traces
 		}
 
 		bp, err := t.client.CreateBreakpointWithExpr(requestedBp, spec, t.substitutePathRules(), false)
@@ -2989,8 +2989,7 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
 		tracePrefix = fmt.Sprintf("goroutine(%d):", th.GoroutineID)
 	}
 
-	// Get verbosity level from terminal
-	verbosity := t.traceVerbosity
+	verbosity := t.TraceVerbosity
 	if th.Breakpoint.Tracepoint {
 		// Print trace only if there was a match on the function while TraceFollowCalls is on or if it's a regular trace
 		if rootindex != -1 || th.Breakpoint.TraceFollowCalls <= 0 {
@@ -3019,7 +3018,6 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
 				}
 			}
 
-			// Join parameters
 			var paramStr string
 			if len(params) > 0 {
 				if verbosity >= 3 {
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -83,7 +83,7 @@ type Term struct {
 	quitting      bool
 
 	traceNonInteractive bool
-	traceVerbosity      int // Verbosity level for trace output (0-4)
+	TraceVerbosity      int // Verbosity level for trace output (0-4)
 
 	downloadsMu         sync.Mutex
 	downloadsInProgress bool
@@ -235,10 +235,6 @@ func (t *Term) SetTraceNonInteractive() {
 
 func (t *Term) IsTraceNonInteractive() bool {
 	return t.traceNonInteractive
-}
-
-func (t *Term) SetTraceVerbosity(v int) {
-	t.traceVerbosity = v
 }
 
 // Close returns the terminal to its previous mode.
--- a/Documentation/usage/dlv_trace.md
+++ b/Documentation/usage/dlv_trace.md
@@ -21,16 +21,16 @@ dlv trace [package] regexp [flags]
 ### Options
 
 ```
-      --ebpf                Trace using eBPF (experimental).
-  -e, --exec string         Binary file to exec and trace.
-      --follow-calls int    Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.
-  -h, --help                help for trace
-      --output string       Output path for the binary.
-  -p, --pid int             Pid to attach to.
-  -s, --stack int           Show stack trace with given depth. (Ignored with --ebpf)
-  -t, --test                Trace a test binary.
-      --timestamp           Show timestamp in the output
-  -V, --trace-verbose int   Parameter verbosity: 0=values, 1=types, 2=inline, 3=expanded, 4=full (default 0)
+      --ebpf               Trace using eBPF (experimental).
+  -e, --exec string        Binary file to exec and trace.
+      --follow-calls int   Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.
+  -h, --help               help for trace
+      --output string      Output path for the binary.
+  -p, --pid int            Pid to attach to.
+  -s, --stack int          Show stack trace with given depth. (Ignored with --ebpf)
+  -t, --test               Trace a test binary.
+      --timestamp          Show timestamp in the output
+  -v, --verbose int        Parameter verbosity: 0=values, 1=types, 2=inline, 3=expanded, 4=full (default 0)
 ```
 
 ### Options inherited from parent commands


cmd/dlv/cmds/commands.go
```diff
@@ -789,14 +792,17 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
 				if traceFollowCalls > 0 && stackdepth == 0 {
 					stackdepth = 20
 				}
+				// Get LoadConfig based on verbosity level
+				loadCfg := api.GetLoadConfigForVerbosity(traceVerbose)
```
aarzilli: Why is this a function in a different package?

cmd/dlv/cmds/commands.go
```diff
@@ -370,6 +371,8 @@ only see the output of the trace operations you can redirect stdout.`,
 	traceCommand.Flags().String("output", "", "Output path for the binary.")
 	must(traceCommand.MarkFlagFilename("output"))
 	traceCommand.Flags().IntVarP(&traceFollowCalls, "follow-calls", "", 0, "Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.")
+	traceCommand.Flags().IntVarP(&traceVerbose, "trace-verbose", "V", 2, "Parameter verbosity: 0=none, 1=types, 2=inline, 3=expanded, 4=full (default 2)")
```
aarzilli: I would expect the default to be the same thing we do now, for backwards compatibility. We don't guarantee backwards compatibility for this but I don't see a compelling reason not to do it in this case.

pkg/terminal/command.go
```diff
@@ -3056,21 +3098,50 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
 		tracePrefix = fmt.Sprintf("goroutine(%d):", th.GoroutineID)
 	}
 
+	// Get verbosity level from breakpoint
+	verbosity := 0
+	if th.Breakpoint != nil {
```
aarzilli: We already know `th.Breakpoint != nil` here.

service/api/types.go
```diff
@@ -142,6 +142,9 @@ type Breakpoint struct {
 	RootFuncName string
 	// TraceFollowCalls indicates the Depth of tracing
 	TraceFollowCalls int
+	// Verbosity level to print function parameters as part of the trace [0-4]
+	TraceVerbosity   int
```
aarzilli: I don't think this belongs here. The only piece of code that uses it (as far as I can see), can just check the value of LoadArgs.

pkg/terminal/command.go
```diff
@@ -3056,21 +3098,50 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
 		if rootindex != -1 || th.Breakpoint.TraceFollowCalls <= 0 {
-			fmt.Fprintf(t.stdout, "%s> %s %s%s(%s)\n", depthPrefix, tracePrefix, bpname, fn.Name(), args)
+			// Format parameters based on verbosity
+			paramStr := formatTraceParameters(th, verbosity)
+			if verbosity >= 3 && paramStr != "" {
+				// Levels 3-4: Multi-line format
+				fmt.Fprintf(t.stdout, "%s> %s %s%s\n%s\n",
+					depthPrefix, tracePrefix, bpname, fn.Name(), paramStr)
+			} else {
+				// Levels 0-2: Single-line format
```
aarzilli: These two branches do the same thing.

pkg/terminal/command.go
```diff
@@ -3056,21 +3098,50 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+
 	if th.Breakpoint.Tracepoint {
 		// Print trace only if there was a match on the function while TraceFollowCalls is on or if it's a regular trace
 		if rootindex != -1 || th.Breakpoint.TraceFollowCalls <= 0 {
-			fmt.Fprintf(t.stdout, "%s> %s %s%s(%s)\n", depthPrefix, tracePrefix, bpname, fn.Name(), args)
+			// Format parameters based on verbosity
+			paramStr := formatTraceParameters(th, verbosity)
+			if verbosity >= 3 && paramStr != "" {
+				// Levels 3-4: Multi-line format
+				fmt.Fprintf(t.stdout, "%s> %s %s%s\n%s\n",
```
aarzilli: Move the common parts of these Fprintf calls out of the if.

pkg/terminal/command.go
```diff
@@ -3056,21 +3098,50 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
 	}
 	if th.Breakpoint.TraceReturn {
-		retVals := make([]string, 0, len(th.ReturnValues))
-		for _, v := range th.ReturnValues {
-			retVals = append(retVals, v.SinglelineString())
-		}
 		// Print trace only if there was a match on the function while TraceFollowCalls is on or if it's a regular trace
 		if rootindex != -1 || th.Breakpoint.TraceFollowCalls <= 0 {
-			fmt.Fprintf(t.stdout, "%s>> %s %s => (%s)\n", depthPrefix, tracePrefix, fn.Name(), strings.Join(retVals, ","))
+			if verbosity > 0 {
```
aarzilli: Move the common parts of the fmt.Fprintf calls out of the if.

pkg/terminal/command_test.go
```diff
@@ -300,9 +300,10 @@ func TestTrace(t *testing.T) {
 	withTestTerminal("issue573", t, func(term *FakeTerminal) {
 		term.MustExec("trace foo")
 		out, _ := term.Exec("continue")
+		fmt.Printf("output %s\n", out)
```
aarzilli: t.Logf

service/api/prettyprint.go
```diff
@@ -68,6 +68,24 @@ func (v *Variable) MultilineString(indent, fmtstr string) string {
 	return buf.String()
 }
 
+// FormatWithVerbosity formats a variable according to trace verbosity level
+func (v *Variable) FormatWithVerbosity(verbosity int) string {
```
aarzilli: This is a helper function that is only used by the pkg/terminal package, it belongs there.

cmd/dlv/cmds/commands.go
derekparker: Yeah, I agree with this.

pkg/terminal/command.go
archanaravindar: The branch under verbosity>=3 has a newline before parameter string paramStr so that the detailed parameter values are clearly visibile and dont mess with the rest of the trace output 
`
                                fmt.Fprintf(t.stdout, "%s> %s %s%s\n%s\n",
`
while the branch in levels 0-2 prints all of them on a single line and hence has  \n only at the end
`
                                        fmt.Fprintf(t.stdout, "%s> %s %s%s(%s)\n",
`

service/api/types.go
archanaravindar: Having a separate variable to track verbosity is cleaner, than having to reverse engineer verbosity everytime by looking at the constituents of LoadArgs. Further if there are any changes to verbosity down the line, having a variable to map verbosity to loadcfg abstracts the details

cmd/dlv/cmds/commands.go
archanaravindar: I will work on making it consistent, thanks

pkg/terminal/command.go
archanaravindar: Ack

pkg/terminal/command.go
archanaravindar: you meant like this? 
```
   fmt.Fprintf(t.stdout, "%s> %s %s%s",
                                        depthPrefix, tracePrefix, bpname, fn.Name())
                        if verbosity >= 3 && paramStr != "" {
                                // Levels 3-4: Multi-line format
                                fmt.Fprintf(t.stdout, "\n%s\n", paramStr)
                        } else {
                                // Levels 0-2: Single-line format
                                if paramStr != "" {
                                        fmt.Fprintf(t.stdout, "(%s)\n", paramStr)
                                } else {
                                        fmt.Fprintf(t.stdout, "()\n")
                                }
                        }

```

pkg/terminal/command_test.go
archanaravindar: Ack

service/api/prettyprint.go
archanaravindar: Wouldnt it be better for all variable formatting logic to stay together? this function operates on the variable and uses internal prettyprint structures, moving it to pkg/terminal may confuse the developers down the line 
For example SinglelineStringWithShortTypes which resides in service/api is used by the DAP server

service/api/types.go
aarzilli: Look how longLoadConfig is used in command.go, that has been in place for 10 years.

pkg/terminal/command.go
aarzilli: I mean the branches below this comment. If `paramStr` is `""` you are printing "()" if it isn't you are printing "(%s)" and passing paramStr to it. They do the same thing.

pkg/terminal/command.go
aarzilli: Yes.

service/api/prettyprint.go
aarzilli: The verbosity parameter and its interpretation are too tied to one specific use of prettyprint. SinglelineStringWithShortTypes is borderline, I'd say that the fact that FormatWithVerbosity is being written means we have to delete SinglelineStringWithShortTypes and SinglelineStringFormatted too and make a `StringWithOptions(fmtstr string, flags PrettyFlags) string` function.

service/api/prettyprint.go
```diff
@@ -68,6 +68,47 @@ func (v *Variable) MultilineString(indent, fmtstr string) string {
...
+			return fmt.Sprintf("(unreadable %s)", v.Unreadable)
+		}
+
+		// For complex types (slices, maps, structs, etc.), use writeTo without type info
+		// This gives us the value representation without the type name prefix
+		var buf bytes.Buffer
+		switch v.Kind {
+		case reflect.Slice, reflect.Array, reflect.Map, reflect.Struct, reflect.Ptr, reflect.Interface, reflect.Chan:
+			// For complex types, format without type names
+			v.writeTo(&buf, 0, "", "")
```
aarzilli: I think calling `writeTo` on a variable returned by eval without passing the prettyTop flag to it is wrong. I think it works on your specific use case but might not in general, for expressions that don't have an address.

service/api/prettyprint.go
```diff
@@ -68,6 +68,47 @@ func (v *Variable) MultilineString(indent, fmtstr string) string {
...
+			return buf.String()
+		case reflect.String:
+			// For strings, use quoted format
+			s := v.Value
+			if len(s) != int(v.Len) {
+				s = fmt.Sprintf("%s...+%d more", s, int(v.Len)-len(s))
+			}
+			return fmt.Sprintf("%q", s)
+		default:
+			// For basic types, return just the value
```
aarzilli: This branch doesn't just cover basic types, you are not handling all possible values of Kind.

service/api/prettyprint.go
```diff
@@ -68,6 +68,47 @@ func (v *Variable) MultilineString(indent, fmtstr string) string {
 	return buf.String()
 }
 
+// FormatWithVerbosity formats a variable according to trace verbosity level
+func (v *Variable) FormatWithVerbosity(verbosity int) string {
+	// Level 0: just the value (master branch compatibility)
```
aarzilli: This comment will make no sense to someone reading it in the future. Also how comes backwards compatiblity requires all this new code?

cmd/dlv/dlv_test.go
```diff
@@ -794,7 +794,7 @@ func TestTraceDirRecursion(t *testing.T) {
 	t.Parallel()
 	dlvbin := protest.GetDlvBinary(t)
 
-	expected := []byte("> goroutine(1):frame(1) main.A(5, 5)\n > goroutine(1):frame(2) main.A(4, 4)\n  > goroutine(1):frame(3) main.A(3, 3)\n   > goroutine(1):frame(4) main.A(2, 2)\n    > goroutine(1):frame(5) main.A(1, 1)\n    >> goroutine(1):frame(5) main.A => (1)\n   >> goroutine(1):frame(4) main.A => (2)\n  >> goroutine(1):frame(3) main.A => (6)\n >> goroutine(1):frame(2) main.A => (24)\n>> goroutine(1):frame(1) main.A => (120)\n")
+	    expected := []byte("> goroutine(1):frame(1) main.A(5, 5)\n > goroutine(1):frame(2) main.A(4, 4)\n  > goroutine(1):frame(3) main.A(3, 3)\n   > goroutine(1):frame(4) main.A(2, 2)\n    > goroutine(1):frame(5) main.A(1, 1)\n    >> goroutine(1):frame(5) main.A => (1)\n   >> goroutine(1):frame(4) main.A => (2)\n  >> goroutine(1):frame(3) main.A => (6)\n >> goroutine(1):frame(2) main.A => (24)\n>> goroutine(1):frame(1) main.A => (120)\n")
```
aarzilli: go fmt.

service/api/prettyprint.go
archanaravindar: I agree, will make the necessary change

service/api/prettyprint.go
archanaravindar: you are right its very confusing, changed it to match default behavior

aarzilli: There are still several places where this nomenclature is used.

service/api/prettyprint.go
archanaravindar: ack

service/api/prettyprint.go
archanaravindar: Thanks for catching it! made the changes

pkg/terminal/command.go
```diff
@@ -3060,20 +3061,66 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
 		tracePrefix = fmt.Sprintf("goroutine(%d):", th.GoroutineID)
 	}
 
+	// Get verbosity level from terminal
+	verbosity := t.traceVerbosity
```
derekparker: Do we actually need this? This is for tracepoints set during a debug session, other non-interactive tracepoints get printed via the `traceCmd`. Verbosity level wouldn't translate to this.

archanaravindar: Hi @derekparker printTracepoint is invoked in non-interactive breakpoint based tracing, (such as the code path within IsTraceNonInteractive()) 
this is also needed when we use -V in conjunction with the  --follow-calls option to print parameters of child calls 
Since in all these cases, we now support printing according to verbosity level, setting this variable is a requirement

derekparker: Ack, ok makes sense.

pkg/terminal/command_test.go
```diff
@@ -300,6 +300,7 @@ func TestTrace(t *testing.T) {
 	withTestTerminal("issue573", t, func(term *FakeTerminal) {
 		term.MustExec("trace foo")
 		out, _ := term.Exec("continue")
+		t.Logf("output %q", out)
```
derekparker: Leftover?

cmd/dlv/dlv_test.go
```diff
@@ -856,6 +857,166 @@ func TestTraceMultipleGoroutines(t *testing.T) {
...
+	rdr, err := cmd.StderrPipe()
+	assertNoError(err, t, "stderr pipe")
+	defer rdr.Close()
+
+	cmd.Dir = filepath.Join(fixtures, "buildtest")
+
+	assertNoError(cmd.Start(), t, "running trace")
+	output, _ := io.ReadAll(rdr)
+
+	// For verbosity 4, both first-level and nested structs should be fully expanded
```
derekparker: Is this output deterministic? If so, we should maybe just assert on the entire output, including formatting, like we do with other trace tests, instead of asserting that we see certain struct members, etc...

This also gives a visual of what the output should look like in the test.

cmd/dlv/cmds/commands.go
```diff
@@ -1262,3 +1287,54 @@ func must(err error) {
 		log.Fatal(err)
 	}
 }
+
+// getLoadConfigForVerbosity returns the LoadConfig for a given verbosity level
+func getLoadConfigForVerbosity(verbosity int) api.LoadConfig {
+	switch verbosity {
+	case 0:
+		// Level 0: default case(ShortLoadConfig)
+		return api.LoadConfig{
```
derekparker, line 1294: Maybe these should be named values that live in the `api` package, that way we don't have to store verbosity level in terminal and we can key off this.

For example:

```go
var LoadConfigTraceVerbosity1 = LoadConfig{ ... }
var LoadConfigTraceVerbosity2 = LoadConfig{ ... }
```

And then instead of storing and switching on verbosity level you could just do:

```go
switch verbosity {
case api.LoadConfigTraceVerbosity1:
   ...
...
}
```

Since we already are storing that information when creating the breakpoint. If this ends up not working the way I envision it storing the trace verbosity isn't the end of the world.

archanaravindar: @derekparker Tying verbosity to loadconfig removes the abstraction of what to display and how to display it,  Down the line if there are changes to verbosity the change in the code wont be apparent to the user, I still feel having a variable is cleaner than storing it in a mapped form to various loadconfigs

derekparker: Ack, I'm ok with keeping the verbosity boolean around.

cmd/dlv/dlv_test.go
```diff
@@ -856,6 +857,166 @@ func TestTraceMultipleGoroutines(t *testing.T) {
...
+	}
+}
+
+// Test trace verbosity level 1 (types only)
+func TestTraceVerbosityLevel1(t *testing.T) {
+	t.Parallel()
+	dlvbin := protest.GetDlvBinary(t)
+
+	fixtures := protest.FindFixturesDir()
+	cmd := exec.Command(dlvbin, "trace", "--output", filepath.Join(t.TempDir(), "__debug"),
```
aarzilli, line 917: @derekparker why do the trace tests pass an explicit `--output` option? It looks like it could create problems along with `t.Parallel`.

cmd/dlv/dlv_test.go
archanaravindar: @derekparker I will refer to the existing trace tests and follow their lead to assert on the output including formatting and modify the test, thanks

cmd/dlv/cmds/commands.go
```diff
@@ -370,6 +370,8 @@ only see the output of the trace operations you can redirect stdout.`,
 	traceCommand.Flags().String("output", "", "Output path for the binary.")
 	must(traceCommand.MarkFlagFilename("output"))
 	traceCommand.Flags().IntVarP(&traceFollowCalls, "follow-calls", "", 0, "Trace all children of the function to the required depth. Trace also supports defer functions and cases where functions are dynamically returned and passed as parameters.")
+	traceCommand.Flags().IntVarP(&traceVerbose, "trace-verbose", "V", 0, "Parameter verbosity: 0=values, 1=types, 2=inline, 3=expanded, 4=full (default 0)")
```
aarzilli: I think this should just be `verbose` and the short form should be `v`, to be consistent with the `version` command.

cmd/dlv/cmds/commands.go
```diff
@@ -370,6 +370,8 @@ only see the output of the trace operations you can redirect stdout.`,
...
+	must(traceCommand.RegisterFlagCompletionFunc("trace-verbose", cobra.NoFileCompletions))
```
aarzilli: I don't know exactly how those work but the other boolean flag don't have this. Probably neither should this one.

cmd/dlv/cmds/commands.go
```diff
@@ -874,11 +893,17 @@ func traceCmd(cmd *cobra.Command, args []string, conf *config.Config) int {
...
 								fmt.Fprintf(os.Stderr, ">> goroutine(%d): %s => (%s)\n", t.GoroutineID, t.FunctionName, strings.Join(retVals, ","))
 							} else {
-								fmt.Fprintf(os.Stderr, "> goroutine(%d): %s(%s)\n", t.GoroutineID, t.FunctionName, params.String())
+								fmt.Fprintf(os.Stderr, "> goroutine(%d): %s(", t.GoroutineID, t.FunctionName)
+								if traceVerbose >= 3 {
+									// Levels 3-4: Multi-line format
+									fmt.Fprintf(os.Stderr, "\n")
+								}
+								fmt.Fprintf(os.Stderr, "\n%s)\n", paramStr)
+
```
aarzilli: Superfluous empty line.

pkg/terminal/command.go
```diff
@@ -2988,20 +2989,66 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
 		tracePrefix = fmt.Sprintf("goroutine(%d):", th.GoroutineID)
 	}
 
+	// Get verbosity level from terminal
```
aarzilli: Redundant comment.

pkg/terminal/command.go
```diff
@@ -2988,20 +2989,66 @@ func printTracepoint(t *Term, th *api.Thread, bpname string, fn *api.Function, a
...
+						prefix := ""
+						if verbosity >= 3 {
+							prefix = "  " // Indent for multi-line format
+						}
+						params = append(params, fmt.Sprintf("%s%s: %s", prefix, arg.Name, formatted))
+					}
+				}
+			}
+
+			// Join parameters
```
aarzilli: Redundant comment.

pkg/terminal/terminal.go
```diff
@@ -83,6 +83,7 @@ type Term struct {
 	quitting      bool
 
 	traceNonInteractive bool
+	traceVerbosity      int // Verbosity level for trace output (0-4)
```
aarzilli: I'd just make this member public and skip the setter.

cmd/dlv/dlv_test.go
derekparker: IIRC to enable `t.Parallel` since it's prefixing the output with `t.TempDir()` which will be unique per test.

cmd/dlv/cmds/commands.go
archanaravindar: I agree. since they are subcommands of trace, trace-verbose looks redundant

cmd/dlv/cmds/commands.go
archanaravindar: the documentation says it works for autocompleting when you press tab with file names etc, i guess that s why it is added for some parameters such as pid/stack, no harm in not having it for the verbose variable as it does not exist for follow-calls either

pkg/terminal/terminal.go
archanaravindar: Ack

pkg/terminal/command.go
archanaravindar: Ack

pkg/terminal/command_test.go
archanaravindar: removed

cmd/dlv/dlv_test.go
aarzilli: Ack.

PR #4223, opened 2025-12-22
terminal/starlark: fix string output with `write_file()`

89oso:
Fixes #4222

`example.star`:
```starlark
def main():
    write_file("test1", "hello\nworld")
    # write_file("test2", 123)  # error
```

`test1`:
```
hello
world
```

PR #4252, opened 2026-02-06
terminal/starbind: add 'append_file" built-in func

### LINKED ISSUE #4222: Unexpected output when using `write_file()` in starlark script

1.25.2

go version go1.25.4 linux/amd64

Linux/NixOS - amd64

### Description

Writing any output using `write_file(...)` inside a starlark script results in that output being quoted.

`example.star`:
```starlark
def main():
    write_file("foo", "line1\nline2")
```

```
...
(dlv) source example.star
...
```

The contents of `foo`:
```
"line1\nline2"
```

Expected contents of `foo`:
```
line1
line2
```

This behavior prevents making functions like:
```starlark
def command_save_breakpoints():
    buffer = ""
    bps = breakpoints().Breakpoints

    for bp in bps:
        if bp.ID <= 0:
            continue
        buffer += "b " + bp.File + ":" + str(bp.Line)
        if bp.Cond != "":
            buffer += " if " + bp.Cond
        buffer += "\n"

    if buffer != "":
        write_file(".breakpoints", buffer)
        print("Breakpoints saved to .breakpoints")

def command_load_breakpoints():
    dlv_command("source .breakpoints")
```

The issue is caused by the usage of the `String()` method on a `starlark.Value`.

```go
// vendor/go.starlark.net/starlark/value.go
...
// Value is a value in the Starlark interpreter.
type Value interface {
	// String returns the string representation of the value.
	// Starlark string values are quoted as if by Python's repr.
	String() string
...
}
```

```go
// pkg/terminal/starbind/starlark.go:128
err := os.WriteFile(string(path), []byte(args[1].String()), 0o640)
```

sding3:
This change adds the 'append_file' which will append the specified string to the specified OS file.
This update also fixes #4222.

### Testing

append.star:
```py
def command_append(args):
  append_file("./debug.txt", 'xxx\n')
  dlv_command("continue")
```

/debug.txt:

```
xxx
xxx
xxx
```

### DIFF
--- a/pkg/terminal/starbind/starlark.go
+++ b/pkg/terminal/starbind/starlark.go
@@ -131,16 +131,7 @@ func New(ctx Context, out EchoWriter) *Env {
 			return nil, decorateError(thread, err)
 		}
 		defer f.Close()
-		var data []byte
-		switch v := args[1].(type) {
-		case starlark.String:
-			data = []byte(string(v))
-		case starlark.Bytes:
-			data = []byte(v)
-		default:
-			data = []byte(args[1].String())
-		}
-		_, err = f.Write(data)
+		_, err = f.Write(toBytes(args[1]))
 		return starlark.None, decorateError(thread, err)
 	})
 	builtindoc(appendFileBuiltinName, "(Path, Text)", "append text to the specified file.")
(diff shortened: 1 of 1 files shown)


pkg/terminal/starbind/starlark.go
```diff
@@ -117,6 +118,33 @@ func New(ctx Context, out EchoWriter) *Env {
...
+		path, ok := args[0].(starlark.String)
+		if !ok {
+			return nil, decorateError(thread, errors.New("first argument of append_file was not a string"))
+		}
+		f, err := os.OpenFile(string(path), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o640)
+		if err != nil {
+			return nil, decorateError(thread, err)
+		}
+		defer f.Close()
+		var data []byte
```
aarzilli: It's weird that this and write_file end up having different behavior with respects to how they treat their second argument. There is an issue opened about write_file #4223 but it seems to have been abandoned. I'd say make write_file do the same thing (by calling a common function).

sding3: Sounds good - updated in c9485d2c471fc6cdb227f6bce014a4c37510c6f2

aarzilli (approved):
LGTM

PR #4255, opened 2026-02-06
goversion: parse rc version with trailing suffix

alexsaezm:
While updating Delve in Fedora 44, I found out that it is failing due to our currently uncommon Go version output: `go version go1.26rc2-X:nodwarf5 linux/amd64`.

The problem as I see it is that it fails to identify long release candidate versions with flags.

```log
+ echo '=== Start tests ==='
++ go list ./...
++ grep -v cmd
++ grep -v scripts
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf -skip TestGuessSubstitutePath
ok  	github.com/go-delve/delve/pkg/dwarf	0.003s
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf/dwarfbuilder -skip TestGuessSubstitutePath
?   	github.com/go-delve/delve/pkg/dwarf/dwarfbuilder	[no test files]
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf/frame -skip TestGuessSubstitutePath
ok  	github.com/go-delve/delve/pkg/dwarf/frame	0.003s
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf/godwarf -skip TestGuessSubstitutePath
ok  	github.com/go-delve/delve/pkg/dwarf/godwarf	0.003s
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf/leb128 -skip TestGuessSubstitutePath
ok  	github.com/go-delve/delve/pkg/dwarf/leb128	0.003s
+ for d in $(go list ./... | grep -v cmd | grep -v scripts)
+ go test github.com/go-delve/delve/pkg/dwarf/line -skip TestGuessSubstitutePath
--- FAIL: TestDebugLinePrologueParser (0.19s)
    line_parser_test.go:121: Include dirs not parsed correctly
FAIL
FAIL	github.com/go-delve/delve/pkg/dwarf/line	0.421s
FAIL
RPM build errors:
```

### DIFF
--- a/pkg/goversion/go_version.go
+++ b/pkg/goversion/go_version.go
@@ -119,9 +119,15 @@ func Parse(ver string) (GoVersion, bool) {
 				// old boringcrypto version goX.YbZ
 				_, err3 = strconv.Atoi(rest)
 			case hasPrefix("rc"):
-				// old rc release goX.YrcZ
+				i := 0
+				for i < len(rest) && rest[i] >= '0' && rest[i] <= '9' {
+					i++
+				}
 				var rc int
-				rc, err3 = strconv.Atoi(rest)
+				rc, err3 = strconv.Atoi(rest[:i])
+				if err3 != nil || i == 0 {
+					return GoVersion{}, false
+				}
 				r.Rev = rcRev(rc)
 			default:
 				// what is this?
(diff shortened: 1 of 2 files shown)


aarzilli:
Do you know how you get the version string to be like that? I want to try with a released version, this bug may also exist with non-rc releases.

alexsaezm:
> Do you know how you get the version string to be like that? I want to try with a released version, this bug may also exist with non-rc releases.

In that very specific case, I modified [the GOEXPERIMENT inside the go.env file](https://src.fedoraproject.org/rpms/golang/blob/rawhide/f/0001-Modify-go.env.patch#_39) during bootstrapping. Another way could be to edit the VERSION file of a release branch ([we do this in RHEL for example](https://gitlab.com/redhat/centos-stream/rpms/golang/-/blob/c10s/golang.spec?ref_type=heads#L357)).

PR #4257, opened 2026-02-09
deps: migrate from gopkg.in/yaml.v3 to go.yaml.in/yaml/v3

scop:
gopkg.in/yaml.v3 is unmaintained, ref
https://github.com/yaml/go-yaml#project-status

Upgrade to 3.0.4 while at it.

3.0.1...3.0.2 diff (3.0.1 are identical):
https://gist.github.com/scop/6ec72debf62a9603cff9dc97e6814ddd

3.0.2...3.0.4 diff:
https://github.com/yaml/go-yaml/compare/v3.0.2...v3.0.4

*Note*: I did not include the vendor/ changes in this PR to keep it small, as a simple `go mod vendor` will take care of it. Let me know if I should include those changes here, too.


go.mod
```diff
@@ -28,4 +28,5 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	golang.org/x/exp v0.0.0-20230224173230-c95f2b4c22f2 // indirect
 	golang.org/x/sync v0.8.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
scop: I suspect upgrading cobra will make this one disappear, as current versions of it have migrated to go.yaml.in/yaml/v3 already too.

derekparker:
Hey thanks for the patch.

Yes please include vendor updates as well, in a separate commit for easier reviewing of the main change.

Might as well wrap the cobra update in this patch to eliminate the dependency if it's not too noisy and there are no breaking changes.

scop:
Cheers, added three commits, one for vendoring go.yaml.in/yaml/v3, another to upgrade cobra, and finally one to vendor the cobra upgrade changes.

go.mod
scop: Verified, it does.

aarzilli:
See test failures.

scop:
Fixed obvious ones, I'm not sure if the remaining ones have anything to do with these changes. There was some variance between runs that smell like flakiness.

derekparker (approved):
LGTM

PR #4258, opened 2026-02-09
pkg/debugdetect: add package for detecting debugger attachment

### LINKED ISSUE #4243: Add helper package to detect if program is currently being debugged
For some application, e.g. TUI applications, it is not really feasible to start them with delve itself to debug them. Instead, the debugger has to be attached to the program later. 
But if the problem you try to debug happens on startup there is no time to attach the debugger.
`lazygit`, for example implements a [helper function](https://github.com/jesseduffield/lazygit/blob/19605ad47699888da471ddf17bd2ae6dbb1abf3a/pkg/integration/clients/injector/main.go#L72-L88) to wait until a debugger is attached. However, it is not really compatible with anything other than macOS.

So I think it would be great if delve could provide a subpackage that provides such a `isDebuggerAttached` function that is compatible with different platforms.

* Linux: https://stackoverflow.com/a/68210536 (Uses `/proc/self/status`)
* MacOS: https://stackoverflow.com/a/2200786 (Uses `sysctl`-syscall)
* Windows: https://stackoverflow.com/a/16687225 (Uses Function from win32 `debugapi.h`)

What do you think? Would a feature like this be "in scope" for delve?

derekparker:
Implement new pkg/debugdetect package that detects if a program is running under a debugger on Linux, macOS, Windows, and FreeBSD.

Platform-specific detection methods:
- Linux: Parse /proc/self/status for TracerPid field
- macOS: Use sysctl with KERN_PROC_PID to check P_TRACED flag
- Windows: Call IsDebuggerPresent() Win32 API
- FreeBSD: Try sysctl first, fall back to /proc/curproc/status

Fixes #4243

### DIFF
--- a/pkg/debugdetect/detect_test.go
+++ b/pkg/debugdetect/detect_test.go
@@ -39,6 +39,50 @@ func TestIntegration_NotAttached(t *testing.T) {
 	}
 }
 
+func TestIntegration_WaitForDebugger(t *testing.T) {
+	// This test verifies that WaitForDebugger() blocks until a debugger
+	// attaches and then returns successfully by running the fixture
+	// under Delve.
+	protest.AllowRecording(t)
+
+	dlvbin := protest.GetDlvBinary(t)
+	fixturesDir := protest.FindFixturesDir()
+	fixtureSrc := filepath.Join(fixturesDir, "waitfordebugger.go")
+
+	const listenAddr = "127.0.0.1:40581"
+	cmd := exec.Command(dlvbin, "debug", fixtureSrc, "--headless", "--continue", "--accept-multiclient", "--listen", listenAddr)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stdout.Close()
+
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Read stdout until we see the program output
+	scanner := bufio.NewScanner(stdout)
+	foundOutput := false
+	for scanner.Scan() {
+		line := scanner.Text()
+		t.Log(line)
+		if strings.Contains(line, "DEBUGGER_FOUND") {
+			foundOutput = true
+			break
+		}
+	}
+
+	// Clean up - connect and detach
+	client := rpc2.NewClient(listenAddr)
+	client.Detach(true)
+	cmd.Wait()
+
+	if !foundOutput {
+		t.Error("expected 'DEBUGGER_FOUND' in output when running under debugger")
+	}
+}
+
 func TestIntegration_Attached(t *testing.T) {
 	// This test verifies that IsDebuggerAttached() returns true when
 	// the process is actually running under a debugger by using
--- a/_fixtures/waitfordebugger.go
+++ b/_fixtures/waitfordebugger.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/go-delve/delve/pkg/debugdetect"
+)
+
+func main() {
+	err := debugdetect.WaitForDebugger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "ERROR: %v\n", err)
+		os.Exit(2)
+	}
+	fmt.Println("DEBUGGER_FOUND")
+}
--- a/pkg/debugdetect/detect.go
+++ b/pkg/debugdetect/detect.go
@@ -3,6 +3,7 @@ package debugdetect
 import (
 	"fmt"
 	"runtime"
+	"time"
 )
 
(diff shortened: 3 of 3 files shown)


pkg/debugdetect/detect_darwin.go
```diff
@@ -0,0 +1,43 @@
...
+		uintptr(unsafe.Pointer(&size)),
+		0,
+		0,
+	)
+
+	if errno != 0 {
+		return false, fmt.Errorf("sysctl failed: %w", errno)
+	}
+
+	// Check if P_TRACED flag is set
```
aarzilli: Superfluous comment.

pkg/debugdetect/detect_freebsd.go
```diff
@@ -0,0 +1,85 @@
...
+	if errProcfs == nil {
+		return attachedProcfs, nil
+	}
+
+	// Both methods failed
+	return false, fmt.Errorf("sysctl failed: %v; procfs fallback failed: %v", err, errProcfs)
+}
+
+func detectFreeBSDSysctl() (bool, error) {
+	// Use kinfo_getproc to get process information
```
aarzilli: Superfluous comment.

pkg/debugdetect/detect_freebsd.go
```diff
@@ -0,0 +1,85 @@
...
+	kp, err := C.kinfo_getproc(C.int(os.Getpid()))
+	if err != nil {
+		return false, fmt.Errorf("kinfo_getproc failed: %v", err)
+	}
+	defer C.free(unsafe.Pointer(kp))
+
+	// Check if P_TRACED flag is set in ki_flag
```
aarzilli: Superfluous comment.

pkg/debugdetect/detect_freebsd.go
```diff
@@ -0,0 +1,85 @@
...
+	// ki_flag contains the process flags including P_TRACED
+	return (int(kp.ki_flag) & pTracedFlag) != 0, nil
+}
+
+func detectFreeBSDProcfs() (bool, error) {
+	// Try reading /proc/curproc/status (similar to Linux)
```
aarzilli: Does this really exist on FreeBSD? The [procfs manpage](https://man.freebsd.org/cgi/man.cgi?query=procfs&apropos=0&sektion=0&manpath=FreeBSD+16.0-CURRENT&format=html) says the status file is a single line.

pkg/debugdetect/detect_test.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Fatalf("fixture failed: %v\n%s", err, out)
+	}
+
+	output := string(out)
+	if !strings.Contains(output, "NOT_ATTACHED") {
+		t.Errorf("expected 'NOT_ATTACHED' in output, got: %s", output)
+	}
+}
+
+func TestIsDebuggerAttached_NotAttached(t *testing.T) {
```
aarzilli: This one looks redundant to me with TestIntegration_NotAttached. I'd only keep one.

pkg/debugdetect/doc.go
```diff
@@ -0,0 +1,26 @@
+// Package debugdetect provides utilities for detecting if a program
+// is running under a debugger.
+//
+// This is useful for applications that need to modify their behavior
+// when being debugged, such as TUI applications that need to wait for
+// debugger attachment before continuing startup.
+//
+// Example usage:
+//
+//	attached, err := debugdetect.IsDebuggerAttached()
```
aarzilli, line 10: Maybe we should make this an helper function.

pkg/debugdetect/detect_test.go
derekparker: Agreed, removed the unit test kept the integration.

pkg/debugdetect/detect_freebsd.go
derekparker: Good catch, I should have caught this when looking through the freebsd procfs documentation. I ended up just removing the fallback since the usage of procfs on freebsd is deprecated anyway.

pkg/debugdetect/doc.go
derekparker: I thought about that but it seems easy enough for an end user to implement and there's too many potential variables for that the API should look like. Should it accept a context? Should it accept a timeout param? How opinionated should it be? Seems like users writing a solution on top of our building block thats suitable for their use case is better. Happy to discuss further though.

aarzilli: I'd say just the basic loop:

```
for {
    attached, err := IsDebuggerAttached()
    if attached || err != nil {
        return err
    }
    time.Sleep(500 * time.Millisecond)
}
```

anything different written by the user using IsDebuggerAttached.

PR #4262, opened 2026-02-19
pkg/terminal/starbind: remove some duplicate code

aarzilli:
introduce helper type starlarkUnhashable to avoid several duplicate
method definitions.


derekparker (approved):
LGTM

PR #4263, opened 2026-02-20
proc: support debugging stripped non-Go binaries

derekparker:
Add support for debugging non-Go stub binaries that dynamically load Go shared libraries via dlopen or other dynamic linkage modes. When the initial binary has no Go runtime symbols or DWARF info, Delve now marks it as non-Go and defers Go-specific initialization instead of exiting. A breakpoint on the dynamic linker's _dl_debug_state function detects shared library loads, allowing Delve to pick up Go debug info when a Go .so is loaded at runtime.

Should allow for use cases such as https://github.com/golang/go/issues/77484.

### DIFF
--- a/pkg/proc/target.go
+++ b/pkg/proc/target.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/go-delve/delve/pkg/dwarf/op"
 	"github.com/go-delve/delve/pkg/goversion"
@@ -79,6 +80,10 @@ type Target struct {
 	fakeMemoryRegistryMap map[string]*compositeMemory
 
 	partOfGroup bool
+
+	// onInitialGoImage ensures that Go-specific breakpoints are only set up
+	// once when transitioning from no Go images to having a Go image.
+	onInitialGoImage sync.Once
 }
 
 type KeepSteppingBreakpoints uint8
@@ -452,14 +457,18 @@ func (t *Target) sharedLibCallback(th Thread, tgt *Target) (bool, error) {
 		return false, nil
 	}
 
-	logger := logflags.DebuggerLogger()
-	logger.Info("Go shared library detected, setting up Go-specific breakpoints")
+	didRun := false
+	tgt.onInitialGoImage.Do(func() {
+		didRun = true
+		logger := logflags.DebuggerLogger()
+		logger.Info("Go shared library detected, setting up Go-specific breakpoints")
 
-	tgt.createUnrecoveredPanicBreakpoint()
-	tgt.createFatalThrowBreakpoint()
-	tgt.createPluginOpenBreakpoint()
+		tgt.createUnrecoveredPanicBreakpoint()
+		tgt.createFatalThrowBreakpoint()
+		tgt.createPluginOpenBreakpoint()
+	})
 
-	return true, nil
+	return didRun, nil
 }
 
 // CurrentThread returns the currently selected thread which will be used
@@ -677,8 +686,7 @@ func (lbp *LogicalBreakpoint) isSuspendedOnGroup(grp *TargetGroup) bool {
 	return true
 }
 
-type dummyRecordingManipulation struct {
-}
+type dummyRecordingManipulation struct{}
 
 // Recorded always returns false for the native proc backend.
 func (*dummyRecordingManipulation) Recorded() (bool, string) { return false, "" }


derekparker:
@Pawloland are you able to confirm if this enables your use case?

aarzilli:
The test failures look real.

pkg/proc/linutil/auxv.go
```diff
@@ -7,9 +7,34 @@ import (
 
 const (
 	_AT_NULL  = 0
+	_AT_BASE  = 7
 	_AT_ENTRY = 9
 )
 
+// BaseFromAuxv searches the elf auxiliary vector for the AT_BASE value,
```
aarzilli: I'd say we should deduplicate this by making BaseFromAuxv and EntryPointFromAuxv both call a `searchAuxv(int)` function.

pkg/proc/bininfo.go
```diff
@@ -1023,6 +1023,10 @@ type Image struct {
 	symTable *gosym.Table
 	Trimpath bool // trimpath used
 
+	// IsNonGo is true when the image has been confirmed to not be a Go binary.
+	// This is set when both DWARF loading and Go runtime symbol table loading fail.
+	IsNonGo bool
```
aarzilli: Negative flags are always hard to reason about, I'd change this to IsGo.

pkg/proc/bininfo.go
```diff
@@ -1052,6 +1056,17 @@ func (image *Image) Stripped() bool {
 	return image.dwarf == nil
 }
 
+// HasGoImage returns true if any loaded image is a Go binary with
+// successfully loaded debug info.
+func (bi *BinaryInfo) HasGoImage() bool {
```
aarzilli, line 1061: If I'm not mistaken this doesn't need to be exported.

pkg/proc/bininfo.go
```diff
@@ -1709,6 +1749,14 @@ func loadBinaryInfoElf(bi *BinaryInfo, image *Image, path string, addr uint64, w
 		}
 	}
 
+	// Mark images that have DWARF but no Go producer string as non-Go.
+	// Without this check, shared libraries that have DWARF debug info would
+	// be incorrectly treated as Go images by HasGoImage(), since IsNonGo is
+	// only set when both DWARF and Go symbol table loading fail.
+	if !hasGoProducer(image.dwarf) {
```
aarzilli: We should do this as part of loadDebugInfoMaps instead of reparsing the whole debug_info section twice, which is quite slow.

pkg/terminal/command.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/cosiner/argv"
 	"github.com/go-delve/delve/pkg/config"
 	"github.com/go-delve/delve/pkg/locspec"
+	"github.com/go-delve/delve/pkg/proc"
```
aarzilli: Importing proc directly from pkg/terminal is weird.

derekparker: Yeah I agree, however I did it to try and deduplicate the string in multiple places. I'll rework this.

pkg/proc/bininfo.go
derekparker: Ack, makes sense I'll refactor.

pkg/proc/linutil/auxv.go
derekparker: Done, extracted a `searchAuxv` helper that both `BaseFromAuxv` and `EntryPointFromAuxv` call.

pkg/proc/bininfo.go
derekparker: Done, renamed to `IsGo` with inverted semantics.

pkg/proc/bininfo.go
derekparker: `HasGoImage` is used from `proc_test.go` which is `package proc_test` (external test package), so it needs to be exported. Kept as-is.

pkg/proc/bininfo.go
derekparker: Done, removed the separate `hasGoProducer` pass. `image.IsGo` is now set inside `loadDebugInfoMaps` when a compile unit has `DW_AT_language == DW_LANG_Go`, which is already checked there for `cu.isgo`.

pkg/terminal/command.go
derekparker: Done, removed the `proc` import. Using the string literal `"shared library loaded"` directly.

derekparker:
Ended up removing the failing DAP test as I don't think it's relevant. In the container we can attach to PID 1 and with this change it won't fail if it's non-Go. Locally it passes because when not in container the user typically won't have privileges to attach to PID 1.

Used claude code for review feedback, so updated the CLAUDE.md as well with a hint.

Pawloland:
> @Pawloland are you able to confirm if this enables your use case?

I am positive, that the behavior changed from the previous one as the dlv in server mode now waits for the dlv in client mode for connections instead of closing immediately. As for debugging, it is still not working.  I will explain more in https://github.com/golang/go/issues/77484#issuecomment-3937069465.

pkg/proc/target.go
```diff
@@ -419,6 +428,40 @@ func (t *Target) createPluginOpenBreakpoint() {
...
+	}
+	bp, err := t.SetBreakpoint(0, rBrkAddr, SharedLibBreakpoint, nil)
+	if err != nil {
+		t.BinInfo().logger.Errorf("could not set shared lib breakpoint at %#x: %v", rBrkAddr, err)
+		return
+	}
+	bp.Breaklets[len(bp.Breaklets)-1].callback = t.sharedLibCallback
+}
+
+func (t *Target) sharedLibCallback(th Thread, tgt *Target) (bool, error) {
```
aarzilli, line 452: The code in this callback should only do something when we transition from not having any go image to having a go image, after the first go image is opened it will do its work every time another shared object is openend. Maybe calling createUnrecoveredPanicBreakpoint multiple times does nothing bad, but I'm not sure. It still seems wrong, though.

aarzilli: Thinking more about this: if this function happened inside AddImage, it could easily check if the current image is the first Go image and only do the work then. Implementing it like that would mean that the callback function could simply return false (or maybe there would be no callback function at all and the SharedLibrary breakpoint could be hardcoded to not trigger in checkCond.

derekparker: Agreed, updated.

aarzilli (approved):
LGTM, don't know if you want to merge it like this or also address further problems from the upstream issue.

derekparker:
> LGTM, don't know if you want to merge it like this or also address further problems from the upstream issue.

The other issues upstream seem to be related to path issues, so maybe lacking proper substitute-path settings?

PR #4270, opened 2026-02-28
pkg/proc/debuginfod: fix debuginfod download progress

aarzilli:
debuginfod-find added a --verbose flag and in the process broke the old
way to get download progress information:

https://sourceware.org/bugzilla/show_bug.cgi?id=33937

Restore this functionality.

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -141,6 +141,9 @@ func New(client service.Client, conf *config.Config) *Term {
 				t.downloadsMu.Unlock()
 				if !firstEventBinaryInfoDownload {
 					fmt.Fprintf(t.stdout, "\r")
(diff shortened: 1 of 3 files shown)


pkg/terminal/terminal.go
```diff
@@ -141,6 +141,9 @@ func New(client service.Client, conf *config.Config) *Term {
 				t.downloadsMu.Unlock()
 				if !firstEventBinaryInfoDownload {
 					fmt.Fprintf(t.stdout, "\r")
+					if strings.ToLower(os.Getenv("TERM")) != "dumb" {
+						fmt.Fprintf(t.stdout, "\x1b[J")
```
derekparker: Please add some a comment explaining this escape code.

aarzilli: Done.

PR #4274, opened 2026-03-04
pkg/proc: add GOEXPERIMENT=mapsplitgroup support

prattmic:
https://go.dev/cl/711560 adds GOEXPERIMENT=mapsplitgroup which changes the layout of map groups from KVKVKVKV to KKKKVVVV.

This GOEXPERIMENT may or may not be enabled permanently, depending on its performance impact, but I needed delve support for some debugging, so I went ahead and implemented support.

In the map iterator we determine which layout the map is using based on the presence of keys and elems fields in the group type instead of a slots field.

### DIFF
--- a/pkg/proc/mapiter.go
+++ b/pkg/proc/mapiter.go
@@ -322,7 +322,32 @@ type mapIteratorSwiss struct {
 	keyTypeIsPtr, elemTypeIsPtr bool
 	tableType, groupType        *godwarf.StructType
 
-	tableFieldIndex, tableFieldGroups, groupsFieldLengthMask, groupsFieldData, groupFieldCtrl, groupFieldSlots, slotFieldKey, slotFieldElem *godwarf.StructField
+	tableFieldIndex, tableFieldGroups, groupsFieldLengthMask, groupsFieldData, groupFieldCtrl *godwarf.StructField
+
+	// GOEXPERIMENT=nomapsplitgroup
+	//
+	// Interleaved slot layout (KVKVKVKV):
+	//
+	// type group struct {
+	//     ctrl  uint64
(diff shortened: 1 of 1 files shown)


prattmic:
I'm not sure that you want to merge this yet, but I needed it for testing, so I figured I might as well send it.

cc @jakebailey

pkg/proc/mapiter.go
```diff
@@ -482,23 +530,39 @@ func (it *mapIteratorSwiss) next() bool {
...
-				cur, err := it.group.slots.sliceAccess(int(it.slotIdx))
-				if err != nil {
-					it.v.Unreadable = fmt.Errorf("error accessing swiss map in table %d, group %d, slot %d", it.dirIdx, it.groupIdx, it.slotIdx)
-					return false
-				}
+				if it.mapsplitgroup {
+					var err1, err2 error
+					it.curKey, err1 = it.group.keys.sliceAccess(int(it.slotIdx))
+					it.curValue, err2 = it.group.elems.sliceAccess(int(it.slotIdx))
+					if err1 != nil || err2 != nil {
```
aarzilli: I'd move this outside of the if.

aarzilli (commented):
It would need a test after the CL gets merged, more or less a copy of TestClassicMap with just the `t.Setenv` line changed (or better put the code of TestClassicMap in an helper function and pass in the value for GOEXPERIMENT as argument)

aarzilli:
I see that the experiment was merged. Does this still match the golang implementation?

derekparker:
Closing in favor of https://github.com/go-delve/delve/pull/4370.

Thanks for the initial patch!

PR #4277, opened 2026-03-06
pkg/proc,service,pkg/terminal: add stacktrace caching

### LINKED ISSUE #989: Allow to load the full stacktrace

### LINKED ISSUE #3809: Navigating the stack in a coredump resulting from a stack overflow is slow
I was debugging a coredump produced by a stack overflow recently, and I noticed that Delve gets increasingly slower as I go up the stack. I suspect that stack navigation commands in Delve (`up`, `down` etc) could be linear in stack size. Does that sound right? 

I usually need to go to the top of the stack to understand what caused the stack overflow, and that easily takes minutes, because each step along the stack is slow (and is only getting slower).

I assume this should be fairly easy to reproduce, but let me know if you need more details.

---

   1.23.0

   go1.23-20240626-RC01

   Linux, x64

aarzilli:
Adds caching to stacktraces so that stacktrace requests are not
quadratic. Also add a Skip parameter to the Stacktrace RPC request so
that clients can only request the frames they need.

Fixes #3809
Fixes #989

### DIFF
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -111,6 +111,14 @@ func (frame *Stackframe) contains(off int64) bool {
 func ThreadStacktrace(tgt *Target, thread Thread, depth int) ([]Stackframe, error) {
 	g, _ := GetG(thread)
 	if g == nil {
+		if cachedStack := tgt.scache.get(0, thread.ThreadID()); cachedStack != nil {
+			frames, err := cachedStack.it.stacktrace(depth, cachedStack.frames)
+			if err != nil {
+				return nil, err
+			}
+			tgt.scache.put(0, thread.ThreadID(), cachedStack.it, frames)
+			return limitframes(frames, depth+1), nil
+		}
 		regs, err := thread.Registers()
 		if err != nil {
 			return nil, err
@@ -119,7 +127,13 @@ func ThreadStacktrace(tgt *Target, thread Thread, depth int) ([]Stackframe, erro
 		dwarfRegs := *(thread.BinInfo().Arch.RegistersToDwarfRegisters(so.StaticBase, regs))
 		dwarfRegs.ChangeFunc = thread.SetReg
 		it := newStackIterator(tgt, thread.BinInfo(), thread.ProcessMemory(), dwarfRegs, 0, nil, 0)
-		return it.stacktrace(depth)
+		frames, err := it.stacktrace(depth, nil)
+		if err != nil {
+			return nil, err
+		}
+		tgt.scache.put(0, thread.ThreadID(), it, frames)
+		return frames, nil
+
 	}
 	return GoroutineStacktrace(tgt, g, depth, 0)
 }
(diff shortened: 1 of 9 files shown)


aarzilli:
The test failure on darwin/amd64 looks like it could be related, I'll have to look into it first.

pkg/proc/stack.go
```diff
@@ -165,17 +179,30 @@ const (
 // GoroutineStacktrace returns the stack trace for a goroutine.
 // Note the locations in the array are return addresses not call addresses.
 func GoroutineStacktrace(tgt *Target, g *G, depth int, opts StacktraceOptions) ([]Stackframe, error) {
+	if opts == 0 {
```
derekparker, line 182: Would it make sense to include `opts` in the cache key to be able to cache these calls as well?

pkg/proc/stack.go
```diff
@@ -385,11 +412,16 @@ func (it *stackIterator) newStackframe(ret, retaddr uint64) Stackframe {
...
 
-func (it *stackIterator) stacktrace(depth int) ([]Stackframe, error) {
+func (it *stackIterator) stacktrace(depth int, initialFrames []Stackframe) ([]Stackframe, error) {
 	if depth < 0 {
 		return nil, errors.New("negative maximum stack depth")
 	}
-	frames := make([]Stackframe, 0, depth+1)
+	var frames []Stackframe
+	if len(initialFrames) > 0 {
+		frames = initialFrames
```
derekparker, line 425: Couldn't we return early here if `len(frames) >= depth`? I know that `it.stacktraceFunc` will return pretty quickly in this case, so maybe it's not worth it here.

pkg/proc/stack.go
aarzilli: The only way where we do a stacktrace with non-default options is if a user asks for a stacktrace at the command line by passing a bunch of arguments to the `stack` command. I don't think we would get to reuse that cache, so I don't think it's worth it.

pkg/proc/stack.go
aarzilli: Done.

pkg/proc/stack.go
derekparker: Ack.

pkg/proc/stack.go
derekparker: We should slice `frames[:depth+1]` before returning, no?

aarzilli: No, if we did the callers of `stacktrace` would cache the truncated version. The contract of the stacktrace functions was always that they could return more than the requested depth, because of inline expansion, so everything should be fine.

The truncation to the requested depth is done by the rpc call, when it matters (sending more would be expensive). DAP already did its own truncation before so nothing to change there.

PR #4278, opened 2026-03-06
build: add goreleaser release automation with cosign signing

### LINKED ISSUE #2241: Suggestions for release precompiled binary

4. What did you do?
6. What did you see instead?

Thanks for this great debugger, I use a lot in my golang projects. And I wonder whether it's possible to release precompiled binary? Because recently I try to debug the compiled binary in my vm that wasn't installed golang before.
And I had to install golang, then install delve.

Thanks!

derekparker:
Adds goreleaser-based release automation integrated with GitHub Releases.

On pushing a v* tag, the release workflow cross-compiles dlv for 8 platforms (linux/darwin/windows × amd64/arm64/386/ppc64le where supported), packages them as .tar.gz/.zip archives, generates a SHA256 checksums file, signs checksums.txt with cosign keyless signing via GitHub Actions OIDC, and publishes everything to GitHub Releases.

A workflow_dispatch trigger supports dry-run testing without publishing.

Changes:
- .goreleaser.yaml: goreleaser v2 config for builds, archives, checksums, cosign signing, and GitHub Release publishing
- .github/workflows/release.yml: release workflow with tag push and manual dispatch triggers
- _scripts/make.go: make build now delegates to goreleaser --single-target so local and release builds share identical config
- Documentation/installation/README.md: document binary release download and cosign verification
- CONTRIBUTING.md: document goreleaser v2 as a build prerequisite

Fixes #2241

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,64 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run only (snapshot build, not published to GitHub Releases)'
+        type: boolean
+        default: true
+
+# Cancel in-progress runs for the same ref.
+concurrency:
+  group: "${{ github.ref }}-release"
+  cancel-in-progress: false
+
+permissions:
+  contents: write   # create GitHub Release and upload assets
+  id-token: write   # cosign OIDC for keyless signing
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # full history required for goreleaser release notes
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3
+
+      - name: Run goreleaser (release)
+        if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && inputs.dry_run == false)
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: ~> v2
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run goreleaser (snapshot)
+        if: github.event_name == 'workflow_dispatch' && inputs.dry_run == true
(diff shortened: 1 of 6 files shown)


ofek:
Thank you so much!

loeffel-io:
Yes, thanks! Great news

_scripts/make.go
```diff
@@ -23,9 +23,6 @@ var NOTimeout bool
 var TestIncludePIE bool
 var TestSet, TestRegex, TestBackend, TestBuildMode string
 var Tags *[]string
-var Architecture string
```
aarzilli, line 26: Why remove those?

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # full history required for goreleaser release notes
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
```
aarzilli, line 35: We usually release when the new version of go is in rc1, using the stable version here means we build we the wrong version of go when we do the .0 release. However we also don't want to build with tip when we do >1.X.0 releases. What do we do?

_scripts/make.go
derekparker: Since `goreleaser` is handling cross-compilation, I just wanted to simplify the script.

.github/workflows/release.yml
derekparker: Good catch. We can specify a version in this field, but looking at how it resolves that via a known manifest, I don't think we're able to specify an RC version. However, the only time this would bite us is if we relied on an update to the standard library such as DWARF parsing or something to debug the newer release, which if that's the case we can temporarily vendor as we've done before. I think also if this becomes an issue in practice I believe we always have the escape hatch to be able to define our own custom step to download whatever version we require.

_scripts/make.go
aarzilli: I don't know, people might have scripts that rely on those. Maybe even reasons not to use goreleaser. I'd keep them.

.github/workflows/release.yml
aarzilli: I think there was only a couple of times in 10 years where we needed to build with a specific version go. Let's leave it as it is and see.

_scripts/make.go
derekparker: Merged the changes together to give the option to use the old flags if needed.

PR #4280, opened 2026-03-09
service/dap: set seq number for response messages

### LINKED ISSUE #2245: service/dap: set `seq` field in responses

aarzilli:
The DAP specification says that messages sent from the server to the
client should have a seq number but we never did do that because the
original DAP extension didn't.

Add sequence numbers, even though it doesn't seem to cause problems in
practice, to conform to the protocol specification.

Fixes #2245

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -31,6 +31,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-delve/delve/pkg/gobuild"
@@ -230,6 +231,8 @@ type Session struct {
 
 	// preTerminatedWG the WaitGroup that needs to wait before sending a terminated event.
 	preTerminatedWG sync.WaitGroup
+
+	seq atomic.Uint64
 }
 
 // Config is all the information needed to start the debugger, handle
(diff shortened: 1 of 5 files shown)


service/dap/server.go
```diff
@@ -4002,10 +4003,17 @@ func (s *Session) sendNotYetImplementedErrorResponse(request dap.Request) {
 		fmt.Sprintf("cannot process %q request", request.Command))
 }
 
-func newResponse(request dap.Request) *dap.Response {
+func (s *Session) getResponseSeq() int {
```
derekparker, line 4006: Mutex seems overkill here, couldn't we make seq a `atomic.Uint64` and then this becomes `return s.seq.Add(1)`.

aarzilli: Done.

PR #4282, opened 2026-03-09
config,terminal: better documentation for configuration file

aarzilli:
Provide an online documentation for the configuration file parameters
and the 'config' command.

Until now the configuration parameters were only documented in the
default configuration file and in the source code, neither of which is
very discoverable.


aarzilli:
Done.

PR #4283, opened 2026-03-09
build: allow release workflow to be manually triggered against a tag

derekparker:
Add a 'tag' input to the workflow_dispatch trigger so releases can be re-run or initiated manually for a specific tag. The checkout step uses the provided tag, and GORELEASER_CURRENT_TAG is set so goreleaser picks up the correct version. A validation step ensures non-dry-run releases require a tag.


ofek:
Is it possible to use this now to publish artifacts for the previous release?

PR #4284, opened 2026-03-12
pkg/terminal,pkg/config: add ability to configure Delve's prompt

### LINKED ISSUE #4001: Is it possible to change the default prompt `(dlv) `?
I want to customize the default prompt `(dlv) ` like `$PS1` in shell.

For example, I want to add newlines around the default prompt for better readability.

aarzilli:
Fixes #4001

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -116,6 +116,10 @@ type Config struct {
 	// TraceShowTimestamp controls whether to show timestamp in the trace
 	// output.
 	TraceShowTimestamp bool `yaml:"trace-show-timestamp"`
+
+	// Prompt is the string printed before each command. If empty, the
+	// default prompt "(dlv) " is used.
+	Prompt string `yaml:"prompt,omitempty"`
 }
 
 var Documentation = map[string]string{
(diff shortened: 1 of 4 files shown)


pkg/terminal/terminal.go
```diff
@@ -697,6 +703,92 @@ func (t *Term) goVersion() *goversion.GoVersion {
...
+func (t *Term) expandPrompt(prompt string) string {
+	const escape = '$'
+	var malformedEscape = []byte("<malformed escape>")
+
+	var hasState bool
+	var pid, tid int
+	var goid int64
+
+	getState := func() {
+		if hasState {
```
derekparker, line 715: We could wrap all this in a `sync.Once` but I don't feel strongly about it.

pkg/config/config.go
```diff
@@ -375,6 +389,15 @@ debug-info-directories: ["/usr/lib/debug/.build-id"]
 #  - 'disassembly' always show disassembly
 #  - 'default' show disassembly after step-instruction, source otherwise
 # position default
+
+# Uncomment to change Delve's default prompt.
+# The prompt string can contain various escape codes:
```
derekparker, line 394: This is awesome!

pkg/terminal/terminal.go
aarzilli: I think sync.Once is overkill for a local variable that doesn't need synchronization.

PR #4285, opened 2026-03-12
proc/internal/ebpf,build: extend eBPF tracing type support

derekparker:
Add support for additional Go types in the eBPF tracing backend:

- Bool: use BoolType with ByteSize 1 instead of IntType with ByteSize 8
- Int/Uint: use actual DWARF size instead of hardcoded ByteSize 8
- Float32/Float64: synthesize FloatType, detect XMM registers as unreadable
- Complex64/Complex128: synthesize ComplexType with XMM detection
- Pointer: capture pointer address and dereference pointed-to value via
  bpf_probe_read_user in eBPF C code
- Slice: capture slice header (ptr+len+cap) and dereference element data
- Graceful degradation: map, chan, interface, func, struct, and array
  types now produce specific "not yet supported" errors instead of
  generic "type not supported by ebpf"

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -1326,10 +1326,9 @@ func TestTraceEBPFTypes(t *testing.T) {
 	dlvbin := protest.GetDlvBinaryEBPF(t)
 	fixtures := protest.FindFixturesDir()
 
-	// Test small integer types (int8, uint16, int32) to verify correct byte sizes
-	t.Run("SmallInts", func(t *testing.T) {
-		t.Parallel()
-		cmd := exec.Command(dlvbin, "trace", "--ebpf", "--output", filepath.Join(t.TempDir(), "__debug"), filepath.Join(fixtures, "ebpf_trace_types.go"), "main.tracedSmallInts")
+	runEBPFTrace := func(t *testing.T, funcName string) []byte {
+		t.Helper()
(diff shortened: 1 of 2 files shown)


cmd/dlv/dlv_test.go
```diff
@@ -1303,6 +1304,96 @@ func TestTraceBackendParity(t *testing.T) {
...
+	}
+	if usr.Uid != "0" {
+		t.Skip("test must be run as root")
+	}
+
+	dlvbin := protest.GetDlvBinaryEBPF(t)
+	fixtures := protest.FindFixturesDir()
+
+	// Test small integer types (int8, uint16, int32) to verify correct byte sizes
+	t.Run("SmallInts", func(t *testing.T) {
```
aarzilli: IMO these could all be a single test run that checks for all three. Otherwise the common code should be in a helper function, at least.

pkg/proc/internal/ebpf/helpers.go
```diff
@@ -245,6 +284,23 @@ func parseFunctionParameterList(rawParamBytes []byte) RawUProbeParams {
...
+// registers, which are not accessible from eBPF uprobes.
+// On amd64, XMM0-XMM15 correspond to DWARF register numbers 17-32.
+// This file is only built for amd64 (see build tag), so the threshold
+// is hardcoded to 17.
+func usesXMMRegisters(param function_parameter_t) bool {
+	if !param.in_reg {
+		return false
+	}
+	for i := 0; i < int(param.n_pieces); i++ {
+		if param.reg_nums[i] >= 17 {
```
aarzilli: regnum.AMD64_XMM0 instead of 17

PR #4288, opened 2026-03-13
proc: implement frame pointer unwinding

alexsaezm:
Add hybrid stack unwinding that attempts frame-pointer-based unwinding before falling back to DWARF. Also adds BenchmarkStacktrace

Only affects AMD64 and ARM64. I tried to make it work on PPC64LE without luck. But in those two architectures, the improvement is noticeable. A reduction of almost ~50%. From ~20k allocs/op to 10k allocs/op.

I would like to add that this patch was mostly possible thanks to @derekparker. I've been trying to attempt this for a few weeks without any sucess, until he gave me the original patch this is based on. I mostly went over it to add ARM64 specifics because before that only AMD64 was using frame-pointer.


aarzilli:
Delve is already supposed to be doing frame pointer chasing. See amd64_arch.go:

```
		// When there's no frame descriptor entry use BP (the frame pointer) instead
		// - return register is [bp + a.PtrSize()] (i.e. [cfa-a.PtrSize()])
		// - cfa is bp + a.PtrSize()*2
		// - bp is [bp] (i.e. [cfa-a.PtrSize()*2])
		// - sp is cfa

		// When the signal handler runs it will move the execution to the signal
		// handling stack (installed using the sigaltstack system call).
		// This isn't a proper stack switch: the pointer to g in TLS will still
		// refer to whatever g was executing on that thread before the signal was
		// received.
		// Since go did not execute a stack switch the previous value of sp, pc
		// and bp is not saved inside g.sched, as it normally would.
		// The only way to recover is to either read sp/pc from the signal context
		// parameter (the ucontext_t* parameter) or to unconditionally follow the
		// frame pointer when we get to runtime.sigreturn (which is what we do
		// here).
```

and

```
	// We assume that RBP is the frame pointer, and we want to keep it updated,
	// so that we can use it to unwind the stack even when we encounter frames
	// without descriptor entries.
	// If there isn't a rule already we emit one.
```

Does it not work anymore?

I'm not surprised that frame pointer chasing is faster than dwarf, we're talking dereferencing a pointer vs running the bytecode for a general purpose language. However just doing that, or doing that first and then looking at dwarf, is not the correct thing to do.

derekparker:
> Delve is already supposed to be doing frame pointer chasing. See amd64_arch.go:
> 
> ```
> 		// When there's no frame descriptor entry use BP (the frame pointer) instead
> 		// - return register is [bp + a.PtrSize()] (i.e. [cfa-a.PtrSize()])
> 		// - cfa is bp + a.PtrSize()*2
> 		// - bp is [bp] (i.e. [cfa-a.PtrSize()*2])
> 		// - sp is cfa
> 
> 		// When the signal handler runs it will move the execution to the signal
> 		// handling stack (installed using the sigaltstack system call).
> 		// This isn't a proper stack switch: the pointer to g in TLS will still
> 		// refer to whatever g was executing on that thread before the signal was
> 		// received.
> 		// Since go did not execute a stack switch the previous value of sp, pc
> 		// and bp is not saved inside g.sched, as it normally would.
> 		// The only way to recover is to either read sp/pc from the signal context
> 		// parameter (the ucontext_t* parameter) or to unconditionally follow the
> 		// frame pointer when we get to runtime.sigreturn (which is what we do
> 		// here).
> ```
> 
> and
> 
> ```
> 	// We assume that RBP is the frame pointer, and we want to keep it updated,
> 	// so that we can use it to unwind the stack even when we encounter frames
> 	// without descriptor entries.
> 	// If there isn't a rule already we emit one.
> ```
> 
> Does it not work anymore?

No this still works, and we should reuse more of this.

> 
> I'm not surprised that frame pointer chasing is faster than dwarf, we're talking dereferencing a pointer vs running the bytecode for a general purpose language. However just doing that, or doing that first and then looking at dwarf, is not the correct thing to do.

This patch does a hybrid, similar to what the Go runtime does for the execution tracer, however we get more fidelity as we combine with DWARF or other methods when necessary (CGO transitions, frameless functions). The spirit of this patch is to try and prioritize the quick and efficient frame pointer based unwinding when possible, and fall back to DWARF or other methods when necessary. Essentially the opposite of what we do now, which is using frame pointer unwinding as the fallback. Even mentioned in the code you qouted:

> 		// When there's no frame descriptor entry use BP (the frame pointer) instead

Why fall back to the most efficient method instead of use it until we can't?

aarzilli:
If this is about performance, then we should only do it for go functions, where we know that it is a valid method. See the test failures too, they do look related.

pkg/proc/stack.go
```diff
@@ -537,15 +544,146 @@ func (it *stackIterator) appendInlineCalls(callback func(Stackframe) bool, frame
...
+			}
+		}
+	}
+	return callFrameRegs, ret, retaddr
+}
+
+func (it *stackIterator) tryFramePointerUnwind() (callFrameRegs op.DwarfRegisters, ret uint64, retaddr uint64, ok bool) {
+	// Only enable frame pointer unwinding on tested architectures
+	if it.bi.Arch.Name != "amd64" && it.bi.Arch.Name != "arm64" {
+		// All other architectures (ppc64le, 386, riscv64, loong64) use DWARF only
```
aarzilli: This should say something along the lines that we don't implement this optimization. I think those other architectures have frame pointers as well (386 definitely does).

pkg/proc/stack.go
```diff
@@ -537,15 +544,146 @@ func (it *stackIterator) appendInlineCalls(callback func(Stackframe) bool, frame
...
+		"runtime.goexit", "runtime.goexit0", "runtime.goexit1",
+		"runtime.mstart", "runtime.mstart0", "runtime.mstart1",
+		"runtime.systemstack_switch",
+		"runtime.sigreturn", "runtime.sigtrampgo",
+		"runtime.sigpanic",
+		"crosscall2":
+		return op.DwarfRegisters{}, 0, 0, false
+	}
+
+	if it.useDWARF {
```
aarzilli: This is a free check so it should be done before looking up the function (maybe the first or second check).

pkg/proc/stack.go
```diff
@@ -537,15 +544,146 @@ func (it *stackIterator) appendInlineCalls(callback func(Stackframe) bool, frame
...
+
+	cfa := int64(bp + 2*ptrSize)
+
+	if ret == 0 || it.bi.PCToFunc(ret) == nil {
+		return op.DwarfRegisters{}, 0, 0, false
+	}
+
+	it.regs.CFA = cfa
+
+	callimage := it.bi.PCToImage(it.pc)
```
aarzilli: This should be `fn.cu.image` or `bi.funcToImage(fn)` to avoid searching for the function of `it.pc` twice.

pkg/proc/stack.go
```diff
@@ -243,6 +243,13 @@ type stackIterator struct {
 
 	count int
 
+	// useDWARF is true when the frame pointer is not yet stable for
+	// FP-based unwinding. This happens at the top of the stack where
+	// the topmost function may be frameless (BP inherited from caller).
+	// FP unwinding becomes safe when BP + 2*PtrSize == CFA, indicating
+	// that BP is the current frame's own frame pointer.
+	useDWARF bool
```
aarzilli: Maybe I don't understand but shouldn't this be named `canUseFramePointerUnwind` or something like that, more than `useDWARF`?

alexsaezm: I renamed it and changed the logic a little bit; I think now it is less confusing.

_fixtures/deeprecursion.go
```diff
@@ -0,0 +1,16 @@
+package main
```
aarzilli: This file doesn't seem to be used.

pkg/proc/stack.go
```diff
@@ -537,15 +544,145 @@ func (it *stackIterator) appendInlineCalls(callback func(Stackframe) bool, frame
...
 // advanceRegs calculates the DwarfRegisters for a next stack frame
-// (corresponding to it.pc).
-//
-// The computation uses the registers for the current stack frame (it.regs) and
-// the corresponding Frame Descriptor Entry (FDE) retrieved from the DWARF info.
-//
-// The new set of registers is returned. it.regs is not updated, except for
-// it.regs.CFA; the caller has to eventually switch it.regs when the iterator
-// advances to the next frame.
+// (corresponding to it.pc). It tries frame-pointer-based unwinding first
```
aarzilli: I think too much of the comment was deleted.

```
advanceRegs calculates the DwarfRegisters for the next stack frame (corresponding to it.pc).

The computation uses the registers for the current stack frame in it.regs. When possible a simple frame pointer based unwinding is done, otherwise the Frame Descriptor Entry (FDE) corresponding to the current frame, retrieved from DWARF, is used.

The new set of registers is returned. it.regs is not updated, except for it.regs.CFA; the caller has to eventually switch it.regs when the iterator advances to the next frame.
```

pkg/proc/stack.go
aarzilli: Still TBD.

pkg/proc/stack.go
```diff
@@ -537,15 +544,145 @@ func (it *stackIterator) appendInlineCalls(callback func(Stackframe) bool, frame
...
+		"runtime.cgocallback_gofunc", "runtime.cgocallback",
+		"runtime.goexit", "runtime.goexit0", "runtime.goexit1",
+		"runtime.mstart", "runtime.mstart0", "runtime.mstart1",
+		"runtime.systemstack_switch",
+		"runtime.sigreturn", "runtime.sigtrampgo",
+		"runtime.sigpanic",
+		"crosscall2":
+		return op.DwarfRegisters{}, 0, 0, false
+	}
+
```
aarzilli, line 646: Since we can not guarantee that the frame pointer chasing is correct for non-go code this should also check `fn.cu.isgo`.

pkg/proc/stack.go
alexsaezm: I already changed this to `callimage := it.bi.funcToImage(fn)` but for some reason GitHub doesn't show it right here in the discussion view :-/

_fixtures/deeprecursion.go
alexsaezm: Right, I used it at the beginning, but I totally forgot about it. I will delete it in the next commit.

pkg/proc/stack.go
alexsaezm: I added your suggestion, thanks.

pkg/proc/stack.go
alexsaezm: Thanks. I added the check.

aarzilli:
I believe the test failures, in TeamCity, are related to this change set.

alexsaezm:
> I believe the test failures, in TeamCity, are related to this change set.

It looks like it works with Go tip but no with 1.24 and 1.25. I disable the feature for Windows ARM with Go 1.24 and 1.25. I think it might be related to https://github.com/golang/go/issues/63630

Let's see if the tests pass because I don't have a Windows system to try locally.

Edit: I didn't see the other errors. I will work on them.

alexsaezm:
I think I properly fixed the TeamCity failures with the latest commit.

I'm confused with the Windows tests... I thought they were fixed, but now 1.25 fails?

derekparker:
> I think I properly fixed the TeamCity failures with the latest commit.
> 
> I'm confused with the Windows tests... I thought they were fixed, but now 1.25 fails?

We're having issues with Windows/arm64 lately with intermittent failures. Doesn't seem related to this change.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4289, opened 2026-03-14
dap/command: port examinemem from dlv/terminal

BrotifyPacha:
I find `examinemem` command  sometimes quite useful. As i mostly use delve through dap, i wanted to port it from dlv/terminal package.

My current implementation in this PR is as follows:

1. Extract the argument parsing from [pkg/terminal/command examineMemoryCmd()](https://github.com/go-delve/delve/blob/master/pkg/terminal/command.go#L2090) into a separate `func ParseExamineMemoryArg(out *ExamineMemoryArgs, argstr string) error`
2. Import `pkg/terminal` in `service/dap/command` and reuse `func ParseExamineMemoryArg`
3. Use `debug.ExamineMemory` and `api.PrettyExamineMemory` to achieve the same examine memory functionality in dap

But as soon as i got it working, i tried to write test, and got an import cycle error:

```
$ make test
package github.com/go-delve/delve/pkg/terminal
        imports github.com/go-delve/delve/service/rpccommon from command_test.go
        imports github.com/go-delve/delve/service/dap from server.go
        imports github.com/go-delve/delve/pkg/terminal from command.go: import cycle not allowed in test
make: *** [vet] Error 1
```

Basically the step 2, where i imported `pkg/terminal` in `service/dap/command` is the root cause of an import cycle.

I can see a few solutions:

1. Just copy the argument parsing implementation
2. Extract argument parsing out of `pkg/terminal` somewhere else, maybe `service/api`?

The first one would require of you guys to keep both functions in sync, which seems bad, so im more in favor of a second one, but im not yet accustomed to delve code base to make a judgement call of where to extract it to.

I'd like to go through with the porting of this command to dap, so im looking forward for your feedback.

### DIFF
--- a/pkg/terminal/command.go
+++ b/pkg/terminal/command.go
@@ -2088,101 +2088,15 @@ func watchpoint(t *Term, ctx callContext, args string) error {
 }
 
 func examineMemoryCmd(t *Term, ctx callContext, argstr string) error {
-	var (
-		address uint64
-		err     error
-		ok      bool
-		args    = strings.Split(argstr, " ")
-	)
-
-	// Default value
-	priFmt := byte('x')
-	count := int64(1)
-	size := int64(1)
-	isExpr := false
-	rawout := false
-
-	// nextArg returns the next argument that is not an empty string, if any, and
-	// advances the args slice to the position after that.
-	nextArg := func() string {
-		for len(args) > 0 {
-			arg := args[0]
-			args = args[1:]
-			if arg != "" {
-				return arg
-			}
-		}
-		return ""
+	args, err := api.ParseExamineMemoryArg(argstr)
+	if err != nil {
+		return err
 	}
 
-loop:
-	for {
-		switch cmd := nextArg(); cmd {
-		case "":
-			// no more arguments
-			break loop
-		case "-fmt":
-			arg := nextArg()
-			if arg == "" {
-				return errors.New("expected argument after -fmt")
-			}
-			if arg == "raw" {
-				rawout = true
-			} else {
-				fmtMapToPriFmt := map[string]byte{
-					"oct":         'o',
-					"octal":       'o',
-					"hex":         'x',
-					"hexadecimal": 'x',
-					"dec":         'd',
-					"decimal":     'd',
-					"bin":         'b',
-					"binary":      'b',
-				}
-				priFmt, ok = fmtMapToPriFmt[arg]
-				if !ok {
-					return fmt.Errorf("%q is not a valid format", arg)
-				}
-			}
-		case "-count", "-len":
-			arg := nextArg()
-			if arg == "" {
-				return errors.New("expected argument after -count/-len")
-			}
-			var err error
-			count, err = strconv.ParseInt(arg, 0, 64)
-			if err != nil || count <= 0 {
-				return errors.New("count/len must be a positive integer")
-			}
-		case "-size":
-			arg := nextArg()
-			if arg == "" {
-				return errors.New("expected argument after -size")
-			}
-			var err error
-			size, err = strconv.ParseInt(arg, 0, 64)
-			if err != nil || size <= 0 || size > 8 {
-				return errors.New("size must be a positive integer (<=8)")
-			}
-		case "-x":
-			isExpr = true
-			break loop // remaining args are going to be interpreted as expression
-		default:
-			if len(args) > 0 {
-				return fmt.Errorf("unknown option %q", args[0])
-			}
-			args = []string{cmd}
-			break loop // only one arg left to be evaluated as a uint
-		}
-	}
+	var address uint64
 
-	if len(args) == 0 {
-		return errors.New("no address specified")
-	}
-
-	if isExpr {
-		expr := strings.Join(args, " ")
-		val, err := t.client.EvalVariable(ctx.Scope, expr, t.loadConfig())
+	if args.IsExpr {
+		val, err := t.client.EvalVariable(ctx.Scope, args.Operand, t.loadConfig())
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


service/dap/command.go
```diff
@@ -133,6 +156,55 @@ func (s *Session) helpMessage(_, _ int, args string) (string, error) {
...
+			return fmt.Errorf("unsupported expression type: %s", val.Kind).Error(), nil
+		}
+	} else {
+		address, err = strconv.ParseUint(args.Operand, 0, 64)
+		if err != nil {
+			return fmt.Errorf("convert address into uintptr type failed, %s", err).Error(), nil
+		}
+	}
+
+	memory, err := s.debugger.ExamineMemory(address, int(args.Count*args.Size))
```
aarzilli: You should limit `args.Count*args.Size` to some maximum.

pkg/terminal/command.go
```diff
@@ -2089,18 +2098,92 @@ func watchpoint(t *Term, ctx callContext, args string) error {
...
+type ExamineMemoryArgs struct {
+	Operand string
+	Count   int64
+	Size    int64
+	IsExpr  bool
+	Format  byte
+	RawOut  bool
+}
+
+func ParseExamineMemoryArg(out *ExamineMemoryArgs, argstr string) error {
```
aarzilli: It should return ExamineMemoryArgs as an output parameter, rather than taking out as an argument.

pkg/terminal/command.go
```diff
@@ -2180,55 +2265,10 @@ loop:
...
-		if err != nil {
-			return fmt.Errorf("convert address into uintptr type failed, %s", err)
-		}
-	}
-
-	t.stdout.pw.PageMaybe(nil)
-
-	start := address
-	remsz := int(count * size)
+	out.Operand = args[0]
```
aarzilli: You are overriding the value set from the `"-x"` case.

service/dap/command.go
```diff
@@ -133,6 +156,55 @@ func (s *Session) helpMessage(_, _ int, args string) (string, error) {
 	return buf.String(), nil
 }
 
+func (s *Session) examineMemory(goid, frame int, argstr string) (string, error) {
+
```
aarzilli: Superfluous empty line.

pkg/terminal/command.go
```diff
@@ -2180,55 +2265,10 @@ loop:
...
-		if rawout {
-			t.stdout.Write(memArea)
-		} else {
-			fmt.Fprint(t.stdout, api.PrettyExamineMemory(uintptr(start), memArea, isLittleEndian, priFmt, int(size)))
-		}
-		start += uint64(reqsz)
-		remsz -= reqsz
-	}
 	return nil
+
```
aarzilli: Superfluous empty line.

BrotifyPacha: Fixed

service/dap/command.go
BrotifyPacha: Fixed

pkg/terminal/command.go
BrotifyPacha: Fixed

pkg/terminal/command.go
BrotifyPacha: Fixed

service/dap/command.go
BrotifyPacha: Fixed using `rpc2.ExamineMemoryLengthLimit`

BrotifyPacha:
> I think doing it this way is reasonable. However see the failure in TestEvaluateCommandRequest (I'd remove dlvHelp and just check that the response to `dlv help` is non-empty). And it needs a test

I fixed a dlv help test, and added basic test cases for examinemem, but the issue of cyclic imports still persists

BrotifyPacha:
Tidied everything up according to comments and wrote a proper test. To resolve the import cycle i extracted argument parsing into `service/api` package.

service/dap/server_test.go
```diff
@@ -4529,6 +4529,92 @@ func TestEvaluateCommandRequest(t *testing.T) {
...
+			func() {
+				client.LaunchRequest("exec", fixture.Path, !stopOnEntry)
+			},
+			fixture.Source, []int{}, // Breakpoint set in the program
+			[]onBreakpoint{{ // Stop at first breakpoint
+				execute: func() {
+					checkStop(t, client, 1, "main.foobar", []int{65, 66})
+
+					// get address of b1
+					client.EvaluateRequest("call uint64(&b1)", localsScope, "repl")
```
aarzilli: If you use `call` for this it won't work on platforms where function calls aren't supported, which is why this test is failing on 386.

service/dap/server_test.go
```diff
@@ -4529,6 +4529,92 @@ func TestEvaluateCommandRequest(t *testing.T) {
...
+	})
+}
+
+func mustEvaluateAddress(t *testing.T, got *dap.EvaluateResponse) int64 {
+	t.Helper()
+
+	var addressStr string
+	if parts := strings.Split(got.Body.Result, " "); len(parts) < 1 {
+		t.Fatalf("could not evaluate address: %q must contain a pointer", got.Body.Result)
+	} else {
```
aarzilli: No need for an else after a Fatalf.

service/dap/server_test.go
BrotifyPacha: Is it a good solution to exclude these tests on 386, or should i better rewrite them to be less exact, for example: just check that the output **contains** correct value?

service/dap/server_test.go
BrotifyPacha: Fixed

service/dap/server_test.go
BrotifyPacha: Looked around at other tests, and decided to make tests more generic.

BrotifyPacha:
@aarzilli Is there anything else i should do for this PR to be merged?

aarzilli:
Wait for @derekparker review.

PR #4295, opened 2026-03-24
proc: fix stack watch breaklet index for recorded replay

cuiweixie:
When adding the extra WatchOutOfScope breakpoint at the CALL before the return PC, the code used len(retbp.Breaklets) to index into retbp2.Breaklets. If the return address already had other breaklets (e.g. a user breakpoint), that index could be out of range.

Add TestWatchpointStackRecordedOutOfScopeBreakletIndex (runs with -backend=rr) to cover this path.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -5211,15 +5211,6 @@ func TestWatchpointStackRecordedOutOfScopeBreakletIndex(t *testing.T) {
 	// returns), retbp accumulates an extra breaklet while retbp2 does not, and
 	// using len(retbp)-1 to index retbp2 panics.
 	skipUnlessOn(t, "only for recorded targets", "rr")
-	skipOn(t, "not implemented", "freebsd")
-	skipOn(t, "not implemented", "386")
-	skipOn(t, "not implemented", "ppc64le")
-	skipOn(t, "not implemented", "riscv64")
-	skipOn(t, "not implemented", "loong64")
-	skipOn(t, "see https://github.com/go-delve/delve/issues/2768", "windows")
-	if _, isTeamCityTest := os.LookupEnv("TEAMCITY_VERSION"); isTeamCityTest {
-		skipOn(t, "CI is running a version of macOS that is too old (11.2)", "darwin", "arm64")
-	}
 	protest.AllowRecording(t)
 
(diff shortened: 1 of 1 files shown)


derekparker:
One question, why does this only manifest with RR? The stack watchpoint code isn't RR specific, so this could also be triggered without RR, correct?

cuiweixie:
> One question, why does this only manifest with RR? The stack watchpoint code isn't RR specific, so this could also be triggered without RR, correct?

It should be.But it is a bug obviously and I only test with rr.

pkg/proc/proc_test.go
```diff
@@ -5203,6 +5203,40 @@ func TestWatchpointStackBackwardsOutOfScope(t *testing.T) {
...
 
+func TestWatchpointStackRecordedOutOfScopeBreakletIndex(t *testing.T) {
+	// Regression: setStackWatchBreakpoints (recorded replay) adds a second
+	// WatchOutOfScope breakpoint at the CALL before the return PC. The last
+	// breaklet must be taken from retbp2, not indexed with len(retbp.Breaklets).
+	// If a user breakpoint already exists at the return PC (here: main after f()
+	// returns), retbp accumulates an extra breaklet while retbp2 does not, and
+	// using len(retbp)-1 to index retbp2 panics.
+	skipUnlessOn(t, "only for recorded targets", "rr")
+	skipOn(t, "not implemented", "freebsd")
```
aarzilli: These are all irrelevant, it's all already covered by the `skipUnlessOn(..., "rr")`

aarzilli:
> One question, why does this only manifest with RR? The stack watchpoint code isn't RR specific, so this could also be triggered without RR, correct?

It's in a branch that checks `t.recman.Recorded()` so effectively it only executes with the rr backend. This isn't the normal stack watchpoint, it's the one added to check out-of-stack while running backwards.

aarzilli:
I'm not sure how but this set of changes got mixed up with the examinemem PR.

cuiweixie:
> I'm not sure how but this set of changes got mixed up with the examinemem PR.

remove it. Sorry for that.

derekparker:
> > One question, why does this only manifest with RR? The stack watchpoint code isn't RR specific, so this could also be triggered without RR, correct?
> 
> It's in a branch that checks `t.recman.Recorded()` so effectively it only executes with the rr backend. This isn't the normal stack watchpoint, it's the one added to check out-of-stack while running backwards.

Ack, didn't notice that, thanks.

PR #4296, opened 2026-03-24
dwarf/op: fix DW_OP_shr and DW_OP_shra evaluation

cuiweixie:
DW_OP_shr must perform a logical right shift (zero-fill) and DW_OP_shra an arithmetic right shift (sign-extend). The two cases were swapped, so negative values were wrong for location expressions.

Add a regression test that exercises both opcodes with a negative value.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/go-delve/delve
 
-go 1.2
+go 1.24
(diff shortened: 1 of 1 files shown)


aarzilli (approved):
LGTM

PR #4306, opened 2026-04-13
pkg/proc: Fix flakiness in TestWaitFor

typesanitizer:
TestWaitFor could have the following sequence of events:
1. cmd.Start() is called.
2. The process is started in the kernel.
3. CreateToolhelp32Snapshot (inside waitForSearchProcess) sees
   the process and returns from native.WaitFor.
4. The test immediately deferences cmd.Process, but cmd.Process
   has not been assigned yet in cmd.Start() -> nil panic.

This patch fixes the issue by introducing a channel that is
written to from testWaitForSetup() _after_ cmd.Start() returns.
Unlike the mutex + boolean, this channel can be waited upon
without spinning, and only then cmd.Process is read inside
the test code.

### DIFF
--- a/pkg/proc/proc_test.go
+++ b/pkg/proc/proc_test.go
@@ -24,7 +24,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 	"testing"
 	"text/tabwriter"
 	"time"
@@ -5639,7 +5638,15 @@ func TestReadTargetArguments(t *testing.T) {
 	})
 }
 
-func testWaitForSetup(t *testing.T, mu *sync.Mutex, started *bool) (*exec.Cmd, *proc.WaitFor) {
+type waitForFixture struct {
+	t        *testing.T
+	cmd      *exec.Cmd
+	waitFor  *proc.WaitFor
+	startedc chan struct{}
+	startErr error
+}
+
+func testWaitForSetup(t *testing.T) *waitForFixture {
 	var buildFlags protest.BuildFlags
 	if buildMode == "pie" {
 		buildFlags |= protest.BuildModePIE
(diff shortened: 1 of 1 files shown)


pkg/proc/proc_test.go
```diff
@@ -5667,27 +5677,42 @@ func testWaitForSetup(t *testing.T, mu *sync.Mutex, started *bool) (*exec.Cmd, *
...
 func TestWaitFor(t *testing.T) {
 	skipOn(t, "waitfor implementation is delegated to debugserver", "darwin")
 	skipOn(t, "flaky", "freebsd")
 
-	var mu sync.Mutex
-	started := false
-
-	cmd, waitFor := testWaitForSetup(t, &mu, &started)
+	fixtureState := testWaitForSetup(t)
+	t.Cleanup(func() { fixtureState.cleanup() })
```
aarzilli: This can just be `t.Cleanup(fixtureState.cleanup)`

pkg/proc/proc_test.go
```diff
@@ -5667,27 +5677,42 @@ func testWaitForSetup(t *testing.T, mu *sync.Mutex, started *bool) (*exec.Cmd, *
 		// See: https://lldb.llvm.org/man/lldb.html#cmdoption-lldb-wait-for
 		waitFor.Name = fixture.Path
 	}
+	fixtureState.waitFor = waitFor
+
+	return fixtureState
+}
 
-	return cmd, waitFor
+func (fixtureState *waitForFixture) waitStarted(t *testing.T) {
```
aarzilli: Might as well save the `testing.T` inside `waitForFixture` instead of passing it again.

pkg/proc/proc_test.go
```diff
@@ -5703,38 +5728,29 @@ func TestWaitForAttach(t *testing.T) {
 		return
 	}
 
-	var mu sync.Mutex
-	started := false
-
-	cmd, waitFor := testWaitForSetup(t, &mu, &started)
+	fixtureState := testWaitForSetup(t)
+	t.Cleanup(func() { fixtureState.cleanup() })
```
aarzilli: Same.

typesanitizer: Fixed in https://github.com/go-delve/delve/pull/4306/commits/236ce9dd7842b07e5a284dfa7f8445b537bf10a1

pkg/proc/proc_test.go
typesanitizer: Fixed in https://github.com/go-delve/delve/pull/4306/commits/236ce9dd7842b07e5a284dfa7f8445b537bf10a1

pkg/proc/proc_test.go
typesanitizer: Fixed in https://github.com/go-delve/delve/pull/4306/commits/236ce9dd7842b07e5a284dfa7f8445b537bf10a1

aarzilli (approved):
LGTM

typesanitizer:
@aarzilli sorry, I don't have merge access, can you merge the PR for me? Is there something else you'd like me to do before merging?

derekparker (approved):
LGTM

PR #4307, opened 2026-04-14
cmd/dlv: use goversion.ParseProducer for DW_AT_producer

croepha:
~~For some reason, a prefix is added to the raw go version.  It may be caused by how I built go.  This simply removes the prefix.~~

Actually uses goversion.ParseProducer

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -699,7 +699,7 @@ func (t *Term) goVersion() *goversion.GoVersion {
 		return t.goVersionCache
 	}
 	vers := t.client.GetVersion()
-	v, _ := goversion.Parse(vers.TargetGoVersion)
+	v := goversion.ParseProducer(vers.TargetGoVersion)
 	t.goVersionCache = &v
 	return t.goVersionCache
 }


aarzilli:
>  It may be caused by how I built go

How do you build go?

aarzilli:
Looking around, the version string is generated by `findgoversion` in `src/cmd/dist/build.go` and the only way it could start with like that is a malformed `VERSION` file. You could also be talking about the version string in `DW_AT_producer`, which *always* starts with `Go cmd/compile`, but that's already taken care of by `goversion.ParseProducer`.

croepha:
it is `DW_AT_producer`, I'm hitting this branch:
https://github.com/go-delve/delve/blob/b26393a5eacdaf84a4f046adb9e2d47c206f7478/service/debugger/debugger.go#L2272-L2275

Then here we call `goversion.Parse` instead of `goversion.ParseProducer`
https://github.com/go-delve/delve/blob/b26393a5eacdaf84a4f046adb9e2d47c206f7478/pkg/terminal/terminal.go#L697-L705

croepha:
I rebooted the PR, with a patch to use ParseProducer instead of Parse

aarzilli (approved):
LGTM

croepha:
Do you know how to get this failing test to pass? its complaining about credits...

aarzilli:
Don't worry about it.

derekparker (approved):
LGTM

PR #4308, opened 2026-04-18
make interface type show addr for compare

runsys:


aarzilli:
You can print the address of a variable with `print &v`.

PR #4319, opened 2026-04-24
proc: fix SP calculation for sigpanic frames on arm64

### LINKED ISSUE #3591: arm64 core, the regs command get the wrong SP after nil pointer dereference
```
Delve Debugger
Version: 1.21.2
```

`go version go1.20.10 linux/arm64`

```
cat /etc/redhat-release
CentOS Linux release 7.9.2009 (AltArch)
```
4. What did you do?
```
package main

func main() {
	map_value := make(map[string]*int)

	var value_1 int = 1
	var value_2 int = 2
	map_value["a"] = &value_1
	map_value["b"] = &value_2

	mod_map(map_value)
}

func mod_map(map_value map[string]*int) {
	p_value := map_value["c"]
	*p_value = 3
}
```

```
step 1: Compile the demo.go with debug information (go build -gcflags=all="-N -l" demo.go)
step 2: export GOTRACEBACK=crash, then run the executable file. Now wen can get the coredump file and the information of the program.
step 3: Compare the information of the program and dlv coredump information.
```

When `dlv core demo core_demo_xx`，we expect to get the right SP for the coredump function.
**SP=0x4000192dc0 is printed for the function mod_map which contains nil pointer dereference.**
The following is the inforamtion of the program. The inforamtion contain the SP for each frame stack.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x68bc0]

goroutine 1 [running]:
panic({0x760c0, 0xd1300})
        /home/wsl/go/go1.20.10/src/runtime/panic.go:987 +0x354 fp=0x4000192d60 sp=0x4000192ca0 pc=0x3ee44
runtime.panicmem()
        /home/wsl/go/go1.20.10/src/runtime/panic.go:260 +0x48 fp=0x4000192d80 sp=0x4000192d60 pc=0x3da78
runtime.sigpanic()
        /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:841 +0x1e0 fp=0x4000192db0 sp=0x4000192d80 pc=0x533f0
main.mod_map(0x4000160668)
        /home/wsl/demo2/demo_2.go:16 +0x40 fp=0x4000192e00 sp=0x4000192dc0 pc=0x68bc0
main.main()
        /home/wsl/demo2/demo_2.go:11 +0x158 fp=0x4000192f70 sp=0x4000192e00 pc=0x68b68
runtime.main()
        /home/wsl/go/go1.20.10/src/runtime/proc.go:250 +0x1d4 fp=0x4000192fd0 sp=0x4000192f70 pc=0x41584
runtime.goexit()
        /home/wsl/go/go1.20.10/src/runtime/asm_arm64.s:1172 +0x4 fp=0x4000192fd0 sp=0x4000192fd0 pc=0x66584
```

6. What did you see instead?
Dlv get the wrong SP information for the function mod_map which contains nil pointer dereference.
**Dlv get he wrong SP = 0x4000192db0. The right SP is 0x4000192dc0.**
**We guess the wrong SP is the root cuase of the problem that the print commands for input parameter and local variable  is wrong.**

```
(dlv) bt
 0  0x00000000000673d8 in runtime.raise
    at /home/wsl/go/go1.20.10/src/runtime/sys_linux_arm64.s:158
 1  0x0000000000053474 in runtime.dieFromSignal
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:883
 2  0x000000000005395c in runtime.sigfwdgo
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:1096
 3  0x00000000000526ac in runtime.sigtrampgo
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:432
 4  0x00000000000673d8 in runtime.raise
    at /home/wsl/go/go1.20.10/src/runtime/sys_linux_arm64.s:157
 5  0x0000000000053474 in runtime.dieFromSignal
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:883
 6  0x00000000000535e4 in runtime.crash
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:975
 7  0x000000000003f5a0 in runtime.fatalpanic
    at /home/wsl/go/go1.20.10/src/runtime/panic.go:1172
 8  0x000000000003ee44 in runtime.gopanic
    at /home/wsl/go/go1.20.10/src/runtime/panic.go:987
 9  0x000000000003da78 in runtime.panicmem
    at /home/wsl/go/go1.20.10/src/runtime/panic.go:260
10  0x00000000000533f0 in runtime.sigpanic
    at /home/wsl/go/go1.20.10/src/runtime/signal_unix.go:841
11  0x0000000000068bc0 in main.mod_map
    at ./demo_2.go:15
12  0x0000000000068bc0 in main.mod_map
    at ./demo_2.go:15
13  0x0000000000068bb8 in main.mod_map
    at ./demo_2.go:15
14  0x0000000000000000 in ???
    at :0
    error: NULL address
(truncated)
(dlv) frame 11
> runtime.raise() /home/wsl/go/go1.20.10/src/runtime/sys_linux_arm64.s:158 (PC: 0x673d8)
Frame 11: ./demo_2.go:15 (PC: 68bc0)
    10:
    11:         mod_map(map_value)
    12: }
    13:
    14: func mod_map(map_value map[string]*int) {
=>  15:         p_value := map_value["c"]
    16:         *p_value = 3
    17: }
(dlv) regs
 PC = 0x0000000000068bc0
 SP = 0x0000004000192db0
X29 = 0x00000040001605f8
X30 = 0x0000000000068bc0
```

alexsaezm:
Stack traces from linux/arm64 core dumps showed SP values that differed from the runtime's panic output.

This patch adds a reproducer based on the original snippet from the issue and a test that tries to verify if the SP is the same.

Fixes #3591

### DIFF
--- a/pkg/proc/core/core_test.go
+++ b/pkg/proc/core/core_test.go
@@ -10,7 +10,9 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


alexsaezm:
For the record, I'm not really sure if this is the best approach because it feels wrong to depend on `exec.Command` but as far as I understood from the original report, this only happens on Linux, and they are common tools.

Also, not sure if using `TempDir` is a good idea; maybe this will break on the CI...

pkg/proc/stack.go
```diff
@@ -633,6 +633,14 @@ func (it *stackIterator) advanceRegs() (callFrameRegs op.DwarfRegisters, ret uin
 		}
 	}
 
+	// Issue #3591: On LR architectures, runtime.sigpanic's DWARF CFA points to the
+	// signaled frame's SP, not the caller's SP. Add 2 words (saved FP + saved LR)
+	// to compute the correct caller SP.
+	if fn := it.bi.PCToFunc(it.pc); fn != nil && fn.Name == "runtime.sigpanic" && it.bi.Arch.usesLR {
```
derekparker, line 784: Could we put this into the branch above which checks for this situation? https://github.com/go-delve/delve/pull/4319/changes#diff-39a537b90025d8d637af699fc383b57d2842a0767847b0b0f4402b5942a390baR623

I believe this is a similar issue to what the above branch is fixing where we have a leaf function which ends up looking like it called `runtime.sigpanic` due to the frame injection.

pkg/proc/core/core_test.go
```diff
@@ -545,3 +547,113 @@ func mustSupportCore(t *testing.T) {
...
+
+// TestIssue3591_CoreModMapFrameSP verifies Delve reports the same SP as the
+// Go runtime for sigpanic frames in linux/arm64 core dumps.
+func TestIssue3591_CoreModMapFrameSP(t *testing.T) {
+	if runtime.GOOS != "linux" || runtime.GOARCH != "arm64" {
+		t.Skip("issue #3591 reproduces on linux/arm64 ELF cores only")
+	}
+	mustSupportCore(t)
+
+	tempDir := t.TempDir()
```
aarzilli: Most of this is just withCoreFile right? The only difference is the part where it also reads the stderr? If so we should probably just modify withCoreFile to call `CombinedOutput()` and return it.

alexsaezm: Thanks for the suggestion. I updated `withCoreFile` to call `CombinedOutput`, and then removed all the noise in the new test. Way cleaner for sure.

pkg/proc/stack.go
alexsaezm: I tried to move the code into the branch above, something like this:

```diff
diff --git a/pkg/proc/stack.go b/pkg/proc/stack.go
index dea3d3d8..36c45ac1 100644
--- a/pkg/proc/stack.go
+++ b/pkg/proc/stack.go
@@ -772,20 +772,17 @@ func (it *stackIterator) advanceRegsDWARF() (callFrameRegs op.DwarfRegisters, re
                                                it.err = err
                                        }
                                        binary.Read(bytes.NewReader(buf), binary.LittleEndian, &ret)
+
+                                       // Issue #3591: The DWARF CFA points to the signaled frame's SP, not the
+                                       // caller's SP. Add 2 words (saved FP + saved LR) to compute the correct caller SP.
+                                       callerSP := uint64(it.regs.CFA) + uint64(2*it.bi.Arch.PtrSize())
+                                       callFrameRegs.AddReg(callFrameRegs.SPRegNum, op.DwarfRegisterFromUint64(callerSP))
                                }
                        }
                        retaddr = uint64(it.regs.CFA + regRule.Offset)
                }
        }
 
-       // Issue #3591: On LR architectures, runtime.sigpanic's DWARF CFA points to the
-       // signaled frame's SP, not the caller's SP. Add 2 words (saved FP + saved LR)
-       // to compute the correct caller SP.
-       if fn := it.bi.PCToFunc(it.pc); fn != nil && fn.Name == "runtime.sigpanic" && it.bi.Arch.usesLR {
-               callerSP := uint64(it.regs.CFA) + uint64(2*it.bi.Arch.PtrSize())
-               callFrameRegs.AddReg(callFrameRegs.SPRegNum, op.DwarfRegisterFromUint64(callerSP))
-       }
-
        if it.bi.Arch.usesLR {
                if ret == 0 && it.regs.Reg(it.regs.LRRegNum) != nil {
                        ret = it.regs.Reg(it.regs.LRRegNum).Uint64Val

```

When doing this, the test fails because the SP is off by 16 bytes. I am not sure if I can move it into the loop without breaking it. As far as I understand, I need to do the word addition after the whole register processing.

Did I understand your suggestion right? Or you meant in another branch?

derekparker: Ack, thanks for trying it out.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4322, opened 2026-05-01
native: propagate PtraceGetRegs error on linux/ppc64le

cuiweixie:
`ptraceGetGRegs` ignored the return value from `sys.PtraceGetRegs`, so `registers()` on linux ppc64le never surfaced syscall failures when reading GPRs.

Capture the errno and normalize `syscall.Errno(0)` to nil, consistent with neighbouring ptrace helpers.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4323, opened 2026-05-01
proc: propagate AddrPiece ReadMemory errors in composite memory

cuiweixie:
`newCompositeMemory` ignored errors from `MemoryReadWriter.ReadMemory` for `op.AddrPiece` fragments, so failed memory reads silently produced zeros in the synthesized buffer.

Check the error and return a wrapped message with piece size and address, aligned with existing `RegPiece` error handling in the same function.


aarzilli (approved):

derekparker (approved):
LGTM

PR #4325, opened 2026-05-01
proc: fix OR handling in breakpointConditionSatisfiable

cuiweixie:
When shortcut-checking breakpoint conditions that reference Delve breakpoint hit-counts, breakpointConditionSatisfiable mistakenly applied AND semantics inside token.OR branches: it returned false whenever either subtree was not satisfiable, instead of only when neither side could succeed.

Evaluate disjunction as satisf(X) || satisf(Y), matching token.AND and ordinary boolean logic.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4327, opened 2026-05-01
dwarf/reader: propagate errors during Reader entry iteration

cuiweixie:
Several Reader helpers iterated with `for entry, err := r.Next…(); entry != nil; entry, err = r.Next…()`, so the continuation condition could become false on `(nil, err)` before the loop body ran.

When the body never saw `err`, callers could get `ErrTypeNotFound`, a formatted not-found error, or `(nil, nil)` even though the DWARF walk had already failed. Use an explicit loop that checks `err` and then `entry == nil`.

Touched: SeekToTypeNamed, FindEntryNamed, NextMemberVariable.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4328, opened 2026-05-01
proc: bail out of loadArrayValues after stride overflow

cuiweixie:
`loadArrayValues` sets `v.Unreadable` when `v.Base + uint64(v.stride*count)` wraps (overflow check), mirroring defensive handling elsewhere.

Without an immediate return, the function still ran `cacheMemory` / the element loop afterward, contradicting `v.Unreadable` and risking confusing partially-filled `Children`. Add `return` after that assignment, consistent with the other early exits in the same function.


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4329, opened 2026-05-01
proc: accept uintptr sources in EvalScope pointer casts

cuiweixie:
In `evalTypeCast`, the `*godwarf.PtrType` branch only whitelisted signed int kinds and fixed-size unsigned kinds when turning a loaded constant into a pointer variable.

`reflect.Uintptr` behaves like an unsigned, pointer-sized integer in the Go type system, but it was not in the switch, so expressions such as casting a `uintptr` holding an address to `*T` errored even when the operand was otherwise valid.

Add `reflect.Uintptr` beside the existing uint branches.


aarzilli:
Needs a regression test.

cuiweixie:

PR #4330, opened 2026-05-01
proc: accept uintptr operands in EvalScope primitive casts

cuiweixie:
`evalTypeCast` had two whitelist-style switches that enumerated unsigned operands but omitted `reflect.Uintptr`. That rejected valid casts from Go `uintptr` values when lowering to dwarf `UintType`s or rebuilding `* godwarf.PtrType` views from numeric addresses.

Add `reflect.Uintptr` beside the fixed-width unsigned kinds so both pathways match Go semantics.

Includes the pointer-cast whitelist and the `@godwarf.UintType` branch using `constant.Uint64Val`.

If superseding an older PR that only patched the pointer case, prefer merging this branch and closing the narrower follow-up attempt.


aarzilli:
Needs regression tests. See `TestEvalExpression` in variables_test.go

cuiweixie:

PR #4331, opened 2026-05-01
proc: accept uintptr sources when casting to dwarf IntType

cuiweixie:
In `evalTypeCast`'s `@godwarf.IntType` handling, operands from unsigned reflective kinds reuse `constant.Uint64Val`, then truncate to `ttyp.Size()` via `convertInt`, matching Go two's-complement integer casts.

`reflect.Uintptr` was omitted from that case list despite matching the stored unsigned width semantics, causing explicit casts from Go `uintptr` into dwarf-signed integer descriptors to fail. Add `uintptr` alongside other unsigned kinds.


aarzilli:
This, https://github.com/go-delve/delve/pull/4330 and https://github.com/go-delve/delve/pull/4329 should all be one PR, instead of three.

cuiweixie:

PR #4332, opened 2026-05-01
gobuild: propagate goBuildArgs2 errors from GoBuild

cuiweixie:
`GoBuild` called `goBuildArgs2` but discarded its error, unlike `GoBuildCombinedOutput` / `GoTestBuildCombinedOutput`.

When argument construction fails (e.g. bad `buildflags` shape for the `any` path), return that error instead of running `go build` with missing or partial args.


aarzilli:
This was actually done on purpose, GoBuild is only called in places that pass safe things in buildflags.

PR #4333, opened 2026-05-01
proc: wire LRRegNum into riscv64 Arch

cuiweixie:
`Arch.LRRegNum` is used across `pkg/proc` for call-frame bookkeeping (e.g. `stack.go`, `threads.go` LR updates). ARM64, ppc64le, and loong64 all set it in their `*Arch` constructor; riscv64 did not, leaving the field at the zero default.

Set `LRRegNum` to `regnum.RISCV64_LR` (`x1`), consistent with existing riscv64 unwind / `NewDwarfRegisters` plumbing in the same package.


aarzilli:
Do you have a riscv64 machine or is this done on faith alone?

cuiweixie:
> Do you have a riscv64 machine or is this done on faith alone?

Sorry, I don't have a riscv64 machine to run it.

PR #4335, opened 2026-05-02
proc: compile hit-condition regexp once

cuiweixie:
`parseHitCondition` used `regexp.MustCompile` with an identical pattern each time breakpoints with hit-count conditions were updated.

Hoist `hitConditionRegex` to package scope (`var ... = regexp.MustCompile(...)`) so compilation happens once at package init instead of per call. Parsing remains the same (`FindStringSubmatch` unchanged).


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4336, opened 2026-05-02
refactor: use slices.Backward to simplify the code

criciss:
There is a [new function](https://pkg.go.dev/slices@go1.23.0#Backward)  added in the go1.23 standard library, which can make the code more concise and easy to read.

More info can see  https://github.com/golang/go/issues/61899

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -1,6 +1,7 @@
 package proc
 
 import (
+	"slices"
 	"debug/dwarf"
 	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


aarzilli:
This change contains syntax errors. Also IMO it doesn't simplify the code in any appreciable way and probably makes it slower. I think we should decline.

criciss:
> This change contains syntax errors. Also IMO it doesn't simplify the code in any appreciable way and probably makes it slower. I think we should decline.

Sorry, it does look like there’s a bug... Thank you for your guidance.

PR #4340, opened 2026-05-07
service/dap: add newline to the build message

sagg0t:
Problem: build messages in DAP output get slammed together with the debuggee output.
Solution: add a trailing newline character.


derekparker (approved):
LGTM

PR #4343, opened 2026-05-11
proc: fix range over func stepping for go1.27

aarzilli:
Closure names generation changed in go1.27 because of go.dev/cl/770200.

It is now possible to make the stepping tests with inlining enabled
work.

This patch changes the rangeParentName function to handle the new
naming scheme and reenables most disabled tests.

Some tests still fail and will be investigated later.


derekparker (approved):
lgtm

PR #4345, opened 2026-05-12
proc: fix hasInlines in stacktrace, fix range step with inlining

aarzilli:
The hasInlines flag was not correctly set for inlined flags, causing
stepping through inlined range-over-func bodies to not behave
correctly.


derekparker (approved):
LGTM

PR #4351, opened 2026-05-20
service/dap: relax TestBadLaunchRequest

aarzilli:
Relax TestBadLaunchRequest that fails on go1.27 due to a change in how
a JSON unmarshal error is reported.


derekparker (approved):
LGTM

PR #4352, opened 2026-05-20
proc/internal/ebpf: switch to header+param event ring buffer protocol

derekparker:
Replace the single bpf_ringbuf_reserve/submit of the entire function_parameter_list_t struct with a multi-event protocol: one HEADER event per uprobe hit followed by one PARAM event per parameter. Each PARAM event carries the parameter's kind, val (0x30 bytes), and deref_val (0x30 bytes) inline via bpf_ringbuf_reserve/submit.

BPF changes:
- Define packed event_header_t and param_event_t wire structs
- Snapshot registers into a stack-local u64[16] array so helpers receive PTR_TO_STACK instead of PTR_TO_CTX, which the verifier accepts for indexed access
- Remove goroutine_id from function_parameter_list_t; compute it at runtime and embed it in the HEADER event
- Keep the existing val[0x30]/deref_val[0x30] data layout and the string/pointer/slice deref helpers unchanged

Go-side changes:
- Deadline-based ring buffer polling: set a 10 ms deadline before each Read() call so the goroutine unblocks to check for context cancellation without a concurrent SetDeadline call
- Pending-event reassembly: buffer HEADER events in a map keyed by (goroutine_id, fn_addr, is_ret) and attach incoming PARAM events; emit the completed RawUProbeParams once all params arrive
- Store DWARF types and parameter names during UpdateArgMap and look them up by global index when parsing PARAM events, replacing the old argTypeInfo map
- synthesizeTypeFromKind: create a godwarf.Type from reflect.Kind for parameters that lack a full DWARF type

Other changes:
- CFA offset fix: use FirstPCAfterPrologue and the frame table to compute the CFA-from-RSP offset at the uprobe PC so DW_OP_fbreg expressions evaluate to correct RSP-relative offsets; input and return parameters are now evaluated at different PCs
- Thread a LoadConfig through the RPC, debugger, and target layers so trace output verbosity is configurable
- Upgrade the eBPF Docker build image to Ubuntu 26.04 / clang-22

This sets the stage for additional changes to be submitted later to allow for larger structs / arrays and dereferencing pointers.


aarzilli (commented):
TestGeneratedDoc fails.

aarzilli (approved):
LGTM

PR #4358, opened 2026-05-28
proc: add test for accessing global C variables

aarzilli:
It turns out that we only test this functionality accidentally through
TestDWZCompression which will soon stop working because dwz doesn't
work with 1.25 and later.

Make a copy of the test that doesn't use dwz compression.


derekparker (approved):
lgtm

PR #4359, opened 2026-05-28
gobuild: increase wait length for deleting binaries on Windows

### LINKED ISSUE #4357: gobuild.Remove() fails on Windows with "Access is denied" due to insufficient retry duration

1.26.3

go version go1.26.3 windows/amd64

windows 11 x64

### Description

# `gobuild.Remove()` fails on Windows with "Access is denied" due to insufficient retry duration

## Description

On Windows, when a debugged process is terminated (e.g., via Ctrl+C in the integrated terminal), `gobuild.Remove()` frequently fails to delete the `__debug_bin` temporary binary with an "Access is denied" error. This leaves stale debug binaries behind in the project directory.

The root cause is that the retry logic in `gobuild.Remove()` is too aggressive: it retries 20 times with 1ms intervals, totaling only **20ms** of wait time. On Windows, the kernel may hold the executable file lock longer than 20ms after the process exits, making this window insufficient.

## Steps to Reproduce

1. Open a Go project in VS Code on Windows
2. Start a debug session (F5) for a Go program that has a non-trivial shutdown sequence (e.g., closing network connections, cleaning up resources)
3. Press Ctrl+C in the integrated terminal to terminate the program
4. Observe the following error in the terminal:

```
could not remove C:\...\__debug_bin.exe1234567890: remove C:\...\__debug_bin.exe1234567890: Access is denied.
```

## Root Cause Analysis

The issue is in `pkg/gobuild/gobuild.go`:

```go
func Remove(path string) {
    var err error
    for i := 0; i < 20; i++ {
        err = os.Remove(path)
        if err == nil || runtime.GOOS != "windows" {
            break
        }
        time.Sleep(1 * time.Millisecond)  // 20 x 1ms = 20ms total
    }
    if err != nil {
        fmt.Fprintf(os.Stderr, "could not remove %v: %v\n", path, err)
    }
}
```

The shutdown sequence on Windows is:

1. `kill()` in `pkg/proc/native/proc_windows.go` calls `syscall.TerminateProcess(hProcess, 1)`
2. Then waits for `EXIT_PROCESS_DEBUG_EVENT` via `waitForDebugEvent(waitBlocking)`
3. Then calls `p.Wait()` to confirm OS-level process exit
4. Then calls `dbp.postExit()` to clean up internal state
5. Control returns to `Session.Close()` in `pkg/service/dap/server.go`, which immediately calls `gobuild.Remove()`

Although `p.Wait()` confirms the process handle is signaled, the Windows kernel may still be in the process of releasing file locks on the executable. For larger programs (e.g., 60MB+ binaries), this cleanup can take longer than 20ms.

This is acknowledged in the test suite (`server_test.go:5412-5421`):

```go
rmErrRe := regexp.MustCompile(`could not remove .*\n`)
rmErr := rmErrRe.FindString(string(err))
if rmErr != "" {
    if runtime.GOOS != "windows" || !stringContainsCaseInsensitive(rmErr, "Access is denied") {
        t.Fatalf("Binary removal failure:\n%s\n", rmErr)
    }
}
```

The test explicitly tolerates "Access is denied" on Windows, confirming this is a known limitation.

## Expected Behavior

`gobuild.Remove()` should wait long enough for the Windows kernel to release the file lock, so the temporary debug binary is always cleaned up.

## Actual Behavior

`gobuild.Remove()` gives up after 20ms, leaving `__debug_bin.exe*` files behind in the project directory.

## Proposed Fix

Increase the retry count and interval to provide a more generous retry window on Windows:

```go
func Remove(path string) {
    var err error
    for i := 0; i < 100; i++ {
        err = os.Remove(path)
        if err == nil || runtime.GOOS != "windows" {
            break
        }
        time.Sleep(50 * time.Millisecond)  // 100 x 50ms = 5s total
    }
    if err != nil {
        fmt.Fprintf(os.Stderr, "could not remove %v: %v\n", path, err)
    }
}
```

This increases the maximum wait from 20ms to 5 seconds, which should be sufficient for even large binaries. The happy path (successful deletion on first try) is unaffected.

## Environment

- **OS**: Windows 10/11
- **Go version**: 1.24+
- **Delve version**: 1.25.0 / 1.26.3 (issue present in both)
- **IDE**: VS Code with Go extension (v0.52.2)
- **Binary size**: ~68MB (typical for medium-to-large Go projects)

aarzilli:
Increases the time interval in which we retry deleting the executable's
binary on Windows from 20ms total to ~5s total. Also retries with an
exponential backoff algorithm.

Fixes #4357

### DIFF
--- a/pkg/gobuild/gobuild.go
+++ b/pkg/gobuild/gobuild.go
@@ -4,6 +4,7 @@ package gobuild
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 	"runtime"
@@ -18,15 +19,20 @@ import (
 // This can be used to remove the temporary binary generated for the session.
 func Remove(path string) {
 	var err error
-	for range 20 {
+	// Open files can be removed on Unix, but not on Windows, where there also appears
+	// to be a delay in releasing the binary when the process exits.
+	// Leaving temporary files behind can be annoying to users, so we try again.
+	//
+	// Does backoff exponentially starting at 1ms all the way to ~400ms for a
+	// total of 4.8s of wait time.
+	for i := range 66 {
+		if i != 0 {
+			time.Sleep(time.Millisecond * time.Duration(math.Pow(1.1, float64(i-1))))
+		}
 		err = os.Remove(path)
-		// Open files can be removed on Unix, but not on Windows, where there also appears
-		// to be a delay in releasing the binary when the process exits.
-		// Leaving temporary files behind can be annoying to users, so we try again.
 		if err == nil || runtime.GOOS != "windows" {
 			break
 		}
-		time.Sleep(1 * time.Millisecond)
 	}
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "could not remove %v: %v\n", path, err)


derekparker (approved):
lgtm

Issue #4360, opened 2026-05-28
go1.27: support generic methods


adonovan:
This is just a reminder that Go 1.27 will permit generic methods. The type checker and x/tools packages have now all been updated to support it (and the required changes are generally light). Please test delve with code that uses the new feature to ensure that it is supported at release time, and let us know if any cmd, std, or x/tools changes are needed. Thanks.

Also, struct literals T{F: 0} will permit the use of embedded field selectors F. Please check those as well.

aarzilli:
See #4356 for generic methods.

adonovan:
Great! Glad I remembered to remind you. ;-)

PR #4362, opened 2026-05-29
proc: add test for struct literals with embedded field selectors

aarzilli:
Because we implemented struct literals through assignment we always
supported this even though Go only started in 1.27. Add an explicit
test for this.

Updates #4360


derekparker (approved):
lgtm

PR #4364, opened 2026-06-07
service/dap: add dap write memory request handler

DrSergei:
Add write memory request handler similar to existing read memory handler

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -856,6 +856,8 @@ func (s *Session) handleRequest(request dap.Message) {
 		s.onDisassembleRequest(request)
 	case *dap.ReadMemoryRequest: // Optional (capability 'supportsReadMemoryRequest')
 		s.onReadMemoryRequest(request)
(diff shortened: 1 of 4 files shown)


service/dap/server.go
```diff
@@ -3737,6 +3740,64 @@ func (s *Session) makeReadMemoryResponse(req dap.Request, addr uint64, data []by
...
+		return
+	}
+
+	data, err := base64.StdEncoding.DecodeString(args.Data)
+	if err != nil {
+		s.sendErrorResponse(request.Request, UnableToWriteMemory, "Unable to write memory", err.Error())
+		return
+	}
+
+	addr := ref.addr
```
aarzilli: Are you sure about this? Suppose `args.Offset` is -3 and `ref.addr` is 1000 the result of this if is 1003.

DrSergei: Thansk, fixed

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4365, opened 2026-06-08
proc: fix stepping test on go1.27 / arm64

aarzilli:
Change a stepping test so it passes on arm64 with go1.27.


derekparker (approved):
LGTM

PR #4367, opened 2026-06-11
proc/test: include GOEXPERIMENT in the fixture cache key 

typesanitizer:
BuildFixture used to cache built fixtures by (name, flags),
but `TestClassicMap` builds testvariables2 under
`GOEXPERIMENT=noswissmap`, which produces a
different binary.

Whichever test built the fixture first determined the
binary every other test got.

If we try to parallelize the tests (upcoming work),
`TestClassicMap` runs first and populates the cache
with a `noswissmap` binary, breaking `TestEvalExpression`'s
`swissmap` type assertions.

So earlier, the fully-serial `TestClassicMap`
would incorrectly use the cached `swissmap` binary
and does not actually exercise classic maps.

### DIFF
--- a/pkg/proc/variables_test.go
+++ b/pkg/proc/variables_test.go
@@ -1984,6 +1984,10 @@ func TestClassicMap(t *testing.T) {
 		{"mnil == m1", false, "", "", "", errors.New("can not compare map variables")},
 		{"mnil == nil", false, "true", "true", "", nil},
 		{"m2", true, "map[int]*main.astruct [1: *{A: 10, B: 11}, ]", "map[int]*main.astruct [...]", "map[int]*main.astruct", nil},
+
+		// Check that the fixture binary actually uses the classic map
+		// implementation: runtime.hmap does not exist in swissmap binaries.
+		{`**(**runtime.hmap)(uintptr(&m1))`, false, `…`, `…`, "runtime.hmap", nil},
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/proc/variables_test.go
```diff
@@ -1984,6 +1984,10 @@ func TestClassicMap(t *testing.T) {
 		{"mnil == m1", false, "", "", "", errors.New("can not compare map variables")},
 		{"mnil == nil", false, "true", "true", "", nil},
 		{"m2", true, "map[int]*main.astruct [1: *{A: 10, B: 11}, ]", "map[int]*main.astruct [...]", "map[int]*main.astruct", nil},
+
+		// Check that the fixture binary actually uses the classic map
+		// implementation: runtime.hmap does not exist in swissmap binaries.
+		{`**(**runtime.hmap)(uintptr(&m1))`, false, `…`, `…`, "runtime.hmap", nil},
```
aarzilli, line 1990: Classic maps have been removed, so this test is going to fail soon.

typesanitizer: The test is already gated on:

```
	if goversion.VersionAfterOrEqual(runtime.Version(), 1, 27) {
		t.Skip("noswissmap experiment removed in Go 1.27")
	}
```

So it shouldn't start failing?

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4368, opened 2026-06-11
proc: run independent tests in parallel

typesanitizer:
Add t.Parallel() to 200+ tests to speed up the test suite
(~35% wall time reduction).

The following are excluded:

- Tests using t.Setenv
- Tests attaching by process name (TestWaitFor*, TestAttachDetach)
- Manual stop and concurrent stepping

Test suite execution time on a M4 Macbook Pro (4P + 6E):

- Before this patch (serial): 220s.
  (Mimicked using `go test ./pkg/proc -count=1 -parallel=1`)
- After this patch (parallel): 140s.
  (`go test ./pkg/proc -count=1 -parallel=24`)

This PR is stacked on top of https://github.com/go-delve/delve/pull/4367

### DIFF
--- a/pkg/proc/arm64_disasm.go
+++ b/pkg/proc/arm64_disasm.go
@@ -11,7 +11,9 @@ func arm64AsmDecode(asmInst *AsmInstruction, mem []byte, regs *op.DwarfRegisters
 	asmInst.Size = 4
 	asmInst.Bytes = mem[:asmInst.Size]
 
+	asmDecodeMu.Lock()
 	inst, err := arm64asm.Decode(mem)
+	asmDecodeMu.Unlock()
 	if err != nil {
 		asmInst.Inst = (*arm64ArchInst)(nil)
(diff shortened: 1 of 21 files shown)


aarzilli (commented):
The tests in proc can not be parallelized due to the way operating systems work.

typesanitizer:
@aarzilli could you provide a longer explanation about which OSes/any links to documentation/relevant code?

AFAIK, the limitation is that a single process can have at most one other process debugging it. It's not a global system-wide limit on the number of processes being actively debugged.

At least on macOS and Linux, I have used debugging in parallel on different binaries.

The tests are also passing, which would be unexpected if there's an system-wide limit, unless they're just silently getting queued inside the kernel (but then the speed improvement would likely not be so dramatic).

aarzilli:
It's because of how waitpid and WaitForDebugEvent work.

PR #4370, opened 2026-06-17
pkg/proc: add GOEXPERIMENT=mapsplitgroup support

aarzilli:
This is the same as #4274 but with a test added. I've checked it against the implementation that was merged and it seems to be still matching.


derekparker (approved):
LGTM

derekparker (changes requested):
Actually, there are a lot of failures in CI for 1.26.

derekparker (approved):
LGTM

PR #4373, opened 2026-06-19
service/dap: fix test failures from ExitedEvent change

derekparker:
TestStepOutPreservesGoroutine was not updated to handle the new ExitedEvent that now precedes TerminatedEvent on process exit. On arm64, the test program seems to exit more readily, triggering this path.

Also replace hardcoded listen ports in cmd/dlv tests with 127.0.0.1:0 to let the OS assign a free port. The actual address is parsed from dlv's stdout. This eliminates "address already in use" flakes from parallel test runs in CI.

Fixes TestStepOutPreservesGoroutine on linux/arm64. Fixes TestRemoteDAPClientMulti CI flake.

### DIFF
--- a/cmd/dlv/dlv_test.go
+++ b/cmd/dlv/dlv_test.go
@@ -91,9 +91,18 @@ func filterProcessExitLines(output []byte) []byte {
 	return bytes.Join(filtered, []byte("\n"))
 }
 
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM

PR #4377, opened 2026-06-25
misc: Go 1.27 compatibility and windows/arm64 test fixes

derekparker:
Skip TestCapsLock on Go 1.27 RC/beta/devel builds where the exact
version string format differs from stable releases.

Fix several test failures on windows/arm64. TestGeneratedDoc was
running gen-usage-docs without the exp.winarm64 build tag, which was
lost during the partial revert in b17af91d. TestFrameEvaluation could
panic with an index out of range when variable evaluation returned
garbage values for goroutines deep in runtime frames — added a bounds
check and increased leniency on windows/arm64 to tolerate more than
one unreadable goroutine.

The debugdetect integration tests were failing because dlv exited
without printing the listen address. The root cause was
golang.org/x/sys/windows using the deprecated syscall.Syscall15 in
Proc.Call, which overflows the 792-byte nosplit stack limit when
compiled with -gcflags="all=-N -l" on windows/arm64 (golang/go#78317).
Updated golang.org/x/sys from v0.26.0 to v0.46.0 which uses
syscall.SyscallN instead. Also improved the test to scan for the listen
address marker in a loop rather than assuming it is the first line of
stdout, and to include stderr in the failure message.


aarzilli (approved):
LGTM

aarzilli:
Still draft?

PR #4379, opened 2026-06-26
go.mod: update disassembler

aarzilli:
Supports disassembling more AVX, AVX2 and AVX-512 instructions.


pkg/proc/x86_disasm.go
```diff
@@ -21,7 +21,6 @@ func x86AsmDecode(asmInst *AsmInstruction, mem []byte, regs *op.DwarfRegisters,
 
 	asmInst.Size = inst.Len
 	asmInst.Bytes = mem[:asmInst.Size]
-	patchPCRelX86(asmInst.Loc.PC, &inst)
```
aarzilli, line 24: This change is somewhat scary because this line has been there since 2016. I remove it (and add the handling of x86asm.Rel arguments to resolveCallArgX86) because changing the arguments of CALL instructions to immediates violates assumptions in the formatting code of golang.org/x/arch/x86/x86asm, see the comment in golang.org/x/arch/x86/x86asm/plan9x.go function plan9Arg, case Imm.

It should be fine since the other architectures, in particular arm64, does not patch out relative addresses, so this makes the two codepaths more similar.

derekparker (approved):
LGTM

PR #4383, opened 2026-06-30
proc: fix panic on follow-exec process exit

### LINKED ISSUE #3183: Debugger stops at breakpoint in a loop only once (Windows Arm64)
**1. dlv version**
```
Delve Debugger
Version: 1.9.1
Build: $Id: d81b9fd12bfa603f3cf7a4bc842398bd61c42940 $
```

**2. go version**
`go version go1.19.3 windows/arm64`

**3. Operating system and processor architecture**
```
Edition	       Windows 11 Pro
Version	       22H2
OS build	       22621.521
System type  64-bit operating system, ARM-based processor
```

**6. Steps to reproduce**
   - Install delve for Windows arm64 `go install -tags exp.winarm64 github.com/go-delve/delve/cmd/dlv`
   - Create sample.go file with the outlined below content
      `package main

        import (
            "fmt"
        )

        func main() {
            length, capacity, y := 1000, 2000, 3
            AllResults := make([]int, length, capacity)
            for i := 0; i < length; i++ {
                result := i / y // Hi there
                fmt.Println(result)
                AllResults[i] = result
            }
            j := 0
            fmt.Println(j)
        }`
   - Start debug client: `dlv debug sample.go`
   - Add breakpoint inside loop: `(dlv) break sample.go:12` 
   - Continue debug session  `(dlv) continue` (debugger stops at breakpoint)
   - Continue debug session further `(dlv) continue`  

**7. Expected result**
The debugger stops at the breakpoint in every cycle in a loop.

**8. Actual result**
The debugger stops at the breakpoint only in first cycle of the loop.

example

derekparker:
- Make `postExit` idempotent to prevent double-release of the `ptraceThread` refcount
- Flush instruction cache after writing memory on Windows to fix intermittent missed breakpoints on ARM64

### Fix 1: postExit double-release

In `trapWaitInternal`, a process leader can have `postExit` called via the ESRCH path (when `resumeWithSig` fails because the thread is gone), and then called again when the actual exit event arrives. The double-release drops the `ptraceThread` refcount too low, closing `ptraceChan` while another process in the follow-exec group still needs it. The next ptrace operation on that process panics with "send on closed channel".

Uses atomic `Swap` instead of `Store` so the second call is a no-op.

Observed on linux/riscv64 with Go 1.27rc1 as a panic in `TestLaunchWithFollowExec`.

### Fix 2: FlushInstructionCache on Windows

On ARM64 the instruction and data caches are not coherent. `WriteProcessMemory` only updates the data cache, so breakpoint instructions written by the debugger may not be visible to the instruction fetch unit. This causes intermittent missed breakpoints on Windows/ARM64 — the CPU executes the original instruction from stale I-cache instead of the BRK that was written.

Calls `FlushInstructionCache` after every `WriteProcessMemory` to ensure I-cache coherency. On x86/amd64 this is a documented no-op.

Fixes #3183

### DIFF
--- a/pkg/proc/native/proc.go
+++ b/pkg/proc/native/proc.go
@@ -421,7 +421,9 @@ func (dbp *nativeProcess) execPtraceFunc(fn func()) {
 }
 
 func (dbp *nativeProcess) postExit() {
-	dbp.exited.Store(true)
+	if dbp.exited.Swap(true) {
+		return
+	}
 	dbp.ptraceThread.release()
 	dbp.bi.Close()
 	if dbp.ctty != nil {
--- a/pkg/proc/native/syscall_windows.go
+++ b/pkg/proc/native/syscall_windows.go
@@ -138,6 +138,7 @@ type _DEBUG_EVENT struct {
 //sys	_ResumeThread(threadid syscall.Handle) (prevsuspcount uint32, err error) [failretval==0xffffffff] = kernel32.ResumeThread
 //sys	_ContinueDebugEvent(processid uint32, threadid uint32, continuestatus uint32) (err error) = kernel32.ContinueDebugEvent
 //sys	_WriteProcessMemory(process syscall.Handle, baseaddr uintptr, buffer *byte, size uintptr, byteswritten *uintptr) (err error) = kernel32.WriteProcessMemory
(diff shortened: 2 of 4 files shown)


aarzilli (approved):
LGTM

PR #4385, opened 2026-07-02
dap: make variable load limits configurable

### LINKED ISSUE #4064: Feature: Add possibility to configure max string/bytes limits for on-demand content loading for debug
**Context**
I've read the issue https://github.com/golang/vscode-go/issues/1318 and I'd like to go back to a topic of truncating the long strings..

And in mine case these limits are not sufficient so as @polinasok said in https://github.com/golang/vscode-go/issues/1318#issuecomment-904366663 I'm letting you know :D 

I'm working on a huge payload which comes as `[]byte` and sometimes is visible as a `string`
According to a https://github.com/go-delve/delve/blob/f498dc8c5a8ad01334a9d782893c10bd0addb510/service/dap/server.go#L289 and https://github.com/go-delve/delve/blob/f498dc8c5a8ad01334a9d782893c10bd0addb510/service/dap/server.go#L2919
it's not possible to change it via the config.

I'm using VSCode extension and created topic there (https://github.com/golang/vscode-go/issues/3795#issuecomment-3115090298)
For me using newest adapter is a bit useless as I constantly am hitting the limit and I need to create bunch of `Println` which is rather not sth that should be done..

**Requested feature description**
I like the idea of having different limits depends on context - it's good for the performance. But the hardcoded logic limits are sth bad.
I would like to have on-demand configurable max idea

I would like to be able to configure different kind of limits (but only for on-demand) context
```
maxStringLen: Maximum number of bytes read from a string (default: 64).
maxArrayValues: Maximum number of elements read from an array, slice, or map (default: 64).
maxStructFields: Maximum number of fields read from a struct. A setting of -1 indicates that all fields should be read (default: -1).
maxVariableRecurse: How far to recurse when evaluating nested types (default: 1).
followPointers: Automatically dereference pointers (default: true).
```

larrasket:
the server hardcoded MaxStringLen (512) and MaxArrayValues (64) in DefaultLoadConfig with no way to adjust them, debugging sessions that need to see longer strings or larger collections had to page through reloads or use the dlv connection workaround. 

this adds maxStringLen and maxArrayValues to the launch/attach configuration and to the config evaluation request. 
Fixes #4064

### DIFF
--- a/service/dap/types.go
+++ b/service/dap/types.go
@@ -185,6 +185,14 @@ type LaunchAttachCommonConfig struct {
 	// Default is 50.
 	StackTraceDepth int `json:"stackTraceDepth,omitempty"`
 
+	// Maximum number of bytes loaded for strings.
+	// Default is 512.
+	MaxStringLen int `json:"maxStringLen,omitempty"`
+
+	// Maximum number of array, slice and map elements loaded.
+	// Default is 64.
+	MaxArrayValues int `json:"maxArrayValues,omitempty"`
+
 	// Boolean value to indicate whether global package variables
 	// should be shown in the variables pane or not.
 	ShowGlobalVariables bool `json:"showGlobalVariables,omitempty"`
--- a/service/dap/command.go
+++ b/service/dap/command.go
@@ -166,7 +166,7 @@ func (s *Session) examineMemory(goid, frame int, argstr string) (string, error) 
 	var address uint64
 
 	if args.IsExpr {
-		val, err := s.debugger.EvalVariableInScope(int64(goid), frame, 0, args.Operand, DefaultLoadConfig)
+		val, err := s.debugger.EvalVariableInScope(int64(goid), frame, 0, args.Operand, s.loadConfig())
 		if err != nil {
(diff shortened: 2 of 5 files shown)


service/dap/types.go
```diff
@@ -185,6 +185,14 @@ type LaunchAttachCommonConfig struct {
 	// Default is 50.
 	StackTraceDepth int `json:"stackTraceDepth,omitempty"`
 
+	// Maximum number of bytes loaded for strings.
+	// Default is 512.
+	MaxStringLen int `json:"maxStringLen,omitempty"`
```
aarzilli, line 190: Changes to this struct need to also be documented by updating the corresponding json descriptions in vscode-go. For example: https://go-review.googlesource.com/c/vscode-go/+/753000

aarzilli (approved):
LGTM

service/dap/types.go
larrasket: https://github.com/golang/vscode-go/pull/4054

derekparker (approved):
LGTM

PR #4387, opened 2026-07-03
pkg/terminal: add exit -d to detach without killing the process

### LINKED ISSUE #2324: Exit from attached process without prompt
When using _dlv attach_ to debug local running process, it's really annoying to answer the prompt
`Would you like to kill the process? [Y/n]`
It slows the debugging flow a lot and this question is not something critical when debugging a local process.

I propose that config param is introduced which could set allow user that prefers not to kill running process.
Already did some implementation which introduces config param
`soft-detach: true`
If param is true it will skip the `yesno` question, and will exit from dlv without killing the process.

If this is something that could be useful, I can submit PR request, but wanted to discuss this enhancement here first.

larrasket:
Add an exit -d flag that detaches from the target.

`exitCommand` records the intent on the terminal and `handleExit` skips the kill prompt and detaches with `kill=false`.

Fixes #2324

pkg/terminal/exit_test.go was generated by Claude Sonnet 5

### DIFF
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -76,6 +76,10 @@ type Term struct {
 	// should be resumed before quitting.
 	quitContinue bool
 
+	// detachNoKill is set to true by exitCommand when passed -d to signal that
+	// the target process should be left running, without prompting.
+	detachNoKill bool
+
 	longCommandMu         sync.Mutex
 	longCommandCancelFlag bool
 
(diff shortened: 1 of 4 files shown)


aarzilli (commented):
What's the use case for this? I think something like this came up before and it's actually fewer keystrokes to answer the prompt.

pkg/terminal/exit_test.go
```diff
@@ -0,0 +1,58 @@
+package terminal
```
aarzilli: This can just go in terminal_test.go

pkg/terminal/terminal.go
```diff
@@ -593,7 +597,10 @@ func (t *Term) handleExit() (int, error) {
 
 		if doDetach {
 			kill := true
-			if t.client.AttachedToExistingProcess() {
+			switch {
+			case t.detachNoKill:
```
aarzilli: This branch should be after the `AttachedToExistingProcess` check.

larrasket:
@aarzilli  #2324  leaving an attached process running without answering the prompt every time is the use case.  interactively yes it just saves answering n. The concrete gain is non-interactive use (for example yesno goes through `liner.Prompt`, which returns EOF when stdin isn't a tty, so a scripted exit on an attached process returns 2 without detaching) -d makes the detach deterministic there.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4388, opened 2026-07-08
terminal: extend frame pointer unwinding blocking branch to fix TestScopePrefix index panic on windows/arm64

derekparker:
TestScopePrefix intermittently panics with an index-out-of-range on windows/arm64 CI when the evaluated value of `i` is garbage (likely a stack/pointer address) instead of the expected 0-9, indicating a race in stack unwinding on that backend. Guard the index and fail with the raw eval output, stack, and goroutines listing so the next occurrence is diagnosable instead of a bare panic.


aarzilli (approved):
LGTM

PR #4391, opened 2026-07-14
dwarf/frame: check the error after executing an opcode

minhbq-99:
executeDwarfInstruction will return immediately when there is a previous error. In the execution loop, if we don't check the error, nothing changes after each iteration and we get an infinite loop. Check the error after executing every opcode and break in case of error.

### DIFF
--- a/pkg/dwarf/frame/table.go
+++ b/pkg/dwarf/frame/table.go
@@ -175,7 +175,7 @@ func executeDwarfProgramUntilPC(fde *FrameDescriptionEntry, pc uint64) *FrameCon
 }
 
 func (frame *FrameContext) executeDwarfProgram() {
-	for frame.buf.Len() > 0 {
+	for frame.buf.Len() > 0 && frame.err == nil {
 		executeDwarfInstruction(frame)
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/dwarf/frame/table.go
```diff
@@ -177,6 +177,9 @@ func executeDwarfProgramUntilPC(fde *FrameDescriptionEntry, pc uint64) *FrameCon
 func (frame *FrameContext) executeDwarfProgram() {
 	for frame.buf.Len() > 0 {
```
aarzilli: Better `&& frame.err == nil` here

pkg/dwarf/frame/table.go
```diff
@@ -190,6 +193,9 @@ func (frame *FrameContext) ExecuteUntilPC(instructions []byte) error {
 	// are currently at in the traced process).
 	for frame.address >= frame.loc && frame.buf.Len() > 0 {
```
aarzilli: Same.

pkg/dwarf/frame/table.go
minhbq-99: I've fixed it.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4392, opened 2026-07-14
dwarf/frame: add DW_CFA_GNU_args_size opcode

minhbq-99:
This opcode is used by C++ for fixing the stack when unwinding exception. When backtracing the stack, we don't need to care about it. So just read it and ignore.


minhbq-99:
For more context, I used this pkg/dwarf/frame to parse eh_frame of a C++ binary so I stumbled this opcode. The change is quite small. But if it's not the correct direction for the project (I don't know if we have this opcode in Go binary), please feel free to close the PR.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4393, opened 2026-07-14
proc/core: handle UTF-16LE output from procdump64 in TestMinidump

derekparker:
ProcDump v12.01 (released 2026-07-09) produces UTF-16LE encoded console output, causing TestMinidump to fail because strings.Contains cannot match ASCII against UTF-16LE bytes. The dump itself completes successfully but the success string check breaks.

Add decodeProcdumpOutput which tries the plain string first, then falls back to UTF-16LE decoding via golang.org/x/text if the success marker isn't found. Add golang.org/x/text as a new dependency.


aarzilli (approved):
LGTM

PR #4394, opened 2026-07-15
service/test: skip waitreason test on development versions

aarzilli:
There is no point in updating the code for this early because the list
of waitreasons could change late in the development cycle and then the
test wouldn't catch it.

Disable the test on development builds so that it reminds us of making
an update after the rc comes out.


derekparker (approved):
LGTM

PR #4395, opened 2026-07-15
service/dap,terminal: add option to display expanded/raw strings

### LINKED ISSUE #4245: Console: Option to expand "\n", "\t" and other formatting characters
Currently when `call` returns a multiline string, delve puts in one line without expanding formatting characters. I would like an option to print an expanded multiline string.

https://github.com/golang/vscode-go/issues/3592

hitzhangjie:
This PR adds a new option that controls how escape characters in string
values are displayed. When enabled, newlines (`\n`), tabs (`\t`) and
other control characters are preserved as actual characters in the
output instead of being displayed as escaped sequences.

The implementation adds a `PrettyRawString` flag to the API
pretty-printing layer and a `quoteRawString` helper that only escapes
double-quotes and backslashes, preserving raw control characters.

**DAP (`service/dap`)**

Adds a `showRawStrings` launch configuration option. In `launch.json`:

```json
"showRawStrings": true
```

**Terminal (pkg/terminal)**
Adds a `show-raw-strings` config option. In the dlv REPL or config file:

```
config show-raw-strings true
```

Both options share the same formatting backend.

Fixes #4245

### DIFF
--- a/service/api/prettyprint.go
+++ b/service/api/prettyprint.go
@@ -26,9 +26,7 @@ const (
 	prettyIncludeType
 	PrettyNewlines    // pretty print variable on multiple lines
 	PrettyShortenType // pretty print variable shortening types when they are printed
-	PrettyRawString   // when formatting strings, preserve raw escape characters (\n, \t, etc.)
-	                  // instead of using Go's %q escaping. Intended for DAP where multi-line
-	                  // values are supported by the protocol.
+	PrettyRawString   // print strings as-is without any modification, like %s
 )
 
 func (flags PrettyFlags) top() bool         { return flags&prettyTop != 0 }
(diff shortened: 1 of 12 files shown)


service/api/prettyprint.go
```diff
@@ -267,13 +271,42 @@ func (v *Variable) writeBasicType(buf io.Writer, fmtstr string) {
...
 		}
 		fmt.Fprintf(buf, fmtstr, v.Value)
 	}
 }
 
+// quoteRawString produces a double-quoted Go string literal that preserves
+// raw escape characters (newlines, tabs, etc.) in the output. Only
+// double-quotes and backslashes are escaped. This is intended for DAP where
+// the protocol supports multi-line values in the Variable result field.
+func quoteRawString(s string) string {
```
aarzilli: I'm tempted to say that show-raw-strings should mean raw strings and no quoting at all should be made.

hitzhangjie: Yes, here should print the string as it is. Print the string with %s, remove this function.

hitzhangjie:
The latest Aggregator -- teamcity build failed. It's apt-get failed.

aarzilli (approved):
LGTM

derekparker (changes requested):
Can you update the documentation at `Documentation/api/dap/README.md` in the table with the large list of optional arguments and add this one as well.

hitzhangjie:
> Can you update the documentation at `Documentation/api/dap/README.md` in the table with the large list of optional arguments and add this one as well.

Yes, I have updated the doc. 

fd6530eddce7ce7ac2a17dce6c0e4ce8edaba4653eb592f59360f490812b78c0

derekparker (approved):
LGTM

PR #4397, opened 2026-07-21
service/dap: set type in SetVariable response

### LINKED ISSUE #3170: DAP Evaluate and SetVariables response not containing `type` with `supportsVariableType`.

larrasket:
Fixes #3170

### DIFF
--- a/service/dap/server.go
+++ b/service/dap/server.go
@@ -3688,6 +3688,7 @@ func (s *Session) onSetVariableRequest(request *dap.SetVariableRequest) {
 
 	response := &dap.SetVariableResponse{Response: *s.newResponse(request.Request)}
 	response.Body.Value = arg.Value
+	response.Body.Type = s.getTypeIfSupported(evaluated)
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

Issue #4363, opened 2026-06-06
telemetry report: pkg/proc: evalCallInjectionSetTarget missing an argument


aarzilli:
```
#!stacks
"proc.(*evalStack).fncallPeek" && "runtime.panicBounds" && "proc.(*EvalScope).evalCallInjectionSetTarget"
```
Issue created by [stacks](https://pkg.go.dev/golang.org/x/telemetry/cmd/stacks).

This stack `lwoC8w` was [reported by telemetry](https://telemetry.go.dev/data/2026-06-05):

- `delve/bug`
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).executeOp.func1:+3`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/eval.go#L1101)
- [`runtime.gopanic:+51`](https://cs.opensource.google/go/go/+/go1.26.1:src/runtime/panic.go;l=860)
- [`runtime.panicBounds64:+25`](https://cs.opensource.google/go/go/+/go1.26.1:src/runtime/panic.go;l=236)
- [`runtime.panicBounds:+14`](https://cs.opensource.google/go/go/+/go1.26.1:src/runtime/asm_arm64.s;l=1770)
- `github.com/go-delve/delve/pkg/proc.(*evalStack).fncallPeek:=915`
  - [`github.com/go-delve/delve/pkg/proc.(*EvalScope).evalCallInjectionSetTarget:+1`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/fncall.go#L990)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).executeOp:+151`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/eval.go#L1247)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).run:+4`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/eval.go#L1020)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).resume:+61`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/eval.go#L1013)
- [`github.com/go-delve/delve/pkg/proc.callInjectionProtocol:+36`](https://github.com/go-delve/delve/blob/v1.26.1/pkg/proc/fncall.go#L1187)
[...]

Issue #4085, opened 2025-08-09
proc: telemetry report: index error in fncallPeek while resuming execution of a call injection


aarzilli:
```
#!stacks
"goPanicIndex" && "proc.(*evalStack).fncallPeek"
```
Issue created by [stacks](https://pkg.go.dev/golang.org/x/telemetry/cmd/stacks).

This stack `5cV75w` was [reported by telemetry](https://telemetry.go.dev/data/2025-08-03):

- `delve/bug`
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).executeOp.func1:+3`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/eval.go#L1054)
- [`runtime.gopanic:+50`](https://cs.opensource.google/go/go/+/go1.24.5:src/runtime/panic.go;l=792)
- [`runtime.goPanicIndex:+2`](https://cs.opensource.google/go/go/+/go1.24.5:src/runtime/panic.go;l=115)
- `github.com/go-delve/delve/pkg/proc.(*evalStack).fncallPeek:=873`
- [`github.com/go-delve/delve/pkg/proc.(*EvalScope).evalCallInjectionSetTarget:+1`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/fncall.go#L983)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).executeOp:+151`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/eval.go#L1200)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).run:+4`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/eval.go#L973)
- [`github.com/go-delve/delve/pkg/proc.(*evalStack).resume:+56`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/eval.go#L966)
- [`github.com/go-delve/delve/pkg/proc.callInjectionProtocol:+36`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/fncall.go#L1180)
- [`github.com/go-delve/delve/pkg/proc.(*TargetGroup).Continue:+112`](https://github.com/go-delve/delve/blob/v1.25.0/pkg/proc/target_exec.go#L168)
- [`github.com/go-delve/delve/service/debugger.(*Debugger).Command:+38`](https://github.com/go-delve/delve/blob/v1.25.0/service/debugger/debugger.go#L1100)
[...]

PR #4398, opened 2026-07-21
proc: avoid panic when call injection finishes before SetTarget

derekparker:
Guard CallInjectionSetTarget against an empty fncalls stack and treat premature RestoreRegisters / unknown protocol register values as errors so evaluation fails cleanly instead of panicking (issues #4085, #4363).

### DIFF
--- a/pkg/proc/fncall.go
+++ b/pkg/proc/fncall.go
@@ -93,6 +93,10 @@ type functionCallState struct {
 	// undoInjection is set after evalop.CallInjectionSetTarget runs and cleared by evalCallInjectionComplete
 	// it contains information on how to undo a function call injection without running it
 	undoInjection *undoInjection
+	// targetSet is true after CallInjectionSetTarget runs. Unlike undoInjection it
+	// is not cleared by CallInjectionComplete, so RestoreRegisters can tell whether
+	// the target was ever configured.
+	targetSet bool
 
 	// hasDebugPinner is true if the target has runtime.debugPinner
 	hasDebugPinner bool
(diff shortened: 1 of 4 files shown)


aarzilli:
I don't really see how this could be the problem. For us to get to `funcCallStep` some instruction must have set `callInjectionContinue` which is either CallInjectionComplete or CallInjectionComplete2. Either of those get inserted into the instruction stream after CallInjectionSetTarget by evalcompile.go.

Obviously the bug happens, so there must be a path that leads there (which I am not seeing), but to me it does not look like the path is this one.

derekparker:
> I don't really see how this could be the problem. For us to get to `funcCallStep` some instruction must have set `callInjectionContinue` which is either CallInjectionComplete or CallInjectionComplete2. Either of those get inserted into the instruction stream after CallInjectionSetTarget by evalcompile.go.
> 
> Obviously the bug happens, so there must be a path that leads there (which I am not seeing), but to me it does not look like the path is this one.

`CallInjectionStart` also sets `callInjectionContinue` and enters `funcCallStep` *before* SetTarget runs. After Start, the next expected protocol value is CompleteCall (`0`), then `resume` continues into SetTarget. If that first stop is RestoreRegisters (`16`) instead (bad protocol register / mis-attributed stop / etc.), `funcCallFinish` pops the frame and `stack.run()` hits SetTarget with an empty `fncalls` stack and that's the `fncallPeek` panic from #4085/#4363.

pkg/proc/fncall.go
```diff
@@ -957,10 +976,12 @@ func funcCallStep(callScope *EvalScope, stack *evalStack, thread Thread) bool {
 		fncall.panicvar.Name = "~panic"
 
 	default:
-		// Got an unknown protocol register value, this is probably bad but the safest thing
-		// possible is to ignore it and hope it didn't matter.
+		// Unknown protocol register value: do not continue blindly. Continuing
+		// without SetTarget can finish the call and then panic in
+		// evalCallInjectionSetTarget (see issues #4085, #4363).
+		fncall.err = unknownProtocolRegisterError(regval)
```
aarzilli, line 982: This could just be the fmt.Errorf call directly.

aarzilli (commented):
Good point. But then instead of `targetSet` it should be `callInjectionCompleted` that we check for in the case for `debugCallRegRestoreRegisters`, no?

PR #4399, opened 2026-07-21
proc: fix arm64 crosscall2 SP restore for cgo stacktraces

derekparker:
`crosscall2` adjusts SP (e.g. `SUB $(8*24), RSP` on arm64) but never spills it. Delve was either restoring SP from the saved frame pointer (linux/arm64) or loading `*(SP+N)` (other LR arches). Both are wrong: the FP fallback only works when C keeps a frame pointer, and `*(SP+N)` is the caller's stack contents.

After the TeamCity Ubuntu 20.04 -> 24.04 bump (newer GCC), cgo-compiled C more often omits frame pointers, so `TestCgoStacktrace` failed on linux/arm64 at the C→Go callback step.

Restore the caller SP as `SP + framesize`, matching what each arch's `crosscall2` epilogue adds back. Applied on arm64, riscv64, loong64, and ppc64le (with corrected ppc64le BP/LR save slots). Re-enable `TestCgoStacktrace` on linux/ppc64le.

darwin/arm64 shares the arm64 path and gets the same correct restore.

### DIFF
--- a/Documentation/backend_test_health.md
+++ b/Documentation/backend_test_health.md
@@ -25,8 +25,7 @@ Tests skipped by each supported backend:
 	* 1 broken
 * linux/loong64 skipped = 1
 	* 1 not working on linux/loong64
-* linux/ppc64le skipped = 3
-	* 1 broken - cgo stacktraces
+* linux/ppc64le skipped = 2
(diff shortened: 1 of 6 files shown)


aarzilli (approved):
LGTM

PR #4404, opened 2026-08-02
proc/native: update Windows syscall generator

typesanitizer:
I noticed that the generated `zsyscall_windows.go` checked in was not up-to-date, and that the generation logic relied on the Go toolchain instead of the pinned `sys` module. So this patch fixes that issue.

1. Switches the native Windows wrapper generation entry point to `golang.org/x/sys/windows/mkwinsyscall`, using the pinned `golang.org/x/sys` module version.
  - The command uses `-mod=mod` because the vendored tree does not include the `mkwinsyscall` command.
2. Regenerates `zsyscall_windows.go` with that generator. This removes a generated `DbgUiRemoteBreakin` declaration, so retain that address-only procedure explicitly in `syscall_windows.go`.

Validation:

```text
GOOS=windows GOARCH=amd64 go build ./pkg/proc/native
```


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4405, opened 2026-08-04
build: pin github actions to commit shas and declare workflow permissions

Totara-thib:
Hi, drive-by CI hardening in two commits, one logical change each.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v6` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. The spot that matters here is `release.yml`: goreleaser and the cosign installer run with `contents: write` and the keyless signing identity, so a re-pointed tag there means a compromised release could ship signed. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). All shas were resolved from the upstream repos and cross checked against their release tags. The pins stay maintainable: a dependabot config with the `github-actions` ecosystem bumps them automatically, one small block.

**Commit 2 adds `permissions: contents: read` to the windows arm64 test workflow**, the only one without a block. The job never uses the token, so the scope is exact. `release.yml` already declares scoped permissions (nicely done) and is untouched.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v6`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on master. I will also open a PR which adds it to CI so this does not quietly drift back, that one is a bonus, this PR stands on its own.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,22 +35,22 @@ jobs:
           exit 1
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           fetch-depth: 0  # full history required for goreleaser release notes
           ref: ${{ inputs.tag || github.ref }}
 
       - name: Set up Go
-        uses: actions/setup-go@v5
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4406, opened 2026-08-04
build: add plumber workflow security check

Totara-thib:
Companion to https://github.com/go-delve/delve/pull/4405, merge that one first: the check added here flags the unpinned actions and the missing permissions block until the hardening lands, then it goes green.

This adds [Plumber](https://github.com/getplumber/plumber) to CI, the tool I used to find those issues in the first place. It scans the workflows on each push to master and on each PR, and fails when something regresses: an unpinned action, a job without a permissions block, a known CVE, that kind of thing. The gate passes at 85 points, so one small finding does not block your PRs.

- `plumber.yml`: pinned by sha, minimal permissions, findings go to the security tab as SARIF (skipped on PRs from forks, the report stays as an artifact there).
- `.plumber.yaml`: a 14 line overlay that inherits the CLI's built-in baseline, you only see what differs for this repo: one control is off with a comment saying why (the release job restores the Go build cache on tag builds, scoping that cache key is a good follow-up but its own change). Everything else, including new controls in future releases, follows the defaults automatically.
- `README.md`: one line, the score badge next to your build status one.

## Score badge

I enabled `score-push` on the action. It works like OpenSSF Scorecard's published results: every run, on any branch, publishes the score to score.getplumber.io, and that feeds the badge in the README. Scores are public and the badge always shows the state of master. A failed publish never fails your CI. Until the first run the badge reads UNKNOWN in gray, then it flips to the grade. If you would rather not have it, drop the README line and the `score-push` input, the rest works the same.

With the hardening in, this runs green with a score of A. Set `soft-fail: true` if you prefer report only, without gating PRs.

To be fully transparent: I work on Plumber. If you do not want the tool in your CI, no hard feelings, the hardening PR is the one that matters and it stands on its own.

### DIFF
--- a/.github/workflows/plumber.yml
+++ b/.github/workflows/plumber.yml
@@ -1,0 +1,34 @@
+name: Plumber
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  plumber:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      security-events: write
+      # Needed by score-push to publish the score for the README badge.
+      id-token: write
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
+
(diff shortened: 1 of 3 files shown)


aarzilli (approved):

derekparker (approved):
LGTM

Issue #4407, opened 2026-08-05
v1.27.1 release assets were not uploaded because the GitHub Release was already immutable


FranciscoPombal:
## Summary

The `v1.27.1` release workflow successfully:

* built all configured binaries;
* created the archives;
* generated checksums;
* signed the checksums using Cosign.

It then failed while uploading the resulting artifacts to the GitHub Release:

```text
422 Cannot upload assets to an immutable release
```

The GitHub Release for `v1.27.1` had already been published before GoReleaser attempted to upload the assets.
Because release immutability is enabled for the repository, GitHub does not allow assets to be added to or replaced in a published release.

## Cause

The current release workflow is triggered when a version tag is pushed and expects GoReleaser to manage the GitHub Release lifecycle.

The intended sequence is:

```text
push version tag
    -> GitHub Actions workflow starts
    -> GoReleaser creates the release as a draft
    -> GoReleaser uploads all artifacts
    -> GoReleaser publishes the completed release
```

For `v1.27.1`, the sequence appears to have instead been:

```text
create and publish GitHub Release
    -> version tag triggers the workflow
    -> GoReleaser builds the artifacts
    -> GitHub rejects the uploads because the release is already immutable
```

This therefore does not appear to be a build, signing, permissions, or GoReleaser configuration failure.
It is a release-ordering issue caused by publishing the GitHub Release before the automated workflow had uploaded its assets.

## Recovery

The existing `v1.27.1` release cannot be completed by rerunning the workflow because it is already published and immutable.

The practical recovery is:

1. Create a new patch release, such as `v1.27.2`.
2. Create and push the tag without manually creating or publishing its GitHub Release.
3. Allow the tag-triggered release workflow to run.
4. Let GoReleaser create the draft release, upload the artifacts, and publish it.
5. Optionally update the `v1.27.1` release notes to direct users to the replacement release.

The `v1.27.1` tag name should not be deleted with the expectation that it can be reused: tags associated with published immutable releases cannot be reused even after deleting the release.

## Avoiding recurrence

The release procedure should be documented as:

> Create and push the release tag, but do not manually publish a GitHub Release. GoReleaser owns creation and publication of the release.

As an additional safeguard, the workflow could perform an early check for an already-published release with the same tag.
That would not prevent someone from publishing the release manually, but it would stop the workflow before performing the builds and provide a clear, actionable error instead of failing later during artifact upload.

---

Note: diagnosis of this bug was assisted by an AI agent.
[...]

PR #4408, opened 2026-08-05
ci: fail early when a release is already published

FranciscoPombal:
## Summary

Add a preflight check that aborts a release run when the target tag already has a published GitHub Release.

## Motivation

GitHub does not allow assets to be added to a published immutable release.
Previously, this condition was discovered only after GoReleaser had built, archived, checksummed, and signed all artifacts, when the upload stage failed with HTTP 422 errors.

The new check reports the release-ordering problem before the build steps begin.

## Scope

This change:

* runs only for tag-triggered and non-dry-run releases;
* does not create, modify, publish, or delete any release;
* does not prevent a maintainer from publishing a release manually;
* makes that mistake fail immediately with an actionable explanation;
* leaves snapshot builds unchanged.

This makes the release process more resilient to release-ordering mistakes by replacing a late, ambiguous upload failure with an early diagnostic.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,51 @@ jobs:
           echo "::error::A tag must be specified for non-dry-run releases"
           exit 1
 
+      - name: Reject existing published release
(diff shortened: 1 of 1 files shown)


derekparker (approved):
LGTM

PR #4410, opened 2026-08-05
ci: upgrade actions, pin them, and pin GoReleaser version

FranciscoPombal:
## Summary

* upgrade `actions/checkout` from v4 and v5 to v7.0.1;
* upgrade `actions/setup-go` from v5 to v7.0.0;
* keep `sigstore/cosign-installer` on v3 and pin it to v3.10.1;
* upgrade `goreleaser/goreleaser-action` from v6 to v7.2.3;
* upgrade `actions/upload-artifact` from v4 to v7.0.1;
* pin every action to its full commit SHA; and
* pin the GoReleaser executable to v2.17.1 instead of the floating `~> v2` constraint.

## Motivation

Newer actions versions use the Node 24 action runtime, avoiding the Node 20 deprecation path.

The GoReleaser workflow previously requested `~> v2`, allowing a future GoReleaser v2 release to be selected automatically.
Pinning the exact version ensures that release and snapshot jobs use the same known GoReleaser implementation until the repository deliberately upgrades it.

The cosign-installer action remains on v3 because v4 installs Cosign v3 by default.
Cosign v3 requires migrating the existing detached certificate and signature output to a Sigstore bundle, which should be handled as a separate release-workflow change.

## Scope

This change:

* updates and pins the CI actions and the GoReleaser executable;
* does not change build targets, archives, checksums, signing, or publication configuration;
* is independent of the immutable-release upload failure.

This makes the release process more resilient to unreviewed tooling changes and action-runtime deprecations.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,24 +35,24 @@ jobs:
           exit 1
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
         with:
           fetch-depth: 0  # full history required for goreleaser release notes
           ref: ${{ inputs.tag || github.ref }}
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
         with:
           go-version: stable
 
       - name: Install cosign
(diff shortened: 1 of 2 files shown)


derekparker:
Hello, thank you for this patch! Would you mind rebasing this please?

derekparker:
I rebased this onto current master and opened it as https://github.com/go-delve/delve/pull/4427, since the original had merge conflicts after the action SHA pinning landed. Going to close this one in favor of that.

PR #4411, opened 2026-08-06
proc: export LoadFullValue as return-by-value helper

derekparker:
Replace the unexported mutable loadFullValue var with LoadFullValue() so callers get a fresh copy, and deduplicate matching LoadConfig literals across proc, debugger, rpc2, terminal, and tests.


aarzilli (approved):
LGTM

PR #4413, opened 2026-08-06
proc/native: surface FlushInstructionCache failures on Windows

derekparker:
## Summary
Stop swallowing FlushInstructionCache errors after WriteProcessMemory. If the flush fails, return the error so it shows up instead of looking like a missed breakpoint.

Part of digging into flaky TestNextWhileNexting / TestStepShadowConcurrentBreakpoint on windows/arm64 with Go 1.27.

### DIFF
--- a/pkg/proc/native/threads_windows.go
+++ b/pkg/proc/native/threads_windows.go
@@ -2,8 +2,10 @@ package native
 
 import (
 	"errors"
+	"fmt"
 	"syscall"
 
+	"github.com/go-delve/delve/pkg/logflags"
 	sys "golang.org/x/sys/windows"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/proc/native/threads_windows.go
```diff
@@ -118,7 +120,13 @@ func (t *nativeThread) WriteMemory(addr uint64, data []byte) (int, error) {
 	// On ARM64 the instruction and data caches are not coherent,
 	// WriteProcessMemory only updates the data cache so we must flush
 	// the instruction cache explicitly. On x86/amd64 this is a no-op.
-	_ = _FlushInstructionCache(t.dbp.os.hProcess, uintptr(addr), uintptr(len(data)))
+	// Flush failures must not be silent: a stale I-cache can cause missed
+	// breakpoints or execute the wrong instruction after a patch.
+	if err := _FlushInstructionCache(t.dbp.os.hProcess, uintptr(addr), count); err != nil {
+		err = fmt.Errorf("FlushInstructionCache(%#x, %d): %w", addr, count, err)
+		logflags.DebuggerLogger().Errorf("%v", err)
```
aarzilli, line 127: Why both logging and returning?

derekparker: I know it's redundant, but I wanted to throw everything at it and see if anything is discoverable by CI (maybe the error gets swallowed or something). Kind of grasping at straws here honestly trying to figure out these seemingly never-ending windows/arm64 flakes.

PR #4414, opened 2026-08-06
pkg/proc: Defer pointer cast to syscall invoking function

typesanitizer:
Previously, the logic in `registers(*nativeThread)` would
create a pointer to a stack variable `threadInfo`, immediately
convert it to a `uintptr` and pass that `uintptr` down to
`_NtQueryInformationThread` which would invoke `SyscallN`.

This meant that if the call to `_NtQueryInformationThread`
forced growth of the stack (due to running out of space),
the uintptr (interpreted as a pointer) would be logically
dangling. And the syscall would successfully write to that memory.

After the syscall finished, the `status` would be a success,
but the subsequent read from the local's field
`threadInfo.TebBaseAddress` would not actually be the
correct value, because the write happened at the old address.

This PR changes the signature of _NtQueryInformationThread
to take `*_THREAD_BASIC_INFORMATION` and regenerates the syscall
wrapper. With this change, the `*T` -> `unsafe.Pointer` -> `uintptr`
conversion happens within the `SyscallN` call expression,
which is correctly handled by the compiler.

One downstream impact of this is that it fixes flakiness
in TestCallFunction on Windows with addr=0x1488
which seems to correspond to

```
// Offsets into Thread Environment Block (pointer in GS)
#define TEB_TlsSlots 0x1480
```

in the runtime assembly stubs.

### DIFF
--- a/pkg/proc/native/syscall_windows.go
+++ b/pkg/proc/native/syscall_windows.go
@@ -136,7 +136,7 @@ type _DEBUG_EVENT struct {
 	U              [160]byte
 }
 
-//sys	_NtQueryInformationThread(threadHandle syscall.Handle, infoclass int32, info uintptr, infolen uint32, retlen *uint32) (status _NTSTATUS) = ntdll.NtQueryInformationThread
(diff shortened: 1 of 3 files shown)


derekparker (approved):
LGTM

derekparker:
Github actions (where our Windows CI lives) is apparently having a bad day, but this fix seem very targeted and correct, so I will go ahead and merge now.

PR #4415, opened 2026-08-06
service/dap: stop attach fixtures inheriting test stdio

derekparker:
Fixture processes that inherit os.Stdout/os.Stderr keep go test's pipes open after the package exits, causing flaky WaitDelay / "Test I/O incomplete" failures (seen on linux/386 CI). Discard fixture stdio by default, Kill+Wait on cleanup, and route attach tests that need stdout through a pipe helper.

### DIFF
--- a/service/dap/server_test.go
+++ b/service/dap/server_test.go
@@ -514,15 +514,7 @@ func TestLaunchStopOnEntry(t *testing.T) {
 func TestAttachStopOnEntry(t *testing.T) {
 	runTest(t, "loopprog", func(client *daptest.Client, fixture protest.Fixture) {
 		// Start the program to attach to
-		cmd := exec.Command(fixture.Path)
-		stdout, err := cmd.StdoutPipe()
-		if err != nil {
-			t.Fatal(err)
-		}
-		cmd.Stderr = os.Stderr
-		if err := cmd.Start(); err != nil {
-			t.Fatal(err)
-		}
+		cmd, stdout := startFixtureWithStdoutPipe(t, fixture)
 		// Wait for output.
 		// This will give the target process time to initialize the runtime before we attach,
 		// so we can rely on having goroutines when they are requested on attach.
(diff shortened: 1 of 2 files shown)


service/dap/server_test.go
```diff
@@ -4139,20 +4123,72 @@ func substitutePathTestHelper(t *testing.T, fixture protest.Fixture, client *dap
...
+
+// execFixture starts fixture.Path without attaching its stdio to the test
+// process. Inheriting os.Stdout/os.Stderr shares go test's pipes; if the child
+// outlives the test binary, go test fails with WaitDelay / "I/O incomplete".
+// A Cleanup hook kills and waits for the process.
+func execFixture(t *testing.T, fixture protest.Fixture, args ...string) *exec.Cmd {
 	t.Helper()
-	// TODO(polina): do I need to sanity check testBackend and runtime.GOOS?
-	cmd := exec.Command(fixture.Path)
-	cmd.Stdout = os.Stdout
```
aarzilli, line 4148: Seeing the stdout and stderr is useful for debugging in some cases. Besides passing os.Stdout and os.Stderr is what we do in every other case anyway in pkg/proc/native/proc_linux.go.

It can't be that this is the problem and if it was it would happen on all linux platforms, not just on linux/386.

derekparker: So what lead me here is failures such as https://delve.teamcity.com/buildConfiguration/Delve_linux_386_1_27/80319?logFilter=debug&logView=flowAware&showLog=80319_57096_582. 

```
PASS
*** Test I/O incomplete 1m0s after exiting.
exec: WaitDelay expired before I/O complete
FAIL	github.com/go-delve/delve/service/dap	153.394s
```

So something along these lines is causing intermittent failures in CI.

aarzilli: It could be any of the tests that's doing this, they all give stdout and stderr to a child process. If we print the output of `ps aux` at the end of TestMain maybe we can manage to narrow it down.

derekparker: I believe it is only present on `attach` tests because the attach path uses a different kill mechanism (SIGINT vs SIGKILL).

aarzilli: If we detached properly it shouldn't make any difference. Maybe we are not detaching properly with dap on linux/386?

PR #4417, opened 2026-08-10
service/dap: print "ps aux" before exiting tests

aarzilli:
The tests on linux/386 occasionally fail with a "Test I/O incomplete
1m0s after exiting" error. See if we can use the list of running
processes to determine which test is at fault.


derekparker (approved):
LGTM

PR #4419, opened 2026-08-14
service/dap: honor the frame and goroutine of the SetVariable scope

### LINKED ISSUE #3171: Setting a variable on an upper frame using DAP fails
**Delve Version**: 1.9.1
**Go Version**: 1.16
**Machine**: Linux AMD64

Using `dlv-dap` on VSCode (And also on other DAP compatible editors), after stopping on a breakpoint and trying to set an upper frame's local variable, the following error occurrs:
image

I think the main reason for this is that the code handling the `setVariable` request purposefully uses the stopped goroutine & top frame for the location of the variable to set, while it might not be the case: 

https://github.com/go-delve/delve/blob/beaa04604a7e2811446191e0b3f4c970b4e9d406/service/dap/server.go#L2870-L2879

larrasket:
onSetVariableRequest hardcoded goid, frame := -1, 0, setting a local in a non-top stack frame is either: failed to find the variable or silently mutated the same-named local in frame 0.

Carry the scope's goroutine and frame index on the variable handle, propagate them to child handles, and use them in onSetVariableRequest.

Fixes #3171

### DIFF
--- a/service/dap/handles.go
+++ b/service/dap/handles.go
@@ -27,6 +27,12 @@ type fullyQualifiedVariable struct {
 	// startIndex is the index of the first child for an array or slice.
 	// This variable represents a chunk of the array, slice or map.
 	startIndex int
+	// goroutineID and frameIndex identify the scope (goroutine and stack frame)
+	// this variable belongs to, so that SetVariable acts on the correct frame
+	// rather than always the top frame. They are propagated from the scope
+	// handle to its child variable handles.
+	goroutineID int
+	frameIndex  int
 }
 
 func newHandlesMap[T any]() *handlesMap[T] {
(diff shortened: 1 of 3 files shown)


service/dap/server.go
```diff
@@ -2726,7 +2726,7 @@ func (s *Session) onScopesRequest(request *dap.ScopesRequest) {
 		globScope := &fullyQualifiedVariable{&proc.Variable{
 			Name:     fmt.Sprintf("Globals (package %s)", currPkg),
 			Children: slicePtrVarToSliceVar(globals),
-		}, currPkg, true, 0}
+		}, currPkg, true, 0, goid, frame}
```
aarzilli: globals are globals

service/dap/handles.go
```diff
@@ -27,6 +27,12 @@ type fullyQualifiedVariable struct {
 	// startIndex is the index of the first child for an array or slice.
 	// This variable represents a chunk of the array, slice or map.
 	startIndex int
+	// goroutineID and frameIndex identify the scope (goroutine and stack frame)
+	// this variable belongs to, so operations such as SetVariable act on the
```
aarzilli: Only SetVariableRequest uses this, remove the "such as".

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4420, opened 2026-08-14
proc: fuzz eval stack machine and harden call-injection paths

derekparker:
Follow-up to #4412. Adds two in-process fuzzers for paths `FuzzEvalExpression`
never sees, and uses them to close real eval / call-injection bugs.

- **`FuzzEvalStackOps`** feeds opcode programs to `evalStack`. The decoder is
  total (`tag % N`), jump targets are remapped to the actual program length,
  and the filter is `DepthCheck(ops, -1)` (underflow / joins only — not
  Compile's end-depth 0/1, so `Roll` actually runs).
- **`FuzzCallInjectionProtocol`** drives real `funcCallStep` /
  `funcCallFinish` / `evalStack.run` with a mock protocol register, covering
  premature `RestoreRegisters` without a live `debugCallV2` target.
- **`FuzzEvalExpression`** takes a second `string` input and seeds a sample of
  real and invalid expressions.

Production fixes found along the way:
- `Roll.depthCheck` now reports `N+1` slots, matching `executeOp`.
- `DepthCheck` rejects out-of-range jump targets (used to index `depth[]`
  and panic) and swaps the end-depth error args.
- Empty `fncalls` is a clean `"terminated before target"` error
  (`requireFncall`), not a panic in `SetTarget` / `funcCallStep` / Complete
  (#4085, #4363).
- `evalStack.run` bounds `JumpAlways` cycles with a length-relative step
  limit (`max(1024, len(ops)*256)`). DepthCheck treats a self-jump as valid.
5s smokes for the new fuzzers run in `_scripts/fuzz_eval_expression.sh fuzz`
only (nightly workflow). TeamCity `seed` mode is still setup + corpus check.
FuzzEvalExpression setup dumps are gitignored.

[Disclaimer: fuzz tests added via AI assistance]


pkg/proc/evalop/evalcompile.go
```diff
@@ -235,16 +235,13 @@ func (ctx *compileCtx) pushOp(op Op) {
 	ctx.ops = append(ctx.ops, op)
 }
 
-// depthCheck validates the list of instructions produced by Compile and
-// CompileSet by performing a stack depth check.
-// It calculates the depth of the stack at every instruction in ctx.ops and
```
aarzilli, line 240: I think this part of the comment should be retained.

pkg/proc/evalop/evalop_test.go
```diff
@@ -2,11 +2,115 @@ package evalop
...
+
+	t.Run("OutOfRangeJumpTarget", func(t *testing.T) {
+		ops := []Op{&Jump{When: JumpAlways, Target: 99}}
+		if err := DepthCheck(ops, 0); err == nil {
+			t.Fatal("expected out-of-range jump target error")
+		}
+	})
+
+	t.Run("RollUnderflow", func(t *testing.T) {
+		// Roll{N} needs N+1 stack entries (see executeOp). A lone PushNil
```
aarzilli: This comment doesn't belong.

pkg/proc/evalop/evalop_test.go
```diff
@@ -2,11 +2,115 @@ package evalop
...
+			&PushNil{},
+			&Roll{N: 1},
+		}
+		if err := DepthCheck(ops, 2); err != nil {
+			t.Fatalf("DepthCheck: %v", err)
+		}
+	})
+
+	t.Run("NegativeEndDepthSkipsFinalCheck", func(t *testing.T) {
+		// Two PushNil + Roll{1} is underflow-safe but ends at depth 2.
```
aarzilli: This comment doesn't belong.

pkg/proc/eval.go
```diff
@@ -884,6 +884,24 @@ type evalStack struct {
...
 }
 
+// evalStepLimit bounds how many opcodes a single evalStack.run invocation
+// may execute. Legitimate programs either finish or suspend via call
+// injection (which returns from run and resets the budget on the next
+// resume). Without this limit, a depth-consistent JumpAlways cycle never
+// suspends and hangs the debugger. The budget is length-relative so a
+// 1-op cycle cannot burn a huge constant, but is still well above typical
+// compiled expressions (tens of ops) and nested call-injection resumes.
+func evalStepLimit(nops int) int {
```
aarzilli: IIRC we never jump backwards, there are no loops so this is not necessary. We could add a check that there are no loops in CheckDepth (and/or impose a maximum length there).

pkg/proc/eval.go
```diff
@@ -915,6 +933,16 @@ func (s *evalStack) fncallPeek() *functionCallState {
 	return s.fncalls[len(s.fncalls)-1]
 }
 
+func (s *evalStack) requireFncall() *functionCallState {
```
aarzilli: I don't think we should do this. It's going to obfuscate the telemetry reports. The error also assumes too much (it could be that there are no fncalls because we compiled something wrong).

pkg/proc/eval_stack_fuzz_test.go
```diff
@@ -0,0 +1,355 @@
...
+		return
+	}
+	for _, op := range ops {
+		if j, ok := op.(*evalop.Jump); ok {
+			j.Target = j.Target % n
+		}
+	}
+}
+
+// evalStackOpsDepthOK reports whether ops are underflow-safe and have
```
aarzilli: Unnecessary comment.

pkg/proc/eval_stack_fuzz_test.go
```diff
@@ -0,0 +1,355 @@
...
+		ops := []evalop.Op{
+			&evalop.PushConst{Value: constant.MakeInt64(1)},
+			&evalop.Pop{},
+		}
+		FailIfInternalDebuggerError(t, runEvalStackOps(ops))
+	})
+
+	t.Run("EmptyPopIsInternalOrPanicRecovered", func(t *testing.T) {
+		err := runEvalStackOps([]evalop.Op{&evalop.Pop{}})
+		// Depth-invalid pop is recovered inside executeOp; this is not a
```
aarzilli: I don't get this comment. Every panic is recovered inside executeOp.

pkg/proc/export_test.go
```diff
@@ -0,0 +1,15 @@
+package proc
+
+import (
+	"strings"
+	"testing"
+)
+
+// FailIfInternalDebuggerError fails t if err is an internal debugger error.
+// Visible to proc_test (see variables_fuzz_test.go).
+func FailIfInternalDebuggerError(t testing.TB, err error) {
```
aarzilli: This isn't doing anything that needs to be in a separate file, as far as I can tell.

pkg/proc/fuzz_helpers_test.go
```diff
@@ -0,0 +1,21 @@
+package proc
```
aarzilli: I don't think anything here is used anywhere except in fncall_protocol_fuzz_test.go

pkg/proc/fncall_protocol_fuzz_test.go
```diff
@@ -0,0 +1,467 @@
+// Layer 2 call-injection protocol fuzzing harness.
```
aarzilli: What does this fuzz exactly? If we get a wrong sequence of register values for R12?

pkg/proc/variables_fuzz_test.go
```diff
@@ -59,8 +63,19 @@ func FuzzEvalExpression(f *testing.F) {
 	fns, err := bi.FindFunction("main.main")
 	assertNoError(err, f, "FindFunction main.main")
 	fi.Loc.Fn = fns[0]
-	f.Add(fi.Fuzzbuf)
-	f.Fuzz(func(t *testing.T, fuzzbuf []byte) {
+	f.Add(fi.Fuzzbuf, "i1")
```
aarzilli: I don't think this change belongs to this PR. But I also don't understand it. The fuzz function already calls getEvalExpressionTestCases what good does it do to add a few extra ones? Fuzzing the expression itself is just going to result in a parsing error and the parser is not part of delve.

derekparker:
Going to close for now, I think it should be broken up and reworked a bit.

PR #4421, opened 2026-08-16
proc: advance through FreeBSD process snapshots

typesanitizer:
`waitForSearchProcess` advanced its process pointer only after processing an entry. When an entry was already present in `seen`, `continue` skipped that update, causing the remaining iterations to repeatedly inspect the same process and miss newer entries.

Compute each entry from the process-array base and loop index so every iteration advances. Add a FreeBSD regression test that populates `seen`, starts a new process, and verifies that a second scan reaches it.

Tested on FreeBSD 14.3 amd64 under QEMU/KVM:

```
go test ./pkg/proc/native -run TestWaitForSearchPastSeenProcess -count=100
go test ./pkg/proc/native ./pkg/proc
```

### DIFF
--- a/pkg/proc/native/waitfor_test.go
+++ b/pkg/proc/native/waitfor_test.go
@@ -1,0 +1,59 @@
+package native
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+)
+
+const waitForHelperEnv = "DELVE_WAITFOR_HELPER"
+
+func TestWaitForSearchPastSeenProcess(t *testing.T) {
+	switch runtime.GOOS {
+	case "freebsd", "linux", "windows":
+		// waitForSearchProcess is only implemented on these operating systems.
+	default:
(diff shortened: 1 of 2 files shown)


pkg/proc/native/waitfor_test.go
```diff
@@ -0,0 +1,51 @@
+package native
```
aarzilli, line 1: There's nothing specific to freebsd in this test, however waitForSearchProcess is only implemented in freebsd, linux and windows. Rather than restricting to freebsd with the filename check runtime.GOOS.

typesanitizer: Changed as requested.

typesanitizer:
@aarzilli could you take a second look?

aarzilli (approved):
LGTM

typesanitizer:
@aarzilli sorry I don't have merge access, can you merge?

aarzilli:
@derekparker will after review.

derekparker (approved):
LGTM

PR #4424, opened 2026-08-19
tests: report error with t.Fatal if we can't compile a fixture

aarzilli:
Instead of calling os.Exit(1) and fmt.Printf report the failure with
t.Fatal. This makes our tests behave correctly with TeamCity.


derekparker (approved):
LGTM

PR #4425, opened 2026-08-20
debugger: close resumeNotify early with the halt command

aarzilli:
When we receive an halt command we wait to close the resumeNotify
channel until we acquire the target mutex, which is only possible after
the previous continue command has completed.

This means that upon receiving an halt command, we will not be able to
serve any other request, including asynchronous requests, until the
previous continue command is completed.

This breaks the intended use of the GetEvents request. The one that is
in-flight when we receive the halt command will terminate but any
subsequent GetEvents command will not be able to run until continue
finishes.

With the way the eventsChan in rpc2/server.go is used, this can also
lead to a deadlock (continue gets blocked on sending to eventsChan
through the eventsFn callback, and the eventsChan never gets drained
because GetEvents can't run).

To resolve this problem we close the resumeNotify channel early for
halt commands.

Additionally we drop events in rpc2/server.go if the eventsChan is
filled up, so that improper use of the API does not ever lead to a
deadlock.

### DIFF
--- a/_fixtures/cgotest.go
+++ b/_fixtures/cgotest.go
@@ -6,10 +6,17 @@ char* foo(void) { return "hello, world!"; }
 */
 import "C"
 
-import "fmt"
-import "runtime"
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"time"
+)
 
 func main() {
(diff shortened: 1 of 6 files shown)


derekparker (approved):
LGTM

PR #4426, opened 2026-08-20
tests: regenerate capslock

aarzilli:
Disable capslock if it isn't built with an updated version of
x/go/tools because it currently crashes.

### DIFF
--- a/_scripts/capslock_darwin_amd64-output.txt
+++ b/_scripts/capslock_darwin_amd64-output.txt
@@ -4,7 +4,7 @@ For additional debugging signals, use verbose mode with -output=verbose
 To get machine-readable full analysis output, use -output=json
 
 Analyzed packages:
-  github.com/cilium/ebpf v0.11.0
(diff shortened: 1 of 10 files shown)


derekparker (approved):
LGTM

derekparker (changes requested):
Actually, I'm not sure if pinning (or searching) for a specific version is the best approach here. Could we catch the panic or error message and skip based off that instead?

aarzilli:
Changed

derekparker:
Looks like capslock may actually need a regen on windows.

aarzilli:
Done. Unless I'm missing something the capslock regeneration script was wrong, it passed the GOOS/GOARCH environment variables to `go run` which creates an unrunnable capslock executable.

aarzilli:
Changed to just regenerating capslock files since capslock has been updated to support go1.27.

derekparker (approved):
LGTM

PR #4427, opened 2026-08-20
ci: upgrade GitHub Actions and pin GoReleaser

derekparker:
This upgrades the GitHub Actions we use to current versions so CI is on the Node 24 runtime, and pins GoReleaser itself to v2.17.1 instead of the floating ~> v2 constraint. Cosign installer stays on v3 so we don't have to migrate signing output to Sigstore bundles here.

fuzz.yml still had floating v5 tags after the earlier SHA pinning, so that is pinned too, along with checkout in plumber.yml. The fuzz workflow also now declares contents: read.

This is #4410 rebased onto current master.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -10,6 +10,9 @@ concurrency:
   group: '${{ github.workflow }}'
   cancel-in-progress: true
 
+permissions:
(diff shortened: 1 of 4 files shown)


aarzilli (approved):
LGTM

PR #4428, opened 2026-08-21
ci: upload fuzz crashers as GitHub Actions artifacts

derekparker:
When FuzzEvalExpression finds a failing input, the crasher and the setup files needed to replay it were left on the runner. Upload them on failure or cancellation so they can be inspected after the job ends.


aarzilli (approved):
LGTM

PR #4431, opened 2026-08-22
proc: do not call debuginfod-find right after attach

aarzilli:
Calling debuginfod-find can lead to minute long pauses, if we do it
while we are executing the Attach constructors we have no way of
notifying the user about this and the user has no way of aborting the
operation.

This means that delve will appear to be stuck somewhere, for long
periods of time.

This change disables debuginfod calls immediately after attach (to an
existing process). The libraries that weren't loaded this way will be
loaded after the next continue when clients can set up a notification
channel and disable them with CancelDownloads.

A followup commit will extend the functionality of
DownloadLibraryDebugInfo, so that it too can print notifications, and
also call it automatically from the client after an attach.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -118,6 +118,7 @@ type BinaryInfo struct {
 	debugPinnerFn *Function
 	logger        logflags.Logger
 	eventsFn      func(*Event)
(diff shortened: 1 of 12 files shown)


derekparker (approved):
LGTM

Issue #4430, opened 2026-08-22
FreeBSD CI is gone: Cirrus CI shut down on 2026-06-01


neilpang:
.cirrus.yml no longer runs anywhere.  Cirrus CI shut down:

    "Cirrus CI will shut down effective Monday, June 1, 2026."
    -- https://cirruslabs.org/  (announced 2026-04-07)

The last check-run it posted here is on 8fc4acbd, 2026-05-25, and it was
green:

    https://github.com/go-delve/delve/runs/77678888441

Since then 60 commits have landed on master with no FreeBSD result at
all, 28 of them touching pkg/proc.  A commit today carries the Windows
ARM64 jobs and plumber, nothing else.  PR #4421 changes
pkg/proc/native/proc_freebsd.go and adds a FreeBSD-only regression test
that nothing in CI can run.

cirrus-ci.com still resolves and accepts TCP on 443, but the TLS
handshake returns zero bytes and EOF, from three separate networks,
while github.com from the same hosts is fine.  There is nothing left to
fix in .cirrus.yml.

Its scripts are not the problem.  Unchanged, in a FreeBSD VM:

    pkg install -y go gcc git  ->  go 1.26.6, gcc 14.2.0, git 2.55.0
    make test                  ->  exit 0

    FreeBSD 15.1   211s   559 passed, 0 failed, 77 skipped
    FreeBSD 14.4   227s   559 passed, 0 failed, 77 skipped

(The lang/go packaging problem from #3994 is fixed; go resolves to
go126-1.26.6 again.)

Would you take a GitHub Actions job that runs the same make test in a
FreeBSD VM on ubuntu-latest, so the result lands next to the Windows
one?  One new file, .github/workflows/test-freebsd.yml, modelled on
test-windows-arm64.yml -- actions pinned to commit SHAs, permissions
contents: read, the same concurrency block.  Green on a fork in six
minutes, 1122 passed / 0 failed / 78 skipped:

    https://github.com/neilpang/delve/actions/runs/32358548830

One thing to know before saying yes: the job will be red about as often
as Cirrus was, and for the same reason -- dlv trace hangs intermittently
on FreeBSD, filed as #4429.  Restoring CI does not fix that; it makes it
visible.

I would leave .cirrus.yml in place unless you want it removed.

AI disclosure, per Documentation/AI/AI_POLICY.md: I used Claude Code to
run the FreeBSD VMs, collect the numbers above and draft this text.  I
[...]

Issue #4429, opened 2026-08-22
dlv trace hangs intermittently on FreeBSD (wait4 never returns)


neilpang:
What did you do

Ran dlv trace against the test fixtures in a loop on FreeBSD:

    cd _fixtures/buildtest
    dlv trace --output /tmp/x/__debug ../../_fixtures/goroutines-trace.go callme

What did you expect to see

It prints the trace and exits, as on Linux.

What did you see instead

Some runs never exit.  dlv sits in wait4 while the traced process is
already stopped, so nothing will wake it:

    ps -axwwo pid,ppid,stat,wchan,command
     17473 17437  I   uwait  dlvbin trace --output /tmp/tr1/__debug ...
     17481 17473  TX  -      /tmp/tr1/__debug

SIGQUIT on the hung dlv:

    golang.org/x/sys/unix.Wait4(...)
    pkg/proc/native.(*nativeProcess).wait
        pkg/proc/native/proc_freebsd.go:447
    pkg/proc/native.(*nativeProcess).trapWaitInternal
        pkg/proc/native/proc_freebsd.go:345
    pkg/proc/native.trapWait
        pkg/proc/native/proc_freebsd.go:327
    pkg/proc/native.(*processGroup).ContinueOnce
        pkg/proc/native/proc.go:295

    goroutine 1 is the trace command itself:
    pkg/terminal.(*Commands).cont      pkg/terminal/command.go:1374
    cmd/dlv/cmds.New.func8.traceCmd.1  cmd/dlv/cmds/commands.go:911

Trace output up to that point is normal, it just stops mid-stream.  The
hung pair is never reaped: after a few go test ./cmd/dlv runs there were
13 leftover dlv+__debug pairs on the machine.

How often

One test on its own, 30 iterations, idle bare-metal KVM, 6 vCPU:

    FreeBSD 15.1-RELEASE-p1 amd64   8/30 hung
    FreeBSD 14.4-RELEASE    amd64   6/30 hung

The rate is load dependent, so I would not read one probability off
those.  In whole-suite runs, 4 of 8 died on a trace test, and not always
the same one -- TestTraceMultipleGoroutines (goroutines-trace.go),
TestTrace2 (traceprog.go), TestTraceVerbosityLevel2 (traceverb.go).
Each ends as:

    panic: test timed out after 10m0s
        running tests:
            TestTrace2 (9m51s)
    FAIL github.com/go-delve/delve/cmd/dlv 600.045s

[...]

PR #4432, opened 2026-08-24
ci: run the FreeBSD tests on GitHub Actions

neilpang:
Replaces the Cirrus job, which cannot come back: Cirrus CI shut down on
2026-06-01.  Details and measurements in #4430.

One new file, .github/workflows/test-freebsd.yml, modelled on
test-windows-arm64.yml -- same on:/concurrency: block, actions pinned to
commit SHAs, permissions contents: read.  It runs the same "make test"
that .cirrus.yml ran, in a FreeBSD VM on ubuntu-latest, over both
supported branches.  .cirrus.yml is removed, as you suggested.

The release input is the major only ('14', '15'), so the newest release
of each branch keeps being picked up without editing this file.

prepare needs one line .cirrus.yml did not.  The workspace is rsynced
into the VM keeping the runner's uid, which does not exist there, so git
refuses to read the repository:

    error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
    no Go files in /home/runner/work/delve/delve

The second line is the real damage: $(go list ...) in the vet target
comes back empty, so go vet runs against the repository root instead.
git config --global --add safe.directory "$GITHUB_WORKSPACE" fixes it
and leaves make test byte-identical to what Cirrus ran.

Verified on a fork -- the FreeBSD 15 job, six minutes, 1122 passed,
0 failed, 78 skipped:

    https://github.com/neilpang/delve/actions/runs/32358548830

Expect red runs at roughly the rate Cirrus had.  In that same run the
FreeBSD 14 job died on the dlv trace hang, #4429; this job does not
cause it, it makes it visible again.

AI disclosure, per Documentation/AI/AI_POLICY.md: I used Claude Code to
run the FreeBSD VMs, verify the workflow and draft this text.  I
reviewed and edited it; the numbers come from runs on hardware I have.


.github/workflows/test-freebsd.yml
```diff
@@ -0,0 +1,49 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        # Both supported branches. Only the leading part is given, so the
+        # newest release of each keeps being picked up without an edit here.
+        freebsd-version: ['14', '15']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
```
aarzilli: We are already using `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1` everywhere else.

aarzilli (approved):
LGTM

derekparker (approved):
LGTM - thank you!

PR #4433, opened 2026-08-24
proc: fix infinite loop in mapIteratorSwiss.next

aarzilli:
Fixes a bug found by the fuzzer in
https://github.com/go-delve/delve/actions/runs/32555989728.

If the table length is too large (because of memory corruption) the
multiplication between that value and the size of the group struct can
overflow. If the overflow is just right the resulting stride will be
zero, causing the loop in mapIteratorSwiss.next to never progress.

Also changes said loop to respect maxNumGroups even when we can't load
any tables but dirLen is large. This is unrelated to the fuzzer bug, it
was merely spotted during the investigation.

We might need some other mechanism to stop the search in mapAccess
early if the table is enormous, or appears to be because of memory
corruption, but doesn't cause an integer overflow. However there are no
obvious limits to impose there so leave it for now.

### DIFF
--- a/pkg/proc/eval.go
+++ b/pkg/proc/eval.go
@@ -2883,6 +2883,10 @@ func (v *Variable) sliceAccess(idx int) (*Variable, error) {
 }
 
 func (v *Variable) mapAccess(idx *Variable) (*Variable, error) {
+	// TODO(aarzilli): here if the memory is corrupt we could end up looking at
+	// a lot of memory, and taking a long time. However there is also no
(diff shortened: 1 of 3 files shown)


aarzilli:
What do we want to do with `testdata`? The `fuzz` directory is pretty small but it needs the executable and the coredump to run and those are big and different every time they are run. I don't want to push multiple 40MB objects into the repository, since it's not something users are interested in downloading.

derekparker (approved):
LGTM

PR #4434, opened 2026-08-28
terminal,service,proc: add set next statement (jump) command

### LINKED ISSUE #1045: Support for instruction reordering / jumping

larrasket:
Adds a "jump" (alias "j") command and a SetExecutionPoint RPC method that set the next instruction to be executed by the current thread to a given location, without executing any of the instructions in between.

The target must be inside the function the current thread is stopped in as jumping to a different function would leave the stack frame set up for the old function and corrupt execution.

Fixes #1045

### DIFF
--- a/Documentation/cli/README.md
+++ b/Documentation/cli/README.md
@@ -12,6 +12,7 @@ Command | Description
 --------|------------
 [call](#call) | Resumes process, injecting a function call (EXPERIMENTAL!!!)
 [continue](#continue) | Run until breakpoint or program termination.
+[jump](#jump) | Set the next instruction to be executed (EXPERIMENTAL!!!).
 [next](#next) | Step over to next source line.
 [next-instruction](#next-instruction) | Single step a single cpu instruction, skipping function calls.
 [rebuild](#rebuild) | Rebuild the target executable and restarts it. It does not work if the executable was not built by delve.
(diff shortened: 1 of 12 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4436, opened 2026-08-30
proc: iterate target detaches with slices.Backward

elonfliter:
TargetGroup.Detach must process targets in reverse order. 

Use slices.Backward to express that traversal directly while preserving validity checks, detach ordering, and aggregated errors.


aarzilli:
This just makes the code harder for the compiler for no benefit whatsoever.

PR #4438, opened 2026-09-01
proc,service: emit events while calling DownloadLibraryDebugInfo

aarzilli:
Emit download events produced after calling DownloadLibraryDebugInfo,
fix a bug where the download context wasn't reset.

When -1 is passed to DownloadLibraryDebugInfo download all libraries
that weren't downloaded previously.

Automatically call DownloadLibraryDebugInfo(-1) client side after an
attach.

### DIFF
--- a/pkg/proc/bininfo.go
+++ b/pkg/proc/bininfo.go
@@ -1155,8 +1155,18 @@ func (bi *BinaryInfo) AddImage(path string, addr uint64) error {
 	return err
 }
 
+// ResetDownloadsContext resets the downloads context. If the context was
+// cancelled before this BinaryInfo object will be able to do downloads
+// again.
+func (bi *BinaryInfo) ResetDownloadsContext() {
+	bi.cancelDownloadsMu.Lock()
+	bi.downloadsCtx, bi.cancelDownloads = context.WithCancel(context.Background())
+	bi.cancelDownloadsMu.Unlock()
+}
+
 // LoadImageBinaryInfoAgain loads the n-th image debug symbols if they weren't already loaded.
 func (bi *BinaryInfo) LoadImageBinaryInfoAgain(n int) error {
+	bi.attaching = false
 	if n < 0 || n >= len(bi.Images) || bi.Images[n].loadErr == nil {
 		return nil
 	}
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -166,6 +166,7 @@ func New(client service.Client, conf *config.Config, isAttachCmd bool) *Term {
 				if !firstEventBinaryInfoDownload {
 					fmt.Fprintf(t.stdout, "\n")
 				}
+				firstEventBinaryInfoDownload = true
 			case api.EventBreakpointMaterialized:
 				bp := event.BreakpointMaterializedEventDetails.Breakpoint
 				file := t.formatPath(bp.File)
--- a/pkg/terminal/terminal.go
+++ b/pkg/terminal/terminal.go
@@ -73,6 +73,9 @@ type Term struct {
 
 	substitutePathRulesCache [][2]string
 
+	// isAttachCmd is true if this terminal was stated by 'dlv attach...'
+	isAttachCmd bool
+
 	// quitContinue is set to true by exitCommand to signal that the process
 	// should be resumed before quitting.
 	quitContinue bool
@@ -104,7 +107,7 @@ type displayEntry struct {
 }
 
 // New returns a new Term.
-func New(client service.Client, conf *config.Config) *Term {
+func New(client service.Client, conf *config.Config, isAttachCmd bool) *Term {
 	cmds := DebugCommands(client)
 	if conf != nil && conf.Aliases != nil {
 		cmds.Merge(conf.Aliases)
@@ -120,6 +123,8 @@ func New(client service.Client, conf *config.Config) *Term {
 		line:   liner.NewLiner(),
 		cmds:   cmds,
 		stdout: &transcriptWriter{pw: &pagingWriter{w: os.Stdout}},
+
+		isAttachCmd: isAttachCmd,
 	}
 	t.line.SetCtrlZStop(true)
 
@@ -154,7 +159,7 @@ func New(client service.Client, conf *config.Config) *Term {
 				}
 				fmt.Fprintf(t.stdout, "Downloading debug info for %s: %s (press ^C to cancel)", event.BinaryInfoDownloadEventDetails.ImagePath, event.BinaryInfoDownloadEventDetails.Progress)
 				firstEventBinaryInfoDownload = false
-			case api.EventStopped:
+			case api.EventStopped, api.EventDownloadLibraryInfoDone:
 				t.downloadsMu.Lock()
 				t.downloadsInProgress = false
 				t.downloadsMu.Unlock()
@@ -415,6 +420,10 @@ func (t *Term) Run() (int, error) {
 	// Ensure that the target process is neither running nor recording by
 	// making a blocking call.
 	_, _ = t.client.GetState()
+
+	if t.isAttachCmd {
+		t.client.DownloadLibraryDebugInfo(-1)
+	}
 
 	for {
 		locs = nil
--- a/service/debugger/debugger.go
+++ b/service/debugger/debugger.go
@@ -2664,10 +2664,31 @@ func (d *Debugger) CancelDownloads() bool {
 }
 
 // DownloadLibraryDebugInfo attempts to download the specified library's debug info.
-func (d *Debugger) DownloadLibraryDebugInfo(n int) error {
+func (d *Debugger) DownloadLibraryDebugInfo(n int, eventsFn func(*proc.Event)) error {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
-	return d.target.Selected.BinInfo().LoadImageBinaryInfoAgain(n)
+	if eventsFn != nil {
+		defer eventsFn(&proc.Event{Kind: proc.EventDownloadLibraryInfoDone})
+	}
+	d.target.SetEventsFn(eventsFn)
+	bi := d.target.Selected.BinInfo()
+	bi.ResetDownloadsContext()
+	if n > 0 {
+		return bi.LoadImageBinaryInfoAgain(n)
+	}
+	hadErrors := false
+	for i := range bi.Images {
+		if bi.Images[i].LoadError() != nil {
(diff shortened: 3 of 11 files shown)


service/debugger/debugger.go
```diff
@@ -2664,10 +2664,31 @@ func (d *Debugger) CancelDownloads() bool {
...
 	defer d.targetMutex.Unlock()
-	return d.target.Selected.BinInfo().LoadImageBinaryInfoAgain(n)
+	if eventsFn != nil {
+		defer eventsFn(&proc.Event{Kind: proc.EventDownloadLibraryInfoDone})
+	}
+	d.target.SetEventsFn(eventsFn)
+	bi := d.target.Selected.BinInfo()
+	bi.ResetDownloadsContext()
+	if n > 0 {
+		return bi.LoadImageBinaryInfoAgain(n)
```
derekparker, line 2677: Isn't this off by one? The server calls `DownloadLibraryDebugInfo(arg.N+1, ...)` to reserve 0 for the download-everything case, but here we hand the shifted `n` straight to `LoadImageBinaryInfoAgain` while the loop below passes the unshifted `i`. So `libraries -d 3` reloads `Images[4]`, `-d 0` isn't reachable at all, and the last valid index quietly no-ops because `n >= len(bi.Images)`. Shouldn't this be `LoadImageBinaryInfoAgain(n - 1)`?

service/debugger/debugger.go
```diff
@@ -2664,10 +2664,31 @@ func (d *Debugger) CancelDownloads() bool {
...
 // DownloadLibraryDebugInfo attempts to download the specified library's debug info.
-func (d *Debugger) DownloadLibraryDebugInfo(n int) error {
+func (d *Debugger) DownloadLibraryDebugInfo(n int, eventsFn func(*proc.Event)) error {
 	d.targetMutex.Lock()
 	defer d.targetMutex.Unlock()
-	return d.target.Selected.BinInfo().LoadImageBinaryInfoAgain(n)
+	if eventsFn != nil {
+		defer eventsFn(&proc.Event{Kind: proc.EventDownloadLibraryInfoDone})
+	}
+	d.target.SetEventsFn(eventsFn)
```
derekparker, line 2673: Shouldn't we clear this again on the way out? `Command` opens with `SetEventsFn(nil)` for this reason. As written the target group keeps pushing into the server's eventsChan after the RPC returns with nobody draining it, so anything emitted outside of a command (breakpoint materialization, a spawned process) sits in the queue and gets handed to the next command's `drainEvents`. If a queued terminator comes out first the client's drain goroutine returns early and we lose the rest of that command's output. I think a `defer d.target.SetEventsFn(nil)` covers it.

service/debugger/debugger.go
```diff
@@ -2664,10 +2664,31 @@ func (d *Debugger) CancelDownloads() bool {
...
+	for i := range bi.Images {
+		if bi.Images[i].LoadError() != nil {
+			err := bi.LoadImageBinaryInfoAgain(i)
+			if err != nil {
+				hadErrors = true
+			}
+		}
+	}
+	if hadErrors {
+		return errors.New("errors downloading debuginfo")
```
derekparker: Won't this return an error almost every time in practice? The loop retries every image with a LoadError, including ones no download will ever fix (linux-vdso.so.1, system libraries with no published debuginfo). Those come back with the "could not read debug info (...) and could not read go symbol table (...)" error, so `hadErrors` gets set even when every download that could have succeeded did. We also throw the individual errors away, so a client can't tell "nothing here was downloadable" from "debuginfod is down". Could we count only the images we actually attempted a download for, and aggregate the real errors into the message?

pkg/proc/bininfo.go
```diff
@@ -1155,8 +1155,18 @@ func (bi *BinaryInfo) AddImage(path string, addr uint64) error {
...
+// again.
+func (bi *BinaryInfo) ResetDownloadsContext() {
+	bi.cancelDownloadsMu.Lock()
+	bi.downloadsCtx, bi.cancelDownloads = context.WithCancel(context.Background())
+	bi.cancelDownloadsMu.Unlock()
+}
+
 // LoadImageBinaryInfoAgain loads the n-th image debug symbols if they weren't already loaded.
 func (bi *BinaryInfo) LoadImageBinaryInfoAgain(n int) error {
+	bi.attaching = false
```
derekparker, line 1169: Won't this clear `attaching` even when the call turns out to be a no-op? It happens before the bounds and loadErr check, so a bad index or an already loaded image still flips it for the rest of the session.

Combined with the automatic `DownloadLibraryDebugInfo(-1)` on attach, that means every attach session ends up with `attaching == false`, and we're back to blocking the target in debuginfod-find for every library loaded by a later continue or dlopen, which is what 3e4f098e was trying to avoid.

Could we at least move it below the early return, and ideally decide this in the debugger layer rather than in here?

pkg/proc/bininfo.go
```diff
@@ -1155,8 +1155,18 @@ func (bi *BinaryInfo) AddImage(path string, addr uint64) error {
 	return err
 }
 
+// ResetDownloadsContext resets the downloads context. If the context was
+// cancelled before this BinaryInfo object will be able to do downloads
+// again.
+func (bi *BinaryInfo) ResetDownloadsContext() {
```
derekparker, line 1161: This only gets called from `Debugger.DownloadLibraryDebugInfo`, so doesn't the automatic path still have the same problem? `AddImage` only builds the context `if bi.cancelDownloads == nil`, so once someone hits ^C during a continue the context stays cancelled and every automatic download after that fails immediately with ErrNoDebugInfoFound, with nothing printed to say why. Since ^C during a continue is the usual way to cancel, that's the case I'd most want covered here.

pkg/terminal/terminal.go
```diff
@@ -154,7 +159,7 @@ func New(client service.Client, conf *config.Config) *Term {
 				}
 				fmt.Fprintf(t.stdout, "Downloading debug info for %s: %s (press ^C to cancel)", event.BinaryInfoDownloadEventDetails.ImagePath, event.BinaryInfoDownloadEventDetails.Progress)
 				firstEventBinaryInfoDownload = false
-			case api.EventStopped:
+			case api.EventStopped, api.EventDownloadLibraryInfoDone:
```
derekparker, line 162: Should this case set `firstEventBinaryInfoDownload` back to true?

pkg/terminal/terminal.go
```diff
@@ -416,6 +421,10 @@ func (t *Term) Run() (int, error) {
 	// making a blocking call.
 	_, _ = t.client.GetState()
 
+	if t.isAttachCmd {
+		t.client.DownloadLibraryDebugInfo(-1)
```
derekparker, line 426: Should we check the error here?

service/test/integration2_test.go
```diff
@@ -3499,6 +3499,93 @@ func TestCancelDownload(t *testing.T) {
 	})
 }
 
+func TestCancelDownloadAfterAttach(t *testing.T) {
+	if testBackend == "rr" {
+		return
```
derekparker: Should this be `t.Skip`?

derekparker (changes requested):

service/debugger/debugger.go
aarzilli: The reason it's like this is that ListDynamicLibraries in rpc2/server.go skips Images[0] (because it isn't a library, it's the executable). The +1 in server.go is to make the indexes match with ListDynamicLibraries.

I didn't really intend to reserve 0 for download everything (you can see that I used -1 below for that), that `if n > 0` thing is mostly because we always load the main executable's debuginfo, we could change it to a `if n >= 0`.

As for the invalid indexes, they all quietly no-op. Maybe we should make them all error.

service/debugger/debugger.go
aarzilli: Somebody setting an events function needs to also fully drain the queue using GetEvents until they see `EventStopped` (for `Command`) or `EventDownloadLibraryDone` (for `DownloadLibraryDebugInfo`). Breakpoint materialization and process spawn can not happen between an `EventStopped` and the next `Command` call.

I'm going to document this explicitly.

service/debugger/debugger.go
aarzilli: Good point. I've changed the error message to have a count of how many libraries we tried to download and how many we succeeded. The individual errors aren't thrown away, they are stored in the loadErr of the library and can be see with the `libraries` command.

pkg/proc/bininfo.go
aarzilli: 3e4f098ee1b8126e97f0afcbacce128cc93766bb was trying to avoid calling debuginfod before we have a server running, can show a prompt to the user, give the option to cancel. At the point we can call `DownloadLibraryDebugInfo` it's fine to call debuginfod, it's infact what we should do (otherwise we'd never call it).

What clearing it here does is it stops the `if hadAttaching` loop in `(*TargetGroup).Continue` from running. That loop is there to try to load the debug info of all libraries that we skipped during the first attach. I put this there to avoid trying the download twice when the client is aware of this and has already sent a `DownloadLibraryDebugInfo(-1)`.

This is the correct behavior for our command line client (since it always sends the `DownloadLibraryDebugInfo(-1)` call) for a different client that sends `DownloadLibraryDebugInfo` only for specific libraries I'm not sure what would be the best thing to do. I think it's still fine, if the client is aware about this it can still send a call with `-1`.

pkg/proc/bininfo.go
aarzilli: For the normal path the context is reset in `(*TargetGroup).Continue`. In our client ^C only calls CancelDownloads if we saw a download event, if the target is still running ^C won't cancel downloads.

pkg/terminal/terminal.go
aarzilli: You are right.

pkg/terminal/terminal.go
aarzilli: IMO no, this is essentially the same as the AddImage call in pkg/proc/linutil/dynamic.go, where we also discard the error. Dropping the error here keeps the behavior the same as it was before. An error loading the debug info here is going to show up in the `libraries` command.

service/test/integration2_test.go
aarzilli: Correct.

service/debugger/debugger.go
derekparker: Ack makes sense, thanks.

derekparker (approved):
LGTM

PR #4439, opened 2026-09-02
dap: add types command to list program types

### LINKED ISSUE #3729: Support `dlv types` from DAP (the same way `types` is supported in the command line `dlv` REPL)
Some Delve error messages ask users to run specific commands, for example when a type is not found (e.g. type assertion `.(any)`), users get a message
> use 'types' for a list of valid types

These error messages are also shown to users of the DAP protocol (e.g. via VSCode). The way to invoke commands via DAP is to send them as `evaluate` request, escaped with `dlv`, like `dlv help`. These are configured here:
https://github.com/go-delve/delve/blob/7c7265f4e68a333bb084217e7b7ae065024e428b/service/dap/command.go#L81

As you can see, only three commands are supported, while command line Delve supports a few dozen. I think it would be good to support at least the ones we reference from error messages, like `types`.

1. What version of Delve are you using (`dlv version`)?  1.22.1
2. What version of Go are you using? (`go version`)?  go1.23-20240419-RC02
3. What operating system and processor architecture are you using? Linux, x86_64

larrasket:
Mirror the existing sources command so `dlv types [<regex>]` works from a DAP session the same way it does in the command-line REPL. The handler delegates to Debugger.Types, which already filters, sorts, and deduplicates its result.

Fixes #3729

### DIFF
--- a/service/dap/command.go
+++ b/service/dap/command.go
@@ -80,6 +80,12 @@ Type "help" followed by the name of a command for more information about it.`
 
 If regex is specified only the source files matching it will be returned.`
 
+	msgTypes = `Print list of types.
+
+	dlv types [<regex>]
+
+If regex is specified only the types matching it will be returned.`
+
 	msgTarget = `Manages child process debugging.
 
         target follow-exec [-on [regex]] [-off]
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4440, opened 2026-09-02
proc: decode AVX-512 ZMM16-ZMM31 from XSAVE

alexsaezm:
previously, zmm16 through zmm31 were ignored. This adds decoding support plus a bunch of synthetic tests.

### DIFF
--- a/pkg/proc/amd64util/xsave.go
+++ b/pkg/proc/amd64util/xsave.go
@@ -13,11 +13,13 @@ import (
 // Manual, Volume 1: Basic Architecture.
 type AMD64Xstate struct {
 	AMD64PtraceFpRegs
-	Xsave       []byte // raw xsave area
-	AvxState    bool   // contains AVX state
-	YmmSpace    [256]byte
-	Avx512State bool // contains AVX512 state
-	ZmmSpace    [512]byte
+	Xsave        []byte // raw xsave area
+	AvxState     bool   // contains AVX state
+	YmmSpace     [256]byte
+	Avx512State  bool // contains AVX512 state
+	ZmmSpace     [512]byte
+	Hi16ZmmState bool // contains ZMM16 through ZMM31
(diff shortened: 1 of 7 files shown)


pkg/proc/amd64util/xsave_test.go
```diff
@@ -0,0 +1,113 @@
+package amd64util
```
aarzilli, line 1: These tests all seem very circular, to me. I'm not sure there's any value to having them.

pkg/proc/amd64util/xsave.go
```diff
@@ -118,19 +127,13 @@ func AMD64XstateRead(xstateargs []byte, readLegacy bool, regset *AMD64Xstate, xs
 		return nil
 	}
 
-	if !xstate_bv.hasAVX() {
```
aarzilli, line 121: What's the motivation for rewriting this without the early returns? That some kernel could hypothetically pass weird options to XSAVE so that we have ZMM16 through 31 without having ZMM0 through 15?

alexsaezm: I deleted my previous answer because it was not mine at all. I was trying to write a test locally to see if I can really justify this changed (because in my mind makes sense that somehow you could write to 16 and not to 15 for example) and after not being really able to come with a proof I decided to try Cursor out of curiosity. First and last time I use it: It decided to use the `gh` command line to post a comment on my behalf.

Shame on me for using this crap. Will come with a better answer soon, and mine.

pkg/proc/amd64util/xsave_test.go
alexsaezm: They are circular for sure. I created them because I started this PR on a laptop that doesn't have AVX so I thought that maybe this was a good approach, and also because what if the CI don't have AVX, but they are pretty much useless I think.

PR #4444, opened 2026-09-03
gobuild: keep -C first for test builds

### LINKED ISSUE #4443: gobuild: dlv test places -c before a leading -C flag

Delve 1.27.1

go1.25.5

macOS Darwin arm64

### Description

Delve should invoke Go with `-C` as the first flag, for example:

    `go test -C . -c ...`

The test binary should be built and the debug session should start.

What did you see instead?

Delve constructs the command in this order:

    `go test -c -C . ...`

Go rejects this command with:

    `-C` flag must be first flag on command line

The same argument ordering problem exists for both string and []string buildFlags.

cuishuang:
Fixes https://github.com/go-delve/delve/issues/4443

Keeping a leading -C argument before Delve-generated flags when building test binaries.

The fix applies to both string and []string buildFlags. In test mode, `-c` is now added after the leading `-C` argument, so Delve generates:

    `go test -C <dir> -c ...`


aarzilli (approved):
LGTM

derekparker (approved):
LGTM

PR #4445, opened 2026-09-08
goversion: format versioned development builds correctly

cuishuang:
`GoVersion.String` formats versioned development builds as beta releases because `versionedDevel` is less than `betaStart`.

For example, a development build parsed from `go1.25-devel_...` was formatted as:

    go1.25beta-1000

Handle `versionedDevel` before the beta branch and format it as:

    devel go1.25

Add round-trip and direct formatting tests for versioned development builds.


derekparker (approved):
LGTM

PR #4449, opened 2026-09-12
proc,terminal,service: add timeout to expression evaluation

aarzilli:
Adds a timeout to expression evaluation, applies it to map access and
map load, which currently can run in unbounded time. Defaults the
timeout to 100ms.


derekparker (changes requested):
Can you add a test exercising this somehow? Also, should we check the timeout within `nextBucket` as well?

aarzilli:
> Can you add a test exercising this somehow? Also, should we check the timeout within `nextBucket` as well?

Done.

derekparker (approved):
LGTM

PR #4447, opened 2026-09-09
proc: read from sigtrampgo for ppc64le

alexsaezm:
Also fixes TestCgoStacktrace2 in ppc64le

PR #4451, opened 2026-09-17
Fix TestCGlobal failure 

### LINKED ISSUE #4423: Build failure on ppc64le

1.27.1

1.26.5

Debian

### Description

Hi,

while updating delve in Debian to the latest version, I've noticed that delve shows a new build failure on ppc64le between 1.26 and 1.27. The failing test is `TestCGlobal`, introduced in PR #4358.

The build logs can be seen here: https://buildd.debian.org/status/logs.php?pkg=delve&arch=ppc64el

For now, I've disabled the test on ppc64le, but maybe this is a real regression which you might want to investigate.

Regards,
Tobias

archanaravindar:
The TestCGlobal fails on ppc64le due to the incorrect global variable resolution. Delve places breakpoints at the function global entry point. This address contains a TOC set up stub which is skipped in case of same module callers that enter at local entry point or LEP instead. Hence the breakpoint never got hit. As a result an incorrect scope resulted when the variable in question was accessed in this case function C.fortytwo and variable globalvar. 
LEP = GEP + offset encoded in bits 5-7 of symbols st_other byte (this holds good in any ELF V2 binary) 
The fix is to add a localEntry helper that reads the elf symbol from SymNames and decodes the offset and clamps the FirstPCAfterPrologue forward to the LEP when the computed PC falls before it 
While implementing this encountered another issue- loadSymbolName was using a whole byte st_info == _STT_FUNC comparison which worked only for local functions and dropped global ones like cgo exported C.fortytwo hence had to be changed to elf.ST_TYPE(info)==_STT_FUNC since this adjustment is needed only for ppc64le gated the change for ppc64le only. 
TestCgoStackTrace will be fixed along with TestCgoStackTrace2 in https://github.com/go-delve/delve/pull/4447 as both fixes are related 
Fixes #4423


pkg/proc/bininfo.go
```diff
@@ -414,11 +414,44 @@ func FirstPCAfterPrologue(p Process, fn *Function, sameline bool) (uint64, error
 		// breakpoint with file:line and with the function name always result on
 		// the same instruction being selected.
 		if pc2, _, _, ok := fn.cu.lineInfo.FirstStmt(fn.Entry, fn.End); ok {
-			return pc2, nil
+			pc = pc2
```
aarzilli, line 417:
```
if bi.Arch.Name == "ppc64le" {
     pc2 = ppc64leSkipToLocalEntry(bi, fn, pc)
}
return pc2, nil
```

pkg/proc/bininfo.go
```diff
@@ -414,11 +414,44 @@ func FirstPCAfterPrologue(p Process, fn *Function, sameline bool) (uint64, error
...
+// with a TOC (r2) setup stub. However, Callers in the same module (for example
+// the cgo trampoline calling a C function) already share the TOC and enter at
+// the local entry point. A breakpoint placed at global entry point never executes.
+// gcc does not emit DW_LNS_set_prologue_end, so PrologueEndPC cannot advance past
+// the stub on its own; use the local entry offset recorded in the ELF symbol
+// instead.
+//
+// On other architectures, for functions without a local entry offset, or when
+// pc is already past the local entry point, this becomes a no-op.
+func (bi *BinaryInfo) localEntry(fn *Function, pc uint64) uint64 {
```
aarzilli, line 438: I think this should be renamed to something like `ppc64leSkipToLocalEntry` to make it clear that it is an architecture specific thing.

pkg/proc/bininfo.go
```diff
@@ -414,11 +414,44 @@ func FirstPCAfterPrologue(p Process, fn *Function, sameline bool) (uint64, error
...
 
-	return pc, nil
+	return p.BinInfo().localEntry(fn, pc), nil
+
+}
+
+// localEntry returns the ppc64le ELFv2 local entry point of fn, clamping pc
+// forward to it when pc falls inside the function's global-entry stub.
+//
+// On ppc64le DWARF low_pc (fn.Entry) is the global entry point, which begins
```
aarzilli, line 428: This needs references.

pkg/proc/bininfo.go
```diff
@@ -1874,9 +1907,18 @@ func (bi *BinaryInfo) loadSymbolName(image *Image, file *elf.File, wg *sync.Wait
...
+	// match on the type bits of st_info. Comparing the whole byte to _STT_FUNC
+	// (as done on other arches) keeps only local functions (st_info 0x02) and
+	// drops global ones (st_info 0x12). Other arches keep the existing behavior.
+	ppc64le := bi.Arch.Name == "ppc64le"
 	symSecs, _ := file.Symbols()
 	for _, symSec := range symSecs {
-		if symSec.Info == _STT_FUNC { // TODO(chainhelen), need to parse others types.
+		isFunc := symSec.Info == _STT_FUNC // TODO(chainhelen), need to parse others types.
+		if ppc64le {
+			isFunc = elf.ST_TYPE(symSec.Info) == _STT_FUNC
```
aarzilli, line 1919: Shouldn't we always call `elf.ST_TYPE`?

pkg/proc/bininfo.go
archanaravindar: Done.

pkg/proc/bininfo.go
archanaravindar: Done.

pkg/proc/bininfo.go
archanaravindar: Added, thanks.

pkg/proc/bininfo.go
archanaravindar: Ack.

pkg/proc/bininfo.go
aarzilli: I meant keep the early return.

pkg/proc/bininfo.go
```diff
@@ -1876,7 +1914,12 @@ func (bi *BinaryInfo) loadSymbolName(image *Image, file *elf.File, wg *sync.Wait
 	}
 	symSecs, _ := file.Symbols()
 	for _, symSec := range symSecs {
-		if symSec.Info == _STT_FUNC { // TODO(chainhelen), need to parse others types.
+		// Match on the type bits of st_info. Comparing the whole byte to
```
aarzilli: I think this comment is superfluous.

archanaravindar: replaced with a one liner to explain why we added elf.ST_TYPE(symSec.Info)

aarzilli (approved):
LGTM

derekparker (approved):
LGTM

